PR #43, opened 2018-07-05
Generate new DeclRefExpr's instead of modifying old ones for differentiate() and gradient() args.

efremale:
Currently, after the derivative/gradient function was generated, we put it as a first argument of differentiate()/gradient() by modifying the old argument (which is DeclRefExpr, referencing the old function) by calling setDecl() on it.

A better solution is to create entirely new DeclRefExpr and set is as a first argument. We do not need to pass any placeholders anymore.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -6,14 +6,7 @@
 using namespace clang;
 
 namespace clad {
-  // FIXME:
-  // Currently, the global variable is used to store the pointer to the last
-  // call which requires update.
-  // This is needed to share this information between different invocations of
-  // CladPlugin::HandleTopLevelDecl in which clad::gradient and make_gradient
-  // are processed (see Differentiator.h).
-  clang::CallExpr* global_GradientToUpdate = nullptr;
-
+  static SourceLocation noLoc;
   FunctionDeclInfo::FunctionDeclInfo(FunctionDecl* FD, ParmVarDecl* PVD)
     : m_FD(FD), m_PVD(PVD) {
 #ifndef NDEBUG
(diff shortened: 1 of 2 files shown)


vgvassilev:
Could you remove the merge commit?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -42,27 +34,64 @@ namespace clad {
...
-             = dyn_cast<UnaryOperator>(call->getArg(0))){
-      DRE = dyn_cast<DeclRefExpr>(UnOp->getSubExpr());
+    else if (auto UnOp = dyn_cast<UnaryOperator>(call->getArg(0))){
+      oldDRE = dyn_cast<DeclRefExpr>(UnOp->getSubExpr());
     }
-    // Set the new declaration.
-    if (DRE)
-      DRE->setDecl(FD);
+    else 
+      assert(false && "Trying to differentiate something unsupported");
```
vgvassilev: We should use `llvm_unreachable` instead of `assert(0);`

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -42,27 +34,64 @@ namespace clad {
...
+    
+    ASTContext& C = SemaRef.getASTContext();
+    // Create ref to generated FD.
+    auto DRE = DeclRefExpr::Create(
+      C,
+      oldDRE->getQualifierLoc(), // Text before the FD name, e.g. it is
```
vgvassilev: Could you reword this comment (I do not get it) and maybe move it before the variable DRE definition.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -42,27 +34,64 @@ namespace clad {
...
+    else if (auto oldUnOp = dyn_cast<UnaryOperator>(oldArg)) {
+      auto newUnOp = SemaRef.BuildUnaryOp(
+        nullptr,
+        SourceLocation(),
+        oldUnOp->getOpcode(),
+        DRE).get();
+      call->setArg(0, newUnOp);
+    } 
+    else 
+      assert(false && "Trying to differentiate something unsupported");
```
vgvassilev: Likewise.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -42,27 +34,64 @@ namespace clad {
...
-    // Set the new declaration.
-    if (DRE)
-      DRE->setDecl(FD);
+    else 
+      assert(false && "Trying to differentiate something unsupported");
+    
+    ASTContext& C = SemaRef.getASTContext();
+    // Create ref to generated FD.
+    auto DRE = DeclRefExpr::Create(
+      C,
```
vgvassilev: Can you move the first argument next to the opening paren and align everything else by it? Please do that for all indented calls.

lib/Differentiator/DiffPlanner.cpp
efremale: It's just saying that we need to use the same nested-name-qualifier (whatever is before the ::) as in the oldDRE. It will fail if we pass empty qualifier.

PR #48, opened 2018-07-06
Support differentiation of recursive functions

efremale:
Now clad::differentiate should produce correct results on recursive functions.
Any recursive call should be replaced by a call to a derivative.


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -447,6 +449,23 @@ namespace clad {
       return NodeContext(OverloadedDerivedFn);
     }
 
+    if (CE->getDirectCallee() == m_Function) {
+      // The differentiated function is called recursively.
+      auto derivativeRef = m_Sema.BuildDeclarationNameExpr(
```

PR #51, opened 2018-07-07
Store intermediate results in temporary variables

efremale:
This commit adds functionality to store intermediate derivatives in variables, so that the same expressions are not evaluated multiple times.

Currently, output of both differentiation modes is equivalent of that of symbolic differentiation. Resulting formulas often require to evaluate same expressions multiple times and can be exponentially longer than the original function.

Implemented functionality allows to (for now partially) reduce the size of the formula by storing intermediate results in variables. 

For example:
```
double f(double x, double y, double z) {
  return 2*(x + y + z)*x*y*z;
}
...
clad::gradient(f)
```

Old result would be:
```
void f_grad(double x, double y, double z, double *_result) {
    _result[0UL] += 2 * 1. * z * y * x;
    _result[1UL] += 2 * 1. * z * y * x;
    _result[2UL] += 2 * 1. * z * y * x;
    _result[0UL] += 2 * (x + y + z) * 1. * z * y;
    _result[1UL] += 2 * (x + y + z) * x * 1. * z;
    _result[2UL] += 2 * (x + y + z) * x * y * 1.;
}
```
Note that, e.g., expression `2 * 1. * (z * y * x)` has to be evaluated three times.

After this commit, the output is:
```
void f_grad(double x, double y, double z, double *_result) {
    double _t1 = 1. * z;
    double _t2 = _t1 * y;
    double _t3 = _t2 * x;
    double _t4 = _t3 * (x + y + z);
    double _t5 = 2 * _t3;
    _result[0UL] += _t5;
    _result[1UL] += _t5;
    _result[2UL] += _t5;
    double _t6 = 2 * (x + y + z) * _t2;
    _result[0UL] += _t6;
    double _t7 = 2 * (x + y + z) * x * _t1;
    _result[1UL] += _t7;
    double _t8 = 2 * (x + y + z) * x * y * 1.;
    _result[2UL] += _t8;
}
```

Now the result of the same expression is stored in the variable `_t5` and needs not to be recomputed multiple times.

There are ways to optimize it further. For example, the variable `_t4` is unused (it is the derivative of f w.r.t. constant 2 and it does not end up in the result). It is possible to eliminate such variables.

Additionally, we sometimes still recompute subexpressions of original function, e.g., `2 * (x + y + z)` multiple times. This too can be optimized by preprocessing the function and storing results of its subexpressions in variables.

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -64,9 +64,8 @@ namespace clad {
 
   }
 
-  FunctionDecl* DerivativeBuilder::Derive(
-    FunctionDeclInfo& FDI,
-    const DiffPlan& plan) {
+  FunctionDecl* DerivativeBuilder::Derive(FunctionDeclInfo& FDI,
+                                          const DiffPlan& plan) {
     FunctionDecl* result = nullptr;
     if (plan.getMode() == DiffMode::forward) {
       ForwardModeVisitor V(*this);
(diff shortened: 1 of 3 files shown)


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -645,6 +701,33 @@ namespace clad {
     return m_Builder.Clone(e);
   }
 
+  Expr* ReverseModeVisitor::StoreAndRef(clang::Expr* E, const char * prefix) {
+#ifdef CLAD_STORE_TEMPORARIES
```
efremale: It is currently disabled by default because it is still experimental and would break old tests.

vgvassilev:
I am a little uneasy with this change. I'd expect the LLVM optimizer to handle these cases. Do you have an example where the LLVM IR is not optimized?

PR #52, opened 2018-07-07
Add differentiation of function calls in reverse mode

### LINKED ISSUE #40: Recursive functions are not differentiated correctly
A call to itself inside a function body is not treated correctly.

For example, if we differentiate
```
double r(double x) {
  if (x > 0)
    return r(x-1);
  else
    return x;
```
the call r(x-1) is treated as a call to some other function and we get the warning: **function 'r' was not differentiated because it is not declared in namespace 'custom_derivatives' attempted differention of function 'r', which does not have a definition**. 

The following derivative is generated:
```
double r_darg0(double x) {
  if (x > 0)
    return r(x-1);
  else
    return 1.0;
```
but the correct derivative would be
```
double r_darg0(double x) {
  if (x > 0)
    return r_darg0(x-1) * 1.0;
  else
    return 1.0;
```

efremale:
Now we can differentiate function calls in reverse mode, in a similar manner as it is done in forward mode.
For functions with a single parameter, we look for a `f_darg0` function in `custom_derivatives` namespace,
for more parameters we look for `f_grad`.

Now reverse mode should support everything that was also implemented in forward mode.

Some bugs were also fixed.

Fixes #40.

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -997,7 +997,136 @@ namespace clad {
   }
   
   void ReverseModeVisitor::VisitCallExpr(const CallExpr* CE) {
-    llvm_unreachable("calling functions is not supported yet");
+    auto FD = CE->getDirectCallee();
+    if (!FD) {
+       unsigned unsupported_call
+         = m_Sema.Diags.getCustomDiagID(DiagnosticsEngine::Warning,
+                                        "attempted differentiation of something"
+                                        " that is not a direct call to a"
+                                        " function and is not supported yet." 
+                                        " Ignored.");
+        m_Sema.Diag(noLoc, unsupported_call);
+       return;
+    }
+    IdentifierInfo* II = nullptr;
+    auto NArgs = FD->getNumParams();
+    // If the function has no args then we assume that it is not related
+    // to independent variables and does not contribute to gradient.
+    if (!NArgs)
+      return;
+
+    llvm::SmallVector<Expr*, 16> CallArgs(CE->getNumArgs());
+    std::transform(CE->arg_begin(), CE->arg_end(), std::begin(CallArgs),
+      [this](const Expr* Arg) { return Clone(Arg); });
+
+    VarDecl* ResultDecl = nullptr;
+    Expr* Result = nullptr;
+    // If the function has a single arg, we look for a derivative w.r.t. to 
+    // this arg (it is unlikely that we need gradient of a one-dimensional'
+    // function).
+    if (NArgs == 1)
+      II = &m_Context.Idents.get(FD->getNameAsString() + "_darg0");
+    // If it has more args, we look for its gradient.
+    else {
+      II = &m_Context.Idents.get(FD->getNameAsString() + "_grad");
+      // We also need to create an array to store the result of gradient call.
+      auto size_type_bits = m_Context.getIntWidth(m_Context.getSizeType());    
+      auto ArrayType =
+        m_Context.getConstantArrayType(CE->getType(),
+                                       llvm::APInt(size_type_bits, NArgs),
+                                       ArrayType::ArraySizeModifier::Normal,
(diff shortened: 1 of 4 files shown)


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -1089,17 +1224,21 @@ namespace clad {
       //xi = xl / xr
       //dxi/xl = 1 / xr
       //df/dxl += df/dxi * dxi/xl = df/dxi * (1/xr)
-      auto one = ConstantFolder::synthesizeLiteral(R->getType(), m_Context, 1.0);
       auto clonedR = Clone(R);
-      auto dl = BuildOp(BO_Div, one, clonedR);
+      auto dl = BuildOp(BO_Div, dfdx(), clonedR);
```
efremale: This was missing, bug fixed.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -1089,17 +1224,21 @@ namespace clad {
...
       auto dlTmp = StoreAndRef(dl);
       Visit(L, dlTmp);
       //dxi/xr = -xl / (xr * xr)
       //df/dxl += df/dxi * dxi/xr = df/dxi * (-xl /(xr * xr))
-      auto RxR = BuildOp(BO_Mul, clonedR, clonedR);
+      auto RxR = m_Sema.ActOnParenExpr(
+        noLoc, noLoc,
+        BuildOp(BO_Mul, clonedR, clonedR)).get();
```
efremale: Parentheses were necessary, otherwise it generates code like: `1 / R * R`

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -1089,17 +1224,21 @@ namespace clad {
...
       auto dr = BuildOp(
-        UO_Minus,
-        BuildOp(BO_Div, Clone(L), RxR));
+        BO_Mul,
+        dfdx(),
+        BuildOp(
+          UO_Minus,
+          BuildOp(BO_Div, Clone(L), RxR)));
```
efremale: Multiplication by `dfdx()` was missing, bug fixed.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -1026,7 +1029,139 @@ namespace clad {
...
+      // We also need to create an array to store the result of gradient call.
+      auto size_type_bits = m_Context.getIntWidth(m_Context.getSizeType());    
+      auto ArrayType = m_Context.getConstantArrayType(
+        CE->getType(),
+        llvm::APInt(size_type_bits, NArgs),
+        ArrayType::ArraySizeModifier::Normal,
+        0); // No IndexTypeQualifiers
+
+      // Create {} array initializer to fill it with zeroes.
+      // FIXME: is there better way to create array initializers?
```
vgvassilev: `clang::Sema::ActOnInitList`?

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -1026,7 +1029,139 @@ namespace clad {
...
+        // Not a recursive call, derivative was not found, ignore.
+        // Issue a warning.
+        SourceLocation IdentifierLoc = CE->getDirectCallee()->getLocEnd();
+        unsigned warn_function_not_declared_in_custom_derivatives
+          = m_Sema.Diags.getCustomDiagID(DiagnosticsEngine::Warning,
+                                     "function '%0' was not differentiated "
+                                     "because it is not declared in namespace "
+                                     "'custom_derivatives'");
+        m_Sema.Diag(IdentifierLoc, warn_function_not_declared_in_custom_derivatives)
+          << FD->getNameAsString();
```

lib/Differentiator/DerivativeBuilder.cpp
vgvassilev: Can you turn that into a comment in the source code?

PR #54, opened 2018-07-08
Add differentiation of intermediate variables declarations in forward mode

efremale:
Now we can differentiate functions with intermediate declarations and
keep track of how this declarations are related to the independent
variable.

Example:
```
double f(double x) {
  double t = x * x;
  return t;
}
...
clad::differentiate(f, 0);
```
will result in:
```
double f_darg0(double x) {
    double t = x * x;
    return (1. * x + x * 1.);
}
```

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -225,6 +225,12 @@ namespace clad {
         m_Sema.IdResolver.AddDecl(newPVD);
       }
     }
+    /// Store dx/dx = 1 in the variables map.
+    m_Variables[m_IndependentVar] =
+      ConstantFolder::synthesizeLiteral(m_IndependentVar->getType(),
+                                        m_Context,
+                                        1);
+
     llvm::ArrayRef<ParmVarDecl*> paramsRef
       = llvm::makeArrayRef(params.data(), params.size());
     derivedFD->setParams(paramsRef);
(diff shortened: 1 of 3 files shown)


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -350,9 +356,16 @@ namespace clad {
   NodeContext ForwardModeVisitor::VisitDeclRefExpr(const DeclRefExpr* DRE) {
     DeclRefExpr* clonedDRE = Clone(DRE).getAs<DeclRefExpr>();
     QualType Ty = DRE->getType();
-    if (clonedDRE->getDecl() == m_IndependentVar)
-      // Return 1 literal if this is the independent variable.
-      return ConstantFolder::synthesizeLiteral(Ty, m_Context, 1);
+    if (auto VD = dyn_cast<VarDecl>(clonedDRE->getDecl())) {
+      // If DRE references a variable, try to find if we know something about
+      // how it is related to the independent variable.
+      auto it = m_Variables.find(VD);
```
vgvassilev: Can we move this in the if-stmt?

test/FirstDerivative/Variables.C
```diff
@@ -0,0 +1,76 @@
+// RUN: %cladclang %s -lm -I%S/../../include -oVariables.out 2>&1 | FileCheck %s
```
vgvassilev: Should we run the Variables.out?

PR #56, opened 2018-07-23
Updating installation instruction for Clad

oshadura:


lib/Differentiator/CMakeLists.txt
```diff
@@ -91,8 +91,10 @@ add_llvm_library(cladDifferentiator
   ${version_inc}
   )
 
-#target_link_libraries(cladDifferentiator PRIVATE
-#    clangAST
-#    clangBasic
-#    clangSema
-#  )
+if(LLVM_ENABLE_PLUGINS)
```
vgvassilev: We should use if (TARGET clangAST) to disambiguate if we were in a from-source build.

PR #65, opened 2018-07-27
Fix clang debug assert failures on codegen

efremale:
Two problems were: 

```
bool _t0 = x < y;
return (_t0 ? x : y );
```

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -904,7 +904,12 @@ namespace clad {
 
   void ReverseModeVisitor::VisitConditionalOperator(
     const clang::ConditionalOperator* CO) {
-    auto cond = StoreAndRef(Clone(CO->getCond()));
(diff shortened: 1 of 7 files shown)


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -904,7 +904,12 @@ namespace clad {
 
   void ReverseModeVisitor::VisitConditionalOperator(
     const clang::ConditionalOperator* CO) {
-    auto cond = StoreAndRef(Clone(CO->getCond()));
+    auto condVar = StoreAndRef(Clone(CO->getCond()));
+    auto cond =
+      m_Sema.ActOnCondition(m_CurScope.get(),
+                            noLoc,
+                            condVar,
+                            Sema::ConditionKind::Boolean).get().second;
```
efremale: This should generate all the required casts (i.e. missing Lvalue to Rvalue cast).

vgvassilev:
Superseded by PR #66. Closing.

PR #66, opened 2018-07-27
Fix assert failures in clang codegen.

vgvassilev:
Two problems were: 

```
bool _t0 = x < y;
return (_t0 ? x : y );
```

NOTE: This is PR #65 rebased + trailing whitespaces removed.


vgvassilev:
Superseded by #69

PR #68, opened 2018-07-29
Fix assert failures in clang codegen.

efremale:
Two problems were: 

```
bool _t0 = x < y;
return (_t0 ? x : y );
```

Some trailing whitespaces are removed.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -2,6 +2,7 @@
 
 #include "clang/AST/ASTContext.h"
 #include "clang/Sema/Sema.h"
+#include "clang/Sema/TemplateDeduction.h"
 
 using namespace clang;
 
@@ -27,7 +28,7 @@ namespace clad {
       m_FD->dump();
     else
       llvm::errs() << "<invalid> FD: " << m_FD << '\n';
-    
+
     if (m_PVD)
       m_PVD->dump();
     else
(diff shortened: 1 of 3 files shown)


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -128,6 +129,45 @@ namespace clad {
...
+    auto CladGradientFDeclOld = call->getDirectCallee();
+    auto CladGradientExprOld = call->getCallee();
+    auto CladGradientFTemplate = CladGradientFDeclOld->getPrimaryTemplate();
+
+    FunctionDecl* CladGradientFDeclNew = nullptr;
+    sema::TemplateDeductionInfo info(noLoc);
+    // Create/get template specialization of clad::gradient that matches
+    // argument types. Result is stored to CladGradientFDeclNew.
+    SemaRef.DeduceTemplateArguments(CladGradientFTemplate,
+                                    nullptr,
```
vgvassilev: Could you specify the the parameter name in front eg /* Scope */.  Makes it much more readable saving the reviewer one hop to look this up.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -128,6 +129,45 @@ namespace clad {
...
+                                    [] (llvm::ArrayRef<QualType>) {
+                                      return false;
+                                    });
+    // DeclRefExpr for new specialization.
+    auto CladGradientExprNew =
+      SemaRef.BuildDeclRefExpr(CladGradientFDeclNew,
+                               CladGradientFDeclNew->getType(),
+                               CladGradientExprOld->getValueKind(),
+                               CladGradientExprOld->getLocEnd()).get();
+    auto CastOld = dyn_cast<ImplicitCastExpr>(CladGradientExprOld);
```
vgvassilev: Do you expect CastOld to be nullité? If not use just cast instead of dyn_cast.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -128,6 +129,45 @@ namespace clad {
...
+    // Function to pointer cast.
+    CladGradientExprNew =
+      ImplicitCastExpr::Create(C,
```
vgvassilev: Can’t we delegate the generation of the implicit casts to a higher level sema routine?

Issue #47, opened 2018-07-06
Reduce cloning complexity in forward mode


efremale:
Each call to Clone() has complexity of O(n), where n is the size of the Stmt's subtree, since it recursively calls Clone() on every substatement until leaves are reached.

When differentiating a function via Forward/ReverseModeVisitor, we potentially call Clone() on every node of the AST tree (which recursively clones all subnodes, we are cloning same nodes multiple times) and have the total complexity of O(n^2).

Wouldn't it be more efficient to Clone() the whole tree just once before calling Visit() on it? The complexity should be just 2xO(n).

vgvassilev:
@efremale can we now close this?

efremale:
> @efremale can we now close this?

I believe some work is still required in the reverse mode.

vgvassilev:
Ok, that's #94.

PR #80, opened 2018-08-12
Store reused expressions in intermediate variables instead of reevaluating them in forward mode. Remove unnecessary cloning in forward mode. Refactor DerivativeBuilder.cpp

efremale:
Forward mode now supports storing expression results in intermediate
variables. This allows to avoid reevaluation if derived expressions are
used on several places. For example, for:
```
double t = std::sin(x) * std::cos(x);
```
We now produce:
```
double _t0 = std::sin(x);
double _t1 = std::cos(x);
double _d_t = sin_darg0(x) * (_d_x) * _t1 + _t0 * cos_darg0(x) * (_d_x);
double t = _t0 * _t1;
```
Instead of:
```
double _d_t = sin_darg0(x) * (_d_x) * cos(x) + sin(x) * cos_darg0(x) * (_d_x);
double t = sin(x) * cos(x);
```
which allows us to avoid reevaluation of sin and cos.

This commit solves the issue #47 in case of forward mode.
For example, when deriving BinaryOperator, instead of cloning its whole
tree and visiting it (potentially cloning the subtrees again), e.g.
```
auto ClonedBO = Clone(BO); // both LHS and RHS are cloned
auto LHS = ClonedBO->getLHS();
auto LHSDiff = Visit(LHS) // LHS will be cloned inside again
auto RHS = ClonedBO->getRHS();
auto RHSDiff = Visit(RHS);
auto BODiff = ...// Build BODiff from LHSDiff and RHSDiff
```
we first visit the whole tree and rebuild the original operation from
the results of subtrees, e.g.
```
auto [LHS, LHSDiff] = Visit(BO->getLHS());
auto [RHS, RHSDiff] = Visit(BO->getRHS());
auto ClonedBO = BuildOp(OpKind, LHS, RHS);
auto BODiff = ...// Build BODiff from LHSDiff and RHSDiff.
```

Some refactoring was applied to DerivativeBuilder.cpp to support new way
of visiting. Some functions of DerivativeBuilder and Visitors were moved
to VisitorBase.

### DIFF
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -144,11 +144,11 @@ namespace clad {
                                  clang::Expr* Init = nullptr,
                                  bool DirectInit = false);
     /// Wraps a declaration in DeclStmt.
-    clang::Stmt* BuildDeclStmt(clang::Decl* D);
-    clang::Stmt* BuildDeclStmt(llvm::MutableArrayRef<clang::Decl*> DS);
+    clang::DeclStmt* BuildDeclStmt(clang::Decl* D);
+    clang::DeclStmt* BuildDeclStmt(llvm::MutableArrayRef<clang::Decl*> DS);
 
     /// Builds a DeclRefExpr to a given Decl.
-    clang::Expr* BuildDeclRef(clang::VarDecl* D);
+    clang::DeclRefExpr* BuildDeclRef(clang::VarDecl* D);
 
     /// Stores the result of an expression in a temporary variable (of the same
     /// type as is the result of the expression) and returns a reference to it.
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -168,14 +168,15 @@ namespace clad {
     /// Shorthand to issues a warning or error.
     template <std::size_t N>
     void diag(clang::DiagnosticsEngine::Level level, // Warning or Error
+              clang::SourceLocation loc,
               const char (&format)[N],
-              llvm::ArrayRef<llvm::StringRef> args,
-              clang::SourceLocation loc = {});
-
-    template <std::size_t N>
-    void diag(clang::DiagnosticsEngine::Level level,
-              const char (&format)[N],
-              clang::SourceLocation loc = {});
+              llvm::ArrayRef<llvm::StringRef> args = {}) {
+      unsigned diagID
+        = m_Sema.Diags.getCustomDiagID(level, format);
+      clang::Sema::SemaDiagnosticBuilder stream = m_Sema.Diag(loc, diagID);
+      for (auto arg : args)
+        stream << arg;
+    }
 
     /// Conuter used to create unique identifiers for temporaries
     std::size_t m_tmpId = 0;
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -8,8 +8,8 @@
 #define CLAD_DERIVATIVE_BUILDER_H
 
 #include "clang/AST/RecursiveASTVisitor.h"
-#include "clang/Sema/Sema.h"
 #include "clang/AST/StmtVisitor.h"
+#include "clang/Sema/Sema.h"
 
 #include <array>
 #include <stack>
@@ -104,11 +104,11 @@ namespace clad {
     }
 
     /// Get the latest block of code (i.e. place for statements output).
-    Stmts & getCurrentBlock() {
+    Stmts& getCurrentBlock() {
       return m_Blocks.top();
     }
     /// Create new block.
-    Stmts & beginBlock() {
+    Stmts& beginBlock() {
       m_Blocks.push({});
       return m_Blocks.top();
     }
@@ -169,7 +169,12 @@ namespace clad {
     template <std::size_t N>
     void diag(clang::DiagnosticsEngine::Level level, // Warning or Error
               const char (&format)[N],
-              llvm::ArrayRef<llvm::StringRef> args = {},
+              llvm::ArrayRef<llvm::StringRef> args,
+              clang::SourceLocation loc = {});
+
+    template <std::size_t N>
+    void diag(clang::DiagnosticsEngine::Level level,
+              const char (&format)[N],
               clang::SourceLocation loc = {});
 
     /// Conuter used to create unique identifiers for temporaries
@@ -211,7 +216,7 @@ namespace clad {
       return llvm::cast_or_null<clang::Expr>(getStmt_dx());
     }
     // Stmt_dx goes first!
-    std::array<clang::Stmt*, 2> & getBothStmts() {
+    std::array<clang::Stmt*, 2>& getBothStmts() {
       return data;
     }
   };
@@ -229,7 +234,7 @@ namespace clad {
     clang::VarDecl* getDecl() { return data[1]; }
     clang::VarDecl* getDecl_dx() { return data[0]; }
     // Decl_dx goes first!
-    std::array<clang::VarDecl*, 2> & getBothDecls() {
+    std::array<clang::VarDecl*, 2>& getBothDecls() {
       return data;
     }
   };
@@ -310,7 +315,7 @@ namespace clad {
     ///
     ///\returns The gradient of the function
     ///
-    clang::FunctionDecl* Derive(FunctionDeclInfo & FDI, const DiffPlan& plan);
+    clang::FunctionDecl* Derive(FunctionDeclInfo& FDI, const DiffPlan& plan);
     void VisitCompoundStmt(const clang::CompoundStmt* CS);
     void VisitIfStmt(const clang::IfStmt* If);
     void VisitReturnStmt(const clang::ReturnStmt* RS);
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -8,8 +8,10 @@
 #define CLAD_DERIVATIVE_BUILDER_H
 
 #include "clang/AST/RecursiveASTVisitor.h"
+#include "clang/Sema/Sema.h"
 #include "clang/AST/StmtVisitor.h"
 
+#include <array>
 #include <stack>
 #include <unordered_map>
 
@@ -33,56 +35,9 @@ namespace clad {
 }
 
 namespace clad {
-  class NodeContext {
-  private:
-    typedef llvm::SmallVector<clang::Stmt*, 2> Statements;
-    Statements m_Stmts;
-    NodeContext() {};
-  public:
-    NodeContext(clang::Stmt* s) { m_Stmts.push_back(s); }
-    
-    NodeContext(clang::Stmt* s0, clang::Stmt* s1) {
-      m_Stmts.push_back(s0);
-      m_Stmts.push_back(s1);
-    }
-    
-    //NodeContext(llvm::ArrayRef) : m_Stmt(s) {}
-    
-    bool isSingleStmt() const { return m_Stmts.size() == 1; }
-    
-    clang::Stmt* getStmt() {
-      assert(isSingleStmt() && "Cannot get multiple stmts.");
-      return m_Stmts.front();
-    }
-    
-    //FIXME: warning: all paths through this function will call itself
-    const clang::Stmt* getStmt() const {
-      return const_cast<NodeContext*>(this)->getStmt();
-    }
-    
-    const Statements& getStmts() const { return m_Stmts; }
-    
-    clang::CompoundStmt* wrapInCompoundStmt(clang::ASTContext& C) const;
-        
-    clang::Expr* getExpr() {
-      assert(llvm::isa<clang::Expr>(getStmt()) && "Must be an expression.");
-      return llvm::cast<clang::Expr>(getStmt());
-    }
-
-    //FIXME: warning: all paths through this function will call itself
-    const clang::Expr* getExpr() const {
-      return const_cast<NodeContext*>(this)->getExpr();
-    }
-
-    template<typename T> T* getAs() {
-      if (clang::Expr* E = llvm::dyn_cast<clang::Expr>(getStmt()))
-        return llvm::cast<T>(E);
-      return llvm::cast<T>(getStmt());
-    }
-  };
-
+  static clang::SourceLocation noLoc{};
   class DiffPlan;
-  /// The main builder class which then uses either ForwardModeVisitor or 
+  /// The main builder class which then uses either ForwardModeVisitor or
   /// ReverseModeVisitor based on the required mode.
   class DerivativeBuilder {
   private:
@@ -92,24 +47,8 @@ namespace clad {
 
     clang::Sema& m_Sema;
     clang::ASTContext& m_Context;
-    std::unique_ptr<clang::Scope> m_CurScope;
     std::unique_ptr<utils::StmtClone> m_NodeCloner;
     clang::NamespaceDecl* m_BuiltinDerivativesNSD;
-
-    /// Updates references in newly cloned statements.
-    void updateReferencesOf(clang::Stmt* InSubtree);
-    /// Clones a statement
-    clang::Stmt* Clone(const clang::Stmt* S);
-    /// A shorthand to simplify cloning of expressions.
-    clang::Expr* Clone(const clang::Expr* E);
-    /// A shorthand to simplify syntax for creation of new expressions.
-    /// Uses m_Sema.BuildUnOp internally.
-    clang::Expr* BuildOp(clang::UnaryOperatorKind OpCode, clang::Expr* E);
-    /// Uses m_Sema.BuildBin internally.
-    clang::Expr* BuildOp(
-      clang::BinaryOperatorKind OpCode,
-      clang::Expr* L,
-      clang::Expr* R);
 
     clang::Expr* findOverloadedDefinition(clang::DeclarationNameInfo DNI,
                             llvm::SmallVectorImpl<clang::Expr*>& CallArgs);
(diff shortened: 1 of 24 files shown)


test/FirstDerivative/BasicArithmeticMulDiv.C
```diff
@@ -12,54 +12,60 @@ int m_1(int x) {
...
   return x * y * x * 3 * x; // == 9 * x * x * y
 }
 // CHECK: int m_4_darg0(int x) {
+// CHECK-NEXT: int _d_y = 0;
 // CHECK-NEXT: int y = 4;
-// CHECK-NEXT: return ((((1 * y + x * 0) * x + x * y * 1) * 3 + x * y * x * 0) * x + x * y * x * 3 * 1);
+// CHECK-NEXT: int _t0 = x * y;
+// CHECK-NEXT: int _t1 = _t0 * x;
+// CHECK-NEXT: int _t2 = _t1 * 3;
+// CHECK-NEXT: return 1 * y + x * _d_y * x + _t0 * 1 * 3 + _t1 * 0 * x + _t2 * 1;
```
vgvassilev: This does not seem to produce 9x^2y.

test/FirstDerivative/BasicArithmeticMulDiv.C
```diff
@@ -94,21 +104,29 @@ double issue25(double x, double y) {
...
 }
 
-// CHECK: double issue25_darg0(double x, double y) {
-// CHECK-NEXT: x += 0.;
-// CHECK-NEXT: x -= 0.;
-// CHECK-NEXT: x /= 0.;
-// CHECK-NEXT: x *= 0.;
-// CHECK-NEXT: return 1.;
-// CHECK-NEXT: }
+// FIXME: +=, etc. operators are not currently supported, ignore for now.
```
vgvassilev: Could you open a ticket for this?

test/FirstDerivative/BasicArithmeticMulDiv.C
```diff
@@ -94,21 +104,29 @@ double issue25(double x, double y) {
...
 // CHECK: int md_1_darg0(int x) {
+// CHECK-NEXT: int _d_y = 0;
 // CHECK-NEXT: int y = 4;
-// CHECK-NEXT: return ((((((1 * x + x * 1) * x - x * x * 1) / (x * x) * y + x * x / x * 0) * y - x * x / x * y * 0) / (y * y) * 3 + x * x / x * y / y * 0) * 3 - x * x / x * y / y * 3 * 0) / (3 * 3);
+// CHECK-NEXT: int _t0 = x * x;
+// CHECK-NEXT: int _t1 = _t0 / x;
+// CHECK-NEXT: int _t2 = _t1 * y;
+// CHECK-NEXT: int _t3 = _t2 / y;
+// CHECK-NEXT: int _t4 = _t3 * 3;
+// CHECK-NEXT: return (((1 * x + x * 1 * x - _t0 * 1) / (x * x) * y + _t1 * _d_y * y - _t2 * _d_y) / (y * y) * 3 + _t3 * 0 * 3 - _t4 * 0) / (3 * 3);
```
vgvassilev: Look like this does not evaluate to 1 but to 1/3xy [see here](http://www.wolframalpha.com/input/?i=(((1+*+x+%2B+x+*+1+*+x+-+x+*+x+*+1)+%2F+(x+*+x)+*+y)+%2F+(y+*+y)+*+3)+%2F+(3+*+3)).

test/FirstDerivative/CallArguments.C
```diff
@@ -18,39 +18,40 @@ float g(float x) {
 }
 
 // CHECK: float g_darg0(float x) {
-// CHECK-NEXT: return f_darg0(x * x * x)  * (((1.F * x + x * 1.F) * x + x * x * 1.F));
+// CHECK: float _t0 = x * x;
+// CHECK-NEXT: f_darg0(_t0 * x) * (1.F * x + x * 1.F * x + _t0 * 1.F);
```
vgvassilev: The second term of the expression seems to evaluate to 2x^2*x instead of 3x^2.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -220,183 +293,312 @@ namespace clad {
...
-      stmts.push_back(Visit(*I).getStmt());
+  StmtDiff ForwardModeVisitor::VisitIfStmt(const IfStmt* If) {
+    // Create a block "around" if statement, e.g:
+    // {
+    //   ...
+    //  if (...) {...}
+    // }
+    startBlock();
+    auto init = If->getInit();
+    auto initResult = init ? Visit(init) : StmtDiff{};
```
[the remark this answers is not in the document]
efremale: There is nothing wrong with using StmtDiff() here, IMO {} shows the intent of creating "empty" object better.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -220,183 +293,312 @@ namespace clad {
...
+             "attempted differention w.r.t. a parameter ('%0') which is not of "
+             "a real type",
+             { m_IndependentVar->getNameAsString() });
+          return nullptr;
+        }
+        else
+          continue;
       }
+      // If param is independent variable, its derivative is 1, otherwise 0.
+      auto dValue =
```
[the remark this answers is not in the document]
vgvassilev: That’s the problem with auto :)

Why not: unsigned dValue = param == m_IndependentVar ?

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -220,183 +293,312 @@ namespace clad {
...
+                                          ifFalse.Expr_dx(),
+                                          ifTrue.Expr_dx()->getType(),
+                                          // FIXME: check if we can get lvalue
+                                          // in some case
+                                          VK_RValue,
+                                          OK_Ordinary);
+
+    auto condExprDiffParens = BuildParens(condExprDiff);
+
+    return {
```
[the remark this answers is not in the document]
vgvassilev: I think the former would suffice.

lib/Differentiator/DerivativeBuilder.cpp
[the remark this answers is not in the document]
efremale: Writing 1/0 explicitly looks clearer to me than using whatever the result of implicit cast is, but ok :)

lib/Differentiator/DerivativeBuilder.cpp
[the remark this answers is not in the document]
efremale: OK, but what is the benefit of restating the type of function's return twise (it is already written in its declaration)?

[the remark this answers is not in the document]
vgvassilev: It saves time if the reader to look it up especially when it can be meters away ;)

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -204,24 +234,32 @@ namespace clad {
...
+    StmtDiff VisitIntegerLiteral(const clang::IntegerLiteral* IL);
+    StmtDiff VisitFloatingLiteral(const clang::FloatingLiteral* FL);
+    StmtDiff VisitCallExpr(const clang::CallExpr* CE);
+    StmtDiff VisitDeclStmt(const clang::DeclStmt* DS);
+    StmtDiff VisitImplicitCastExpr(const clang::ImplicitCastExpr* ICE);
+    StmtDiff VisitConditionalOperator(const clang::ConditionalOperator* CO);
+    // Decl is not Stmt, so it cannot be visited directly.
+    // Instead, wrap it in DeclStmt and Visit.
+    // Orginal cloned VarDecl is output to Cloned arg, for easier access.
+    StmtDiff VisitVarDecl(const clang::VarDecl* VD, clang::VarDecl*& Cloned) {
```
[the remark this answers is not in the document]
efremale: This is useful. Decl is not derived from Stmt and must be wrapped in DeclStmt. However, If and For statements can contain a declaration, which is explicitly obtained as VarDecl, not DeclStmt (and their constructors accept only VarDecl too). This VarDecl must be differentiated carefully, its derivative must be emitted prior to If/ForStmt and its clone must be given to constructor without emitting, e.g:
```
if (int x = y) {...}
```
->
{
 int _d_x = _d_y;
 if (int x = y) {...} // if (int _d_x = _d_y, x = y) wouldn't work in this case
}
```

Therefore, we need some function that takes VarDecl and outputs DeclStmt for its derivative and VarDecl for its clone, i.e. ProcessVarDecl: VarDecl* -> (DeclStmt*, VarDecl*).

Maybe it is misleading to name it VisitVarDecl and return StmtDiff where .Stmt() part is not used.
Should I then rename it to ProcessVarDecl or something similar?

File: lib/Differentiator/DerivativeBuilder.cpp [hunk 9]
```diff
@@ -75,52 +65,132 @@ namespace clad {
...
+    auto DGR = DeclGroupRef::Create(m_Context, DS.data(), DS.size());
+    return new (m_Context) DeclStmt(DGR,
+                                    noLoc,
+                                    noLoc);
+  }
+
+  Expr* VisitorBase::BuildDeclRef(ValueDecl* D) {
+    return m_Sema.BuildDeclRefExpr(D,
+                                   D->getType(),
+                                   VK_LValue,
```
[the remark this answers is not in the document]
efremale: Since we are only building references to local variable declarations, it is always LValue. I will change the arg type from ValueDecl* to VarDecl*. Otherwise we'll have to use Sema::BuildDeclarationNameExpr.

include/clad/Differentiator/DerivativeBuilder.h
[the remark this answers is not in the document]
vgvassilev: Yes, I was after picking up a different name for the routine.

test/FirstDerivative/CodeGenSimple.C
```diff
@@ -6,12 +6,13 @@
 #include "clad/Differentiator/Differentiator.h"
 extern "C" int printf(const char* fmt, ...); //expected-warning{{function 'printf' was not differentiated because it is not declared in namespace 'custom_derivatives'}}
 int f_1(int x) {
-   printf("I am being run!\n");
+   printf("I am being run!\n"); //expected-warning{{attempted to differentiate unsupported statement, no changes applied}}
    return x * x;
 }
 // CHECK: int f_1_darg0(int x) {
+// CHECK-NEXT: 0;
```
vgvassilev: Why do we emit this?

test/FirstDerivative/DiffInterface.C
```diff
@@ -40,6 +45,34 @@ int f_3() {
...
+
+int f_try_catch(int x)
+  try {
+    return x;
+  }
+  catch (int) {
+    return 0;
+  } // expected-warning {{attempted to differentiate unsupported statement, no changes applied}}
+
+// CHECK: int f_try_catch_darg0(int x) {
```
vgvassilev: We must not synthesize a wrong function. We should probably promote the warning into an error.

test/FirstDerivative/FunctionCalls.C
```diff
@@ -76,15 +76,15 @@ float test_4(int x) {
 }
 
 // CHECK: float test_4_darg0(int x) {
-// CHECK-NEXT: return overloaded();
+// CHECK-NEXT: return 0;
```
vgvassilev: Why this did not rewire the call? I'd tend to promote the warning into an error and not produce `test_4_darg0`.

test/FirstDerivative/FunctionsOneVariable.C
```diff
@@ -12,23 +12,24 @@ float f_simple(float x) {
...
+//CHECK-NEXT:    return 0.F * x + 0.F * 1.F + 0.F * 1.F + 1.F * 0.F + 0.F * 1.F + 1.F * 0.F + 1.F * 0.F + x * 0.F;
 //CHECK-NEXT:}
 
 int f_simple_negative(int x) {
   //  printf("This is f(x).\n");
   return -x*x;
 }
 // CHECK: int f_simple_negative_darg0(int x) {
-// CHECK-NEXT: return (-1 * x + -x * 1);
+// CHECK-NEXT: int _t0 = -x;
```
vgvassilev: Do we really a temporary here?

test/FirstDerivative/NonContinuous.C
```diff
@@ -16,9 +16,11 @@ int f(int x) {
   return x*x;
 }
 // CHECK: int f_darg0(int x) {
-// CHECK-NEXT: if (x < 0)
-// CHECK-NEXT:   return (-1 * x + -x * 1);
-// CHECK-NEXT: return (1 * x + x * 1);
+// CHECK-NEXT: if (x < 0) {
+// CHECK-NEXT:   int _t0 = -x;
```
vgvassilev: Likewise.

test/FirstDerivative/NonContinuous.C
```diff
@@ -47,12 +51,12 @@ int g(long y) {
     return 2;
 }
 // CHECK: int g_darg0(long y) {
-// CHECK-NEXT: if (y)
+// CHECK-NEXT: if (y) {
```
vgvassilev: We do not really need a compound statement here.

test/FirstDerivative/NonContinuous.C
```diff
@@ -47,12 +51,12 @@ int g(long y) {
...
 // CHECK-NEXT:   return 0;
-// CHECK-NEXT: else
+// CHECK-NEXT:  } else {
```
vgvassilev: Likewise.

test/FirstDerivative/Recursive.C
```diff
@@ -15,10 +15,11 @@ int f_pow(int arg, int p) {
 }
 
 // CHECK: int f_pow_darg0(int arg, int p) {
-// CHECK-NEXT:  if (p == 0)
+// CHECK-NEXT:  if (p == 0) {
```
vgvassilev: Likewise.

test/Misc/RunDemos.C
```diff
@@ -11,15 +11,33 @@
...
 // CHECK_GRADIENT-NOT:{{.*error|warning|note:.*}}
 // CHECK_GRADIENT:float sphere_implicit_func_darg0(float x, float y, float z, float xc, float yc, float zc, float r) {
-// CHECK_GRADIENT:    return ((1.F - (0.F)) * (x - xc) + (x - xc) * (1.F - (0.F))) + (((0.F - (0.F)) * (y - yc) + (y - yc) * (0.F - (0.F)))) + (((0.F - (0.F)) * (z - zc) + (z - zc) * (0.F - (0.F)))) - ((0.F * r + r * 0.F));
+// CHECK_GRADIENT: float _t0 = (x - xc);
+// CHECK_GRADIENT: float _t1 = (x - xc);
+// CHECK_GRADIENT: float _t2 = (y - yc);
+// CHECK_GRADIENT: float _t3 = (y - yc);
+// CHECK_GRADIENT: float _t4 = (z - zc);
+// CHECK_GRADIENT: float _t5 = (z - zc);
+// CHECK_GRADIENT: return (1.F - 0.F) * _t1 + _t0 * (1.F - 0.F) + (0.F - 0.F) * _t3 + _t2 * (0.F - 0.F) + (0.F - 0.F) * _t5 + _t4 * (0.F - 0.F) - (0.F * r + r * 0.F);
```
vgvassilev: IIUC, this yields different result.

test/Misc/RunDemos.C
```diff
@@ -11,15 +11,33 @@
 // RUN: %cladclang %S/../../demos/Gradient.cpp -I%S/../../include -oGradient.out 2>&1 | FileCheck -check-prefix CHECK_GRADIENT %s 
 // CHECK_GRADIENT-NOT:{{.*error|warning|note:.*}}
 // CHECK_GRADIENT:float sphere_implicit_func_darg0(float x, float y, float z, float xc, float yc, float zc, float r) {
-// CHECK_GRADIENT:    return ((1.F - (0.F)) * (x - xc) + (x - xc) * (1.F - (0.F))) + (((0.F - (0.F)) * (y - yc) + (y - yc) * (0.F - (0.F)))) + (((0.F - (0.F)) * (z - zc) + (z - zc) * (0.F - (0.F)))) - ((0.F * r + r * 0.F));
+// CHECK_GRADIENT: float _t0 = (x - xc);
+// CHECK_GRADIENT: float _t1 = (x - xc);
+// CHECK_GRADIENT: float _t2 = (y - yc);
+// CHECK_GRADIENT: float _t3 = (y - yc);
+// CHECK_GRADIENT: float _t4 = (z - zc);
+// CHECK_GRADIENT: float _t5 = (z - zc);
```
vgvassilev: Looks like `_t5`, `_t3` and `_t1` are redundant.

test/Misc/RunDemos.C
```diff
@@ -11,15 +11,33 @@
...
 
 // CHECK_GRADIENT:float sphere_implicit_func_darg2(float x, float y, float z, float xc, float yc, float zc, float r) {
-// CHECK_GRADIENT:    return ((0.F - (0.F)) * (x - xc) + (x - xc) * (0.F - (0.F))) + (((0.F - (0.F)) * (y - yc) + (y - yc) * (0.F - (0.F)))) + (((1.F - (0.F)) * (z - zc) + (z - zc) * (1.F - (0.F)))) - ((0.F * r + r * 0.F));
+// CHECK_GRADIENT: float _t0 = (x - xc);
+// CHECK_GRADIENT: float _t1 = (x - xc);
+// CHECK_GRADIENT: float _t2 = (y - yc);
+// CHECK_GRADIENT: float _t3 = (y - yc);
+// CHECK_GRADIENT: float _t4 = (z - zc);
+// CHECK_GRADIENT: float _t5 = (z - zc);
+// CHECK_GRADIENT: return (0.F - 0.F) * _t1 + _t0 * (0.F - 0.F) + (0.F - 0.F) * _t3 + _t2 * (0.F - 0.F) + (1.F - 0.F) * _t5 + _t4 * (1.F - 0.F) - (0.F * r + r * 0.F);
```
vgvassilev: Likewise.

test/FirstDerivative/BasicArithmeticMulDiv.C
efremale: Done: https://github.com/vgvassilev/clad/issues/85

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -137,52 +76,168 @@ namespace clad {
...
+      return m_Blocks.top();
+    }
+    /// Create new block.
+    Stmts & beginBlock() {
+      m_Blocks.push({});
+      return m_Blocks.top();
+    }
+    /// Remove the block from the stack, wrap it in CompoundStmt and return it.
+    clang::CompoundStmt* endBlock() {
+      auto CS = MakeCompoundStmt(getCurrentBlock());
```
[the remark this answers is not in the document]
efremale: In many cases, it is actually desirable to get a CompoundStmt, even if it has only 1 statement inside. E.g. when visiting another CompoundStmt (especially a function body).

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -137,52 +76,168 @@ namespace clad {
...
+    clang::VarDecl* BuildVarDecl(clang::QualType Type,
+                                 llvm::StringRef prefix = "_t",
+                                 clang::Expr* Init = nullptr,
+                                 bool DirectInit = false);
+    /// Wraps a declaration in DeclStmt.
+    clang::Stmt* BuildDeclStmt(clang::Decl* D);
+    clang::Stmt* BuildDeclStmt(llvm::MutableArrayRef<clang::Decl*> DS);
+
+    /// Builds a DeclRefExpr to a given Decl.
+    clang::Expr* BuildDeclRef(clang::VarDecl* D);
```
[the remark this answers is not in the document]
efremale: Sema::BuildDeclRefExpr returns ExprResult (alias for ActionResult<Expr*>), and .get() on it returns Expr*. We can cast it to DeclRefExpr*, though it seems to be unnecessary since in most of the cases getting Expr* is sufficient.

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -137,52 +76,168 @@ namespace clad {
...
 
-    /// Wraps variable declaration in DeclStmt
-    clang::Stmt* BuildDeclStmt(clang::VarDecl* VD);
+    clang::VarDecl* BuildVarDecl(clang::QualType Type,
+                                 llvm::StringRef prefix = "_t",
+                                 clang::Expr* Init = nullptr,
+                                 bool DirectInit = false);
+    /// Wraps a declaration in DeclStmt.
+    clang::Stmt* BuildDeclStmt(clang::Decl* D);
+    clang::Stmt* BuildDeclStmt(llvm::MutableArrayRef<clang::Decl*> DS);
```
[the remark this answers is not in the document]
efremale: Similarly to the above, with StmtResult/Stmt*.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -75,52 +65,136 @@ namespace clad {
...
+                         llvm::ArrayRef<llvm::StringRef> args,
+                         clang::SourceLocation loc) {
+    unsigned diagID
+      = m_Sema.Diags.getCustomDiagID(level, format);
+    Sema::SemaDiagnosticBuilder stream = m_Sema.Diag(loc, diagID);
+    for (auto arg : args)
+      stream << arg;
+  }
+
+  Expr* VisitorBase::StoreAndRef(Expr* E, llvm::StringRef prefix,
```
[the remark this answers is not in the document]
efremale: See above.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -75,52 +65,136 @@ namespace clad {
...
+                                 bool forceDeclCreation) {
+    return StoreAndRef(E, E->getType(), prefix, forceDeclCreation);
+  }
+
+  Expr* VisitorBase::StoreAndRef(Expr* E,
```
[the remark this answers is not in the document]
efremale: See above.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -945,15 +1167,15 @@ namespace clad {
...
         auto dStmt = new (m_Context) ParenExpr(noLoc,
                                                noLoc,
                                                condExpr);
         Visit(branch, dStmt);
     };
-   
+
     // FIXME: not optimal, creates two (condExpr ? ... : ...) expressions,
-    // so cond is unnesarily checked twice. 
+    // so cond is unnesarily checked twice.
```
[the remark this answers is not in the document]
efremale: Removing the comment, it was left here from older versions, now it is fixed.

test/FirstDerivative/CodeGenSimple.C
efremale: Since there exist no custom_derivatives::printf_darg0, we set the derivative of printf to be zero. In principle, we could detect it and return null instead, but this would break cases when printf is used in expressions, e.g. `int i = x + printf(...);` -> `int _d_i = _d_x + 0;` . This is an obscure case, but it makes sense to use 0 in such expressions (say, if it is not printf, but some non-differentiable math function, we cannot determine that).

It does not make much sense to emit 0 in this case, but it is very hard to detect in general case.
Another option is just to exit with hard error whenever non-differentiable function is visited.

test/FirstDerivative/FunctionCalls.C
efremale: In this case, since there is no custom_derivatives::overloaded_darg0, its derivative is set to 0. Strictly speaking, this should be
```
fload test_4_darg0(int x) {
  return 0;
  return overloaded();
}
```
but there is no reason to emit 2nd return.

test/FirstDerivative/FunctionsOneVariable.C
efremale: I did not add unary operators to a list of exceptions which are not stored in temporaries (e.g. literals and references). I agree, there is not much sense in using a temporary for unary operation like `-x` on built-in types, but e.g. we would have to store `++x` to avoid incrementing several times (though `++` is not supported yet). 

Should I add unary `+` and `-` to the list of things which are not stored? This requires a bit more complicated checking, i.e. first we need to check if it is UnaryOperator, then cast it and check its getOpcode().

test/Misc/RunDemos.C
efremale: It seems to be the same.

test/Misc/RunDemos.C
efremale: Since the original function uses `(x - xc)` on several places, we get several temporaries for it. We do not check the original function for repeated expressions, it would add quite a huge computational overhead. Moreover, since it was acceptable for the original function to use the same expression repeatedly, it is presumably acceptable to do so in the derivative.

efremale: I.e. the original function is `return (x-xc)*(x-xc) + (y-yc)*(y-yc) + (z-zc)*(z-zc) - r*r;`, not `auto t0 = (x - xc); auto t1 = (y - yc); auto t2 = (z - zc); return t0 * t0 + t1 * t1 + t2 * t2 - r * r'`.

include/clad/Differentiator/DerivativeBuilder.h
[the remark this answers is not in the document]
vgvassilev: This is a generic concept in Sema. Here we know what we expect. The return type narrowing has a lot of advantages from making the code clearer to detecting silent change of behavior.

test/FirstDerivative/CodeGenSimple.C
vgvassilev: I see. Maybe we should assign it to a temporary with some special name showing the intent. Could you open an issue for this?

test/FirstDerivative/DiffInterface.C
vgvassilev: Alternatively, we could add a `bool IsValid;` to StmtDiff which we can accumulate and check where necessary.

test/FirstDerivative/FunctionCalls.C
vgvassilev: Maybe we could do:
```
float _fn_overloaded_non_differentiable = 0;
return _fn_overloaded_non_differentiable;
```

test/FirstDerivative/FunctionsOneVariable.C
vgvassilev: I'd opt for opening an issue and addressing it in a subsequent PR.

test/Misc/RunDemos.C
vgvassilev: Ok, let's resolve this by opening an issue to document this.

test/Misc/RunDemos.C
vgvassilev: Indeed, I screwed up checking.

include/clad/Differentiator/DerivativeBuilder.h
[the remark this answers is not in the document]
efremale: Are we sure that BuildDeclRefExpr always returns DeclRefExpr though? For example, BuildBinOp may return either BinaryOperator or CXXOperatorCallExpr.

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -169,7 +169,12 @@ namespace clad {
     template <std::size_t N>
     void diag(clang::DiagnosticsEngine::Level level, // Warning or Error
               const char (&format)[N],
-              llvm::ArrayRef<llvm::StringRef> args = {},
+              llvm::ArrayRef<llvm::StringRef> args,
+              clang::SourceLocation loc = {});
```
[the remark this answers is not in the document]
efremale: Fixed

test/Misc/RunDemos.C
efremale: Done: https://github.com/vgvassilev/clad/issues/88

test/FirstDerivative/FunctionsOneVariable.C
efremale: Done: https://github.com/vgvassilev/clad/issues/89

test/FirstDerivative/FunctionCalls.C
efremale: Issue created: https://github.com/vgvassilev/clad/issues/90

test/FirstDerivative/CodeGenSimple.C
efremale: See issue: https://github.com/vgvassilev/clad/issues/90

test/FirstDerivative/DiffInterface.C
efremale: On the other hand, it makes sense to clone some unsupported things without error, e.g. string literals as in `printf("string")`.

efremale: Issue created: https://github.com/vgvassilev/clad/issues/91

include/clad/Differentiator/DerivativeBuilder.h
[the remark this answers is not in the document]
efremale: The latest commit narrows return types.

PR #81, opened 2018-08-13
Preserve namespace specifiers

efremale:
Currently, clad omits namespace specifiers when cloning DeclRefExprs to
variables/functions or looking up derivatives in custom_derivatives
namespace.

This may lead to problems when unqialified name lookup finds something
other than what was intended.

For example, differentiating
```
double f(double x) {
  double t = std::sin(x);
  return t;
}
```
currently results in
```
double f_darg0(double x) {
  double _d_t = sin_darg0(x);
  double t = sin(x);
  return _d_t;
}
```
Consider if there is a function `sin` in the current namespace:
```
double sin(double x) {
  assert(false);
}
```
Unqualified name lookup would find this function instead of `std::sin` and
`f_darg0` will fail on assert.

This commit changes this behaviour and preserves namespace specifiers and
adds `custom_derivatives` namespace when necessary. The output of previous
example now is:
```
double f_darg0(double x) {
  double _d_t = custom_derivatives::sin_darg0(x);
  double t = std::sin(x);
  return _d_t;
}
```


vgvassilev:
Nice!

What happens after this pr if we have:

namespace std { double sin(double x){...}}
using namespace std;

double f(double x) { return sin(x); }

It is important the clone to omit the namespace qualifier because otherwise we will turn off ADL, which was turned on in the initial function.

efremale:
In your example, the result would be
```
double f_darg0(double x) {
    return custom_derivatives::sin_darg0(x) * (1.);
}
```
which is fine since we require `sin_darg0` to be in `custom_derivatives` anyway.

However, in cases where `sin(x)` is ''copied'' to the derivative function, it is copied as it is (without namespace specifier in this case), i.e. result for
```
double f(double x) {
  double t = sin(x);
  return t;
}
```
is
```
double f_darg0(double x) {
    double _d_t = custom_derivatives::sin_darg0(x) * (1.);
    double t = sin(x);
    return _d_t;
}
```

lib/Differentiator/StmtClone.cpp
```diff
@@ -43,7 +43,11 @@ Stmt* StmtClone::Visit ## CLASS(CLASS *Node)  \
 
 DEFINE_CLONE_EXPR(BinaryOperator, (Clone(Node->getLHS()), Clone(Node->getRHS()), Node->getOpcode(), Node->getType(), Node->getValueKind(), Node->getObjectKind(), Node->getOperatorLoc(), Node->getFPFeatures()))
 DEFINE_CLONE_EXPR(UnaryOperator, (Clone(Node->getSubExpr()), Node->getOpcode(), Node->getType(), Node->getValueKind(), Node->getObjectKind(), Node->getOperatorLoc()))
-DEFINE_CLONE_EXPR(DeclRefExpr, (Node->getDecl(), Node->refersToEnclosingVariableOrCapture(), Node->getType(), Node->getValueKind(), Node->getLocation()))
+Stmt* StmtClone::VisitDeclRefExpr(DeclRefExpr *Node) {
```
vgvassilev: I'd prefer to use the same style here. Could you implement `DEFINE_CLONE_EXPR` for the template arguments. Then we can use `DEFINE_CREATE_EXPR` to use the relevant `DeclRefExpr::Create` call.

efremale: Do you mean to define DEFINE_CREATE_EXPR and use it here instead of DEFINE_CLONE_EXPR?

efremale: Is TemplateArgumentList visitable? In my understanding, StmtClone can only visit types that are derived from Stmt (and TemlateArgumentList is not). It is also copied in the same way in VisitUnresolvedLookupExpr.

PR #82, opened 2018-08-13
Handle scopes properly

efremale:
Handle scopes properly

Previously, scopes were not handled properly and varibable declarations
were not removed from IdResolver on scope exit. This could lead to errors,
in particular with shadowing declarations, e.g.:
```
double f(double x) {
  { double x = 0; return x; }
}
```
would cause an error: reference to 'x' is ambiguous. IdResolver was not
cleared properly, which also causes errors, for example, on
```
double f(double x) {
  { double y = ...; // inside block scope }
  // y is no longer visible here, but was not removed from IdResolver
  double y = ...; // error: reference to 'y' is ambiguous
}
```

This commit introduces proper handling of scopes and declarations in
IdResolver.

### DIFF
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -124,17 +124,17 @@ namespace clad {
     }
 
     /// Get a current scope.
-    clang::Scope* currentScope() {
+    clang::Scope* getCurrentScope() {
       return m_CurScope;
     }
     /// Enters a new scope.
-    void enterScope(unsigned ScopeFlags) {
+    void beginScope(unsigned ScopeFlags) {
       // FIXME: since Sema::CurScope is private, we cannot access it and have
       // to use separate member variable m_CurScope. The only options to set
       // CurScope of Sema seemt to be through Parser or ContextAndScopeRAII.
-      m_CurScope = new clang::Scope(currentScope(), ScopeFlags, m_Sema.Diags);
+      m_CurScope = new clang::Scope(getCurrentScope(), ScopeFlags, m_Sema.Diags);
     }
-    void exitScope() {
+    void endScope() {
       // This will remove all the decls in the scope from the IdResolver. 
       m_Sema.ActOnPopScope(noLoc, m_CurScope);
       auto oldScope = m_CurScope;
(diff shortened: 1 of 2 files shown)


include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -123,6 +123,25 @@ namespace clad {
       currentBlock().push_back(S);
     }
 
+    /// Get a current scope.
+    clang::Scope* currentScope() {
```
vgvassilev: Could you rename this to `getCurScope`?

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -123,6 +123,25 @@ namespace clad {
...
+      return m_CurScope;
+    }
+    /// Enters a new scope.
+    void enterScope(unsigned ScopeFlags) {
```
vgvassilev: I'd prefer `beginScope`/`endScope`.

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -123,6 +123,25 @@ namespace clad {
...
+      // FIXME: since Sema::CurScope is private, we cannot access it and have
+      // to use separate member variable m_CurScope. The only options to set
+      // CurScope of Sema seemt to be through Parser or ContextAndScopeRAII.
```
vgvassilev: I am confused by this comment. What is the reason for not using a `ContextAndScopeRAII` object?

efremale: Is it possible to use `ContextAndScopeRAII`, though handling RAII object lifetime is sometimes inconvenient. 

Also, in most of the cases, we need to enter a new Scope, but not DeclContext, though we can circumvent it by passing the same DeclContext.

Also, destructor of `ContextAndScopeRAII` does not call `Sema::ActOnPopScope` (which we now call on `exitScope`, it is useful to remove all the `Decl`s of popped scope from `IdResolver`). So, if we use `ContextAndScopeRAII` instead of our own `enterScope`/`exitScope`, we have to additionally call `ActOnPopScope` before the RAII is destroyed. Overall, it seems to be inconvenient and not worth it.

PR #83, opened 2018-08-13
Support variable assignments in forward mode

efremale:
This commit allows to keep track of variable reassignments in forward mode.
Example:
```
  y = x;
  return y;
}
->
double f_darg0(double x, double y) {
  double _d_x = 1;
  double _d_y = 0;
  _d_y = _d_x;
  y = x;
  return _d_y;
}
```

Currently assigmnents are only supported for function parameters and
variables declared inside the function body (i.e. assignments to global
variables and struct members are not tracked).

### DIFF
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -269,6 +269,12 @@ namespace clad {
     /// Map used to keep track of variable declarations and match them
     /// with their derivatives.
     std::unordered_map<clang::VarDecl*, clang::Expr*> m_Variables;
+    /// Map contains variable declarations replacements. If the original
+    /// function contains a declaration which name collides with something
+    /// already created inside derivative's body, the declaration is replaced
+    /// with a new one.
+    /// See the example inside ForwardModeVisitor::VisitDeclStmt.
+    std::unordered_map<const clang::VarDecl*, clang::VarDecl*> m_DeclReplacements;
     unsigned m_DerivativeOrder;
     unsigned m_ArgIndex;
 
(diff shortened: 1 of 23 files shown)


include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -238,6 +238,7 @@ namespace clad {
     /// Map used to keep track of variable declarations and match them
     /// with their derivatives.
     std::unordered_map<clang::VarDecl*, clang::Expr*> m_Variables;
+    std::unordered_map<const clang::ValueDecl*, clang::ValueDecl*> m_DeclReplacements;
```
vgvassilev: Could you add documentation to this member.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -329,16 +329,29 @@ namespace clad {
           1 :
           0 ;
       auto dParam =
-        ConstantFolder::synthesizeLiteral(param->getType(),
+        ConstantFolder::synthesizeLiteral(m_Context.IntTy,
```
vgvassilev: Out of curiosity: why did we have to change this? And why not UnsignedIntTy?

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -933,35 +965,70 @@ namespace clad {
...
         VD = BuildVarDecl(VD->getType(),
-                          VD->getIdentifier(),
+                          VD->getNameAsString(),
                           initDiff.Expr(),
                           VD->isDirectInit());
+
+        // Check if decl's name is the same as before. The name may be changed
+        // if decl name collides with something in the derivative body.
+        // This can happen in rare cases, e.g. when the original function
+        // has both y and _d_y (here _d_y collides with the name produced by
```
vgvassilev: Perhaps we should we diagnose such cases.

lib/Differentiator/DerivativeBuilder.cpp
efremale: I did it to make the derivative look cleaner: it is cleaner to use simple int literals (1/0) instead of 1ULL or 1.F. This should not break anything.

PR #98, opened 2018-10-05
Fix issues with differentiation of functions inside namespace

efremale:
Derivaties of functions inside namespace are now output into the same namespace.

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -20,6 +20,8 @@
 #include "clang/Sema/Overload.h"
 #include "clang/Sema/SemaInternal.h"
 
+#include "llvm/Support/SaveAndRestore.h"
+
 using namespace clang;
 
 namespace clad {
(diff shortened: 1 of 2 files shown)


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -113,6 +113,70 @@ namespace clad {
...
+        Parent->addDecl(UD);
+      }
+    }
+    // Namespace scope and declcontext. Must be exited by the user.
+    beginScope(Scope::DeclScope);
+    m_Sema.PushDeclContext(m_CurScope, NDecl);
+    return NDecl;
+  }
+
+  NamespaceDecl* VisitorBase::RebuildEnclosingNamespaces(DeclContext* DC) {
```
vgvassilev: Why not taking a `NamespaceDecl*`, it is also a `DeclContext`.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -249,7 +313,10 @@ namespace clad {
     IdentifierInfo* II = &m_Context.Idents.get(
       derivativeBaseName + "_d" + s + "arg" + std::to_string(m_ArgIndex));
     DeclarationNameInfo name(II, noLoc);
-    FunctionDecl* derivedFD = 0;
+    FunctionDecl* derivedFD = nullptr;
+    NamespaceDecl* enclosingNS = nullptr;
+    DeclContext* oldContext = m_Sema.CurContext;
```
vgvassilev: Maybe use `llvm::SaveAndRestore` for such operations?

tools/ClangPlugin.cpp
```diff
@@ -124,33 +125,37 @@ namespace clad {
...
                  std::error_code err;
                  llvm::raw_fd_ostream f("Derivatives.cpp", err,
                                         llvm::sys::fs::F_Append);
-                 Result->print(f, Policy);
+                 DerivativeDecl->print(f, Policy);
                  f.flush();
               }
               // Call CodeGen only if the produced decl is a top-most decl.
-              if (Result->getDeclContext()
+              Decl* OutputStmt = DerivativeDeclContext ?
```
vgvassilev: Could you use for example `DerivativeDeclOrNamespace` or something similar as a variable name?

lib/Differentiator/DerivativeBuilder.cpp
efremale: At some point, some outer DeclContext that is not a NamespaceDecl will be reached. This is checked inside the function and it makes more sense for me to do so, instead of checking it outside.

PR #108, opened 2019-03-07
Update clad documentation and other NFCs

oshadura:


CMakeLists.txt
```diff
@@ -125,7 +125,9 @@ add_subdirectory(lib)
 add_subdirectory(tools)
 
 # TODO: docs.
-if (NOT CLAD_BUILD_STATIC_ONLY)
+# For tests we need to have lit (LLVM Integrated Tester), we need to add a check if lit was found
+# since it is not distributed via binaries.
```
vgvassilev: We should probably issue a warning before.

PR #109, opened 2019-03-18
Add string arguments for parameter specification

efremale:
The commit introduces another option for the 2nd parameter of a call to
`clad::differentiate` and `clad::gradient`.
Now, there are 3 ways how to provide specific parameter(s), w.r.t. which the
differentiation will happen in `clad::differentiate` and `clad::gradient`:

1) Pass a string literal, containing comma-separated names of function's parameters,
as defined in function's defintion. The function will be differentiated w.r.t. all
the specified parameters. E.g., for a fucntion `double f(double x, double y, double
z) {...}`, calling `clad::gradient(f, "x, y")` will only differentiate it w.r.t. `x`
and `y` but not `z`.

2) Pass a numeric literal representing a single value. The function will be
differentiated w.r.t. to the parameter corresponding to literal's value index. For
now it only works for a single index/parameter.

3) Provide no argument, default argument will be used. The function will be
differentiated w.r.t. to its every parameter. E.g., `clad::gradient(f)`.

For now, gradient results are stored in the `_result` parameter in the same order as
they were provided in the argument string. E.g. on `clad::gradient(f, "y, x")`, `y`'s
derivative is stored in `_result[0]` and `x`'s in `_result[1]`. This will likely be
changed.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -8,37 +8,10 @@ using namespace clang;
 
 namespace clad {
   static SourceLocation noLoc;
-  FunctionDeclInfo::FunctionDeclInfo(FunctionDecl* FD, ParmVarDecl* PVD)
-    : m_FD(FD), m_PVD(PVD) {
-#ifndef NDEBUG
-    if (!PVD) // Can happen if there was error deducing the argument
-      return;
-    bool inArgs = false;
-    for (unsigned i = 0; i < FD->getNumParams(); ++i)
-      if (PVD == FD->getParamDecl(i)) {
-        inArgs = true;
-        break;
-      }
-    assert(inArgs && "Must pass in a param of the FD.");
-#endif
-  }
-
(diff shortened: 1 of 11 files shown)


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -225,31 +153,17 @@ namespace clad {
...
-          if (index != -1)
-            if (FD->getParamDecl(index) == m_TopMostFDI->getPVD()) {
-              index = i - 1; // Decrement by 1, adapting to FD's 0 param list.
-              break;
-            }
-        if (index > -1) {
-          FunctionDeclInfo FDI(FD, FD->getParamDecl(index));
-          getCurrentPlan().push_back(FDI);
-        }
-      }
```
efremale: I am not sure what this part was intended for (likely recursive differentiation of called functions inside the currently differentiated one), it is not used anyhow at the moment. Recursive differentiation will require a bit more work in the future.

PR #110, opened 2019-03-21
Add declarations to reverse mode

### LINKED ISSUE #105: Reverse mode does not support DeclStmt-s
Consider:
```cpp
double breitwigner_pdf(double x, double gamma, double x0) {
  double gammahalf = gamma/2.0;
  return gammahalf/(M_PI * ((x-x0)*(x-x0) + gammahalf*gammahalf));
}
auto cl = clad::gradient(breitwigner_pdf);
```

efremale:
Now we support declarations of variables in the reverse mode. Variable reassignments are not yet supported.

Example:
```
  double a = x * x;
  double b = y * y;
  double c = a + b;
  return c;
}
```

Fix #105

### DIFF
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -106,9 +106,19 @@ namespace clad {
     clang::FunctionDecl* m_Derivative;
     /// The function that is currently differentiated.
     clang::FunctionDecl* m_Function;
+
+    /// Map used to keep track of variable declarations and match them
+    /// with their derivatives.
+    std::unordered_map<clang::VarDecl*, clang::Expr*> m_Variables;
+    /// Map contains variable declarations replacements. If the original
+    /// function contains a declaration which name collides with something
+    /// already created inside derivative's body, the declaration is replaced
+    /// with a new one.
+    /// See the example inside ForwardModeVisitor::VisitDeclStmt.
+    std::unordered_map<const clang::VarDecl*, clang::VarDecl*> m_DeclReplacements;
     /// A stack of all the blocks where the statements of the gradient function
     /// are stored (e.g., function body, if statement blocks).
-    std::stack<Stmts> m_Blocks;
+    std::vector<Stmts> m_Blocks;
 
     template <typename Range>
     clang::CompoundStmt* MakeCompoundStmt(const Range & Stmts) {
@@ -121,17 +131,17 @@ namespace clad {
 
     /// Get the latest block of code (i.e. place for statements output).
     Stmts& getCurrentBlock() {
-      return m_Blocks.top();
+      return m_Blocks.back();
     }
     /// Create new block.
     Stmts& beginBlock() {
-      m_Blocks.push({});
-      return m_Blocks.top();
+      m_Blocks.push_back({});
+      return m_Blocks.back();
     }
     /// Remove the block from the stack, wrap it in CompoundStmt and return it.
     clang::CompoundStmt* endBlock() {
       auto CS = MakeCompoundStmt(getCurrentBlock());
-      m_Blocks.pop();
+      m_Blocks.pop_back();
       return CS;
     }
 
@@ -140,6 +150,7 @@ namespace clad {
     /// Output a statement to the current block. If Stmt is null or is an unused
     /// expression, it is not output and false is returned. 
     bool addToCurrentBlock(clang::Stmt* S);
+    bool addToBlock(clang::Stmt* S, Stmts& block);
 
     /// Get a current scope.
     clang::Scope* getCurrentScope() {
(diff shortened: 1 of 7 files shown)


include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -347,22 +352,95 @@ namespace clad {
...
+    : public clang::ConstStmtVisitor<ReverseModeVisitor, StmtDiff>,
       public VisitorBase {
   private:
     llvm::SmallVector<clang::VarDecl*, 16> m_IndependentVars;
     /// Stack is used to pass the arguments (dfdx) to further nodes
     /// in the Visit method.
     std::stack<clang::Expr*> m_Stack;
     clang::Expr* dfdx () {
-        return m_Stack.top ();
+        if (m_Stack.empty()) {
```
vgvassilev: Could you fix the indentation here and elsewhere to 2 cols.

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -347,22 +352,95 @@ namespace clad {
...
+      return getCurrentBlock(d);
+    }
+    /// Remove the block from the stack, wrap it in CompoundStmt and return it.
+    clang::CompoundStmt* endBlock(direction d = forward) {
+      if (d == forward) {
+        auto CS = MakeCompoundStmt(getCurrentBlock(forward));
+        m_Blocks.pop_back();
+        return CS;
+      }
+      else {
```
vgvassilev: Move the else on the previous line.

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -391,21 +469,23 @@ namespace clad {
...
-    void VisitMemberExpr(const clang::MemberExpr* ME);
-    void VisitArraySubscriptExpr(const clang::ArraySubscriptExpr* ASE);
-    void VisitDeclRefExpr(const clang::DeclRefExpr* DRE);
-    void VisitParenExpr(const clang::ParenExpr* PE);
-    void VisitIntegerLiteral(const clang::IntegerLiteral* IL);
-    void VisitFloatingLiteral(const clang::FloatingLiteral* FL);
-    void VisitCallExpr(const clang::CallExpr* CE);
-    void VisitImplicitCastExpr(const clang::ImplicitCastExpr* ICE);
-    void VisitConditionalOperator(const clang::ConditionalOperator* CO);
+    StmtDiff VisitCompoundStmt(const clang::CompoundStmt* CS);
```
vgvassilev: Maybe order all Visit* alphabetically?

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -832,15 +834,15 @@ namespace clad {
...
-        // If current context is different than the context of the original
-        // declaration (e.g. we are inside lambda), rebuild the DeclRefExpr
-        // with Sema::BuildDeclRefExpr. This is required in some cases, e.g.
-        // Sema::BuildDeclRefExpr is responsible for adding captured fields
-        // to the underlying struct of a lambda.
-        if (clonedDRE->getDecl()->getDeclContext() != m_Sema.CurContext) {
-          auto referencedDecl = cast<VarDecl>(clonedDRE->getDecl());
-          clonedDRE = cast<DeclRefExpr>(BuildDeclRef(referencedDecl));
-        }
+      // If current context is different than the context of the original
```
vgvassilev: For future reference: the blankspace changes should be made only if we touch the area or in a separate commit.

lib/Differentiator/StmtClone.cpp
```diff
@@ -342,9 +341,9 @@ ReferencesUpdater::ReferencesUpdater(Sema& SemaRef, utils::StmtClone* C, Scope*
 bool ReferencesUpdater::VisitDeclRefExpr(DeclRefExpr* DRE) {
   // If the declaration's decl context encloses the derivative's decl
   // context we must not update anything.
-  if (DRE->getDecl()->getDeclContext()->Encloses(m_Sema.CurContext)) {
-    return true;
-  }
+  //if (DRE->getDecl()->getDeclContext()->Encloses(m_Sema.CurContext)) {
```
vgvassilev: Why did we comment that out?

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -1396,44 +1394,56 @@ namespace clad {
...
     // Function body scope.
     beginScope(Scope::FnScope | Scope::DeclScope);
     beginBlock();
     // Start the visitation process which outputs the statements in the current
     // block.
-    Stmt* functionBody = m_Function->getMostRecentDecl()->getBody();
-    Visit(functionBody, dfdf);
+    Stmt* BodyDiffForward = Visit(FD->getBody()).getStmt();
+    if (isa<CompoundStmt>(BodyDiffForward))
+      for (Stmt* S : cast<CompoundStmt>(BodyDiffForward)->body())
```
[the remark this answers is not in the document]
efremale: It's like an iterator range, it can be empty (i.e. if we have an empty block: {}), but it's never null.

lib/Differentiator/StmtClone.cpp
efremale: Previously, we only had to clone statements that were created in the context of the original function. In the reverse mode, I also have to clone statements that I created myself in the context of the derivative function, so I have to ignore that condition. I think we can remove it.

PR #111, opened 2019-03-24
Fix converity issues

vgvassilev:


include/clad/Differentiator/DiffPlanner.h
```diff
@@ -14,6 +14,7 @@ namespace clang {
 namespace clad {
 
   enum class DiffMode {
+    unknown = 0,
```
efremale: What is the point of having that? It can never be in a valid but 'unknown' state. Is it worth it to introduce this solely for the purpose of calming down the static analyzer? Why not just initialize it to be 'forward'?

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -29,12 +30,13 @@ namespace clad {
     unsigned m_RequestedDerivativeOrder = 1;
     unsigned m_CurrentDerivativeOrder = 1;
     clang::Expr* m_DiffArgs = nullptr;
-    DiffMode m_Mode;
+    DiffMode m_Mode = DiffMode::unknown;
```
efremale: See previous comment.

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -29,12 +30,13 @@ namespace clad {
...
   public:
     typedef Functions::iterator iterator;
     typedef Functions::const_iterator const_iterator;
 
     DiffMode getMode() const {
+      assert(m_Mode != DiffMode::unknown && "Must be set!");
```
efremale: See previous comment.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -834,6 +834,7 @@ namespace clad {
         clonedDRE = BuildDeclRef(it->second);
       else
         clonedDRE = cast<DeclRefExpr>(Clone(DRE));
+      assert(clonedDRE);
```
efremale: Can it ever be null? BuildDeclRef returns the result of VarDecl::Create (can it be null?), cast will trigger the assert if it is given a null input.

include/clad/Differentiator/DiffPlanner.h
vgvassilev: That's what I thought initially, however, having an `unknown` flag protects against errors in programming. This essentially will protect you to call `getMode` before it was set.

efremale: I'd prefer to prohibit default initialization at all (i.e. only allow constructor of DiffPlan with given DiffMode). But I guess yours approach is valid too.

vgvassilev: That's better, I will do that.

vgvassilev: Looks like that'd be more refactoring than what I am willing to make at this point. Let's keep it as is.

Issue #116, opened 2019-03-29
Fix nested function calls in the forward mode


efremale:
At the moment, calls in the forward mode work correctly only if called function has one input.
If a function `f` is called with args `a1, a2, ...`, its derivative should be a dot product of `gradient(f)(a1, a2, ...)` and a vector of derivatives of args: `{ da1, da2, ...}`.

This dot product can be efficiently computed by differentiating `f` in the forward mode and providing values of `da1, da2, ...` as seeds. See issue #100.

vaithak:
Similar to #100, this was fixed in https://github.com/vgvassilev/clad/pull/565 and many subsequent PRs which added the vector forward mode support.

We also have many test cases present with the vector mode interface: https://github.com/vgvassilev/clad/blob/master/test/ForwardMode/VectorMode.C.
Demo example: https://github.com/vgvassilev/clad/blob/master/demos/VectorForwardMode.cpp

vgvassilev:
Thanks @vaithak, let's close the issue now.

PR #117, opened 2019-03-29
Enable recursive differentiation of nested calls

efremale:
Now, if a function `f` that is being differentiated contains a call to a function `g`, clad will proceed to try to differentiate `g` even if it has no
overload in `custom_derivatives`.

Example:
```
double f(double x) {
  return g(double x);
}
```
On `clad::differentiate(f, 0)`, clad will proceed to differentiate `g` and so on, even if it is not in `custom_derivatives` namespace.

A new issue with forward mode calls was detected: #116, we should fix it. Some tests were disabled.

### DIFF
--- a/include/clad/Differentiator/DiffPlanner.h
+++ b/include/clad/Differentiator/DiffPlanner.h
@@ -19,72 +19,49 @@ namespace clad {
     reverse
   };
 
-  ///\brief The list of the dependent functions which also need differentiation
-  /// because they are called by the function we are asked to differentitate.
-  ///
-  class DiffPlan {
-  private:
-    typedef llvm::SmallVector<clang::FunctionDecl*, 16> Functions;
-    Functions m_Functions;
-    clang::CallExpr* m_CallToUpdate = nullptr;
-    unsigned m_RequestedDerivativeOrder = 1;
-    unsigned m_CurrentDerivativeOrder = 1;
-    clang::Expr* m_DiffArgs = nullptr;
-    DiffMode m_Mode = DiffMode::unknown;
-  public:
-    typedef Functions::iterator iterator;
-    typedef Functions::const_iterator const_iterator;
+  /// A struct containing information about request to differentiate a function.
+  struct DiffRequest {
+    /// Function to be differentiated.
+    const clang::FunctionDecl* Function = nullptr;
+    /// Name of the base function to be differentiated. Can be different from
+    /// function->getNameAsString() when higher-order derivatives are computed.
+    std::string BaseFunctionName = {};
+    /// Current derivative order to be computed.
+    unsigned CurrentDerivativeOrder = 1;
+    /// Highest requested derivative order.
+    unsigned RequestedDerivativeOrder = 1;
+    /// Context in which the function is being called, or a call to
+    /// clad::gradient/differentiate, where function is the first arg.
+    clang::CallExpr* CallContext = nullptr;
+    /// Args provided to the call to clad::gradient/differentiate.
+    const clang::Expr* Args = nullptr;
+    /// Requested differentiation mode, forward or reverse.
+    DiffMode Mode = DiffMode::unknown;
+    /// If function appears in the call to clad::gradient/differentiate,
+    /// the call must be updated and the first arg replaced by the derivative.
+    bool CallUpdateRequired = false;
+    /// A flag to enable/disable diag warnings/errors during differentiation.
+    bool VerboseDiags = false;
 
-    DiffMode getMode() const {
-      assert(m_Mode != DiffMode::unknown && "Must be set!");
-      return m_Mode;
-    }
-    void setMode(DiffMode mode) {
-      m_Mode = mode;
-    }
-    unsigned getRequestedDerivativeOrder() const {
-      return m_RequestedDerivativeOrder;
-   }
-    void setCurrentDerivativeOrder(unsigned val) {
-      m_CurrentDerivativeOrder = val;
-    }
-    unsigned getCurrentDerivativeOrder() const {
-      return m_CurrentDerivativeOrder;
-    }
-    void push_back(clang::FunctionDecl* FD) { m_Functions.push_back(FD); }
-    iterator begin() { return m_Functions.begin(); }
-    iterator end() { return m_Functions.end(); }
-    const_iterator begin() const { return m_Functions.begin(); }
-    const_iterator end() const { return m_Functions.end(); }
-    size_t size() const { return m_Functions.size(); }
-    void setCallToUpdate(clang::CallExpr* CE) { m_CallToUpdate = CE; }
     void updateCall(clang::FunctionDecl* FD, clang::Sema& SemaRef);
-    clang::Expr* getArgs() const { return m_DiffArgs; }
-    LLVM_DUMP_METHOD void dump();
-
-    friend class DiffCollector;
   };
 
-  typedef llvm::SmallVector<DiffPlan, 16> DiffPlans;
+  using DiffSchedule = llvm::SmallVector<DiffRequest, 16>;
 
   class DiffCollector: public clang::RecursiveASTVisitor<DiffCollector> {
   private:
     ///\brief The diff step-by-step plan for differentiation.
     ///
-    DiffPlans& m_DiffPlans;
+    DiffSchedule& m_DiffPlans;
 
     ///\brief If set it means that we need to find the called functions and
     /// add them for implicit diff.
     ///
-    clang::FunctionDecl* m_TopMostFD;
-
+    const clang::FunctionDecl* m_TopMostFD = nullptr;
     clang::Sema& m_Sema;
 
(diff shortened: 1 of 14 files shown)


test/FirstDerivative/Recursive.C
```diff
@@ -7,28 +7,46 @@
...
+//cCHECK: int f_pow_darg0(int arg, int p) {
+//cCHECK-NEXT:   int _d_arg = 1;
+//cCHECK-NEXT:   int _d_p = 0;
+//cCHECK-NEXT:   if (p == 0)
+//cCHECK-NEXT:     return 0;
+//cCHECK-NEXT:   else {
+//cCHECK-NEXT:     int _t0 = f_pow(arg, p - 1);
+//cCHECK-NEXT:     return _d_arg * _t0 + arg * (f_pow_darg0(arg, p - 1) * (_d_arg + _d_p - 0));
+//cCHECK-NEXT:   }
+//cCHECK-NEXT: }
```
efremale: This test is broken atm and was disabled.

test/FirstDerivative/CompoundAssignments.C
```diff
@@ -91,17 +91,17 @@ double f5(double x, double y) {
...
+//cCHECK-NEXT:       double _d_x = 1;
+//cCHECK-NEXT:       double _d_y = 0;
+//cCHECK-NEXT:       double _t0 = std::pow(2., y);
+//cCHECK-NEXT:       _d_x = (_d_x * _t0 - x * (custom_derivatives::pow_darg0(2., y) * (0. + _d_y))) / (_t0 * _t0);
+//cCHECK-NEXT:       x /= _t0;
+//cCHECK-NEXT:       double _t1 = std::pow(2., y);
+//cCHECK-NEXT:       _d_x = _d_x * _t1 + x * (custom_derivatives::pow_darg0(2., y) * (0. + _d_y));
+//cCHECK-NEXT:       x *= _t1;
+//cCHECK-NEXT:       return _d_x;
+//cCHECK-NEXT:   }
```
efremale: This test is broken atm and was disabled.

test/FirstDerivative/Recursive.C
vgvassilev: Could you prepend FIXME- and open an issue for this?

efremale: OK, the issue is #116

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -19,72 +19,49 @@ namespace clad {
...
-    typedef Functions::iterator iterator;
-    typedef Functions::const_iterator const_iterator;
+  /// A struct containing information about request to differentiate a function.
+  struct DiffRequest {
+    /// Function to be differentiated.
+    const clang::FunctionDecl* function = nullptr;
+    /// Name of the base function to be differentiated. Can be different from
+    /// function->getNameAsString() when higher-order derivatives are computed.
+    std::string baseFunctionName = {};
+    /// Current derivative order to be computer.
```
vgvassilev:
```suggestion
    /// Current derivative order to be computed.
```

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -424,9 +435,10 @@ namespace clad {
     NamespaceDecl* enclosingNS = nullptr;
     llvm::SaveAndRestore<DeclContext*> SaveContext(m_Sema.CurContext);
     llvm::SaveAndRestore<Scope*> SaveScope(m_CurScope);
-    m_Sema.CurContext = m_Function->getDeclContext();
+    DeclContext* DC = const_cast<DeclContext*>(m_Function->getDeclContext());
```
vgvassilev: We should make  m_Function non const.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -1339,9 +1315,10 @@ namespace clad {
...
```
vgvassilev: Likewise

test/FirstDerivative/CallArguments.C
```diff
@@ -96,7 +96,7 @@ float f_const_args_func_6(const float x, const float y, const Vec &v) {
 
 float f_const_helper(const float x) {
   return x * x;
-} // expected-warning 4 {{function 'f_const_helper' was not differentiated because it is not declared in namespace 'custom_derivatives'}}
+}
```
vgvassilev: Nice!

test/FirstDerivative/CodeGenSimple.C
```diff
@@ -4,9 +4,10 @@
 //CHECK-NOT: {{.*error|warning|note:.*}}
 
 #include "clad/Differentiator/Differentiator.h"
-extern "C" int printf(const char* fmt, ...); //expected-warning{{function 'printf' was not differentiated because it is not declared in namespace 'custom_derivatives'}}
+extern "C" int printf(const char* fmt, ...);
+
 int f_1(int x) {
-   printf("I am being run!\n"); //expected-warning{{attempted to differentiate unsupported statement, no changes applied}}
+   printf("I am being run!\n"); //expected-warning{{attempted to differentiate unsupported statement, no changes applied}} //expected-warning{{function 'printf' was not differentiated because clad failed to differentiate it and no suitable overload was found in namespace 'custom_derivatives'}}
```
vgvassilev: We should shorten the diagnostic and make it more accurate. What about “function ... was not differentiated because no definition is available and no suitable replacement ...”?

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -19,72 +19,49 @@ namespace clad {
...
-    unsigned m_CurrentDerivativeOrder = 1;
-    clang::Expr* m_DiffArgs = nullptr;
-    DiffMode m_Mode = DiffMode::unknown;
-  public:
-    typedef Functions::iterator iterator;
-    typedef Functions::const_iterator const_iterator;
+  /// A struct containing information about request to differentiate a function.
+  struct DiffRequest {
+    /// Function to be differentiated.
+    const clang::FunctionDecl* function = nullptr;
```
vgvassilev: We should probably either add m_ to all members or at least capitalize them.

efremale: What is wrong with `request.function`? It is the cleanest way IMO. It is just a struct carrying parameters.

vgvassilev: We seem to be using different pattern elsewhere.

lib/Differentiator/DerivativeBuilder.cpp
efremale: It is not really a solution, since we are assigning `FD` (`const FunctionDecl*` passed to `::Derive`, which is indeed intended to be const, since we are not going to modify the original declaration) to `m_Function`. So that will just move the `const_cast` elsewhere.

It is more reasonable to remove constness from the `DeclContext` since we are indeed going to modify it (e.g. by putting new derivative into the original function's namespace).

lib/Differentiator/DerivativeBuilder.cpp
efremale: See above.

test/FirstDerivative/CodeGenSimple.C
efremale: It is not necessarily that there is no definition. There may be a definition but clad may fail to derive it (e.g. if it meets something it doesn't support).

include/clad/Differentiator/DiffPlanner.h
efremale: It is questionable why are we using m_ thing in the first place. Maybe it makes some sense in classes where only a small part is exposed as a public interface. It doesn't make a lot of sense to use it on a config struct where members are intended to be assigned with something as a public interface.

vgvassilev: We should have accessors but that'd be an overkill for that. Let's use llvm's convention capitalizing the data members.

test/FirstDerivative/CodeGenSimple.C
vgvassilev: I think In that case we should diagnose that in a separate diagnostic.

include/clad/Differentiator/DiffPlanner.h
efremale: Accessors like `request.setCurrentDerivativeOrder(request.getCurrentDerivativeOrder() + 1)`? I don't think we should ever have that.

test/FirstDerivative/CodeGenSimple.C
efremale: That is not so easy to do at the moment and we'll need to work on some diagnostic infrastructure to propagate exact and clear errors from recursive calls to `::Derive`.

vgvassilev: Ok, could you open an issue for that then?

efremale: #119

PR #120, opened 2019-04-03
Add support for arrays inside functions

efremale:
Now arrays can be uses inside functions in both forward and reverse mode.

For now it only works with usual static C arrays (i.e. no custom datastructures like
`std::array`, `std::vector`, no `malloc`, `new`).

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -574,6 +574,26 @@ namespace clad {
       return m_Sema.ActOnInitList(noLoc, {}, noLoc).get();
   }
 
+  std::pair<const clang::Expr*, llvm::SmallVector<const clang::Expr*, 4>>
+  VisitorBase::SplitArraySubscript(const ArraySubscriptExpr* ASE) {
+    llvm::SmallVector<const clang::Expr*, 4> Indices{};
+    const Expr* E = ASE;
+    while (auto S = dyn_cast<ArraySubscriptExpr>(E)) {
+      Indices.push_back(S->getIdx());
+      E = S->getBase()->IgnoreImpCasts();
+    }
+    std::reverse(std::begin(Indices), std::end(Indices));
+    return std::make_pair(E, std::move(Indices));
+  }
+
+  Expr* VisitorBase::BuildArraySubscript(Expr* Base,
+    const llvm::SmallVector<clang::Expr*, 4>& Indices) {
+    Expr* result = Base;
+    for (Expr* I : Indices)
+      result = m_Sema.CreateBuiltinArraySubscriptExpr(result, noLoc, I, noLoc).get();
+    return result;
+  }
+
   StmtDiff ForwardModeVisitor::VisitStmt(const Stmt* S) {
     diag(DiagnosticsEngine::Warning, S->getLocStart(),
          "attempted to differentiate unsupported statement, no changes applied");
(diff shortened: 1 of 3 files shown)


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -836,6 +863,55 @@ namespace clad {
                     ConstantFolder::synthesizeLiteral(Ty, m_Context, 0));
   }
 
+  StmtDiff ForwardModeVisitor::VisitInitListExpr(const InitListExpr* ILE) {
+    llvm::SmallVector<Expr*, 16> clonedExprs(ILE->getNumInits());
+    llvm::SmallVector<Expr*, 16> derivedExprs(ILE->getNumInits());
+    for (unsigned i = 0; i < ILE->getNumInits(); i++) {
```
```suggestion
    for (unsigned i = 0, e = ILE->getNumInits(); i < e; i++) {
```
This way we avoid recomputing the number N times and show the reader that the loop does not mutate the number of the inits of ILE.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -836,6 +863,55 @@ namespace clad {
...
+    const auto& Indices = ASI.second;
+    Expr* clonedBase = Clone(Base);
+    llvm::SmallVector<Expr*, 4> clonedIndices(Indices.size());
+    std::transform(std::begin(Indices), std::end(Indices), std::begin(clonedIndices),
+      [this](const Expr* E) { return Clone(E); });
+    auto cloned = BuildArraySubscript(clonedBase, clonedIndices);
+
+    DeclRefExpr* DRE = dyn_cast<DeclRefExpr>(clonedBase->IgnoreParenImpCasts());
+    auto zero = ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
+    if (!DRE)
```
vgvassilev: Could we use early returns instead? For instance,
 ```if (!isa<DeclRefExpr>(clonedBase->IgnoreParenImpCasts()) return return StmtDiff(cloned, zero);
  auto DRE = cast<DeclRefExpr>(...);
```

File: lib/Differentiator/DerivativeBuilder.cpp [hunk 3]
```diff
@@ -836,6 +863,55 @@ namespace clad {
...
+      return StmtDiff(cloned, zero);
+    auto VD = dyn_cast<VarDecl>(DRE->getDecl());
+    if (!VD)
```

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -1595,11 +1671,31 @@ namespace clad {
     return StmtDiff(BuildParens(subStmtDiff.getExpr()));
   }
 
+  StmtDiff ReverseModeVisitor::VisitInitListExpr(const InitListExpr* ILE) {
+    llvm::SmallVector<Expr*, 16> clonedExprs(ILE->getNumInits());
+    for (unsigned i = 0; i < ILE->getNumInits(); i++) {
```
vgvassilev: Likewise, assuming we do not mutate the number of inits.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -1595,11 +1671,31 @@ namespace clad {
...
+    auto ASI = SplitArraySubscript(ASE);
+    const Expr* Base = ASI.first;
+    const auto& Indices = ASI.second;
+    Expr* clonedBase = Clone(Base);
+    llvm::SmallVector<Expr*, 4> clonedIndices(Indices.size());
+    std::transform(std::begin(Indices), std::end(Indices), std::begin(clonedIndices),
+      [this](const Expr* E) { return Clone(E); });
+    auto cloned = BuildArraySubscript(clonedBase, clonedIndices);
+    
+    DeclRefExpr* DRE = dyn_cast<DeclRefExpr>(clonedBase->IgnoreParenImpCasts());
```
vgvassilev: Likewise. We should really figure out where to extract this code, we are duplicating quite a lot of code.

Issue #94, opened 2018-09-28
Reduce cloning complexity in reverse mode.


vgvassilev:
Extends the work done in #47

efremale:
Solved in #121

PR #121, opened 2019-04-24
Implement reverse pass, store forward pass subexpressions, reduce cloning complexity

efremale:
This commit implements several things:

1. Implement proper forward and reverse pass in the reverse mode
   * The forward pass does all the computations of the original function
and stores them. Reverse pass computes the derivatives using stored
values from the forward pass. To solve the issue with local variables
and scopes, forward pass stores intermediate values in variables
declared in the beginning of derivative function's scope.
   * While separation into forward and reverse pass is not required when
there are no control-flow dependent variable reassignments, it is
crucial if we want to support reassignments (not implemented yet).
   * Since C++ functions may have several control-flow based points of
return, this requires special treatment. When forward pass contains
`return` statement that is not the last statement in the pass,
reverse computation should not start in the beginning of the reverse
pass, but at some further point. Previously, this was implemented by
cloning the corresponding "tail" of the reverse pass and putting that
tail immediately after the point of return. This is not efficient, as
it requires to clone the reverse pass once per each `return` statement
in function's body. Now, we only have a singe forward and reverse pass,
and preliminary return from a forward pass is implemented as a `goto`
statement to a corresponding label in the reverse pass.

2. Store intermediate expression results in temporary variables in the
forward pass of the reverse mode
   * As it was done before for the forward mode, intermediate results of
expressions of the forward pass are now stored in intermediate
variables, which are later reused in both forward and reverse pass.
This allows to avoid recomputing repeated expressions and cloning long
expression trees. Previosly, in the reverse mode this was done only for
"reverse" part, now it is also enabled in the forward pass.

3. Reduce cloning complexity in the reverse pass
   * Now the complexity of building the gradient function via the
reverse mode should be linear instead of quadratic, as the issue of
repeated cloning of expression trees is solved. This solves the issue
#94.


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -1787,9 +1814,30 @@ namespace clad {
...
+    // Save current index in the current block, to potentially put some statements
+    // there later.
+    std::size_t insertionPoint = getCurrentBlock(reverse).size();
+    for (const Expr* Arg : CE->arguments()) {
+      // Create temporary variables corresponding to derivative of each argument,
+      // so that they can be reffered to when arguments is visited. Variables
+      // will be initialized later after aruments is visited. This is done to
+      // reduce cloning complexity and only clone once.
+      Expr* dArg = StoreAndRef(nullptr, Arg->getType(), reverse, "_r", /*force*/true);
+      ArgResultDecls.push_back(cast<VarDecl>(cast<DeclRefExpr>(dArg)->getDecl()));
```
vgvassilev: Are we guaranteed to only have declrefs over vardecls?

test/FirstDerivative/BasicArithmeticAddSub.C
```diff
@@ -1,4 +1,4 @@
-// RUN: %cladclang %s -I%S/../../include -oBasicArithmeticAddSub.out 2>&1 | FileCheck %s
+// RUN: %cladclang %s -lm -I%S/../../include -oBasicArithmeticAddSub.out 2>&1 | FileCheck %s
```
vgvassilev: Why do we need to link against libm?

test/FirstDerivative/DependencyTracking.C
```diff
@@ -1,4 +1,4 @@
-// RUN: %cladclang %s -I%S/../../include -fsyntax-only 2>&1 | FileCheck %s
+// RUN: %cladclang %s -lm -I%S/../../include 2>&1 | FileCheck %s
```
vgvassilev: Did this stop failing and that’s why you remove -fsyntax-only mode? If so please remove the xfail directive, too.

test/FirstDerivative/FunctionsInNamespaces.C
```diff
@@ -1,4 +1,4 @@
-// RUN: %cladclang %s -I%S/../../include -fsyntax-only -Xclang -verify 2>&1 | FileCheck %s
+// RUN: %cladclang %s -I%S/../../include -Xclang -verify 2>&1 | FileCheck %s
```
vgvassilev: Likewise

lib/Differentiator/DerivativeBuilder.cpp
efremale: Yes, in that case the result of `StoreAndRef` with `force = true` is always a `DeclRefExpr` to a `VarDecl`.

test/FirstDerivative/DependencyTracking.C
efremale: It seems -fsyntax-only prevents us from catching linker errors: if we forward-declare f_darg0 and use it somewhere, but clad fails to create the definition. Without -fsyntax-only we'll get a linker error.

test/FirstDerivative/BasicArithmeticAddSub.C
efremale: I will revert changes related to added -lm: it is not necessary. At some point I changed the definitions in BuiltinDerivatives.h from templates to actual functions (and they have cmath inside), then -lm was always required. But I reverted it.

PR #123, opened 2019-04-25
Use `bool` instead of `_Bool` when printing generated code

efremale:
This solves the issue of copying and compiling generated code
containing `_Bool`, as `_Bool` could not be compiled. Now `bool` is
printed.

### DIFF
--- a/test/Gradient/Gradients.C
+++ b/test/Gradient/Gradients.C
@@ -331,7 +331,7 @@ double f_cond1(double x, double y) {
 }
 
 //CHECK:   void f_cond1_grad(double x, double y, double *_result) {
-//CHECK-NEXT:       _Bool _cond0;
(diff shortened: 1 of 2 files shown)


vgvassilev:
Lgtm after landing pr #121. Maybe branch off the master instead as this does not depend on #121

test/FirstDerivative/BasicArithmeticAddSub.C
```diff
@@ -1,4 +1,4 @@
-// RUN: %cladclang %s -I%S/../../include -oBasicArithmeticAddSub.out 2>&1 | FileCheck %s
+// RUN: %cladclang %s -lm -I%S/../../include -oBasicArithmeticAddSub.out 2>&1 | FileCheck %s
```
vgvassilev: Looks like we brought again -lm

PR #124, opened 2019-04-29
Implement variable reassignments in the reverse mode

efremale:
Now the reverse mode supports variable reassignments, potentially depending on control-flow (e.g. IfStmts). 
At the moment it is assumed that each assignment statement is executed at most once (i.e. it is not inside a loop/backwards-pointing goto), as otherwise it is reqired to store result of the assignment once per each execution (e.g. in a stack).

### DIFF
--- a/test/Gradient/Assignments.C
+++ b/test/Gradient/Assignments.C
@@ -1,0 +1,264 @@
+// RUN: %cladclang %s -I%S/../../include -oReverseAssignments.out 2>&1 | FileCheck %s
+// RUN: ./ReverseAssignments.out | FileCheck -check-prefix=CHECK-EXEC %s
+//CHECK-NOT: {{.*error|warning|note:.*}}
+
+#include "clad/Differentiator/Differentiator.h"
+#include <cmath>
+
+// = y
+double f1(double x, double y) {
+  x = y;
+  return y;
+}
+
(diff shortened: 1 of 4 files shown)


test/Gradient/Assignments.C
```diff
@@ -0,0 +1,266 @@
+// RUN: %cladclang %s -lm -I%S/../../include -oReverseAssignments.out 2>&1 | FileCheck %s
```
vgvassilev: Do we still need `-lm`?

test/Gradient/Assignments.C
```diff
@@ -0,0 +1,266 @@
...
+  double result[2] = {};
+  TEST(f1, 3, 4); // CHECK-EXEC: {0.00, 1.00}
+  TEST(f2, 3, 4); // CHECK-EXEC: {0.00, 1.00} 
+  TEST(f3, 2, 100); // CHECK-EXEC: {32.00, 0.00} 
+  TEST(f4, 3, 4); // CHECK-EXEC: {1.00, 0.00} 
+  TEST(f5, -3, 4); // NOT-CHECK-EXEC: {-6.00, 0.00} 
+  TEST(f6, 3, -4); // CHECK-EXEC: {-6.00, 0.00} 
+}
+
+
```
vgvassilev: Extra line ;)

PR #126, opened 2019-05-08
Add missing operators to the reverse mode, improve assignments, add support for nested assignments, fix issues with storing expressions with side-effects, optimize visiting independent expressions

efremale:
Now reverse mode supports operators `+=`, `-=`, `*=`, `/=`, `,`, `++`, `--`.

Assignments to array subscripts are now supported.

Now we also support nested assignments, even as parts of other expressions, like
`a = b * ((c ? d : e) = f = g);`

Storing expressions with side effects (e.g. assignments like `a *= b`) is fixed. The problem was that the expression was stored once and then reused
again in the forward pass, causing it to be executed twice, like:
```
...
_t0 = (x *= y)
...
t = (x *= y) * z
```
which lead to `x` being  updated twice and gave incorrect results in some cases. Now the stored temporary variable is used instead:
```
...
_t0 = (x *= y)
...
t = _t0 * z
```

Additionally, passing `dfdx()` as a parameter for `Visit` was optimized. It is no longer pushed on the stack repeatedly. Also, when `dfdx()` is zero,
derivatives are no longer updated with zero expressions, since it has no effect anyway. Thus we avoid creating useless expressions sometimes.

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -293,7 +293,9 @@ namespace clad {
   }
 
   DeclRefExpr* VisitorBase::BuildDeclRef(DeclaratorDecl* D) {
-    Expr* DRE = m_Sema.BuildDeclRefExpr(D, D->getType(), VK_LValue, noLoc).get();
+    QualType T = D->getType();
+    T = T.getNonReferenceType();
+    Expr* DRE = m_Sema.BuildDeclRefExpr(D, T, VK_LValue, noLoc).get();
     return cast<DeclRefExpr>(DRE);
   }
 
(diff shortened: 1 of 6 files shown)


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -1486,6 +1490,11 @@ namespace clad {
 
     auto idx = 0;
     for (auto arg : args) {
+      // FIXME: fix when adding array inputs
```
vgvassilev: Could you elaborate?

test/Gradient/Gradients.C
```diff
@@ -236,7 +236,7 @@ double f_c(double x, double y) {
 
 //CHECK:   void f_c_grad(double x, double y, double *_result) {
 //CHECK-NEXT:       double _t0;
-//CHECK-NEXT:       double _t1;
+//CHECK-NEXT:       double _t1;
```
vgvassilev: Space ;)

lib/Differentiator/DerivativeBuilder.cpp
efremale: Now we are just skipping all array/pointer inputs (not treating them as independent variables), since we don't have a good way of storing their derivatives yet. When we add this functionality, we should remove the following if-stmt. 

I will add a comment into the code.

Issue #89, opened 2018-08-20
Avoid storing simple expressions like -x in temporaries


efremale:
Example:
```
double f(double x) {
  double t = 2 * (-x);
  return t;
}
```
->
```
double f_darg0(double x) {
    double _t0 = -x;
    double _d_t = 0 * _t0 + 2 * -1.;
    double t = 2 * _t0;
    return _d_t;
}
```

In this case, `double _t0 = -x` seems useless so it makes sense to add such simple unary operators to a list of cases for which temporaries are not generated (now it is literals and references). On the other hand, results of unary operators like `++x` must be stored to avoid repeated increment.

efremale:
Fixed in #129

Issue #128, opened 2019-05-12
Fix potential issues with side effects in conditional operators in forward mode


efremale:
In C++, in conditional operators like `(c ? a : b)` only `a` or `b` is executed, depending on `c`. We did not account for this when visiting in forward mode, therefore side effects of both `a` and `b` could be applied, independent of `c`.

E.g., for an expression like `(c ? (x = a) * b : y)`, a temporary `double t = (x = a);` will be created, making `x = a` to be executed, even when `c` is false.

PR #129, opened 2019-05-12
Optimize stores, fix issues with compound statements in reverse mode

efremale:
Several optimizations for `StoreAndRef` and `GlobalStoreAndRef` are implemented.
In `StoreAndRef`, useless temporaries are no longer created for expressions like `+x`, `-x` and array subscripts. This fixed issue #89.
In `GlobalStoreAndRef`, constant literals are no longer stored. This will be useful when implementing for-loops to avoid stores into stack.

Additionally, in the reverse mode, an issue when visiting conditional operators is fixed:
In C++, in conditional operators like `(c ? a : b)` only `a` or `b` is executed, depending on `c`. We did not account for this when visiting, therefore
side effects of both `a` and `b` could be applied, independent of `c`. Now, an IfStmt is created and only one side will be evaluated. The issue is still present in forward mode: issue #128.

### DIFF
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -133,7 +133,7 @@ namespace clad {
     /// A stack of all the blocks where the statements of the gradient function
     /// are stored (e.g., function body, if statement blocks).
     std::vector<Stmts> m_Blocks;
-
+  public:
     template <typename Range>
     clang::CompoundStmt* MakeCompoundStmt(const Range & Stmts) {
       auto Stmts_ref = llvm::makeArrayRef(Stmts.data(), Stmts.size());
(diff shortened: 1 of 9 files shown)


include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -484,8 +489,33 @@ namespace clad {
     clang::Expr* GlobalStoreAndRef(clang::Expr* E,
                                    llvm::StringRef prefix = "_t");
 
-    //// A reference to the output parameter of the gradient function.
-    clang::Expr* m_Result;
+    struct DelayedStoreResult {
```
vgvassilev: Could we default initialize the members of the struct?

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -540,6 +570,12 @@ namespace clad {
     /// order is maintained.
     StmtDiff DifferentiateSingleStmt(const clang::Stmt* S,
                                      clang::Expr* expr = nullptr);
+    /// A helper method used to keep substatements created by Visit(E, expr) in
+    /// separate forward/reverse blocks instead of putting them into current
+    /// blocks. First result is a StmtDiff of forward/reverse blocks with 
+    /// additionally created Stmts, second is a direct result of call to Visit.
+    std::pair<StmtDiff, StmtDiff> 
+    DifferentiateSingleExpr(const clang::Expr* E, clang::Expr* expr = nullptr);
```
vgvassilev: Could we use more meaningful param names? Waht is `expr`?

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -346,16 +346,30 @@ namespace clad {
...
-    return StoreAndRef(E, E->getType(), block, prefix, forceDeclCreation);
+    QualType Type = E->getType();
+    if (E->isModifiableLvalue(m_Context) == Expr::MLV_Valid)
+      Type = m_Context.getLValueReferenceType(Type);
+    return StoreAndRef(E, Type, block, prefix, forceDeclCreation);
   }
- 
-  bool UsefulToStore(Expr* E) {
+
+  static bool UsefulToStore(Expr* E) {
```
vgvassilev: Can we add some more documentation to this routine?

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -2288,6 +2347,25 @@ namespace clad {
...
 
+  std::pair<StmtDiff, StmtDiff> ReverseModeVisitor::DifferentiateSingleExpr(
+    const Expr* E, Expr* expr) {
+    beginBlock(forward);
+    beginBlock(reverse);
+    StmtDiff EDiff = Visit(E, expr);
+    CompoundStmt* RCS = endBlock(reverse);
+    Stmt* ForwardResult = endBlock(forward);
+    Stmt* ReverseResult = nullptr;
+    if (RCS->body_empty())
```
vgvassilev: Can't we use `unwrapIfSingleStmt` to simplify this block?

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -2345,8 +2423,32 @@ namespace clad {
     return StmtDiff(Clone(ME));
   }
 
-  Expr* ReverseModeVisitor::GlobalStoreAndRef(Expr* E, QualType Type,
-                                              llvm::StringRef prefix) {
+  static bool UsefulToStoreGlobal(Expr* E) {
+    // TODO: it may be worth to be more strict and use this, when storing in a
```
vgvassilev: Could we use a single greppable form eg: TODO->FIXME.

include/clad/Differentiator/DerivativeBuilder.h
efremale: It just a wrapper that forwards its arguments to `Visit(E, expr)`, where `expr` is current accumulated derivative w.r.t. `E`. I don't think I have a better explanation.

include/clad/Differentiator/DerivativeBuilder.h
efremale: I do not think it is a good idea, having default member initializers before C++14 does not allow to use aggregate initialization, and makes us write constructors. This class is not intended to be default-initialized, it is allways constructed with all members provided.

include/clad/Differentiator/DerivativeBuilder.h
vgvassilev: I meant to ask to change the parameter name to dfdx or smth like that.

PR #131, opened 2019-05-16
Enable differentiation of for-loops in the reverse mode

efremale:


include/clad/Differentiator/Tape.h
```diff
@@ -0,0 +1,99 @@
...
+#define CLAD_TAPE_H
+
+#include <cassert>
+#include <memory>
+#include <type_traits>
+#include <utility>
+
+namespace clad {
+  template <typename T>
+  class tape_impl {
```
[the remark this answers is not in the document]
vgvassilev: We should probably force constructing a `clad::tape` with an argument giving a guesstimate of the capacity. Most of the loops go far beyond 32 and we want to reduce the redundant reallocations.

include/clad/Differentiator/Tape.h
```diff
@@ -0,0 +1,99 @@
...
+    void pop_back() {
+      assert(_size);
+      _size -= 1;
+      end()->~T();
+    }
+
+  private:
+    constexpr static std::size_t _init_capacity = 32;
+    void grow() {
+      if (!_capacity)
```
vgvassilev:
```suggestion
      if (LLVM_UNLIKELY(!_capacity))
```

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -618,6 +623,118 @@ namespace clad {
...
+    DeclarationName Name = &m_Context.Idents.get(name);
+    LookupResult R(m_Sema, Name, noLoc, Sema::LookupOrdinaryName);
+    m_Sema.LookupQualifiedName(R, CladNS, CSS);
+    assert(!R.empty() && isa<FunctionTemplateDecl>(R.getRepresentativeDecl()) &&
+           "cannot find requested name");
+    return R;
+  }
+
+  LookupResult& VisitorBase::GetCladTapePush() {
+    static llvm::Optional<LookupResult> Result{};
```
vgvassilev: Why not just `LookupResult` and checking if `!empty`. We do not expect not finding the `clad::tape` entities, so we would have stopped before that point (in debug) and crashed later (in release).

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -618,6 +623,118 @@ namespace clad {
...
+
+  LookupResult& VisitorBase::GetCladTapeBack() {
+    static llvm::Optional<LookupResult> Result{};
+    if (Result)
+      return Result.getValue();
+    Result = LookupCladTapeMethod("back");
+    return Result.getValue();
+  }
+
+  QualType VisitorBase::GetCladTapeOfType(QualType T) {
```
vgvassilev: Could you document this routine a little better? Is the NestedNameSpecifier forcing an instantiation? I'd expect `ActOnCallExpr` in the caller to instantiate.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -618,6 +623,118 @@ namespace clad {
...
+    Expr* BackDRE = V.m_Sema.BuildDeclarationNameExpr(CSS, Back,
+                                                      /*ADL*/ false).get();
+    Expr* Call = V.m_Sema.ActOnCallExpr(V.getCurrentScope(), BackDRE, noLoc, Ref,
+                                        noLoc).get();
+    return Call;
+  }
+
+  ReverseModeVisitor::CladTapeResult ReverseModeVisitor::MakeCladTapeFor(Expr* E) {
+    assert(E && "must be provided");
+    QualType TapeType = GetCladTapeOfType(E->getType());
```
vgvassilev: The common notion for names of clang::Type is `TapeTy` this avoid weird naming clashes with existing classes (not the case here but for future reference).

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -748,6 +865,34 @@ namespace clad {
     return StmtDiff(condExpr, condExprDiff);
   }
 
+  // A function used to wrap result of visiting E in a lambda. Returns a call
+  // to the built lambda. Func is a functor that will be invoked inside lambda
+  // scope and block. Statements inside lambda are expected to be added by
+  // addToCurrentBlock from func invocation.
+  template <typename F>
+  static Expr* wrapInLambda(VisitorBase& V, Sema& S, const Expr* E, F&& func) {
+    // FIXME: Here we make use some of the things that are used from Parser, it
```
```suggestion
    // FIXME: Here we make use of some of the things that are used from Parser, it
```

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -1594,16 +1719,45 @@ namespace clad {
...
+      // skip corresponding push.
+      cond = StoreAndRef(cond.getExpr(), forward, "_t", /*force*/ true);
+      StmtDiff condPushPop = GlobalStoreAndRef(cond.getExpr(), "_cond");
+      PushCond = condPushPop.getExpr();
+      PopCond = condPushPop.getExpr_dx();
+    }
+    else
+      cond = GlobalStoreAndRef(cond.getExpr(), "_cond");
+    // Convert cond to boolean condition.
+    for (Stmt*& S : cond.getBothStmts())
```
vgvassilev: `Stmt*& S` is worth documenting a bit more.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -1685,12 +1843,13 @@ namespace clad {
...
     const clang::ConditionalOperator* CO) {
-    Expr* cond = Clone(CO->getCond());
+    StmtDiff cond = Clone(CO->getCond());
     // Condition has to be stored as a "global" variable, to take the correct
     // branch in the reverse pass.
-    cond = GlobalStoreAndRef(cond, "_cond");
-    cond = m_Sema.ActOnCondition(m_CurScope, noLoc, cond,
-                                 Sema::ConditionKind::Boolean).get().second;
+    cond = GlobalStoreAndRef(cond.getExpr(), "_cond");
+    for (Stmt*& S : cond.getBothStmts())
```
vgvassilev: Likewise, it'd be great if we can move it in a utility function.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -1747,8 +1908,148 @@ namespace clad {
...
+    if (condVarDecl) {
+       VarDeclDiff condVarResult = DifferentiateVarDecl(condVarDecl);
+       condVarClone = condVarResult.getDecl();
+       // If there is a variable declared, store its derivative as a global,
+       // as we usually do with declarations in reverse mode.
+       if (condVarResult.getDecl_dx())
+         addToBlock(BuildDeclStmt(condVarResult.getDecl_dx()), m_Globals);
+    }
+
+    // FIXME: for now we assume that cond has no differentiable effects,
```
vgvassilev: Could you add a test case marking it `XFAIL` with this case?

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -1747,8 +1908,148 @@ namespace clad {
     return StmtDiff(condExpr);
   }
 
+  StmtDiff ReverseModeVisitor::VisitForStmt(const ForStmt* FS) {
```
vgvassilev: We should add documentation about this transformation. Either describing this routine or in a text form elsewhere or even better: both.

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -399,6 +415,8 @@ namespace clad {
     Stmts m_Globals;
     //// A reference to the output parameter of the gradient function.
     clang::Expr* m_Result;
+    /// A flag indicating if the Stmt we are currently visiting is inside loop.
+    bool isInsideLoop = false;
```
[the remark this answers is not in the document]
efremale: I am more in favor of skipping it everywhere, we should unify that at some point.

include/clad/Differentiator/Tape.h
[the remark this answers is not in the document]
efremale: There is usually no way to make a reasonable estimate the capacity. However, we may allow user to make that guess globally, e.g. by modifying `_init_capacity` of `clad::tape`.

include/clad/Differentiator/Tape.h
```diff
@@ -0,0 +1,99 @@
...
+
+#include <cassert>
+#include <memory>
+#include <type_traits>
+#include <utility>
+
+namespace clad {
+  template <typename T>
+  class tape_impl {
+    void* _data = nullptr;
```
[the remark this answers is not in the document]
efremale: This is user facing, so it needs not to be of the same style as the rest of the project (user doesn't see the rest). It seems more reasonable to follow `std::` style here.

include/clad/Differentiator/Tape.h
efremale: This is user-facing, user does not necessarily need to include LLVM headers when compiling clad-generated code, so this may fail. There will be ``[[unlikely]]`` attribute for that in C++20.

lib/Differentiator/DerivativeBuilder.cpp
efremale: Result here is `static` to avoid repeating lookup on each request. However, before `GetCladTapePush` is called first time, `Result` does not exist, hence it is optional. Just using uninitialized `LookupResult` doesn't work, since it has no default constructor.

lib/Differentiator/DerivativeBuilder.cpp
efremale: Instantiation actually happens on `CheckTemplateIdType`. Who might have thought :laughing:

lib/Differentiator/DerivativeBuilder.cpp
efremale: Just a reference to the underlying value of the range, which is of type `Stmt*` (as if it were `auto& S` here).

lib/Differentiator/DerivativeBuilder.cpp
efremale: See above. It is a reference, since we are assigning to `S`, it is just the usual range-for loop with reference.

lib/Differentiator/DerivativeBuilder.cpp
efremale: I think there are enough comments inside it to make it fairly clear what is going on. Is there anything particular that you'd like to address?

lib/Differentiator/DerivativeBuilder.cpp
efremale: I'd prefer to open an issue for that.

include/clad/Differentiator/DerivativeBuilder.h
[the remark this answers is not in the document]
vgvassilev: Yeah, before that, let's keep it consistent to the rest ;)

include/clad/Differentiator/Tape.h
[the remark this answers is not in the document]
vgvassilev: The class `CladFunction` is also user-facing and follows, the regular convention. Let's stick to it for now.

include/clad/Differentiator/Tape.h
vgvassilev: Ah, good point... sigh...

lib/Differentiator/DerivativeBuilder.cpp
vgvassilev: Doesn't that allow you to modify "in-place" `S`. Eg. we rewrite the content of `cond.getBothStmts()`.

include/clad/Differentiator/Tape.h
```diff
@@ -0,0 +1,114 @@
+#ifndef CLAD_TAPE_H
+#define CLAD_TAPE_H
+
+#include <cassert>
+#include <memory>
+#include <type_traits>
+#include <utility>
```
vgvassilev: Can we reduce the includes and at what cost? All this will correspond into bigger AST to work with.

lib/Differentiator/DerivativeBuilder.cpp
vgvassilev: I'd like to see documentation, describing the transformation A->B, eg. this for loop becomes 'that'. This will help farmer John (myself included) understanding the use of tape, push, pop, etc.

efremale: Lets leave it for some more elaborate documentation outside of code.

lib/Differentiator/DerivativeBuilder.cpp
efremale: Yes, exactly.

include/clad/Differentiator/DerivativeBuilder.h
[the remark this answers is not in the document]
efremale: There is already a convention to use normal style for bool flags, e.g. `silenceDiags`.

include/clad/Differentiator/Tape.h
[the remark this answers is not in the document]
efremale: What is the convention here? Even `llvm::SmallVector` uses "unconventional" names like push_back, pop_back etc. This also make sense if we will ever change clad::tape alias to std:: type. So I'd rather stay with normal convention, maybe even rewrite CladFunction, since the amount of user facing code there is minimal.

include/clad/Differentiator/Tape.h
efremale: I can remove `<type_traits>` and `<utility>` without errors, but I guess that is because they are already included somewhere else, and that is not entirely correct.

lib/Differentiator/DerivativeBuilder.cpp
vgvassilev: That's a non-standard way of use in the for loop. Could you just document it, adding something like "We change cond.getBothStmts() content in-place."

include/clad/Differentiator/Tape.h
vgvassilev: Yes, you are right, I experimented a bit. If we want to reduce anything we need to remove `memory`. Let's leave it like this.

vgvassilev:
Please remove `lib/Differentiator/Tape.cpp` or register it in cmake so that the bots can start building.

PR #141, opened 2019-07-26
Hessian Implementation

540KJ:
Verified test results with Wolfram Alpha
Reduced tests to only feature relevant ones that test Hessians specifically
Removed 'noise' from previous version of PR.

### DIFF
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -169,5 +169,26 @@ namespace clad {
       reinterpret_cast<void (C::*) (Args..., R*)>(f) /* will be replaced by gradient*/,
       code);
   }
+
+  /// Function for Hessian matrix computation
+  /// Given  a function f, clad::hessian generates all the second derivatives
+  /// of the original function, (they are also columns of a Hessian matrix
+  template<typename ArgSpec = const char *, typename R, typename... Args>
+  CladFunction<false, void, Args..., R*> __attribute__((annotate("H")))
+  hessian(R (*f)(Args...), ArgSpec args = "", const char* code = "") {
+    assert(f && "Must pass in a non-0 argument");
+    return CladFunction<false, void, Args..., R*>(
+      reinterpret_cast<void (*) (Args..., R*)>(f) /* will be replaced by Hessian*/,
+      code);
+  }
+    
+  template<typename ArgSpec = const char *, typename R, typename C, typename... Args>
+  CladFunction<true, void, C, Args..., R*> __attribute__((annotate("H")))
+  hessian(R (C::*f)(Args...), ArgSpec args = "", const char* code = "") {
+    assert(f && "Must pass in a non-0 argument");
+    return CladFunction<true, void, C, Args..., R*>(
+      reinterpret_cast<void (C::*) (Args..., R*)>(f) /* will be replaced by Hessian*/,
+      code);
+  }
 }
 #endif // CLAD_DIFFERENTIATOR
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -59,7 +59,7 @@ namespace clad {
     // If FD is only a declaration, try to find its definition.
     if (!FD->getDefinition()) {
       if (request.VerboseDiags)
-        diag(DiagnosticsEngine::Error, 
+        diag(DiagnosticsEngine::Error,
              request.CallContext ? request.CallContext->getLocStart() : noLoc,
              "attempted differentiation of function '%0', which does not have a "
              "definition", { FD->getNameAsString() });
@@ -74,6 +74,9 @@ namespace clad {
     else if (request.Mode == DiffMode::reverse) {
       ReverseModeVisitor V(*this);
       result = V.Derive(FD, request);
+    } else if (request.Mode == DiffMode::hessian) {
+      HessianModeVisitor H(*this);
+      result = H.Derive(FD, request);
     }
 
     if (result.first)
@@ -317,7 +320,7 @@ namespace clad {
       id += 1;
     for (;;) {
       IdentifierInfo* name = &m_Context.Idents.get(nameBase.str() + idStr);
-      LookupResult R(m_Sema, DeclarationName(name), noLoc, 
+      LookupResult R(m_Sema, DeclarationName(name), noLoc,
                      Sema::LookupOrdinaryName);
       m_Sema.LookupName(R, m_CurScope, /*AllowBuiltinCreation*/ false);
       if (R.empty()) {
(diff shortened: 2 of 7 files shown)


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -398,6 +401,235 @@ namespace clad {
...
+
+  HessianModeVisitor::~HessianModeVisitor() {}
+
+  DeclWithContext HessianModeVisitor::Derive(const clang::FunctionDecl* FD,
+                                             const DiffRequest& request) {
+    int numIndependentArgs = request.Function->getNumParams();
+    std::vector<FunctionDecl*> secondDerivativeColumns;
+
+    // Copied and adapted from clang::Sema::ActOnStringLiteral.
+    QualType CharTyConst = m_Context.CharTy;
```
```suggestion
    QualType CharTyConst = m_Context.CharTy.withConst();
```

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -398,6 +401,235 @@ namespace clad {
...
+    CharTyConst.addConst();
```
```suggestion

```

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -398,6 +401,235 @@ namespace clad {
...
+    beginScope(Scope::FnScope | Scope::DeclScope);
+    m_DerivativeFnScope = getCurrentScope();
+
+    // Creates callExprs to the second derivative functions genereated
+    // and creates maps array elements to input array.
+    int numSecDerivatives = secDerivFuncs.size();
+    for (int i = 0; i < numSecDerivatives; i++)
+    {
+      Expr* exprFunc = BuildDeclRef(secDerivFuncs[i]);
+      int numIndependentArgs = secDerivFuncs[i]->getNumParams();
```
```suggestion
      const unsigned numIndependentArgs = secDerivFuncs[i]->getNumParams();
```

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -398,6 +401,235 @@ namespace clad {
...
+    Expr* m_Result = BuildDeclRef(params.back());
+    std::vector<Stmt*> CompStmtSave;
+
+    beginScope(Scope::FnScope | Scope::DeclScope);
+    m_DerivativeFnScope = getCurrentScope();
+
+    // Creates callExprs to the second derivative functions genereated
+    // and creates maps array elements to input array.
+    int numSecDerivatives = secDerivFuncs.size();
+    for (int i = 0; i < numSecDerivatives; i++)
```
vgvassilev: This usually is done by:
`for (size_t i = 0, e = secDerivFuncs.size(); i < e; ++i)`

This reduces the scope of numSecDerivatives, too.

include/clad/Differentiator/Differentiator.h
```diff
@@ -169,5 +169,26 @@ namespace clad {
...
+
+  /// Function for Hessian matrix computation
+  /// Given  a function f, clad::hessian generates all the second derivatives
+  /// of the original function, (they are also columns of a Hessian matrix
+  template<typename ArgSpec = const char *, typename R, typename... Args>
+  CladFunction<false, void, Args..., R*> __attribute__((annotate("H")))
+  hessian(R (*f)(Args...), ArgSpec args = "", const char* code = "") {
+    assert(f && "Must pass in a non-0 argument");
+    return CladFunction<false, void, Args..., R*>(
+      reinterpret_cast<void (*) (Args..., R*)>(f),
```
vgvassilev:
```suggestion
      reinterpret_cast<void (*) (Args..., R*)>(f) /* will be replaced by gradient*/,
```

include/clad/Differentiator/Differentiator.h
```diff
@@ -169,5 +169,26 @@ namespace clad {
...
+      code);
+  }
+    
+  template<typename ArgSpec = const char *, typename R, typename C, typename... Args>
+  CladFunction<true, void, C, Args..., R*> __attribute__((annotate("H")))
+  hessian(R (C::*f)(Args...), ArgSpec args = "", const char* code = "") {
+    assert(f && "Must pass in a non-0 argument");
+    return CladFunction<true, void, C, Args..., R*>(
+      reinterpret_cast<void (C::*) (Args..., R*)>(f),
```
vgvassilev:
```suggestion
      reinterpret_cast<void (C::*) (Args..., R*)>(f) /* will be replaced by gradient*/,
```

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -398,6 +401,235 @@ namespace clad {
...
   }
+  
+  HessianModeVisitor::HessianModeVisitor(DerivativeBuilder& builder):
+  VisitorBase(builder) {}
+
+  HessianModeVisitor::~HessianModeVisitor() {}
+
+  DeclWithContext HessianModeVisitor::Derive(const clang::FunctionDecl* FD,
+                                             const DiffRequest& request) {
+    int numIndependentArgs = request.Function->getNumParams();
```
vgvassilev: See other suggestion...

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -398,6 +401,235 @@ namespace clad {
...
+    // Copied and adapted from clang::Sema::ActOnStringLiteral.
+    QualType CharTyConst = m_Context.CharTy;
+    CharTyConst.addConst();
+    // Get an array type for the string, according to C99 6.4.5. This includes
+    // the nul terminator character as well as the string length for pascal
+    // strings.
+
+    // Generates the second derivative functions corresponding to a column of
+    // a hessian matrix.
+    for (int i = 1; i <= numIndependentArgs; i++)
```
vgvassilev: Why don't we start from 0.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -398,6 +401,235 @@ namespace clad {
...
+
+    // Copied and adapted from clang::Sema::ActOnStringLiteral.
+    QualType CharTyConst = m_Context.CharTy;
+    CharTyConst.addConst();
+    // Get an array type for the string, according to C99 6.4.5. This includes
+    // the nul terminator character as well as the string length for pascal
+    // strings.
+
+    // Generates the second derivative functions corresponding to a column of
+    // a hessian matrix.
```
vgvassilev: Could we enhance the documentation explaining on a higher level what does the loop do.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -163,7 +163,7 @@ namespace clad {
     // clad::differentiate(...) __attribute__((annotate("D")))
     // TODO: why not check for its name? clad::differentiate/gradient?
     const AnnotateAttr* A = FD->getAttr<AnnotateAttr>();
-    if (A && (A->getAnnotation().equals("D") || A->getAnnotation().equals("G"))) {
+    if (A && (A->getAnnotation().equals("D") || A->getAnnotation().equals("G") || A->getAnnotation().equals("H"))) {
```
vgvassilev: We try to enforce 80 column width. Could you set up your editor?

test/Hessian/Hessians.C
```diff
@@ -0,0 +1,285 @@
...
+
+#include "clad/Differentiator/Differentiator.h"
+
+double f_cubed_add1(double a, double b) {
+  return a * a * a + b * b * b;
+}
+
+void f_cubed_add1_darg0_grad(double a, double b, double *_result);
+//CHECK:void f_cubed_add1_darg0_grad(double a, double b, double *_result) {
+//CHECK-NEXT:    double _d__d_a = 0;
```
vgvassilev: The gradient verification is not needed in that test. You can remove all the // CHECK directives but the hessian specific ones.

include/clad/Differentiator/Differentiator.h
efremale: /* will be replaced by gradient*/ -> /* will be replaced by Hessian */

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -398,6 +401,235 @@ namespace clad {
...
+
+    llvm::ArrayRef<ParmVarDecl*> paramsRef = llvm::makeArrayRef(params.data(),
+                                                                params.size());
+    hessianFD->setParams(paramsRef);
+    Expr* m_Result = BuildDeclRef(params.back());
+    std::vector<Stmt*> CompStmtSave;
+
+    beginScope(Scope::FnScope | Scope::DeclScope);
+    m_DerivativeFnScope = getCurrentScope();
+
```
efremale: It would be great if we could extract some part of the code above into the separate function, since now it is used by  Forward, Reverse, and Hessian visitors in almost the same way.

test/Hessian/Hessians.C
540KJ: Do we want one just in case -  test if ForwardModeVisitor and ReverseModeVisitor have been used correctly

vgvassilev: It would have been caught elsewhere from the test suite, but if you want add for one case. This reminds me that we also need to add a demo of the new functionality.

lib/Differentiator/DiffPlanner.cpp
540KJ: Does it have to exactly under 80 column? In general, some lines go over 80 chars slightly

vgvassilev: We must avoid going over 80 cols but we are all humans and go over it. Especially the new contributions should follow the 80 cols.

vgvassilev:
@540KJ ping...

lib/Differentiator/DerivativeBuilder.cpp
540KJ: Isn't the code between beginScope and endScope different for each visitor? Do you just want a function that encapsulates beginScope and getCurrentScope?

efremale: Isn't the whole setup above pretty much the same for differentiate/gradient/hessian? There are some differences in function type signature (`hessianFunctionType` here) and function's name (_darg, _grad, _hessian). Can't we factor the code above out into some separate function, which will be parameterized, e.g. will take function type and name as args?

540KJ: Should I put this in a separate PR?

efremale: Sure, lets put it into a separate PR.

PR #142, opened 2019-10-14
vscode support

alexander-penev:


vgvassilev:
That's already in the master after PR #144

PR #152, opened 2019-11-26
Enable code coverage (codecov.io) for clad project.

alexander-penev:


vgvassilev:
That should fix #114.

PR #157, opened 2019-12-01
[travis] Run clad with clang which has asserts enabled.

vgvassilev:
This gives us a great deal of coverage. Clang has well-developed assert system
which often can detect malformed AST. This is very essential for clad as until
today I would need to run that by hand, locally on every pull request.


vgvassilev:
Resolved by https://github.com/vgvassilev/clad/pull/662

PR #161, opened 2020-06-24
Jacobian v2

540KJ:
For cross comparison purposes.


vgvassilev:
@540KJ, thanks a lot! @blide has incorporated good amount of this PR in PR #162 

I will close that one now.

PR #164, opened 2020-07-30
Demo: ODE solver sensitivity

blide:


vgvassilev:
@blide, can you run that demo adding a line in `test/Misc/RunDemos.C`

demos/ODESolverSensitivity.cpp
```diff
@@ -0,0 +1,92 @@
...
+}
+
+double solution(double a, double b, double c, double x) {
+  return rungeKutta(0, 0, x, 0.001, a, b, c);
+}
+
+// Evaluate b-sensitivity over x domain
+double bSensitivity(double x) {
+  // Using reverse mode since forward mode doesn't support
+  // differentiation of multi-arg calls
```
vgvassilev, line 66: Can you open an issue and reference it in a FIXME?

blide:
Should I add checks for a generated code? I guess it's too long to make sense testing it.

vgvassilev:
No, we want to make sure it runs successfully, no need to check generated code. How long does it run and can make it faster by reducing iterations?

blide:
It runs in milliseconds.

vgvassilev:
Then we are good. Let's add it there.

demos/ODESolverSensitivity.cpp
```diff
@@ -0,0 +1,93 @@
...
+  // FIXME: switch to clad::differentiate after fixing issue#168
```
vgvassilev, line 67:
```suggestion
  // FIXME: switch to clad::differentiate after fixing vgvassilev/clad#168
```

PR #170, opened 2020-09-10
Find LLVM before Clang. Fix ROOT-10886

pcanal:
Find LLVM before Clang.
    
    This fixes ROOT-10886.
    
    In the environment described in ROOT-10886, CMAKE_PREFIX_PATH is set to point
    to the clang (compiler) installation with includes an installation of the
    LLVM and Clang libraries.
    
    When finding Clang, the find_package for Clang will (in the ROOT environment)
    execute interpreter/llvm/src/tools/clang/lib/cmake/clang/ClangConfig.cmake
    which contains the line:
    
      find_package(LLVM REQUIRED CONFIG
                   HINTS path_where_ROOT_found_LLVM)
    
    However, CMAKE_PREFIX_PATH has search priority of the HINTS and circunvents
    Clad's CMakeLists.txt to constrain the search first to a version then
    to paths when they are supplied.
    
    This result in Cling and Clad using 2 distincts (and incompatible) version
    of LLVM/Clang.
    
    The solution is to search for LLVM first and then for Clang.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,46 +26,6 @@ if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
     set(LLVM_DIR ${Clang_DIR})
   endif()
 
-  ## Find supported Clang
-
-  set(CMAKE_FIND_PACKAGE_SORT_ORDER NATURAL)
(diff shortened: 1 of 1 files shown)


pcanal:
@vgvassilev Could you review this?

vgvassilev (approved):
Thanks Philippe! LGTM!

PR #174, opened 2020-09-29
Fix section order in README.md

efremale:
It seems the "How Clad works" section was incorrectly placed in the middle of "How to install" section. Fixed


vgvassilev:
Great to hear from you @efremale! Thanks!

PR #175, opened 2020-09-30
Improved scope resolution vector

jatin22-exe:


vgvassilev:
@jatin22-exe thanks for your contribution. Can you elaborate a little more why this change is needed in the commit message?

jatin22-exe:
Yeah sure. The main purpose is that it increases the readibility of the
code. Which helps to understand more.

vgvassilev:
Adding spaces between qualification only for std::cout would be inconsistent to the other uses of ::. If you change them, too, that'd make code less standard looking to most of C++ devs.

jatin22-exe:
I agree with you but most cases is of that only. I want to tell you that I
have 4 star at hacker and verified by google digital garage.
I will be thankful to you if you pass my PR.
yours truly.

vgvassilev:
Unfortunately, we cannot accept this PR. Feel free to close this one. If you find bugs in the code I'd be happy to review future PRs!

klmr:
@vgvassilev The user seems to create spam PRs (see also klmr/named-operator#5). Close and block.

vgvassilev:
@klmr, thanks for letting me know!

PR #183, opened 2020-11-09
 Differentiate boolean conditions

### LINKED ISSUE #132: Differentiate boolean conditions
Currently, we are not differentiating boolean conditions (e.g. `C` in `if (C) ...` or `(C ? ... : ...)`, we are just cloning them. 
However, they may have differentiable side-effects, e.g. `if (x = y) ...`. In that example, if result of the function depends on `x`, our derivative will be incorrect. Therefore, we should also call `Visit` on conditions.

grimmmyshini:
Differentiate boolean conditions to account for differentiable side effects. 

Fixes  #132

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -1557,9 +1557,7 @@ namespace clad {
       return StmtDiff(op, diff.getExpr_dx());
     }
     else {
-      diag(DiagnosticsEngine::Warning, UnOp->getEndLoc(),
-           "attempt to differentiate unsupported unary operator, derivative \
-            set to 0");
+      unsupportedOpWarn(UnOp->getEndLoc());
       auto zero = ConstantFolder::synthesizeLiteral(m_Context.IntTy,
                                                     m_Context, 0);
       return StmtDiff(op, zero);
@@ -1648,9 +1646,7 @@ namespace clad {
     }
     if (!opDiff) {
       //FIXME: add support for other binary operators
-      diag(DiagnosticsEngine::Warning, BinOp->getEndLoc(),
-           "attempt to differentiate unsupported binary operator, derivative \
-            set to 0");
+      unsupportedOpWarn(BinOp->getEndLoc());
       opDiff = ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
     }
     opDiff = folder.fold(opDiff);
@@ -2024,6 +2020,7 @@ namespace clad {
     // the if statement.
     Expr* PushCond = nullptr;
     Expr* PopCond = nullptr;
+    auto condExpr = Visit(cond.getExpr());
     if (isInsideLoop) {
       // If we are inside for loop, cond will be stored in the following way:
       // forward:
@@ -2036,13 +2033,13 @@ namespace clad {
       // if (clad::push(..., _t) { ... }
       // is incorrect when if contains return statement inside: return will
       // skip corresponding push.
-      cond = StoreAndRef(cond.getExpr(), forward, "_t", /*force*/ true);
+      cond = StoreAndRef(condExpr.getExpr(), forward, "_t", /*force*/ true);
       StmtDiff condPushPop = GlobalStoreAndRef(cond.getExpr(), "_cond");
       PushCond = condPushPop.getExpr();
       PopCond = condPushPop.getExpr_dx();
     }
     else
-      cond = GlobalStoreAndRef(cond.getExpr(), "_cond");
+      cond = GlobalStoreAndRef(condExpr.getExpr(), "_cond");
     // Convert cond to boolean condition. We are modifying each Stmt in StmtDiff.
     for (Stmt*& S : cond.getBothStmts())
       if (S)
@@ -2137,7 +2134,7 @@ namespace clad {
     StmtDiff cond = Clone(CO->getCond());
     // Condition has to be stored as a "global" variable, to take the correct
     // branch in the reverse pass.
-    cond = GlobalStoreAndRef(cond.getExpr(), "_cond");
+    cond = GlobalStoreAndRef(Visit(cond.getExpr()).getExpr(), "_cond");
     // Convert cond to boolean condition. We are modifying each Stmt in StmtDiff.
     for (Stmt*& S : cond.getBothStmts())
       S = m_Sema.ActOnCondition(m_CurScope, noLoc, cast<Expr>(S),
@@ -2241,8 +2238,9 @@ namespace clad {
     // but it is not generally true, e.g. for (...; (x = y); ...)...
     StmtDiff cond;
     if (FS->getCond())
-      cond = Clone(FS->getCond());
-    const Expr* inc = FS->getInc();
+      cond = Visit(FS->getCond());
+    auto IDRE = dyn_cast<DeclRefExpr>(FS->getInc());
+    const Expr* inc = IDRE? Visit(FS->getInc()).getExpr(): FS->getInc();
 
     // Save the isInsideLoop value (we may be inside another loop).
     llvm::SaveAndRestore<bool> SaveIsInsideLoop(isInsideLoop);
@@ -2705,8 +2703,10 @@ namespace clad {
       diff = Visit(UnOp->getSubExpr(), dfdx());
     }
     else {
-      diag(DiagnosticsEngine::Warning, UnOp->getEndLoc(),
-           "attempt to differentiate unsupported unary operator, ignored");
+      // We should not output any warning on visiting boolean conditions
+      // FIXME: We should support boolean differentiation or ignore it completely
+      if(opCode != UO_LNot)
+        unsupportedOpWarn(UnOp->getEndLoc());
 
       Expr* subExpr = UnOp->getSubExpr();
       if(auto SDRE = dyn_cast<DeclRefExpr>(subExpr))
@@ -2991,8 +2991,10 @@ namespace clad {
       ResultRef = Ldiff.getExpr();
     }
     else {
-      diag(DiagnosticsEngine::Warning, BinOp->getEndLoc(),
-           "attempt to differentiate unsupported binary operator, ignored");
+      // We should not output any warning on visiting boolean conditions
+      // FIXME: We should support boolean differentiation or ignore it completely
+      if(!BinOp->isComparisonOp() && !BinOp->isLogicalOp())
+        unsupportedOpWarn(BinOp->getEndLoc());
 
       // If either LHS or RHS is a declaration reference, visit it to avoid naming collision
       auto LDRE = dyn_cast<DeclRefExpr>(L);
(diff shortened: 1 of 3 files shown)


vgvassilev:
Thanks a lot for this PR!

We need to decide how to handle the warnings as they are valid. Maybe we should ignore conditionals?

grimmmyshini:
Technically we should ignore boolean expressions, however, in a few cases, it might be necessary to visit them. The cases are as follows: 

- Naming collisions

If any variable in the loop/conditional operator/If statement was renamed by clad, we have to visit that variable to make sure the declaration for the new name is built and replaced appropriately. For example, consider the following piece of code

```cpp 
int _r0 = 53;
if( _r0 < 30)
  return _r0;
```  
A part of the clad generated code will then look like this (without calling visit on the conditional expression)

```cpp
int _d__r00 = 0;
_Bool _cond0;
int _r00 = 53;
 _cond0 = _r0 < 30;
```
As you can see, the variable in the condition for the `if` statement does not get replaced by the correct reference, hence making clad crash with a `DeclRefExpr for Decl not entered in LocalDeclMap?` message. Similar is the case with the conditional and looping constructs.

- Differentiable statements in boolean expressions 

As mentioned in #132, a case like

```cpp
if( x = y)
//do something
```
or 
```cpp
z = (x = (x+y))? ... : ...;
```

may have differentiable side effects if the function depends on variable x and we chose to not visit the conditional statements.

Perhaps we could, in the `ReverseModeVisitor::VisitBinaryOperator` and its unary counterpart, add another branch for conditional operators and then do as we were previously, just clone the statement there. In any case, they _should not be_ differentiated in the first place (however boolean differential calculus does exist but I am not sure if I know much about it to comment on this situation) but that does not mean that clad _cannot_ differentiate them. So, I feel like even if we move them into a separate branch for now, it probably would not be a big issue.

vgvassilev:
Thanks for the detailed explanation. Let's make a separate branch in the Bin/UnOp and clone those. There we should add a fixme explaining that we should probably revisit this with the boolean differental calculus in mind.

grimmmyshini:
Yep, that sounds good. I'll get to that soon!

grimmmyshini:
@vgvassilev I just avoided the warning for now. Let me know if you want me to make a separate branch instead.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -2705,8 +2707,11 @@ namespace clad {
       diff = Visit(UnOp->getSubExpr(), dfdx());
     }
     else {
-      diag(DiagnosticsEngine::Warning, UnOp->getEndLoc(),
-           "attempt to differentiate unsupported unary operator, ignored");
+      // We should not output any warning on visiting boolean conditions
+      // FIXME: We should support boolean differentiation or ignore it completely
+      if(opCode != UO_LNot)
+        diag(DiagnosticsEngine::Warning, UnOp->getEndLoc(),
```
vgvassilev, line 2713: Can you move those diagnostics in a separate routine?

vgvassilev (commented):
Thanks a lot for the changes. Can you also add tests testing the diag codepaths?

lib/Differentiator/DerivativeBuilder.cpp
grimmmyshini: Would that help? The diagnostic message is just one line so would it not be wise to keep it as it is to avoid the function call overhead?

grimmmyshini:
> Thanks a lot for the changes. Can you also add tests testing the diag codepaths?

Sure! should I put those in a separate PR?

vgvassilev:
Please add them to the same commit of this PR. This way the tests can also describe in a technical way the introduced changes.

grimmmyshini:
Sure! will do that.

lib/Differentiator/DerivativeBuilder.cpp
vgvassilev: If we pick a good function name, I think it will make the code clearer. If it is a file static the function will likely be inlined.

grimmmyshini:
Okay, so regarding the unary and binary operators, there are some operators we should definitely ignore as it makes no mathematical sense to derive them, like address and indirection operators (`*` and `&`), or pointer-to-member operations (`->` and `->*`) so we should technically not warn for them either.

 Not sure where bitwise operators stand here, would it make sense to differentiate them?

vgvassilev:
> Okay, so regarding the unary and binary operators, there are some operators we should definitely ignore as it makes no mathematical sense to derive them, like address and indirection operators (`*` and `&`), or pointer-to-member operations (`->` and `->*`) so we should technically not warn for them either.

Agreed. Let's make a conservative set of those and not warn against them.

> 
> Not sure where bitwise operators stand here, would it make sense to differentiate them?

I think we need to look into specific implementations of math functions (unless we have a clear guidance from existing AD literature). I presume the bitwise operators are mostly used for efficient implementation of parts of those functions.

efremale:
> > Okay, so regarding the unary and binary operators, there are some operators we should definitely ignore as it makes no mathematical sense to derive them, like address and indirection operators (`*` and `&`), or pointer-to-member operations (`->` and `->*`) so we should technically not warn for them either.
> 
> Agreed. Let's make a conservative set of those and not warn against them.
> 
> > Not sure where bitwise operators stand here, would it make sense to differentiate them?
> 
> I think we need to look into specific implementations of math functions (unless we have a clear guidance from existing AD literature). I presume the bitwise operators are mostly used for efficient implementation of parts of those functions.

Wouldn’t it make sense to be able to differentiate & and * though, when the pointer points to something differentiable?

For example
```
double f(double x) {
  double * p = &x;
  return *p;
}
```

grimmmyshini:
> Wouldn’t it make sense to be able to differentiate & and * though, when the pointer points to something differentiable?

This is a fair point. I feel like the address operator `&` should be left untouched (since the derivative of `p` does not yield any meaningful result) but the dereferencing operator `*` should be differentiated. Looking at it in a purely mathematical sense, `*p` is nothing but the variable `x` itself but perhaps another _alias_  for it. So technically, whenever we encounter `*p` we should act as if we have encountered `x` and that would make `_d_*p` store the address of the derivative of `x` (`_d_x`). Does this make sense?

vgvassilev:
> > > Okay, so regarding the unary and binary operators, there are some operators we should definitely ignore as it makes no mathematical sense to derive them, like address and indirection operators (`*` and `&`), or pointer-to-member operations (`->` and `->*`) so we should technically not warn for them either.
> > 
> > 
> > Agreed. Let's make a conservative set of those and not warn against them.
> > > Not sure where bitwise operators stand here, would it make sense to differentiate them?
> > 
> > 
> > I think we need to look into specific implementations of math functions (unless we have a clear guidance from existing AD literature). I presume the bitwise operators are mostly used for efficient implementation of parts of those functions.
> 
> Wouldn’t it make sense to be able to differentiate & and * though, when the pointer points to something differentiable?
> 
> For example
> 
> ```
> double f(double x) {
>   double * p = &x;
>   return *p;
> }
> ```

We should be careful when tracking this information. In that case one can consider `p` an alias of `x`. However, we cannot assume that in the general case.

vgvassilev:
> > Wouldn’t it make sense to be able to differentiate & and * though, when the pointer points to something differentiable?
> 
> This is a fair point. I feel like the address operator `&` should be left untouched (since the derivative of `p` does not yield any meaningful result) but the dereferencing operator `*` should be differentiated. Looking at it in a purely mathematical sense, `*p` is nothing but the variable `x` itself but perhaps another _alias_ for it. So technically, whenever we encounter `*p` we should act as if we have encountered `x` and that would make `_d_*p` store the address of the derivative of `x` (`_d_x`). Does this make sense?

Yes, it makes sense to me. I think we need to see several different use-cases and allow only a conservative set of operations to be differentiated avoiding generating invalid programs.

The `&` operator can be seen as an alias introducer:
```
void f(double * val) {
  *val = 1;
}
void g(double x) {
  double *addr = &x;
  f(addr);
}
```

However, I am not sure if there is a good amount of use-cases using a variation of this (or similar) idiom.

grimmmyshini:
> The `&` operator can be seen as an alias introducer:

That makes a lot of sense, the only time an alias will change is if the address is changed either via assignment through the address-of operator ( `p = &var`) or direct assignment of some other pointer value/ address containing value  ( `p = ptr` ). 

 Another interesting thing to probably note is that while we should differentiate ` ++*p`, we might not want to differentiate `++p`. And these post/pre-increment operators and other pointer arithmetic will also change the variable to which `p` is an alias (if one was to stick with the alias idiom).  

> However, I am not sure if there is a good amount of use-cases using a variation of this (or similar) idiom.

So, I will just add the tests here in this PR and maybe this discussion could be moved to a separate issue? I also had a few ideas regarding the differentiation of other operators like `%` and  `<<` and it might be interesting to discuss those somewhere.

vgvassilev:
> So, I will just add the tests here in this PR and maybe this discussion could be moved to a separate issue?

Yes, the test should be in the same commit as the code that introduces the new behavior.

> I also had a few ideas regarding the differentiation of other operators like `%` and `<<` and it might be interesting to discuss those somewhere.

Please feel free to open a "new issue" and we can continue that particular discussion there.

vgvassilev:
Hi @grimmmyshini, ping on this. Was there anything else you planned to do with this PR besides appeasing the codecov bot?

grimmmyshini:
Not really, I think we were discussing supporting other operators, however, that isn't very relevant to this PR. 

Apart from that, I am not sure how to add a meaningful test since the uncovered parts of the code are areas where we would want to plug in rules for differentiating other differentiable operators. I could make a sneaky substitution in the existing tests (such as  `x = x << 1` instead of `x = x * 2` ) however, then I would have to let the tests pass even on the emission of warnings.

Any ideas on how I could add a meaningful test?

vgvassilev:
What about adding a function say:
```cpp
double f(double x) {
   return x << 1; // expected-warning{{attempted to differentiate unsupported ...}}
}
```

That's the approach we've taken in the past for such checks.

test/FirstDerivative/UnsupportedOpsWarn.C
```diff
@@ -0,0 +1,58 @@
...
+// CHECK-NEXT:   int *pnt = &x;
+// CHECK-NEXT:   int unOpWarn_1_return = x;
+// CHECK-NEXT:   goto _label0;
+// CHECK-NEXT:   _label0:
+// CHECK-NEXT:   _result[0UL] += 1;
+// CHECK-NEXT: }
+
+int main(){
+
+    clad::differentiate(binOpWarn_0, 0).dump();
```
vgvassilev, line 53: I suspect we don’t need the `dump` command. The test suite has a flag to print the derivatives bodies.

vgvassilev (commented):
Thanks for working this! Can you squash all commits into one?

test/FirstDerivative/UnsupportedOpsWarn.C
grimmmyshini: Sorry I completely forgot to remove those! Fixed it.

vgvassilev (approved):
LGTM!

PR #186, opened 2020-12-13
Fix cmake search of LLVM and Clang.

alexander-penev:
In some Linux distrubutions with multiple LLVM/Clang versions
instaled Clad detect wrong (and uncompatioble) versions of
LLVM and Clang.


vgvassilev:
Looks like we have OSX travis outage.

PR #190, opened 2021-01-17
Visit only regions annotated with #pragma clad ON.

vgvassilev:
This PR implements more sophisticated handling of #pragma clad ON/OFF/DEFAULT
and tracks the regions where it was enabled/disabled.

Later, we only collect the calls to `clad::*` within that region without
descending into the clad-produced derivatives/gradients.


vgvassilev:
The codecov cannot detect the coverage -- it comes from the *-runtime5 builds, so we should be good to go.

PR #165, opened 2020-07-30
Demo: gradient descent

blide:

PR #191, opened 2021-01-20
Add gradient descent demo

grimmmyshini:
Add a demo that illustrates using `clad::gradient` to perform gradient descent to solve a simple linear regression problem. Inspired by #165

Running this demo and subsequently the gnuplot, one should observe a similar output:
graph

**Things to be kept in mind:** Changing the dataset size may cause the algorithm to fail to converge. However, this is expected behavior and can be corrected by preprocessing the dataset.


demos/GradientDescent.cpp
```diff
@@ -0,0 +1,139 @@
...
+// theta_0, theta_1 are the features to learn and x is the input data
+double f(double theta_0, double theta_1, double x) {
+    return theta_0 + theta_1 * x;
+}
+
+// Structure to represent input dataset
+// x, y represents the input and output data respectively
+// size represents the number of samples
+// learning_rate represents the learning rate
+struct Dataset{
```
vgvassilev, line 41:
```suggestion
struct Dataset {
```

vgvassilev: Can you add spaces in these cases.

demos/GradientDescent.cpp
```diff
@@ -0,0 +1,139 @@
...
+    size_t size = 50;
+    double learning_rate = 1e-5;
+
+    // Populate the data set with random data
+    Dataset(){
+
+        // For plotting data later
+        std::ofstream out("dataset_gd.dat");
+
+        for(size_t i = 0; i < size; i++){
```
vgvassilev, line 54: Likewise.

demos/GradientDescent.cpp
```diff
@@ -0,0 +1,139 @@
...
+// learning_rate represents the learning rate
+struct Dataset{
+
+    std::vector<double> x;
+    std::vector<double> y;
+    size_t size = 50;
+    double learning_rate = 1e-5;
+
+    // Populate the data set with random data
+    Dataset(){
```
vgvassilev, line 49: Likewise.

demos/GradientDescent.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+// Function to optimize the cost function of interest
+// theta contains the feature parameters
+// maxSteps are the maximum steps to perform
+std::vector<double> optimize(std::vector<double> theta, Dataset dt, unsigned int maxSteps, double eps) {
+
+    std::vector<double> diff = theta;
+    bool hasConverged = false;
+    int currentStep = 0;
+
```
vgvassilev, line 95: We can do here `clad_grad = clad::gradient(cost);` and remove the forward declaration.

Then when we need to compute the gradient we can call `clad_grad.execute`.

Actually I'd be quite interested how that would compare against the current approach in terms of performance. I suspect some overhead.

demos/GradientDescent.cpp
```diff
@@ -0,0 +1,139 @@
...
+double cost(double theta_0, double theta_1, double x, double y){
+
+    double f_x = f(theta_0, theta_1, x);
+    return (f_x - y) * (f_x - y);
+
+}
+
+int main() {
+
+  clad::gradient(cost);
```
vgvassilev, line 122: You may want to move that in the `optimize` function.

vgvassilev (commented):
Overall looks good -- some tiny nits inline.

I wonder why codecov did not report coverage improvement for this demo -- do we really run it?

demos/GradientDescent.cpp
grimmmyshini: Yep, using `clad_grad.execute` increases the cost of the total function ( i.e `clad_grad.execute` + `cost_grad`) to 65.55 %

grimmmyshini:
> I wonder why codecov did not report coverage improvement for this demo -- do we really run it?
 
I have added the line to run this demo under misc tests. Weirdly enough, codecov actually reports a 0.01% decrease in coverage immediately in the file `include/clad/Differentiator/Compatibility.h`, not sure what that means at all.

grimmmyshini:
Also, @vgvassilev do you think I should change the name of the demo file?

demos/GradientDescent.cpp
vgvassilev: Is the example compiled with -O2 or -O3?

vgvassilev:
Do you have a specific name in mind?

demos/GradientDescent.cpp
grimmmyshini: It was compiled with -O0 when I profiled it

grimmmyshini:
Not really, but what about 'LinearRegression' ?

demos/GradientDescent.cpp
vgvassilev: Ok, could you also check when compiling with O2 — I expect improvement due to inlining.

demos/GradientDescent.cpp
```diff
@@ -0,0 +1,139 @@
...
+// -I../include/ -x c++ -lstdc++ -lm GradientDescent.cpp
+//
+// To plot the results install gnuplot and type:
+// gnuplot -e "plot 'dataset_gd.dat' with points pt 7; replot 'out_gd.dat' using 1:2 with lines; pause -1"
+
+#include <iostream>
+#include <fstream>
+#include <cstdlib>
+#include <vector>
+#include <algorithm>
```
vgvassilev, line 26: Can you order the includes alphabetically?

demos/GradientDescent.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+    // Populate the data set with random data
+    Dataset() {
+
+        // For plotting data later
+        std::ofstream out("dataset_gd.dat");
+
+        for(size_t i = 0; i < size; i++) {
+
+            x.push_back( (std::rand()%20) );
```
vgvassilev, line 56:
```suggestion
            x.push_back(std::rand()%20);
```

demos/GradientDescent.cpp
```diff
@@ -0,0 +1,139 @@
...
+            y.push_back( f(2 + std::rand()%25, 5, x[i]) );
```
vgvassilev, line 57:
```suggestion
            y.push_back(f(2 + std::rand()%25, 5, x[i]));
```

Although we may need extra variables to describe the meaning of these values.

demos/GradientDescent.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+// Function to perform a minimization step
+// theta_x are the feature parameters
+// dt is the generated dataset
+// clad_grad is the gradient function generated by Clad
+template <typename T>
+void performStep(double& theta_0, double& theta_1, Dataset dt, T clad_grad) {
+    double J_theta[4], result[2] = {0, 0};
+
+    for (size_t i = 0; i < dt.size; i++){
```
vgvassilev, line 74:
```suggestion
    for (size_t i = 0; i < dt.size; i++) {
```

demos/GradientDescent.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+    theta_0 -=  dt.learning_rate * result[0] / (2 * dt.size);
+    theta_1 -=  dt.learning_rate * result[1] / (2 * dt.size);
+}
+
+// The cost function to minimize using gradient descent
+// theta_0, theta_1 are the features to learn
+// x, y are the inputs and outputs of f respectively
+double cost(double theta_0, double theta_1, double x, double y) {
+
```
vgvassilev, line 90: No new lines when we begin a new construct, eg. for-loop/function body/function body end, etc.

demos/GradientDescent.cpp
```diff
@@ -0,0 +1,139 @@
...
+    do {
+        performStep(theta[0], theta[1], dt, clad_grad);
+
+        std::cout << "Steps #" << currentStep << " Theta 0: " << theta[0] << " Theta 1: " << theta[1] << std::endl;
+
+        hasConverged = abs(diff[0] - theta[0]) <= eps && abs(diff[1] - theta[1]) <= eps;
+
+        diff = theta;
+
+    }while(currentStep++ < maxSteps && !hasConverged);
```
vgvassilev, line 117:
```suggestion
    } while(currentStep++ < maxSteps && !hasConverged);
```

demos/GradientDescent.cpp
```diff
@@ -0,0 +1,139 @@
...
+    return theta;
+}
+
+int main() {
+
+  std::vector<double> theta = {0, 0};
+  Dataset dt;
+  auto results = optimize(theta, dt, 10000, 1e-6);
+
+  std::cout << "Result: " << "(" << results[0] << ", " << results[1] << ")" << std::endl;
```
vgvassilev, line 128: We also break at 80 columns. Could you configure your editor?

demos/GradientDescent.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+  // For plotting data
+  std::ofstream out("out_gd.dat");
+  for(size_t i = 0; i < dt.size; i++){
```
vgvassilev, line 132:
```suggestion
  for (size_t i = 0; i < dt.size; i++){
```

demos/GradientDescent.cpp
```diff
@@ -0,0 +1,139 @@
...
+// Necessary for clad to work include
+#include "clad/Differentiator/Differentiator.h"
+
+// The hypothesis function for the quadratic problem
+// theta_0, theta_1 are the features to learn and x is the input data
+double f(double theta_0, double theta_1, double x) {
+    return theta_0 + theta_1 * x;
+}
+
+// Structure to represent input dataset
```
vgvassilev, line 37: Could you align this to 80 columns?

demos/GradientDescent.cpp
grimmmyshini: -O2 makes the algorithm numerically unstable so I suppose there isn't much we can look at performance-wise.

vgvassilev: Ah, that's quite bad -- I think -O2 should have only "safe" optimizations. Can we improve the stability of the algorithm?

grimmmyshini: I'm not entirely sure but I'll try to inspect the code a bit, maybe I'm going wrong somewhere.

demos/GradientDescent.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+// Function to perform a minimization step
+// theta_x are the hypothesis parameters, t is the generated dataset and
+// clad_grad is the gradient function generated by Clad
+template <typename T>
+void performStep(double &theta_0, double &theta_1, Dataset dt, T clad_grad) {
+  double J_theta[4], result[2] = {0, 0};
+
+  for (size_t i = 0; i < dt.size; i++) {
+    clad_grad.execute(theta_0, theta_1, dt.x[i], dt.y[i], J_theta);
```
sudo-panda, line 78: Possible fix for the `-O2` error is initializing `J_theta`elements to `0`.
```suggestion
    J_theta[0] = J_theta[1] = J_theta[2] = J_theta[3] = 0;
    clad_grad.execute(theta_0, theta_1, dt.x[i], dt.y[i], &J_theta[0], &J_theta[1], &J_theta[2], &J_theta[3]);
```

grimmmyshini: @sudo-panda Thanks! That + tuning the hyperparameters solved it! Super silly issue :P

PR #201, opened 2021-02-27
Add info about non-real differentiability

### LINKED ISSUE #198: Clad is not considering truncation of variables inside function for differentiation
```
double f(double x) {
  int a = x;
  return 4*x*x*x + 5*a*a;   //basically 4*x^3 + 5*floor(x)*floor(x), where floor = floor function
}
```
For this function, Intuitively variable `a` should be treated as `floor function of x`. And thus differentiation should be done accordingly. For floor functions differentiation should be undefined at integral points and 0 for non-integral points.  
But in the program result is same whether type of `variable a` is `double` or `int`, and `variable a` is not being considered as floor function of x

For calcuating derivative of this function I have used this code, 
```
int main() {
  double x_val = 2;
  auto f_dx_1 = clad::differentiate(f, "x");
  std::cout << "f_dx_1.execute(x_val) = " << f_dx_1.execute(x_val) << "\n";
// f_dx_1.execute(x_val) = 68 // output 
}
```

grimmmyshini:
Added to README information regarding how Clad does not type discriminate during differentiation.

Closes #198 for now, Should discuss dropping differentiability support for int types somewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,8 @@ The following subset of C++ syntax is supported at the moment:
 * Control flow: `if` statements and `for` loops (*work on loops in the reverse-mode is in progress*)
 * Calls to other functions, including recursion
 
+Note: Clad currently differentiates types such as `int`/`char`/`boolean` as any real type (such as `float`, `double`, etc.) would be differentiated. Users should keep in mind that Clad *does not* warn against lossy casts, which on differentiation may result in incorrect derivaties.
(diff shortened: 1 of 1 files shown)


vgvassilev (approved):
LGTM!

vgvassilev:
Codecov seems to be acting up ;)

PR #206, opened 2021-03-03
Fix pow fails on different arg types in forward mode

sudo-panda:
Fixes the issue where pow doesn't work for forward mode when it has different type of arguments

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -80,18 +80,18 @@ namespace custom_derivatives {
   }
 #endif
 
-  template<typename T>
-  T pow_darg0(T x, T exponent) {
-    return exponent * pow(x, exponent-((T)1));
+  template<typename T1, typename T2>
+  decltype(pow(T1(), T2())) pow_darg0(T1 x, T2 exponent) {
+    return exponent * pow(x, exponent-((T2)1));
   }
 
-  template <typename T>
-  T pow_darg1(T x, T exponent) {
+  template <typename T1, typename T2>
+  decltype(pow(T1(), T2())) pow_darg1(T1 x, T2 exponent) {
     return pow(x, exponent) * log(x);
   }
 
-  template <typename T>
-  void pow_grad(T x, T exponent, T* result) {
+  template <typename T1, typename T2>
+  void pow_grad(T1 x, T2 exponent, decltype(pow(T1(), T2())) * result) {
     result[0] += pow_darg0(x, exponent);
     result[1] += pow_darg1(x, exponent);
   }
@@ -100,6 +100,18 @@ namespace custom_derivatives {
   T log_darg0(T x) {
     return 1.0/x;
   }
+
+  // FIXME: These math functions depend on promote_2 just like pow:
+  // atan2
+  // fmod
+  // copysign
+  // fdim
+  // fmax
+  // fmin
+  // hypot
+  // nextafter
+  // remainder
+  // remquo
 
 } // end namespace builtin_derivatives
 
--- a/test/FirstDerivative/BuiltinDerivatives.C
+++ b/test/FirstDerivative/BuiltinDerivatives.C
@@ -75,6 +75,43 @@ float f6(float x) {
 // CHECK-NEXT: return custom_derivatives::exp_darg0(x * x) * (_d_x * x + x * _d_x);
 // CHECK-NEXT: }
 
+float f7(float x) {
+  return pow(x, 2.0);
+}
+
+// CHECK: float f7_darg0(float x) {
+// CHECK-NEXT:   float _d_x = 1;
+// CHECK-NEXT:   return custom_derivatives::pow_darg0(x, 2.) * (_d_x + 0.);
+// CHECK-NEXT: }
+
+double f8(float x) {
+  return pow(x, 2);
+}
+
+// CHECK: double f8_darg0(float x) {
+// CHECK-NEXT:   float _d_x = 1;
+// CHECK-NEXT:   return custom_derivatives::pow_darg0(x, 2) * (_d_x + 0);
+// CHECK-NEXT: }
+
+float f9(float x, float y) {
+  return pow(x, y);
+}
+
+// CHECK: float f9_darg0(float x, float y) {
+// CHECK-NEXT:     float _d_x = 1;
+// CHECK-NEXT:     float _d_y = 0;
+// CHECK-NEXT:     return custom_derivatives::pow_darg0(x, y) * (_d_x + _d_y);
+// CHECK-NEXT: }
+
+double f10(float x, int y) {
+  return pow(x, y);
+}
+
+// CHECK: double f10_darg0(float x, int y) {
+// CHECK-NEXT:   float _d_x = 1;
+// CHECK-NEXT:   int _d_y = 0;
+// CHECK-NEXT:   return custom_derivatives::pow_darg0(x, y) * (_d_x + _d_y);
+// CHECK-NEXT: }
 
 int main () { //expected-no-diagnostics
   auto f1_darg0 = clad::differentiate(f1, 0);
@@ -100,6 +137,18 @@ int main () { //expected-no-diagnostics
 
   auto f6_darg0 = clad::differentiate(f6, 0);
   printf("Result is = %f\n", f6_darg0.execute(3)); //CHECK-EXEC: Result is = 48618.503906
+
+  auto f7_darg0 = clad::differentiate(f7, 0);
+  printf("Result is = %f\n", f7_darg0.execute(3)); //CHECK-EXEC: Result is = 6.000000
+
+  auto f8_darg0 = clad::differentiate(f8, 0);
+  printf("Result is = %f\n", f8_darg0.execute(3)); //CHECK-EXEC: Result is = 6.000000
+
+  auto f9_darg0 = clad::differentiate(f9, 0);
+  printf("Result is = %f\n", f9_darg0.execute(3, 4)); //CHECK-EXEC: Result is = 108.000000
+
+  auto f10_darg0 = clad::differentiate(f10, 0);
+  printf("Result is = %f\n", f10_darg0.execute(3, 4)); //CHECK-EXEC: Result is = 108.000000
 
   return 0;
 }


sudo-panda:
@vgvassilev I have fixed reverse mode differentiation of pow and forward mode differentiation of pow with respect to the first variable. I found the issue of forward mode differentiation not working for second variable while writing the tests so I am working on it now. Meanwhile could you please check the implementation of the first two fixes and let me know if I could improve it?

test/FirstDerivative/BuiltinDerivatives.C
```diff
@@ -1,3 +1,7 @@
+//
+// Created by sudo-panda on 06/03/21.
+//
```
vgvassilev, line 3: Do we need this?

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -2539,10 +2539,10 @@ namespace clad {
     std::size_t insertionPoint = getCurrentBlock(reverse).size();
     for (const Expr* Arg : CE->arguments()) {
       // Create temporary variables corresponding to derivative of each argument,
-      // so that they can be reffered to when arguments is visited. Variables
+      // so that they can be referred to when arguments is visited. Variables
       // will be initialized later after arguments is visited. This is done to
       // reduce cloning complexity and only clone once.
-      Expr* dArg = StoreAndRef(nullptr, Arg->getType(), reverse, "_r", /*force*/true);
+      Expr* dArg = StoreAndRef(nullptr, CE->getType(), reverse, "_r", /*force*/true);
```
vgvassilev, line 2545: Could you improve the commit message including why the current state was incorrect ( you may find [this](https://chris.beams.io/posts/git-commit/) a good read ).

In addition, the changes in the test cases should be squashed in that commit for additional documentation.

cc: @efremale

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -80,18 +124,21 @@ namespace custom_derivatives {
   }
 #endif
 
-  template<typename T>
-  T pow_darg0(T x, T exponent) {
-    return exponent * pow(x, exponent-((T)1));
+  template<typename T1, typename T2,
+      typename T3 = typename promote_2<T1, T2>::type>
+  T3 pow_darg0(T1 x, T2 exponent) {
```
vgvassilev, line 129: Same comments for commit message holds for this change. I'd like to clarify the rationale and the intent.

vgvassilev (commented):
Thanks for working on this! I added a few comments from the partial review of this PR.

test/FirstDerivative/BuiltinDerivatives.C
sudo-panda: No, sorry, Clion put it and it escaped my eye.

lib/Differentiator/DerivativeBuilder.cpp
sudo-panda: Sure

include/clad/Differentiator/BuiltinDerivatives.h
sudo-panda: Sure

sudo-panda:
@vgvassilev the fix for finding the derivative of pow w.r.t the exponent is actually the fix for multiarg expressions in forward mode. I will put up another pull request for that as it will cover more than just pow. The fix is to basically find the partial derivative of the the function with respect to each argument and then multiply it with the derivative of the argument w.r.t to the variable being differentiated.

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -46,6 +47,49 @@ namespace custom_derivatives {
     //}
   }// end namespace std
 
+  // Using anonymous namespace as these structs are only for local
+  // usage
+  namespace {
+    // promote sets type as double for all integer types
```
vgvassilev, line 53: Why do we need to promote integral types to doubles?

sudo-panda: If you see the commit message I have written the reason. 

To reiterate the return type of pow is decided by first promoting the integral types of the args to `double` and then the return type is set to the wider type of both the args. For example an `int` base with a `float` exponent will first get converted to a `double` base and a `float` exponent and the pow function will finally return a `double` being the wider of the two. So to get the appropriate return type of the functions `pow_darg0` and `pow_darg1` and the type of the array `result` in `pow_grad` we need to do the same.

vgvassilev: Ok, I was missing the fact that `std::pow` is very weirdly specified. Maybe worth noting...

I'd like to remove the reliance on the type_traits header. Can you think of a way forward?

lib/Differentiator/DerivativeBuilder.cpp
vgvassilev, line 2545: That change likely requires to update/extend the comment above and move the `CE->getType()` invariant out of the loop.

vgvassilev (commented):
Thanks for working on this. See comments inline.

include/clad/Differentiator/BuiltinDerivatives.h
sudo-panda: We can replace the `is_integral` with our own implementation and we will be able to remove the dependence.

vgvassilev: Alternatively, we could return `T` in the cases of `float` and `long double` and `double` for everything else (according to the pow2 spec).

sudo-panda: I was thinking where this would fail and the only possible case where it could have caused problems was when the user supplies data types that are not intended to be used but since we are passing the arguments to `pow`, `pow` itself will throw an error and hence we can proceed with this implementation.

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -80,18 +119,21 @@ namespace custom_derivatives {
   }
 #endif
 
-  template<typename T>
-  T pow_darg0(T x, T exponent) {
-    return exponent * pow(x, exponent-((T)1));
+  template<typename T1, typename T2,
+      typename T3 = typename promote_2<T1, T2>::type>
```
vgvassilev, line 123: I still think we can express the specification without having the `promote` specializations.

sudo-panda: Sure I will remove them.

sudo-panda:
@vgvassilev I had a doubt about the type of the temporary variables `_r0`, `_r1`, etc. in reverse mode. The value for these variables are obtained by the multiplication of `dfdx()` and the `OverloadedFunction` or the `ithResult` and later it(probably) gets stored in the `result` array. My question is which type is the correct type for _r0? Is it:

- The wider of `dfdx()` and `OverloadedFunction` / `ithResult` 
- The return type of the helper function (`OverloadedFunction` / `ithResult`)
- The type of the `result` array

vgvassilev:
> @vgvassilev I had a doubt about the type of the temporary variables `_r0`, `_r1`, etc. in reverse mode. The value for these variables are obtained by the multiplication of `dfdx()` and the `OverloadedFunction` or the `ithResult` and later it (might) get stored in the `result` array. My question is which type is the correct type for _r0? Is it:
> 
>     * The wider of `dfdx()` and `OverloadedFunction` / `ithResult`
> 
>     * The return type of the helper function (`OverloadedFunction` / `ithResult`)
> 
>     * The type of the `result` array

That's a tricky question -- I guess we don't want to lose precision so the type of the `result` array would be my conservative take.

sudo-panda:
@vgvassilev 
> That's a tricky question -- I guess we don't want to lose precision so the type of the `result` array would be my conservative take.

That would be my take too. I will proceed with this then.

sudo-panda:
@vgvassilev I will make another PR that fixes #214 and pow in reverse mode. This PR now only deals with fixing pow in forward mode.

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -80,18 +119,18 @@ namespace custom_derivatives {
...
-    return exponent * pow(x, exponent-((T)1));
+  template<typename T1, typename T2>
+  typename promote_2<T1, T2>::type pow_darg0(T1 x, T2 exponent) {
+    return exponent * pow(x, exponent-((T2)1));
   }
 
-  template <typename T>
-  T pow_darg1(T x, T exponent) {
+  template <typename T1, typename T2>
+  typename promote_2<T1, T2>::type pow_darg1(T1 x, T2 exponent) {
```
vgvassilev, line 128: The comments such as https://github.com/vgvassilev/clad/pull/215/files#r595950668 are valid here too.

sudo-panda: @vgvassilev could you please reply to my questions on the comments that you linked

vgvassilev: I do not see any questions -- could you paste a link?

sudo-panda: @vgvassilev I had a question here: https://github.com/vgvassilev/clad/pull/215#discussion_r596902201

And a doubt here: https://github.com/vgvassilev/clad/pull/215#discussion_r596899463

vgvassilev: I cannot open any of those. And I have not seen emails. Did you click on 'submit review' if they were pending. Otherwise please re-post them here.

sudo-panda: Yeah I didn't see the submit review button when I wrote those comments. Thanks for the tip.

vgvassilev: > I don't think that will work because T1 and T2 need to be promoted first before choosing the wider of the two.

Why?

sudo-panda: Because `decltype(T1() + T2())` will just select the wider type out of `T1` and `T2`, right? For example if `T1` and `T2` both are `int` then `T1() + T2()` will be `int` whereas we want it to be `double`

vgvassilev: Ok, fair point -- I had a different implementation in mind but apparently I cannot make it cover the cases this one covers.

Sevaral things to still improve:
* find a better name for `promote_2` -- this seems to be `pow2` specific;
* replace typedefs with template aliases (aka template<typename T> using...);
* use decltype instead of ```__typeof__```;
* remove the anonymous namespace -- anonymous namespace means internal linkage which means if the header file is included in two different translation units it will cause an ODR violation for its contents.
* move the opening curly on the same line -- in fact the entire specialization can go in a single line (assuming it fits in 80 columns).

sudo-panda: @vgvassilev  I applied the fix pointed out by @grimmmyshini [here](https://github.com/vgvassilev/clad/pull/215#discussion_r598321223)

This seems to be the most elegant looking solution. Could you please review it?

vgvassilev: That's what I was after -- one thing to keep in mind that using `decltype(pow)` coming `math.h` is a potential source of issues as `math.h` is very platform-dependent header. Could you add some tests around that area? I also see that we need too use the same 'promotion' techniques for a whole bunch of math functions (look for promote_2 in the cmath header). I suspect we should add those here, either in this PR or in a subsequent.

sudo-panda: @vgvassilev I wasn't able to find any issues that would creep up in the use of `decltype(pow)` due to `math.h` being platform-dependent. The only problem I found was that the accuracy of `pow` might be different in different platforms. Were you referring this?

Also I looked up the functions other than `pow` whose return type is dependent on `promote_2` :
- `atan2`
- `fmod`
- `copysign`
- `fdim`
- `fmax`
- `fmin`
- `hypot`
- `nextafter`
- `remainder`
- `remquo`

We don't have support for any of them in `BuiltinDerivatives.h`. Should we add support for more `math.h` functions?

vgvassilev: > @vgvassilev I wasn't able to find any issues that would creep up in the use of `decltype(pow)` due to `math.h` being platform-dependent. The only problem I found was that the accuracy of `pow` might be different in different platforms. Were you referring this?

Yes, I think so.


> 
> Also I looked up the functions other than `pow` whose return type is dependent on `promote_2` :
> 
>     * `atan2`
> 
>     * `fmod`
> 
>     * `copysign`
> 
>     * `fdim`
> 
>     * `fmax`
> 
>     * `fmin`
> 
>     * `hypot`
> 
>     * `nextafter`
> 
>     * `remainder`
> 
>     * `remquo`
> 
> 
> We don't have support for any of them in `BuiltinDerivatives.h`. Should we add support for more `math.h` functions?

I think for the moment we can just enumerate them in a `// FIXME:` comment.

sudo-panda: > > @vgvassilev I wasn't able to find any issues that would creep up in the use of `decltype(pow)` due to `math.h` being platform-dependent. The only problem I found was that the accuracy of `pow` might be different in different platforms. Were you referring this?
> 
> Yes, I think so.

I tried to find a way to run `pow` on a c++ compiler online that supports different platforms but could not find any. Do you have any other way that I could add those tests?

vgvassilev: Unfortunately, I have no other suggestions but I think we should move forward with that approach.

sudo-panda: @vgvassilev do you mean to move forward with the `decltype` approach or the adding the platform dependent tests approach?

vgvassilev (commented):
I think we are almost good to go. Can you squash all commits in one?

sudo-panda:
> I think we are almost good to go. Can you squash all commits in one?

@vgvassilev Done

vgvassilev (approved):
LGTM!

Issue #139, opened 2019-07-19
Codegen Error for Class Function Differentiation in clad::gradient


jacklqiu:
Differentiating class functions yields codegen errors
```
struct Test {
  double a_b_cubed_class(double a, double b, double c)
  {
    return a * a * a + b * b * b + c * c * c;
  }
};


int main()
{
  auto h = clad::gradient(&Test::a_b_cubed_class, "a");
  h.dump();
}
```

```
#0 0x00007f1ff14991fa llvm::sys::PrintStackTrace(llvm::raw_ostream&) (/usr/lib/llvm-5.0/bin/../lib/libLLVM-5.0.so.1+0x7a91fa)
#1 0x00007f1ff1497416 llvm::sys::RunSignalHandlers() (/usr/lib/llvm-5.0/bin/../lib/libLLVM-5.0.so.1+0x7a7416)
#2 0x00007f1ff149754b (/usr/lib/llvm-5.0/bin/../lib/libLLVM-5.0.so.1+0x7a754b)
#3 0x00007f1ff4312890 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x12890)
#4 0x000055c02d53efd4 clang::CodeGen::CGCXXABI::canCopyArgument(clang::CXXRecordDecl const*) const (/usr/lib/llvm-5.0/bin/clang+0x5cefd4)
#5 0x000055c02d4df64b (/usr/lib/llvm-5.0/bin/clang+0x56f64b)
#6 0x000055c02d51ad16 (/usr/lib/llvm-5.0/bin/clang+0x5aad16)
#7 0x000055c02d540e8e clang::CodeGen::CodeGenTypes::arrangeLLVMFunctionInfo(clang::CanQual<clang::Type>, bool, bool, llvm::ArrayRef<clang::CanQual<clang::Type> >, clang::FunctionType::ExtInfo, llvm::ArrayRef<clang::FunctionProtoType::ExtParameterInfo>, clang::CodeGen::RequiredArgs) (/usr/lib/llvm-5.0/bin/clang+0x5d0e8e)
#8 0x000055c02d541caa (/usr/lib/llvm-5.0/bin/clang+0x5d1caa)
[...]

PR #212, opened 2021-03-14
Add support for using casts, * and & operators with fn argument

### LINKED ISSUE #211: Error when using function pointer variable to pass function address in differentiation calls
Minimal reproducible example:
```c++
#include <iostream>

struct Expression {
  Expression(double p_x=0,double p_y=0): x(p_x), y(p_y) {}
  double x,y;
  double mem_fn(double i,double j) {
    return x*i + y*j;
  }
};
int main()
{
  auto ptr = &Expression::mem_fn;
  auto d_mem_fn = clad::differentiate(ptr,"i");
  Expression expr(3,5);
  std::cout<<d_mem_fn.execute(expr,2,3)<<"\n";
}
```
<details>
<summary>
Error stack trace
</summary>

```

clang-9: /home/parth/Programming/gsoc/clad-dev-env/src/include/llvm/Support/Casting.h:264: typename llvm::cast_retty<X, Y*>::ret_type llvm::cast(Y*) [with X = clang::FunctionDecl; Y = clang::ValueDecl; typename llvm::cast_retty<X, Y*>::ret_type = clang::FunctionDecl*]: Assertion `isa<X>(Val) && "cast<Ty>() argument of incompatible type!"' failed.
Stack dump:
0.	Program arguments: /home/parth/Programming/gsoc/clad-dev-env/inst/bin/clang-9 -cc1 -triple x86_64-unknown-linux-gnu -emit-obj -mrelax-all -disable-free -main-file-name mem-fn-forward-bug.cpp -mrelocation-model static -mthread-model posix -mdisable-fp-elim -fmath-errno -masm-verbose -mconstructor-aliases -munwind-tables -fuse-init-array -target-cpu x86-64 -dwarf-column-info -debug-info-kind=limited -dwarf-version=4 -debugger-tuning=gdb -resource-dir /home/parth/Programming/gsoc/clad-dev-env/inst/lib/clang/9.0.1 -I /home/parth/Programming/gsoc/clad-dev-env/inst/include -internal-isystem /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9 -internal-isystem /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/x86_64-linux-gnu/c++/9 -internal-isystem /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/x86_64-linux-gnu/c++/9 -internal-isystem /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/backward -internal-isystem /usr/local/include -internal-isystem /home/parth/Programming/gsoc/clad-dev-env/inst/lib/clang/9.0.1/include -internal-externc-isystem /usr/include/x86_64-linux-gnu -internal-externc-isystem /include -internal-externc-isystem /usr/include -std=c++11 -fdeprecated-macro -fdebug-compilation-dir /home/parth/Programming/gsoc/clad-tutorial -ferror-limit 19 -fmessage-length 0 -fobjc-runtime=gcc -fcxx-exceptions -fexceptions -fdiagnostics-show-option -add-plugin clad -load /home/parth/Programming/gsoc/clad-dev-env/inst/lib/clad.so -faddrsig -o /tmp/mem-fn-forward-bug-4402a0.o -x c++ mem-fn-forward-bug.cpp 
 #0 0x000055b330c4f5cf llvm::sys::PrintStackTrace(llvm::raw_ostream&) /home/parth/Programming/gsoc/clad-dev-env/src/lib/Support/Unix/Signals.inc:533:22
 #1 0x000055b330c4f666 PrintStackTraceSignalHandler(void*) /home/parth/Programming/gsoc/clad-dev-env/src/lib/Support/Unix/Signals.inc:594:1
 #2 0x000055b330c4d500 llvm::sys::RunSignalHandlers() /home/parth/Programming/gsoc/clad-dev-env/src/lib/Support/Signals.cpp:68:20
 #3 0x000055b330c4ef59 SignalHandler(int) /home/parth/Programming/gsoc/clad-dev-env/src/lib/Support/Unix/Signals.inc:385:1
 #4 0x00007f0f3cb4e3c0 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x153c0)
 #5 0x00007f0f3b04e18b raise /build/glibc-eX1tMB/glibc-2.31/signal/../sysdeps/unix/sysv/linux/raise.c:51:1
 #6 0x00007f0f3b02d859 abort /build/glibc-eX1tMB/glibc-2.31/stdlib/abort.c:81:7
 #7 0x00007f0f3b02d729 get_sysdep_segment_value /build/glibc-eX1tMB/glibc-2.31/intl/loadmsgcat.c:509:8
 #8 0x00007f0f3b02d729 _nl_load_domain /build/glibc-eX1tMB/glibc-2.31/intl/loadmsgcat.c:970:34
 #9 0x00007f0f3b03ef36 (/lib/x86_64-linux-gnu/libc.so.6+0x36f36)
#10 0x00007f0f3aea64d8 llvm::cast_retty<clang::FunctionDecl, clang::ValueDecl*>::ret_type llvm::cast<clang::FunctionDecl, clang::ValueDecl>(clang::ValueDecl*) /home/parth/Programming/gsoc/clad-dev-env/src/include/llvm/Support/Casting.h:264:3
#11 0x00007f0f3aea429b clad::DiffCollector::VisitCallExpr(clang::CallExpr*) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clad/lib/Differentiator/DiffPlanner.cpp:232:42
#12 0x00007f0f3aed65b4 clang::RecursiveASTVisitor<clad::DiffCollector>::WalkUpFromCallExpr(clang::CallExpr*) /home/parth/Programming/gsoc/clad-dev-env/obj/tools/clad/../clang/include/clang/AST/StmtNodes.inc:765:1
#13 0x00007f0f3aeea509 clang::RecursiveASTVisitor<clad::DiffCollector>::TraverseCallExpr(clang::CallExpr*, llvm::SmallVectorImpl<llvm::PointerIntPair<clang::Stmt*, 1u, bool, llvm::PointerLikeTypeTraits<clang::Stmt*>, llvm::PointerIntPairInfo<clang::Stmt*, 1u, llvm::PointerLikeTypeTraits<clang::Stmt*> > > >*) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clad/../clang/include/clang/AST/RecursiveASTVisitor.h:2473:1
#14 0x00007f0f3aec71df clang::RecursiveASTVisitor<clad::DiffCollector>::dataTraverseNode(clang::Stmt*, llvm::SmallVectorImpl<llvm::PointerIntPair<clang::Stmt*, 1u, bool, llvm::PointerLikeTypeTraits<clang::Stmt*>, llvm::PointerIntPairInfo<clang::Stmt*, 1u, llvm::PointerLikeTypeTraits<clang::Stmt*> > > >*) /home/parth/Programming/gsoc/clad-dev-env/obj/tools/clad/../clang/include/clang/AST/StmtNodes.inc:765:1
#15 0x00007f0f3aeb1f55 clang::RecursiveASTVisitor<clad::DiffCollector>::TraverseStmt(clang::Stmt*, llvm::SmallVectorImpl<llvm::PointerIntPair<clang::Stmt*, 1u, bool, llvm::PointerLikeTypeTraits<clang::Stmt*>, llvm::PointerIntPairInfo<clang::Stmt*, 1u, llvm::PointerLikeTypeTraits<clang::Stmt*> > > >*) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clad/../clang/include/clang/AST/RecursiveASTVisitor.h:655:7
#16 0x00007f0f3aeb553b clang::RecursiveASTVisitor<clad::DiffCollector>::TraverseVarHelper(clang::VarDecl*) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clad/../clang/include/clang/AST/RecursiveASTVisitor.h:2086:5
#17 0x00007f0f3aeabe61 clang::RecursiveASTVisitor<clad::DiffCollector>::TraverseVarDecl(clang::VarDecl*) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clad/../clang/include/clang/AST/RecursiveASTVisitor.h:2090:1
#18 0x00007f0f3aea5e72 clang::RecursiveASTVisitor<clad::DiffCollector>::TraverseDecl(clang::Decl*) /home/parth/Programming/gsoc/clad-dev-env/obj/tools/clad/../clang/include/clang/AST/DeclNodes.inc:447:1
#19 0x00007f0f3aedc7d1 clang::RecursiveASTVisitor<clad::DiffCollector>::TraverseDeclStmt(clang::DeclStmt*, llvm::SmallVectorImpl<llvm::PointerIntPair<clang::Stmt*, 1u, bool, llvm::PointerLikeTypeTraits<clang::Stmt*>, llvm::PointerIntPairInfo<clang::Stmt*, 1u, llvm::PointerLikeTypeTraits<clang::Stmt*> > > >*) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clad/../clang/include/clang/AST/RecursiveASTVisitor.h:2169:1
#20 0x00007f0f3aec633f clang::RecursiveASTVisitor<clad::DiffCollector>::dataTraverseNode(clang::Stmt*, llvm::SmallVectorImpl<llvm::PointerIntPair<clang::Stmt*, 1u, bool, llvm::PointerLikeTypeTraits<clang::Stmt*>, llvm::PointerIntPairInfo<clang::Stmt*, 1u, llvm::PointerLikeTypeTraits<clang::Stmt*> > > >*) /home/parth/Programming/gsoc/clad-dev-env/obj/tools/clad/../clang/include/clang/AST/StmtNodes.inc:97:1
#21 0x00007f0f3aeb1f55 clang::RecursiveASTVisitor<clad::DiffCollector>::TraverseStmt(clang::Stmt*, llvm::SmallVectorImpl<llvm::PointerIntPair<clang::Stmt*, 1u, bool, llvm::PointerLikeTypeTraits<clang::Stmt*>, llvm::PointerIntPairInfo<clang::Stmt*, 1u, llvm::PointerLikeTypeTraits<clang::Stmt*> > > >*) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clad/../clang/include/clang/AST/RecursiveASTVisitor.h:655:7
#22 0x00007f0f3aeb50ae clang::RecursiveASTVisitor<clad::DiffCollector>::TraverseFunctionHelper(clang::FunctionDecl*) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clad/../clang/include/clang/AST/RecursiveASTVisitor.h:2031:5
#23 0x00007f0f3aeab490 clang::RecursiveASTVisitor<clad::DiffCollector>::TraverseFunctionDecl(clang::FunctionDecl*) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clad/../clang/include/clang/AST/RecursiveASTVisitor.h:2036:1
#24 0x00007f0f3aea5cda clang::RecursiveASTVisitor<clad::DiffCollector>::TraverseDecl(clang::Decl*) /home/parth/Programming/gsoc/clad-dev-env/obj/tools/clad/../clang/include/clang/AST/DeclNodes.inc:395:1
#25 0x00007f0f3aea3b9c clad::DiffCollector::DiffCollector(clang::DeclGroupRef, std::vector<clang::SourceRange, std::allocator<clang::SourceRange> >&, llvm::SmallSet<clang::Decl const*, 16u, std::less<clang::Decl const*> > const&, llvm::SmallVector<clad::DiffRequest, 16u>&, clang::Sema&) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clad/lib/Differentiator/DiffPlanner.cpp:159:19
#26 0x00007f0f3adb5665 clad::plugin::CladPlugin::HandleTopLevelDecl(clang::DeclGroupRef) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clad/tools/ClangPlugin.cpp:128:35
#27 0x000055b33171ed18 clang::MultiplexConsumer::HandleTopLevelDecl(clang::DeclGroupRef) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clang/lib/Frontend/MultiplexConsumer.cpp:271:25
#28 0x000055b3336581fc clang::ParseAST(clang::Sema&, bool, bool) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clang/lib/Parse/ParseAST.cpp:162:20
#29 0x000055b3316ddb35 clang::ASTFrontendAction::ExecuteAction() /home/parth/Programming/gsoc/clad-dev-env/src/tools/clang/lib/Frontend/FrontendAction.cpp:1041:11
#30 0x000055b331f83701 clang::CodeGenAction::ExecuteAction() /home/parth/Programming/gsoc/clad-dev-env/src/tools/clang/lib/CodeGen/CodeGenAction.cpp:1060:1
#31 0x000055b3316dd48c clang::FrontendAction::Execute() /home/parth/Programming/gsoc/clad-dev-env/src/tools/clang/lib/Frontend/FrontendAction.cpp:938:38
#32 0x000055b331675b4c clang::CompilerInstance::ExecuteAction(clang::FrontendAction&) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clang/lib/Frontend/CompilerInstance.cpp:944:42
#33 0x000055b331844bf2 clang::ExecuteCompilerInvocation(clang::CompilerInstance*) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clang/lib/FrontendTool/ExecuteCompilerInvocation.cpp:291:38
#34 0x000055b32f5f27b7 cc1_main(llvm::ArrayRef<char const*>, char const*, void*) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clang/tools/driver/cc1_main.cpp:249:40
#35 0x000055b32f5e74ad ExecuteCC1Tool(llvm::ArrayRef<char const*>, llvm::StringRef) /home/parth/Programming/gsoc/clad-dev-env/src/tools/clang/tools/driver/driver.cpp:309:64
#36 0x000055b32f5e7b9b main /home/parth/Programming/gsoc/clad-dev-env/src/tools/clang/tools/driver/driver.cpp:381:26
#37 0x00007f0f3b02f0b3 __libc_start_main /build/glibc-eX1tMB/glibc-2.31/csu/../csu/libc-start.c:342:3
#38 0x000055b32f5e5c6e _start (/home/parth/Programming/gsoc/clad-dev-env/inst/bin/clang-9+0x2248c6e)
clang-9: error: unable to execute command: Aborted (core dumped)
clang-9: error: clang frontend command failed due to signal (use -v to see invocation)
clang version 9.0.1 (https://github.com/llvm-mirror/clang.git bed37d4b18ce6993fc2567e8de61f6b30fd0cb75) (https://github.com/llvm-mirror/llvm.git c2f5309b216cb02a0aae17353549d985b5b05635)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /home/parth/Programming/gsoc/clad-dev-env/inst/bin
clang-9: note: diagnostic msg: PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace, preprocessed source, and associated run script.
clang-9: note: diagnostic msg: 
********************

PLEASE ATTACH THE FOLLOWING FILES TO THE BUG REPORT:
Preprocessed source(s) and associated run script(s) are located at:
clang-9: note: diagnostic msg: /tmp/mem-fn-forward-bug-94d82b.cpp
clang-9: note: diagnostic msg: /tmp/mem-fn-forward-bug-94d82b.sh
clang-9: note: diagnostic msg: 

********************
Namespace(clad_inst=PosixPath('/home/parth/Programming/gsoc/clad-dev-env/inst'), clang=PosixPath('/home/parth/Programming/gsoc/clad-dev-env/inst/bin/clang++'), i=PosixPath('mem-fn-forward-bug.cpp'), o=PosixPath('build/mem-fn-forward-bug'), plugin_arg=None, plugin_arg_clad=None, run=True)
/home/parth/Programming/gsoc/clad-dev-env/inst/bin/clang++ -g -Xclang -add-plugin -Xclang clad -Xclang -load -Xclang /home/parth/Programming/gsoc/clad-dev-env/inst/lib/clad.so -I/home/parth/Programming/gsoc/clad-dev-env/inst/include -x c++ -std=c++11 mem-fn-forward-bug.cpp -o build/mem-fn-forward-bug && build/mem-fn-forward-bug


```

</details>

The cause of the issue is the incorrect determination of the function type by _code_ being used to determine if the function to be differentiated is member or non-member function in `lib/Differentiator/DiffPlanner.cpp`. 
In this file, in functions `updateCall` and `getArgFunction`, code responsible for determining if function to be differentiated is member or non-member function is, 
```c++
  // Handle the case of function pointer.
  if (auto ICE = dyn_cast<ImplicitCastExpr>(call->getArg(0))){
      oldDRE = dyn_cast<DeclRefExpr>(ICE->getSubExpr());
  }
    // Handle the case of member function.
  else if (auto UnOp = dyn_cast<UnaryOperator>(call->getArg(0))){
    oldDRE = dyn_cast<DeclRefExpr>(UnOp->getSubExpr());
  }
```
but in example provided above(_minimal reproducible example_), `clad::differentiate(ptr,"i")`, will incorrectly map to the if condition which should be responsible for handling non-member functions , and this is causing the issue. I am working on the fix for this issue.

parth-07:
Progress: 
- [x] Fix the issue
- [x] Add relevant regression tests

This PR aims to add support for differentiation calls with Unary operators(`&`, `*`), parantheses and casts (both C and C++ style) with the function argument.
For instance, with this PR, differentiation calls like these will work correctly: 

```c++
auto d_memFnPtrToPtr_1ParPar = clad::differentiate(*(*(&memFnPtrToPtr)),"i");
auto d_memFnPtrToPtrToPtr_1Par = clad::differentiate(***(&memFnPtrToPtrToPtr),"i");
auto d_memFnPtrToPtr_1 = clad::differentiate(**&memFnPtrToPtr,"i");
auto d_nonMemFnStaticCast = clad::differentiate(static_cast<decltype(&nonMemFn)>(nonMemFn),"i");
auto d_memFnCStyleCast = clad::differentiate((decltype(&Expression::memFn))(&Expression::memFn),"i");
```

Fixes #211

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -1,6 +1,7 @@
 #include "clad/Differentiator/DiffPlanner.h"
 
 #include "clang/AST/ASTContext.h"
+#include "clang/AST/RecursiveASTVisitor.h"
 #include "clang/Basic/SourceManager.h"
 #include "clang/Sema/Sema.h"
 #include "clang/Sema/TemplateDeduction.h"
@@ -13,6 +14,47 @@ using namespace clang;
 
 namespace clad {
   static SourceLocation noLoc;
+  
+  // Returns DeclRefExpr of function argument of differentiation calls.
+  // If argument is passed for relevantAncestor, then it's value will be 
+  // modified in-place to contain relevant ancestor of the function 
+  // argument's DeclRefExpr
+  // Here relevant ancestor is nearest ancestor of ImplicitCastExpr or 
+  // UnaryOperator type.
+  DeclRefExpr* getArgFunction(CallExpr* call,
+                              Expr** relevantAncestor = nullptr) {
+    struct Finder :
+      RecursiveASTVisitor<Finder> {
+        DeclRefExpr* m_FnDRE = nullptr;
+        Expr** m_RelevantAncestor = nullptr;
+        Finder(Expr** relevantAncestor) : m_RelevantAncestor(relevantAncestor) {}
+        bool VisitExpr(Expr* E) {
+          if (m_RelevantAncestor) {
+            switch (E->getStmtClass()) {
+              case Stmt::StmtClass::ImplicitCastExprClass: LLVM_FALLTHROUGH;
+              case Stmt::StmtClass::UnaryOperatorClass:
+                *m_RelevantAncestor = E;
+                break; 
+            }
+          }
+          if (auto DRE = dyn_cast<DeclRefExpr>(E)) {
+            if (auto VD = dyn_cast<VarDecl>(DRE->getDecl()))
+              TraverseStmt(VD->getInit());
+            else if (auto FD = dyn_cast<FunctionDecl>(DRE->getDecl()))
+              m_FnDRE = DRE;
+            return false;  
+          }
+          return true; 
+        }
+      } finder(relevantAncestor);
+
+      assert(cast<NamespaceDecl>(call->getDirectCallee()->
+             getDeclContext())->getName() == "clad" &&
+             "Should be called for clad:: special functions!");
+
+      finder.TraverseStmt(call->getArg(0));
+      return finder.m_FnDRE;
+  }
 
   void DiffRequest::updateCall(FunctionDecl* FD, Sema& SemaRef) {
     CallExpr* call = this->CallContext;
@@ -21,16 +63,10 @@ namespace clad {
     assert(call && "Must be set");
     assert(FD && "Trying to update with null FunctionDecl");
 
-    DeclRefExpr* oldDRE = nullptr;
-    // Handle the case of function pointer.
-    if (auto ICE = dyn_cast<ImplicitCastExpr>(call->getArg(0))){
-      oldDRE = dyn_cast<DeclRefExpr>(ICE->getSubExpr());
-    }
-    // Handle the case of member function.
-    else if (auto UnOp = dyn_cast<UnaryOperator>(call->getArg(0))){
-      oldDRE = dyn_cast<DeclRefExpr>(UnOp->getSubExpr());
-    }
-    else
+    Expr* oldArgDREParent = nullptr;
+    DeclRefExpr* oldDRE = getArgFunction(call, &oldArgDREParent);
+
+    if (!oldDRE)
       llvm_unreachable("Trying to differentiate something unsupported");
 
     ASTContext& C = SemaRef.getASTContext();
@@ -42,8 +78,7 @@ namespace clad {
     // using call->setArg(0, DRE) seems to be sufficient,
     // though the real AST allways contains the ImplicitCastExpr (function ->
     // function ptr cast) or UnaryOp (method ptr call).
-    auto oldArg = call->getArg(0);
-    if (auto oldCast = dyn_cast<ImplicitCastExpr>(oldArg)) {
+    if (auto oldCast = dyn_cast<ImplicitCastExpr>(oldArgDREParent)) {
       // Cast function to function pointer.
       auto newCast = ImplicitCastExpr::Create(C,
                                               C.getPointerType(FD->getType()),
@@ -54,7 +89,7 @@ namespace clad {
                                               CLAD_COMPAT_CLANG12_CastExpr_GetFPO(oldCast));
       call->setArg(0, newCast);
     }
-    else if (auto oldUnOp = dyn_cast<UnaryOperator>(oldArg)) {
+    else if (auto oldUnOp = dyn_cast<UnaryOperator>(oldArgDREParent)) {
       // Add the "&" operator
       auto newUnOp = SemaRef.BuildUnaryOp(nullptr,
                                           noLoc,
@@ -159,20 +194,6 @@ namespace clad {
         continue;
       TraverseDecl(D);
     }
-  }
-
-  DeclRefExpr* getArgFunction(CallExpr* E) {
-    if (E->getNumArgs() == 0)
-      return nullptr;
-    Expr* arg = E->getArg(0);
-    // Handle the case of function.
-    if (ImplicitCastExpr* ICE = dyn_cast<ImplicitCastExpr>(arg))
-      return dyn_cast<DeclRefExpr>(ICE->getSubExpr());
-    // Handle the case of member function.
-    else if (UnaryOperator* UnOp = dyn_cast<UnaryOperator>(arg))
-      return dyn_cast<DeclRefExpr>(UnOp->getSubExpr());
-    else
-      return nullptr;
   }
 
   bool DiffCollector::isInInterval(SourceLocation Loc) const {
--- a/test/ForwardMode/Pointers.C
+++ b/test/ForwardMode/Pointers.C
@@ -1,0 +1,163 @@
+// RUN: %cladclang %s -I%S/../../include -oPointers.out 2>&1 | FileCheck %s
+// RUN: ./Pointers.out | FileCheck -check-prefix=CHECK-EXEC %s
+// CHECK-NOT: {{.*error|warning|note:.*}}
+
+#include "clad/Differentiator/Differentiator.h"
+
+class Expression {
+public:
+  Expression() {}
+  double memFn(double i) {
+    return i*i;
+  }
+
+  // CHECK: double memFn_darg0(double i) {
+  // CHECK-NEXT:     double _d_i = 1;
+  // CHECK-NEXT:     return _d_i * i + i * _d_i;
+  // CHECK-NEXT: }
+
+};
+
+double nonMemFn(double i) {
+  return 5*i;
+}
+
+// CHECK: double nonMemFn_darg0(double i) {
+// CHECK-NEXT:     double _d_i = 1;
+// CHECK-NEXT:     return 0 * i + 5 * _d_i;
+// CHECK-NEXT: }
+
+#define MEM_FN_TEST(var)\
+printf("%.2f\n",var.execute(expr,7));
+
+#define NON_MEM_FN_TEST(var)\
+printf("%.2f\n",var.execute(3));
+
+int main() {
+  Expression expr;
+
+  auto memFnPtr = &Expression::memFn;
+  auto memFnPtrToPtr = &memFnPtr;
+  auto memFnPtrToPtrToPtr = &memFnPtrToPtr;
+  auto memFnIndirectPtr = memFnPtr;
+  auto memFnIndirectIndirectPtr = memFnIndirectPtr;
+
+  auto d_memFnPtr = clad::differentiate(memFnPtr, "i");
+  auto d_memFnPtrPar = clad::differentiate((memFnPtr), "i");
+  auto d_memFnPtrToPtr = clad::differentiate(*memFnPtrToPtr, "i");
+  auto d_memFnPtrToPtr_1 = clad::differentiate(**&memFnPtrToPtr, "i");
+  auto d_memFnPtrToPtr_1Par = clad::differentiate(**(&memFnPtrToPtr), "i");
+  auto d_memFnPtrToPtr_1ParPar = clad::differentiate(*(*(&memFnPtrToPtr)), "i");
+  auto d_memFnPtrToPtrToPtr = clad::differentiate(**memFnPtrToPtrToPtr, "i");
+  auto d_memFnPtrToPtrToPtrPar = clad::differentiate(*(*memFnPtrToPtrToPtr), "i");
+  auto d_memFnPtrToPtrToPtrParPar = clad::differentiate(((*(*memFnPtrToPtrToPtr))), "i");
+  auto d_memFnPtrToPtrToPtr_1 = clad::differentiate(***&memFnPtrToPtrToPtr, "i");
+  auto d_memFnPtrToPtrToPtr_1Par = clad::differentiate(***(&memFnPtrToPtrToPtr), "i");
+  auto d_memFnPtrToPtrToPtr_1ParPar = clad::differentiate(**(*(&memFnPtrToPtrToPtr)), "i");
+  auto d_memFnPtrToPtrToPtr_1ParParPar = clad::differentiate(*(*(*(&memFnPtrToPtrToPtr))), "i");
+  auto d_memFnIndirectPtr = clad::differentiate(memFnIndirectPtr, "i");
+  auto d_memFnIndirectIndirectPtr = clad::differentiate(memFnIndirectIndirectPtr, "i");
+  auto d_memFnStaticCast = clad::differentiate(static_cast<decltype(&Expression::memFn)>(&Expression::memFn), "i");
+  auto d_memFnReinterpretCast = clad::differentiate(reinterpret_cast<decltype(&Expression::memFn)>(&Expression::memFn), "i");
+  auto d_memFnCStyleCast = clad::differentiate((decltype(&Expression::memFn))(&Expression::memFn), "i");
+
+  MEM_FN_TEST(d_memFnPtr); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnPtrPar); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnPtrToPtr); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnPtrToPtr_1); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnPtrToPtr_1Par); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnPtrToPtr_1ParPar); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnPtrToPtrToPtr); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnPtrToPtrToPtrPar); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnPtrToPtrToPtrParPar); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnPtrToPtrToPtr_1); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnPtrToPtrToPtr_1Par); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnPtrToPtrToPtr_1ParPar); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnPtrToPtrToPtr_1ParParPar); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnIndirectPtr); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnIndirectIndirectPtr); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnStaticCast); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnReinterpretCast); // CHECK-EXEC: 14.00
+
+  MEM_FN_TEST(d_memFnCStyleCast); // CHECK-EXEC: 14.00
+
+  auto nonMemFnPtr = &nonMemFn;
+  auto nonMemFnPtrToPtr = &nonMemFnPtr;
+  auto nonMemFnPtrToPtrToPtr = &nonMemFnPtrToPtr;
+  auto nonMemFnIndirectPtr = nonMemFnPtr;
+  auto nonMemFnIndirectIndirectPtr = nonMemFnIndirectPtr;
+
+  auto d_nonMemFn = clad::differentiate(nonMemFn, "i");
+  auto d_nonMemFnPar = clad::differentiate((nonMemFn), "i");
+  auto d_nonMemFnPtr = clad::differentiate(nonMemFnPtr, "i");
+  auto d_nonMemFnPtrToPtr = clad::differentiate(*nonMemFnPtrToPtr, "i");
+  auto d_nonMemFnPtrToPtrPar = clad::differentiate((*(nonMemFnPtrToPtr)), "i");
+  auto d_nonMemFnPtrToPtr_1 = clad::differentiate(**&nonMemFnPtrToPtr, "i");
+  auto d_nonMemFnPtrToPtr_1Par = clad::differentiate(**(&nonMemFnPtrToPtr), "i");
+  auto d_nonMemFnPtrToPtr_1ParPar = clad::differentiate(*(*(&nonMemFnPtrToPtr)), "i");
+  auto d_nonMemFnPtrToPtrToPtr = clad::differentiate(**nonMemFnPtrToPtrToPtr, "i");
+  auto d_nonMemFnPtrToPtrToPtr_1 = clad::differentiate(***&nonMemFnPtrToPtrToPtr, "i");
+  auto d_nonMemFnPtrToPtrToPtr_1Par = clad::differentiate(***(&nonMemFnPtrToPtrToPtr), "i");
+  auto d_nonMemFnPtrToPtrToPtr_1ParPar = clad::differentiate(*(**(&nonMemFnPtrToPtrToPtr)), "i");
+  auto d_nonMemFnPtrToPtrToPtr_1ParParPar = clad::differentiate((*(**((&nonMemFnPtrToPtrToPtr)))), "i");
+  auto d_nonMemFnIndirectPtr = clad::differentiate(nonMemFnIndirectPtr, "i");
+  auto d_nonMemFnIndirectIndirectPtr = clad::differentiate(nonMemFnIndirectIndirectPtr, "i");
+  auto d_nonMemFnStaticCast = clad::differentiate(static_cast<decltype(&nonMemFn)>(nonMemFn), "i");
+  auto d_nonMemFnReinterpretCast = clad::differentiate(reinterpret_cast<decltype(&nonMemFn)>(nonMemFn), "i");
+  auto d_nonMemFnCStyleCast = clad::differentiate((decltype(&nonMemFn))(nonMemFn), "i");
+
+  NON_MEM_FN_TEST(d_nonMemFn); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPar); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtr); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtr); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtrPar); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtr_1); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtr_1Par); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtr_1ParPar); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtrToPtr); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtrToPtr_1); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtrToPtr_1Par); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtrToPtr_1ParPar); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtrToPtr_1ParParPar); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnIndirectPtr); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnIndirectIndirectPtr); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnStaticCast); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnReinterpretCast); // CHECK-EXEC: 5.00
+
+  NON_MEM_FN_TEST(d_nonMemFnCStyleCast); // CHECK-EXEC: 5.00
+
+
+
+}
--- a/test/Gradient/Pointers.C
+++ b/test/Gradient/Pointers.C
@@ -1,0 +1,96 @@
+// RUN: %cladclang %s -I%S/../../include -oPointers.out 2>&1 | FileCheck %s
+// RUN: ./Pointers.out | FileCheck -check-prefix=CHECK-EXEC %s
+// CHECK-NOT: {{.*error|warning|note:.*}}
+
+#include "clad/Differentiator/Differentiator.h"
+
+double nonMemFn(double i) {
+  return i*i;
+}
+
+// CHECK: void nonMemFn_grad(double i, double *_result) {
+// CHECK-NEXT:     double _t0;
+// CHECK-NEXT:     double _t1;
+// CHECK-NEXT:     _t1 = i;
+// CHECK-NEXT:     _t0 = i;
+// CHECK-NEXT:     double nonMemFn_return = _t1 * _t0;
+// CHECK-NEXT:     goto _label0;
+// CHECK-NEXT:   _label0:
+// CHECK-NEXT:     {
+// CHECK-NEXT:         double _r0 = 1 * _t0;
+// CHECK-NEXT:         _result[0UL] += _r0;
+// CHECK-NEXT:         double _r1 = _t1 * 1;
+// CHECK-NEXT:         _result[0UL] += _r1;
+// CHECK-NEXT:     }
+// CHECK-NEXT: }
+
+#define NON_MEM_FN_TEST(var)\
+res[0]=0;\
+var.execute(5,res);\
+printf("%.2f\n", res[0]);
+
+int main() {
+  auto nonMemFnPtr = &nonMemFn;
+  auto nonMemFnPtrToPtr = &nonMemFnPtr;
+  auto nonMemFnPtrToPtrToPtr = &nonMemFnPtrToPtr;
+  auto nonMemFnIndirectPtr = nonMemFnPtr;
+  auto nonMemFnIndirectIndirectPtr = nonMemFnIndirectPtr;
+
+  double res[2];
+
+  auto d_nonMemFn = clad::gradient(nonMemFn, "i");
+  auto d_nonMemFnPar = clad::gradient((nonMemFn), "i");
+  auto d_nonMemFnPtr = clad::gradient(nonMemFnPtr, "i");
+  auto d_nonMemFnPtrToPtr = clad::gradient(*nonMemFnPtrToPtr, "i");
+  auto d_nonMemFnPtrToPtrPar = clad::gradient((*(nonMemFnPtrToPtr)), "i");
+  auto d_nonMemFnPtrToPtr_1 = clad::gradient(**&nonMemFnPtrToPtr, "i");
+  auto d_nonMemFnPtrToPtr_1Par = clad::gradient(**(&nonMemFnPtrToPtr), "i");
+  auto d_nonMemFnPtrToPtr_1ParPar = clad::gradient(*(*(&nonMemFnPtrToPtr)), "i");
+  auto d_nonMemFnPtrToPtrToPtr = clad::gradient(**nonMemFnPtrToPtrToPtr, "i");
+  auto d_nonMemFnPtrToPtrToPtr_1 = clad::gradient(***&nonMemFnPtrToPtrToPtr, "i");
+  auto d_nonMemFnPtrToPtrToPtr_1Par = clad::gradient(***(&nonMemFnPtrToPtrToPtr), "i");
+  auto d_nonMemFnPtrToPtrToPtr_1ParPar = clad::gradient(*(**(&nonMemFnPtrToPtrToPtr)), "i");
+  auto d_nonMemFnPtrToPtrToPtr_1ParParPar = clad::gradient((*(**((&nonMemFnPtrToPtrToPtr)))), "i");
+  auto d_nonMemFnIndirectPtr = clad::gradient(nonMemFnIndirectPtr, "i");
+  auto d_nonMemFnIndirectIndirectPtr = clad::gradient(nonMemFnIndirectIndirectPtr, "i");
+  auto d_nonMemFnStaticCast = clad::gradient(static_cast<decltype(&nonMemFn)>(nonMemFn), "i");
+  auto d_nonMemFnReinterpretCast = clad::gradient(reinterpret_cast<decltype(&nonMemFn)>(nonMemFn), "i");
+  auto d_nonMemFnCStyleCast = clad::gradient((decltype(&nonMemFn))(nonMemFn), "i");
+
+
+  NON_MEM_FN_TEST(d_nonMemFn); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPar); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtr); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtr); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtrPar); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtr_1); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtr_1Par); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtr_1ParPar); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtrToPtr); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtrToPtr_1); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtrToPtr_1Par); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtrToPtr_1ParPar); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnPtrToPtrToPtr_1ParParPar); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnIndirectPtr); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnIndirectIndirectPtr); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnStaticCast); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnReinterpretCast); // CHECK-EXEC: 10.00
+
+  NON_MEM_FN_TEST(d_nonMemFnCStyleCast); // CHECK-EXEC: 10.00
+}
--- a/test/Hessian/Pointers.C
+++ b/test/Hessian/Pointers.C
@@ -1,0 +1,153 @@
+// RUN: %cladclang %s -I%S/../../include -oPointers.out 2>&1 | FileCheck %s
+// RUN: ./Pointers.out | FileCheck -check-prefix=CHECK-EXEC %s
+// CHECK-NOT: {{.*error|warning|note:.*}}
+
+#include "clad/Differentiator/Differentiator.h"
+
+double nonMemFn(double i, double j) {
+  return i*j;
+}
+
+// CHECK: double nonMemFn_darg0(double i, double j) {
+// CHECK-NEXT:     double _d_i = 1;
+// CHECK-NEXT:     double _d_j = 0;
+// CHECK-NEXT:     return _d_i * j + i * _d_j;
+// CHECK-NEXT: }
+
+// CHECK: void nonMemFn_darg0_grad(double i, double j, double *_result) {
+// CHECK-NEXT:     double _d__d_i = 0;
+// CHECK-NEXT:     double _d__d_j = 0;
+// CHECK-NEXT:     double _t0;
+// CHECK-NEXT:     double _t1;
+// CHECK-NEXT:     double _t2;
+// CHECK-NEXT:     double _t3;
+// CHECK-NEXT:     double _d_i = 1;
+// CHECK-NEXT:     double _d_j = 0;
+// CHECK-NEXT:     _t1 = _d_i;
+// CHECK-NEXT:     _t0 = j;
+// CHECK-NEXT:     _t3 = i;
+// CHECK-NEXT:     _t2 = _d_j;
+// CHECK-NEXT:     double nonMemFn_darg0_return = _t1 * _t0 + _t3 * _t2;
+// CHECK-NEXT:     goto _label0;
+// CHECK-NEXT:   _label0:
+// CHECK-NEXT:     {
+// CHECK-NEXT:         double _r0 = 1 * _t0;
+// CHECK-NEXT:         _d__d_i += _r0;
+// CHECK-NEXT:         double _r1 = _t1 * 1;
+// CHECK-NEXT:         _result[1UL] += _r1;
+// CHECK-NEXT:         double _r2 = 1 * _t2;
+// CHECK-NEXT:         _result[0UL] += _r2;
+// CHECK-NEXT:         double _r3 = _t3 * 1;
+// CHECK-NEXT:         _d__d_j += _r3;
+// CHECK-NEXT:     }
+// CHECK-NEXT: }
+
+// CHECK: double nonMemFn_darg1(double i, double j) {
+// CHECK-NEXT:     double _d_i = 0;
+// CHECK-NEXT:     double _d_j = 1;
+// CHECK-NEXT:     return _d_i * j + i * _d_j;
+// CHECK-NEXT: }
+
+// CHECK: void nonMemFn_darg1_grad(double i, double j, double *_result) {
+// CHECK-NEXT:     double _d__d_i = 0;
+// CHECK-NEXT:     double _d__d_j = 0;
+// CHECK-NEXT:     double _t0;
+// CHECK-NEXT:     double _t1;
+// CHECK-NEXT:     double _t2;
+// CHECK-NEXT:     double _t3;
+// CHECK-NEXT:     double _d_i = 0;
+// CHECK-NEXT:     double _d_j = 1;
+// CHECK-NEXT:     _t1 = _d_i;
+// CHECK-NEXT:     _t0 = j;
+// CHECK-NEXT:     _t3 = i;
+// CHECK-NEXT:     _t2 = _d_j;
+// CHECK-NEXT:     double nonMemFn_darg1_return = _t1 * _t0 + _t3 * _t2;
+// CHECK-NEXT:     goto _label0;
+// CHECK-NEXT:   _label0:
+// CHECK-NEXT:     {
+// CHECK-NEXT:         double _r0 = 1 * _t0;
+// CHECK-NEXT:         _d__d_i += _r0;
+// CHECK-NEXT:         double _r1 = _t1 * 1;
+// CHECK-NEXT:         _result[1UL] += _r1;
+// CHECK-NEXT:         double _r2 = 1 * _t2;
+// CHECK-NEXT:         _result[0UL] += _r2;
+// CHECK-NEXT:         double _r3 = _t3 * 1;
+// CHECK-NEXT:         _d__d_j += _r3;
+// CHECK-NEXT:     }
+// CHECK-NEXT: }
+
+// CHECK: void nonMemFn_hessian(double i, double j, double *hessianMatrix) {
+// CHECK-NEXT:     nonMemFn_darg0_grad(i, j, &hessianMatrix[0UL]);
+// CHECK-NEXT:     nonMemFn_darg1_grad(i, j, &hessianMatrix[2UL]);
+// CHECK-NEXT: }
+
+#define NON_MEM_FN_TEST(var)\
+res[0]=res[1]=res[2]=res[3]=0;\
+var.execute(3, 4, res);\
+printf("{%.2f %.2f %.2f %.2f}\n", res[0], res[1], res[2], res[3]);
+
+int main() {
+  auto nonMemFnPtr = &nonMemFn;
+  auto nonMemFnPtrToPtr = &nonMemFnPtr;
+  auto nonMemFnPtrToPtrToPtr = &nonMemFnPtrToPtr;
+  auto nonMemFnIndirectPtr = nonMemFnPtr;
+  auto nonMemFnIndirectIndirectPtr = nonMemFnIndirectPtr;
+
+  double res[4];
+
+  auto d_nonMemFn = clad::hessian(nonMemFn);
+  auto d_nonMemFnPar = clad::hessian((nonMemFn));
+  auto d_nonMemFnPtr = clad::hessian(nonMemFnPtr);
+  auto d_nonMemFnPtrToPtr = clad::hessian(*nonMemFnPtrToPtr);
+  auto d_nonMemFnPtrToPtrPar = clad::hessian((*(nonMemFnPtrToPtr)));
+  auto d_nonMemFnPtrToPtr_1 = clad::hessian(**&nonMemFnPtrToPtr);
+  auto d_nonMemFnPtrToPtr_1Par = clad::hessian(**(&nonMemFnPtrToPtr));
+  auto d_nonMemFnPtrToPtr_1ParPar = clad::hessian(*(*(&nonMemFnPtrToPtr)));
+  auto d_nonMemFnPtrToPtrToPtr = clad::hessian(**nonMemFnPtrToPtrToPtr);
+  auto d_nonMemFnPtrToPtrToPtr_1 = clad::hessian(***&nonMemFnPtrToPtrToPtr);
+  auto d_nonMemFnPtrToPtrToPtr_1Par = clad::hessian(***(&nonMemFnPtrToPtrToPtr));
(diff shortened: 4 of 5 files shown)


parth-07:
@vgvassilev 
Please consider this code example, 
```c++
Expr* expr;  // expr have some valid value
DeclRefExpr* ref = nullptr;
if(auto ICE = dyn_cast<ImplicitCastExpr>(expr)) {
  ref = dyn_cast<DeclRefExpr>(ICE->getSubExpr());
}
```
Assuming `dyn_cast` did not return nullptr. Is there any way to retrieve `expr` using `ref` ? This will help to reduce code in this commit.

parth-07:
@vgvassilev Please see why check on `Clad-CI/ ubuntu-gcc9-runtime7` is failing .

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -14,6 +14,51 @@ using namespace clang;
...
 
+  // returns pointer to parent of first DeclRefExpr descendant 
+  // of Expr pointed by E in AST.
+  // returns nullptr, if node E is itself of type DeclRefExpr
+  // returns nullptr, if node of type other than ImplicitCastExpr
+  // or UnaryOperator comes before DeclRefExpr
+  Expr* getParentOfFirstDreDescendant(Expr* E) {
+    Expr* last_E = nullptr;
+    while(1) {
+      if(auto ICE = dyn_cast<ImplicitCastExpr>(E)) {
```
vgvassilev, line 25: Can we use `Expr::IgnoreImpCasts`?

vgvassilev:
> @vgvassilev
> Please consider this code example,
> 
> ```c++
> Expr* expr;  // expr have some valid value
> DeclRefExpr* ref = nullptr;
> if(auto ICE = dyn_cast<ImplicitCastExpr>(expr)) {
>   ref = dyn_cast<DeclRefExpr>(ICE->getSubExpr());
> }
> ```
> 
> Assuming `dyn_cast` did not return nullptr. Is there any way to retrieve `expr` using `ref` ? This will help to reduce code in this commit.

Can you elaborate? Do you mean to get a `DeclRefExpr` from the `ICE->getSubExpr()`?

lib/Differentiator/DiffPlanner.cpp
parth-07: No, it do not skip pasts `UnaryOperator`expressions.

parth-07:
> > @vgvassilev
> > Please consider this code example,
> > ```c++
> > Expr* expr;  // expr have some valid value
> > DeclRefExpr* ref = nullptr;
> > if(auto ICE = dyn_cast<ImplicitCastExpr>(expr)) {
> >   ref = dyn_cast<DeclRefExpr>(ICE->getSubExpr());
> > }
> > ```
> > 
> > 
> > Assuming `dyn_cast` did not return nullptr. Is there any way to retrieve `expr` using `ref` ? This will help to reduce code in this commit.
> 
> Can you elaborate? Do you mean to get a `DeclRefExpr` from the `ICE->getSubExpr()`?

I meant, to get original `expr` from `ref`.  I now know, that we can use `ASTContext.getParent` for it, I will change code to use it.

parth-07:
`ASTContext::getParents()` is incorrectly returning empty parents list in some cases, mutating the AST might be responsible for this( relevent links:  [ASTContext::getParents returning empty list](http://clang-developers.42468.n3.nabble.com/ASTContext-getParents-returning-empty-list-td4062620.html) , [ASTContext::getParents discussion](http://clang-developers.42468.n3.nabble.com/ASTContext-getParents-discussion-td4064656.html) ) . Is there some other way to get parent of `DeclRefExpr` node ?

lib/Differentiator/DiffPlanner.cpp
parth-07: Now I am using, `Expr::IgnoreParenImpCasts`, it skips past over both implicit casts and parantheses. This helps to add support for using parantheses too. Now I am just handling `UnaryOperator` expressions manually.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -14,6 +14,53 @@ using namespace clang;
...
+  // Returns address of parent of Expr node obtained 
+  // by skipping past any parantheses, implicit casts
+  // and UnaryOperators from Expr pointed by `E`.
+  // If considerParanAsParent is false,then returns last non-parantheses
+  // ancestor of Expr node obtained by skipping.
+  // Returns nullptr, if there is no skipping of nodes
+  // from given Expr
+  Expr* getParentOfSkipPastParImpCastUnOp(Expr* E,bool considerParanAsParent = true) {
+    Expr* last_E = nullptr;
+    while(1) {
```
vgvassilev, line 26:
```suggestion
    while (1) {
```

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -14,6 +14,53 @@ using namespace clang;
...
+      if(auto ICE = dyn_cast<ImplicitCastExpr>(E)) {
```
vgvassilev, line 27:
```suggestion
      if (auto ICE = dyn_cast<ImplicitCastExpr>(E)) {
```
And in the rest of the changes.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -14,6 +14,53 @@ using namespace clang;
...
+        last_E = E;
+        E = cast<Expr>(ICE->getSubExpr());
+      }
+      else if(auto pExp = dyn_cast<ParenExpr>(E)) {
+        if(considerParanAsParent)
+          last_E = E;
+        E = cast<Expr>(pExp->getSubExpr());
+      }
+      else if(auto UnOp = dyn_cast<UnaryOperator>(E)) {
```
vgvassilev, line 36: The `else if` goes on the previous line, and the trailing `else`, too.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -22,18 +69,33 @@ namespace clad {
...
     DeclRefExpr* oldDRE = nullptr;
-    // Handle the case of function pointer.
-    if (auto ICE = dyn_cast<ImplicitCastExpr>(call->getArg(0))){
-      oldDRE = dyn_cast<DeclRefExpr>(ICE->getSubExpr());
-    }
-    // Handle the case of member function.
-    else if (auto UnOp = dyn_cast<UnaryOperator>(call->getArg(0))){
-      oldDRE = dyn_cast<DeclRefExpr>(UnOp->getSubExpr());
+    // obtain parent of DeclRefExpr of function 
+    // to be differentiated
```
vgvassilev, line 73: That can go on a single line.

vgvassilev (commented):
We seem to lack coverage, please take a look where codecov hints and create a test case to cover that particular control flow.

lib/Differentiator/DiffPlanner.cpp
parth-07: I will take care of the spaces. 
Can we use some formatter to take care of formatting issues in github workflows ?

lib/Differentiator/DiffPlanner.cpp
parth-07: I am not sure if I understood this, do the else if goes on the previous line means it should be with the ending brace ( `}` ) of last `if` statement, as in,

```c++
} else if {
...
}
```

lib/Differentiator/DiffPlanner.cpp
parth-07: Okay :)

lib/Differentiator/DiffPlanner.cpp
vgvassilev: We can and we should, if you have experience with that I'd be happy to merge your PR (cc @reikdas).

parth-07:
@vgvassilev 

> We seem to lack coverage, please take a look where codecov hints and create a test case to cover that particular control flow.

```c++
      else if(auto pExp = dyn_cast<ParenExpr>(E)) {
        if(considerParanAsParent)
          last_E = E;
        E = cast<Expr>(pExp->getSubExpr());
      }
```
@vgvassilev This code under if condition isn't covered by the tests, because `considerParanAsParent` parameter will be `false` for all our use cases. But I feel uncomfortable removing it, because then function name will lose it's meaning. And if I change function implementation and it's name then the function will become too specific .

Here function is, `getParentOfSkipPastParImpCastUnOp(Expr* E)`, basically returns the parent of `Expr` node obtained by skipping `parenExpr`, `implicitCast` and `UnaryOperator`, this optional `considerParenAsParent` `bool` variable can slightly change the function meaning and it's return value, that is if this value is false, then nearest ancestor which is not of parenExpr type will be returned instead. Please guide me what should I do in this case to improve code coverage as well as not to make the function too specific.

parth-07:
> We seem to lack coverage, please take a look where codecov hints and create a test case to cover that particular control flow.

Please guide me how should I test coverage for code which should not be reached if everything goes well (function can be differentiated) ?
For example, how to test coverage for code like, 

```c++
else {
  llvm_unreachable("Trying to differentiate something unsupported");
}
```

vgvassilev:
> > We seem to lack coverage, please take a look where codecov hints and create a test case to cover that particular control flow.
> 
> ```c++
>       else if(auto pExp = dyn_cast<ParenExpr>(E)) {
>         if(considerParanAsParent)
>           last_E = E;
>         E = cast<Expr>(pExp->getSubExpr());
>       }
> ```
> 
> @vgvassilev This code under if condition isn't covered by the tests, because `considerParanAsParent` parameter will be `false` for all our use cases. But I feel uncomfortable removing it, because then function name will lose it's meaning. And if I change function implementation and it's name then the function will become too specific .

I am wondering if there is a more elegant way to solve this issue.

> 
> Here function is, `getParentOfSkipPastParImpCastUnOp(Expr* E)`, it basically returns the parent of `Expr` node obtained by skipping `parenExpr`, `implicitCast` and `UnaryOperator` from the `Expr` pointed by `E` , this optional `considerParenAsParent` `bool` variable can slightly change the function meaning and it's return value, that is if this value is false, then nearest ancestor which is not of `parenExpr` type will be returned instead. Please guide me what should I do in this case to improve code coverage as well as not to make the function too specific.

Alternatively, we can name the function `strip(Expr *E, StmtClass SC...)`. Then the usage would be:
```cpp
  Expr *E;
  Expr OldE = E;
  E = strip(E, {ParenExprClass,...})
```
https://godbolt.org/z/j4qvaf

I think we should use a `StmtVisitor` in the implementation of that function instead of a `while (1)`. Is all of that to strip the pointers and get to a 'clean' DeclRefExpr?

parth-07:
@vgvassilev Yes, `StmtVisitor` is much more flexible and convenient to solve this problem. I will use it instead of `while(1)` loop. 

> Is all of that to strip the pointers and get to a 'clean' DeclRefExpr?

In 2 places, I have also used `while(1)` loop to get to `DeclRefExpr` of a `FunctionDecl`, in these, loop also need to skip past the `DeclRefExpr` of  `VarDecl` using `ValueDecl::getInit()` method. 
And in `updateCall` function, we need nearest ancestor of `DeclRefExpr` of `FunctionDecl`  that isn't of type `ParenExpr` instead of `DeclRefExpr` itself. But this can also be easily done using `StmtVisitor`.

And in `strip(Expr* E, StmtClass SC...)` , where `StmtClass...` denotes types of expressions to skip, can I use `llvm::SmallSet` to store enums of StmtClasses to skip ? As I will need to frequenty check, if the current expression is to be skipped or not and `SmallSet` will give better search complexity.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -14,52 +17,126 @@ using namespace clang;
...
-  // ancestor of Expr node obtained by skipping.
-  // Returns nullptr, if there is no skipping of nodes
-  // from given Expr
-  Expr* getParentOfSkipPastParImpCastUnOp(Expr* E,bool considerParanAsParent = true) {
-    Expr* last_E = nullptr;
-    while(1) {
-      if(auto ICE = dyn_cast<ImplicitCastExpr>(E)) {
-        last_E = E;
-        E = cast<Expr>(ICE->getSubExpr());
+  // Visitor class to conveniently obtain DeclRefExpr descendants
```
vgvassilev, line 20: I would expect this to become something like:
```cpp
DeclRefExpr* findFirstDeclRefExpr(Stmt* S) {
  struct Finder : public StmtVisitor<Finder> {
     DeclRefExpr* m_FoundDRE = nullptr;
     void VisitDeclRefExpr(const DeclRefExpr* DRE) {
         // FIXME: Figure out how to abort visitation when found
         if (!m_FoundDRE)
           m_FoundDRE = DRE;
     }
  } F;
  return F.Visit(S).m_Found;
}```

File: lib/Differentiator/DiffPlanner.cpp (hunk 3 above)
Comment (parth-07) [reply lib/Differentiator/DiffPlanner.cpp]: I don't have much experience with this yet, I will try to do it, if I succeed I will send the PR :)

File: lib/Differentiator/DiffPlanner.cpp (hunk 6 above)
Comment (parth-07) [reply lib/Differentiator/DiffPlanner.cpp:20]: In this case, I would like to keep `StmtVisitor` to use, outside the functions, since it will help to reduce code duplication.
Two functions will need this same `StmtVisitor`, `findFirstDeclRefExpr` and `findRelevantAncestorOfFirstDeclRef` (`Relevant` because we are only interested in the nearest ancestor having some particular values for`StmtType` )

Comment (vgvassilev) [reply lib/Differentiator/DiffPlanner.cpp:20]: You could do that in a single pass and possibly changing the signature and name of `findFirstDeclRefExpr`. You will need to add another overload `void VisitStmt(const Stmt *S)` where you can check for its `StmtClass` and store the parent expression.

Comment (parth-07) [reply lib/Differentiator/DiffPlanner.cpp:20]: And also, I think `findFirstDeclRefExpr` should be general to use, so we should allow to pass some list of `StmtClass` to specify which statements can be skipped while finding `DeclRefExpr`. 

```c++
DeclRefExpr* findFirstDeclRefExpr(Stmt* S, const ArrayRef<StmtClass>& stmtTypesToSkip) {
    DeclRefVisitor dreVisitor(stmtTypesToSkip,{});
    DeclRefExpr* DRE = nullptr; 
    while (DRE = dyn_cast_or_null<DeclRefExpr>(dreVisitor.skipToDeclRef(E))) {
      if (auto VD = dyn_cast<VarDecl>(DRE->getDecl())) {
        E = VD->getInit();
      } else {
        break;
      }
    }
    return DRE;
  }
```

Is this implementation good for `findFirstDeclRefExpr` ?

parth-07: Yes, I can do both of these work in single pass, but I still think there should be 2 separate functions for these. Because in some situations we will just need `DeclRefExpr` and in some we will need it's ancestor. These functions will be easier to use, if they do just one task.

parth-07: If I do both the tasks of finding `DeclRefExpr` and it's relevant ancestor and try to keep function generic as well, then function signature will become something like this,

```c++
std::pair<DeclRefExpr*, Expr*> findFirstFnDeclAndRelAncestor(llvm::ArrayRef<Stmt::StmtClass> stmtsTypesToSkip,
                                                            llvm::ArrayRef<Stmt::StmtClass> RelevantAncTypes);
```
or,

```c++
Expr* findFirstFnDecl(llvm::ArrayRef<Stmt::StmtClass> stmtsTypesToSkip,
                    llvm::ArrayRef<Stmt::StmtClass> RelevantAncTypes={},
                    Expr* ancestor = nullptr);
```

These function do not seem very friendly to use, or is there some other different and much better way to represent a function which can do both of these tasks.

vgvassilev: We should not overengineer the functions, let's start with something simple and see what else we are missing. The implementation of `findFirstFnDeclRefExpr` should not have loops inside -- that's taken care of the visitor. Take a look at the limited documentation about the [RecursiveASTVisitor](https://clang.llvm.org/doxygen/classclang_1_1RecursiveASTVisitor.html).

parth-07: Yes, I think I was overengineering the functions a bit too much, I have made functions much more specific now. Please review the latest commit and suggest changes,

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -13,6 +14,52 @@ using namespace clang;
...
+  // argument's DeclRefExpr
+  // Here relevant ancestor is nearest ancestor of ImplicitCastExpr or 
+  // UnaryOperator type.
+  DeclRefExpr* getArgFunction(CallExpr* call,Expr** relevantAncestor = nullptr) {
+    struct Finder :
+      ConstStmtVisitor<Finder> {
+        const DeclRefExpr* m_FnDRE = nullptr;
+        Expr** m_RelevantAncestor;
+        Finder(Expr** relevantAncestor) : m_RelevantAncestor(relevantAncestor) {}
+        void Visit(const Stmt* stmt) {
```
vgvassilev, line 30: All of the logic can probably go in a `void VisitExpr` which will be called for every expression.

parth-07: `void VisitExpr` will only be called for every expression if I don't provide more specialized functions ( `void ImplicitCastExpr`, `void DeclRefExpr` etc). In `constStmtVisitor` only the most specialized `Visit*` function is called. 

And I am reluctant to place logic for `ParenExpr`, `ImplicitCastExpr` and `UnaryOperator` in the common `void VisitExpr` function because it will make it more difficult to add logic for different types of expressions later on. And also then the code of `VisitExpr` will be very dependent on unchecked `dyn_cast` if we do not want to add lots of `if` conditions.

parth-07: I have placed most of the common logic in `void Visit` function, If I move common logic to `void VisitExpr` then I will need to either:
1) Manually call `Visit(E)` from each of `void VisitImplicitCastExpr`, `void VisitUnaryOperator` and `void ParenExpr`.
2) Place logic of these three functions in `void VisitExpr` using `if` conditions with `dyn_cast`.

I thought it be easier to maintain logic of expressions using `Visit*` functions, and to place common logic in `Visit` function, should I change it ?

vgvassilev: > `void VisitExpr` will only be called for every expression if I don't provide more specialized functions ( `void ImplicitCastExpr`, `void DeclRefExpr` etc). In `constStmtVisitor` only the most specialized `Visit*` function is called.

Correct, I do not think adding specialized functions for that particular use case buys us a lot.

> 
> And I am reluctant to place logic for `ParenExpr`, `ImplicitCastExpr` and `UnaryOperator` in the common `void VisitExpr` function because it will make it more difficult to add logic for different types of expressions later on. 

What other logic you are envisioning? Remember -- that is a super specific function for a very specific use case.

> And also then the code of `VisitExpr` will be very dependent on unchecked `dyn_cast` if we do not want to add lots of `if` conditions.

Hm... not sure I understand -- we just need to compare the stmt class and decide to store either in the `DeclRefExpr` field or the other field for the non-implicitcast, non-unaryop, non-parenexpr parent expression.

parth-07: > > `void VisitExpr` will only be called for every expression if I don't provide more specialized functions ( `void ImplicitCastExpr`, `void DeclRefExpr` etc). In `constStmtVisitor` only the most specialized `Visit*` function is called.
> 
> Correct, I do not think adding specialized functions for that particular use case buys us a lot.
>
> > And I am reluctant to place logic for `ParenExpr`, `ImplicitCastExpr` and `UnaryOperator` in the common `void VisitExpr` function because it will make it more difficult to add logic for different types of expressions later on.
> 
> What other logic you are envisioning? Remember -- that is a super specific function for a very specific use case.


By other logic I meant considering more expressions, at the moment we are only considering `ImplicitCast`, `UnaryOperator` and `Parantheses`, in the future we may need to consider lot more types of expressions as well(`static_cast`, `subscript` operator etc) . So, I was thinking for separate function for each type of expression, but yes, single function with `if` and `dyn_cast` will be good too. 
 
> > And also then the code of `VisitExpr` will be very dependent on unchecked `dyn_cast` if we do not want to add lots of `if` conditions.
> 
> Hm... not sure I understand -- we just need to compare the stmt class and decide to store either in the `DeclRefExpr` field or the other field for the non-implicitcast, non-unaryop, non-parenexpr parent expression.


Sorry, I wasn't very clear here. Here I was thinking using `VisitExpr` for `ImplicitCastExpr`, `ParenExpr` and `UnaryOperator` and `VisitDeclRefExpr` for `DeclRefExpr`. Then by unchecked `dyn_cast` I meant code like,
 
`dyn_cast<decltype(E)>(E)->getSubExpr())`

this way we will not need any separate `if` conditions for different types of Expressions, but yes, this code is very dangerous and wrong, so using `if` conditions is a lot better.

vgvassilev: > > > `void VisitExpr` will only be called for every expression if I don't provide more specialized functions ( `void ImplicitCastExpr`, `void DeclRefExpr` etc). In `constStmtVisitor` only the most specialized `Visit*` function is called.
> > 
> > 
> > Correct, I do not think adding specialized functions for that particular use case buys us a lot.
> > > And I am reluctant to place logic for `ParenExpr`, `ImplicitCastExpr` and `UnaryOperator` in the common `void VisitExpr` function because it will make it more difficult to add logic for different types of expressions later on.
> > 
> > 
> > What other logic you are envisioning? Remember -- that is a super specific function for a very specific use case.
> 
> By other logic I meant considering more types of expressions, at the moment we are only considering `ImplicitCast`, `UnaryOperator` and `Parantheses`, in the future we may need to consider lot more types of expressions as well(`static_cast`, `subscript` operator etc) . So, I was thinking for separate function for each type of expression, but yes, single function with `if` and `dyn_cast` will be good too.

That's the point I am missing why would you need that `dyn_cast` -- that is done by the visitor itself, you do not need to do `Visit(X->getY)`.

> 
> > > And also then the code of `VisitExpr` will be very dependent on unchecked `dyn_cast` if we do not want to add lots of `if` conditions.
> > 
> > 
> > Hm... not sure I understand -- we just need to compare the stmt class and decide to store either in the `DeclRefExpr` field or the other field for the non-implicitcast, non-unaryop, non-parenexpr parent expression.
> 
> Sorry, I wasn't very clear here. Here I was thinking using `VisitExpr` for `ImplicitCastExpr`, `ParenExpr` and `UnaryOperator` and `VisitDeclRefExpr` for `DeclRefExpr`. Then by unchecked `dyn_cast` I meant code like,
> 
> `dyn_cast<decltype(E)>(E)->getSubExpr())`

Why do you need this? The visitor does it for you.

parth-07: I was under the impression in `StmtVisitor` we manually have to call `Visit` function for each `stmt` that we have to visit, it's in `RecursiveASTVisitors` where visitor automatically visits every `stmts` (along with expressions and declarations)

vgvassilev: Sure, we can use the `RecursiveASTVisitor`. The generic visit routine for a `StmtVisitor` looks like [this](https://github.com/llvm/llvm-project/blob/20a24af01de2eeb3d1f88fa22cee7201bf13d608/clang/include/clang/AST/EvaluatedExprVisitor.h#L100-L104). In general, you can see how such things are used in the clang codebase.

parth-07: yeah, I did see `StmtVisitor` example in `lib/AST/StmtPrinter.cpp` file, it do seems there we manually need to call `Visit` for `stmt` that we want it to visit. 
I did consider using `RecursiveASTVisitor`, but it have it's own issues, first we do not just need to traverse down the given AST Node, as if we encounter any `VarDecl` then we will need to traverse down it's value (using `VarDecl::getInit()`), so we will need to call `TraverseDecl(VD->getInit())` to get to the DeclRefExpr of the function.

parth-07: And I was under the impression that `StmtVisitor` is more efficient than `RecursiveASTVisitor` as it just visit statements, so I preferred using `StmtVisitor`.

parth-07: Is `RecursiveASTVisitor` better fit for this task ? Should I change code to use `RecursiveASTVisitor` instead of `ConstStmtVisitor` ? 
And is there any significant difference in efficiency between`StmtVisitor` and `RecursiveASTVisitor` that I should be aware of while choosing between them ?

parth-07: Using `RecursiveASTVisitor`, we will not need to manually take `E->getSubExpr()`  exactly as you were saying. And, we also won't need to consider `ParenExpr`, 'ImplicitCastExpr` etc exactly as you were saying. 
Should I make changes in the code to use `RecursiveASTVisitor` instead of `ConstStmtVisitor` ? 

With `RecursiveASTVisitor`, `VisitExpr` function will be like this:
```c++
 bool VisitExpr(Expr* E) {
          if (m_RelevantAncestor) {
            if (E->getStmtClass() == Stmt::StmtClass::ImplicitCastExprClass ||
                E->getStmtClass() == Stmt::StmtClass::UnaryOperatorClass) {
              *m_RelevantAncestor = const_cast<Expr*>(E);
            }
          }
          if (auto DRE = dyn_cast<DeclRefExpr>(E)) {
            if (auto VD = dyn_cast<VarDecl>(DRE->getDecl())) {
              TraverseStmt(VD->getInit());
            } else if (auto FD = dyn_cast<FunctionDecl>(DRE->getDecl())) {
              m_FnDRE = DRE;
            }
            return false;  
          }
          return true; 
        }
      }
```

parth-07:
@vgvassilev Please review the PR again and suggest any required changes,

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -13,6 +14,43 @@ using namespace clang;
...
+  // If argument is passed for relevantAncestor, then it's value will be 
+  // modified in-place to contain relevant ancestor of the function 
+  // argument's DeclRefExpr
+  // Here relevant ancestor is nearest ancestor of ImplicitCastExpr or 
+  // UnaryOperator type.
+  DeclRefExpr* getArgFunction(CallExpr* call,Expr** relevantAncestor = nullptr) {
+    struct Finder :
+      RecursiveASTVisitor<Finder> {
+        DeclRefExpr* m_FnDRE = nullptr;
+        Expr** m_RelevantAncestor;
```
vgvassilev, line 28: Why not `Expr* m_RelevantAncestor = nullptr;`

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -13,6 +14,43 @@ using namespace clang;
...
+        Finder(Expr** relevantAncestor) : m_RelevantAncestor(relevantAncestor) {}
+        bool VisitExpr(Expr* E) {
+          if (m_RelevantAncestor) {
+            if (E->getStmtClass() == Stmt::StmtClass::ImplicitCastExprClass ||
```
vgvassilev, line 32: I would use a switch statement here.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -13,6 +14,43 @@ using namespace clang;
...
+            }
+            return false;  
+          }
+          return true; 
+        }
+      } finder(relevantAncestor);
+      if (call->getNumArgs() < 1) {
+        return nullptr;
+      }
+      finder.TraverseStmt(call->getArg(0));
```
vgvassilev, line 51: This part is unclear to me -- why only the first argument, could you add a comment? I presume it is because it comes from `clad::differentiate` then, I'd drop `if (call->getNumArgs() < 1)` and add an assert to catch calls outside of the context of clad:: functions.

test/ForwardMode/Pointers.C
```diff
@@ -0,0 +1,455 @@
...
+  }
+  // CHECK: double memFn_darg0(double i, double j) {
+  // CHECK-NEXT:     double _d_i = 1;
+  // CHECK-NEXT:     double _d_j = 0;
+  // CHECK-NEXT:     double _t0 = i * i;
+  // CHECK-NEXT:     double _t1 = _t0 * j;
+  // CHECK-NEXT:     double &_t2 = this->x;
+  // CHECK-NEXT:     double _t3 = j * j;
+  // CHECK-NEXT:     double &_t4 = this->y;
+  // CHECK-NEXT:     return ((_d_i * i + i * _d_i) * j + _t0 * _d_j) * _t2 + _t1 * 0. + (_d_j * j + j * _d_j) * _t4 + _t3 * 0.;
```
vgvassilev, line 24: I feel all those check clauses are redundant as this test is not really testing the output but rather the way we take the function pointers. I'd suggest make the function body really simple -- like `return 1;`.

parth-07: Yes, agreed. Will make this change.

lib/Differentiator/DiffPlanner.cpp
parth-07: Because only constructor of Finder takes 1 argument `relevantAncestor` ( function parameter) to initialize `m_RelevantAncestor` and `relevantAncestor` parameter have default value of `nullptr`. Therefore, if value of `relevantAncestor` is not specified, then value of both `relevantAncestor` and `m_RelevantAncestor` will be `nullptr`

lib/Differentiator/DiffPlanner.cpp
parth-07: Yes, switch statement seems better here. Will make the changes.

lib/Differentiator/DiffPlanner.cpp
parth-07: It's because all the differentiation calls will have _atleast_ 1 argument ( for example `clad::gradient` can accept 1 argument) and in all the differentiation calls first argument is always the function to be differentiated. I will add comment for this. 
I will add `assert` statement here to make sure that this function is called from clad:: functions only .

lib/Differentiator/DiffPlanner.cpp
vgvassilev: The computational cost to set it either way is close to 0.

parth-07: Yes, now I have added default initialization value of `nullptr` to `Expr** m_RelevantAncestor` as well.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -13,6 +14,48 @@ using namespace clang;
...
 namespace clad {
   static SourceLocation noLoc;
+  
+  // Returns DeclRefExpr of function argument of differentiation calls.
+  // If argument is passed for relevantAncestor, then it's value will be 
+  // modified in-place to contain relevant ancestor of the function 
+  // argument's DeclRefExpr
+  // Here relevant ancestor is nearest ancestor of ImplicitCastExpr or 
+  // UnaryOperator type.
+  DeclRefExpr* getArgFunction(CallExpr* call,Expr** relevantAncestor = nullptr) {
```
vgvassilev, line 24:
```suggestion
  DeclRefExpr* getArgFunction(CallExpr* call,
                                                    Expr** relevantAncestor = nullptr) {
```

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -13,6 +14,48 @@ using namespace clang;
...
+    struct Finder :
+      RecursiveASTVisitor<Finder> {
+        DeclRefExpr* m_FnDRE = nullptr;
+        Expr** m_RelevantAncestor = nullptr;
+        Finder(Expr** relevantAncestor) : m_RelevantAncestor(relevantAncestor) {}
+        bool VisitExpr(Expr* E) {
+          if (m_RelevantAncestor) {
+            switch (E->getStmtClass()) {
+              case Stmt::StmtClass::ImplicitCastExprClass:
```
vgvassilev, line 33:
```suggestion
              case Stmt::StmtClass::ImplicitCastExprClass: LLVM_FALLTHROUGH;
```

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -13,6 +14,48 @@ using namespace clang;
...
+              case Stmt::StmtClass::UnaryOperatorClass:
+                *m_RelevantAncestor = E;
+                break; 
+            }
```
vgvassilev, line 37: Isn't that the inverse? You want to se the `m_RelevantAncestor` when it is not `UnaryOperatorClass` and `ImplicitCastExprClass` -- in that case you need to move the setter in the `default`.

If that's the case, that tells us that we do not have test coverage.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -13,6 +14,48 @@ using namespace clang;
...
+          }
+          if (auto DRE = dyn_cast<DeclRefExpr>(E)) {
+            if (auto VD = dyn_cast<VarDecl>(DRE->getDecl())) {
```
vgvassilev, line 40: No `{`, `}` for single stmt if-s.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -13,6 +14,48 @@ using namespace clang;
...
+              TraverseStmt(VD->getInit());
+            } else if (auto FD = dyn_cast<FunctionDecl>(DRE->getDecl())) {
+              m_FnDRE = DRE;
+            }
+            return false;
```
vgvassilev, line 45: Returning `false` abort the visitation -- I'd think we want to do that only when setting the `m_FnDRE`.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -13,6 +14,48 @@ using namespace clang;
...
+            return false;  
+          }
+          return true; 
+        }
+      } finder(relevantAncestor);
+
+      auto callFn = call->getDirectCallee();
+      auto cladNS = dyn_cast<NamespaceDecl>(callFn->getDeclContext());
+      assert(cladNS->getName() == "clad" &&
+        "Only clad:: function call expression can be passed as call argument");
```
vgvassilev, line 54:
```suggestion
      assert(cast<NamespaceDecl>(call->getDirectCallee())->
             getDeclContext())->getName() == "clad" &&
             "Should be called for clad:: special functions!");
```

asserts are only expanded in debug builds and `cladNS` will become a unused variable in release.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -21,16 +64,10 @@ namespace clad {
...
-    if (auto ICE = dyn_cast<ImplicitCastExpr>(call->getArg(0))){
-      oldDRE = dyn_cast<DeclRefExpr>(ICE->getSubExpr());
-    }
-    // Handle the case of member function.
-    else if (auto UnOp = dyn_cast<UnaryOperator>(call->getArg(0))){
-      oldDRE = dyn_cast<DeclRefExpr>(UnOp->getSubExpr());
-    }
-    else
+    Expr* oldArgDREParent = nullptr;
+    DeclRefExpr* oldDRE = getArgFunction(call,&oldArgDREParent);
```
vgvassilev, line 68:
```suggestion
    DeclRefExpr* oldDRE = getArgFunction(call, &oldArgDREParent);
```

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -42,8 +79,7 @@ namespace clad {
     // using call->setArg(0, DRE) seems to be sufficient,
     // though the real AST allways contains the ImplicitCastExpr (function ->
     // function ptr cast) or UnaryOp (method ptr call).
-    auto oldArg = call->getArg(0);
-    if (auto oldCast = dyn_cast<ImplicitCastExpr>(oldArg)) {
+    if (auto oldCast = dyn_cast<ImplicitCastExpr>(oldArgDREParent)) {
```
vgvassilev, line 82: Isn't that unreachable now? Can `oldArgDREParent` be an `ImplicitCastExpr`?

lib/Differentiator/DiffPlanner.cpp
parth-07: No, we need the nearest Expression ancestor of `FunctionDecl` which is of type `ImplicitCastExpr` and `UnaryOperator`. As they will provide us information on how to construct `DeclRefExpr` for derived function in `DiffRequest::updateCall` function.

lib/Differentiator/DiffPlanner.cpp
parth-07: Okay, I will make this change and I will also read about `LLVM_FALLTHROUGH`:)

lib/Differentiator/DiffPlanner.cpp
parth-07: No, here we want to return `false` on finding `DeclRefExpr`. Here I have considered 2 cases, 
1) `DeclRefExpr->getDecl` is of type `FunctionDecl`, then `m_FnDRE` will be set to correct value and visitation will abort.
2) `DeclRefExpr->getDecl` is of type `VarDecl`, then considering, `VarDecl` is pointer to the function or pointer to pointer to the function (and so on), we will need to traverse the value of this variable, thus I have used the call `TraverseStmt(VD->getInit())`. I have returned `false` in this case too, since we do not need to traverse down the `VarDecl`, we only need to traverse down it's value.

lib/Differentiator/DiffPlanner.cpp
parth-07: Oh okay, I will make this change, and I will remember this for future PRs :)

lib/Differentiator/DiffPlanner.cpp
parth-07: Yes, `oldArgDREParent` can be`ImplicitCastExpr`,  actually, as per `getArgFunction` implementation, `oldArgDREParent` can only be of type `ImplicitCastExpr` and `UnaryOperator`.

parth-07:
@vgvassilev Should I add tests of using casts on function argument in clad:: functions ? For example, calls like these, 
```c++
auto d_nonMemFnStaticCast = clad::differentiate(static_cast<decltype(&nonMemFn)>(nonMemFn),"i");
auto d_memFnCStyleCast = clad::differentiate((decltype(&Expression::memFn))(&Expression::memFn),"i");
```

And should I follow max 80 charactes per line convention for test files too ?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -13,6 +14,47 @@ using namespace clang;
...
   static SourceLocation noLoc;
+  
+  // Returns DeclRefExpr of function argument of differentiation calls.
+  // If argument is passed for relevantAncestor, then it's value will be 
+  // modified in-place to contain relevant ancestor of the function 
+  // argument's DeclRefExpr
+  // Here relevant ancestor is nearest ancestor of ImplicitCastExpr or 
+  // UnaryOperator type.
+  DeclRefExpr* getArgFunction(CallExpr* call,
+    Expr** relevantAncestor = nullptr) {
```
vgvassilev, line 25: Please indent this below the first parameter.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -13,6 +14,47 @@ using namespace clang;
...
+              m_FnDRE = DRE;
+            return false;  
+          }
+          return true; 
+        }
+      } finder(relevantAncestor);
+
+      assert(cast<NamespaceDecl>(call->getDirectCallee()->
+        getDeclContext())->getName() == "clad" &&
+        "Should be called for clad:: special functions!");
```
vgvassilev, line 53: Likewise.

lib/Differentiator/DiffPlanner.cpp
vgvassilev: Ah, okay, sounds good.

lib/Differentiator/DiffPlanner.cpp
vgvassilev: Ok, sounds reasonable to me.

vgvassilev:
> @vgvassilev Should I add tests of using casts on function argument in clad:: functions ? For example, calls like these,
> 
> ```c++
> auto d_nonMemFnStaticCast = clad::differentiate(static_cast<decltype(&nonMemFn)>(nonMemFn),"i");
> auto d_memFnCStyleCast = clad::differentiate((decltype(&Expression::memFn))(&Expression::memFn),"i");
> ```

If that would increase testing coverage, yes.

> 
> And should I follow max 80 charactes per line convention for test files too ?

Yes, where possible.

test/ForwardMode/Pointers.C
```diff
@@ -0,0 +1,331 @@
...
+// RUN: ./Pointers.out | FileCheck -check-prefix=CHECK-EXEC %s
+// CHECK-NOT: {{.*error|warning|note:.*}}
+
+#include "clad/Differentiator/Differentiator.h"
+
+class Expression {
+public:
+  Expression() {}
+  double memFn(double i) {
+    return 1;
```
vgvassilev, line 11: Let's return `i` here and then CHECK if the return result is 1. This would be make the execution result more reliable as in we will know clad differentiated something.

test/ForwardMode/Pointers.C
```diff
@@ -0,0 +1,331 @@
...
+
+  // CHECK: double memFn_darg0(double i) {
+  // CHECK-NEXT:     double _d_i = 1;
+  // CHECK-NEXT:     return 0;
+  // CHECK-NEXT: }
+
+  // CHECK: double memFn_darg0(double i) {
+  // CHECK-NEXT:     double _d_i = 1;
+  // CHECK-NEXT:     return 0;
+  // CHECK-NEXT: }
```
vgvassilev, line 102: This is a recurring pattern which I'd like to remove. We have two options:
  * Keep only one such block -- The next use of `CHECK` (not `CHECK-NEXT`) will filter those out;
  * Fix the actual bug in printing the derivatives -- Currently each new produced derivative function declaration gets blindly printed. We can store a map of what's printed already and not reprint. Could you check if we produce N different `memFn_darg0`, if so that'd be wrong.

test/ForwardMode/Pointers.C
```diff
@@ -0,0 +1,331 @@
...
+  auto d_memFnPtrToPtrToPtrParPar = clad::differentiate(((*(*memFnPtrToPtrToPtr))), "i");
+  auto d_memFnPtrToPtrToPtr_1 = clad::differentiate(***&memFnPtrToPtrToPtr, "i");
+  auto d_memFnPtrToPtrToPtr_1Par = clad::differentiate(***(&memFnPtrToPtrToPtr), "i");
+  auto d_memFnPtrToPtrToPtr_1ParPar = clad::differentiate(**(*(&memFnPtrToPtrToPtr)), "i");
+  auto d_memFnPtrToPtrToPtr_1ParParPar = clad::differentiate(*(*(*(&memFnPtrToPtrToPtr))), "i");
+  auto d_memFnIndirectPtr = clad::differentiate(memFnIndirectPtr, "i");
+  auto d_memFnIndirectIndirectPtr = clad::differentiate(memFnIndirectIndirectPtr, "i");
+  auto d_memFnStaticCast = clad::differentiate(static_cast<decltype(&Expression::memFn)>(&Expression::memFn), "i");
+  auto d_memFnReinterpretCast = clad::differentiate(reinterpret_cast<decltype(&Expression::memFn)>(&Expression::memFn), "i");
+  auto d_memFnCStyleCast = clad::differentiate((decltype(&Expression::memFn))(&Expression::memFn), "i");
```
vgvassilev, line 232: Is this block of checks supposed to work with clad::gradient/clad::hessian/clad::jacobian? If so let's add couple of test cases for them too.

vgvassilev (commented):
Looks like we are converging!

test/ForwardMode/Pointers.C
parth-07: Yes, this will be better, I will make this change.

test/ForwardMode/Pointers.C
parth-07: I am more inclined to fix the actual bug.
We are indeed differentiating `memFn_darg0` as many times as it is requested (there seems to be no check to see if we have already differentiated the said function before differentiating it).
I will start working on this today.

test/ForwardMode/Pointers.C
parth-07: Yes, it is supposed to work with `clad::gradient`, `clad::hessian` and `clad::jacobian` as well. I will add tests for them. 
I would like to add, I think we should not add tests for differentiating member functions using `clad::gradient` due to issue #139

parth-07:
@vgvassilev To stop producing same derived function multiple times,  we need to know the derived function name in `CladPlugin::ProcessDiffRequest`. Currently we are producing derived function name in `clad::ForwardModeVisitor::derive` and `clad::ReverseModeVisitor::derive`. 

I propose one common function which can be accessed in `CladPlugin::ProcessDiffRequest` to produce derived function name. We can use this derived function name to check if we have already produced this derived function, if yes, then we can choose to not produce it again. 

And I would also like to add `std::string derivedFnName` in `DiffRequest`. This way we will be able to pass derived function name produced in `CladPlugin::ProcessDiffRequest` to ` m_DerivativeBuilder->Derive`, with this we will not need to produce derived function name again in `clad::ForwardModeVisitor::derive` and `clad::ReverseModeVisitor::derive`, and also we would be able to  remove the code responsible for producing derived function name from these two functions.

test/ForwardMode/Pointers.C
vgvassilev: ping...

parth-07: Please reply to this comment https://github.com/vgvassilev/clad/pull/212#issuecomment-808967358 on this thread. 
I would like to fix the issue of same functions being differentiated and printed multiple times.

vgvassilev:
I would think having some map that stores mapping FD->`DeclWithContext` (possibly somehow extended with DiffMode if necessary) somewhere around [here](https://github.com/vgvassilev/clad/blob/5b844d017a6aa3e9e0c29e458ff05a7e46e765ef/lib/Differentiator/DerivativeBuilder.cpp#L110).

parth-07:
To use map to check if the function has already been differentiated, we will need map as either
1) Key as combination of `IdentifierInfo*` of function to be differentiated, `DeclContext*` of function to be differentiated, differentiation mode, differentiation order and differentiation argument information  -> `FunctionDecl*` of differentiated function as value
Or
2) Key as `IdentifierInfo*` of the differentiated function and `DeclContext` of differentiated function -> `FunctionDecl*` of differentiated function as value.

Here, option 2 seems better.

If we are checking if a function is already differentiated in `DerivativeBuilder::Derive`, will it be okay to make a function `getDerivedFnName(const DiffRequest& request)` as member function of `DerivativeBuilder` ? For this we will also need `parseDiffArgs` function in `DerivativeBuilder` too. 

Or we can make this check in `CladPlugin::ProcessDiffRequest` and avoid the call to `DerivativeBuilder::Derive` altogether. But then we will need functions for finding derived function name in `CladPlugin`.

vgvassilev:
We cannot use `IdentifierInfo*` as a key because they are not unique per function but per identifier. Consider:
```cpp
  double ident(double){}
  double ident(double, double){}
  int main() { int ident = 1; return ident; }
```

Here the `IdentifierInfo*` will be the same for all 3 different types of declarations.

parth-07:
I was hoping `IdentifierInfo*` combined with `DeclContext*` would be unique, since we do not support overloaded functions in clad.

parth-07:
@vgvassilev Can you please review the PR.

vgvassilev:
Apologies for the delay. Could you squash the commit history into several atomic commits?

parth-07:
> Could you squash the commit history into several atomic commits?

Yes, I have squashed all the commits into 2 commits.

vgvassilev:
The tests should come in the same commit as the feature.

parth-07:
> The tests should come in the same commit as the feature.

Okay, I have now squashed commit for tests with the feature commit.

vgvassilev:
Could you rebase this PR?

parth-07:
> Could you rebase this PR?

Done.

vgvassilev (approved):
LGTM! The failures are due to a recent regression in the CI setup.

PR #215, opened 2021-03-17
Reverse mode type fix

### LINKED ISSUE #214: Differentiation of call expressions in reverse mode missisng cast
Minimal reproducible example:

```cpp
float helper (float a) {
    return a;
}

double fn (float a) {
    double x = helper(a);
    return x;
}

clad::gradient(fn);
```
results in 

```
void llvm::StoreInst::AssertOK(): Assertion `getOperand(0)->getType() == cast<PointerType>(getOperand(1)->getType())->getElementType() && "Ptr must be a pointer to Val type!"' failed.
```

The dump for ```fn``` reveals the following code...

```cpp
void fn_grad(float a, double *_result) {
    float _t0;
    double _d_x = 0;
    _t0 = a;
    double x = helper(_t0);
    double fn_return = x;
    goto _label0;
  _label0:
    _d_x += 1;
    {
        float _grad0[1] = {};
        helper_grad(_t0, _grad0);
        float _r0 = _d_x * _grad0[0UL];  //Line of interest, this fails but it actually should not
        _result[0UL] += _r0;
    }
}

```

Here, at the line of interest, we are possibly missing an implicit cast. 

@sudo-panda, this issue might be the core reason behind failure of ```pow``` in reverse mode. Maybe you would want to look into this and if possible, perhaps split your current PR (#206) to accommodate changes for both. 

Another thing to note is, this might not be the only place implicit casts are missing. So we should carefully find out such places and make sure the implicit casts are built.

sudo-panda:
Fixes #214 and also assigns the correct type to the temporary variables

### DIFF
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -299,6 +299,19 @@ namespace clad {
     clang::LookupResult& GetCladTapeBack();
     /// Instantiate clad::tape<T> type.
     clang::QualType GetCladTapeOfType(clang::QualType T);
+
+    /// Assigns the Init expression to VD after performing the necessary
+    /// implicit conversion. This is required as clang doesn't add implicit
+    /// conversions while assigning values to variables which are initialized
+    /// after it is already declared.
+    void PerformImplicitConversionAndAssign(clang::VarDecl* VD,
+                                             clang::Expr* Init) {
+      // Implicitly convert Init into the type of VD
+      clang::Expr* ICE = m_Sema.PerformImplicitConversion(
+          Init, VD->getType(), clang::Sema::AA_Casting).get();
+      // Assign the resulting expression to the variable declaration
+      VD->setInit(ICE);
+    }
   };
   /// A class that represents the result of Visit of ForwardModeVisitor.
   /// Stmt() allows to access the original (cloned) Stmt and Stmt_dx() allows
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -987,7 +987,6 @@ namespace clad {
     return CladTapeResult{*this, PushExpr, PopExpr, TapeRef};
   }
 
-
   StmtDiff ForwardModeVisitor::VisitStmt(const Stmt* S) {
     diag(DiagnosticsEngine::Warning, S->getBeginLoc(),
          "attempted to differentiate unsupported statement, no changes applied");
@@ -2537,14 +2536,17 @@ namespace clad {
     // Save current index in the current block, to potentially put some statements
     // there later.
     std::size_t insertionPoint = getCurrentBlock(reverse).size();
+    // Store the type to reduce call overhead that would occur if used in the loop
+    auto CEType = CE->getType();
     for (const Expr* Arg : CE->arguments()) {
       // Create temporary variables corresponding to derivative of each argument,
-      // so that they can be reffered to when arguments is visited. Variables
+      // so that they can be referred to when arguments is visited. Variables
       // will be initialized later after arguments is visited. This is done to
-      // reduce cloning complexity and only clone once.
-      Expr* dArg = StoreAndRef(nullptr, Arg->getType(), reverse, "_r", /*force*/true);
+      // reduce cloning complexity and only clone once. The type is same as the
+      // call expression as it is the type used to declare the _gradX array
+      Expr* dArg = StoreAndRef(nullptr, CEType, reverse, "_r", /*force*/true);
       ArgResultDecls.push_back(cast<VarDecl>(cast<DeclRefExpr>(dArg)->getDecl()));
-      // Visit using unitialized reference.
+      // Visit using uninitialized reference.
       StmtDiff ArgDiff = Visit(Arg, dArg);
       // Save cloned arg in a "global" variable, so that it is accesible from the
       // reverse pass.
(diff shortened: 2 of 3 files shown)


include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -46,6 +46,45 @@ namespace custom_derivatives {
...
+    struct promote<double>
+    {
+      typedef double type;
+    };
+
+    template <>
+    struct promote<float>
+    {
+      typedef float type;
+    };
```
vgvassilev, line 76: Can that be folder to ```template <typename T = double> using type = T;```

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -80,18 +119,18 @@ namespace custom_derivatives {
   }
 #endif
 
-  template<typename T>
-  T pow_darg0(T x, T exponent) {
-    return exponent * pow(x, exponent-((T)1));
+  template<typename T1, typename T2>
+  typename promote_2<T1, T2>::type pow_darg0(T1 x, T2 exponent) {
```
vgvassilev, line 123: I suspect that could become: ```template <typename T1, typename T2>
  decltype(T1() + T2()) pow_darg1(T1 x, T2 exponent) {...}```

sudo-panda (commented):
@vgvassilev it should now be visible

include/clad/Differentiator/BuiltinDerivatives.h
grimmmyshini: > I don't think that will work because T1 and T2 need to be promoted first before choosing the wider of the two.

you may use the following,

```template <typename T1, typename T2> decltype(std::pow(T1(), T2())) pow_darg1(T1 x, T2 exponent) {...}```

This will make ```decltype``` look at all overloads of ```std::pow``` and select the appropriate one (without evaluating that expression), saving you a lot of headache of manually promoting the types. Here we can take advantage of the fact that ```std::pow``` is already defined!

Comment (sudo-panda) [reply include/clad/Differentiator/BuiltinDerivatives.h:123]: Thanks @grimmmyshini. That's actually quite an elegant solution. I will try it out.

File: lib/Differentiator/DerivativeBuilder.cpp [hunk 3]
```diff
@@ -2664,7 +2673,14 @@ namespace clad {
...
           auto di = BuildOp(BO_Mul, dfdx(), ithResult);
-          ArgResultDecls[i]->setInit(di);
+
+          // Since clang won't automatically promote di to the type of the variable
+          // in which it is being stored, we need to implicitly convert the type
+          // so that clang won't throw an error.
+          auto res = m_Sema.PerformImplicitConversion(di, CEType,
+                                                      clang::Sema::AA_Casting);
+
+          ArgResultDecls[i]->setInit(res.get());
```
vgvassilev, line 2683: It would be worth moving this repetitive code in a separate utility function.

vgvassilev: ping.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -987,6 +987,11 @@ namespace clad {
     return CladTapeResult{*this, PushExpr, PopExpr, TapeRef};
   }
 
+  Expr* ReverseModeVisitor::GetImplicitConversion(
```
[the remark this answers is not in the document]
vgvassilev: You can pass it by reference.

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -299,6 +299,15 @@ namespace clad {
     clang::LookupResult& GetCladTapeBack();
     /// Instantiate clad::tape<T> type.
     clang::QualType GetCladTapeOfType(clang::QualType T);
+
+    /// Assigns the Init expression to VD after performing the necessary
+    /// implicit conversion
+    void PerformImplicitConversionAndAssign(clang::VarDecl* VD,
```
sudo-panda, line 305: @vgvassilev This seems to be better imo as it can be used in all visitors and also reduces code duplication.

vgvassilev: Ha, okay, fair point.

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -299,6 +299,15 @@ namespace clad {
...
+                                             clang::Expr* Init) {
+      VD->setInit(m_Sema.PerformImplicitConversion(
```
vgvassilev, line 307: Can you split this into two statements, it will make it clearer to read.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -2664,7 +2670,12 @@ namespace clad {
           auto ithResult = m_Sema.CreateBuiltinArraySubscriptExpr(Result, noLoc,
                                                                   I, noLoc).get();
           auto di = BuildOp(BO_Mul, dfdx(), ithResult);
-          ArgResultDecls[i]->setInit(di);
+
+          // Since clang won't automatically promote di to the type of the variable
+          // in which it is being stored, we need to implicitly convert the type
+          // so that clang won't throw an error.
```
vgvassilev, line 2676: I think both comments are redundant now, as we have a call to `PerformImplicitConversionAndAssign`. What do you think?

vgvassilev:
Looks good. Can you squash all commits into one?

vgvassilev:
The osx failures seem real.

sudo-panda:
> The osx failures seem real.

This might be due to the platform dependency of `math.h`.

vgvassilev:
Yes, the `__promote_2` is an implementation detail for libstdc++ and seems different for libc++. The `CHECK` directives can take regex and I'd suggest skipping all up to `f7_return`.

sudo-panda:
It should work now

vgvassilev:
Thanks a lot!

PR #216, opened 2021-03-23
Add cloning of function attributes

ioanaif:


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -734,6 +734,9 @@ namespace clad {
                                        CLAD_COMPAT_CLANG10_FunctionDecl_Create_ExtraParams(FD->getTrailingRequiresClause())
                                        );
     }
+    for (const auto *Attr : FD->attrs()) {
+      derivedFD->addAttr(Attr->clone(m_Context));
+    }
```
vgvassilev, line 739:
```suggestion
    // Clone original function's attributes.
    for (const auto *Attr : FD->attrs())
      derivedFD->addAttr(Attr->clone(m_Context));
```

vgvassilev: In fact, I would suggest to move these lines https://github.com/vgvassilev/clad/blob/0af111f9c5e73daf7679572966aed91315066b45/lib/Differentiator/DerivativeBuilder.cpp#L518-L547 and https://github.com/vgvassilev/clad/blob/0af111f9c5e73daf7679572966aed91315066b45/lib/Differentiator/DerivativeBuilder.cpp#L1827-L1857 and https://github.com/vgvassilev/clad/blob/0af111f9c5e73daf7679572966aed91315066b45/lib/Differentiator/DerivativeBuilder.cpp#L714-L736 into a static helper (say `CloneFunction`) function as they are code duplication. The attribute cloning should become part of that function.

vgvassilev (commented):
Nice work! I think we should adjust the failing test in `ClassMethodCall.C` and we should add a couple of more tests that make sure the attribute cloning works. We could add the ```__attribute__((always_inline))``` to some of the existing tests.

PR #219, opened 2021-03-26
Improvement/clone function attributes

ioanaif:


216.diff
```diff
@@ -0,0 +1,14 @@
...
+@@ -734,6 +734,9 @@ namespace clad {
+                                        CLAD_COMPAT_CLANG10_FunctionDecl_Create_ExtraParams(FD->getTrailingRequiresClause())
+                                        );
+     }
++    for (const auto *Attr : FD->attrs()) {
++      derivedFD->addAttr(Attr->clone(m_Context));
++    }
+     m_Derivative = derivedFD;
+ 
+     llvm::SmallVector<ParmVarDecl*, 4> params;
```
vgvassilev, line 14: We probably accidentally added this file.

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -44,9 +44,8 @@ namespace clad {
   using DeclWithContext = std::pair<clang::FunctionDecl*, clang::Decl*>;
 
   using DiffParams = llvm::SmallVector<const clang::VarDecl*, 16>;
-
   using VectorOutputs = std::vector<std::unordered_map<const clang::VarDecl*, clang::Expr*>>;
-
+
```
vgvassilev, line 48: I think that's a deed of your editor. Can we remove that change and configure it?

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -57,13 +56,16 @@ namespace clad {
     friend class ReverseModeVisitor;
     friend class HessianModeVisitor;
     friend class JacobianModeVisitor;
-
+
```
vgvassilev, line 59: Likewise.

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -57,13 +56,16 @@ namespace clad {
...
     clang::Sema& m_Sema;
     plugin::CladPlugin& m_CladPlugin;
     clang::ASTContext& m_Context;
     std::unique_ptr<utils::StmtClone> m_NodeCloner;
     clang::NamespaceDecl* m_BuiltinDerivativesNSD;
-
+    clang::FunctionDecl* cloneFunction(const clang::FunctionDecl* FD, clang::DeclContext* DC, 
+                            clang::Sema& m_Sema, clang::ASTContext&  m_Context, 
+                            clang::SourceLocation& noLoc, clang::DeclarationNameInfo name,
+                            clang::QualType functionType);
```
vgvassilev, line 68:
```suggestion
    clang::FunctionDecl* cloneFunction(const clang::FunctionDecl* FD,
                                                               clang::DeclContext* DC, 
                                                               clang::Sema& m_Sema,
                                                               clang::ASTContext&  m_Context, 
                                                               clang::SourceLocation& noLoc,
                                                               clang::DeclarationNameInfo name,
                                                               clang::QualType functionType);
```

We tend to follow the 80 column width -- I guess that indent should follow the rule.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -51,7 +51,42 @@ namespace clad {
       return;
     // Register the function on the redecl chain.
     derivedFD->setPreviousDecl(cast<FunctionDecl>(R.getFoundDecl()));
+  }
 
+  clang::FunctionDecl* DerivativeBuilder::cloneFunction(const clang::FunctionDecl* FD,
```
vgvassilev, line 56: We should do the same hyphenation as above. (Usually clang-format gives a good guidance.)

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -51,7 +51,42 @@ namespace clad {
...
+                                        clad_compat::Function_GetConstexprKind(FD)
+                                        CLAD_COMPAT_CLANG10_FunctionDecl_Create_ExtraParams(FD->getTrailingRequiresClause())
+                                        );
+    } else {
+      diag(DiagnosticsEngine::Error, FD->getEndLoc(),
+        "attempted differentiation of '%0' which is of unsupported type",
+         { FD->getNameAsString() });   
+      return {};
+    }
+    for (const auto *Attr : FD->attrs()) {
```
vgvassilev, line 86: When we have single like operations we do not add `{` and `}`.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -509,42 +544,13 @@ namespace clad {
     // Create the gradient function declaration.
     DeclContext* DC = const_cast<DeclContext*>(m_Function->getDeclContext());
 
-    FunctionDecl* hessianFD = nullptr;
     NamespaceDecl* enclosingNS = nullptr;
     llvm::SaveAndRestore<DeclContext*> SaveContext(m_Sema.CurContext);
     llvm::SaveAndRestore<Scope*> SaveScope(m_CurScope);
     m_Sema.CurContext = DC;
+    enclosingNS = RebuildEnclosingNamespaces(DC);
```
vgvassilev, line 551: Before the patch we called `RebuildEnclosingNamespaces` only in the case when we clone a FunctionDecl. I'd propose to do the same and make `cloneFunction` take a `VisitorBase&`.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -509,42 +544,13 @@ namespace clad {
...
-        clad_compat::Function_GetConstexprKind(m_Function)
-        CLAD_COMPAT_CLANG10_FunctionDecl_Create_ExtraParams(m_Function->getTrailingRequiresClause())
-        );
-    } else {
-      diag(DiagnosticsEngine::Error, m_Function->getEndLoc(),
-         "attempted differentiation of '%0' which is of unsupported type",
-          { m_Function->getNameAsString() });
-    return {};
-    }
+    FunctionDecl* hessianFD = m_Builder.cloneFunction(m_Function, DC, m_Sema, m_Context, noLoc, name, hessianFunctionType);
```
vgvassilev, line 553: We seem to exceed the 80 col limit here.

test/Gradient/Gradients.C
```diff
@@ -6,6 +6,26 @@
 #include "clad/Differentiator/Differentiator.h"
 #include <cmath>
 
+__attribute__((always_inline)) double f_attr(double x) {
```
vgvassilev, line 9: It seems our intent is to test mostly the attribute cloning, what about adding `__attribute__((always_inline))` to an already existing function in this file (eg. `f_add1`) and adjust the expected reference output?

test/Hessian/Hessians.C
```diff
@@ -5,6 +5,36 @@
 
 #include "clad/Differentiator/Differentiator.h"
 
+__attribute__((always_inline)) double f_attr (double x) {
```
vgvassilev, line 8: Likewise.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -51,7 +51,58 @@ namespace clad {
...
+  }
+
+  static bool hasAttribute(const Decl *D, attr::Kind Kind) {
+    for (const auto *Attribute : D->attrs())
+      if (Attribute->getKind() == Kind)
+        return true;
+    return false;
+  }
+
+  NamespaceDeclWithContext
```
vgvassilev, line 63: We should return `DeclWithContext` here.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -705,35 +729,17 @@ namespace clad {
...
-                                       CLAD_COMPAT_CLANG10_FunctionDecl_Create_ExtraParams(FD->getTrailingRequiresClause())
-                                       );
-    }
+    NamespaceDeclWithContext result{};
+
+    result = m_Builder.cloneFunction(FD, *this, DC, m_Sema,
+                                                      m_Context, noLoc, name,
+                                                      FD->getType());
+    FunctionDecl* derivedFD = result.first;
+    NamespaceDecl* enclosingNS = result.second;
```
vgvassilev, line 742: We don't need this definition, we can just return `result` [here](https://github.com/vgvassilev/clad/pull/219/files#diff-ca86e90607539e50579f9bc81aba290d66ebb00ca004fd021b1b68ac7a1e4774L825)

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -705,35 +729,17 @@ namespace clad {
...
-                                       /*default*/
-                                       FD->isInlineSpecified(),
-                                       FD->hasWrittenPrototype(),
-                                       clad_compat::Function_GetConstexprKind(FD)
-                                       CLAD_COMPAT_CLANG10_FunctionDecl_Create_ExtraParams(FD->getTrailingRequiresClause())
-                                       );
-    }
+    NamespaceDeclWithContext result{};
+
+    result = m_Builder.cloneFunction(FD, *this, DC, m_Sema,
```
vgvassilev, line 738:
```suggestion
    DeclWithContext result = m_Builder.cloneFunction(FD, *this, DC, m_Sema,
```

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -51,7 +51,58 @@ namespace clad {
...
+    } else if (isa<FunctionDecl>(FD)) {
+      enclosingNS = VD.RebuildEnclosingNamespaces(DC);
+      returnedFD = FunctionDecl::Create(m_Context, m_Sema.CurContext, noLoc,
+                                        name, functionType,
+                                        FD->getTypeSourceInfo(),
+                                        FD->getStorageClass(),
+                                        FD->isInlineSpecified(),
+                                        FD->hasWrittenPrototype(),
+                                      clad_compat::Function_GetConstexprKind(FD)
+                                        CLAD_COMPAT_CLANG10_FunctionDecl_Create_ExtraParams(FD->getTrailingRequiresClause())
```
vgvassilev, line 91: The failures are due to the fact that `CLAD_COMPAT_CLANG10_FunctionDecl_Create_ExtraParams` was invoked in the `VisitorBase` derived class and it has a `Clone` function which is conditionally called.

Here we can either pass the `VD`, or just modify the `CLAD_COMPAT_CLANG10_FunctionDecl_Create_ExtraParams` macro to do `VD.Clone` rather than just `Clone` for clang10.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -51,7 +51,62 @@ namespace clad {
...
+                                        FD->getStorageClass(),
+                                        FD->isInlineSpecified(),
+                                        FD->hasWrittenPrototype(),
+                                        clad_compat::Function_GetConstexprKind(FD)CLAD_COMPAT_CLANG10_FunctionDecl_Create_ExtraParams(FD->getTrailingRequiresClause()));
+    } else {
+      diag(DiagnosticsEngine::Error, FD->getEndLoc(),
+        "attempted differentiation of '%0' which is of unsupported type",
+         { FD->getNameAsString() });
+      return {};
+    }
```
vgvassilev, line 102: It seems that this else condition is dead code. I do not see how we can reach it. Let's delete it and turn this:```} else if (isa<FunctionDecl>(FD)) {``` into: 
```
} else {
  assert(isa<FunctionDecl>(FD) && "Unexpected!");
  ...
}
  ```

File: test/Gradient/Gradients.C [hunk 16]
```diff
@@ -6,11 +6,11 @@
 #include "clad/Differentiator/Differentiator.h"
 #include <cmath>
 
-double f_add1(double x, double y) {
+__attribute__((always_inline)) double f_add1(double x, double y) {
```
vgvassilev, line 9: To test the `hasAttribute` function: 
```suggestion
__attribute__((always_inline)) double f_add1(double x, double y);
__attribute__((always_inline)) double f_add1(double x, double y) {
```

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -63,7 +63,14 @@ namespace clad {
     clang::ASTContext& m_Context;
     std::unique_ptr<utils::StmtClone> m_NodeCloner;
     clang::NamespaceDecl* m_BuiltinDerivativesNSD;
-
+    DeclWithContext cloneFunction(const clang::FunctionDecl* FD,
+                                        clad::VisitorBase VB,
```
vgvassilev, line 67: Something happened with indentation here.

vgvassilev (approved):
Great job!

PR #224, opened 2021-04-15
Modify Forward Mode VisitCallExpr to support multiarg calls

sudo-panda:
This is done by:
- Differentiating the called function w.r.t each of its arguments
- Multiplying the results with the derivatives of the arguments itself
- Summing them all up

A faster way would be to use the Reverse mode gradient of the function and operate on it. But since forward mode Visit functions return the derivative of the expression we need to enclose them in a lambda function and call it in place. I feel this might become messy.

@vgvassilev could you please suggest which one to follow. 

P.S. the tests will fail because they need to be updated to match the changes.

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -1459,85 +1459,97 @@ namespace clad {
     if (m_DerivativeOrder == 1)
       s = "";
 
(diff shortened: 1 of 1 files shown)


vgvassilev:
@sudo-panda can you rebase this PR?

vgvassilev:
@sudo-panda what is the fate of this PR?

sudo-panda:
I am closing it for now I think using reverse mode to achieve the multiarg diff is better way so I will put up a PR for that

PR #225, opened 2021-04-19
Split DerivativeBuilder.cpp based on class

sudo-panda:
New files created are:
- ForwardModeVisitor.cpp
- HessianModeVisitor.cpp
- JacobianModeVisitor.cpp
- ReverseModeVisitor.cpp
- VisitorBase.cpp


sudo-panda:
@efremale Could you please review this PR?

vgvassilev:
@efremale what's your take of this refactoring?

efremale:
> @efremale what's your take of this refactoring?

Looks good to me!

efremale (approved):

vgvassilev:
@sudo-panda, can you make sure you run clang-format on these changes. The PR is a good opportunity to fix things.

sudo-panda:
> @sudo-panda, can you make sure you run clang-format on these changes. The PR is a good opportunity to fix things.

I already have. But just in case I have missed one I will do it again and check.

include/clad/Differentiator/VisitorBase.h
```diff
@@ -0,0 +1,316 @@
...
+    /// implicit conversion. This is required as clang doesn't add implicit
+    /// conversions while assigning values to variables which are initialized
+    /// after it is already declared.
+    void PerformImplicitConversionAndAssign(clang::VarDecl* VD,
+                                            clang::Expr* Init) {
+      // Implicitly convert Init into the type of VD
+      clang::Expr* ICE = m_Sema
+                             .PerformImplicitConversion(Init, VD->getType(),
+                                                        clang::Sema::AA_Casting)
+                             .get();
```
vgvassilev, line 309: I suspect we should tune the clang-format rules, this does not look good...

sudo-panda: What do you suggest it should look like?

Is this one better:
```
      clang::Expr* ICE = m_Sema
                             .PerformImplicitConversion(Init, VD->getType(),
                                                        clang::Sema::AA_Casting)
                                                        .get();
```

sudo-panda:
I had missed one.

include/clad/Differentiator/VisitorBase.h
vgvassilev:
```clang::Expr* ICE = m_Sema.PerformImplicitConversion(Init,
                                                                            VD->getType(),
                                                           clang::Sema::AA_Casting).get();
```
neither is nice...

vgvassilev:
> I had missed one.

We should only run clang-format on the diff not the files as that would bloat the history.

sudo-panda:
> > I had missed one.
> 
> We should only run clang-format on the diff not the files as that would bloat the history.
> 
> For example:
> 
> ```shell
> git checkout split-derivative-builder
> wget https://raw.githubusercontent.com/llvm/llvm-project/main/clang/tools/clang-format/git-clang-format
> chmod +x git-clang-format
> git rebase -i -x "git-clang-format master && git commit -a --allow-empty --fixup=HEAD" --strategy-option=theirs origin/master
> git log --oneline # to inspect the results
> git rebase --autosquash -i master # squash without poluting the history
> ```

I had created the file for the refactor so I ran it for the whole file.

include/clad/Differentiator/VisitorBase.h
vgvassilev: The `.get()` which makes things quite ugly is not meant to be used that way but to check if there was no error (in the StmtResult object) and then call `get()`..

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -53,86 +49,75 @@ namespace clad {
...
-                                   clang::Sema& m_Sema,
-                                   clang::ASTContext& m_Context,
-                                   clang::SourceLocation& noLoc,
-                                   clang::DeclarationNameInfo name,
-                                   clang::QualType functionType) {
+  DeclWithContext DerivativeBuilder::cloneFunction(
+      const clang::FunctionDecl* FD, clad::VisitorBase VD,
+      clang::DeclContext* DC, clang::Sema& m_Sema, clang::ASTContext& m_Context,
+      clang::SourceLocation& noLoc, clang::DeclarationNameInfo name,
+      clang::QualType functionType) {
```
vgvassilev, line 63: That also looks a bit weird We should probably prefer they way it was currently written...

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -53,86 +49,75 @@ namespace clad {
...
-                                         functionType, 
-                                         FD->getTypeSourceInfo(),
-                                         FD->getStorageClass(),
-                                         FD->isInlineSpecified(),
-                                         clad_compat::Function_GetConstexprKind
-                                         (FD), noLoc);
+      returnedFD = CXXMethodDecl::Create(
+          m_Context, CXXRD, noLoc, name, functionType, FD->getTypeSourceInfo(),
+          FD->getStorageClass(), FD->isInlineSpecified(),
+          clad_compat::Function_GetConstexprKind(FD), noLoc);
```
vgvassilev, line 71: I guess that's controlled by the same switch as above.

lib/Differentiator/DerivativeBuilder.cpp
sudo-panda: Do I manually fix this or try to change clang-format to do so automatically?

vgvassilev: I'd think we should try to find the rule that fixes this. It may have to do with the break new line 'penalty'.

vgvassilev:
We may want to use that opportunity to try to improve the coverage with tests (in separate commits).

lib/Differentiator/DerivativeBuilder.cpp
sudo-panda: I have updated the clang-format file please take a look.

include/clad/Differentiator/VisitorBase.h
```diff
@@ -303,12 +303,13 @@ namespace clad {
...
-                             .get();
-      // Assign the resulting expression to the variable declaration
-      VD->setInit(ICE);
+      clang::ActionResult<clang::Expr*> ICAR = m_Sema
+          .PerformImplicitConversion(Init, VD->getType(),
+                                     clang::Sema::AA_Casting);
+      if(!ICAR.isInvalid()) {
+        // Assign the resulting expression to the variable declaration
+        VD->setInit(ICAR.get());
+      }
```
vgvassilev, line 312: Else, we should report error? Or assert is invalid.

vgvassilev:
Do we want to go an extra mile and try to improve the test coverage misses according to the codecov bot comments?

sudo-panda:
> Do we want to go an extra mile and try to improve the test coverage misses according to the codecov bot comments?

I am unable to access the full codecov report can you check which file is the culprit?

vgvassilev:
If you browse https://github.com/vgvassilev/clad/pull/225/files you will see some comments from the bot. Are they sufficient?

sudo-panda:
> If you browse https://github.com/vgvassilev/clad/pull/225/files you will see some comments from the bot. Are they sufficient?
Yeah

They seem to be for very specific conditions so I don't think it is very important.

vgvassilev:
@sudo-panda, can you rebase this PR please?

vgvassilev (commented):
Apologies for the delay. LGTM! The CI failures are due to some unrelated packaging issue with clang-9.

PR #228, opened 2021-04-26
Initial LLVM/Clang-12 support.

alexander-penev:


vgvassilev (commented):
Thanks, let's see if the knowall-bot will trigger after this review comments.

vgvassilev:
The llvm-12 support is complete. There are some unrelated failures for the llvm-9 which apparently somehow changed in the packaging system -- still to be investigated.

PR #230, opened 2021-05-23
Recent llvm-lit requires fstrings which are part of python3.6.

vgvassilev:
Thanks @reikdas.


vgvassilev:
The failures are unrelated to this PR. We experience failures in the llvm setup and clang-9. They will be fixed as part of another PR.

PR #232, opened 2021-05-26
Feature/gradient cuda support

ioanaif:


include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -81,23 +87,23 @@ namespace custom_derivatives {
...
 
   template <typename T1, typename T2>
-  decltype(pow(T1(), T2())) pow_darg1(T1 x, T2 exponent) {
+  CUDA_HOST_DEVICE decltype(pow(T1(), T2())) pow_darg1(T1 x, T2 exponent) {
     return pow(x, exponent) * log(x);
   }
 
   template <typename T1, typename T2>
-  void pow_grad(T1 x, T2 exponent, decltype(pow(T1(), T2())) * result) {
+  CUDA_HOST_DEVICE void pow_grad(T1 x, T2 exponent, decltype(pow(T1(), T2())) * result) {
```
vgvassilev, line 100: Mind the 80 column limitation. Maybe run clang-format on this PR.

include/clad/Differentiator/Tape.h
```diff
@@ -28,70 +33,94 @@ namespace clad {
...
+    CUDA_HOST_DEVICE void emplace_back(ArgsT&&... args) {
       if (_size >= _capacity)
         grow();
       ::new (end()) T(std::forward<ArgsT>(args)...);
       _size += 1;
     }
-   
-    std::size_t size() const {
+    
+    CUDA_HOST_DEVICE std::size_t size() const {
```
vgvassilev, line 43: Spacing.

include/clad/Differentiator/Tape.h
```diff
@@ -28,70 +33,94 @@ namespace clad {
...
+    CUDA_HOST_DEVICE reference back() {
       assert(_size);
       return begin()[_size - 1];
     }
-    const_reference back() const {
+    CUDA_HOST_DEVICE const_reference back() const {
       assert(_size);
       return begin()[_size - 1];
     }
-
```
vgvassilev, line 64: We should keep that.

include/clad/Differentiator/Tape.h
```diff
@@ -28,70 +33,94 @@ namespace clad {
...
-      // Move values from old storage to the new storage. Should call move
-      // constructors on non-trivial types, otherwise is expected to use
-      // memcpy/memmove.
-      std::uninitialized_copy(std::make_move_iterator(begin()),
-                              std::make_move_iterator(end()), new_data);
-      // Destroy all values in the old storage.
-      destroy(begin(), end());
-      _data = new_data;
-    }
+    #ifdef __CUDACC__
```
vgvassilev, line 78: Likewise to the other suggestion -- please consider splitting the variant code into a separate function while keeping the rest unmodified.

vgvassilev (commented):
Thanks a lot for working on this! We should also have test coverage and a demo showing how to enable cuda.

include/clad/Differentiator/Differentiator.h
```diff
@@ -28,28 +28,45 @@ extern "C" {
 }
 
 namespace clad {
+
+  /// Length utily for CUDA environment
```
vgvassilev, line 32:
```suggestion
  /// \returns the size of a c-style string 
```

include/clad/Differentiator/Tape.h
```diff
@@ -26,68 +32,88 @@ namespace clad {
...
+    /// Move values from old to new storage
+    CUDA_HOST_DEVICE T* MoveValues(std::size_t _capacity) {
+      #ifdef __CUDACC__
+        // Allocate raw storage (without calling constructors of T) of new capacity.
+        T* new_data = static_cast<T*>(::operator new(_capacity * sizeof(T)));
+        assert(new_data);
+        // Move values from old storage to the new storage. Should call move
+        // constructors on non-trivial types, otherwise is expected to use
+        // memcpy/memmove.
+        for (auto i=begin(); i != end(); ++i, ++new_data) {
```
vgvassilev, line 49:
```suggestion
        for (auto i=begin(), e = end(); i != e; ++i, ++new_data) {
```
This way we tell the reader that we do not modify the collection in the loop.

include/clad/Differentiator/Tape.h
```diff
@@ -26,68 +32,88 @@ namespace clad {
...
+            new(new_data) T(std::move(*i));
+        }
```
vgvassilev, line 51: Is that code block because older versions of cuda do not implement `uninitialized_copy` but newer have it? If that is the case we should only use that code when there is not implementation of `uninitialized_copy`.

include/clad/Differentiator/Tape.h
```diff
@@ -26,68 +32,88 @@ namespace clad {
...
     /// Remove the last value from the tape.
-    void pop_back() {
+    CUDA_HOST_DEVICE void pop_back() {
       assert(_size);
       _size -= 1;
       end()->~T();
     }
 
   private:
+
```
vgvassilev, line 105: Stray space.

test/CUDA/GradientCutda.C
```diff
@@ -0,0 +1,54 @@
...
+    cudaMemcpy(d_x, x, N*sizeof(double), cudaMemcpyHostToDevice);
+
+    double *result, *d_result;
+
+    result = (double*)malloc(N*sizeof(double));
+    cudaMalloc(&d_result, N*sizeof(double));
+
+    compute<<<1, 1>>>(d_x, d_result, N);
+    cudaMemcpy(result, d_result, N*sizeof(double), cudaMemcpyDeviceToHost);
+    printf("Result = {%.2f, %.2f, %.2f}\n", result[0], result[1], result[2]);   // CHECK-EXEC: Result is = {0.00, 0.00, 0.00}
```
vgvassilev, line 52: Would it make sense to verify against the result got from the cpu version? FileCheck has a way to remember variables but I am not sure if that's across runs...

vgvassilev (commented):
Thanks. I think we are getting there. Could you also modify the commit message by having a blank line after the first sentence.

vgvassilev (approved):
Thanks for working on this! LGTM!

PR #235, opened 2021-05-31
Add support for building doxygen documentation using cmake

parth-07:
CMake options and variables added:
- `CLAD_INCLUDE_DOCS`: Generate targets of all the _enabled_ documentation tools
- `CLAD_ENABLE_DOXYGEN`: Enables the generation of browsable HTML documentation using doxygen. Defaults to OFF.

To build documentation,
1) Run CMake with `-DCLAD_ENABLE_DOXYGEN=ON` and `-DCLAD_INCLUDE_DOCS=ON` flags
2) Run `make doxygen-clad` in the build directory.

This PR also adds configuration files needed to easily host clad documentation on [readthedocs.org](https://readthedocs.org/).
I have temporarily hosted clad documentation [here](https://trying-read-the-docs.readthedocs.io/en/latest/index.html).


vgvassilev:
Great work! Could you also consider uploading the documentation to a website such as readthedocs.org? There are number of workarounds to include doxygen-style documentation there.

parth-07:
@vgvassilev Yes definitely.

.readthedocs.yaml
```diff
@@ -0,0 +1,13 @@
...
+sphinx:
+  configuration: docs/sphinx/conf.py
+
+build:
+  image: latest
+  apt_packages:
+    - clang-9
+    - cmake
+    - libclang-9-dev
+    - llvm-9-dev
```
vgvassilev, line 12: Would it make sense to use the latest supported version? As of today it's llvm-12.

parth-07: Yes, I will modify `llvm` packages to the latest versions available in ubuntu 18.04 (readthedocs runner uses ubuntu 18.04). But since we are just using these packages for configuring with CMake, and not actually building clad (we are just building docs), updated llvm packages should not make much difference.

vgvassilev: Thanks!

parth-07: As per the package list [here](https://packages.ubuntu.com/bionic/allpackages), `llvm-10` and `clang-10` are the latest versions available, I have updated the configuration file to use them.

vgvassilev (approved):
LGTM!

Issue #168, opened 2020-07-31
Support of differentiation of multi-arg calls in the forward mode


blide:
```c++
double multiarg(double x, double y) {
  return x + y;
}
double f(double x) {
  return multiarg(x, 1);
}
clad::differentiate(f);
```
Now will fail with:
```
clang-9: /src/tools/clad/lib/Differentiator/DerivativeBuilder.cpp:1456: clad::StmtDiff
clad::ForwardModeVisitor::VisitCallExpr(const clang::CallExpr*): Assertion `(CE->getNumArgs() <= 1) &&
"forward differentiation of multi-arg calls is currently broken"' failed.
```

sudo-panda:
@vgvassilev please close this issue too

sudo-panda:
@efremale suggests in #237:

>By the way, it would be good to make a similar change of parameter structure for forward mode too at some point to fix #168
>
>The idea (not specific for arrays but works in the same for arrays) is:
>For `double f(double a, double b)`, any application of clad::differentiate on f will generate/use
>double `f_jvp(double a, double da, double b, double db)`, where
>`f_da(a, b)` is `f_jvp(a, 1, b, 0)` and
>`f_db(a, b)` is `f_jvp(a, 0, b, 1)` and in general
>for any `x`, `y`, `f_jvp(a, x, b, y)` is equal to matrix-vector product `matmul(f_jacobian(a, b), {x, y})`, i.e the Jacobian-vector product (JVP).
>In particular in nested function calls the value of `f_jvp(a, _d_a, b, _d_b)` is exactly what we need to fix the issue #168.
>
>Note that this does not require explicit evaluation of jacobian, forward mode naturally computes JVP faster than evaluating Jacobian and matmul on it.
>
>This does not require any change in forward mode algorithm code, only in function signature. Current code works in the same way, it only hardcodes `_d_a = 1`, `_d_b = 0` e.g. inside `f_d_a` rather than accepting general values of `_d_a`, `_d_b` (which will give JVP in the result).

PR #237, opened 2021-06-03
Add array diff support to reverse mode and hessian mode

sudo-panda:
Reverse mode and hessian mode now support array diff.

The signature of the grad function has been changed. For example, if the input function has the signature:
```c++
int f(int a, int *b)
```
then the grad function for it will have the signature:
```c++
void f_grad(int a, int *b, int* _d_a, int *_d_b)
```
Reverse mode currently doesn't differentiate w.r.t individual elements of an array. If individual elements of an array are provided the whole array is differentiated instead. This is because it doesn't cost much to do the extra work but instead makes the the output more clear.

The signature of hessian function has not changed, but there are some additional requirements when differentiating arrays. The user will need to specify the index interval of the array that needs to be differentiated using the `args` param of `clad::hessian`. Eg.
```c++
double f(int a, int b[4]); 
clad::hessian(f, "a, b[0:4]")
```

### DIFF
--- a/include/clad/Differentiator/Array.h
+++ b/include/clad/Differentiator/Array.h
@@ -1,0 +1,66 @@
+#ifndef CLANG_ARRAY_H
+#define CLANG_ARRAY_H
+
+#include "clad/Differentiator/CladConfig.h"
+
+#include <type_traits>
+
+namespace clad {
+  /// This class is not meant to be used by user. It is used by clad internally
+  /// only
+  template <typename T> class array {
+  private:
+    /// The pointer to the underlying array
+    T* m_arr = nullptr;
+    /// The size of the array
+    std::size_t m_size = 0;
+
+  public:
+    /// Delete default constructor
+    array() = delete;
+    /// Constructor to create an array of the specified size
+    CUDA_HOST_DEVICE array(std::size_t size)
+        : m_arr(new T[size]{static_cast<T>(0)}), m_size(size) {}
+
+    /// Destructor to delete the array if it was created by array_ref
+    CUDA_HOST_DEVICE ~array() { delete[] m_arr; }
+
+    /// Returns the size of the underlying array
+    CUDA_HOST_DEVICE std::size_t size() { return m_size; }
+    /// Returns the ptr of the underlying array
+    CUDA_HOST_DEVICE T* ptr() { return m_arr; }
+    /// Returns the reference to the location at the index of the underlying
+    /// array
+    CUDA_HOST_DEVICE T& operator[](std::size_t i) { return m_arr[i]; }
+    /// Returns the reference to the underlying array
+    CUDA_HOST_DEVICE T& operator*() { return *m_arr; }
+
+    // Arithmetic overloads
+    /// Divides the number from every element in the array
+    CUDA_HOST_DEVICE array<T>& operator/=(T n) {
+      for (std::size_t i = 0; i < m_size; i++)
+        m_arr[i] /= n;
+      return *this;
+    }
+    /// Multiplies the number to every element in the array
+    CUDA_HOST_DEVICE array<T>& operator*=(T n) {
+      for (std::size_t i = 0; i < m_size; i++)
+        m_arr[i] *= n;
+      return *this;
+    }
+    /// Adds the number to every element in the array
+    CUDA_HOST_DEVICE array<T>& operator+=(T n) {
+      for (std::size_t i = 0; i < m_size; i++)
+        m_arr[i] += n;
+      return *this;
+    }
+    /// Subtracts the number from every element in the array
+    CUDA_HOST_DEVICE array<T>& operator-=(T n) {
+      for (std::size_t i = 0; i < m_size; i++)
+        m_arr[i] -= n;
+      return *this;
+    }
+  };
+} // namespace clad
+
+#endif // CLANG_ARRAY_H
--- a/include/clad/Differentiator/ArrayRef.h
+++ b/include/clad/Differentiator/ArrayRef.h
@@ -1,0 +1,79 @@
+#ifndef CLAD_ARRAY_REF_H
+#define CLAD_ARRAY_REF_H
+
+#include "clad/Differentiator/Array.h"
+#include "clad/Differentiator/CladConfig.h"
+
+#include <assert.h>
+#include <type_traits>
+
+namespace clad {
+  /// Stores the pointer to and the size of an array and provides some helper
+  /// functions for it. The array is supplied should have a life greater than
+  /// that of the array_ref
+  template <typename T> class array_ref {
+  private:
+    /// The pointer to the underlying array
+    T* m_arr = nullptr;
+    /// The size of the array
+    std::size_t m_size = 0;
+
+  public:
+    /// Delete default constructor
+    array_ref() = delete;
+    /// Constructor to store the pointer to and size of an array supplied by the
+    /// user
+    CUDA_HOST_DEVICE array_ref(T* arr, std::size_t size)
+        : m_arr(arr), m_size(size) {}
+    /// Constructor for arrays having size equal to 1 or non pointer types to
+    /// store their addresses
+    CUDA_HOST_DEVICE array_ref(T* a) : m_arr(a), m_size(1) {}
+    /// Constructor for clad::array types
+    CUDA_HOST_DEVICE array_ref(array<T>& a)
+        : m_arr(a.ptr()), m_size(a.size()) {}
+
+    /// Returns the size of the underlying array
+    CUDA_HOST_DEVICE std::size_t size() { return m_size; }
+    /// Returns the reference to the location at the index of the underlying
+    /// array
+    CUDA_HOST_DEVICE T& operator[](std::size_t i) { return m_arr[i]; }
+    /// Returns the reference to the underlying array
+    CUDA_HOST_DEVICE T& operator*() { return *m_arr; }
+
+    // Arithmetic overloads
+    /// Divides the arrays element wise
+    CUDA_HOST_DEVICE array_ref<T>& operator/=(array_ref<T>& Ar) {
+      assert(m_size == Ar.size() && "Size of both the array_refs must be equal "
+                                    "for carrying out addition assignment");
+      for (std::size_t i = 0; i < m_size; i++)
+        m_arr[i] /= Ar[i];
+      return *this;
+    }
+    /// Multiplies the arrays element wise
+    CUDA_HOST_DEVICE array_ref<T>& operator*=(array_ref<T>& Ar) {
+      assert(m_size == Ar.size() && "Size of both the array_refs must be equal "
+                                    "for carrying out addition assignment");
+      for (std::size_t i = 0; i < m_size; i++)
+        m_arr[i] *= Ar[i];
+      return *this;
+    }
+    /// Adds the arrays element wise
+    CUDA_HOST_DEVICE array_ref<T>& operator+=(array_ref<T>& Ar) {
+      assert(m_size == Ar.size() && "Size of both the array_refs must be equal "
+                                    "for carrying out addition assignment");
+      for (std::size_t i = 0; i < m_size; i++)
+        m_arr[i] += Ar[i];
+      return *this;
+    }
+    /// Subtracts the arrays element wise
+    CUDA_HOST_DEVICE array_ref<T>& operator-=(array_ref<T>& Ar) {
+      assert(m_size == Ar.size() && "Size of both the array_refs must be equal "
+                                    "for carrying out addition assignment");
+      for (std::size_t i = 0; i < m_size; i++)
+        m_arr[i] -= Ar[i];
+      return *this;
+    }
+  };
+} // namespace clad
+
+#endif // CLAD_ARRAY_REF_H
--- a/include/clad/Differentiator/ArrayRef.h
+++ b/include/clad/Differentiator/ArrayRef.h
@@ -34,6 +34,14 @@ namespace clad {
 
     /// Returns the size of the underlying array
     CUDA_HOST_DEVICE std::size_t size() { return m_size; }
+    /// Returns an array_ref to a part of the underlying array starting at
+    /// offset and having the specified size
+    CUDA_HOST_DEVICE array_ref<T> slice(std::size_t offset, std::size_t size) {
+      assert((offset >= 0) && (offset + size <= m_size) &&
+             "Window is outside array. Please provide an offset and size "
+             "inside the array size.");
+      return array_ref<T>(&m_arr[offset], size);
+    }
     /// Returns the reference to the location at the index of the underlying
     /// array
     CUDA_HOST_DEVICE T& operator[](std::size_t i) { return m_arr[i]; }
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -11,6 +11,7 @@
 // should go.
 namespace custom_derivatives{}
 
+#include "clad/Differentiator/ArrayRef.h"
 #include "clad/Differentiator/CladConfig.h"
 
 #include <math.h>
@@ -93,10 +94,11 @@ namespace custom_derivatives {
   }
 
   template <typename T1, typename T2>
-  CUDA_HOST_DEVICE void 
-  pow_grad(T1 x, T2 exponent, decltype(pow(T1(), T2())) * result) {
-    result[0] += pow_darg0(x, exponent);
-    result[1] += pow_darg1(x, exponent);
+  CUDA_HOST_DEVICE void
+  pow_grad(T1 x, T2 exponent, clad::array_ref<decltype(pow(T1(), T2()))> _d_x,
+           clad::array_ref<decltype(pow(T1(), T2()))> _d_y) {
+    *_d_x += pow_darg0(x, exponent);
+    *_d_y += pow_darg1(x, exponent);
   }
 
   template <typename T>
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -60,8 +60,15 @@ namespace clad {
 
 namespace clad {
   /// A pair of FunctionDecl and potential enclosing context, e.g. a function
-  // in nested namespaces
+  /// in nested namespaces.
+  // This is the type returned by cloneFunction. Using OverloadedDeclWithContext
+  // instead would lead to unnecessarily returning a nullptr in the overloaded
+  // FD
   using DeclWithContext = std::pair<clang::FunctionDecl*, clang::Decl*>;
+  /// A tuple which consists of a FunctionDecl, it's potential enclosing context
+  /// and optionally it's overload FunctionDecl
+  using OverloadedDeclWithContext =
+      std::tuple<clang::FunctionDecl*, clang::Decl*, clang::FunctionDecl*>;
   using DiffParams = llvm::SmallVector<const clang::VarDecl*, 16>;
   using IndexIntervalTable = llvm::SmallVector<IndexInterval, 16>;
   using DiffParamsWithIndices = std::pair<DiffParams, IndexIntervalTable>;
@@ -121,8 +128,8 @@ namespace clad {
     ///\returns The differentiated function and potentially created enclosing
     /// context.
     ///
-    DeclWithContext Derive(const clang::FunctionDecl* FD,
-                           const DiffRequest & request);
+    OverloadedDeclWithContext Derive(const clang::FunctionDecl* FD,
+                                     const DiffRequest& request);
   };
 
 } // end namespace clad
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -342,12 +342,10 @@ namespace clad {
   /// Function for Hessian matrix computation
   /// Given  a function f, clad::hessian generates all the second derivatives
   /// of the original function, (they are also columns of a Hessian matrix)
-  template <typename ArgSpec = const char*,
-            typename F,
-            typename DerivedFnType = ExtractDerivedFnTraits_t<F>>
+  template <typename ArgSpec = const char*, typename F,
+            typename DerivedFnType = HessianDerivedFnTraits_t<F>>
   CladFunction<DerivedFnType> __attribute__((annotate("H")))
-  hessian(F f,
-          ArgSpec args = "",
+  hessian(F f, ArgSpec args = "",
           DerivedFnType derivedFn = static_cast<DerivedFnType>(nullptr),
           const char* code = "") {
     assert(f && "Must pass in a non-0 argument");
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -67,17 +67,48 @@ namespace clad {
   }
 
   /// Access return the last value in the tape.
-  template <typename T>
-  CUDA_HOST_DEVICE T& back(tape<T>& of) {
+  template <typename T> CUDA_HOST_DEVICE T& back(tape<T>& of) {
     return of.back();
+  }
+
+  /// Pad the args supplied with nullptr(s) or zeros to match the the num of
+  /// params of the function and then execute the function using the padded args
+  /// i.e. we are adding default arguments as we cannot do that with
+  /// meta programming
+  ///
+  /// For example:
+  /// Let's assume we have a function with the signature:
+  ///   fn_grad(double i, double j, int k, int l);
+  /// and f is a pointer to fn_grad
+  /// and args are the supplied arguments- 1.0, 2.0 and Args has their type
+  /// (double, double)
+  ///
+  /// When pad_and_execute(DropArgs_t<sizeof...(Args), decltype(f)>{}, f, args)
+  /// is run, the Rest variadic argument will have the types (int, int).
+  /// pad_and_execute will then make up for the remaining args by appending 0s
+  /// and the return statement translates to:
+  ///   return f(1.0, 2.0, 0, 0);
+  // for executing non-member functions
+  template <class... Rest, class F, class... Args>
+  return_type_t<F> execute_with_default_args(list<Rest...>, F f,
+                                             Args&&... args) {
+    return f(static_cast<Args>(args)..., static_cast<Rest>(0)...);
+  }
+
+  // for executing member-functions
+  template <class... Rest, class ReturnType, class C, class Obj, class... Args>
+  auto execute_with_default_args(list<Rest...>, ReturnType C::*f, Obj&& obj,
+                                 Args&&... args) -> return_type_t<decltype(f)> {
+    return (static_cast<Obj>(obj).*f)(static_cast<Args>(args)...,
+                                      static_cast<Rest>(0)...);
   }
 
   // Using std::function and std::mem_fn introduces a lot of overhead, which we
   // do not need. Another disadvantage is that it is difficult to distinguish a
   // 'normal' use of std::{function,mem_fn} from the ones we must differentiate.
-  /// Explicitly passing `FunctorT` type is necessary for maintaining 
+  /// Explicitly passing `FunctorT` type is necessary for maintaining
   /// const correctness of functor types.
-  /// Default value of `Functor` here is temporary, and should be removed 
+  /// Default value of `Functor` here is temporary, and should be removed
   /// once all clad differentiation functions support differentiating functors.
   template <typename F, typename FunctorT = ExtractFunctorTraits_t<F>>
   class CladFunction {
@@ -182,7 +213,8 @@ namespace clad {
       template <class Fn, class... Args>
       return_type_t<CladFunctionType> execute_helper(Fn f, Args&&... args) {
         // `static_cast` is required here for perfect forwarding.
-        return f(static_cast<Args>(args)...);
+        return execute_with_default_args(DropArgs_t<sizeof...(Args), F>{}, f,
+                                         static_cast<Args>(args)...);
       }
 
       /// Helper functions for executing member derived functions.
@@ -198,7 +230,10 @@ namespace clad {
       return_type_t<CladFunctionType>
       execute_helper(ReturnType C::*f, Obj&& obj, Args&&... args) {
         // `static_cast` is required here for perfect forwarding.
-        return (static_cast<Obj>(obj).*f)(static_cast<Args>(args)...);
+        return execute_with_default_args(DropArgs_t<sizeof...(Args),
+                                                    decltype(f)>{},
+                                         f, static_cast<Obj>(obj),
+                                         static_cast<Args>(args)...);
       }
       /// If user have not passed object explicitly, then this specialization
       /// will be used and derived function will be called through the object
@@ -210,7 +245,10 @@ namespace clad {
                "No default object set, explicitly pass an object to "
                "CladFunction::execute");
         // `static_cast` is required here for perfect forwarding.
-        return (m_Functor->*f)(static_cast<Args>(args)...);
+        return execute_with_default_args(DropArgs_t<sizeof...(Args),
+                                                    decltype(f)>{},
+                                         f, *m_Functor,
+                                         static_cast<Args>(args)...);
       }
   };
 
@@ -271,15 +309,13 @@ namespace clad {
   /// \param[in] args independent parameters information
   /// \returns `CladFunction` object to access the corresponding derived
   /// function.
-  template <typename ArgSpec = const char*,
-            typename F,
-            typename DerivedFnType = ExtractDerivedFnTraits_t<F>,
+  template <typename ArgSpec = const char*, typename F,
+            typename DerivedFnType = GradientDerivedFnTraits_t<F>,
             typename = typename std::enable_if<
                 !std::is_class<remove_reference_and_pointer_t<F>>::value>::type>
   CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>> __attribute__((
       annotate("G"))) CUDA_HOST_DEVICE
-  gradient(F f,
-           ArgSpec args = "",
+  gradient(F f, ArgSpec args = "",
            DerivedFnType derivedFn = static_cast<DerivedFnType>(nullptr),
            const char* code = "") {
     assert(f && "Must pass in a non-0 argument");
@@ -290,15 +326,13 @@ namespace clad {
   /// Specialization for differentiating functors.
   /// The specialization is needed because objects have to be passed
   /// by reference whereas functions have to be passed by value.
-  template <typename ArgSpec = const char*,
-            typename F,
-            typename DerivedFnType = ExtractDerivedFnTraits_t<F>,
+  template <typename ArgSpec = const char*, typename F,
+            typename DerivedFnType = GradientDerivedFnTraits_t<F>,
             typename = typename std::enable_if<
                 std::is_class<remove_reference_and_pointer_t<F>>::value>::type>
   CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>> __attribute__((
       annotate("G"))) CUDA_HOST_DEVICE
-  gradient(F&& f,
-           ArgSpec args = "",
+  gradient(F&& f, ArgSpec args = "",
            DerivedFnType derivedFn = static_cast<DerivedFnType>(nullptr),
            const char* code = "") {
     return CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>>(
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -7,8 +7,10 @@
 #ifndef CLAD_DIFFERENTIATOR
 #define CLAD_DIFFERENTIATOR
 
-#include "CladConfig.h"
+#include "Array.h"
+#include "ArrayRef.h"
 #include "BuiltinDerivatives.h"
+#include "CladConfig.h"
 #include "FunctionTraits.h"
 #include "Tape.h"
 
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -355,12 +355,10 @@ namespace clad {
                                        code);
   }
 
-  template <typename ArgSpec = const char*,
-            typename F,
-            typename DerivedFnType = ExtractDerivedFnTraits_t<F>>
+  template <typename ArgSpec = const char*, typename F,
+            typename DerivedFnType = JacobianDerivedFnTraits_t<F>>
   CladFunction<DerivedFnType> __attribute__((annotate("J")))
-  jacobian(F f,
-           ArgSpec args = "",
+  jacobian(F f, ArgSpec args = "",
            DerivedFnType derivedFn = static_cast<DerivedFnType>(nullptr),
            const char* code = "") {
     assert(f && "Must pass in a non-0 argument");
--- a/include/clad/Differentiator/FunctionTraits.h
+++ b/include/clad/Differentiator/FunctionTraits.h
@@ -449,12 +449,75 @@ namespace clad {
     using type = NoFunction*;
   };
 
+  template <class... Args> struct SelectLast;
+
+  template <class... Args>
+  using SelectLast_t = typename SelectLast<Args...>::type;
+
+  template <class T> struct SelectLast<T> { using type = T; };
+
+  template <class T, class... Args> struct SelectLast<T, Args...> {
+    using type = typename SelectLast<Args...>::type;
+  };
+
+  template <class T> struct JacobianDerivedFnTraits {};
+
+  // JacobianDerivedFnTraits is used to deduce type of the derived functions
+  // derived using jacobian mode
+  template <class T>
+  using JacobianDerivedFnTraits_t = typename JacobianDerivedFnTraits<T>::type;
+
+  // JacobianDerivedFnTraits specializations for pure function pointer types
+  template <class ReturnType, class... Args>
+  struct JacobianDerivedFnTraits<ReturnType (*)(Args...)> {
+    using type = void (*)(Args..., SelectLast_t<Args...>);
+  };
+
+  /// These macro expansions are used to cover all possible cases of
+  /// qualifiers in member functions when declaring JacobianDerivedFnTraits.
+  /// They need to be read from bottom to top. Starting from the use of AddCON,
+  /// the call to which is used to pass the cases with and without C-style
+  /// varargs, then as the macro name AddCON says it adds cases of const
+  /// qualifier. The AddVOL and AddREF macro similarly add cases for volatile
+  /// qualifier and reference respectively. The AddNOEX adds cases for noexcept
+  /// qualifier only if it is supported and finally AddSPECS declares the
+  /// function with all the cases
+#define JacobianDerivedFnTraits_AddSPECS(var, cv, vol, ref, noex)              \
+  template <typename R, typename C, typename... Args>                          \
+  struct JacobianDerivedFnTraits<R (C::*)(Args...) cv vol ref noex> {          \
+    using type = void (C::*)(Args..., SelectLast_t<Args...>) cv vol ref noex;  \
+  };
+
+#if __cpp_noexcept_function_type > 0
+#define JacobianDerivedFnTraits_AddNOEX(var, con, vol, ref)                    \
+  JacobianDerivedFnTraits_AddSPECS(var, con, vol, ref, )                       \
+      JacobianDerivedFnTraits_AddSPECS(var, con, vol, ref, noexcept)
+#else
+#define JacobianDerivedFnTraits_AddNOEX(var, con, vol, ref)                    \
+  JacobianDerivedFnTraits_AddSPECS(var, con, vol, ref, )
+#endif
+
+#define JacobianDerivedFnTraits_AddREF(var, con, vol)                          \
+  JacobianDerivedFnTraits_AddNOEX(var, con, vol, )                             \
+      JacobianDerivedFnTraits_AddNOEX(var, con, vol, &)                        \
+          JacobianDerivedFnTraits_AddNOEX(var, con, vol, &&)
+
+#define JacobianDerivedFnTraits_AddVOL(var, con)                               \
+  JacobianDerivedFnTraits_AddREF(var, con, )                                   \
+      JacobianDerivedFnTraits_AddREF(var, con, volatile)
+
+#define JacobianDerivedFnTraits_AddCON(var)                                    \
+  JacobianDerivedFnTraits_AddVOL(var, )                                        \
+      JacobianDerivedFnTraits_AddVOL(var, const)
+
+  JacobianDerivedFnTraits_AddCON(()); // Declares all the specializations
+
   // ExtractDerivedFnTraits is used to deduce type of the derived functions
   // derived using hessian and jacobian differentiation modes
   // It SHOULD NOT be used to get traits of derived functions derived using
   // forward or reverse differentiation mode
-  template<class ReturnType, class = void>
-  struct ExtractDerivedFnTraits {};
+  template <class ReturnType, class = void> struct ExtractDerivedFnTraits {};
+
   template<class T>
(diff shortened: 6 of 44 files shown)


vgvassilev (commented):
Can you run clang-format for each commit individually. The command should be something like:

```bash
git checkout <branch_name>
wget https://raw.githubusercontent.com/llvm/llvm-project/main/clang/tools/clang-format/git-clang-format
chmod +x git-clang-format
git rebase -i -x "git-clang-format master && git commit -a --allow-empty --fixup=HEAD" --strategy-option=theirs origin/master
git log --oneline # to inspect the results
git rebase --autosquash -i master # squash without polluting the history

**Comment by vgvassilev** [comment]
@efremale could you review this PR?

File: lib/Differentiator/HessianModeVisitor.cpp [hunk 1]
```diff
@@ -31,8 +31,47 @@ namespace clad {
...
+  DeriveUsingForwardAndReverseMode(clad::plugin::CladPlugin& CP,
+                                   DiffRequest IndependentArgRequest,
+                                   const Expr* ForwardModeArgs,
+                                   const Expr* ReverseModeArgs) {
+    // Derives function once in forward mode w.r.t to ForwardModeArgs
+    IndependentArgRequest.Args = ForwardModeArgs;
+    IndependentArgRequest.Mode = DiffMode::forward;
+    IndependentArgRequest.CallUpdateRequired = false;
+    FunctionDecl* firstDerivative =
+        plugin::ProcessDiffRequest(CP, IndependentArgRequest);
```
vgvassilev, line 60: That patch did not introduce it but I think that's a layering violation worth a FIXME comment.

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -41,56 +80,78 @@ namespace clad {
...
+        // arg
+        auto argIndex = it - args.begin();
+        if (isArrayOrPointerType(PVD->getType())) {
+          if (indexIntervalTable.size() == 0 ||
+              indexIntervalTable[argIndex].size() == 0) {
+            diag(DiagnosticsEngine::Error,
+                 request.Args ? request.Args->getEndLoc() : noLoc,
+                 "Hessian mode differentiation w.r.t to array or pointer "
+                 "parameters needs explicit declaration of the indices of "
+                 "the array using the args parameter");
```
vgvassilev, line 104: Can we add a `; did you mean XXX` to this diagnostic to give an example what the user should/can do.

lib/Differentiator/HessianModeVisitor.cpp
sudo-panda: Could you explain the violation?

vgvassilev: The `plugin::` namespace is supposed to be only the 'driver' to initiate the process. It seems unnatural the deeper implementation such as the YYYVisitors to call back into it to schedule some derivation process.

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -62,6 +62,7 @@ namespace clad {
   /// A pair of FunctionDecl and potential enclosing context, e.g. a function
   // in nested namespaces
   using DeclWithContext = std::pair<clang::FunctionDecl*, clang::Decl*>;
+  using OverloadedDeclWithContext = std::tuple<clang::FunctionDecl*, clang::Decl*, clang::FunctionDecl*>;
```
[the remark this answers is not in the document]
sudo-panda: The clone function uses `DeclWithContext`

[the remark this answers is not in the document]
sudo-panda: Using `DeclWithContext` as suggested causes problems while using `std::tie` to receive the output. Should I still go ahead with this and remove `std::tie` and create methods to `get` the required variable?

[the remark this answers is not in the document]
vgvassilev: No, please keep the `std::tie` for now but add a comment to capture that discussion for future.

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -31,8 +31,47 @@ namespace clad {
...
+    QualType CharTyConst = C.CharTy.withConst();
+    QualType StrTy = clad_compat::getConstantArrayType(
+        C,
+        CharTyConst,
+        llvm::APInt(32, independentArgString.size() + 1),
+        nullptr,
+        ArrayType::Normal,
+        /*IndexTypeQuals*/ 0);
+    const StringLiteral* IndependentArgStringLiteral = StringLiteral::Create(
+        C, independentArgString, StringLiteral::Ascii, false, StrTy, noLoc);
```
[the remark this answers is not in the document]
sudo-panda: Yes

sudo-panda:
> Can you run clang-format for each commit individually. The command should be something like:
> 
> ```shell
> git checkout <branch_name>
> wget https://raw.githubusercontent.com/llvm/llvm-project/main/clang/tools/clang-format/git-clang-format
> chmod +x git-clang-format
> git rebase -i -x "git-clang-format master && git commit -a --allow-empty --fixup=HEAD" --strategy-option=theirs origin/master
> git log --oneline # to inspect the results
> git rebase --autosquash -i master # squash without polluting the history
> ```

@vgvassilev any way to make it not touch the test commits

lib/Differentiator/HessianModeVisitor.cpp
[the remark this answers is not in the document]
vgvassilev: That's ugly :(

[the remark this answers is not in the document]
sudo-panda: Should I change the `clang-format` file?

vgvassilev:
> > Can you run clang-format for each commit individually. The command should be something like:
> > ```shell
> > git checkout <branch_name>
> > wget https://raw.githubusercontent.com/llvm/llvm-project/main/clang/tools/clang-format/git-clang-format
> > chmod +x git-clang-format
> > git rebase -i -x "git-clang-format master && git commit -a --allow-empty --fixup=HEAD" --strategy-option=theirs origin/master
> > git log --oneline # to inspect the results
> > git rebase --autosquash -i master # squash without polluting the history
> > ```
> 
> @vgvassilev any way to make it not touch the test commits

Not sure, maybe:

`git rebase -i -x "git-clang-format master . ':!test' && git commit -a --allow-empty --fixup=HEAD" --strategy-option=theirs origin/master` ?

lib/Differentiator/HessianModeVisitor.cpp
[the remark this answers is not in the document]
vgvassilev: The question would be what looks nicer:

```cpp
const StringLiteral* IndependentArgStringLiteral =
    StringLiteral::Create(C, independentArgString, StringLiteral::Ascii,
                                      false, StrTy, noLoc);
```
I guess I am more used to that but does not look much nicer. The example from `clad_compat::getConstantArrayType` is also funny...

include/clad/Differentiator/FunctionTraits.h
```diff
@@ -223,132 +223,267 @@ namespace clad {
...
+
+#define HessianDerivedFnTraits_AddVOL(var, con) \
+  HessianDerivedFnTraits_AddREF(var, con,) \
+  HessianDerivedFnTraits_AddREF(var, con, volatile)
+
+#define HessianDerivedFnTraits_AddCON(var) \
+  HessianDerivedFnTraits_AddVOL(var,) \
+  HessianDerivedFnTraits_AddVOL(var, const)
+
+  HessianDerivedFnTraits_AddCON(()) // Declares all the specializations
```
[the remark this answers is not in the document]
sudo-panda: I have added it here for now: https://github.com/vgvassilev/clad/pull/237/files#diff-321620026993392cecb8c4421c6bea9f107c577d113fc51a51e2b4bb1e701bcdR288-R295

lib/Differentiator/HessianModeVisitor.cpp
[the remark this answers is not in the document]
sudo-panda: How about
```
    QualType StrTy = 
        clad_compat::getConstantArrayType(C,
                                          CharTyConst,
                                          llvm::APInt(32, independentArgString.size() + 1),
                                          nullptr,
                                          ArrayType::Normal,
                                          /*IndexTypeQuals*/ 0);
```
and
```
const StringLiteral* IndependentArgStringLiteral = 
    StringLiteral::Create(C,
                          independentArgString,
                          StringLiteral::Ascii,
                          false,
                          StrTy,
                          noLoc);
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -295,7 +346,119 @@ namespace clad {
...
+        callExpr = BuildCallExprToMemFn(gradientCMD, callArgsRef);
+      } else {
+        assert(isa<FunctionDecl>(gradientFD) && "Unexpected!");
+        callExpr = BuildCallExprToFunction(gradientFD, callArgsRef);
+      }
+      addToCurrentBlock(callExpr);
+      Stmt* gradientOverloadBody = endBlock();
+
+      gradientOverloadFD->setBody(gradientOverloadBody);
+    }
```
parth-07, line 457: I believe we are missing ending scopes here. 
```suggestion
      endScope(); // Function body scope
      m_Sema.PopFunctionScopeInfo();
      m_Sema.PopDeclContext();
      endScope(); // Function decl scope
    }
```

tools/ClangPlugin.cpp
```diff
@@ -169,7 +174,9 @@ namespace clad {
         // If this is the last required derivative order, replace the function
         // inside a call to clad::differentiate/gradient with its derivative.
         if (request.CallUpdateRequired && lastDerivativeOrder)
-          request.updateCall(DerivativeDecl, m_CI.getSema());
+          request.updateCall(DerivativeDecl,
```
parth-07, line 177: I think we can avoid sending an extra argument to `updateCall` by deciding here only which of `DerivativeDecl` and `OverloadedDerivativeDecl` should be `replacementFD`. 

```suggestion
          request.updateCall((OverloadedDerivativeDecl ? OverloadedDerivativeDecl
                                                 : DerivativeDecl),
                       m_CI.getSema());
```

sudo-panda: The extra argument is also required for `dump` ing the actual gradient function. Also when Hessian Mode uses the gradient it can use the actual gradient instead of the overloaded gradient.

parth-07: Oh okay, I didn't consider these things. Thanks for the details.

include/clad/Differentiator/Array.h
```diff
@@ -0,0 +1,71 @@
+#ifndef CLANG_ARRAY_H
+#define CLANG_ARRAY_H
+
+#ifdef __CUDACC__
+#define CUDA_HOST_DEVICE __host__ __device__
+#else
+#define CUDA_HOST_DEVICE
+#endif
```
vgvassilev, line 8: Could you move these into some dedicated header maybe called `CladConfig.h`? Maybe in a separate commit or PR.

include/clad/Differentiator/ArrayRef.h
```diff
@@ -0,0 +1,92 @@
+#ifndef CLAD_ARRAY_REF_H
+#define CLAD_ARRAY_REF_H
+
+#ifdef __CUDACC__
+#define CUDA_HOST_DEVICE __host__ __device__
+#else
+#define CUDA_HOST_DEVICE
+#endif
+
+#include "Array.h"
```
vgvassilev, line 10:
```suggestion
#include "clad/Differentiator/Array.h"
```

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -18,6 +18,7 @@
 namespace custom_derivatives{}
 
 #include <math.h>
+#include "ArrayRef.h"
```
vgvassilev, line 21:
```suggestion
#include "clad/Differentiator/ArrayRef.h"

#include <math.h>
```

include/clad/Differentiator/FunctionTraits.h
```diff
@@ -1,6 +1,8 @@
 #ifndef FUNCTION_TRAITS
 #define FUNCTION_TRAITS
 
+#include "ArrayRef.h"
```
vgvassilev, line 4:
```suggestion
#include "clad/Differentiator/ArrayRef.h"
```

lib/Differentiator/HessianModeVisitor.cpp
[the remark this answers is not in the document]
vgvassilev: Looks better I think..

lib/Differentiator/StmtClone.cpp
```diff
@@ -279,6 +279,16 @@ Stmt* StmtClone::VisitShuffleVectorExpr(ShuffleVectorExpr* Node) {
...
 
+Stmt* StmtClone::VisitExprWithCleanups(ExprWithCleanups* Node) {
+  ExprWithCleanups* result =
+      ExprWithCleanups::Create(Ctx,
+                               Node->getSubExpr(),
+                               Node->cleanupsHaveSideEffects(),
+                               {}); // Sending empty CleanupObjects list because
+                                    // we cannot create proper clones of them
+  return result;
+}
```
vgvassilev, line 290: The implementation looks trivial, can it go with just a `DEFINE_CREATE_EXPR`?

test/Arrays/Arrays.C
```diff
@@ -89,46 +89,46 @@ double const_dot_product(double x, double y, double z) {
...
-//CHECK-NEXT:           _d_consts[2] += _r5;
-//CHECK-NEXT:       }
-//CHECK-NEXT:       {
-//CHECK-NEXT:           _result[0UL] += _d_vars[0];
-//CHECK-NEXT:           _result[1UL] += _d_vars[1];
-//CHECK-NEXT:           _result[2UL] += _d_vars[2];
-//CHECK-NEXT:       }
-//CHECK-NEXT:   }
+// CHECK:   void const_dot_product_grad(double x, double y, double z, clad::array_ref<double> _d_x, clad::array_ref<double> _d_y, clad::array_ref<double> _d_z) {
+// CHECK-NEXT:       double _d_vars[3] = {};
```
vgvassilev, line 93: Can you remove the formatting changes from the tests and keep only the changes introduced by this PR?

test/ROOT/TFormula.C
```diff
@@ -25,31 +25,31 @@ Double_t TFormula_example(Double_t* x, Double_t* p) {
...
-//CHECK-NEXT:           _result[1] += _r1;
-//CHECK-NEXT:           _result[2] += _r1;
-//CHECK-NEXT:           Double_t _r2 = 1 * custom_derivatives::Exp_darg0(_t2);
-//CHECK-NEXT:           _result[0] += -_r2;
-//CHECK-NEXT:           Double_t _r3 = 1 * custom_derivatives::Abs_darg0(_t3);
-//CHECK-NEXT:           _result[1] += _r3;
-//CHECK-NEXT:       }
-//CHECK-NEXT:   }
+void TFormula_example_grad_1(Double_t* x, Double_t* p, Double_t* _d_p);
+// CHECK:   void TFormula_example_grad_1(Double_t *x, Double_t *p, clad::array_ref<Double_t> _d_p) {
```
vgvassilev, line 29: That maybe a breaking change for ROOT. If it is, we should make sure we mention it in the release notes.

test/Arrays/Arrays.C
```diff
@@ -89,46 +89,46 @@ double const_dot_product(double x, double y, double z) {
 //CHECK-NEXT:       return _d_vars[0] * consts[0] + vars[0] * _d_consts[0] + _d_vars[1] * consts[1] + vars[1] * _d_consts[1] + _d_vars[2] * consts[2] + vars[2] * _d_consts[2];
```
vgvassilev, line 89: Maybe it would be worth adding a demo explaining the recommended use of arrays with clad..

vgvassilev (commented):
Great work! Amazing!

I've added some comments from the partial review.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -45,6 +45,7 @@ namespace clad {
     unsigned outputArrayCursor = 0;
     unsigned numParams = 0;
     bool isVectorValued = false;
+    unsigned arrLen = 10;
```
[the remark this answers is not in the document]
sudo-panda: That needs to be removed. It was the workaround used before now we know the array size.

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -97,10 +98,12 @@ namespace custom_derivatives {
...
 
   template <typename T1, typename T2>
-  CUDA_HOST_DEVICE void 
-  pow_grad(T1 x, T2 exponent, decltype(pow(T1(), T2())) * result) {
-    result[0] += pow_darg0(x, exponent);
-    result[1] += pow_darg1(x, exponent);
+  CUDA_HOST_DEVICE void pow_grad(T1 x,
+                                 T2 exponent,
+                                 decltype(pow(T1(), T2()))* _d_x,
+                                 decltype(pow(T1(), T2()))* _d_y) {
```
sudo-panda, line 104: Add `clad::array_ref`

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -62,6 +62,10 @@ namespace clad {
   /// A pair of FunctionDecl and potential enclosing context, e.g. a function
   // in nested namespaces
```
sudo-panda, line 63: Add comment why we need DeclWithContext

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -224,7 +295,29 @@ namespace clad {
...
+            noLoc);
+        // Create the size argument.
+        auto SizeArg = IntegerLiteral::Create(
+            m_Context,
+            llvm::APInt(size_type_bits, IndependentArgsSize[j]),
+            size_type,
+            noLoc);
+        SmallVector<Expr*, 2> Args({OffsetArg, SizeArg});
+        // Create the hessianMatrix.get_window(OffsetArg, SizeArg) expression.
+        auto WindowAccessExpr = GetArrayRefGetWindowExpr(m_Result, Args);
```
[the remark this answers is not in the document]
sudo-panda: Changed to `slice`

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -815,12 +1029,19 @@ namespace clad {
...
+    if (isArrayOrPointerType(target->getType()))
       // Create the _result[idx] expression.
       result = BuildArraySubscript(target, reverseIndices);
+    else if (isArrayRefType(target->getType())) {
+      result = m_Sema
+                   .ActOnArraySubscriptExpr(getCurrentScope(),
+                                            target,
+                                            ASE->getExprLoc(),
+                                            reverseIndices.back(),
+                                            noLoc)
```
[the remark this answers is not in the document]
sudo-panda: Cannot switch because `BuildArraySubscript` only works on arrays or pointers or vectors

lib/Differentiator/HessianModeVisitor.cpp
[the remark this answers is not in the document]
sudo-panda: @vgvassilev please checkout PR #268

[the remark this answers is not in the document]
vgvassilev: @sudo-panda, does that mean that we will also allow eg:
```void fn (
  arg1,
  arg2) {```
  
?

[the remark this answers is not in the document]
Comment (sudo-panda) [reply lib/Differentiator/HessianModeVisitor.cpp]: Yeah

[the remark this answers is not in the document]
Comment (sudo-panda) [reply lib/Differentiator/HessianModeVisitor.cpp]: Is that unacceptable?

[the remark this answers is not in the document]
Comment (vgvassilev) [reply lib/Differentiator/HessianModeVisitor.cpp]: I risk to become annoying but it does not seem to be consistent to what we have been following. What would we/I usually do is:
```cpp
void
long_fn_name(arg1, arg2,
             argN, argN+1) {
```

I guess we cannot have this both ways.

At the call sites:

```cpp
auto name =
   long_fn_name(arg1, arg2,...
                argN, argN+1);
```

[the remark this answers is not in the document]
sudo-panda: I will make clang-format follow this then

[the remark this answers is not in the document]
vgvassilev: That would be great, I am not sure if that's possible though.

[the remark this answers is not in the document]
sudo-panda: @vgvassilev this is what I have achieved so far I am not sure if I can align it with the opening bracket:
```c++
    // The output paremeter "_result".
    params.back() =
        ParmVarDecl::Create(m_Context, gradientFD, noLoc, noLoc,
                            &m_Context.Idents.get(resultArg()),
                            paramTypes.back(),
                            m_Context
                                .getTrivialTypeSourceInfo(paramTypes.back(),
                                                          noLoc),
                            params.front()->getStorageClass(),
                            /* No default value */ nullptr);
    if (params.back()->getIdentifier())
      m_Sema.PushOnScopeChains(params.back(), getCurrentScope(),
                               /*AddToContext*/ false);
```
and
```c++
  static inline CXXMemberCallExpr*
  CXXMemberCallExpr_Create(ASTContext& Ctx, Expr* Fn, ArrayRef<Expr*> Args,
                           QualType Ty, ExprValueKind VK, SourceLocation RP) {
    return CXXMemberCallExpr::Create(const_cast<ASTContext&>(Ctx), Fn, Args, Ty,
                                     VK, RP);
  }
```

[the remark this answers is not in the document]
sudo-panda: Like for example in this one:
```c++
  static inline CXXOperatorCallExpr* CXXOperatorCallExpr_Create(
      ASTContext& Ctx, OverloadedOperatorKind OpKind, Expr* Fn,
      ArrayRef<Expr*> Args, QualType Ty, ExprValueKind VK,
      SourceLocation OperatorLoc,
      CLAD_COMPAT_CLANG11_CXXOperatorCallExpr_Create_ExtraParamsOverride
          FPFeatures
              CLAD_COMPAT_CLANG11_CXXOperatorCallExpr_Create_ExtraParamsPar) {
#if CLANG_VERSION_MAJOR < 8
    return new (Ctx) CXXOperatorCallExpr(Ctx, OpKind, Fn, Args, Ty, VK,
                                         OperatorLoc, FPFeatures);
#elif CLANG_VERSION_MAJOR >= 8
    return CXXOperatorCallExpr::Create(
        Ctx, OpKind, Fn, Args, Ty, VK, OperatorLoc,
        FPFeatures
            CLAD_COMPAT_CLANG11_CXXOperatorCallExpr_Create_ExtraParamsUse);
#endif
  }
```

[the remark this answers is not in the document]
vgvassilev: > am not sure if I can align it with the opening bracket

what do you mean? Otherwise what you showed looks pretty good.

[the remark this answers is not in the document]
sudo-panda: > what do you mean? Otherwise what you showed looks pretty good.

If the args or params are huge like in the [above](https://github.com/vgvassilev/clad/pull/237#discussion_r676042479) case it will start them from the next line.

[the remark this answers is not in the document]
vgvassilev: Ah, i see now. That’s something we can live with. And humans should not be looking at Compatibility.h anyway ;)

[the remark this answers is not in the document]
sudo-panda: Pushed the changes to the PR

efremale (commented):
Great! This is a good idea.

Is there a specific reason for having pointers (or `array_ref`s) for output gradient parameter for scalar input parameters? Wouldn't it be better (in terms of being easier to use and having smaller code size) to pass them as references? E.g.  `double f(double a, double b)` -> `f_grad(double a, double & da, double b, double & db)`.

efremale:
By the way, it would be good to make a similar change of parameter structure for forward mode too at some point to fix https://github.com/vgvassilev/clad/issues/168

The idea (not specific for arrays but works in the same for arrays) is:
For `double f(double a, double b)`, any application of `clad::differentiate` on `f` will generate/use
`double f_jvp(double a, double da, double b, double db)`, where
`f_da(a, b)` is `f_jvp(a, 1, b, 0)` and
`f_db(a, b)` is `f_jvp(a, 0, b, 1)` and in general
for any `x, y`, `f_jvp(a, x, b, y)` is equal to matrix-vector product `matmul(f_jacobian(a, b), {x, y})`, i.e the Jacobian-vector product (JVP). 
In particular in nested function calls the value of `f_jvp(a, _d_a, b, _d_b)` is exactly what we need to fix the issue 168.

Note that this does not require explicit evaluation of jacobian, forward mode naturally computes JVP faster than evaluating Jacobian and matmul on it.

This does not require any change in forward mode algorithm code, only in function signature. Current code works in the same way, it only hardcodes `_d_a = 1, _d_b = 0` e.g. inside `f_d_a` rather than accepting general values of `_d_a, _d_b` (which will give JVP in the result).

sudo-panda:
> Great! This is a good idea.
> 
> Is there a specific reason for having pointers (or `array_ref`s) for output gradient parameter for scalar input parameters? Wouldn't it be better (in terms of being easier to use and having smaller code size) to pass them as references? E.g. `double f(double a, double b)` -> `f_grad(double a, double & da, double b, double & db)`.

That was the initial plan but for `execute` in `CladFunction` to work `clad` needs to determine the type of the derived function before the plugin can parse the arguments of the function. This works well when the user wants to differentiate the function w.r.t. all parameters but if they want to differentiate it w.r.t. only the specified number of parameters the signature of the derived function can no longer be determined. So to solve this we had to do some changes:

1. We had to change all diff parameters to have the same type so we could no longer distinguish between derived types of scaler and array. 
2. We had to shift all the diff parameters to the end.
3. When the user requested only partial number of params had to create an overload for the derived gradient function that would have diff params for all the original params in all the parameters. The extra arguments are then given default argument like functionality using the [pad_and_execute](https://github.com/vgvassilev/clad/blob/1fbce9342831541c081010975ab34a7b94d8f9ed/include/clad/Differentiator/Differentiator.h#L92-L104) helper function.
Eg.
   ```c++
   // Requested function:
   int f(int a, int* b) {
     . . .
   }

   // Derived function:
   void f_grad1(int a, int* b, int* d_b) {
     . . .
   }

   // Derived function overload:
   void f_grad1(int a, int* b, int* d_b, int* d_0) {
     return f_grad1(a, b, d_b);
   }
   ```

include/clad/Differentiator/Differentiator.h
```diff
@@ -70,15 +70,36 @@ namespace clad {
   }
 
   // for executing non-member functions
-  template<class F, class... Args> 
+  template <class... Rest, class F, class... Args>
+  return_type_t<F> pad_and_execute(list<Rest...>, F f, Args&&... args) {
```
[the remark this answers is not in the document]
vgvassilev: How about changing the name to `execute_with_default_args`?

include/clad/Differentiator/FunctionTraits.h
```diff
@@ -2,6 +2,8 @@
 #define FUNCTION_TRAITS
 #include <type_traits>
 
+#include "clad/Differentiator/ArrayRef.h"
```
vgvassilev, line 5:
```suggestion
#include "clad/Differentiator/ArrayRef.h"

#include <type_traits>
```

include/clad/Differentiator/FunctionTraits.h
```diff
@@ -167,279 +169,429 @@ namespace clad {
...
-    using type = void (C::*)(Args..., ReturnType*) const volatile& noexcept;
+#define REM_CTOR(...) __VA_ARGS__
+
+  // Setup for DropArgs
+  template <typename... T> struct list {};
+
+  struct dummy {};
+
+  template <typename T> struct wrap {
+    constexpr operator dummy() const { return {}; }
```
vgvassilev, line 279: Did you ask what's the idea of `operator dummy` here?

include/clad/Differentiator/FunctionTraits.h
```diff
@@ -167,279 +169,429 @@ namespace clad {
...
+
+#define JacobianDerivedFnTraits_AddVOL(var, con)                               \
+  JacobianDerivedFnTraits_AddREF(var, con, )                                   \
+      JacobianDerivedFnTraits_AddREF(var, con, volatile)
+
+#define JacobianDerivedFnTraits_AddCON(var)                                    \
+  JacobianDerivedFnTraits_AddVOL(var, )                                        \
+      JacobianDerivedFnTraits_AddVOL(var, const)
+
+  JacobianDerivedFnTraits_AddCON(()) // Declares all the specializations
```
vgvassilev, line 513: Maybe add a semicolon to fix the indentation of the next few lines?

include/clad/Differentiator/FunctionTraits.h
```diff
@@ -167,279 +169,429 @@ namespace clad {
...
+#define HessianDerivedFnTraits_AddVOL(var, con)                                \
+  HessianDerivedFnTraits_AddREF(var, con, )                                    \
+      HessianDerivedFnTraits_AddREF(var, con, volatile)
+
+#define HessianDerivedFnTraits_AddCON(var)                                     \
+  HessianDerivedFnTraits_AddVOL(var, ) HessianDerivedFnTraits_AddVOL(var, const)
+
+  HessianDerivedFnTraits_AddCON(()) // Declares all the specializations
+
+      /// Compute type of derived function of function, method or functor when
```
vgvassilev, line 567: This seems misindented due to a missing semicolon of the prev PP macro.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -294,6 +287,14 @@ namespace clad {
            "attempt to differentiate unsupported operator, ignored.",
            args);
     }
+    /// Builds an overload for the gradient function that has derived params for
+    /// all the arguments of the requested function and it calls the original
+    /// gradient function internally
+    clang::FunctionDecl* CreateGradientOverload(
+        llvm::SmallVector<clang::QualType, 16>& GradientParamTypes,
+        llvm::SmallVector<clang::ParmVarDecl*, 4>& GradientParams,
```
vgvassilev, line 295:
```suggestion
        llvm::SmallVectorImpl<QualType>& GradientParamTypes,
        llvm::SmallVectorImpl<ParmVarDecl*>& GradientParams,
```

include/clad/Differentiator/VisitorBase.h
```diff
@@ -71,7 +71,7 @@ namespace clad {
           m_DerivativeInFlight(false), m_Derivative(nullptr),
           m_Function(nullptr) {}
 
-    using Stmts = llvm::SmallVector<clang::Stmt*, 16>;
+    using Stmts = llvm::SmallVector<clang::Stmt*, 32>;
```
vgvassilev, line 74: Can you revert this change. We should debug the underlying problem.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -185,10 +186,13 @@ namespace clad {
...
     ASTContext& C = SemaRef.getASTContext();
+
+    FunctionDecl* replacementFD = OverloadedFD ? OverloadedFD : FD;
     // Create ref to generated FD.
-    Expr* DRE = DeclRefExpr::Create(C, oldDRE->getQualifierLoc(), noLoc,
-                                    FD, false, FD->getNameInfo(), FD->getType(),
-                                    oldDRE->getValueKind());
+    Expr* DRE =
+        DeclRefExpr::Create(C, oldDRE->getQualifierLoc(), noLoc, replacementFD,
+                            false, replacementFD->getNameInfo(),
```
vgvassilev, line 194:
```suggestion
                            /*RefersToEnclosingVariableOrCapture=*/false, replacementFD->getNameInfo(),
```

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -224,7 +307,30 @@ namespace clad {
...
+
+      size_t columnIndex = 0;
+      // Create Expr parameters for each independent arg in the CallExpr
+      for (size_t j = 0, n = IndependentArgsSize.size(); j < n; j++) {
+        // Create the offset argument.
+        auto OffsetArg =
+            IntegerLiteral::Create(m_Context,
+                                   llvm::APInt(size_type_bits,
+                                               HessianMatrixStartIndex +
+                                                   columnIndex),
```
vgvassilev, line 319: Can we move that into a separate local variable? And maybe move it outside of the loop as it seems loop invariant.

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -224,7 +307,30 @@ namespace clad {
...
+                                   size_type, noLoc);
+        // Create the size argument.
+        auto SizeArg =
+            IntegerLiteral::Create(m_Context,
+                                   llvm::APInt(size_type_bits,
+                                               IndependentArgsSize[j]),
```
vgvassilev, line 325: Could you move this into a separate local variable?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -79,8 +79,149 @@ namespace clad {
 
   ReverseModeVisitor::~ReverseModeVisitor() {}
 
-  DeclWithContext ReverseModeVisitor::Derive(const FunctionDecl* FD,
-                                             const DiffRequest& request) {
+  FunctionDecl* ReverseModeVisitor::CreateGradientOverload(
+      SmallVector<QualType, 16>& GradientParamTypes,
+      SmallVector<ParmVarDecl*, 4>& GradientParams,
```
vgvassilev, line 84:
```suggestion
      SmallVectorImpl<QualType>& GradientParamTypes,
      SmallVectorImpl<ParmVarDecl*>& GradientParams,
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -79,8 +79,149 @@ namespace clad {
...
+
+    QualType gradientFunctionOverloadType =
+        m_Context
+            .getFunctionType(m_Context.VoidTy,
+                             llvm::ArrayRef<QualType>(GradientParamTypes.data(),
+                                                      GradientParamTypes
+                                                          .size()),
+                             // Cast to function pointer.
+                             dyn_cast<FunctionProtoType>(m_Function->getType())
+                                 ->getExtProtoInfo());
```
vgvassilev, line 109: Move both parameters in local variables please.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -79,8 +79,149 @@ namespace clad {
...
+    overloadParams.reserve(m_Function->getNumParams() * 2);
+    callArgs.reserve(GradientParamTypes.size());
+
+    for (auto* GVD : GradientParams) {
+      auto* VD =
+          ParmVarDecl::Create(m_Context, gradientOverloadFD, noLoc, noLoc,
+                              GVD->getIdentifier(), GVD->getType(),
+                              GVD->getTypeSourceInfo(), GVD->getStorageClass(),
+                              // Clone default arg if present.
+                              (GVD->hasDefaultArg()
```
vgvassilev, line 135: Please move that into a separate variable.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -79,8 +79,149 @@ namespace clad {
...
+    for (size_t i = 0; i < remainingArgs; i++) {
+      IdentifierInfo* DVDII = &m_Context.Idents.get("_d_" + std::to_string(i));
+
+      auto DVD = ParmVarDecl::
+          Create(m_Context, gradientOverloadFD, noLoc, noLoc, DVDII,
+                 DerivedOutputParamType,
+                 m_Context.getTrivialTypeSourceInfo(DerivedOutputParamType,
+                                                    noLoc),
+                 StorageClass::SC_None,
+                 // Clone default arg if present.
```
vgvassilev, line 154: Is this comment outdated?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -79,8 +79,149 @@ namespace clad {
...
+      // and above
+      callExpr = BuildCallExprToMemFn(thisExpr,
+                                      /*isArrow=*/true, GradientFD->getName(),
+                                      callArgsRef);
+#else
+      // NOUR_None is not supported in clang versions 6 and below
+      auto memberExpr =
+          MemberExpr::Create(m_Context, thisExpr, true, noLoc,
+                             NestedNameSpecifierLoc(gradientCMD->getQualifier(),
+                                                    nullptr),
```
vgvassilev, line 193:
```suggestion
                                                    /*Data=*/nullptr),
```
Could you move this in a separate variable?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -79,8 +79,149 @@ namespace clad {
...
+                             noLoc, gradientCMD,
+                             DeclAccessPair::make(gradientCMD,
+                                                  gradientCMD->getAccess()),
```
vgvassilev, line 196: Likewise.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -79,8 +79,149 @@ namespace clad {
...
+#if CLANG_VERSION_MAJOR <= 9
+      // This method is unable to lookup the GradientFD in clang versions 10
+      // and above
+      callExpr = BuildCallExprToMemFn(thisExpr,
+                                      /*isArrow=*/true, GradientFD->getName(),
+                                      callArgsRef);
+#else
+      // NOUR_None is not supported in clang versions 6 and below
+      auto memberExpr =
+          MemberExpr::Create(m_Context, thisExpr, true, noLoc,
```
vgvassilev, line 191:
```suggestion
          MemberExpr::Create(m_Context, thisExpr, /*IsArrow=*/true, noLoc,
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -79,8 +79,149 @@ namespace clad {
...
+                             NestedNameSpecifierLoc(gradientCMD->getQualifier(),
+                                                    nullptr),
+                             noLoc, gradientCMD,
+                             DeclAccessPair::make(gradientCMD,
+                                                  gradientCMD->getAccess()),
+                             gradientCMD->getNameInfo(), nullptr,
```
vgvassilev, line 197:
```suggestion
                             gradientCMD->getNameInfo(), /*TemplateArgs=*/nullptr,
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -168,98 +327,124 @@ namespace clad {
...
+    llvm::SmallVector<ParmVarDecl*, 4> params;
+    llvm::SmallVector<ParmVarDecl*, 4> outputParams;
+    params.reserve(m_Function->getNumParams() + args.size());
+    for (auto* PVD : m_Function->parameters()) {
+      auto VD =
+          ParmVarDecl::Create(m_Context, gradientFD, noLoc, noLoc,
+                              PVD->getIdentifier(), PVD->getType(),
+                              PVD->getTypeSourceInfo(), PVD->getStorageClass(),
+                              // Clone default arg if present.
+                              (PVD->hasDefaultArg()
```
vgvassilev, line 339: Please move out into a local variable.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -168,98 +327,124 @@ namespace clad {
...
+                                   ? Clone(PVD->getDefaultArg())
+                                   : nullptr));
+      if (VD->getIdentifier())
+        m_Sema.PushOnScopeChains(VD, getCurrentScope(),
+                                 /*AddToContext*/ false);
```
vgvassilev, line 344:
```suggestion
                                 /*AddToContext=*/false);
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -168,98 +327,124 @@ namespace clad {
...
+              &m_Context.Idents.get("_d_" + PVD->getNameAsString());
+
+          auto DVD = ParmVarDecl::
+              Create(m_Context, gradientFD, noLoc, noLoc, DVDII,
+                     DerivedOutputParamType,
+                     m_Context.getTrivialTypeSourceInfo(DerivedOutputParamType,
+                                                        noLoc),
+                     PVD->getStorageClass(),
+                     // Clone default arg if present.
+                     nullptr);
```
vgvassilev, line 364:
```suggestion

                     /*DefArg=*/nullptr);
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -168,98 +327,124 @@ namespace clad {
...
+
+    if (isVectorValued) {
+      // The output parameter "_jacobianMatrix".
+      params.push_back(
+          ParmVarDecl::
+              Create(m_Context, gradientFD, noLoc, noLoc,
+                     &m_Context.Idents.get("jacobianMatrix"),
+                     DerivedOutputParamType,
+                     m_Context.getTrivialTypeSourceInfo(DerivedOutputParamType,
+                                                        noLoc),
```
vgvassilev, line 387: Move into a local variable.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -946,20 +1149,35 @@ namespace clad {
       // done to reduce cloning complexity and only clone once. The type is same
       // as the call expression as it is the type used to declare the _gradX
       // array
-      Expr* dArg = StoreAndRef(nullptr, CEType, reverse, "_r", /*force*/ true);
+      Expr* dArg;
+      if (isVectorValued) {
+        dArg = StoreAndRef(nullptr, CEType, reverse, "_r", /*force*/ true);
```
vgvassilev, line 1154:
```suggestion
        dArg = StoreAndRef(/*E=*/nullptr, CEType, reverse, "_r", /*force=*/true);
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -946,20 +1149,35 @@ namespace clad {
...
+      } else if (isArrayOrPointerType(Arg->getType())) {
+        dArg =
+            StoreAndRef(m_Sema.ActOnCXXNullPtrLiteral(noLoc).get(),
```
vgvassilev, line 1157: Please move into a separate variable.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -946,20 +1149,35 @@ namespace clad {
...
+                        GetCladArrayRefOfType(CEType), reverse, "_r",
+                        /*force*/ true, VarDecl::InitializationStyle::CallInit);
```
vgvassilev, line 1159:
```suggestion
                        /*force=*/true, VarDecl::InitializationStyle::CallInit);
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -946,20 +1149,35 @@ namespace clad {
...
+      } else {
+        dArg = StoreAndRef(nullptr, CEType, reverse, "_r",
+                           /*force*/ true);
```
vgvassilev, line 1162:
```suggestion
        dArg = StoreAndRef(/*E=*/nullptr, CEType, reverse, "_r",
                           /*force=*/true);
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1267,35 +1556,39 @@ namespace clad {
...
+            std::unordered_map<const clang::VarDecl*, clang::Expr*>
+                temp_m_Variables;
+            for (unsigned i = 0; i < numParams; i++) {
+              auto size_type = m_Context.getSizeType();
+              unsigned size_type_bits = m_Context.getIntWidth(size_type);
+              auto idx =
+                  IntegerLiteral::Create(m_Context,
+                                         llvm::APInt(size_type_bits,
+                                                     i + (outputArrayCursor *
+                                                          numParams)),
```
vgvassilev, line 1580: Move in a separate variable.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1590,6 +1882,14 @@ namespace clad {
     return StmtDiff(Clone(ME));
   }
 
+  StmtDiff
+  ReverseModeVisitor::VisitExprWithCleanups(const ExprWithCleanups* EWC) {
+    StmtDiff subExprDiff = Visit(EWC->getSubExpr(), dfdx());
+    // FIXME: We are unable to create cleanup objects currently, this can be
+    //  potentially problematic
```
vgvassilev, line 1889:
```suggestion
    // potentially problematic
```

lib/Differentiator/VisitorBase.cpp
```diff
@@ -185,35 +185,31 @@ namespace clad {
...
-                              SC_None);
-
-    if (Init)
+  VarDecl* VisitorBase::BuildVarDecl(QualType Type, IdentifierInfo* Identifier,
+                                     Expr* Init, bool DirectInit,
+                                     TypeSourceInfo* TSI,
+                                     VarDecl::InitializationStyle IS) {
+
+    auto VD =
+        VarDecl::Create(m_Context, m_Sema.CurContext, m_Function->getLocation(),
```
vgvassilev, line 194: Why do we break here, wouldn't it be better to continue on the same line and break the args?

lib/Differentiator/VisitorBase.cpp
```diff
@@ -387,19 +382,18 @@ namespace clad {
...
     if (!forceDeclCreation) {
       // If Expr is simple (i.e. a reference or a literal), there is no point
       // in storing it as there is no evaluation going on.
       if (!UsefulToStore(E))
         return E;
     }
     // Create variable declaration.
-    VarDecl* Var = BuildVarDecl(Type, CreateUniqueIdentifier(prefix), E);
+    VarDecl* Var = BuildVarDecl(Type, CreateUniqueIdentifier(prefix), E, false,
+                                nullptr, IS);
```
vgvassilev, line 396:
```suggestion
    VarDecl* Var = BuildVarDecl(Type, CreateUniqueIdentifier(prefix), E, /*DirectInit=*/false,
                                /*TSI=*/nullptr, IS);
```

vgvassilev (commented):
Thanks for the updated PR. See some minor comments inline.

lib/Differentiator/VisitorBase.cpp
sudo-panda: I have made a change in `.clang-format` to be achieve this. Please check #270.

include/clad/Differentiator/FunctionTraits.h
sudo-panda: Not yet let me ask rn.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -972,39 +1190,102 @@ namespace clad {
...
+
+      unsigned size_type_bits = m_Context.getIntWidth(m_Context.getSizeType());
+      if (isVectorValued) {
+        // We also need to create an array to store the result of gradient call.
+        auto ArrayType = clad_compat::
+            getConstantArrayType(m_Context, CEType,
+                                 llvm::APInt(size_type_bits, NArgs),
+                                 /*SizeExpr=*/nullptr,
+                                 ArrayType::ArraySizeModifier::Normal,
+                                 0); // No IndexTypeQualifiers
```
vgvassilev, line 1212: Can you add a default argument comment here?

vgvassilev (approved):
Thanks for the patience. Great work!

sudo-panda:
> Thanks for the patience. Great work!

Thanks for reviewing the PR :)

PR #238, opened 2021-06-08
Match gcov version to compiler version in CI

sudo-panda:


vgvassilev:
@sudo-panda, I am starting to think we need something like https://github.com/codecov/codecov-action

vgvassilev:
There are some changes on the codecov side and that PR turned into #239.

PR #240, opened 2021-06-08
Add support for differentiating functors in forward mode

parth-07:
This PR aims to add support for differentiating functors in the forward differentiation mode.

- [x] Implicitly differentiate `operator()` member function when functor is passed
- [x] Implicitly pass functor while executing derived function using `CladFunction`
- [x] Add tests
- [x] Add user-friendly errors using `Sema` diagnostics when an object of class without overloaded `operator()` or multiple overloads of call operator is passed to be differentiated.
- [x] Add demo

By differentiating functors, I mean, differentiating the `operator()` member function when a functor is passed, and implicitly executing the derived function using the passed functor.

Functors can be passed to `clad::differentiate` both by reference and as pointers. For example:
```c++
Experiment E; // a functor
auto d_fn = clad::differentiate(&E, "i");
auto d_fnRef = clad::differentiate(E, "i");
...
...
std::cout<<d_fn.execute(1,  3)<<"\n"; // no need to pass object while executing derived function.
```
## Implemented solution

The solution is implemented in 2 steps:
- Differentiate `operator()` member function whenever functor (class type object) is passed to be differentiated.
- Execute derived function through the functor object.

**1st Step Implementation:**

DiffPlanner `getArgFunction` is responsible for obtaining function to be differentiated from the call expression of clad differentiation functions.

It currently behaves as follows:
- Return `DeclRefExpr` node of the 1st argument (which is, the function to be differentiated), and update the `RelevantAncestor` parameter to the nearest ancestor of 1st argument which is of type `ImplicitCastExpr` or `UnaryOperator`.

`getArgFunction` has been modified such that, whenever a class type object is to be differentiated, it provides all the data for the `operator()` member function of the class type. By all the data, I am referring to `DeclRefExpr` of the `operator()` method, the correct nearest ancestor for the `operator()` method.

But since the `operator()` method hasn't actually passed, these nodes (`DeclRefExpr` and ancestors of the `operator()` method) does not actually exist, and needs to be created inside `getArgFunction`. This effectively converts calls to differentiate class type object to calls to differentiate a member function for the rest of the clad.

**2nd Step Implementation:**

`CladFunction` has been modified to save an object through which derived member functions should be called when the user does not explicitly pass an object while calling `CladFunction::execute`

This behaviour is summarised in the following code snippet:

```c++
SomeClass A, B;
auto d_memFn = clad::differentiate(&SomeClass::SomeMemberFn, "i");
d_memFn.setObject(&A); // set A as default object for executing derived fn
d_memFn.execute(1, 3); // calls A.derivedFn(1, 3);
d_memFn.execute(B, 1, 3); // calls B.derivedFn(1, 3);
``` 
When a functor object is passed to be differentiated, it is saved in the corresponding `CladFunction` object so that it can be automatically taken by the `CladFunction::execute` method to call the derived member function.

**Modifications to `clad::differentiate`**

`clad::differentiate` is modified to have 2 overloads, one for differentiating functions (and methods), and the other for differentiating functors.
2 separate overloads are required because function (and method) pointers need to be passed by value whereas objects need to be passed by reference (because we need to determine `constness` property (among others) of passed functor object)

### DIFF
--- a/demos/Functor.cpp
+++ b/demos/Functor.cpp
@@ -1,0 +1,47 @@
+//--------------------------------------------------------------------*- C++ -*-
+// clad - The C++ Clang-based Automatic Differentiator
+//
+// A demo, describing how to differentiate a functor (an object of class
+// type with user defined call operator)
+//
+//----------------------------------------------------------------------------//
+
+// To run the demo please type:
+// path/to/clang  -Xclang -add-plugin -Xclang clad -Xclang -load -Xclang \
+// path/to/libclad.so  -I../include/ -x c++ -std=c++11 Functor.cpp
+//
+// A typical invocation would be:
+// ../../../../obj/Debug+Asserts/bin/clang  -Xclang -add-plugin -Xclang clad \
+// -Xclang -load -Xclang ../../../../obj/Debug+Asserts/lib/libclad.dylib     \
+// -I../include/ -x c++ -std=c++11 Functor.cpp
+
+// Necessary for clad to work include
+#include "clad/Differentiator/Differentiator.h"
+
+class Equation {
+  double m_x, m_y;
+
+  public:
+  Equation(double x, double y) : m_x(x), m_y(y) {}
+  double operator()(double i, double j) {
+    return m_x*i*j; + m_y*i*j;
+  }
+};
+
+int main() {
+  Equation E(3, 5);
+  
+  // Functor object can be passed in 2 ways:
+
+  // 1) by reference
+  auto d_E = clad::differentiate(E, "i");
+
+  // 2) as pointer
+  auto d_E_pointer = clad::differentiate(&E, "i");
+
+  // no need to explicitly pass functor object as the first argument.
+  double res1 = d_E.execute(7, 9);
+  double res2 = d_E_pointer.execute(7, 9);
+
+  printf("%.2f %.2f", res1, res2);
+}
--- a/demos/Functor.cpp
+++ b/demos/Functor.cpp
@@ -36,19 +36,59 @@ int main() {
   Equation E(3, 5);
 
   // Functor is an object of any type which have user defined call operator.
+  //
   // Clad differentiation functions can directly differentiate functors.
-  // Functors can be passed to clad differentiation functions in two distinct ways:
+  // Functors can be passed to clad differentiation functions in two distinct
+  // ways:
 
   // 1) Pass by reference
   // differentiates `E` wrt parameter `i`
+  // object `E` is saved in the `CladFunction` object `d_E`
   auto d_E = clad::differentiate(E, "i");
 
   // 2) Pass as pointers
   // differentiates `E` wrt parameter `i`
+  // object `E` is saved in the `CladFunction` object `d_E_pointer`
   auto d_E_pointer = clad::differentiate(&E, "i");
 
   // calculate differentiation of `E` when (i, j) = (7, 9)
   double res1 = d_E.execute(7, 9);
   double res2 = d_E_pointer.execute(7, 9);
+  printf("%.2f %.2f", res1, res2);
+
+  Equation AnotherE(11, 13);
+
+  // If differentiation of `E` is not needed, then `d_E` and `d_E_pointer` can
+  // be reused for computing differentiation of another functor of the same
+  // type by modifying the saved object in `CladFunction`. This can save
+  // computation time.
+  d_E.setObject(&AnotherE);
+  d_E_pointer.setObject(&AnotherE);
+
+  // calculate differentiation of `AnotherE` when (i, j) = (7, 9)
+  res1 = d_E.execute(7, 9);
+  res2 = d_E_pointer.execute(7, 9);
+  printf("%.2f %.2f", res1, res2);
+
+  // Differentiation of any other functor of the same type can also be computed
+  // without changing the saved object in `CladFunction` object by explicitly
+  // passing the functor as the first argument while calling
+  // `CladFunction::execute`.
+  // Calculates differentiation of `E` when (i, j) = (7, 9)
+  res1 = d_E.execute(E, 7, 9);
+  res2 = d_E_pointer.execute(E, 7, 9);
+  printf("%.2f %.2f", res1, res2);
+
+  // Saved object in `CladFunction` object can be removed using
+  // `CladFunction::clearObject`
+  d_E.clearObject();
+  d_E_pointer.clearObject();
+
+  // Now it is necessary to explicitly pass the functor object while calling
+  // `CladFunction::execute`, since no saved object is available.
+  // Calculates differentiation of `E` when (i, j) = (7, 9)
+  res1 = d_E.execute(E, 7, 9);
+  // Calculates differentiation of `AnotherE` when (i, j) = (7, 9)
+  res2 = d_E_pointer.execute(AnotherE, 7, 9);
   printf("%.2f %.2f", res1, res2);
 }
--- a/demos/Functor.cpp
+++ b/demos/Functor.cpp
@@ -18,6 +18,7 @@
 // Necessary for clad to work include
 #include "clad/Differentiator/Differentiator.h"
 
+// A class type with user-defined call operator
 class Equation {
   double m_x, m_y;
 
@@ -26,22 +27,28 @@ class Equation {
   double operator()(double i, double j) {
     return m_x*i*j; + m_y*i*j;
   }
+  void setX(double x) {
+    m_x = x;
+  }
 };
 
 int main() {
   Equation E(3, 5);
-  
-  // Functor object can be passed in 2 ways:
 
-  // 1) by reference
+  // Functor is an object of any type which have user defined call operator.
+  // Clad differentiation functions can directly differentiate functors.
+  // Functors can be passed to clad differentiation functions in two distinct ways:
+
+  // 1) Pass by reference
+  // differentiates `E` wrt parameter `i`
   auto d_E = clad::differentiate(E, "i");
 
-  // 2) as pointer
+  // 2) Pass as pointers
+  // differentiates `E` wrt parameter `i`
   auto d_E_pointer = clad::differentiate(&E, "i");
 
-  // no need to explicitly pass functor object as the first argument.
+  // calculate differentiation of `E` when (i, j) = (7, 9)
   double res1 = d_E.execute(7, 9);
   double res2 = d_E_pointer.execute(7, 9);
-
   printf("%.2f %.2f", res1, res2);
 }
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -157,10 +157,9 @@ namespace clad {
     }
     /// Constructor overload for initializing `m_Functor` when functor
     /// is passed by reference.
-    // ! Please confirm if `CUDA_HOST_DEVICE` should be added in the constructor overload.
     CUDA_HOST_DEVICE
     CladFunction(CladFunctionType f, const char* code, FunctorType& functor)
-        : CladFunction(f, code, &functor){};
+        : CladFunction(f, code, &functor) {};
 
     // Intentionally leak m_Code, otherwise we have to link against c++ runtime,
     // i.e -lstdc++.
@@ -209,7 +208,7 @@ namespace clad {
   /// please see `clad::gradient`. 
   /// \param[in] fn function to differentiate 
   /// \param[in] args independent parameter information 
-  /// \return `CladFunction` object to access the corresponding derived function.
+  /// \returns `CladFunction` object to access the corresponding derived function.
   template <unsigned N = 1,
             typename ArgSpec = const char*,
             typename F,
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -72,14 +72,15 @@ namespace clad {
   /// Helper function for executing non-member derived functions.
   /// RedundantType is only present to keep signature same as of execute_helper
   /// member function counterpart.
-  template<class F, class RedundantType, class... Args> 
-  return_type_t<F> execute_helper(F f, RedundantType* redundant, Args&&... args) {
-      return f(static_cast<Args>(args)... );
+  template <class F, class RedundantType, class... Args>
+  return_type_t<F>
+  execute_helper(F f, RedundantType* redundant, Args&&... args) {
+    return f(static_cast<Args>(args)...);
   }
 
   /// Helper functions for executing member derived functions.
-  /// If user have passed object explicitly, then this specialization will be used and 
-  /// derived fn will be called through the passed object.
+  /// If user have passed object explicitly, then this specialization will be
+  /// used and derived fn will be called through the passed object.
   template <class ReturnType,
             class C,
             class FunctorType,
@@ -97,8 +98,9 @@ namespace clad {
     return (static_cast<Obj>(obj).*f)(static_cast<Args>(args)...);
   }
 
-  // If user have not passed object explicitly, then this specialization will be used
-  // and derived fn will be called through the object saved in `CladFunction`.
+  /// If user have not passed object explicitly, then this specialization will
+  /// be used and derived fn will be called through the object saved in
+  /// `CladFunction`.
   template <class ReturnType,
             class C,
             class FunctorType,
@@ -108,6 +110,8 @@ namespace clad {
             class... Args>
   auto execute_helper(ReturnType C::*f, FunctorType* functor, Args&&... args)
       -> return_type_t<decltype(f)> {
+    assert(functor && "No default object set, explicitly pass an object to "
+                      "CladFunction::execute");
     return (functor->*f)(static_cast<Args>(args)...);
   }
 
@@ -118,7 +122,8 @@ namespace clad {
   /// const correctness of functor types.
   /// Default value of `Functor` here is temporary, and should be removed once all
   /// clad differentiation functions support differentiating functors.
-  template <typename F, typename FunctorT = ExtractFunctorTraits_t<F>> class CladFunction {
+  template <typename F, typename FunctorT = ExtractFunctorTraits_t<F>>
+  class CladFunction {
   public:
     using CladFunctionType = F;
     using FunctorType = FunctorT;
@@ -196,13 +201,14 @@ namespace clad {
   // This will be useful in fucture when we are ready to support partial diff.
   //
 
-  /// \brief Differentiates function using forward mode.
+  /// Differentiates function using forward mode.
   ///
-  /// Performs partial differentiation of the `fn` argument using forward mode wrt
-  /// parameter specified in `args`. To differentiate `fn` wrt several parameters, 
-  /// please see `clad::gradient`.
-  /// \param[in] fn function to differentiate
-  /// \param[in] args independent parameter information
+  /// Performs partial differentiation of the `fn` argument using forward mode
+  /// wrt parameter specified in `args`. Template parameter `N` denotes
+  /// the derivative order. To differentiate `fn` wrt several parameters,
+  /// please see `clad::gradient`. 
+  /// \param[in] fn function to differentiate 
+  /// \param[in] args independent parameter information 
   /// \return `CladFunction` object to access the corresponding derived function.
   template <unsigned N = 1,
             typename ArgSpec = const char*,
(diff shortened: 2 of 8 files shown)


include/clad/Differentiator/Differentiator.h
```diff
@@ -111,6 +155,12 @@ namespace clad {
         m_Code = nullptr;
       }
     }
+    /// Constructor overload for initializing `m_Functor` when functor
+    /// is passed by reference.
+    // ! Please confirm if `CUDA_HOST_DEVICE` should be added in the constructor overload.
```
vgvassilev, line 160: @ioanaif, could you confirm that's the way to go here?

include/clad/Differentiator/Differentiator.h
```diff
@@ -111,6 +155,12 @@ namespace clad {
...
+    CUDA_HOST_DEVICE
+    CladFunction(CladFunctionType f, const char* code, FunctorType& functor)
+        : CladFunction(f, code, &functor){};
```
vgvassilev, line 163:
```suggestion
        : CladFunction(f, code, &functor) {};
```

include/clad/Differentiator/FunctionTraits.h
```diff
@@ -349,6 +350,144 @@ namespace clad {
...
 #endif
+
+  /// Utility type trait to remove both reference and pointer
+  /// from type `T`.
+  ///
+  /// First removes any reference qualifier associated with `T`,
+  /// then removes any associated pointer. Resulting type is provided
+  /// as member typedef `type`.
+  template <typename T> struct remove_reference_and_pointer {
+    using type = typename std::remove_pointer<
```
vgvassilev, line 361: What about references to pointers?

include/clad/Differentiator/FunctionTraits.h
```diff
@@ -349,6 +350,144 @@ namespace clad {
...
+  struct ExtractDerivedFnTraitsForwMode<
+      F,
+      typename std::enable_if<
+          std::is_member_function_pointer<F>::value>::type> {
+    using type = typename std::decay<F>::type;
+  };
+
+  /// Specialization for class types
+  template <class F>
+  struct ExtractDerivedFnTraitsForwMode<
```
vgvassilev, line 427: Why we have `ForwMode` in the name? Why this trait is specific to forward mode? Can you add an explanation in a comment?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -14,20 +14,40 @@ using namespace clang;
 
 namespace clad {
   static SourceLocation noLoc;
-  
+
+  /// Find and returns overloaded call operator.
```
vgvassilev, line 18:
```suggestion
  /// Find and return an overloaded call operator.
```

include/clad/Differentiator/Differentiator.h
```diff
@@ -148,19 +201,47 @@ namespace clad {
...
-  ///\brief N is the derivative order.
+  /// Differentiates function using forward mode.
   ///
+  /// Performs partial differentiation of the `fn` argument using forward mode
+  /// wrt parameter specified in `args`. Template parameter `N` denotes
+  /// the derivative order. To differentiate `fn` wrt several parameters,
+  /// please see `clad::gradient`. 
+  /// \param[in] fn function to differentiate 
+  /// \param[in] args independent parameter information 
+  /// \return `CladFunction` object to access the corresponding derived function.
```
vgvassilev, line 212:
```suggestion
  /// \returns `CladFunction` object to access the corresponding derived function.
```

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -14,20 +14,40 @@ using namespace clang;
...
+  /// \param[in] SemaRef reference to sema
+  /// \param[in] RD
+  /// \returns found overloaded call operator
+  CXXMethodDecl* getOverloadedCallOperator(Sema& SemaRef, CXXRecordDecl* RD) {
+    auto callOperatorDeclName =
+        SemaRef.getASTContext().DeclarationNames.getCXXOperatorName(
+            OverloadedOperatorKind::OO_Call);
+    auto LR = RD->lookup(callOperatorDeclName);
+
+    assert(LR.size() == 1 && "Overloaded call operators "
```
vgvassilev, line 29: Could you add a comment explaining why we are guaranteed to have 1 overloaded call operator. Eg. why it can't be 0.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -14,20 +14,40 @@ using namespace clang;
...
   // If argument is passed for relevantAncestor, then it's value will be 
   // modified in-place to contain relevant ancestor of the function 
   // argument's DeclRefExpr
   // Here relevant ancestor is nearest ancestor of ImplicitCastExpr or 
   // UnaryOperator type.
   DeclRefExpr* getArgFunction(CallExpr* call,
+                              Sema& SemaRef,
                               Expr** relevantAncestor = nullptr) {
     struct Finder :
       RecursiveASTVisitor<Finder> {
```
vgvassilev, line 45: Why that is not on the previous line?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -38,15 +58,55 @@ namespace clad {
             }
           }
           if (auto DRE = dyn_cast<DeclRefExpr>(E)) {
-            if (auto VD = dyn_cast<VarDecl>(DRE->getDecl()))
-              TraverseStmt(VD->getInit());
-            else if (auto FD = dyn_cast<FunctionDecl>(DRE->getDecl()))
+            if (auto VD = dyn_cast<VarDecl>(DRE->getDecl())) {
+              auto varType = VD->getType().getTypePtr();
+              /// If variable is of class type, set `m_FnDRE` to
```
vgvassilev, line 63: No need of doxygen style comments in the function bodies.

test/ForwardMode/Functors.C
```diff
@@ -0,0 +1,210 @@
...
+  EXECUTE(d_ERef, 1, 3);  // CHECK-EXEC: 121.00
+  EXECUTE(d_NSE, 1, 3);   // CHECK-EXEC: 289.00
+  EXECUTE(d_NSERef, 1, 3);// CHECK-EXEC: 289.00
+
+  EXECUTE(d_W, 1, 3, 5);          // CHECK-EXEC: 9.00
+  EXECUTE(d_WRef, 1, 3, 5);       // CHECK-EXEC: 9.00
+  EXECUTE(d_constW, 1, 3, 5);     // CHECK-EXEC: 9.00
+  EXECUTE(d_constWRef, 1, 3, 5);  // CHECK-EXEC: 9.00
+  EXECUTE(d_NSW, 1, 3, 5);        // CHECK-EXEC: 121.00
+  EXECUTE(d_NSWRef, 1, 3, 5);     // CHECK-EXEC: 121.00
```
vgvassilev, line 209: Can you add a case of a class which does not have an overloaded call operator and two call operators?

include/clad/Differentiator/Differentiator.h
ioanaif: Hi! Yes, it is.

include/clad/Differentiator/FunctionTraits.h
parth-07: This type trait will define member typedef as `T` for types `T*`, `T&` and `T*&` so references to pointers are covered.

include/clad/Differentiator/FunctionTraits.h
parth-07: It is because the rules for computing the signature of derived functions are different in forward and reverse mode, I will add an explanation for this in the comments.

lib/Differentiator/DiffPlanner.cpp
parth-07: Okay

lib/Differentiator/DiffPlanner.cpp
parth-07: This is what clang-format suggested, I will look into clang-format configurations for this.

include/clad/Differentiator/Differentiator.h
parth-07: The clang-format suggests removing space between constructor signature and constructor initializer list. I am adding the space, and I will also check regarding configuration in clang-format to add space before the constructor initializer list.

lib/Differentiator/DiffPlanner.cpp
parth-07: done

test/ForwardMode/Functors.C
parth-07: The current solution does not support multiple call operators, that's why class types with only a single call operator are present in tests.

lib/Differentiator/DiffPlanner.cpp
parth-07: Done.

test/ForwardMode/Functors.C
vgvassilev: But it should be able to complain if there is no call operator. Do we have a test for this?

parth-07: No, I will add a test for this.

parth-07: tests added for class with no call operator defined and class with multiple call operators defined.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -109,22 +112,23 @@ namespace clad {
           // decomposed to function pointers and thus a temporary is
           // created for the function pointer.
           auto newFnDRE = clad_compat::GetResult<Expr*>(
-              m_SemaRef.BuildDeclRefExpr(callOperator,
-                                         callOperator->getType(),
-                                         ExprValueKind::VK_RValue,
-                                         noLoc,
-                                         &CSS));
+              m_SemaRef.BuildDeclRefExpr(/*D=*/callOperator,
```
vgvassilev, line 115: We don't need the parameter names for each parameter but just for things like constants, `nullptr` which are hard to guess because they are not passed in a variable-like style.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -21,38 +21,123 @@ namespace clad {
...
+            const char diagFmt[] =
+                "No call operator is defined for the class %0";
+            auto diagId =
+                m_SemaRef.Diags.getCustomDiagID(DiagnosticsEngine::Level::Error,
+                                                diagFmt);
+            m_SemaRef.Diag(m_Root->getBeginLoc(), diagId) << RD->getName();
+          } else if (LR.size() > 1) {
+            const char diagFmt[] =
+                "Class %0 defines multiple overloads of call operator. "
+                "Multiple overloads of call operators are not supported.";
```
vgvassilev, line 88:
```suggestion
                "'%0' has multiple definitions of operator(). "
                "Multiple definitions of call operators are not supported.";
```

We should probably issue a note for each definition -- see how it is done in clang on ambiguous overload calls.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -21,38 +21,123 @@ namespace clad {
...
+            auto diagId =
+                m_SemaRef.Diags.getCustomDiagID(DiagnosticsEngine::Level::Error,
+                                                diagFmt);
+            m_SemaRef.Diag(m_Root->getBeginLoc(), diagId) << RD->getName();
           }
-          return true; 
+
+          // Multiple overloads of call operators are not supported,
+          // thus class should have exactly one user-defined call operator.
+          if (LR.size() != 1) {
```
vgvassilev, line 97: We should have used an early exit after we issue the diagnostic and turn this into an assert.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -21,38 +21,123 @@ namespace clad {
...
+          } else if (auto FD = dyn_cast<FunctionDecl>(DRE->getDecl()))
+            m_FnDRE = DRE;
+          return false;
+        }
+
+        bool VisitCXXRecordDecl(CXXRecordDecl* RD) {
+          auto callOperatorDeclName =
+              m_SemaRef.getASTContext().DeclarationNames.getCXXOperatorName(
+                  OverloadedOperatorKind::OO_Call);
+          auto LR = RD->lookup(callOperatorDeclName);
```
vgvassilev, line 75: Is that what clang uses to lookup a call operator?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -21,38 +21,123 @@ namespace clad {
   // argument's DeclRefExpr
   // Here relevant ancestor is nearest ancestor of ImplicitCastExpr or 
   // UnaryOperator type.
-  DeclRefExpr* getArgFunction(CallExpr* call,
+  DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef,
```
vgvassilev, line 24: Does the documentation still describe what this function does?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -21,38 +21,123 @@ namespace clad {
...
+          auto LR = RD->lookup(callOperatorDeclName);
+
+          // Emit error diagnostics
+          if (LR.empty()) {
+            const char diagFmt[] =
+                "No call operator is defined for the class %0";
+            auto diagId =
+                m_SemaRef.Diags.getCustomDiagID(DiagnosticsEngine::Level::Error,
+                                                diagFmt);
+            m_SemaRef.Diag(m_Root->getBeginLoc(), diagId) << RD->getName();
```
vgvassilev, line 84: Why can't we use the `RD`'s location and drop the need for `m_Root`?

lib/Differentiator/DiffPlanner.cpp
parth-07: Okay.

include/clad/Differentiator/Differentiator.h
```diff
@@ -69,31 +69,75 @@ namespace clad {
...
+                std::is_same<typename std::remove_cv<FunctorType>::type,
+                             C>::value>::type,
+            class = typename std::enable_if<
+                std::is_same<typename std::decay<Obj>::type, C>::value>::type,
+            class... Args>
+  auto execute_helper(ReturnType C::*f,
+                      FunctorType* functor,
+                      Obj&& obj,
+                      Args&&... args) -> return_type_t<decltype(f)> {
+    return (static_cast<Obj>(obj).*f)(static_cast<Args>(args)...);
```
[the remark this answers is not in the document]
parth-07: It is for doing perfect forwarding, I will add a comment explaining this.

lib/Differentiator/DiffPlanner.cpp
parth-07: Okay.

lib/Differentiator/DiffPlanner.cpp
parth-07: I am not sure, what's the ideal way of checking how clang lookup up an operator?

lib/Differentiator/DiffPlanner.cpp
parth-07: RD's location is the location of the type definition, we need the error squiggle on object argument in the call to `clad::differentiate`

lib/Differentiator/DiffPlanner.cpp
parth-07: No, it needs to be updated. I will update this.

lib/Differentiator/DiffPlanner.cpp
vgvassilev: Then why don't we pass only that source location.

lib/Differentiator/DiffPlanner.cpp
vgvassilev: The usual approach: you can create a two line example, run clang on it and break at `DeclContext::lookup` to see the call stack.

lib/Differentiator/DiffPlanner.cpp
parth-07: Okay, I will see how it is done in clang on ambiguous overload calls, and try to add it here.

lib/Differentiator/DiffPlanner.cpp
parth-07: Okay :), I will check how clang handles this.

lib/Differentiator/DiffPlanner.cpp
parth-07: Yes, that seems better, I will modify the code to just pass that source location

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -21,38 +21,123 @@ namespace clad {
...
+        bool VisitCXXRecordDecl(CXXRecordDecl* RD) {
+          auto callOperatorDeclName =
+              m_SemaRef.getASTContext().DeclarationNames.getCXXOperatorName(
+                  OverloadedOperatorKind::OO_Call);
+          auto LR = RD->lookup(callOperatorDeclName);
+
+          // Emit error diagnostics
+          if (LR.empty()) {
+            const char diagFmt[] =
+                "No call operator is defined for the class %0";
```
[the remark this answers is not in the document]
parth-07: done.

lib/Differentiator/DiffPlanner.cpp
parth-07: Done, now clad issues a note for each candidate function.

lib/Differentiator/DiffPlanner.cpp
parth-07: clang uses `Sema::LookupQualifiedName` which internally uses `DeclContext::lookup` for call operator lookup. Should we use `Sema::LookupQualifiedName` instead of `DeclContext::lookup` ?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -101,43 +101,70 @@ namespace clad {
               m_SemaRef.NoteOverloadCandidate(candidateFn,
                                               cast<FunctionDecl>(candidateFn));
             return false;
+          } else if (LR.size() == 1 &&
+                     cast<CXXMethodDecl>(LR.front())->getAccess() !=
+                         AccessSpecifier::AS_public) {
+            const char diagFmt[] =
```
vgvassilev, line 107: We should re-use the diagnostics from Sema which checks for visibility.

include/clad/Differentiator/Differentiator.h
```diff
@@ -69,31 +69,29 @@ namespace clad {
...
-      return (static_cast<Obj>(obj).*f)(static_cast<Args>(args)... );
-  }
-
   // Using std::function and std::mem_fn introduces a lot of overhead, which we
   // do not need. Another disadvantage is that it is difficult to distinguish a
   // 'normal' use of std::{function,mem_fn} from the ones we must differentiate.
-  template <typename F> class CladFunction {
+  /// Explicitly passing `FunctorT` type is necessary for maintaining 
+  /// const correctness of functor types.
+  /// Default value of `Functor` here is temporary, and should be removed
```
vgvassilev, line 77: `removed` or `replaced`?

demos/Functor.cpp
```diff
@@ -0,0 +1,54 @@
...
+  auto d_E = clad::differentiate(E, "i");
+
+  // 2) Pass as pointers
+  // differentiates `E` wrt parameter `i`
+  auto d_E_pointer = clad::differentiate(&E, "i");
+
+  // calculate differentiation of `E` when (i, j) = (7, 9)
+  double res1 = d_E.execute(7, 9);
+  double res2 = d_E_pointer.execute(7, 9);
+  printf("%.2f %.2f", res1, res2);
```
vgvassilev, line 53: Can we demonstrate the usage of `setObject` and `clearObject` as well?

lib/Differentiator/DiffPlanner.cpp
vgvassilev: Yes, I'd prefer so.

include/clad/Differentiator/Differentiator.h
parth-07: The default value should be removed from `Cladfunction` once all clad differentiation functions support functors since the functor type will then be computed in the corresponding clad differentiation function.

demos/Functor.cpp
parth-07: Yes, I will modify the demo to include the demonstration of `setObject` and `clearObject`.

lib/Differentiator/DiffPlanner.cpp
parth-07: Done

lib/Differentiator/DiffPlanner.cpp
parth-07: Done, now we are using the `Sema::LookupQualifiedName` function to lookup call operators.

vgvassilev (approved):
Well done! LGTM!

PR #243, opened 2021-06-09
Remove dependency of how the fn is passed to clad diff functions

parth-07:
Currently clad update calls like these:
```c++
auto d_fn_1 = clad::differentiate(freeFn, ...);
auto d_fn_2 = clad::differentiate(&freeFn, ...);
auto d_memFn = clad::differentiate(&SomeClass::SomeMethod, ...);
```
to
```c++
auto d_fn_1 = clad::differentiate(derivedFn, ...);
auto d_fn_2 = clad::differentiate(&derivedFn, ...);
auto d_memFn = clad::differentiate(&SomeClass::SomeMethodDerivedFn, ...);
```

While updating arguments, `&` is only added to the derived fn argument, if `&` was originally present. To achieve this behaviour clad have to find the relevant parent of fn argument `DeclRefExpr`, then act differently depending on if the relevant parent found was `UnaryOperator` or some cast. 

We can instead always add `&` with the derived fn argument. 
This way we will not need to keep track of whether the most recent parent was `ImplicitCastExpr` or `UnaryOperator`, we will not need to differentiate the 2 cases. And this will have no other change in AST other than `ImplicitCastExpr` changed to `UnaryOperator` in fn arguments of clad differentiation calls, since as per clad implementation, function types are always decomposed to function pointer types while passing them to clad differentiation functions.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -20,43 +20,19 @@ namespace clad {
   /// Returns `DeclRefExpr` node corresponding to the function, method or
   /// functor argument which is to be differentiated.
   ///
-  /// Argument corresponding to `relevantAncestor` parameter will be updated to
-  /// have value of type `UnaryOperator` if `&` (address of) operator was
-  /// explicitly used to pass the function, otherwise it will have value of type
-  /// `ImplicitCastExpr`.
-  /// Arugment for `relevantAncestor` only needs to be passed if this
-  /// information is needed.
   /// \param[in] call A clad differentiation function call expression
   /// \param SemaRef Reference to Sema
-  /// \param[out] relevantAncestor
-  DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef,
-                              Expr** relevantAncestor = nullptr) {
+  DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
     struct Finder :
       RecursiveASTVisitor<Finder> {
         DeclRefExpr* m_FnDRE = nullptr;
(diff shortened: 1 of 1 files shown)


vgvassilev (approved):
Good job! Can you squash the two commits and copy the PR description as a commit message? I guess we cannot make codecov happy here for some reason but that's ok.

parth-07:
@vgvassilev Thank you for approving the PR, but if we decide to go with passing function/functor argument by universal reference in my other PR ([#240](https://github.com/vgvassilev/clad/pull/240)) then that will do similar work as done here and will overwrite the work done in this PR. Should we wait to merge this PR until we come to a conclusion regarding passing function/functor argument by universal reference or not ?

vgvassilev:
@parth-07, okay, let's wait. Could you update the description of the other PR to capture the design, intent and the proposed solution?

parth-07:
@vgvassilev I have updated the PR description to include more details.

vgvassilev:
@parth-07, what is the fate of this PR?

parth-07:
Its fate depends on one design decision that we have to decide in PR #240. That design decision is, should we pass function/functor by universal reference or not. We need to pass functors by reference otherwise we cannot detect `constness` attribute. Should I convert this PR to draft in the meantime? 

Given the recent developments, PR #240 needs some modifications and can be made simpler. I will make those modifications and rebase #240 on top of the master branch so it will be easier to review.

vgvassilev (approved):
LGTM!

vgvassilev:
The codecov complaint is coming from removing code which reduces the coverage compared to previous master but that's what we want -- less code.

PR #244, opened 2021-06-09
Add array diff to forward mode

sudo-panda:
`parseDiffArgs` now supports specifying the elements w.r.t. to which the function is to be differentiated.

`ForwardModeVisitor` can now differentiate w.r.t. to any input array element. To support array diff forward mode stores the index of the requested array. It then checks the `ArraySubscriptExpr` being differentiated and if it is equal to the array element requested it returns `1`.

N.B. More tests need to be added before this is merged

### DIFF
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -37,6 +37,25 @@ namespace clad {
     class CladPlugin;
     clang::FunctionDecl* ProcessDiffRequest(CladPlugin& P, DiffRequest& request);
   }
+
+  struct IndexInterval {
+    size_t Start;
+    size_t Finish;
+
+    IndexInterval() : Start(0), Finish(0) {}
+
+    IndexInterval(size_t first, size_t last) : Start(first), Finish(last + 1) {}
+
+    IndexInterval(size_t index) : Start(index), Finish(index + 1) {}
+
+    size_t size() {
+      return Finish - Start;
+    }
+
+    bool isInInterval(size_t n) {
+      return n >= Start && n <= Finish;
+    }
+  };
 }
 
 namespace clad {
@@ -44,8 +63,11 @@ namespace clad {
   // in nested namespaces
   using DeclWithContext = std::pair<clang::FunctionDecl*, clang::Decl*>;
   using DiffParams = llvm::SmallVector<const clang::VarDecl*, 16>;
+  using IndexIntervalTable = llvm::SmallVector<IndexInterval, 16>;
+  using DiffParamsWithIndices = std::pair<DiffParams, IndexIntervalTable>;
 
-  using VectorOutputs = std::vector<std::unordered_map<const clang::VarDecl*, clang::Expr*>>;
+  using VectorOutputs =
+      std::vector<std::unordered_map<const clang::VarDecl*, clang::Expr*>>;
 
   static clang::SourceLocation noLoc{};
   class VisitorBase;
--- a/include/clad/Differentiator/ForwardModeVisitor.h
+++ b/include/clad/Differentiator/ForwardModeVisitor.h
@@ -25,6 +25,7 @@ namespace clad {
         public VisitorBase {
   private:
     const clang::VarDecl* m_IndependentVar = nullptr;
+    unsigned m_IndependentVarIndex = ~0;
     unsigned m_DerivativeOrder = ~0;
     unsigned m_ArgIndex = ~0;
 
--- a/lib/Differentiator/ForwardModeVisitor.cpp
+++ b/lib/Differentiator/ForwardModeVisitor.cpp
@@ -57,8 +57,7 @@ namespace clad {
       return {};
     // Check that only one arg is requested and if the arg requested is of array
     // or pointer type, only one of the indices have been requested
-    if (args.size() > 1 || ((args[0]->getType()->isArrayType() ||
-                             args[0]->getType()->isPointerType()) &&
+    if (args.size() > 1 || (isArrayOrPointerType(args[0]->getType()) &&
                             (indexIntervalTable.size() != 1 || indexIntervalTable[0].size() != 1))) {
       diag(DiagnosticsEngine::Error,
            request.Args ? request.Args->getEndLoc() : noLoc,
@@ -70,10 +69,24 @@ namespace clad {
     }
 
     m_IndependentVar = args.back();
-    // If param is not real (i.e. floating point or integral), we cannot
-    // differentiate it.
+    std::string derivativeSuffix("");
+    // If param is not real (i.e. floating point or integral), a pointer to a
+    // real type, or an array of a real type we cannot differentiate it.
     // FIXME: we should support custom numeric types in the future.
-    if (!m_IndependentVar->getType()->isRealType()) {
+    if (isArrayOrPointerType(m_IndependentVar->getType())) {
+      if (!m_IndependentVar->getType()
+               ->getPointeeOrArrayElementType()
+               ->isRealType()) {
+        diag(DiagnosticsEngine::Error,
+             m_IndependentVar->getEndLoc(),
+             "attempted differentiation w.r.t. a parameter ('%0') which is not"
+             " an array or pointer of a real type",
+             {m_IndependentVar->getNameAsString()});
+        return {};
+      }
+      m_IndependentVarIndex = indexIntervalTable[0].Start;
+      derivativeSuffix = "_" + std::to_string(m_IndependentVarIndex);
+    } else if (!m_IndependentVar->getType()->isRealType()) {
       diag(DiagnosticsEngine::Error,
            m_IndependentVar->getEndLoc(),
            "attempted differentiation w.r.t. a parameter ('%0') which is not "
@@ -94,8 +107,9 @@ namespace clad {
     m_ArgIndex = std::distance(
         FD->param_begin(),
         std::find(FD->param_begin(), FD->param_end(), m_IndependentVar));
-    IdentifierInfo* II = &m_Context.Idents.get(
-        derivativeBaseName + "_d" + s + "arg" + std::to_string(m_ArgIndex));
+    IdentifierInfo* II =
+        &m_Context.Idents.get(derivativeBaseName + "_d" + s + "arg" +
+                              std::to_string(m_ArgIndex) + derivativeSuffix);
     DeclarationNameInfo name(II, noLoc);
     llvm::SaveAndRestore<DeclContext*> SaveContext(m_Sema.CurContext);
     llvm::SaveAndRestore<Scope*> SaveScope(m_CurScope);
@@ -500,6 +514,25 @@ namespace clad {
     if (!isa<VarDecl>(DRE->getDecl()))
       return StmtDiff(cloned, zero);
     auto VD = cast<VarDecl>(DRE->getDecl());
+    if (VD == m_IndependentVar) {
+      llvm::APSInt index;
+      Expr* diffExpr = nullptr;
+
+      if (!clad_compat::Expr_EvaluateAsInt(
+              clonedIndices.back(), index, m_Context)) {
+        diffExpr = BuildParens(
+            BuildOp(BO_EQ,
+                    clonedIndices.back(),
+                    ConstantFolder::synthesizeLiteral(
+                        m_Context.IntTy, m_Context, m_IndependentVarIndex)));
+      } else if (index.getExtValue() == m_IndependentVarIndex) {
+        diffExpr =
+            ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 1);
+      } else {
+        diffExpr = zero;
+      }
+      return StmtDiff(cloned, diffExpr);
+    }
     // Check DeclRefExpr is a reference to an independent variable.
     auto it = m_Variables.find(VD);
     if (it == std::end(m_Variables))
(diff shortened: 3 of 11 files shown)


include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -44,8 +44,11 @@ namespace clad {
   // in nested namespaces
   using DeclWithContext = std::pair<clang::FunctionDecl*, clang::Decl*>;
   using DiffParams = llvm::SmallVector<const clang::VarDecl*, 16>;
+  using IndexTable = llvm::SmallVector<std::set<size_t>, 16>;
```
vgvassilev, line 47: It feels a little awkward to model with a set of indices. How about replacing the `std::set` with a index interval (eg. some sort of custom IndexRange struct)?

sudo-panda: @vgvassilev correct me if I am wrong but by index nterval do you mean for requested indexes `1, 2, 3, 7, 8`. The data structure will store `1, 4, 7, 9` (`1,4` representing `1, 2, 3` and `7,9` representing `7, 8`)

vgvassilev: For now I think we can assume a single contiguous interval. Later, upon request, we can add others.

sudo-panda: Alright!

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -37,15 +37,37 @@ namespace clad {
...
+  };
 }
 
 namespace clad {
   /// A pair of FunctionDecl and potential enclosing context, e.g. a function
   // in nested namespaces
   using DeclWithContext = std::pair<clang::FunctionDecl*, clang::Decl*>;
   using DiffParams = llvm::SmallVector<const clang::VarDecl*, 16>;
+  using IndexIntervalTable = llvm::SmallVector<IndexInterval, 16>;
+  using DiffParamsWithIndexes = std::pair<DiffParams, IndexIntervalTable>;
```
vgvassilev, line 67:
```suggestion
  using DiffParamsWithIndices = std::pair<DiffParams, IndexIntervalTable>;
```

include/clad/Differentiator/ForwardModeVisitor.h
```diff
@@ -25,6 +25,7 @@ namespace clad {
         public VisitorBase {
   private:
     const clang::VarDecl* m_IndependentVar = nullptr;
+    size_t m_IndependentVarIndex = -1;
```
vgvassilev, line 28: Why not make it a bit more consistent to the surrounding code. How about `unsigned m_IndependentVarIndex = ~0;`

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -44,19 +44,23 @@ namespace clad {
...
+    IndexIntervalTable indexIntervalTable{};
     if (request.Args)
-      args = parseDiffArgs(request.Args, FD);
+      std::tie(args, indexIntervalTable) = parseDiffArgs(request.Args, FD);
     else {
       // FIXME: implement gradient-vector products to fix the issue.
       assert((FD->getNumParams() <= 1) &&
              "nested forward mode differentiation for several args is broken");
       std::copy(FD->param_begin(), FD->param_end(), std::back_inserter(args));
+      indexIntervalTable.push_back(IndexInterval());
```
vgvassilev, line 55: Why we need to push an 'empty' argument here. Can't we later check if `indexIntervalTable.empty`?

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -65,10 +69,25 @@ namespace clad {
...
     m_IndependentVar = args.back();
-    // If param is not real (i.e. floating point or integral), we cannot
-    // differentiate it.
+    std::string derivativeSuffix("");
+    // If param is not real (i.e. floating point or integral), a pointer to a
+    // real type, or an array of a real type we cannot differentiate it.
     // FIXME: we should support custom numeric types in the future.
-    if (!m_IndependentVar->getType()->isRealType()) {
+    if (m_IndependentVar->getType()->isPointerType() ||
+        m_IndependentVar->getType()->isArrayType()) {
```
vgvassilev, line 77: Can we move this repeating code into a utility function?

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -37,15 +37,37 @@ namespace clad {
     class CladPlugin;
     clang::FunctionDecl* ProcessDiffRequest(CladPlugin& P, DiffRequest& request);
   }
+
+  struct IndexInterval {
+    size_t begin;
+    size_t end;
```
vgvassilev, line 43: That can be confused with the iterators. How about maybe first/last or something else...

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -495,6 +515,25 @@ namespace clad {
...
+      llvm::APSInt index;
+      if (!clad_compat::Expr_EvaluateAsInt(
+              clonedIndices.back(), index, m_Context)) {
+        return StmtDiff(cloned,
+                        BuildParens(BuildOp(BO_EQ,
+                                            clonedIndices.back(),
+                                            ConstantFolder::synthesizeLiteral(
+                                                m_Context.IntTy,
+                                                m_Context,
+                                                m_IndependentVarIndex))));
```
vgvassilev, line 528: Can you use a temporary here and pass it to the StmtDiff?

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -495,6 +515,25 @@ namespace clad {
     if (!isa<VarDecl>(DRE->getDecl()))
       return StmtDiff(cloned, zero);
     auto VD = cast<VarDecl>(DRE->getDecl());
+    if (VD == m_IndependentVar && Indices.size() == 1) {
```
vgvassilev, line 518: Can the `Indices.size()` be different than 1?

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -495,6 +515,25 @@ namespace clad {
...
+                        BuildParens(BuildOp(BO_EQ,
+                                            clonedIndices.back(),
+                                            ConstantFolder::synthesizeLiteral(
+                                                m_Context.IntTy,
+                                                m_Context,
+                                                m_IndependentVarIndex))));
+      } else if (index.getExtValue() == m_IndependentVarIndex) {
+        return StmtDiff(
+            cloned,
+            ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 1));
```
vgvassilev, line 532: We should probably move that interface in the VisitorBase as it reads awkward.

lib/Differentiator/VisitorBase.cpp
```diff
@@ -60,40 +61,61 @@ namespace clad {
       for (auto PVD : FD->parameters())
         param_names_map.emplace_back(PVD->getName(), PVD);
       for (const auto& name : names) {
+        int loc = name.find('[');
+        loc = (loc == name.npos) ? name.size() : loc;
```
vgvassilev, line 65: I think the loop variable can be declared as StringRef which would give us clearer and handier string manipulation unity.

lib/Differentiator/VisitorBase.cpp
```diff
@@ -60,40 +61,61 @@ namespace clad {
...
-      DiffParams unique_params{};
-      for (const auto param : params) {
-        auto it = std::find(std::begin(unique_params),
-                            std::end(unique_params),
-                            param);
-        if (it != std::end(unique_params)) {
+
+        auto f_it = std::find(std::begin(params), std::end(params), it->second);
+
+        if (f_it == params.end()) {
```
vgvassilev, line 87: How about reducing the cyclomatic complexity by considering the inverse condition, diagnosing and returning; and removing the if?

lib/Differentiator/VisitorBase.cpp
```diff
@@ -60,40 +61,61 @@ namespace clad {
...
+          params.push_back(it->second);
+
+          if (loc != name.size()) {
+            llvm::StringRef interval(
+                name.substr(loc + 1, name.find(']') - loc - 1));
+            unsigned loc = interval.find("-");
```
vgvassilev, line 93: I am not sure if we shouldn't use [1:2] as that's not (yet?) valid C++ and can signify better that's something specific to clad..

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -34,8 +34,9 @@ namespace clad {
   DeclWithContext HessianModeVisitor::Derive(const clang::FunctionDecl* FD,
                                              const DiffRequest& request) {
     DiffParams args{};
+    IndexIntervalTable indexes{};
```
vgvassilev, line 37:
```suggestion
    IndexIntervalTable indices{};
```

vgvassilev (commented):
Thanks for working on this. As the PR description says we will need to add a few more tests. I have added some minor nits.

include/clad/Differentiator/ForwardModeVisitor.h
sudo-panda: Sure!

lib/Differentiator/ForwardModeVisitor.cpp
sudo-panda: Good point!

lib/Differentiator/ForwardModeVisitor.cpp
sudo-panda: It can be in case of multi dimensional array but probably not here

lib/Differentiator/ForwardModeVisitor.cpp
sudo-panda: Could you please elaborate?

lib/Differentiator/VisitorBase.cpp
sudo-panda: Could you please tell me which loop variable are you referring to?

lib/Differentiator/VisitorBase.cpp
sudo-panda: I have changed it to ":" because I feel it's better to be closer to C++ syntax (even though its not valid yet, it will be more intuitive)

include/clad/Differentiator/DerivativeBuilder.h
sudo-panda: I have renamed them to start and finish.

lib/Differentiator/ForwardModeVisitor.cpp
vgvassilev: How about turning it into an assert for now?

lib/Differentiator/ForwardModeVisitor.cpp
vgvassilev: `ConstantFolder::synthesizeLiteral` reads awkward. Maybe add a routine `VisitorBase::synthesizeLiteral` which for now calls `ConstantFolder::synthesizeLiteral`.

lib/Differentiator/VisitorBase.cpp
vgvassilev: I added a suggestion.

lib/Differentiator/VisitorBase.cpp
```diff
@@ -60,40 +61,69 @@ namespace clad {
       for (auto PVD : FD->parameters())
         param_names_map.emplace_back(PVD->getName(), PVD);
       for (const auto& name : names) {
```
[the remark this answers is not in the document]
sudo-panda: Isn't `llvm::StringRef` being deduced automatically?

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -54,9 +55,10 @@ namespace clad {
     }
     if (args.empty())
       return {};
-    if (args.size() > 1) {
+    if (args.size() > 1 || (isArrayOrPointer(args[0]) &&
+                            (indexIntervalTable.size() != 1 || indexIntervalTable[0].size() != 1))) {
```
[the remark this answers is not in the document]
sudo-panda: Forward mode should throw an error if exactly one array index isn't specified.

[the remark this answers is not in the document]
vgvassilev: Ah, ok, let's add a comment there then.

lib/Differentiator/VisitorBase.cpp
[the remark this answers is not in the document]
vgvassilev: Ah, overlooked, sorry.

test/Arrays/ArrayInputsForwardMode.C
```diff
@@ -0,0 +1,56 @@
...
+//CHECK-NEXT:               _d_ret += (i == 1);
+//CHECK-NEXT:               ret += arr[i];
+//CHECK-NEXT:           }
+//CHECK-NEXT:       }
+//CHECK-NEXT:       return _d_ret;
+//CHECK-NEXT:   }
+
+int main() {
+  double arr[] = {1, 2, 3, 4, 5};
+  auto multiply_dx = clad::differentiate(multiply, "arr[1]");
```
vgvassilev, line 48: I realize that ensuring we refer to the correct argument might be a problem. Eg. we should complain if the parameter name we specified is different from the one in the differentiated function.

 Consider:
`double multiply(double*); double multiply(double* p); double multiply(double* arr) {}` is valid C++. I think we should require (and document it) that the argument should match only at the definition (the FD that we will transform).

test/Arrays/ArrayErrorsForwardMode.C
```diff
@@ -0,0 +1,33 @@
+// RUN: not %cladclang %s -I%S/../../include -oFunctionCallsWithResults.out 2>&1 | FileCheck %s
+// REQUIRES: lit-not
```
vgvassilev, line 2: I think we can use -verify here for checking on the diagnostics. You can take a look at the other tests.

vgvassilev (approved):
We seem to be almost good to go. Please squash all commits into one with a proper commit log.

PR #249, opened 2021-06-17
Preserve original member function qualifiers in the derived function.

### LINKED ISSUE #210: function qualifiers of derived functions in reverse differentiation is inconsistent with function qualifiers of derivied functions in forward differentiation
In forward differentiation mode, derived function have same qualifiers(`const`,  `volatile`, `reference qualifiers`, `noexcept`) as the member function to be differentiated. 
Whereas, in reverse differentiation mode, derived function ignores all the qualifiers, and thus derived function have same type for both the functions with qualifiers and without qualifiers (analysed using `-ast-dump` and `-fdump-derived-fn`)

Users of the library will expect consistent behaviour for this in the library, and derived function having same qualifiers as member function, will also allow users to exercise more control over the type of derived function.

parth-07:
When differentiating member functions using `clad::gradient`, member function qualifiers (`const`, `volatile`, `noexcept`, `&`, `&&`) were not preserved. For example, differentiating,

```c++
double const_volatile_mem_fn(double i, double j) const volatile;
```

will create the gradient function with the following prototype.

```c++
void const_volatile_mem_fn_grad(double i, double j, double *_result); // const volatile qualifiers lost
```

This PR fixes this behaviour, and the gradient function will now have all the qualifiers of the original function.

Benefits of preserving member function qualifiers
- Increased consistency in the library, since differentiation using `clad::differentiate` preserve original qualifiers.
- Increased usability of derived functions in cases where constness is involved. For example, const member functions can be called on both `const` and non-const objects, so users will also expect the same behaviour from derived functions of those const member functions, but if the `const` qualifier is not preserved, then the derived function can only be called on non-const objects, thus restricting its usage. 

Closes #210

### DIFF
--- a/test/Gradient/MemberFunctions.C
+++ b/test/Gradient/MemberFunctions.C
@@ -13,652 +13,652 @@ public:
   SimpleFunctions(double p_x = 0, double p_y = 0) : x(p_x), y(p_y) {}
   double x, y;
   double mem_fn(double i, double j)  { 
-   return (x+y)*i + i*j; 
- } 
+    return (x+y)*i + i*j; 
+  } 
 
- // CHECK: void mem_fn_grad(double i, double j, double *_result) {
- // CHECK-NEXT:     double _t0;
- // CHECK-NEXT:     double _t1;
- // CHECK-NEXT:     double _t2;
- // CHECK-NEXT:     double _t3;
- // CHECK-NEXT:     _t1 = (this->x + this->y);
- // CHECK-NEXT:     _t0 = i;
- // CHECK-NEXT:     _t3 = i;
- // CHECK-NEXT:     _t2 = j;
- // CHECK-NEXT:     double mem_fn_return = _t1 * _t0 + _t3 * _t2;
- // CHECK-NEXT:     goto _label0;
- // CHECK-NEXT:   _label0:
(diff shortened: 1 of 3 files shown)


test/Gradient/MemberFunctions.C
```diff
@@ -13,652 +13,652 @@ public:
...
-  // CHECK-NEXT:     {
-  // CHECK-NEXT:         double _r0 = 1 * _t0;
-  // CHECK-NEXT:         double _r1 = _t1 * 1;
-  // CHECK-NEXT:         _result[0UL] += _r1;
-  // CHECK-NEXT:         double _r2 = 1 * _t2;
-  // CHECK-NEXT:         _result[0UL] += _r2;
-  // CHECK-NEXT:         double _r3 = _t3 * 1;
-  // CHECK-NEXT:         _result[1UL] += _r3;
-  // CHECK-NEXT:     }
-  // CHECK-NEXT: }
```
vgvassilev, line 661: Can you preserve the indentation, even if wrong, to outline the real changes to this file?

parth-07: Done, now the previous indentation is preserved.

vgvassilev (approved):
LGTM! Thanks!

PR #250, opened 2021-06-19
Modify clad::hessian to support differentiating member functions.

parth-07:
Progress Checklist:
- [x] Modify `clad::hessian` to support differentiating member functions for clang>=9
- [x] Fix compatibility issues to support earlier clang versions
- [x] Add basic tests

### DIFF
--- a/include/clad/Differentiator/Compatibility.h
+++ b/include/clad/Differentiator/Compatibility.h
@@ -431,6 +431,34 @@ static inline QualType getConstantArrayType(const ASTContext &Ctx,
    #define CLAD_COMPAT_CLANG12_LR_ExtraParams(Node) ,Node->getLParenLoc(),Node->getRParenLoc()
 #endif
 
+/// In Clang < 8, `CXXMethodDecl::getThisType()` member function requires
+/// `ASTContext` to be passed as an argument.
+static inline QualType CXXMethodDecl_getThisType(Sema& SemaRef,
+                                                 const CXXMethodDecl* method) {
+#if CLANG_VERSION_MAJOR >= 8
+  auto thisType = method->getThisType();
+#elif CLANG_VERSION_MAJOR < 8
+  auto thisType = method->getThisType(SemaRef.getASTContext());
+#endif
+  return thisType;
+}
+
+/// Clang < 9, do not provide `Sema::BuildCXXThisExpr` function.
+static inline CXXThisExpr* Sema_BuildCXXThisExpr(Sema& SemaRef,
+                                                 const CXXMethodDecl* method) {
+  auto thisType = CXXMethodDecl_getThisType(SemaRef, method);
+  SourceLocation noLoc;
+#if CLANG_VERSION_MAJOR >= 9
+  return cast<CXXThisExpr>(
+      SemaRef.BuildCXXThisExpr(noLoc, thisType, /*IsImplicit=*/true));
+#elif CLANG_VERSION_MAJOR < 9
+  auto thisExpr = new (SemaRef.getASTContext())
+      CXXThisExpr(noLoc, thisType, /*IsImplicit=*/true);
+  SemaRef.CheckCXXThisCapture(thisExpr->getExprLoc());
+  return thisExpr;
+#endif
+}
+
 } // namespace clad_compat
 
 #endif //CLAD_COMPATIBILITY
--- a/lib/Differentiator/HessianModeVisitor.cpp
+++ b/lib/Differentiator/HessianModeVisitor.cpp
@@ -195,7 +195,6 @@ namespace clad {
     // Creates callExprs to the second derivative functions genereated
     // and creates maps array elements to input array.
     for (size_t i = 0, e = secDerivFuncs.size(); i < e; ++i) {
-      Expr* exprFunc = BuildDeclRef(secDerivFuncs[i]);
       const int numIndependentArgs = secDerivFuncs[i]->getNumParams();
 
       auto size_type = m_Context.getSizeType();
@@ -226,15 +225,7 @@ namespace clad {
                      });
       DeclRefToParams.pop_back();
       DeclRefToParams.push_back(addressArrayExpr);
-
-      Expr* call =
-          m_Sema
-              .ActOnCallExpr(getCurrentScope(),
-                             exprFunc,
-                             noLoc,
-                             llvm::MutableArrayRef<Expr*>(DeclRefToParams),
-                             noLoc)
-              .get();
+      Expr* call = BuildCallExprToFunction(secDerivFuncs[i], DeclRefToParams);
       CompStmtSave.push_back(call);
     }
 
--- a/include/clad/Differentiator/VisitorBase.h
+++ b/include/clad/Differentiator/VisitorBase.h
@@ -316,6 +316,26 @@ namespace clad {
       VD->setInit(ICAR.get());
 
     }
+
+    /// Build a call to member function through this pointer.
+    ///
+    /// \param[in] FD callee member function
+    /// \param[in] argExprs function arguments expressions
+    /// \returns Built member function call expression
+    clang::Expr*
+    BuildCallExprToMemFn(clang::CXXMethodDecl* FD,
+        llvm::MutableArrayRef<clang::Expr*> argExprs);
+
+    /// Build a call to a free function or member function through
+    /// this pointer depending on whether the `FD` argument corresponds to a
+    /// free function or a member function.
+    ///
+    /// \param[in] FD callee function
+    /// \param[in] argExprs function arguments expressions
+    /// \returns Built call expression
+    clang::Expr* 
+    BuildCallExprToFunction(clang::FunctionDecl* FD,
+        llvm::MutableArrayRef<clang::Expr*> argExprs);
   };
 } // end namespace clad
 
--- a/lib/Differentiator/VisitorBase.cpp
+++ b/lib/Differentiator/VisitorBase.cpp
@@ -546,4 +546,54 @@ namespace clad {
     // i.e. tape<T> -> clad::tape<T>
     return m_Context.getElaboratedType(ETK_None, NS, TT);
   }
+
+  clang::Expr* 
+  VisitorBase::BuildCallExprToMemFn(clang::CXXMethodDecl* FD,
+                  llvm::MutableArrayRef<clang::Expr*> argExprs) {
+    auto thisExpr = clad_compat::Sema_BuildCXXThisExpr(m_Sema, FD);
+    CXXScopeSpec CS;
+    CS.Adopt(FD->getQualifierLoc());
+    auto memExpr = m_Sema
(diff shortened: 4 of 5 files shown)


lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -250,4 +242,49 @@ namespace clad {
 
     return result;
   }
+
+  Expr* HessianModeVisitor::BuildRelevantCallExpr(
```
vgvassilev, line 246: How about something like `BuildCallExprToFunction`? The word `relevant` does not have any meaning here.

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -250,4 +242,49 @@ namespace clad {
...
+      FunctionDecl* FD, llvm::MutableArrayRef<Expr*> declRefToParams) {
+    Expr* call = nullptr;
+    if (auto derMethod = dyn_cast<CXXMethodDecl>(FD)) {
+      auto* thisExpr =
+          new (m_Context) CXXThisExpr(noLoc, derMethod->getThisType(), true);
+      m_Sema.CheckCXXThisCapture(thisExpr->getExprLoc());
+      auto memExpr = MemberExpr::Create(
```
vgvassilev, line 253: Maybe run clang-format on this PR.

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -250,4 +242,49 @@ namespace clad {
 
     return result;
   }
+
+  Expr* HessianModeVisitor::BuildRelevantCallExpr(
+      FunctionDecl* FD, llvm::MutableArrayRef<Expr*> declRefToParams) {
+    Expr* call = nullptr;
+    if (auto derMethod = dyn_cast<CXXMethodDecl>(FD)) {
+      auto* thisExpr =
+          new (m_Context) CXXThisExpr(noLoc, derMethod->getThisType(), true);
```
vgvassilev, line 251:
```suggestion
          new (m_Context) CXXThisExpr(noLoc, derMethod->getThisType(), /*IsImplicit=*/ true);
```

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -250,4 +242,49 @@ namespace clad {
...
+      m_Sema.CheckCXXThisCapture(thisExpr->getExprLoc());
+      auto memExpr = MemberExpr::Create(
+          m_Context,
+          thisExpr,
+          true,
```
vgvassilev, line 256: We should add the parameter name as a comment. That way we improve documentation and allow clang-tidy to diagnose if parameters were inverted.

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -250,4 +242,49 @@ namespace clad {
...
+          derMethod,
+          DeclAccessPair::make(derMethod, derMethod->getAccess()),
+          derMethod->getNameInfo(),
+          nullptr,
+          m_Context.BoundMemberTy,
+          ExprValueKind::VK_RValue,
+          ExprObjectKind::OK_Ordinary
+              CLAD_COMPAT_CLANG9_MemberExpr_ExtraParams_NOUR_None);
+      call = m_Sema
+                 .BuildCallToMemberFunction(getCurrentScope(),
```
vgvassilev, line 269: That hyphenation is really bad for reading. Is that what our clang-format config suggests?

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -250,4 +242,49 @@ namespace clad {
...
+          noLoc,
+          NestedNameSpecifierLoc(derMethod->getQualifier(), nullptr),
+          noLoc,
+          derMethod,
+          DeclAccessPair::make(derMethod, derMethod->getAccess()),
+          derMethod->getNameInfo(),
+          nullptr,
+          m_Context.BoundMemberTy,
+          ExprValueKind::VK_RValue,
+          ExprObjectKind::OK_Ordinary
```
vgvassilev, line 266: Hardcoding these values looks worrying. Can't we use `Sema::BuildMemberReferenceExpr` instead?

lib/Differentiator/HessianModeVisitor.cpp
parth-07: Okay, I will add argument names while calling clang functions. 

I have one doubt related to compatibility issues here,
We can create `this` expression easily using `Sema::BuildCXXThisExpr` for clang >= 9.
For clang 8, we can create `this` expression, using `CXXThis` constructor and `m_Sema.CheckCXXThisCapture`.
For clang <= 7, creating `this` is much more complicated, since `CXXMethodDecl::getThisType()` is missing along with some other useful methods which can help to create qualified `this` type. 

While being in a situation like this, should we aim to use code that works for all clang versions or should we create a macro in the `compatibility.h` file, such that it creates `this` expression with the most readable code for any given clang version?

lib/Differentiator/HessianModeVisitor.cpp
parth-07: Yes, this is what our clang-format configuration suggests. I will try to improve our clang-config configuration for nicely formatting this code.

lib/Differentiator/HessianModeVisitor.cpp
parth-07: Yes, agreed. I only used those values because they seemed reasonable, and I did not know about any option other than hard-coding them. 
I will look into `Sema::BuilldMemberReferenceExpr`.

lib/Differentiator/HessianModeVisitor.cpp
parth-07: Okay, I will make the necessary changes.

lib/Differentiator/HessianModeVisitor.cpp
parth-07: clang-format suggests this exact same code. I will try to improve clang-format configuration to improve formatting for cases like these.

lib/Differentiator/HessianModeVisitor.cpp
vgvassilev: I would prefer the latter option. I think that's the take of @alexander-penev, too but let's ask him.

lib/Differentiator/HessianModeVisitor.cpp
parth-07: I wanted to convey that this function can build calls to free functions as well as member functions depending on the passed function. Yes, `relevant` don't do a very good job for this purpose. `BuildCallExprToFunction` can be a bit misleading since in clang sema `sema::BuildCallExpr` only builds call to free functions. I will try to come up with a good name for this function.

sudo-panda: @parth-07  could you please put this function in a separate PR(so that it can be merged early) and in `VisitorBase`. I require this functionality in `ReverseModeVisitor` and it would be great if I could use your function.

parth-07: As of now, I only have to complete this function to complete the PR (the PR consists mostly of this function only), if the PR requires any additional tasks then I will create a separate PR for this function. I will try to complete this function as soon as possible :)

lib/Differentiator/HessianModeVisitor.cpp
parth-07: The current `ContinuationIndentWidth` value for clang-format configuration is 4 , setting it to 2, slightly improves the formatting in this case. 
But the method name still gets formatted on the next line, I am unable to find any configuration option which will format the method name on the same line as `m_Sema` in this case. Should I change `ContinuationIndentWidth` to 2 ?

lib/Differentiator/HessianModeVisitor.cpp
parth-07: Yes, `Sema::BuildMemberReference` works perfectly, I have modified the code to use it instead of `MemberExpr::create`

lib/Differentiator/HessianModeVisitor.cpp
parth-07: I am unable to come up with a good function name for this, I have decided on the suggested name `BuildCallExprToFunction`, it also goes well with other similar functions in `VisitorBase`. Please tell me if I should modify this function name.

include/clad/Differentiator/Compatibility.h
```diff
@@ -431,6 +431,34 @@ static inline QualType getConstantArrayType(const ASTContext &Ctx,
...
+}
+
+/// Clang < 9, do not provide `Sema::BuildCXXThisExpr` function.
+static inline CXXThisExpr* Sema_BuildCXXThisExpr(Sema& SemaRef,
+                                                 const CXXMethodDecl* method) {
+  auto thisType = CXXMethodDecl_getThisType(SemaRef, method);
+  SourceLocation noLoc;
+#if CLANG_VERSION_MAJOR >= 9
+  return cast<CXXThisExpr>(
+      SemaRef.BuildCXXThisExpr(noLoc, thisType, /*IsImplicit*/ true));
```
vgvassilev, line 453:
```suggestion
      SemaRef.BuildCXXThisExpr(noLoc, thisType, /*IsImplicit=*/ true));
```

include/clad/Differentiator/Compatibility.h
```diff
@@ -431,6 +431,34 @@ static inline QualType getConstantArrayType(const ASTContext &Ctx,
...
+#elif CLANG_VERSION_MAJOR < 9
+  auto thisExpr = new (SemaRef.getASTContext())
+      CXXThisExpr(noLoc, thisType, /*IsImplicit*/ true);
```
vgvassilev, line 456:
```suggestion
      CXXThisExpr(noLoc, thisType, /*IsImplicit=*/ true);
```

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -250,4 +241,5 @@ namespace clad {
 
     return result;
   }
+
```
vgvassilev, line 244:
```suggestion
```

vgvassilev (commented):
Have you run clang-format on the code?

Could you also update all uses of the kind `/*Base*/` to `/*Base=*/` which enables a clang-tidy check on the parameter names?

include/clad/Differentiator/Compatibility.h
parth-07: I ran clang-format on some parts of the code, I will run it on all the added code now. 
Yes, I will update all uses of the kind /*Base*/ to /*Base=*/

parth-07: done

include/clad/Differentiator/Compatibility.h
parth-07: done

vgvassilev (approved):
Lgtm! Thanks!

PR #254, opened 2021-07-04
Add support for reference variables in reverse mode

### LINKED ISSUE #253: Support reference variables in reverse mode differentiation
The reverse mode does not support differentiating reference variables, so if the original function uses reference variables then the program fails to compile. 

Minimum reproducible example:
```cpp
#include <iostream>

  double &a = i;
  return a;
}

int main() {
  auto d_fn = clad::gradient(&fn, "i");
}
```
This program fails to compile with error: `non-const lvalue reference to type 'double' cannot bind to an initializer list temporary`.

**Cause of the error**
Using `-fdump-derived-fn`, we obtain the following definition of the derived function:

```c++
void fn_grad_0(double i, double j, double *_result) {
    double &_d_a;
    double &a = i;
    double fn_return = a;
    goto _label0;
  _label0:
    _d_a += 1;
    _result[0UL] += _d_a;
}
```
Here we can observe the 1st line of the function, `double &_d_a` is ill-formed and needs to be fixed.

parth-07:
Progress checklist:
- [x] Add support for reference variables in reverse mode
- [x] Add tests

### Solution:
```c++
double &a = b;
```
Generally, derived variables are initialised with 0, but in the case of reference variables, they should be initialised with the derived variable of the variable they are referencing to. Thus, the updated definition of `_d_a` will be, 
```c++
double& _d_a = _d_b;
```
And since, reference variables are referencing an already declared and derived variable, their definition does not need to be derived.

Closes #253

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1465,14 +1465,40 @@ namespace clad {
   }
 
   VarDeclDiff ReverseModeVisitor::DifferentiateVarDecl(const VarDecl* VD) {
-    auto zero = getZeroInit(VD->getType());
+    StmtDiff initDiff;
+    Expr* VDDerivedInit = nullptr;
+    auto VDDerivedType = getNonConstType(VD->getType(), m_Context, m_Sema);
+    bool isVDRefType = VD->getType()->isReferenceType();
+    // If VD is a reference to a local variable, then the initial value is set
+    // to the derived variable of the corresponding local variable.
+    // If VD is a reference to a non-local variable (global variable, struct
+    // member etc), then no derived variable is available, thus `VDDerived` 
+    // does not need to reference any variable, consequentially the 
+    // `VDDerivedType` is the corresponding non-reference type and the initial
+    // value is set to 0. 
+    // Otherwise, for non-reference types, the initial value is set to 0.
+    if (isVDRefType) {
+      initDiff = Visit(VD->getInit());
+      VDDerivedInit = initDiff.getExpr_dx();
+      if (!VDDerivedInit) {
+        VDDerivedType = VDDerivedType.getNonReferenceType();
+        VDDerivedInit = getZeroInit(VDDerivedType);
+      }
+    } else {
+      VDDerivedInit = getZeroInit(VD->getType());
+    }
     VarDecl* VDDerived =
-        BuildVarDecl(getNonConstType(VD->getType(), m_Context, m_Sema),
+        BuildVarDecl(VDDerivedType,
                      "_d_" + VD->getNameAsString(),
-                     zero);
-    StmtDiff initDiff = VD->getInit()
-                            ? Visit(VD->getInit(), BuildDeclRef(VDDerived))
-                            : StmtDiff{};
+                     VDDerivedInit);
+
+    // If `VD` is a reference to a local variable, then it is already
+    // differentiated and should not be differentiated again.
+    // If `VD` is a reference to a non-local variable then also there's no
+    // need to call `Visit` since non-local variables are not differentiated.
+    if (!isVDRefType)
+      initDiff = VD->getInit() ? Visit(VD->getInit(), BuildDeclRef(VDDerived))
+                               : StmtDiff{};
     VarDecl* VDClone = BuildVarDecl(VD->getType(),
                                     VD->getNameAsString(),
                                     initDiff.getExpr(),
--- a/test/Gradient/Assignments.C
+++ b/test/Gradient/Assignments.C
@@ -553,6 +553,181 @@ double f13(double x, double y) {
 //CHECK-NEXT:       }
 //CHECK-NEXT:   }
 
+double f14(double i, double j) {
+  double& a = i;
+  a = 2*i;
+  a += i;
+  a *= i;
+  return i;
+}
+
+// CHECK: void f14_grad(double i, double j, double *_result) {
+// CHECK-NEXT:     double &_d_a = _result[0UL];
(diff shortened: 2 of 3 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1465,14 +1465,38 @@ namespace clad {
...
+        BuildVarDecl(VDDerivedType,
                      "_d_" + VD->getNameAsString(),
-                     zero);
-    StmtDiff initDiff = VD->getInit()
-                            ? Visit(VD->getInit(), BuildDeclRef(VDDerived))
-                            : StmtDiff{};
+                     VDDerivedInit);
+    // If `VD` is a reference to a local variable, then it is already
+    // differentiated and should not be differentiated again.
+    if (!VD->getType()->isReferenceType() || !initDiff.getExpr_dx())
```
vgvassilev, line 1497: Can you move this to a local variable `!VD->getType()->isReferenceType()` as we use it twice? Is that condition meant to be in the `else` part of the `VD->getType()->isReferenceType()`?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1465,14 +1465,38 @@ namespace clad {
...
 
   VarDeclDiff ReverseModeVisitor::DifferentiateVarDecl(const VarDecl* VD) {
-    auto zero = getZeroInit(VD->getType());
+    StmtDiff initDiff;
+    Expr* VDDerivedInit = nullptr;
+    auto VDDerivedType = getNonConstType(VD->getType(), m_Context, m_Sema);
+
+    // If VD is a reference variable referencing a variable in local scope,
+    // then initial value is set to the derived variable of the variable 
+    // it's referencing to.
```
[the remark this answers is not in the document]
parth-07: Okay.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: This body of this condition uses `VDDerived` which is created after the if block of `VD->getType()->isReferenceType`. But yes, it handles cases for non-reference types, similar to the `else` part of the first if block (just above the `VDDerived` definition).

Currently, 
```c++
initDiff = VD->getInit() ? Visit(VD->getInit(), BuildDeclRef(VDDerived))
                               : StmtDiff{};
```

is also being called for variables that are referencing a struct member (for example, `double& u = this->x`) (hence, the extra `|| !initDiff.getExpr_dx()` part) , since, for cases like these, the variable has not already been differentiated. But I just realized this extra condition can be safely removed, since, no differentiation is performed when the value (here, `VD->getInit()`) is a struct variable. I will modify the if condition here.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1465,14 +1465,39 @@ namespace clad {
   }
 
   VarDeclDiff ReverseModeVisitor::DifferentiateVarDecl(const VarDecl* VD) {
-    auto zero = getZeroInit(VD->getType());
+    StmtDiff initDiff;
+    Expr* VDDerivedInit = nullptr;
+    auto VDDerivedType = getNonConstType(VD->getType(), m_Context, m_Sema);
+    bool isVDRefType = VD->getType()->isReferenceType();
+    // If VD is a reference to a local variable, then initial value is set to
```
vgvassilev, line 1472:
```suggestion
    // If VD is a reference to a local variable, then the initial value is set to
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1465,14 +1465,39 @@ namespace clad {
...
+    // the derived variable of the corresponding local variable.
+    // If VD is a reference to a non-local variable (global variable, struct
+    // member etc), then no derived variable is available, thus `VDDerived` do
```
vgvassilev, line 1475:
```suggestion
    // member etc), then no derived variable is available, thus `VDDerived` does
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1465,14 +1465,39 @@ namespace clad {
...
+    // not need to reference any variable, consequentially reference type is
+    // removed from `VDDerivedType` and initial value is set to 0.
```
vgvassilev, line 1477:
```suggestion
    // not need to reference any variable, consequentially the `VDDerivedType`
    // is the corresponding non-reference type and the initial value is set to 0.
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1465,14 +1465,39 @@ namespace clad {
...
+    Expr* VDDerivedInit = nullptr;
+    auto VDDerivedType = getNonConstType(VD->getType(), m_Context, m_Sema);
+    bool isVDRefType = VD->getType()->isReferenceType();
+    // If VD is a reference to a local variable, then initial value is set to
+    // the derived variable of the corresponding local variable.
+    // If VD is a reference to a non-local variable (global variable, struct
+    // member etc), then no derived variable is available, thus `VDDerived` do
+    // not need to reference any variable, consequentially reference type is
+    // removed from `VDDerivedType` and initial value is set to 0. 
+    // Otherwise, initial value is set to 0.
```
vgvassilev, line 1478:
```suggestion
    // Otherwise, for non-reference types, the initial value is set to 0.
```

vgvassilev (commented):
Please consider squashing all into a single commit.

vgvassilev (approved):
LGTM!

PR #260, opened 2021-07-13
Add support for differentiating functors using clad::jacobian

parth-07:
Progress checklist:
- [x] Add support for computing jacobian of functor using `clad::jacobian`.
- [x] Add tests


vgvassilev (approved):
LGTM! Consider pinging @sudo-panda on how to support differentiating functors with arrays as data members.

PR #262, opened 2021-07-15
Fix building of nested name specifiers in getArg function

parth-07:
This PR modifies code in getArg function to build correct and complete nested name specifiers for the call operator.
It also fixes the failing asserts in clang debug build on running tests for differentiating functors. Credit goes to @grimmmyshini and @sudo-panda for bringing this error to attention. 

The root cause of the failing assert was lambda types not having any associated identifiers. This was causing issues while creating correct nested name specifiers for the call operator.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -46,7 +46,7 @@ namespace clad {
             } else {
               TraverseStmt(VD->getInit());
             }
-          } else if (auto FD = dyn_cast<FunctionDecl>(DRE->getDecl()))
+          } else if (isa<FunctionDecl>(DRE->getDecl()))
             m_FnDRE = DRE;
           return false;
         }
@@ -144,10 +144,7 @@ namespace clad {
           // method, to maintain consistency with member function
           // differentiation.
           CXXScopeSpec CSS;
-          CSS.Extend(m_SemaRef.getASTContext(),
-                     RD->getIdentifier(),
-                     /*IdentifierLoc=*/noLoc,
-                     /*ColonColonLoc=*/noLoc);
+          BuildNNS(callOperator->getDeclContext(), CSS);
 
           // `ExprValueKind::VK_RValue` is used because functions are
           // decomposed to function pointers and thus a temporary is
@@ -158,9 +155,50 @@ namespace clad {
                                          ExprValueKind::VK_RValue,
                                          noLoc,
                                          &CSS));
-          m_FnDRE = cast<DeclRefExpr>(newFnDRE);
-
+          m_FnDRE = cast<DeclRefExpr>(newFnDRE);          
           return false;
+        }
+      private:
+        /// Creates nested name specifier associated with declaration context
+        /// argument `DC`. 
+        ///
+        /// For example, given a structure defined as,
+        /// namespace A {
+        /// namespace B {
+        ///   struct SomeStruct {};
+        /// }
+        /// }
+        ///
+        /// Passing `SomeStruct` as declaration context will create
+        /// nested name specifier of the form `::A::B::struct SomeClass::`
+        /// in `CXXScopeSpec` argument `CSS`.
+        /// \note Currently only namespace and class/struct nested name specifiers
+        /// are supported.
+        ///
+        /// \param[in] DC
+        /// \param[out] CSS
+        void BuildNNS(DeclContext* DC, CXXScopeSpec& CSS) {
+          assert(DC && "Must provide a non null DeclContext");
+
+          // parent name specifier should be added first
+          if (DC->getParent())
+            BuildNNS(DC->getParent(), CSS);
+
+          ASTContext& Context = m_SemaRef.getASTContext();
+
+          if (auto ND = dyn_cast<NamespaceDecl>(DC)) {
+            CSS.Extend(Context, ND,
+                       /*NamespaceLoc=*/noLoc,
+                       /*ColonColonLoc=*/noLoc);
+          } else if (auto RD = dyn_cast<CXXRecordDecl>(DC)) {
+            auto RDQType = RD->getTypeForDecl()->getCanonicalTypeInternal();
+            auto RDTypeSourceInfo = Context.getTrivialTypeSourceInfo(RDQType);
+            CSS.Extend(Context,
+                       /*TemplateKWLoc=*/noLoc, RDTypeSourceInfo->getTypeLoc(),
+                       /*ColonColonLoc=*/noLoc);
+          } else if (isa<TranslationUnitDecl>(DC)) {
+            CSS.MakeGlobal(Context, /*ColonColonLoc=*/noLoc);
+          }
         }
     } finder(SemaRef, call->getArg(0)->getBeginLoc());
     finder.TraverseStmt(call->getArg(0));
--- a/test/ForwardMode/Functors.C
+++ b/test/ForwardMode/Functors.C
@@ -100,6 +100,10 @@ namespace outer {
       // CHECK-NEXT:     return (0. * i + _t0 * _d_i) * j + _t1 * _d_j;
       // CHECK-NEXT: }   
     };
+
+    auto lambdaNNS = [](double i, double j) {
+      return i*i*j;
+    };
   }
 }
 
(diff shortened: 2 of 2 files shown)


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -144,10 +144,24 @@ namespace clad {
...
+            NNSBuilder.Extend(m_SemaRef.getASTContext(),
+                              /*TemplateKWLoc=*/noLoc,
+                              RD->getLambdaTypeInfo()->getTypeLoc(),
+                              /*ColonColonLoc=*/noLoc);
+
+            // Creating nested name specifier location using nested name specifier
+            // of `TypeLoc` without any associated source location. 
+            CSS.MakeTrivial(m_SemaRef.getASTContext(),
+                            NNSBuilder.getRepresentation(),
+                            SourceRange());
```
vgvassilev, line 160: Can't we sink this chunk of code in `BuildNNS`?

parth-07: We can, but then in `BuildNNS` function we will need to check at each recursion if the declaration context passed is of lambda type or not. 
Or we can make 2 functions, `BuildNNS` and `BuildNNS_impl`. Then we can perform check just once in `BuildNNS` and do the current work of `BuildNNS` in `BuildNNS_impl`.  Should I go with this approach ?

parth-07: @vgvassilev Now lambda types do not have to be handled differently, thus `BuildNNS` handles both lambda and non-lambda record declarations now.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -46,7 +46,7 @@ namespace clad {
             } else {
               TraverseStmt(VD->getInit());
             }
-          } else if (auto FD = dyn_cast<FunctionDecl>(DRE->getDecl()))
+          } else if (dyn_cast<FunctionDecl>(DRE->getDecl()))
```
vgvassilev, line 49:
```suggestion
          } else if (isa<FunctionDecl>(DRE->getDecl()))
```

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -159,9 +156,52 @@ namespace clad {
                                          noLoc,
                                          &CSS));
           m_FnDRE = cast<DeclRefExpr>(newFnDRE);
-
+
```
vgvassilev, line 159: Stray space.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -159,9 +156,52 @@ namespace clad {
                                          noLoc,
                                          &CSS));
           m_FnDRE = cast<DeclRefExpr>(newFnDRE);
-
+          
           return false;
         }
+      private:
+        /// Creates nested name specifier associated with declaration context
+        /// argument `DC`
```
vgvassilev, line 164:
```suggestion
        /// argument `DC`.
```

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -159,9 +156,52 @@ namespace clad {
...
+        /// namespace A {
+        /// namespace B {
+        ///   struct SomeStruct {};
+        /// }
+        /// }
+        ///
+        /// Passing `SomeStruct` as declaration context will create
+        /// nested name specifier of the form `::A::B::struct SomeClass::`
+        /// in `CXXScopeSpec` argument `CSS`.
+        /// ! Currently only, namespace and class/struct nested name specifiers
```
vgvassilev, line 176: What does `!` mean here for doxygen?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -159,9 +156,52 @@ namespace clad {
...
+
+          if (auto ND = dyn_cast<NamespaceDecl>(DC)) {
+            CSS.Extend(m_SemaRef.getASTContext(),
+                       ND,
+                       /*NamespaceLoc=*/noLoc,
+                       /*ColonColonLoc=*/noLoc);
+          } else if (auto RD = dyn_cast<CXXRecordDecl>(DC)) {
+            auto RDTypeSourceInfo =
+                m_SemaRef.getASTContext().CreateTypeSourceInfo(
+                    RD->getTypeForDecl()->getCanonicalTypeInternal());
```
vgvassilev, line 196: Please move `RD->getTypeForDecl()->getCanonicalTypeInternal()` into a local QualType variable.

parth-07: Okay

lib/Differentiator/DiffPlanner.cpp
parth-07: I thought `!` stands for attention in Doxygen, I was wrong. Doxgen have `\attention` command for this purpose.

parth-07: Similarly, there's also a `\note` command which I think suits better in this case. Thus, I have replace `!` with `\note` here.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -158,10 +155,52 @@ namespace clad {
...
+        /// nested name specifier of the form `::A::B::struct SomeClass::`
+        /// in `CXXScopeSpec` argument `CSS`.
+        /// \note Currently only namespace and class/struct nested name specifiers
+        /// are supported.
+        ///
+        /// \param[in] DC
+        /// \param[out] CSS
+        void BuildNNS(DeclContext* DC, CXXScopeSpec& CSS) {
+          if (!DC)
+            return;
```
vgvassilev, line 182:
```suggestion
         assert(DC && "Must provide a non null DeclContext");
```

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -158,10 +155,52 @@ namespace clad {
...
+
+          if (auto ND = dyn_cast<NamespaceDecl>(DC)) {
+            CSS.Extend(m_SemaRef.getASTContext(),
+                       ND,
+                       /*NamespaceLoc=*/noLoc,
+                       /*ColonColonLoc=*/noLoc);
+          } else if (auto RD = dyn_cast<CXXRecordDecl>(DC)) {
+            auto RDQType = RD->getTypeForDecl()->getCanonicalTypeInternal();
+            auto RDTypeSourceInfo =
+                m_SemaRef.getASTContext().CreateTypeSourceInfo(RDQType);
```
vgvassilev, line 195: I think here we `getTrivialTypeSourceInfo` should be better.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -158,10 +155,52 @@ namespace clad {
...
+        /// \param[out] CSS
+        void BuildNNS(DeclContext* DC, CXXScopeSpec& CSS) {
+          if (!DC)
+            return;
+
+          // parent name specifier should be added first
+          BuildNNS(DC->getParent(), CSS);
+
+          if (auto ND = dyn_cast<NamespaceDecl>(DC)) {
+            CSS.Extend(m_SemaRef.getASTContext(),
```
vgvassilev, line 188: Please move the ASTContext into a temporary variable.

vgvassilev:
Please also update the commit message to match the PR description.

parth-07:
@vgvassilev I have updated the commit message to match the PR description.

vgvassilev (approved):
LGTM!

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -158,10 +155,51 @@ namespace clad {
...
+            CSS.Extend(Context, ND,
+                       /*NamespaceLoc=*/noLoc,
+                       /*ColonColonLoc=*/noLoc);
+          } else if (auto RD = dyn_cast<CXXRecordDecl>(DC)) {
+            auto RDQType = RD->getTypeForDecl()->getCanonicalTypeInternal();
+            auto RDTypeSourceInfo = Context.getTrivialTypeSourceInfo(RDQType);
+            CSS.Extend(Context,
+                       /*TemplateKWLoc=*/noLoc, RDTypeSourceInfo->getTypeLoc(),
+                       /*ColonColonLoc=*/noLoc);
+          } else if (dyn_cast<TranslationUnitDecl>(DC)) {
```
vgvassilev, line 199:
```suggestion
          } else if (isa<TranslationUnitDecl>(DC)) {
```

PR #263, opened 2021-07-15
Modify code to use effective fn name to create derived fn identifiers 

parth-07:
This PR modifies derived function and corresponding return variable name to valid identifiers when differentiating overloaded call operator in reverse mode. Credit goes to @grimmmyshini and @sudo-panda for bringing this error to attention. 

Extending the original function creates invalid identifiers for the derived function and the return variable name when the original function is an overloaded operator, due to parentheses in the original function name.

Effective name is the function name itself for ordinary functions and a unique name for each operator kind for overloaded operators.

Current derived function name:
```c++
void operator()_grad(...);
```

Modified derived function name:
```c++
void operator_call_grad(...);
```

Current return variable name:
```
operator()_return;
```

Modified return variable name:
```
operator_call_return;
```

### DIFF
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -1,0 +1,15 @@
+#include "clad/Differentiator/CladUtils.h"
+
+#include "clang/AST/Decl.h"
+
+namespace clad {
+  namespace utils {
+    std::string ComputeEffectiveFnName(const clang::FunctionDecl* FD) {
+      // TODO: Add cases for more operators
+      switch (FD->getOverloadedOperator()) {
+        case clang::OverloadedOperatorKind::OO_Call: return "operator_call";
+        default: return FD->getNameAsString();
+      }
+    }
+  } // namespace utils
+} // namespace clad
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -10,6 +10,7 @@
 
 #include "llvm/Support/SaveAndRestore.h"
 
+#include "clad/Differentiator/CladUtils.h"
 #include "clad/Differentiator/Compatibility.h"
 
 using namespace clang;
@@ -317,7 +318,7 @@ namespace clad {
       request.Args = E->getArg(1);
       auto derivedFD = cast<FunctionDecl>(DRE->getDecl());
       request.Function = derivedFD;
-      request.BaseFunctionName = derivedFD->getNameAsString();
+      request.BaseFunctionName = utils::ComputeEffectiveFnName(request.Function);
 
       // FIXME: add support for nested calls to clad::differentiate/gradient
       // inside differentiated functions
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -26,6 +26,7 @@
 #include <algorithm>
 #include <numeric>
 
+#include "clad/Differentiator/CladUtils.h"
 #include "clad/Differentiator/Compatibility.h"
 
 using namespace clang;
@@ -247,7 +248,7 @@ namespace clad {
           GetCladArrayRefOfType(m_Function->getReturnType());
     }
 
-    auto derivativeBaseName = m_Function->getNameAsString();
+    auto derivativeBaseName = request.BaseFunctionName;
     std::string gradientName = derivativeBaseName + funcPostfix();
     // To be consistent with older tests, nothing is appended to 'f_grad' if
     // we differentiate w.r.t. all the parameters at once.
@@ -958,7 +959,7 @@ namespace clad {
     // computation (e.g. assignments), we also have to emit it to execute it.
     StoreAndRef(ExprDiff.getExpr(),
                 forward,
-                m_Function->getNameAsString() + "_return",
+                utils::ComputeEffectiveFnName(m_Function) + "_return",
                 /*force*/ true);
     // Create goto to the label.
     return m_Sema.ActOnGotoStmt(noLoc, noLoc, LD).get();
--- a/include/clad/Differentiator/CladUtils.h
+++ b/include/clad/Differentiator/CladUtils.h
@@ -1,0 +1,22 @@
+// This file contains utility functions which do not belong anywhere else
+
+#ifndef CLAD_UTILS_CLADUTILS_H
+#define CLAD_UTILS_CLADUTILS_H
+
+#include <string>
+
+namespace clang {
+  class FunctionDecl;
+}
+
+namespace clad {
+  namespace utils {
+    /// If `FD` is an overloaded operator, returns a name, unique for
+    /// each operator, that can be used to create valid C++ identifiers.
+    /// Otherwise if `FD` is an ordinary function, returns the name of the
+    /// function `FD`.
+    std::string ComputeEffectiveFnName(const clang::FunctionDecl* FD);
+  }
+}
+
+#endif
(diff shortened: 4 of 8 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -98,8 +98,13 @@ namespace clad {
       args.pop_back();
     }
 
-    auto derivativeBaseName = m_Function->getNameAsString();
-    std::string gradientName = derivativeBaseName + funcPostfix();
+    switch (m_Function->getOverloadedOperator()) {
+      case OverloadedOperatorKind::OO_Call:
+        m_EffectiveOriginalFnName = "operator_call";
+        break;
+      default: m_EffectiveOriginalFnName = m_Function->getNameAsString(); break;
```
vgvassilev, line 105: Did you run clang-format on this?

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -45,7 +45,7 @@ namespace clad {
     unsigned outputArrayCursor = 0;
     unsigned numParams = 0;
     bool isVectorValued = false;
-
+    std::string m_EffectiveOriginalFnName;
```
vgvassilev, line 48: Can we avoid using a data member for getting the function name? We should try to reduce the shared data members to minimum as that shares state which is hard to debug.

parth-07: Then we will need to compute the value of the effective original function name 2 times, once in `ReverseModeVisitor::Derive` and once in `ReverseModeVisitor::VisitReturnStmt`. Should I compute it 2 times rather than using a shared data member in this case?

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: Yes I did.

include/clad/Differentiator/ReverseModeVisitor.h
vgvassilev: Yes, assuming that you extract out the relevant code into a file static utility function.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: What should be the proper format for this code?

vgvassilev: I'd expect it to be:

```cpp
default:
  m_EffectiveOriginalFnName = m_Function->getNameAsString();
  break;
```

But that would disappear if you address the other comment...

include/clad/Differentiator/ReverseModeVisitor.h
parth-07: Instead of a file static function, can I make it a class private function? 
This way, we will not need to pass function to compute the effective name.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: Oh okay.

include/clad/Differentiator/ReverseModeVisitor.h
vgvassilev: It should be generic so passing a function as a parameter should be fine.

parth-07: I have created `getEffectiveFnName` as a static function in the `VisitorBase` class since this function can also be used in `HessianModeVisitor` class. Please tell if some other place would be better for this function definition.

include/clad/Differentiator/VisitorBase.h
```diff
@@ -336,7 +336,24 @@ namespace clad {
     clang::Expr* 
     BuildCallExprToFunction(clang::FunctionDecl* FD,
         llvm::MutableArrayRef<clang::Expr*> argExprs);
+    
+    /// If `FD` is an overloaded operator, returns a name, unique for
+    /// each operator, which can be used to create valid C++ identifiers.
+    /// Otherwise if `FD` is an ordinary function, returns the name of the
+    /// function `FD`.
+    static std::string getEffectiveFnName(const clang::FunctionDecl* FD) {
```
vgvassilev, line 344:
```suggestion
    static std::string getNameAsString(const clang::FunctionDecl* FD, const std::string& suffix) {
```

parth-07: Can you please tell what should be the purpose of `suffix` parameter here?

vgvassilev: So that you can sink the "_return" and "_hessian" from the call sites.

include/clad/Differentiator/VisitorBase.h
```diff
@@ -336,7 +336,24 @@ namespace clad {
...
+      // TODO: Add cases for more operators
+      switch (FD->getOverloadedOperator()) {
+        case clang::OverloadedOperatorKind::OO_Call:
+          return "operator_call";
+          break;
```
vgvassilev, line 349: You don't need a 'break' after the return.

parth-07: oh, yes. I will remove the break statements from here.

include/clad/Differentiator/ReverseModeVisitor.h
vgvassilev: We can always have some file with utility functions. Eg. `CladASTUtils` or something like that..

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -276,6 +276,19 @@ namespace clad {
...
+  /// Otherwise if `FD` is an ordinary function, returns the name of the
+  /// function `FD`.
+  static std::string ComputeEffectiveFnName(const FunctionDecl* FD) {
+    // TODO: Add cases for more operators
+    switch (FD->getOverloadedOperator()) {
+      case clang::OverloadedOperatorKind::OO_Call:
+        return "operator_call";
+      default: return FD->getNameAsString();
+    }
+  }
```
vgvassilev, line 290: We should move that duplicated code into a separate file -- lib/Differentiator/CladUtils.{h,cpp}

parth-07: Okay.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -923,7 +924,7 @@ namespace clad {
   StmtDiff
   ReverseModeVisitor::VisitCXXBoolLiteralExpr(const CXXBoolLiteralExpr* BL) {
     return Clone(BL);
-  }
+  }
```
vgvassilev, line 927:
```suggestion
  }
```

lib/Differentiator/CladUtils.cpp
```diff
@@ -0,0 +1,15 @@
+#include "clad/Differentiator/CladUtils.h"
+
+#include "clang/AST/Decl.h"
+
+namespace clad {
+  namespace utils {
+    std::string ComputeEffectiveFnName(const clang::FunctionDecl* FD) {
```
vgvassilev, line 7: Maybe add some doxygen documentation?

parth-07: I have added the doxygen documentation in the corresponding header file. 
Should I add the doxygen documentation in the implementation file instead of the header file?

vgvassilev: ah, totally overlooked. sorry..

vgvassilev (approved):
LGTM!

PR #266, opened 2021-07-19
Create derived variables for parameters which are not independent

### LINKED ISSUE #265: Reverse mode do not create derived variables for all parameters
Reverse mode differentiation does not create derived variables for all parameters, this can produce an incorrect result in many cases.

Minimum reproducible example:
```c++
  j = i;
  return j;
}

int main() {
  auto d_fn = clad::gradient(fn, "i");
  double res[2] = {};
  d_fn.execute(3, 5, res);
  std::cout<<res[0]<<"\n";  // gives 0; expected: 1
}
```

derived function produced by clad: 

```c++
void fn_grad_0(double i, double j, double *_result) {
    j = i;
    double fn_return = j;
    goto _label0;
  _label0:
    ;
}
```

The root cause of the incorrect result here is that we are not producing derived variables for parameters that are not independent variables, and thus can be easily fixed by creating derived variables for them.

parth-07:
This PR modifies reverse mode differentiation by creating a derived variable for all parameters. And thus produces correct derived function for cases like these:
```c++
  j = i;
  return j;
}

int main() {
  auto d_fn = clad::gradient(fn, "i");
  double res[2] = {};
  d_fn.execute(3, 5, res);
  std::cout<<res[0]<<"\n";  // gives 0; expected: 1
}
```

Derived function produced:
```c++
void fn_grad_0(double i, double j, double *_result) {
    double _d_j = 0;
    j = i;
    double fn_return = j;
    goto _label0;
  _label0:
    _d_j += 1;
    {
        double _r_d0 = _d_j;
        _result[0UL] += _r_d0;
        _d_j -= _r_d0;
    }
}
```

Closes #265

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -374,7 +374,7 @@ namespace clad {
         }
       }
     }
-
+    auto nonDiffParams = params;
     if (isVectorValued) {
       TypeSourceInfo* paramTSI =
           m_Context.getTrivialTypeSourceInfo(DerivedOutputParamType, noLoc);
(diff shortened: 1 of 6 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -266,6 +266,23 @@ namespace clad {
...
+
+    // create derived variables for parameters which are not part of
+    // independent variables (args).
+    for (auto param : params) {
+      if (m_Variables.count(param) || param == params.back())
+        continue;
+      auto VDDerivedType = param->getType();
+      // We cannot initialize derived variable for pointer types without
+      // knowing the correct size.
+      if (VDDerivedType->isArrayType() || VDDerivedType->isPointerType())
```
vgvassilev, line 278:
```suggestion
      if (isArrayOrPointerType(VDDerivedType))
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -449,6 +449,26 @@ namespace clad {
     beginScope(Scope::FnScope | Scope::DeclScope);
     m_DerivativeFnScope = getCurrentScope();
     beginBlock();
+    // create derived variables for parameters which are not part of
+    // independent variables (args).
+    for (auto param : nonDiffParams) {
```
vgvassilev, line 454: Why not just `params`? Can you spell out the actual type rather than `auto`?

vgvassilev (commented):
Could you update the commit message to the PR description?

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: Because `params` have both the original function parameters and output parameters created to store differentiation results.
We only want to create derived variables for original function parameters.

Yes, we can `VarDecl` instead of auto, should I modify the code to use specific type instead ?

parth-07:
@vgvassilev I have updated the commit message to the PR description.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: I have modified code to use `ParmVarDecl*` instead of `auto`

vgvassilev (approved):
LGTM!

PR #269, opened 2021-07-25
Add support for differentiating switch statement in forward mode

parth-07:
Progress checklist:
- [x] Add support for differentiating switch statements in forward mode
- [x] Add tests

### DIFF
--- a/include/clad/Differentiator/Compatibility.h
+++ b/include/clad/Differentiator/Compatibility.h
@@ -479,6 +479,35 @@ Sema_ActOnWhileStmt(Sema& SemaRef, Sema::ConditionResult cond, Stmt* body) {
                                 /*RParenLoc=*/noLoc, body);
 #endif
 }
+
+/// Clang >= 12 has more source locations parameters in `Sema::ActOnStartOfSwitchStmt`
+static inline StmtResult
+Sema_ActOnStartOfSwitchStmt(Sema& SemaRef, Stmt* initStmt,
+                            Sema::ConditionResult Cond) {
+  SourceLocation noLoc;
+#if CLANG_VERSION_MAJOR >= 12
+  return SemaRef.ActOnStartOfSwitchStmt(
+      /*SwitchLoc=*/noLoc,
+      /*LParenLoc=*/noLoc, initStmt, Cond,
+      /*RParenLoc=*/noLoc);
+#elif CLANG_VERSION_MAJOR < 12
+  return SemaRef.ActOnStartOfSwitchStmt(/*SwitchLoc=*/noLoc, initStmt, Cond);
+#endif
+}
+
+/// Clang 9 added an extra parameter for result storage kind in
+/// ConstantExpr::Create 
+/// Clang 11 added an extra parameter for immediate invocation in
+/// ConstantExpr::Create
+#if CLANG_VERSION_MAJOR < 9
+#define CLAD_COMPAT_ConstantExpr_Create_ExtraParams
+#elif CLANG_VERSION_MAJOR < 11
+#define CLAD_COMPAT_ConstantExpr_Create_ExtraParams\
+  , Node->getResultStorageKind()
+#elif CLANG_VERSION_MAJOR >= 11
+#define CLAD_COMPAT_ConstantExpr_Create_ExtraParams\
+  , Node->getResultStorageKind(), Node->isImmediateInvocation()
+#endif
 } // namespace clad_compat
 
 #endif //CLAD_COMPATIBILITY
--- a/include/clad/Differentiator/ForwardModeVisitor.h
+++ b/include/clad/Differentiator/ForwardModeVisitor.h
@@ -77,6 +77,27 @@ namespace clad {
     StmtDiff VisitWhileStmt(const clang::WhileStmt* WS);
     StmtDiff VisitDoStmt(const clang::DoStmt* DS);
     StmtDiff VisitContinueStmt(const clang::ContinueStmt* ContStmt);
+
+    StmtDiff VisitSwitchStmt(const clang::SwitchStmt* SS);
+    StmtDiff VisitBreakStmt(const clang::BreakStmt* BS);
+    
+  private:
+    /// Helper function for differentiating the switch statement body.
+    ///
+    /// It manages scopes and blocks for the switch case labels, checks if
+    /// compound statement to be differentiated is supported and returns the
+    /// active switch case label after processing the given `stmt` argument.
+    ///
+    /// Scope and and block for the last switch case label have to be managed
+    /// manually outside the function because this function have no way of
+    /// knowing when all the statements belonging to last switch case label have
+    /// been processed.
+    ///
+    /// \param[in] stmt Current statement to derive
+    /// \param[in] activeSC Current active switch case label
+    /// \return active switch case label after processing `stmt`
+    clang::SwitchCase* DeriveSwitchStmtBodyHelper(const clang::Stmt* stmt,
+                                                  clang::SwitchCase* activeSC);
   };
 } // end namespace clad
 
--- a/include/clad/Differentiator/StmtClone.h
+++ b/include/clad/Differentiator/StmtClone.h
@@ -10,7 +10,7 @@
 
 #include "clang/AST/StmtVisitor.h"
 #include "clang/AST/RecursiveASTVisitor.h"
-
+#include "clang/Basic/Version.h"
 #include "clang/Sema/Sema.h"
 #include "clang/Sema/Scope.h"
 
@@ -113,7 +113,11 @@ namespace utils {
     DECLARE_CLONE_FN(CXXConstructExpr)
     DECLARE_CLONE_FN(CXXTemporaryObjectExpr)
     DECLARE_CLONE_FN(MaterializeTemporaryExpr)
-
+    // `ConstantExpr` node is only available after clang 7.
+    #if CLANG_VERSION_MAJOR > 7
+    DECLARE_CLONE_FN(ConstantExpr)
+    #endif
+    
     clang::Stmt* VisitStmt(clang::Stmt*);
   };
 
--- a/lib/Differentiator/ForwardModeVisitor.cpp
+++ b/lib/Differentiator/ForwardModeVisitor.cpp
@@ -17,6 +17,7 @@
 #include "clang/Sema/Lookup.h"
 #include "clang/Sema/Overload.h"
 #include "clang/Sema/Scope.h"
+#include "clang/Sema/ScopeInfo.h"
 #include "clang/Sema/Sema.h"
 #include "clang/Sema/SemaInternal.h"
 #include "clang/Sema/Template.h"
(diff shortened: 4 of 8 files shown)


include/clad/Differentiator/Compatibility.h
```diff
@@ -468,6 +468,32 @@ static inline CXXThisExpr* Sema_BuildCXXThisExpr(Sema& SemaRef,
 #endif
 }
 
+/// Clang >= 12 has more source locations parameters in `Sema::ActOnStartOfSwitchStmt`
+static inline StmtResult Sema_ActOnStartOfSwitchStmt(Sema& SemaRef, Stmt* initStmt, Sema::ConditionResult Cond) {
```
vgvassilev, line 472:
```suggestion
static inline StmtResult 
Sema_ActOnStartOfSwitchStmt(Sema& SemaRef, Stmt* initStmt, Sema::ConditionResult Cond) {
```

include/clad/Differentiator/StmtClone.h
```diff
@@ -113,6 +113,9 @@ namespace utils {
     DECLARE_CLONE_FN(CXXConstructExpr)
     DECLARE_CLONE_FN(CXXTemporaryObjectExpr)
     DECLARE_CLONE_FN(MaterializeTemporaryExpr)
+    #if CLANG_VERSION_MAJOR > 7
+    DECLARE_CLONE_FN(ConstantExpr)
```
vgvassilev, line 117: Did `ConstantExpr` appear after clang7? Maybe worth a comment.

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -1018,4 +1008,186 @@ namespace clad {
         IntegerLiteral::Create(m_Context, zero, m_Context.IntTy, noLoc);
     return StmtDiff(Clone(BL), constant0);
   }
-} // end namespace clad
\ No newline at end of file
+
+  /// returns first switch case label contained in the compound statement `CS`
+  static SwitchCase* getContainedSwitchCaseStmt(const CompoundStmt* CS) {
+    for (auto stmt : CS->body()) {
```
vgvassilev, line 1014:
```suggestion
    for (Stmt* stmt : CS->body()) {
```

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -1018,4 +1008,186 @@ namespace clad {
...
+      if (auto SC = dyn_cast<SwitchCase>(stmt))
+        return SC;
+      else if (auto nestedCS = dyn_cast<CompoundStmt>(stmt)) {
+        auto nestedRes = getContainedSwitchCaseStmt(nestedCS);
+        if (nestedRes)
```
vgvassilev, line 1019:
```suggestion
        if (SwitchCase* nestedRes = getContainedSwitchCaseStmt(nestedCS))
```

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -1018,4 +1008,186 @@ namespace clad {
...
+    return nullptr;
+  }
+
+  void setSwitchCaseSubStmt(SwitchCase* SC, Stmt* subStmt) {
+    if (auto caseStmt = dyn_cast<CaseStmt>(SC)) {
+      caseStmt->setSubStmt(subStmt);
+    } else if (auto defaultStmt = dyn_cast<DefaultStmt>(SC)) {
+      defaultStmt->setSubStmt(subStmt);
+    } else {
+      assert("Unsupported switch case statement");
```
vgvassilev, line 1032:
```suggestion
      assert(0 && "Unsupported switch case statement");
```

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -1018,4 +1008,186 @@ namespace clad {
...
+
+    // stores currently active switch case label. It is used to determine
+    // the case label corresponding to statements currently being processed.
+    // It will always be equal to the last visited case/default label.
+    SwitchCase* activeSC = nullptr;
+
+    if (auto CS = dyn_cast<CompoundStmt>(SS->getBody())) {
+      // Visit(CS) cannot be used because then we will not be easily able to
+      // determine when active switch case label should be changed.
+      for (auto stmt : CS->body()) {
```
vgvassilev, line 1077:
```suggestion
      for (Stmt* stmt : CS->body()) {
```

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -1018,4 +1008,186 @@ namespace clad {
...
+      beginBlock();
+
+      activeSC = newActiveSC;
+      DeriveSwitchStmtBodyHelper(SC->getSubStmt(), activeSC);
+    } else {
+      if (auto CS = dyn_cast<CompoundStmt>(stmt)) {
+        if (auto containedSC = getContainedSwitchCaseStmt(CS)) {
+          // FIXME: One way to support this is strategically modifying the
+          // compound statement blocks such that the meaning of code remains
+          // same and no switch case label is contained in the compound
```
vgvassilev, line 1136:
```suggestion
          // the same and no switch case label is contained in the compound
```

lib/Differentiator/StmtClone.cpp
```diff
@@ -386,6 +386,15 @@ Stmt* StmtClone::VisitDeclStmt(DeclStmt* Node) {
   return new (Ctx) DeclStmt(clonedDecls, Node->getBeginLoc(), Node->getEndLoc());
 }
 
+// clang <=7 do not have `ConstantExpr` node.
+#if CLANG_VERSION_MAJOR > 7
+Stmt* StmtClone::VisitConstantExpr(ConstantExpr* Node) {
```
vgvassilev, line 391: Why not directly using `DEFINE_CREATE_EXPR`?

test/FirstDerivative/Switch.C
```diff
@@ -0,0 +1,321 @@
...
+int main() {
+  INIT(fn1, "i");
+  INIT(fn2, "i");
+  INIT(fn3, "i");
+  INIT(fn4, "i");
+
+  TEST_SWITCH_CASES(fn1, 0, 4);   // CHECK-EXEC: 7.00 7.00 6.00 0.89 7.00
+  TEST_SWITCH_CASES(fn2, 0, 4);   // CHECK-EXEC: 2.00 0.92 2.20 7.00 7.00
+  TEST_SWITCH_CASES(fn3, 0, 2);   // CHECK-EXEC: 27.00 27.00 27.00
+  TEST_SWITCH_CASES(fn4, 0, 3);   // CHECK-EXEC: 7.00 7.20 10.00 10.00
```
vgvassilev, line 320: Did you make sure that results are expected/correct mathematically speaking?

include/clad/Differentiator/StmtClone.h
parth-07: Yes, `ConstantExpr` appeared after clang 7. I will add a comment for this.

test/FirstDerivative/Switch.C
parth-07: Yes, I have verified that the results are correct.

lib/Differentiator/StmtClone.cpp
parth-07: I did not try using `DEFINE_CREATE_EXPR`. I have modified the code to use `DEFINE_CREATE_EXPR`.

test/FirstDerivative/Switch.C
```diff
@@ -0,0 +1,321 @@
+// RUN: %cladclang %s -lm -I%S/../../include -oSwitch.out 2>&1 | FileCheck %s
```
vgvassilev, line 1: Can you also add tests for nested switch statements (eg switch-stmt in a case clause) and a fall-through switch case clauses (eg a `case:` without a `break`)?

include/clad/Differentiator/ForwardModeVisitor.h
```diff
@@ -77,6 +77,15 @@ namespace clad {
...
     StmtDiff VisitContinueStmt(const clang::ContinueStmt* ContStmt);
+
+    StmtDiff VisitSwitchStmt(const clang::SwitchStmt* SS);
+    StmtDiff VisitBreakStmt(const clang::BreakStmt* BS);
+    
+  private:
+    /// Helper function for differentiating the switch statement body, it manages 
+    /// scopes and blocks for the case labels. Scope and and block for the last switch
+    /// case label have to be popped manually outside the function.
+    void DeriveSwitchStmtBodyHelper(const clang::Stmt* stmt, clang::SwitchCase*& activeSC);
```
vgvassilev, line 88: Isn't it clearer the newActiveSC to be returned as a result of that function, rather than modifying it in-place?

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -1128,4 +1129,204 @@ namespace clad {
...
+      // Visit(CS) cannot be used because then we will not be easily able to
+      // determine when active switch case label should be changed.
+      for (Stmt* stmt : CS->body()) {
+        DeriveSwitchStmtBodyHelper(stmt, activeSC);
+      }
+    } else {
+      DeriveSwitchStmtBodyHelper(SS->getBody(), activeSC);
+    }
+
+    // scope and block of the last switch case label is not
```
vgvassilev, line 1225: Can you clarify the comment. Why can't `DeriveSwitchStmtBodyHelper` take care of this?

test/FirstDerivative/Switch.C
parth-07: I have added a test for nested switch statements, please check `fn4` in `test/FirstDerivative/Switch.C`. 
I have added a test for fall-through switch case clauses as well, please check `fn3` in `test/FirstDerivative/Switch.C`

include/clad/Differentiator/ForwardModeVisitor.h
parth-07: `DeriveSwitchStmtBodyHelper` take one statement at a time, thus, the active switch case (`activeSC`) will only change when a case label is passed to `DeriveSwitchStmtBodyHelper`. 
And since, active switch case statement needs to be passed to `DeriveSwitchStmtBodyHelper`, I preferred passing by reference, should I modify the function to return the updated value of active switch case instead?

test/FirstDerivative/Switch.C
vgvassilev: Ah, nice!

lib/Differentiator/ForwardModeVisitor.cpp
parth-07: Yes, I will include more details for this. 
It's due to how the `DeriveSwitchStmtBodyHelper` function determines when the corresponding scope and block of a case label should be popped. For all switch case labels other than the last switch case label, scope and block are popped when the next switch label has been found (scope and block of 2nd switch case label are popped when 3rd switch case label have been found, and so on). But for the last switch case label, scope and block should be popped when all statements of the switch statement body have been processed. `DeriveSwitchStmtBodyHelper` have no way of knowing when all statements of switch statement body have been visited and processed. Thus, last switch case label scope and block have to be popped outside the function.

include/clad/Differentiator/ForwardModeVisitor.h
vgvassilev: I'd prefer to see if the implementation would change significantly if we return the value. That makes the code much less subtle to the casual reader.

lib/Differentiator/ForwardModeVisitor.cpp
vgvassilev: Ok, then please clarify that in the comment and we should be good.

include/clad/Differentiator/ForwardModeVisitor.h
parth-07: Done, I have modified `DeriveSwitchStmtBodyHelper` to return the updated active switch case label instead. Please review the modified implementation.

lib/Differentiator/ForwardModeVisitor.cpp
parth-07: Done.

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -1128,4 +1129,209 @@ namespace clad {
...
+      // substatement.
+      if (activeSC) {
+        setSwitchCaseSubStmt(activeSC, endBlock());
+        endScope();
+      }
+      // sub statement will be updated later, either when the corresponding
+      // next label is visited or the corresponding switch statement ends.
+      SwitchCase* newActiveSC = nullptr;
+
+      // Not sure if cloning switch case is the right thing to do, since
```
parth-07, line 1267: @vgvassilev Please tell if cloning the switch case label is the right way to go here, or should we create switch case labels instead, this will avoid unnecessary cloning of substatement of the switch case labels.

vgvassilev: Yeah, if we don't need to clone the substatements probably we should just call `::Create` and add a relevant comment.

include/clad/Differentiator/ForwardModeVisitor.h
vgvassilev: Looks better to me. Thanks!

lib/Differentiator/ForwardModeVisitor.cpp
parth-07: Okay.

parth-07: Done, now we are creating switch case labels instead of cloning them.

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -1128,4 +1129,218 @@ namespace clad {
...
+      defaultStmt->setSubStmt(subStmt);
+    } else {
+      assert(0 && "Unsupported switch case statement");
+    }
+  }
+  
+  /// Returns top switch statement in the `SwitchStack` of the given
+  /// Function Scope.
+  static inline SwitchStmt*
+  getTopSwitchStmtOfSwitchStack(typename sema::FunctionScopeInfo* FSI) {
```
vgvassilev, line 1159: Why `typename` here?

vgvassilev: We don't need the inline here, too.

parth-07: We do not need typename here, I have removed it, sorry for the mistake
Won't making this function inline be good for performance since the function contains very few lines?

vgvassilev: No, the compiler is allowed to ignore the `inline` keyword. It is used more to signal weak linkage (when a symbol is being defined in a header file) to avoid odr violations.

include/clad/Differentiator/ForwardModeVisitor.h
```diff
@@ -77,6 +77,26 @@ namespace clad {
...
+    ///
+    /// Scope and and block for the last switch case label have to be managed
+    /// manually outside the function because this function have no way of 
+    /// knowing when all the statements belonging last switch case label have 
+    /// been processed. 
+    ///
+    /// \param[in] stmt Current statement to derive
+    /// \param[in] activeSC Current active switch case label
+    /// \return active switch case label after processing `stmt`
+    clang::SwitchCase* DeriveSwitchStmtBodyHelper(const clang::Stmt* stmt, clang::SwitchCase* activeSC);
```
vgvassilev, line 99: Can you run clang format on this?

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -1128,4 +1129,218 @@ namespace clad {
...
+      // sub statement will be updated later, either when the corresponding
+      // next label is visited or the corresponding switch statement ends.
+      SwitchCase* newActiveSC = nullptr;
+
+      // We are not cloning the switch case label here because cloning will 
+      // also unnecessary clone substatement of the switch case label.
+      if (auto newCaseSC = dyn_cast<CaseStmt>(SC)) {
+        newActiveSC = clad_compat::
+            CaseStmt_Create(m_Sema.getASTContext(),
+                            (newCaseSC->getLHS() ? Clone(newCaseSC->getLHS())
```
vgvassilev, line 1272: Please use local variables for these.

lib/Differentiator/ForwardModeVisitor.cpp
parth-07: Oh, okay, thank you so much for the detailed explanation.

include/clad/Differentiator/ForwardModeVisitor.h
parth-07: Done, now line 99 has been refactored properly.

vgvassilev (approved):
LGTM!

PR #270, opened 2021-07-26
Modify .clang-format

sudo-panda:
Make penalty for breaking after assignment equal to the penalty for breaking before first call parameter

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -11,6 +11,7 @@ PointerAlignment: Left
 
 AlignAfterOpenBracket: Align
 PenaltyBreakBeforeFirstCallParameter: 1000
+PenaltyBreakAssignment: 1000
 AlignOperands: true
 AllowShortCaseLabelsOnASingleLine: true
 IndentCaseLabels: true


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -36,21 +36,21 @@ namespace clad {
...
-                        .BuildDeclarationNameExpr(CSS,
-                                                  Back,
+                        .BuildDeclarationNameExpr(CSS, Back,
                                                   /*ADL*/ false)
                         .get();
-    Expr* Call =
-        V.m_Sema.ActOnCallExpr(V.getCurrentScope(), BackDRE, noLoc, Ref, noLoc)
-            .get();
+    Expr* Call = V.m_Sema
+                     .ActOnCallExpr(V.getCurrentScope(), BackDRE, noLoc, Ref,
```
vgvassilev, line 43: I'd expect ActOnCallExpr to be on the same and the args indented.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -36,21 +36,21 @@ namespace clad {
...
     return Call;
   }
 
   ReverseModeVisitor::CladTapeResult
   ReverseModeVisitor::MakeCladTapeFor(Expr* E) {
     assert(E && "must be provided");
-    QualType TapeType =
-        GetCladTapeOfType(getNonConstType(E->getType(), m_Context, m_Sema));
+    QualType TapeType = GetCladTapeOfType(
+        getNonConstType(E->getType(), m_Context, m_Sema));
```
vgvassilev, line 53: That change seems worse imo.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -125,8 +127,7 @@ namespace clad {
       unsigned lastArgN = m_Function->getNumParams() - 1;
       outputArrayStr = m_Function->getParamDecl(lastArgN)->getNameAsString();
     }
-    std::transform(m_Function->param_begin(),
-                   m_Function->param_end(),
+    std::transform(m_Function->param_begin(), m_Function->param_end(),
```
vgvassilev, line 130: That's better!

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -150,12 +151,8 @@ namespace clad {
...
     DeclContext* DC = const_cast<DeclContext*>(m_Function->getDeclContext());
     m_Sema.CurContext = DC;
-    DeclWithContext result = m_Builder.cloneFunction(m_Function,
-                                                     *this,
-                                                     DC,
-                                                     m_Sema,
-                                                     m_Context,
-                                                     noLoc,
+    DeclWithContext result = m_Builder.cloneFunction(m_Function, *this, DC,
+                                                     m_Sema, m_Context, noLoc,
```
vgvassilev, line 155: That's better too!

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -169,63 +166,48 @@ namespace clad {
...
+               /* No default value */ nullptr);
     if (params.back()->getIdentifier())
-      m_Sema.PushOnScopeChains(params.back(),
-                               getCurrentScope(),
+      m_Sema.PushOnScopeChains(params.back(), getCurrentScope(),
                                /*AddToContext*/ false);
 
-    llvm::ArrayRef<ParmVarDecl*> paramsRef =
-        llvm::makeArrayRef(params.data(), params.size());
+    llvm::ArrayRef<ParmVarDecl*> paramsRef = llvm::makeArrayRef(params.data(),
```
vgvassilev, line 199: That's also nicer.

lib/Differentiator/ReverseModeVisitor.cpp
sudo-panda: There is no parameter to set penalty for breaking before a `.` operator so I don't think I can fix this.

lib/Differentiator/ReverseModeVisitor.cpp
sudo-panda: I am afraid this can't be resolved without messing with the improvements. I think in this case the better thing to do would be to move the argument into a local variable.

PR #278, opened 2021-08-05
Reset derived variables of loops local variables to 0 at each iteration.

### LINKED ISSUE #277: Derived variables of variables defined in loop are not reset to 0 at each iteration in reverse mode
Minimum reproducible example:
```c++
  double a = 0;
  for (int counter = 0; counter < 3; ++counter) {
    double localVar = i;
    a += localVar;
  }
  return a;
}

int main() {
  double di[1] = {};
  clad::array_ref<double> di_ref(di, 1);
  auto d_fn = clad::gradient(fn, "i");
  d_fn.execute(3, 5, di_ref);
  std::cout<<di[0]<<"\n";
}
```
**Output:** 6
**Expected output:** 3

The root cause of the bug is derived variables of variables defined inside a loop, `_d_localVar` for the given example, are not reset to 0 at each iteration.
This problem can be easily solved by adding an extra statement to reset necessary derived variables to 0 at each iteration.

parth-07:
This PR resets derived variables of variables defined inside loops to 0 before each iteration of the loop. This is required to correctly evaluate derivatives. 

**Implemented solution:**

While differentiating a variable we can check if the variable is defined inside a loop using the `isInsideLoop` member variable, if it is, then the implemented solution adds an additional statement to assign 0 to the corresponding derived variable.

Closes #277

### DIFF
--- a/test/Gradient/Loops.C
+++ b/test/Gradient/Loops.C
@@ -383,7 +383,7 @@ double f_const(const double a, const double b) {
   return r;
 }
 
-void f_const_grad(const double, const double, double *);
+void f_const_grad(const double, const double, clad::array_ref<double>, clad::array_ref<double>);
 //CHECK:   void f_const_grad(const double a, const double b, clad::array_ref<double> _d_a, clad::array_ref<double> _d_b) {
 //CHECK-NEXT:       int _d_r = 0;
 //CHECK-NEXT:       unsigned long _t0;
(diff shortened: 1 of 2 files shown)


vgvassilev (commented):
@sudo-panda, that looks good to me. Could you take a look?

test/Gradient/Loops.C
```diff
@@ -455,5 +530,7 @@ int main() {
   result[1] = 0;
   auto f_const_d = clad::gradient(f_const);
   f_const_d.execute(2, 3, &result[0], &result[1]);
-  printf("{%.2f, %.2f}\n", result[0], result[1]); // CHECK-EXEC: {0.00, 18.00}
+  printf("{%.2f, %.2f}\n", result[0], result[1]); // CHECK-EXEC: {0.00, 12.00}
```
sudo-panda, line 533: The `f_const` test can use the `TEST_2` macro too.

test/Gradient/Loops.C
```diff
@@ -455,5 +530,7 @@ int main() {
   result[1] = 0;
```
sudo-panda, line 530: [L#386](https://github.com/vgvassilev/clad/blob/3ddc99809c004fb661edf9a3f2502c106506ec30/test/Gradient/Loops.C#L386) should be
```
void f_const_grad(const double, const double clad::array_ref<double>, clad::array_ref<double>)
```
It should have been changed by me but I must have missed it. Could you please change it?

sudo-panda (changes requested):
The PR looks good. Just some minor changes in the tests.

test/Gradient/Loops.C
parth-07: Yes sure.

test/Gradient/Loops.C
parth-07: Yes, thanks for the observation.

sudo-panda (approved):

PR #279, opened 2021-08-06
Fix overload derivative decl not being added to codegen

sudo-panda:
- Fixes linker not able to find overload code
- Fixes `setPreviousDecl` called for wrong decl

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -45,10 +45,16 @@ namespace clad {
     // otherwise it would end up in the LookupResult.
     derivedFD->getDeclContext()->addDecl(derivedFD);
 
-    if (R.empty())
-      return;
-    // Register the function on the redecl chain.
-    derivedFD->setPreviousDecl(cast<FunctionDecl>(R.getFoundDecl()));
+    for (NamedDecl* I : R) {
+      if (auto* FD = dyn_cast<FunctionDecl>(I)) {
+        if (semaRef.getASTContext()
+                .hasSameFunctionTypeIgnoringExceptionSpec(derivedFD->getType(),
+                                                          FD->getType())) {
+          // Register the function on the redecl chain.
+          derivedFD->setPreviousDecl(FD);
+        }
+      }
+    }
   }
 
   static bool hasAttribute(const Decl *D, attr::Kind Kind) {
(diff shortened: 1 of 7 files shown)


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -45,10 +45,14 @@ namespace clad {
     // otherwise it would end up in the LookupResult.
     derivedFD->getDeclContext()->addDecl(derivedFD);
 
-    if (R.empty())
-      return;
-    // Register the function on the redecl chain.
-    derivedFD->setPreviousDecl(cast<FunctionDecl>(R.getFoundDecl()));
+    for (NamedDecl* I : R) {
```
vgvassilev, line 48: I think we should either try using `Sema::CheckFunctionDeclaration` and if that does not work we should use `ASTContext::hasSameFunctionTypeIgnoringExceptionSpec(FD->getType(), derivedFD->getType())`

sudo-panda: Alright!

lib/Differentiator/VisitorBase.cpp
```diff
@@ -573,31 +573,21 @@ namespace clad {
...
+        Create(m_Context, thisExpr, /*isArrow=*/true, noLoc, NNS, noLoc, FD,
+               DAP, FD->getNameInfo(),
+               /*TemplateArgs=*/nullptr, m_Context.BoundMemberTy,
+               ExprValueKind::VK_RValue,
+               ExprObjectKind::OK_Ordinary
+               CLAD_COMPAT_CLANG9_MemberExpr_ExtraParams(NOUR_None));
+    return m_Sema
+        .BuildCallToMemberFunction(getCurrentScope(), memberExpr, noLoc,
+                                   argExprs, noLoc)
+        .get();
```
sudo-panda, line 590: @parth-07 could you please see if this implementation fails in any cases that you need it for.

parth-07: No, this implementation should work fine for cases I needed this function for.

sudo-panda: Thanks!

vgvassilev (approved):
LGTM!

vgvassilev:
codecov seems confused again.

PR #283, opened 2021-08-08
Add support for differentiating while and do-while stmts in reverse mode.

parth-07:
Progress checklist:
- [ ] Add support for differentiating while statement
- [ ] Add support for differentiating do-while statement
- [ ] Add tests


vgvassilev:
@parth-07, could you rebase that PR.

@grimmmyshini, could you take a closer look at this PR in order to parallelize the review work a little.

parth-07:
@vgvassilev Yes, I have rebased the PR.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -340,6 +342,44 @@ namespace clad {
...
+    /// Allows to easily create and manage a counter for counting the number of
+    /// executed iterations of a loop. 
+    ///
+    /// It is required to save the number of executed iterations to use the
+    /// same number of iterations in the reverse pass.
+    /// If we are currently inside a loop, then a clad tape object is created
+    /// to be used as the counter; otherwise, a temporary global variable (in
+    /// function scope) is created to be used as the counter.
+    class LoopCounter {
+      clang::Expr *m_Ref=nullptr, *m_Pop=nullptr, *m_Push=nullptr;
```
vgvassilev, line 355:
```suggestion
      clang::Expr *m_Ref=nullptr;
      clang::Expr *m_Pop=nullptr;
      clang::Expr *m_Push=nullptr;
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2114,4 +2099,205 @@ namespace clad {
...
+      // for forward-pass while statement body
+      beginScope(Scope::DeclScope);
+      beginBlock(forward);
+      addToCurrentBlock(counterIncrement);
+      bodyDiff = DifferentiateSingleStmt(body);
+      addToCurrentBlock(bodyDiff.getStmt());
+      bodyDiff = {endBlock(forward), bodyDiff.getStmt_dx()};
+      // for forward-pass while statement body
+      endScope();
+    }
```
vgvassilev, line 2255: Can you extract the common code in a separate file static function?

parth-07: This code is heavily using member variables and member functions. Can I extract the common code in a member function?

vgvassilev: Yeah.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -340,6 +342,70 @@ namespace clad {
...
+      /// for counter; otherwise, returns nullptr.
+      clang::Expr* getPush() const { return m_Push; }
+
+      /// Returns `clad::pop(_t)` expression if clad tape is used for 
+      /// for counter; otherwise, returns nullptr.
+      clang::Expr* getPop() const { return m_Pop; }
+
+      /// Returns reference to the last object of the clad tape if clad tape 
+      /// is used as the counter; otherwise returns reference to the counter
+      /// variable.
```
vgvassilev, line 372:
```suggestion
      /// is used as the counter; otherwise, returns nullptr.
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -910,36 +869,26 @@ namespace clad {
...
-    Expr* CounterCondition = m_Sema
-                                 .ActOnCondition(m_CurScope,
-                                                 noLoc,
-                                                 Counter,
-                                                 Sema::ConditionKind::Boolean)
-                                 .get()
-                                 .second;
-    Expr* CounterDecrement = BuildOp(UO_PostDec, Counter);
+    Expr*
+        CounterCondition = loopCounter.getCounterConditionResult().get().second;
```
vgvassilev, line 873: Is that what clang-format suggests here?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -910,36 +869,26 @@ namespace clad {
...
-    std::reverse(ReverseBody->body_begin(), ReverseBody->body_end());
+    Stmt* ReverseBody = BodyDiff.getStmt_dx();
     Stmt* ReverseResult = unwrapIfSingleStmt(ReverseBody);
     if (!ReverseResult)
       ReverseResult = new (m_Context) NullStmt(noLoc);
     Stmt* Reverse = new (m_Context) ForStmt(m_Context,
                                             nullptr,
                                             CounterCondition,
-                                            condVarClone,
+                                            nullptr,
```
vgvassilev, line 883: Is that the fix for issue #303? If so it's likely worth to be put in a separate commit.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2148,4 +2097,204 @@ namespace clad {
                                 /*isInsideLoop*/ false};
     }
   }
+
+  ReverseModeVisitor::LoopCounter::LoopCounter(ReverseModeVisitor& RMV)
+      : m_RMV(RMV) {
+    ASTContext& context = m_RMV.m_Context;
+    if (RMV.isInsideLoop) {
+      auto zero = ConstantFolder::synthesizeLiteral(context.getSizeType(),
+                                                    context, 0);
```
vgvassilev, line 2106: Add the name of the default parameter.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2148,4 +2097,204 @@ namespace clad {
...
+
+    llvm::SaveAndRestore<bool> SaveIsInsideLoop(isInsideLoop);
+    isInsideLoop = true;
+
+    Expr* condClone = (WS->getCond() ? Clone(WS->getCond()) : nullptr);
+    const VarDecl* condVarDecl = WS->getConditionVariable();
+    StmtDiff condVarRes;
+    if (condVarDecl) {
+      condVarRes = DifferentiateSingleStmt(WS->getConditionVariableDeclStmt());
+    }
```
vgvassilev, line 2137:
```suggestion
    if (condVarDecl)
      condVarRes = DifferentiateSingleStmt(WS->getConditionVariableDeclStmt());
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2148,4 +2103,189 @@ namespace clad {
...
+    } else {
+      // for forward-pass loop statement body
+      beginScope(Scope::DeclScope);
+      beginBlock(forward);
+      addToCurrentBlock(counterIncrement);
+      bodyDiff = DifferentiateSingleStmt(body, /*dfdS=*/nullptr,
+                                         /*shouldEmit=*/false);
+      addToCurrentBlock(bodyDiff.getStmt());
+      // Emit some statemnts later to maintain correct statement order.
+      if (m_ErrorEstimationEnabled) {
```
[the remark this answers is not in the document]
vgvassilev: ping.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2148,4 +2097,204 @@ namespace clad {
...
+    // Create reverse pass loop body statements by arranging various
+    // differentiated statements in the correct order.
+    // Order used:
+    //
+    // 1) `for` loop increment differentiation statements
+    // 2) loop body differentiation statements
+    // 3) condition variable differentiation statements
+    // 4) counter decrement expression
+    beginBlock(reverse);
+    // `for` loop have counter decrement expression in the
```
vgvassilev, line 2290:
```suggestion
    // `for` loops have counter decrement expression in the
```

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: Yes. 
What should be the ideal formatting of this line as per you? I can try to modify clang-format configuration file to improve the formatting.

include/clad/Differentiator/ReverseModeVisitor.h
parth-07: There are 2 cases to consider here. Suggested comment can change the intended meaning.
The loop counter variable is clad tape (we were already inside a loop when the loop counter is created), or it's an ordinary variable.
The first part of the comment refers to the first case - if the loop counter variable is a clad tape, then `getRef` will return a reference to the top element of the clad tape. And otherwise, if the loop counter is an ordinary variable, then `getRef` will return a reference to the variable.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: No, fix for issue #303 is done with refactoring of `VisitForStmt` to use `DifferentiateLoopBody`. 

This change here is unrelated to the fix. `condVarClone` should be used only with the forward pass of the for loop, currently, we are using it in both the forward and reverse pass, and it is causing inconsistencies in the AST of the reverse pass `for` statement but no error has been noticed due to this so far.

Inconsistency is due to the same condition variable (point to the same `VarDecl` node) being declared twice, once in the forward pass and once in the reverse pass. And also, the reverse pass does not use the condition variable.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: Okay.

lib/Differentiator/ReverseModeVisitor.cpp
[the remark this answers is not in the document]
grimmmyshini: @parth-07 yep looks good to me as of yet. Once you get this in and I address another issue, I will add more tests.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2148,4 +2097,203 @@ namespace clad {
                                 /*isInsideLoop*/ false};
     }
   }
+
+  ReverseModeVisitor::LoopCounter::LoopCounter(ReverseModeVisitor& RMV)
+      : m_RMV(RMV) {
+    ASTContext& context = m_RMV.m_Context;
```
vgvassilev, line 2103:
```suggestion
    ASTContext& C = m_RMV.m_Context;
```
This is the usual notion in the codebase, it will save also additional space.

vgvassilev (commented):
With that tiny comment, this lgtm. Sorry for the delayed review.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: Done. I have modified the `context` variable to `C`.

PR #284, opened 2021-08-10
Add changes necessary to integrate clad into root

sudo-panda:
The relevant PR for root is https://github.com/root-project/root/pull/8830

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -63,6 +63,11 @@ namespace custom_derivatives {
   }
 
   template<typename T>
+  CUDA_HOST_DEVICE T exp_darg0_darg0(T x) {
+    return exp(x);
+  }
+
+  template<typename T>
   CUDA_HOST_DEVICE T sin_darg0(T x) {
     return cos(x);
   }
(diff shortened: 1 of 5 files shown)


tools/ClangPlugin.cpp
```diff
@@ -103,6 +103,10 @@ namespace clad {
     // We cannot use HandleTranslationUnit because codegen already emits code on
     // HandleTopLevelDecl calls and makes updateCall with no effect.
     bool CladPlugin::HandleTopLevelDecl(DeclGroupRef DGR) {
+      // if HandleTopLevelDecl was called through clad we don't need to process
+      // it for diff requests
+      if (m_HandleTopLevelDeclInternal)
```
vgvassilev, line 108: We seem to have access here to clad::DiffRequest. In there you should have BaseFunctionName where we can check if the current decl is a FunctionDecl with that name? Alternatively, we can add a new member to hold the derivative name for the derivative. I am happy to have a FIXME here an more forward with the current approach.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -317,7 +317,8 @@ namespace clad {
 
   bool DiffCollector::VisitCallExpr(CallExpr* E) {
     // Check if we should look into this.
-    if (!isInInterval(E->getEndLoc()))
+    SourceLocation endLoc = E->getEndLoc();
```
vgvassilev, line 320:
```suggestion
    // FIXME: Generated code does not usually have valid source locations. In that case we should ask the enclosing ast nodes for a source location and check if it is within range.
    SourceLocation endLoc = E->getEndLoc();
```

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -62,6 +62,11 @@ namespace custom_derivatives {
     return exp(x);
   }
 
+  template<typename T>
+  CUDA_HOST_DEVICE T exp_darg0_darg0(T x) {
```
vgvassilev, line 66: Can we add test cases for these?

vgvassilev (approved):
LGTM!

PR #287, opened 2021-08-12
Fix typo in clad::tape_impl::size().

grimmmyshini:
This PR fixes a small typo in ```clad::tape_impl::size()``` that causes the calls to ```size()``` to fail.


vgvassilev (approved):
Wow, that must have had serious implications. We likely did not see it because we used in the implementation directly _size. LGTM!

PR #288, opened 2021-08-13
Update code to fix clad build warnings

parth-07:
This PR updates code to fix clad build warnings so that we can use `-Werror` for building clad.


vgvassilev (approved):
LGTM! Codecov report seems bogus.

PR #261, opened 2021-07-15
Add numerical differentiation support to clad

grimmmyshini:
This functionality will allow clad to be able to use numerical differentiation as a backup (or even standalone) in the case that a differentiation request is made for a function that is not visible to clad. Full list of possible functionalities is as given below:

**PR Checklist:**
- [x]  Add support for using numerical diff as a backup for unsupported function calls in _forward-mode_.
- [x]  Add support for using numerical diff as a backup for unsupported function calls in _reverse-mode_.
- [x]  Add support for usage of numerical diff as a standalone function (i.e. ```clad::numerical_diff```).
- [x]  Add support for numerical differentiation of multi-arg calls.
- [x]  Extend support of numerical diff to differentiate unsupported operators.
- [x]  Support printing of error information for the numerical methods. 

**Good to implement:**
- [x] Add support for non-primitive types like user structs, tensors etc.

**Known bugs:**
- [ ] Standalone code does not work because central_difference is not specialized correctly.

PR #290, opened 2021-08-13
Add support for automatic testing of reverse mode using forward mode

parth-07:
Progress checklist:
- [x] Add feature
- [ ] Modify tests

This PR adds support for automatic testing of reverse mode differentiation using forward mode differentiation. 
If additional testing is enabled then if the forward mode and reverse mode differentiation give inconsistent results then the program will fail with assert message containing information regarding the result of which function with respect to which parameter gave inconsistent results.

To enable reverse mode testing, run clad with the `-fenable-reverse-mode-testing` flag. 

For example, consider the code in `fn.cpp` file

```c++
  return i + 2*j;
}

int main() {
}
```

Generated gradient function when clad is run with `-fenable-reverse-mode-testing`: 

```c++
    double _t0;
    _t0 = j;
    double fn_return = i + 2 * _t0;
    goto _label0;
  _label0:
    {
        * _d_i += 1;
        double _r0 = 1 * _t0;
        double _r1 = 2 * 1;
        * _d_j += _r1;
    }
    if (!(* _d_i == fn_darg0(i, j)))
        __assert_fail("Inconsistent differentiation result with respect to the parameter 'i' in forward and reverse differentiation mode", "fn.cpp", 0, "fn_grad");
    if (!(* _d_j == fn_darg1(i, j)))
        __assert_fail("Inconsistent differentiation result with respect to the parameter 'j' in forward and reverse differentiation mode", "fn.cpp", 0, "fn_grad");
}
```

Current limitations of automatic testing of reverse-mode differentiation using forward mode differentiation:
- We cannot do automatic testing of array parameters, because we do not know the size of the array parameters at compile time.
  We need their size at compile time to generate forward mode derivatives for all the valid indexes of the array parameters. 
  Currently array parameters are silently skipped during testing.
  For example, consider this code,

  ```c++
  double arr[3];
  double res[3] = {};
  clad::array_ref<double> res_ref(res, 3);
  auto d_someFn = clad::gradient(someFn, "arr");
  d_someFn.execute(arr, res_ref);
  ```

  To automatically test derivatives with respect to `arr` parameter indexes, we will need to generate forward mode derivatives for 
  all the valid indexes (0 to 2) of `arr`.
  Therefore, we will need to compute `fn_darg0_0`, `fn_darg0_1` and `fn_darg0_2` at compile-time:

  But the size of the `arr` is only known at run-time when a user calls `CladFunction::execute`. Therefore we cannot create 
  forward mode derivatives of array parameters, and thus cannot do testing of array parameters. 

- Currently, reverse mode supports multi-arg function differentiation, but the forward mode does not. So we cannot automatically test functions that contain multi-arg function calls. This limitation will be lifted once we add support of multi-arg function call differentiation in the forward mode.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -201,10 +201,10 @@ namespace clad {
     return baseFilename;
   }
 
-  /// Returns specialization of `clad::EssentiallyEqual` function template
+  /// Returns specialization of `clad::VerifyResult` function template
   /// that should be used with `args`.
   static FunctionDecl*
-  getCladEssentiallyEqualSpecialization(Sema& SemaRef,
+  getCladVerifyResultSpecialization(Sema& SemaRef,
                                         llvm::ArrayRef<Expr*> args) {
     FunctionDecl* specialization = nullptr;
 
@@ -217,12 +217,12 @@ namespace clad {
     SemaRef.LookupQualifiedName(cladR, C.getTranslationUnitDecl());
     NamespaceDecl* cladND = cast<NamespaceDecl>(cladR.getFoundDecl());
 
-    // looking up `clad::EssentiallyEqual` template declaration.
-    IdentifierInfo& essentiallyEqualIdentifier = C.Idents.get(
-        "EssentiallyEqual");
+    // looking up `clad::VerifyResult` template declaration.
+    IdentifierInfo& VerifyResultIdentifier = C.Idents.get(
+        "VerifyResult");
     CXXScopeSpec SS;
     SS.Extend(C, cladND, noLoc, noLoc);
-    LookupResult R(SemaRef, DeclarationName(&essentiallyEqualIdentifier), noLoc,
+    LookupResult R(SemaRef, DeclarationName(&VerifyResultIdentifier), noLoc,
                    Sema::LookupNameKind::LookupOrdinaryName);
     // R.setTemplateNameLookup(true);
     SemaRef.LookupQualifiedName(R, cladND, SS);
@@ -586,7 +586,8 @@ namespace clad {
                                          .getFilename(m_Function->getBeginLoc())
                                          .str();
       std::string baseFileName = getBaseFilename(completeFileName);
-      std::string qualifiedGradientName = gradientFD->getQualifiedNameAsString();
+      std::string qualifiedGradientName = gradientFD
+                                              ->getQualifiedNameAsString();
       for (const auto& arg : args) {
         // We cannot do testing for the array parameters because we do not know
         // the size of the array parameters at compile time. We need their size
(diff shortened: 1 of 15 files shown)


test/Gradient/TemplateFunctors.C
```diff
@@ -178,6 +198,10 @@ template <> struct ExperimentConstVolatile<long double> {
 // CHECK-NEXT:         long double _r9 = _t9 * 1;
 // CHECK-NEXT:         * _d_i += _r9;
 // CHECK-NEXT:     }
+// CHECK-NEXT:       if (!(clad::EssentiallyEqual(* _d_i, this->operator_call_darg0(_p_i0, _p_j0))))
+// CHECK-NEXT:           clad::assert_fail("Inconsistent differentiation result with respect to the parameter 'i' in forward and reverse differentiation mode", "TemplateFunctors.C", 0, "ExperimentConstVolatile<long double>::operator_call_grad");
+// CHECK-NEXT:       if (!(clad::EssentiallyEqual(* _d_j, this->operator_call_darg1(_p_i0, _p_j0))))
+// CHECK-NEXT:           clad::assert_fail("Inconsistent differentiation result with respect to the parameter 'j' in forward and reverse differentiation mode", "TemplateFunctors.C", 0, "ExperimentConstVolatile<long double>::operator_call_grad");
```
vgvassilev, line 204: Wouldn't it be easier to synthesize a less complex expressions and move the check in the function itself? Eg this code can become a single line: `clad::VerifyEqual(...)`.

parth-07: Yes, but if we want to include the assert message function and file names in assertions, then we will also need to pass them in the call to `clad::VerifyEqual`. Wouldn't it be unintuitive to pass these details to a function for verifying results?

vgvassilev: I would not consider that not intuitive.

parth-07: Okay, then I will create and use `clad::VerifyEqual(...)` to do the verification and give assert if verification fails.

vgvassilev:
Just heads up that we might also want to include #261 as a third way to cross check.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -483,6 +575,52 @@ namespace clad {
     // given it is not a DeclRefExpr.
     if (m_ErrorEstimationEnabled)
       errorEstHandler->EmitFinalErrorStmts(params, m_Function->getNumParams());
```
vgvassilev, line 577: I think we should move that code as a separate callback class in the same way as we have done for the error estimation handler. Eg, we should create a multiplexing `errorEstHandler` (which is a misnomer for ReverseModeCallbacks) and implement the relevant code there. Then in case the option is specified we will just construct the callback and attach it. We will need the multiplexer when we have an error estimation model handler and the automatic testing. You can take inspiration from `MultiplexExternalSemaSource`. 

cc: @grimmmyshini

PR #291, opened 2021-08-15
Add tests for ROOT interface

sudo-panda:
- Added hessian mode test in TFormula.C
- Added test for array_ref interface in ROOT


vgvassilev (approved):
LGTM!

PR #298, opened 2021-08-20
Modify MoveData in Tape.h to make it CUDA compatible

sudo-panda:
- Change `std::address` to `__builtin_addressof`
- Change `fprintf` to `printf`
- Add `trap` function to replace `exit` function

### DIFF
--- a/include/clad/Differentiator/Tape.h
+++ b/include/clad/Differentiator/Tape.h
@@ -100,13 +100,13 @@ namespace clad {
       // allocation properly.
       for (; first != last; ++first, (void)++current) {
         auto new_data = ::new (const_cast<void*>(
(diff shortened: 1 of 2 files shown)


include/clad/Differentiator/Tape.h
```diff
@@ -100,13 +100,13 @@ namespace clad {
       // allocation properly.
       for (; first != last; ++first, (void)++current) {
         auto new_data = ::new (const_cast<void*>(
-            static_cast<const volatile void*>(std::addressof(*current))))
+            static_cast<const volatile void*>(current)))
```
vgvassilev, line 103: @grimmmyshini, could you take a look at this change?

vgvassilev (approved):
LGTM!

PR #302, opened 2021-08-25
Fix crash due to null operands to BuildOp.

### LINKED ISSUE #245: Array index expression cause segfault in reverse mode
Having a ```BO_Sub``` expression as an array subscript index expression causes a null pointer exception in reverse mode. Consider the following case:

```cpp
float running_sum(float* p, int n){
  for (int i = 1; i < n; i++){
    p[i] += p[i - 1] ;      <-- here since " i-1 " is being used as a index expression, it has no dfdx() 
  }
  return p[n-1];
}
```

*Cause:*

In ```ReverseModeVisitor::VisitBinaryOperator``` while visiting the subtaction expression
```cpp
else if (opCode == BO_Sub) {
      // xi = xl - xr
      // dxi/xl = 1.0
      // df/dxl += df/dxi * dxi/xl = df/dxi
      Ldiff = Visit(L, dfdx());
      // dxi/xr = -1.0
      // df/dxl += df/dxi * dxi/xr = -df/dxi
      auto dr = BuildOp(UO_Minus, dfdx());   <-- here dfdx() is null
      Rdiff = Visit(R, dr);
}
```
```m_Sema.BuildUnaryOp``` (which is called by ```BuildOp```)  does not internally check for any null values before dereferencing the pointer, hence causing this crash.

*Proposed solution:*

Just do a simple not null check in ```BuildOp```. (Note: ```m_Sema.BuildBinOp``` does have a usability check internally and hence it does not require such an addition)

If the proposed solution is acceptable, I'd like to put a PR in!

### PR DESCRIPTION
**Pull request by grimmmyshini**
Closes #245 .

### CODE REVIEW

File: lib/Differentiator/VisitorBase.cpp [hunk 1]
```diff
@@ -418,11 +418,15 @@ namespace clad {
...
                              SourceLocation OpLoc) {
+    if (!E)
+      return nullptr;
     return m_Sema.BuildUnaryOp(nullptr, OpLoc, OpCode, E).get();
   }
 
   Expr* VisitorBase::BuildOp(clang::BinaryOperatorKind OpCode, Expr* L, Expr* R,
                              SourceLocation OpLoc) {
+    if (!L || !R)
+      return nullptr;
```
vgvassilev, line 429: Can you add a test for this.

PR #304, opened 2021-09-07
Add initial support for reverse mode visitor plugin system

parth-07:
This PR adds initial support for the reverse mode visitor plugin system. 

Reverse mode visitor plugins are external sources that can provide additional functionality to the reverse mode visitor by utilising callbacks at crucial locations. 

Here we are making an assumption that most plugins will only require modifying the behaviour of reverse mode at few crucial locations that are less in number and spread throughout the `ReverseModeVisitor` class. Reverse mode visitor plugins should only be used for adding relatively simple features.

## Implementation details

Basic design is inspired by [ExternalSemaSource](https://clang.llvm.org/doxygen/classclang_1_1ExternalSemaSource.html) and [MultiplexExternalSemaSource](https://clang.llvm.org/doxygen/MultiplexExternalSemaSource_8cpp.html) class.

Callbacks are available of the following flavours:

- `ActOnStartOf(Event)` and `ActOnEndOf(Event)`
   Here _Event_ should be a `ReverseModeVisitor` member function. The behaviour of these 2 flavours of events can be 
   described as:

   ```
   ReturnType ReverseModeVisitor::MemberFunction(...) {
     ActOnStartOf(Event)
     fn step 1
     fn step 2
     ...
     ...
     fn step N
     ActOnEndOf(Event)
   }

- `ActBefore(Event)` and `ActAfter(Event)`
   Here _Event_ should be an event containing single or multiple steps. The behaviour of these 2 flavours can be 
   described as:

   ```
   ActBefore(Event)
   Event step 1
   Event step 2
   ...
   ...
   Event step N
   ActAfter(Event)
   ```
- `ActBeforeFinalising(Event)`
   Again, here event is an event that contains single or multiple steps. Finalising an event means the cleanup/adding 
   statements to the current block/Getting current forward and reverse block /returning statements or anything 
   that can be roughly fit under these categories. An example of this flavour of event:

   ```
   ActBeforeFinalising(Event)
   Stmt* Forward = unwrapIfSingleStmt(endBlock(direction::forward));
   Stmt* Reverse = unwrapIfSingleStmt(BranchDiff.getStmt_dx());
   return StmtDiff(Forward, Reverse);
   ```

Any new callbacks that need to be added should also be of one of the flavours described above to maintain consistency in the plugin callbacks API.

## Usage details 

To create a reverse mode visitor plugin, a new class that inherits from `ExternalRMVSource` class should be created to describe the plugin behaviour.
This plugin class should override any callback functions that it needs to implement the required functionality. 

Any number of plugins can be registered to a reverse mode visitor object using the `ReverseModeVisitor::AddExternalSource` member function. 

Internally, reverse mode visitor uses a `MultiplexExternalRMVSource` object for this purpose. For each crucial location, multiplexer calls the callback function for each plugin in the same sequence in which the plugins were registered. 

Ideally, plugins functionality should be isolated and independent from each other. 

Please feel free to give any suggestions for improving the reverse mode plugin system.

### DIFF
--- a/include/clad/Differentiator/ErrorEstimator.h
+++ b/include/clad/Differentiator/ErrorEstimator.h
@@ -2,232 +2,278 @@
 #define CLAD_ERROR_ESTIMATOR_H
 
 #include "EstimationModel.h"
-#include "ReverseModeVisitor.h"
+#include "clad/Differentiator/ExternalRMVSource.h"
 #include "clad/Differentiator/ReverseModeVisitorDirectionKinds.h"
+
+#include "clang/AST/OperationKinds.h"
+
+#include <stack>
+
 namespace clang {
-  class Stmt;
-  class Expr;
-  class Decl;
-  class VarDecl;
-  class ParmVarDecl;
-  class FunctionDecl;
-  class CompoundStmt;
+class Stmt;
+class Expr;
+class Decl;
+class VarDecl;
+class ParmVarDecl;
+class FunctionDecl;
+class CompoundStmt;
 } // namespace clang
 
 namespace clad {
-  /// The estimation handler which interfaces with Clad's reverse mode visitors
-  /// to fetch derivatives.
-  class ErrorEstimationHandler : public ReverseModeVisitor {
-    /// Keeps a track of the delta error expression we shouldn't emit.
-    bool m_DoNotEmitDelta;
-    /// Reference to the final error parameter in the augumented target
-    /// function.
-    clang::Expr* m_FinalError;
-    /// Reference to the return error expression.
-    clang::Expr* m_RetErrorExpr;
-    /// An instance of the custom error estimation model to be used.
-    FPErrorEstimationModel* m_EstModel; // We do not own this.
-    /// A set of assignments resulting for declaration statments.
-    VisitorBase::Stmts m_ForwardReplStmts;
-    /// A vector to keep track of error statements for delayed emission.
-    VisitorBase::Stmts m_ReverseErrorStmts;
-    /// The index expression for emitting final errors for input param errors.
-    clang::Expr* m_IdxExpr;
-    /// A set of declRefExprs for parameter value replacements.
-    std::unordered_map<const clang::VarDecl*, clang::Expr*> m_ParamRepls;
+/// The estimation handler which interfaces with Clad's reverse mode visitors
+/// to fetch derivatives.
+class ErrorEstimationHandler : public ExternalRMVSource {
+  // FIXME: Find a good way to modularize using declarations that are used in
+  // multiple header files. 
+  // `Stmts` is originally defined in `VisitorBase`.
+  using Stmts = llvm::SmallVector<clang::Stmt*, 16>;
+  /// Keeps a track of the delta error expression we shouldn't emit.
+  bool m_DoNotEmitDelta;
+  /// Reference to the final error parameter in the augumented target
+  /// function.
+  clang::Expr* m_FinalError;
+  /// Reference to the return error expression.
+  clang::Expr* m_RetErrorExpr;
+  /// An instance of the custom error estimation model to be used.
+  FPErrorEstimationModel* m_EstModel; // We do not own this.
+  /// A set of assignments resulting for declaration statments.
+  Stmts m_ForwardReplStmts;
+  /// A vector to keep track of error statements for delayed emission.
+  Stmts m_ReverseErrorStmts;
+  /// The index expression for emitting final errors for input param errors.
+  clang::Expr* m_IdxExpr;
+  /// A set of declRefExprs for parameter value replacements.
+  std::unordered_map<const clang::VarDecl*, clang::Expr*> m_ParamRepls;
 
-  public:
-    ErrorEstimationHandler(DerivativeBuilder& builder)
-        : ReverseModeVisitor(builder), m_DoNotEmitDelta(false),
-          m_FinalError(nullptr), m_RetErrorExpr(nullptr), m_EstModel(nullptr),
-          m_IdxExpr(nullptr) {}
-    ~ErrorEstimationHandler() {}
+  std::stack<bool> m_ShouldEmit;
+  ReverseModeVisitor* m_RMV;
+  clang::Expr* m_DeltaVar = nullptr;
+  llvm::SmallVectorImpl<clang::QualType>* m_ParamTypes = nullptr;
+  llvm::SmallVectorImpl<clang::ParmVarDecl*>* m_Params = nullptr;
 
-    /// Function to set the error estimation model currently in use.
-    ///
-    /// \param[in] estModel The error estimation model, can be either
-    /// an in-built one (TaylorApprox) or one provided by the user.
-    void SetErrorEstimationModel(FPErrorEstimationModel* estModel);
+public:
+  using direction = rmv::direction;
+  ErrorEstimationHandler()
+      : m_DoNotEmitDelta(false), m_FinalError(nullptr), m_RetErrorExpr(nullptr),
+        m_EstModel(nullptr), m_IdxExpr(nullptr) {}
+  ~ErrorEstimationHandler() {}
 
-    /// \param[in] finErrExpr The final error expression.
-    void SetFinalErrorExpr(clang::Expr* finErrExpr) {
-      m_FinalError = finErrExpr;
-    }
+  /// Function to set the error estimation model currently in use.
+  ///
+  /// \param[in] estModel The error estimation model, can be either
+  /// an in-built one (TaylorApprox) or one provided by the user.
+  void SetErrorEstimationModel(FPErrorEstimationModel* estModel);
 
-    /// Shorthand to get array subscript expressions.
-    ///
-    /// \param[in] arrBase The base expression of the array.
-    /// \param[in] idx The index expression.
-    /// \param[in] isCladSpType Keeps track of if we have to build a clad
-    /// special type (i.e. clad::Array or clad::ArrayRef).
-    ///
-    /// \returns An expression of the kind arrBase[idx].
-    clang::Expr* getArraySubscriptExpr(clang::Expr* arrBase, clang::Expr* idx,
-                                       bool isCladSpType = true);
+  /// \param[in] finErrExpr The final error expression.
+  void SetFinalErrorExpr(clang::Expr* finErrExpr) { m_FinalError = finErrExpr; }
 
-    /// \returns The final error expression so far.
-    clang::Expr* GetFinalErrorExpr() { return m_FinalError; }
+  /// Shorthand to get array subscript expressions.
+  ///
+  /// \param[in] arrBase The base expression of the array.
+  /// \param[in] idx The index expression.
+  /// \param[in] isCladSpType Keeps track of if we have to build a clad
+  /// special type (i.e. clad::Array or clad::ArrayRef).
+  ///
+  /// \returns An expression of the kind arrBase[idx].
+  clang::Expr* getArraySubscriptExpr(clang::Expr* arrBase, clang::Expr* idx,
+                                     bool isCladSpType = true);
 
-    /// Function to build the final error statemnt of the function. This is the
-    /// last statement of any target function in error estimation and
-    /// aggregates the error in all the registered variables.
-    void BuildFinalErrorStmt();
+  /// \returns The final error expression so far.
+  clang::Expr* GetFinalErrorExpr() { return m_FinalError; }
 
-    /// Function to emit error statements into the derivative body.
-    ///
-    /// \param[in] var The variable whose error statement we want to emit.
-    /// \param[in] deltaVar The "_delta_" expression of the variable 'var'.
-    /// \param[in] errorExpr The error expression (LHS) of the variable 'var'.
-    /// \param[in] isInsideLoop A flag to indicate if 'val' is inside a loop.
-    void AddErrorStmtToBlock(clang::Expr* var, clang::Expr* deltaVar,
-                             clang::Expr* errorExpr, bool isInsideLoop = false);
+  /// Function to build the final error statemnt of the function. This is the
+  /// last statement of any target function in error estimation and
+  /// aggregates the error in all the registered variables.
+  void BuildFinalErrorStmt();
 
-    /// Emit the error estimation related statements that were saved to be
-    /// emitted at later points into specific blocks.
-    ///
-    /// \param[in] d The direction of the block in which to emit the
-    /// statements.
-    void EmitErrorEstimationStmts(direction d = direction::forward);
+  /// Function to emit error statements into the derivative body.
+  ///
+  /// \param[in] var The variable whose error statement we want to emit.
+  /// \param[in] deltaVar The "_delta_" expression of the variable 'var'.
+  /// \param[in] errorExpr The error expression (LHS) of the variable 'var'.
+  /// \param[in] isInsideLoop A flag to indicate if 'val' is inside a loop.
+  void AddErrorStmtToBlock(clang::Expr* var, clang::Expr* deltaVar,
+                           clang::Expr* errorExpr, bool isInsideLoop = false);
 
-    /// We should save the return value if it is an arithmetic expression,
-    /// since we also need to take into account the error in that expression.
-    ///
-    /// \param[in] retExpr The return expression.
-    /// \param[in] retDeclRefExpr The temporary value in which the return
-    /// expression is stored.
-    void SaveReturnExpr(clang::Expr* retExpr,
-                        clang::DeclRefExpr* retDeclRefExpr);
+  /// Emit the error estimation related statements that were saved to be
+  /// emitted at later points into specific blocks.
+  ///
+  /// \param[in] d The direction of the block in which to emit the
+  /// statements.
+  void EmitErrorEstimationStmts(direction d = direction::forward);
 
-    /// Emit the error for parameters of nested functions.
-    ///
-    /// \param[in] fnDecl The function declaration of the nested function.
-    /// \param[in] CallArgs The orignal call arguments of the function call.
-    /// \param[in] ArgResultDecls The differentiated call arguments.
-    /// \param[in] numArgs The number of call args.
-    void EmitNestedFunctionParamError(
-        clang::FunctionDecl* fnDecl,
-        llvm::SmallVectorImpl<clang::Expr*>& CallArgs,
-        llvm::SmallVectorImpl<clang::VarDecl*>& ArgResultDecls, size_t numArgs);
+  /// We should save the return value if it is an arithmetic expression,
+  /// since we also need to take into account the error in that expression.
+  ///
+  /// \param[in] retExpr The return expression.
+  /// \param[in] retDeclRefExpr The temporary value in which the return
+  /// expression is stored.
+  void SaveReturnExpr(clang::Expr* retExpr, clang::DeclRefExpr* retDeclRefExpr);
 
-    /// Save values of registered variables so that they can be replaced
-    /// properly in case of re-assignments.
-    ///
-    /// \param[in] val The value to save.
-    /// \param[in] isInsideLoop A flag to indicate if 'val' is inside a loop.
-    ///
-    /// \returns The saved variable and its derivative.
-    StmtDiff SaveValue(clang::Expr* val, bool isInLoop = false);
+  /// Emit the error for parameters of nested functions.
+  ///
+  /// \param[in] fnDecl The function declaration of the nested function.
+  /// \param[in] CallArgs The orignal call arguments of the function call.
+  /// \param[in] ArgResultDecls The differentiated call arguments.
+  /// \param[in] numArgs The number of call args.
+  void EmitNestedFunctionParamError(
+      clang::FunctionDecl* fnDecl,
+      llvm::SmallVectorImpl<clang::Expr*>& CallArgs,
+      llvm::SmallVectorImpl<clang::VarDecl*>& ArgResultDecls, size_t numArgs);
 
-    /// Save the orignal values of the input parameters in case of
-    /// re-assignments.
-    ///
-    /// \param[in] paramRef The DeclRefExpr of the input parameter.
-    void SaveParamValue(clang::DeclRefExpr* paramRef);
+  /// Save values of registered variables so that they can be replaced
+  /// properly in case of re-assignments.
+  ///
+  /// \param[in] val The value to save.
+  /// \param[in] isInsideLoop A flag to indicate if 'val' is inside a loop.
+  ///
+  /// \returns The saved variable and its derivative.
+  StmtDiff SaveValue(clang::Expr* val, bool isInLoop = false);
 
-    /// Register variables to be used while accumulating error.
-    /// Register variable declarations so that they may be used while
-    /// calculating the final error estimates. Any unregistered variables will
-    /// not be considered for the final estimation.
-    ///
-    /// \param[in] VD The variable declaration to be registered.
-    /// \param[in] toCurrentScope Add the created "_delta_" variable declaration
-    /// to the current scope instead of the global scope.
-    ///
-    /// \returns The Variable declaration of the '_delta_' prefixed variable.
-    clang::Expr* RegisterVariable(clang::VarDecl* VD,
-                                  bool toCurrentScope = false);
+  /// Save the orignal values of the input parameters in case of
+  /// re-assignments.
+  ///
+  /// \param[in] paramRef The DeclRefExpr of the input parameter.
+  void SaveParamValue(clang::DeclRefExpr* paramRef);
 
-    /// Checks if a variable can be registered for error estimation.
-    ///
-    /// \param[in] VD The variable declaration to be registered.
-    ///
-    /// \returns True if the variable can be registered, false otherwise.
-    bool CanRegisterVariable(clang::VarDecl* VD);
+  /// Register variables to be used while accumulating error.
+  /// Register variable declarations so that they may be used while
+  /// calculating the final error estimates. Any unregistered variables will
+  /// not be considered for the final estimation.
+  ///
+  /// \param[in] VD The variable declaration to be registered.
+  /// \param[in] toCurrentScope Add the created "_delta_" variable declaration
+  /// to the current scope instead of the global scope.
+  ///
+  /// \returns The Variable declaration of the '_delta_' prefixed variable.
+  clang::Expr* RegisterVariable(clang::VarDecl* VD,
+                                bool toCurrentScope = false);
 
-    /// Calculate aggregate error from m_EstimateVar.
-    /// Builds the final error estimation statement.
-    clang::Stmt* CalculateAggregateError();
+  /// Checks if a variable can be registered for error estimation.
+  ///
+  /// \param[in] VD The variable declaration to be registered.
+  ///
+  /// \returns True if the variable can be registered, false otherwise.
+  bool CanRegisterVariable(clang::VarDecl* VD);
 
-    /// Get the underlying DeclRefExpr type it it exists.
-    ///
-    /// \param[in] expr The expression whose DeclRefExpr is requested.
-    ///
-    /// \returns The DeclRefExpr of input or null.
-    clang::DeclRefExpr* GetUnderlyingDeclRefOrNull(clang::Expr* expr);
+  /// Calculate aggregate error from m_EstimateVar.
+  /// Builds the final error estimation statement.
+  clang::Stmt* CalculateAggregateError();
 
-    /// Get the parameter replacement (if any).
-    ///
-    /// \param[in] VD The parameter variable declaration to get replacement
-    /// for.
-    ///
-    /// \returns The underlying replaced Expr.
-    clang::Expr* GetParamReplacement(const clang::ParmVarDecl* VD);
+  /// Get the underlying DeclRefExpr type it it exists.
+  ///
+  /// \param[in] expr The expression whose DeclRefExpr is requested.
+  ///
+  /// \returns The DeclRefExpr of input or null.
+  clang::DeclRefExpr* GetUnderlyingDeclRefOrNull(clang::Expr* expr);
 
-    /// An abstraction of the error estimation model's AssignError.
-    ///
-    /// \param[in] val The variable to get the error for.
-    /// \param[in] valDiff The derivative of the variable 'var'
-    ///
-    /// \returns The error in the variable 'var'.
-    clang::Expr* GetError(clang::Expr* var, clang::Expr* varDiff) {
-      return m_EstModel->AssignError({var, varDiff});
-    }
+  /// Get the parameter replacement (if any).
+  ///
+  /// \param[in] VD The parameter variable declaration to get replacement
+  /// for.
+  ///
+  /// \returns The underlying replaced Expr.
+  clang::Expr* GetParamReplacement(const clang::ParmVarDecl* VD);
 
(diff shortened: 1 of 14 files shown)


vgvassilev:
Could you rebase this PR?

parth-07:
> Could you rebase this PR?

Yes! I have rebased the PR. Sorry for the delay.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -31,7 +30,7 @@
 using namespace clang;
 
 namespace clad {
-
+  // Is there any particular reason for creating 'errorEstHandler` a global variable?
```
vgvassilev, line 33: That's probably a question for @grimmmyshini.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -209,6 +221,7 @@ namespace clad {
     if (args.empty())
       return {};
 
+    m_ExternalSource->ActAfterParsingDiffArgs(request, args);
```
vgvassilev, line 224: I think we should always check if we have `m_ExternalSource` for the cases where no listeners are attached.

vgvassilev (commented):
I think this is heading in a good direction!

I have a couple of high-level comments:
  * We should try to increase the test coverage - that may happen if we move the custom error estimation model to use the new API (if we have not).
  * In order to claim success in disentangling the classes, we should be check if now we can hide headers such as `ReverseModeVisitor.h` into the source directory.
  * We will probably need another round of review from @grimmmyshini

lib/Differentiator/DerivativeBuilder.cpp
grimmmyshini: It was because it is used in 2 translation units (cpp files) this and ReverseModeVisitor.

grimmmyshini:
Okay, I did have a look and here are my final thoughts: 

I think as long as the tests pass everything should be good, I also went through the code and everything looks fine there as well. I still believe we have created something very rigid and specialized to ErrorEstimationHandler that might not be extremely useful to the next person who wants to build their own plugins with clad. However, if the interdependency issue is really significant, then maybe it is a good idea to keep this despite the tradeoffs.

lib/Differentiator/DerivativeBuilder.cpp
parth-07: Can we make it a member variable of `DerivativeBuilder` in this case?

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -19,14 +20,21 @@
...
   /// Used to compute derivatives by clad::gradient.
   class ReverseModeVisitor
       : public clang::ConstStmtVisitor<ReverseModeVisitor, StmtDiff>,
         public VisitorBase {
+  
   private:
+    friend class ErrorEstimationHandler;
     /// Determines if an error estimation is in process; helps decide whether
     /// to visit error estimation specific code in calls to VisitStmt.
+    // TODO: remove this member variable after converting error estimator to callback.
```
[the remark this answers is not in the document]
parth-07: I totally forgot about this `todo`, I think we can safely remove the `m_ErrorEstimationEnabled` member variable now.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: `m_ExternalSource` is of type `MultiplexExternalRMVSource`. So, if no listeners are attached it basically does nothing. 
`m_ExternalSource->CallbackFunction` calls `CallbackFunction` by each attached listener in the order they were attached.

parth-07: Should we add an `if` check for the `m_ExternalSource` member function calls?

include/clad/Differentiator/ReverseModeVisitor.h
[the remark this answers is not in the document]
parth-07: I have removed the `m_ErrorEstimationEnabled` member variable.

lib/Differentiator/DerivativeBuilder.cpp
grimmmyshini: I think the scope is reduced now so you can do that.

grimmmyshini:
> I do believe the plugin system does need several improvements to be useful and easy to use and that will be an iterative process. Most importantly, we need to improve the hook naming and standardize the hook places. But the reason it is very rigid and specialized to `ErrorEstimationHandler` is: currently that's the only place where it's used.

This is not what I mean, by specialized I mean users will likely have to edit the source to modify the function parameters. Usually, implementations like these are successful when some sort of state transfer happens, or when clients willingly subscribe to the events they want. These "state" transfers usually happen through shared memory or some sort of message passing system. Now, take into account this scenario: 

A user comes in and decides to want to implement a plugin using clad, it is very unlikely that they will want to use ```ActBeforeCreatingDerivedFnParamTypes(unsigned& numExtraParam)``` with that specific signature itself, then likely that they will have to add or remove from the rmv class multiple times to get to a configuration they want.  This will eventually lead to one of two things:- the code becomes bloated or the user just ends up adding things directly to where they want in their own forks.

In my very humble opinion, I do believe this is a good idea, the only doubt I have is maybe clad just is not the right tool for it. I can obviously be proved wrong here though.

If we only want to keep implementations separate, we can dedicate a branch for extra stuff or perhaps even a whole new repository so as to keep the main branch clean and concise with a nice and robust AD tool. This would also allow us to reuse some aspects of your implementation here but I think they will be reversed - instead of the requests served by the plugin, they are served by clad.

> The more we will use the plugin system the more it should improve. Two immediate use cases of the plugin system are: Cross verification of the reverse mode results using the forward mode and cross-verification of automatic differentiation results using the numerical differentiation.

What I said before applies here as well, I feel that this will grow more and more complicated as more plugins are added. Something that may or may not be detrimental to this tool.

>  cross-verification of automatic differentiation results using the numerical differentiation

Not sure how the plugin system comes to play here, numerical diff is a target program run time facility. The only thing you can do with that is compare at the end with a floating-point sensitive ```==```. Anything else you will need very drastic changes to numerical differentiation and/or the plugin system. Not to also mention that numerical diff as a concept is worse than AD in terms of accuracy and numerical stability so the best you can do is build 3 lines of code and inject it into the main program. I would suggest to not rely heavily on this if you are planning to.

parth-07:
> This is not what I mean, by specialized I mean users will likely have to edit the source to modify the function parameters. 

Yes, I totally agree with this. And thus, the plugin system has been designed to be used internally only, and I don't believe it will be useful for users to use the plugin system directly at all. 

The main benefit of the plugin system that I see is the separation of concerns. `ReverseModeVisitor` should only focus on automatic differentiation, but the way we are going right now, we are adding more and more stuff to it that are unrelated to automatic differentiation. The plugin system provides a (unstandard) way of refactoring the `ReverseModeVisitor` codebase. I don't think it's the best idea for our project, but I do think it's better than filling `ReverseModeVisitor` with more and more unrelated things. 

> If we only want to keep implementations separate, we can dedicate a branch for extra stuff or perhaps even a whole new repository so as to keep the main branch clean and concise with a nice and robust AD tool.

I am totally open to this idea, but I don't see how we can separate Error Estimation Framework in a separate branch/repository. It is a tightly coupled framework with the `ReverseModeVisitor`. It is also not feasible or practical to maintain multiple branches/repositories containing `ReverseModeVisitor` implementation.

> What I said before applies here as well, I feel that this will grow more and more complicated as more plugins are added. Something that may or may not be detrimental to this tool.

If it grows more and more that implies we needed to add more and more stuff unrelated to automatic differentiation inside `ReverseModeVisitor`. Thus, using the plugin system will certainly reduce the amount of code inside `ReverseModeVisitor`.
 
> Not sure how the plugin system comes to play here, numerical diff is a target program run time facility. The only thing you can do with that is compare at the end with a floating-point sensitive ==.

We can use a plugin system to modify the derived functions such that they cross-check results (derivatives) using forward mode and/or numerical differentiation. 
Comparing at the end is all we need most probably. Using the plugin system provides a way to perform this cross-checking without any additional code/dependency. All we will need to do to enable the cross-checking will be to pass appropriate flags to Clad. 

> Not to also mention that numerical diff as a concept is worse than AD in terms of accuracy and numerical stability.

Yes. But the fact that I was concentrating on here was: automatic differentiation implementations are generally more error-prone. Cross-checking will allow us to catch bugs early on in the development process.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -157,16 +156,18 @@ namespace clad {
...
       // Set the handler.
-      errorEstHandler.reset(new ErrorEstimationHandler(*this));
+      errorEstHandler.reset(new ErrorEstimationHandler());
       // Set error estimation model. If no custom model provided by user,
       // use the built in Taylor approximation model.
       if (!m_EstModel) {
         m_EstModel.reset(new TaylorApprox(*this));
       }
       errorEstHandler->SetErrorEstimationModel(m_EstModel.get());
+      R.AddExternalSource(*errorEstHandler);
```
vgvassilev, line 168: This operation should be already handled outside. So we can remove this line.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -157,16 +156,18 @@ namespace clad {
...
       // Finally begin estimation.
-      result = errorEstHandler->Derive(FD, request);
+      result = R.Derive(FD, request);
```
vgvassilev, line 170: That operation should be removed, it will happen naturally.

lib/Differentiator/ErrorEstimator.cpp
```diff
@@ -1,4 +1,5 @@
 #include "clad/Differentiator/ErrorEstimator.h"
+#include "clad/Differentiator/CladUtils.h"
```
vgvassilev, line 2:
```suggestion

#include "clad/Differentiator/CladUtils.h"
```

lib/Differentiator/ErrorEstimator.cpp
```diff
@@ -18,23 +19,25 @@ namespace clad {
...
+                                           noLoc);
       finExpr = m_EstModel->AssignError(StmtDiff(m_RetErrorExpr, flitr));
     }
 
     // Build the final error statement with the sum of all _delta_*.
     Expr* addErrorExpr = m_EstModel->CalculateAggregateError();
     if (addErrorExpr) {
       if (finExpr)
-        addErrorExpr = BuildOp(BO_Add, addErrorExpr, finExpr);
+        addErrorExpr = m_RMV->BuildOp(BO_Add, addErrorExpr, finExpr);
```
vgvassilev, line 32: I suspect we will need to move all Build operations which do not really depend on RMV into a separate file (as we have discussed). I think it is ok to be in a following PR.

lib/Differentiator/ErrorEstimator.cpp
```diff
@@ -273,70 +290,76 @@ namespace clad {
...
+          if (m_RMV->isArrayRefType(LdiffExpr->getType())) {
+            Ldiff = m_RMV->m_Sema
+                        .ActOnArraySubscriptExpr(m_RMV->getCurrentScope(),
+                                                 LdiffExpr, decl->getLocation(),
+                                                 m_IdxExpr, noLoc)
                         .get();
           } else {
-            Ldiff = m_Sema
+            Ldiff = m_RMV->m_Sema
                         .CreateBuiltinArraySubscriptExpr(LdiffExpr, noLoc,
```
vgvassilev, line 322: @grimmmyshini, can you help with the test coverage issues exposed by this PR in the error estimation?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1698,29 +1677,27 @@ namespace clad {
...
       while(!m_PopIdxValues.empty())
-        addToCurrentBlock(m_PopIdxValues.pop_back_val(), reverse);
-      Expr* deltaVar = nullptr;
-      if (m_ErrorEstimationEnabled)
-        deltaVar = errorEstHandler->RegisterBinaryOpLHS(LCloned, R,
-                                                        /*isAssign=*/opCode ==
-                                                            BO_Assign);
+        addToCurrentBlock(m_PopIdxValues.pop_back_val(), direction::reverse);
+
+      m_ExternalSource->ActAfterCloningLHSOfAssignOp(LCloned, R, opCode);
```
vgvassilev, line 1686: I think we should check if there is an external source and only then call it (saving a virtual call).

vgvassilev (commented):
I think this is heading in a very promising way.

vgvassilev:
> > This is not what I mean, by specialized I mean users will likely have to edit the source to modify the function parameters.
> 
> Yes, I totally agree with this. And thus, the plugin system has been designed to be used internally only, and I don't believe it will be useful for users to use the plugin system directly at all.

One very easy to implement and quite different tool to implement with the proposed design is a trace-based debugging. For example if the differentiation crashes/fails we can ask the users to attach a "plugin" that prints to disk the each call and attach the output.

@parth-07, can you rebase this PR and run clang-format on it?

include/clad/Differentiator/ErrorEstimator.h
```diff
@@ -215,8 +226,36 @@ namespace clad {
     /// \param[in] isInsideLoop A flag to keep track of if we are inside a
     /// loop.
     void EmitDeclErrorStmts(VarDeclDiff VDDiff, bool isInsideLoop);
-  };
 
+    void InitialiseRMV(ReverseModeVisitor& RMV) override;
```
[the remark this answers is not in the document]
parth-07: > find the common denominator of functionality for all visitors and use this as a common way to listen to what they do?

That's something we should definitely aim for in the future. I will add a FIXME for this.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -19,15 +20,18 @@
...
   /// Used to compute derivatives by clad::gradient.
   class ReverseModeVisitor
       : public clang::ConstStmtVisitor<ReverseModeVisitor, StmtDiff>,
         public VisitorBase {
+  
   private:
-    /// Determines if an error estimation is in process; helps decide whether
-    /// to visit error estimation specific code in calls to VisitStmt.
-    bool m_ErrorEstimationEnabled = false;
+    friend class ErrorEstimationHandler;
```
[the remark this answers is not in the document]
parth-07: Because `ErrorEstimationHandler` needs to access several private members of `ReverseModeVisitor`. We can avoid making it a friend, but then we will need to add several getters/setters member functions in the `ReverseModeVisitor`.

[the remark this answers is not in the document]
vgvassilev: If these are AST-building functions we can move them into the CladUtils files.

include/clad/Differentiator/ReverseModeVisitorDirectionKinds.h
```diff
@@ -0,0 +1,11 @@
...
+#define CLAD_REVERSE_MODE_VISITOR_DIRECTION_KINDS
+
+namespace clad {
+  namespace rmv {
+    /// An enum to operate between forward and reverse passes.
+    enum direction : int { forward, reverse };
+  } // namespace rmv
+} // namespace clad
+
+#endif
```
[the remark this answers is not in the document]
parth-07: We require the enum in both the two header files: `ReverseModeVisitor.h` and `ErrorEstimator.h`. And I think, we cannot forward declare enums. 
Now, since, we do not want `ErrorEstimator.h` to include `ReverseModeVisitor.h`. We need a separate file for the enum. But yes, I think we can perhaps improve this design.

include/clad/Differentiator/VisitorBase.h
```diff
@@ -68,6 +68,7 @@ namespace clad {
   /// A base class for all common functionality for visitors
   class VisitorBase {
   protected:
+    friend class ErrorEstimationHandler;
```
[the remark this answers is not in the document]
parth-07: Private/protected functions such as 'GetCladTapePop`, `GetCladNamespace` etc. Private member variables such as `m_Context`, `m_Sema` etc.

Currently, we are only using the plugin system internally, thus we can always forward declare the plugin class in `ReverseModeVisitor` and add a friend class declaration. This will not add any additional dependency between the plugin class and the `ReverseModeVisitor`.

lib/Differentiator/CladUtils.cpp
```diff
@@ -42,5 +43,25 @@ namespace clad {
       auto stmtsRef = llvm::makeArrayRef(block.begin(), block.end());
       return clad_compat::CompoundStmt_Create(C, stmtsRef, noLoc, noLoc);
     }
+
+    StringLiteral* CreateStringLiteral(ASTContext& C, llvm::StringRef str) {
```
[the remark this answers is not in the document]
parth-07: codecov seems to be innocent here. At least at the state of this PR, this function is indeed not being used anywhere. But there are places where it can and should be used, especially in `DiffPlanner`. I will use this API there.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: Okay. I will add an if condition surrounding the calls to `m_ExternalSoure` functions.

include/clad/Differentiator/ErrorEstimator.h
```diff
@@ -215,8 +226,36 @@ namespace clad {
...
+        const clang::CallExpr*& CE, clang::Expr*& fnDecl,
+        llvm::SmallVector<clang::Expr*, 16>& CallArgs,
+        llvm::SmallVector<clang::VarDecl*, 16>& ArgResultDecls) override;
+    void ActAfterCloningLHSOfAssignOp(
+        clang::Expr*& LCloned, clang::Expr*& R,
+        clang::BinaryOperator::Opcode& opCode) override;
+    void ActBeforeFinalisingAssignOp(clang::Expr*&, clang::Expr*&) override;
+    void ActBeforeFinalizingDifferentiateSingleStmt(const direction& d) override;
+    void ActBeforeFinalizingDifferentiateSingleExpr(const direction& d) override;
+    void ActAfterDifferentiatingVDInVisitDeclStmt(VarDeclDiff& VD) override;
```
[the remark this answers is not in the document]
grimmmyshini: Yes, I think so, everything that had to be coupled was already coupled in functions before so I think this is good to go.

lib/Differentiator/ErrorEstimator.cpp
grimmmyshini: Yes sure, how do I add more tests? In this PR or a new one?

vgvassilev: Either way should be fine.

grimmmyshini: Okay, I don't think I can directly commit to this PR but I'll make a new PR once this one gets merged and put in some missing tests. It seems like @sudo-panda's recent changes cause some of the coverage loss.

include/clad/Differentiator/ReverseModeVisitor.h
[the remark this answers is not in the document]
parth-07: Yes, many of them are AST-building functions, but some are not. Shall I add a FIXME to remove the friend-dependency of plugins from the visitor classes? Or should we resolve this in this PR only?

lib/Differentiator/ErrorEstimator.cpp
parth-07: Okay :).

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -157,16 +156,18 @@ namespace clad {
       JacobianModeVisitor J(*this);
       result = J.Derive(FD, request);
     } else if (request.Mode == DiffMode::error_estimation) {
+      ReverseModeVisitor R(*this);
       // Set the handler.
-      errorEstHandler.reset(new ErrorEstimationHandler(*this));
+      errorEstHandler.reset(new ErrorEstimationHandler());
```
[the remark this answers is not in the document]
parth-07:
```cpp
errorEstHandler.reset(new ErrorEstimationHandler());
```
Please do correct me if I am missing something. Here `errorEstHandler` is of type `ErrorEstimationHandler` that inherits from `ExternalRMVSource`. Thus, this operation is on the error estimation side. 

> The registration of a new source should be out of the RMV.

Currently, we are attaching a new source in the following way:

```cpp
R.AddExternalSource(*errorEstHandler);
```

Here, `R` is an object of type `ReverseModeVisitor`. We need to add sources each time we create `ReverseModeVisitor` object.
Should we change the interface to attach sources?

lib/Differentiator/DerivativeBuilder.cpp
parth-07: No, we need to explicitly attach the sources to this particular instance of `ReverseModeVisitor`. It is also necessary because we do not want to attach the same sources each time we create a `ReverseModeVisitor` class.

lib/Differentiator/DerivativeBuilder.cpp
parth-07: If we want to remove this operation, then we will need to merge the `if` branch for `DiffMode::error_estimation` inside the `if` branch for `DiffMode::reverse`. In this case the `if` branch will look as follows: 

```cpp
else if (request.Mode == DiffMode::reverse || request.Mode == DiffMode::error_estimation) {
  ReverseModeVisitor V(*this);
  
  if (request.Mode == DiffMode::error_estimation) {
   errorEstHandler.reset(new ErrorEstimationHandler());
   // Set error estimation model. If no custom model provided by user,
   // use the built in Taylor approximation model.
   if (!m_EstModel) {
     m_EstModel.reset(new TaylorApprox(*this));
   }
   errorEstHandler->SetErrorEstimationModel(m_EstModel.get());
   R.AddExternalSource(*errorEstHandler);
  }

  if (request.Mode == DiffMode::error_estimation) {
    m_EstModel->clearEstimationVariables();
  }
}
```

lib/Differentiator/DerivativeBuilder.cpp
parth-07: `errorEstHandler` is currently already of a type that inherits from `ExternalRMVSource`. 
Should I make `errorEstHandler` a member variable of `DerivativeBuilder`?

include/clad/Differentiator/ExternalRMVSource.h
```diff
@@ -0,0 +1,166 @@
...
+
+namespace clad {
+
+  class ReverseModeVisitor;
+  class DiffRequest;
+  class StmtDiff;
+  class VarDeclDiff;
+
+  using direction = rmv::direction;
+  // Should we include `DerivativeBuilder.h` instead? `DiffParams` is originally
```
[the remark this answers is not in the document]
parth-07: We have already separated out `DiffParams` into `ParsedDiffArgsTypes.h` file in a recent PR. Thus, we can now include that directly :)

include/clad/Differentiator/ErrorEstimator.h
```diff
@@ -215,8 +226,36 @@ namespace clad {
...
+    void ActBeforeFinalisingVisitBranchSingleStmtInIfVisitStmt() override;
+    void ActBeforeDifferentiatingLoopInitStmt() override;
+    void ActBeforeDifferentiatingSingleStmtLoopBody() override;
+    void ActAfterProcessingSingleStmtBodyInVisitForLoop() override;
+    void ActBeforeFinalisingVisitReturnStmt(StmtDiff& ExprDiff, clang::Expr*& retDeclRefExpr) override;
+    void ActBeforeFinalisingPostIncDecOp(StmtDiff& diff) override;
+    void ActBeforeFinalisingVisitCallExpr(
+        const clang::CallExpr*& CE, clang::Expr*& fnDecl,
+        llvm::SmallVector<clang::Expr*, 16>& CallArgs,
+        llvm::SmallVector<clang::VarDecl*, 16>& ArgResultDecls) override;
```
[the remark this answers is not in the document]
parth-07: I just realized we can also use, `llvm::ArrayRef` here. What should we prefer between `llvm::SmallVectorImpl` and `llvm::ArrayRef` in these situations?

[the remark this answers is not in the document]
vgvassilev: ArrayRef is shorter, not sure how the call sites would look like.

include/clad/Differentiator/ReverseModeVisitor.h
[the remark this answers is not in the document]
vgvassilev: Yes, FIXME works.

parth-07:
> One very easy to implement and quite different tool to implement with the proposed design is a trace-based debugging.

This is a good idea. Maybe we should add native support for such a plugin as well. It can very helpful during debugging. 

And yes, I have rebased the PR, sorry for the delay.

lib/Differentiator/CladUtils.cpp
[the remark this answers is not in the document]
parth-07: I have used this API in `DiffPlanner.cpp`.

lib/Differentiator/DerivativeBuilder.cpp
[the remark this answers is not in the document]
vgvassilev: > ```c++
> errorEstHandler.reset(new ErrorEstimationHandler());
> ```
> 
> Please do correct me if I am missing something. Here `errorEstHandler` is of type `ErrorEstimationHandler` that inherits from `ExternalRMVSource`. Thus, this operation is on the error estimation side.

I think here the type of `errorEstHandler` should be the generic `ExternalRMVSource` which is conditionally created from the outside if there are external "sources".

> 
> > The registration of a new source should be out of the RMV.
> 
> Currently, we are attaching a new source in the following way:
> 
> ```c++
> R.AddExternalSource(*errorEstHandler);
> ```
> 
> Here, `R` is an object of type `ReverseModeVisitor`. We need to add sources each time we create `ReverseModeVisitor` object. Should we change the interface to attach sources?

This should go on the error estimation side, at its initialization time.

lib/Differentiator/DerivativeBuilder.cpp
vgvassilev: See https://github.com/vgvassilev/clad/pull/304#discussion_r803068517

lib/Differentiator/DerivativeBuilder.cpp
vgvassilev: We can kill `DiffMode::error_estimation` as we do not need special case work here. The plugin system will create a listener in the form of the external source and call the error estimation whenever an interesting event happens.

We should also drop `m_ErrorEstimationEnabled` as well.

I think the overall idea of this approach is the RMV and other AD facilities to have no knowledge of error estimation.

parth-07: But we need some kind of trigger to know when we should attach error estimation plugin source. Currently `DiffMode::error_estimation` is working as this trigger.

Error Estimation is definitely a kind of special case here. For most plugins, we can (should) have a command-line interface, and then we can use that as the trigger. For example, if a particular flag has been passed then we should enable a particular plugin.

include/clad/Differentiator/ErrorEstimator.h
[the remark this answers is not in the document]
parth-07: `llvm::SmallVector` is implicitly convertible to `llvm::ArrayRef`. Therefore call sites should not change.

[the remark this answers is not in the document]
vgvassilev: Then maybe we can use `ArrayRef`.

include/clad/Differentiator/VisitorBase.h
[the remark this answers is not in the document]
vgvassilev: Maybe add a FIXME here as well?

include/clad/Differentiator/ErrorEstimator.h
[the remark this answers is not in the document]
parth-07: As it turned out, we cannot `ArrayRef` because we need to pass `llvm::SmallVector` by reference. I have modified the function parameters to `llvm::SmallVectorImpl`.

lib/Differentiator/DerivativeBuilder.cpp
vgvassilev: Yes, please.

include/clad/Differentiator/ExternalRMVSource.h
```diff
@@ -0,0 +1,160 @@
...
+  virtual void ActBeforeFinalizingDifferentiateSingleStmt(const direction& d) {}
+
+  /// This is called just before finalising
+  /// `ReverseModeVisitor::DifferentiateSingleExpr`.
+  virtual void ActBeforeFinalizingDifferentiateSingleExpr(const direction& d) {}
+
+  virtual void ActBeforeFinalizingVisitDeclStmt(
+      llvm::SmallVectorImpl<clang::Decl*>& decls,
+      llvm::SmallVectorImpl<clang::Decl*>& declsDiff) {}
+};
```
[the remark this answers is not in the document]
vgvassilev: Ok, let's then ignore these complains for now.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -209,6 +222,8 @@ namespace clad {
     if (args.empty())
       return {};
 
+    if (!m_ExternalSource->empty())
```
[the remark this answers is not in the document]
parth-07: We can have `m_ExternalSource` be `nullptr` if there are no external sources attached. 
But then, we will need to modify `ReverseModeVisitor::AddExternalSource` function to always check if `m_ExternalSource` is available or not. I have made necessary changes for this.

lib/Differentiator/DerivativeBuilder.cpp
parth-07: I have made `errorEstHandler` a member variable of `DerivativeBuilder`.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1901,7 +1898,7 @@ namespace clad {
     for (auto D : DS->decls()) {
       if (auto VD = dyn_cast<VarDecl>(D)) {
         VarDeclDiff VDDiff = DifferentiateVarDecl(VD);
-
+
```
vgvassilev, line 1901: Stray space. Please rerun clang-format.

parth-07: done.

lib/Differentiator/MultiplexExternalRMVSource.cpp
```diff
@@ -0,0 +1,195 @@
...
+  }
+}
+
+void MultiplexExternalRMVSource::
+    ActAfterProcessingSingleStmtBodyInVisitForLoop() {
+  for (auto source : m_Sources) {
+    source->ActAfterProcessingSingleStmtBodyInVisitForLoop();
+  }
+}
+void MultiplexExternalRMVSource::ActBeforeFinalisingVisitReturnStmt(
```
vgvassilev, line 137: we should choose space or no space between methods.

parth-07: We are generally following space between methods. I am adding spaces here.

parth-07: I have added space here.

vgvassilev (approved):
LGTM! The complaints from codecov are understood and we can ignore them for now.

PR #305, opened 2021-09-15
Fix large clad::tape for trivial types

sudo-panda:
MoveData function in clad::tape was missing a specialization for trivial types.


vgvassilev:
I think that problem is addressed by #306. Let's reopen if we see the issue again.

PR #311, opened 2021-10-14
Make clad compatible with clang 14.0.x

parth-07:
This PR makes clad compatible with the clang 14.0.x version and adds GitHub runners for testing Clad with clang-13 and clang-14 runtimes.

### DIFF
--- a/test/lit.cfg
+++ b/test/lit.cfg
@@ -252,6 +252,8 @@ if not lit_config.quiet:
 if not lit_config.quiet:
     lit_config.note('using auto diff lib: %r' % config.cladlib)
 
+lit.util.usePlatformSdkOnDarwin(config, lit_config)
+
 #Start clad forwarding to:
 #clang -x c++ -std=c++11 some-input-file.c -Xclang -add-plugin -Xclang ad -Xclang
 #\ -plugin-arg-ad -Xclang -fdump-derived-fn -Xclang -load -Xclang../../Debug+Asserts/lib/libclad.so


vgvassilev:
@parth-07, I think we have this mostly but can you rebase and check?

parth-07:
> @parth-07, I think we have this mostly but can you rebase and check?

PR is ready for review. I have added a Ubuntu 20.04 GitHub runner with Clang-14 both as compiler and runtime. Tests are passing on it. But it seems that code coverage is only taking contributions from `ubu18-gcc6-runtime11-analyzers` GitHub runner.

CMakeLists.txt
```diff
@@ -21,9 +21,9 @@ if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
   ## Define clad supported version of clang and llvm
 
   set(CLANG_MIN_SUPPORTED 5.0)
-  set(CLANG_MAX_SUPPORTED 13.0.0-rc1)
+  set(CLANG_MAX_SUPPORTED 14.0.4)
```
vgvassilev, line 24: Can we make this open ended - I doubt patch releases will introduce breaking changes.

vgvassilev: @parth-07 ping

parth-07: I have modified CMake file such that all `14.0.x` LLVM versions would be supported.

.github/workflows/ci.yml
```diff
@@ -393,6 +393,16 @@ jobs:
...
+          
+          - name: ubu20-clang13-runtime13
+            os: ubuntu-20.04
+            compiler: clang-13
+            clang-runtime: '13'
+
+          - name: ubu20-clang14-runtime14
+            os: ubuntu-20.04
+            compiler: clang-14
+            clang-runtime: '14'
```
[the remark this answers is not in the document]
vgvassilev: You can drop onward, we dropped most of runtime-5 now we should drop most of runtime-6. I don't think we can fix the coverage issue as you added code which is conditionally executed/tested. I think it is fine to ignore codecov for this one.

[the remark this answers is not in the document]
parth-07: We only have runtimes 5, 6, and 7 in Mac systems. Since we have added two new runtimes 13, and 14, I have removed most of the runners with runtime-8, and some of the runners with runtime-9. Please let me know if this requires any changes.

> I think it is fine to ignore codecov for this one.

Oh, okay.

PR #323, opened 2021-10-29
Add information about numerical differentiation

grimmmyshini:
Adds information about clad's numerical differentiation abilities.


vgvassilev (approved):

grimmmyshini:
Squahed!

PR #333, opened 2021-11-03
Fix multiple index pop issue.

### LINKED ISSUE #246: Multiple pops to array index causes failed assertions in reverse mode
Consider the following (dummy) example:

```cpp 
float func1(float x){
  float arr[10];
    for (int i = 0; i < 9; i++){
      arr[i] += x;
    }
  return arr[9];
}
```

The generated derivative of this in reverse mode is as follows: 

```cpp
void func1_grad(float x, float *_result) {
    float _d_arr[10] = {};
    unsigned long _t0;
    int _d_i = 0;
    clad::tape<int> _t1 = {};
    float arr[10];
    _t0 = 0;
    for (int i = 0; i < 9; i++) {
        _t0++;
        arr[clad::push(_t1, i)] += x;     <--  only one push of index per iteration
    }
    float func1_return = arr[9];
    goto _label0;
  _label0:
    _d_arr[9] += 1;
    for (; _t0; _t0--) {
        {
            float _r_d0 = _d_arr[clad::pop(_t1)];    <-- pop one
            _d_arr[clad::pop(_t1)] += _r_d0;          <-- pop two
            _result[0UL] += _r_d0;
            _d_arr[clad::pop(_t1)] -= _r_d0;           <-- pop three
        }
    }
}

```

Here, popping the index more than the number of times it was pushed causes a ```clad::tape_impl::reference clad::tape_impl<int>::back() [T = int]: Assertion `_size' failed.```

**Proposed solution(s):**

1. Replace ```clad::pop``` in the reverse block with ```clad::back```. The code will then look like so:

```cpp
{
       float _r_d0 = _d_arr[clad::back(_t1)];     <-- use back in expressions 
       _d_arr[clad::back(_t1)] += _r_d0;          
       _result[0UL] += _r_d0;
       _d_arr[clad::back(_t1)] -= _r_d0;           
       clad::pop(_t1);                                        <--finally pop here after use
}
```
This is a bit "harder" to develop in the sense that we would have to backpropagate 3 calls -- push, pop, and back. This would require either changing the structure of the ```VisitArraySubscriptExpr``` by some amount or developing a hacky fix to send the values back.

2. build a temporary replacement for the popped value

```cpp
{
       int _t0 = clad::pop(_t1);      <-- store pop value in expression
       float _r_d0 = _d_arr[_t0];     <-- use temp value here
       _d_arr[_t0]] += _r_d0;          
       _result[0UL] += _r_d0;
       _d_arr[_t0]] -= _r_d0;                                        
}
```
Easier to implement, no major change in the structure of ```VisitArraySubscriptExpr```. However, it introduces a new variable (although it is local) to the compound statement scope. 

Any ideas on which one would be better? Or any other better suggestions?

Once a proposed solution is decided, I would like to put in a PR for this!

grimmmyshini:
Possibly closes #246.

I am not 100% happy with the way the solution is implemented. However, I really cannot think of a better way to do this. Any and all suggestions are welcome!

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -45,6 +45,10 @@ namespace clad {
     bool isInsideLoop = false;
     /// Output variable of vector-valued function
     std::string outputArrayStr;
+    /// Stores the pop index values for arrays in reverse mode.This is required
+    /// to maintain the correct statement order when the current block has
+    /// delayed emission i.e. assignment LHS.
+    Stmts m_PopIdxValues;
     unsigned outputArrayCursor = 0;
     unsigned numParams = 0;
     bool isVectorValued = false;
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -985,6 +985,19 @@ namespace clad {
     for (std::size_t i = 0; i < Indices.size(); i++) {
       StmtDiff IdxDiff = Visit(Indices[i]);
       StmtDiff IdxStored = GlobalStoreAndRef(IdxDiff.getExpr());
+      if (isInsideLoop) {
+        // Here we make sure that we are popping each time we push.
+        // Since the max no of pushes = no. of array index expressions in the
+        // loop.
+        Expr* popExpr = IdxStored.getExpr_dx();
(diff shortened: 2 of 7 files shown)


vgvassilev (approved):
LGTM!

vgvassilev:
@grimmmyshini, could you squash and improve the commit message?

PR #335, opened 2021-11-19
[cmake] Try to fix the subsequent benchmarks.

vgvassilev:
We now have added another target benchmark-clad which will rebuild what's necessary and run the benchmarks. In addition we now track more rigorously when a branch changes and re-configure clad so that we can have two different json files which can be compared against.


vgvassilev:
The osx failure is due to a glitch and the benchmark bot should fail as we updated the ci.yml and there is no equivalent of the new logic in the master (before the PR).

PR #339, opened 2021-12-04
Add support for differentiating switch stmt in the reverse mode AD.

parth-07:
This PR adds support for differentiating switch statements in the reverse mode AD.

The basic idea used to differentiate switch statement is that in the forward pass, processing of the statements of the switch statement body always starts from a `case`/`default` label and ends at a `break` statement or at the end of the switch body. 
Similarly, in the reverse pass, processing of the differentiated statements of the switch statement body will start from the statement just above the `break` statement that was hit or from the last differentiated statement in the case when no `break` statement was hit.

Thus, we can keep track of which `break` statement was hit in the forward pass or if no `break` statement was hit at all in a variable. This information is further used by an auxiliary switch statement in the reverse pass to jump the execution to the correct point (that is, differentiated statement of the statement just before the `break` statement that was hit in the forward pass).

In this strategy, each switch case statement of the original function gets transformed to an `if` condition in the reverse pass. The `if` condition decides at runtime whether the processing of the differentiated statements of the switch statement body should stop or continue. This is again based on the fact that the processing of statements of the switch statement body always starts at a case statement.

For an example, consider this code snippet:

```c++
switch (count) {
    case 0: a += i; break;
    case 2: a += 4 * i; break;
    default: a += 10 * i;
  }
```
`case 0` of this code snippet gets transformed to the following in the differentiated function:

forward pass:
```c++
{
  case 0: a += i;
}
{
  clad::push(_t0, 1UL); // this is used to keep track if this break was hit; 1UL is used to represent the case number
  break;
}
```

reverse pass:

```c++
case 1UL:;  // this case is selected if the corresponding break was hit in the forward pass
  {
    {
      double _r_d0 = _d_a;
      _d_a += _r_d0;
      *_d_i += _r_d0;
      _d_a -= _r_d0;
    }
    if (0 == _cond0)  // If case 0: was selected in the forward pass, we should break out of processing differentiated switch stmt body here.
      break;
  }
 ```

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -374,6 +374,9 @@ namespace clad {
     StmtDiff
     VisitMaterializeTemporaryExpr(const clang::MaterializeTemporaryExpr* MTE);
     StmtDiff VisitCXXStaticCastExpr(const clang::CXXStaticCastExpr* SCE);
+    StmtDiff VisitSwitchStmt(const clang::SwitchStmt* SS);
+    StmtDiff VisitCaseStmt(const clang::CaseStmt* CS);
+    StmtDiff VisitDefaultStmt(const clang::DefaultStmt* DS);
     VarDeclDiff DifferentiateVarDecl(const clang::VarDecl* VD);
     StmtDiff VisitSubstNonTypeTemplateParmExpr(
         const clang::SubstNonTypeTemplateParmExpr* NTTP);
@@ -485,7 +488,7 @@ namespace clad {
                                    clang::Stmt* forLoopIncDiff = nullptr,
                                    bool isForLoop = false);
 
-    /// This class modifies forward and reverse blocks of the loop
+    /// This class modifies forward and reverse blocks of the loop/switch
     /// body so that `break` and `continue` statements are correctly
     /// handled. `break` and `continue` statements are handled by
     /// enclosing entire reverse block loop body in a switch statement
@@ -528,6 +531,7 @@ namespace clad {
 
       ReverseModeVisitor& m_RMV;
 
+      const bool m_IsInvokedBySwitchStmt = false;
       /// Builds and returns a literal expression of type `std::size_t` with
       /// `value` as value.
       clang::Expr* CreateSizeTLiteralExpr(std::size_t value);
@@ -542,7 +546,8 @@ namespace clad {
       clang::Expr* CreateCFTapePushExpr(std::size_t value);
 
     public:
-      BreakContStmtHandler(ReverseModeVisitor& RMV) : m_RMV(RMV) {}
+      BreakContStmtHandler(ReverseModeVisitor& RMV, bool forSwitchStmt = false)
+          : m_RMV(RMV), m_IsInvokedBySwitchStmt(forSwitchStmt) {}
 
       /// Begins control flow switch statement scope.
       /// Control flow switch statement is used to refer to the
@@ -574,8 +579,8 @@ namespace clad {
     BreakContStmtHandler* GetActiveBreakContStmtHandler() {
       return &m_BreakContStmtHandlers.back();
     }
-    BreakContStmtHandler* PushBreakContStmtHandler() {
-      m_BreakContStmtHandlers.emplace_back(*this);
+    BreakContStmtHandler* PushBreakContStmtHandler(bool forSwitchStmt = false) {
+      m_BreakContStmtHandlers.emplace_back(*this, forSwitchStmt);
       return &m_BreakContStmtHandlers.back();
     }
     void PopBreakContStmtHandler() {
@@ -609,6 +614,26 @@ namespace clad {
 
     clang::QualType ComputeAdjointType(clang::QualType T);
     clang::QualType ComputeParamType(clang::QualType T);
+    /// Stores data required for differentiating a switch statement.
+    struct SwitchStmtInfo {
+      llvm::SmallVector<clang::SwitchCase*, 16> cases;
+      clang::Expr* switchStmtCond = nullptr;
+      clang::IfStmt* defaultIfBreakExpr = nullptr;
+    };
+
+    /// Maintains a stack of `SwitchStmtInfo`.
+    llvm::SmallVector<SwitchStmtInfo, 4> m_SwitchStmtsData;
+
+    SwitchStmtInfo* GetActiveSwitchStmtInfo() {
+      return &m_SwitchStmtsData.back();
+    }
+
+    SwitchStmtInfo* PushSwitchStmtInfo() {
+      m_SwitchStmtsData.emplace_back();
+      return &m_SwitchStmtsData.back();
+    }
+
+    void PopSwitchStmtInfo() { m_SwitchStmtsData.pop_back(); }
   };
 } // end namespace clad
 
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -632,5 +632,12 @@ namespace clad {
       finder.TraverseStmt(const_cast<Stmt*>(S));
       return finder.hasCallExpr;
     }
+
+    void SetSwitchCaseSubStmt(SwitchCase* SC, Stmt* subStmt) {
+      if (auto caseStmt = dyn_cast<CaseStmt>(SC))
+        caseStmt->setSubStmt(subStmt);
+      else
+        cast<DefaultStmt>(SC)->setSubStmt(subStmt);
+    }
   } // namespace utils
 } // namespace clad
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -634,7 +634,7 @@ namespace clad {
     }
 
     void SetSwitchCaseSubStmt(SwitchCase* SC, Stmt* subStmt) {
-      if (auto *caseStmt = dyn_cast<CaseStmt>(SC))
+      if (auto* caseStmt = dyn_cast<CaseStmt>(SC))
         caseStmt->setSubStmt(subStmt);
       else
         cast<DefaultStmt>(SC)->setSubStmt(subStmt);
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -634,7 +634,7 @@ namespace clad {
     }
 
     void SetSwitchCaseSubStmt(SwitchCase* SC, Stmt* subStmt) {
-      if (auto caseStmt = dyn_cast<CaseStmt>(SC))
+      if (auto *caseStmt = dyn_cast<CaseStmt>(SC))
         caseStmt->setSubStmt(subStmt);
       else
         cast<DefaultStmt>(SC)->setSubStmt(subStmt);
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -9,6 +9,7 @@
 #include "ConstantFolder.h"
 
 #include "TBRAnalyzer.h"
+#include "clad/Differentiator/DerivativeBuilder.h"
 #include "clad/Differentiator/DiffPlanner.h"
 #include "clad/Differentiator/ErrorEstimator.h"
 #include "clad/Differentiator/ExternalRMVSource.h"
@@ -17,7 +18,9 @@
 
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/Expr.h"
+#include "clang/AST/Stmt.h"
 #include "clang/AST/TemplateBase.h"
+#include "clang/Basic/TokenKinds.h"
 #include "clang/Sema/Lookup.h"
 #include "clang/Sema/Overload.h"
 #include "clang/Sema/Scope.h"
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: include/clad/Differentiator/ReverseModeVisitor.h [hunk 1]
```diff
@@ -504,13 +509,37 @@ namespace clad {
...
+    BreakContStmtHandler* PushBreakContStmtHandler(bool forSwitchStmt=false) {
+      m_BreakContStmtHandlers.emplace_back(*this, forSwitchStmt);
       return &m_BreakContStmtHandlers.back();
     }
     void PopBreakContStmtHandler() {
       m_BreakContStmtHandlers.pop_back();
     }
+
+    /// Stores data required for differentiating a switch statement.
+    class SwitchStmtData {
```
vgvassilev, line 521: We usually add `Info`, eg the class name becomes `SwitchStmtInfo`

lib/Differentiator/CladUtils.cpp
```diff
@@ -42,5 +42,15 @@ namespace clad {
...
     }
+
+    void SetSwitchCaseSubStmt(SwitchCase* SC, Stmt* subStmt) {
+      if (auto caseStmt = dyn_cast<CaseStmt>(SC)) {
+        caseStmt->setSubStmt(subStmt);
+      } else if (auto defaultStmt = dyn_cast<DefaultStmt>(SC)) {
+        defaultStmt->setSubStmt(subStmt);
+      } else {
+        assert(0 && "Unsupported switch case statement");
+      }
```
vgvassilev, line 53:
```suggestion
      if (auto caseStmt = dyn_cast<CaseStmt>(SC))
        caseStmt->setSubStmt(subStmt);
      else
        cast<DefaultStmt>(SC)->setSubStmt(subStmt);
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -77,7 +77,7 @@ namespace clad {
     Expr* PushExpr =
         m_Sema.ActOnCallExpr(getCurrentScope(), PushDRE, noLoc, CallArgs, noLoc)
             .get();
-    return CladTapeResult{*this, PushExpr, PopExpr, TapeRef};
+    return CladTapeResult{this, PushExpr, PopExpr, TapeRef};
```
vgvassilev, line 80: Can we avoid passing `this` by pointer?

vgvassilev (commented):
This is a partial review. We will need some benchmarks which IIRC are on my plate, at least for the tapenade infrastructure.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -231,7 +231,7 @@ namespace clad {
                                                 llvm::StringRef prefix = "_t");
 
     struct CladTapeResult {
-      ReverseModeVisitor& V;
+      ReverseModeVisitor* const  V;
```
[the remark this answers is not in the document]
parth-07: This will change the intended meaning. 
`ReverseModeVisitor* const V` -> pointer is constant, `ReverseModeVisitor` is non-const
`const ReverseModeVisitor* V` -> pointer is non-const, `ReverseModeVisitor` is const.

include/clad/Differentiator/ReverseModeVisitor.h
parth-07: Okay, I will make this change.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: I will look into it.

parth-07: Yes, we could avoid passing `this` by pointer here, we now pass `*this` by reference.

parth-07:
@vgvassilev Should I add clad benchmarking for switch statement with this PR?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3237,6 +3237,214 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    // Handles condition variable
+    if (SS->getConditionVariable()) {
+      StmtDiff condVarDiff = DifferentiateSingleStmt(SS->getConditionVariableDeclStmt());
+      addToCurrentBlock(condVarDiff.getStmt(), direction::forward);
+      addToCurrentBlock(condVarDiff.getStmt_dx(), direction::reverse);
+    }
+    // Condition is only cloned, and not differentiated.
+    // Its because conditions generally contain non-differentiable constructs, but
+    // this behaviour will lead to incorrect results if the condition expression
+    // modifies any variable.
```
PetroZarytskyi, line 3276: What non-differentiable constructs are you talking about here? Can you give an example? I think we differentiate conditions in loops, are switch stmts somehow different?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3237,6 +3237,214 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+      // switch (clad::push(..., cond)) { ... }
+      //
+      // reverse block:
+      // switch (...) { ... }
+      // clad::pop(...);
+      condTape.emplace(MakeCladTapeFor(condClone, "_cond"));
+      condExpr = condTape->Push;
+    } else {
+      condExpr = GlobalStoreAndRef(condClone, "_cond").getExpr();
+    }
```
PetroZarytskyi, line 3295: Do we have the same inconsistency here as described in #431?

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: > I think we differentiate conditions in loops, are switch stmts somehow different?
No, switch statement conditions are not different. I am not quite sure why I previously did not differentiate the switch statement condition. Thank you for catching this. I have fixed this now.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: Here `GlobalStoreAndRef` is used when the switch statement is _outside_ the loop. So, it does not suffer from the inconsistency issue.

PetroZarytskyi: I was wondering whether fixing #431 would allow us to use GlobalStoreAndRef inside loops as well.

parth-07: Yes, it will. It's one of the main motivations for the issue.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -607,6 +612,27 @@ namespace clad {
 
     clang::QualType ComputeAdjointType(clang::QualType T);
     clang::QualType ComputeParamType(clang::QualType T);
+    /// Stores data required for differentiating a switch statement.
+    class SwitchStmtInfo {
+    public:
```
vgvassilev, line 617: Maybe we should use a struct?

test/Gradient/Switch.C
```diff
@@ -0,0 +1,596 @@
...
+// CHECK-NEXT:                 * _d_j += i * i * j * _r_d0;
+// CHECK-NEXT:             }
+// CHECK-NEXT:         }
+// CHECK-NEXT:     }
+// CHECK-NEXT: }
+
+double fn3(double i, double j) {
+  double res = 0;
+  int counter = 2;
+  while (counter--) {
```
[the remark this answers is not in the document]
parth-07: I have added a test with switch statement under a for-loop.

include/clad/Differentiator/ReverseModeVisitor.h
parth-07: Yes, we can use struct. I have changed the code to use struct.

parth-07:
> Most of the clang-tidy remarks look reasonable. Can you address them?

I have addressed all the clang-tidy remarks except this one:

```
warning: initializing non-owner 'clang::DefaultStmt *' with a newly created 'gsl::owner<>' [cppcoreguidelines-owning-memory]
```

I am not sure how to address this.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3237,6 +3240,211 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    utils::SetSwitchCaseSubStmt(newSC, diff.getStmt());
+    addToCurrentBlock(diff.getStmt_dx(), direction::reverse);
+    return {endBlock(direction::forward), endBlock(direction::reverse)};
+  }
+
+  StmtDiff ReverseModeVisitor::VisitDefaultStmt(const DefaultStmt* DS) {
+    beginBlock(direction::reverse);
+    beginBlock(direction::forward);
+    auto* SSData = GetActiveSwitchStmtInfo();
+    auto* newDefaultStmt =
```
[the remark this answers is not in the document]
vgvassilev: Could we use Sema::ActOnDefaultStmt? Now we should have better handling of scopes in Sema and maybe would work out of the box?

vgvassilev:
> @vgvassilev Should I add clad benchmarking for switch statement with this PR?

That would be a good idea assuming we can figure out how to make it scale for given N of switch statements. I suspect a benchmark would be used to check against potentially a future alternative approach which uses less tape...

> I have addressed all the clang-tidy remarks except this one

I don't seem to find the changes for `auto Type =...` to `auto * Type =...`. Did you push?

lib/Differentiator/ReverseModeVisitor.cpp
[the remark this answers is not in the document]
parth-07: We cannot (easily) use `Sema::ActOnDefaultStmt` here because `Sema::ActOnDefaultStmt` and `Sema::ActOnCaseStmt` access the _current_ `switch` statement to register the `case`/`default` with the `switch` statement. The issue with this is that we are creating two `switch` statements simultaneously -- one for the forward pass and another for the reverse pass, so there is no _one current_ switch statement. Currently, we manually register `case` and `default` with their corresponding switch statements.

If we use `Sema::ActOnDefaultStmt` but not `Sema::ActOnCaseStmt`, then the order of `case` and `default` in the `switch` statement will be disturbed.

If we use both `Sema::ActOnDefaultStmt` and `Sema::ActOnCaseStmt` then we  will have to use these functions for the forward `switch` statement, but still stick to the old ways for the reverse `switch` statement. This is because `Sema::ActOn(Default|Case)Stmt` functions access the current `switch` statement as `Sema::getCurFunction()->SwitchStack.back()` and we cannot have both the forward and the reverse `switch` statements as top of the stack here. I don't like to have disparity in the implementation for fixing a warning.

parth-07:
> That would be a good idea assuming we can figure out how to make it scale for given N of switch statements. I suspect a benchmark would be used to check against potentially a future alternative approach which uses less tape...

Yes. I would like the benchmark to be added in a separate pull request, if possible.

> I don't seem to find the changes for auto Type =... to auto * Type =.... Did you push?

Yes, I think I did. the latest clang-tidy review also only have one suggestion, and that is due to using the placement new operator (I think).

lib/Differentiator/ReverseModeVisitor.cpp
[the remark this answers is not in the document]
vgvassilev: Ok, understood! Thanks!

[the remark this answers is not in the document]
parth-07: Please let me know what are your thoughts on this. We can definitely use `Sema::ActOnDefaultStmt` for building `default` because we only need to build the `default` case for the forward `switch` statement. But then we will need to use `ActOn[Case|Default]Stmt` for the forward `switch` statement, and `Build[Case|Default]Stmt` for the reverse `switch` statement.

vgvassilev:
> > That would be a good idea assuming we can figure out how to make it scale for given N of switch statements. I suspect a benchmark would be used to check against potentially a future alternative approach which uses less tape...
> 
> Yes. I would like the benchmark to be added in a separate pull request, if possible.
> 
> > I don't seem to find the changes for auto Type =... to auto * Type =.... Did you push?
> 
> Yes, I think I did. the latest clang-tidy review also only have one suggestion, and that is due to using the placement new operator (I think).

Clang-tidy does not re-upload comments that it already has commented. It seems that you have not pushed the changes as the diff does not contain the suggestion. We can generally do it via github by just clicking on commit if you prefer.

parth-07:
> Clang-tidy does not re-upload comments that it already has commented. It seems that you have not pushed the changes as the diff does not contain the suggestion. 

Oh, thank you for the information. I did not know that Clang-tidy does not re-upload comments. I will check the formatting.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3237,6 +3237,214 @@
...
+
+    PopBreakContStmtHandler();
+    PopSwitchStmtInfo();
+    return {endBlock(direction::forward), endBlock(direction::reverse)};
+  }
+
+  StmtDiff ReverseModeVisitor::VisitCaseStmt(const CaseStmt* CS) {
+    beginBlock(direction::forward);
+    beginBlock(direction::reverse);
+    auto SSData = GetActiveSwitchStmtInfo();
```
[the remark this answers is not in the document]
parth-07: Done.

vgvassilev (approved):
Lgtm!

PR #340, opened 2021-12-06
Add no-op cast to function arg in num diff.

grimmmyshini:
Add a no-op c-style cast in the function argument to numerical differentiation so that the printing of the generated code retains the context(type) of the function argument. Earlier the printed code could not be run standalone, however, if we explicitly specify the type of the function as a cast, there is no ambiguity in template instantiations of the numerical diff function.

This solution allows us to cover the case of function templates and overloads without introducing unnecessary complexity.

### DIFF
--- a/lib/Differentiator/VisitorBase.cpp
+++ b/lib/Differentiator/VisitorBase.cpp
@@ -736,7 +736,15 @@ namespace clad {
     DeclarationNameInfo DNInfo(name, noLoc);
     // Build function args.
     llvm::SmallVector<Expr*, 16U> NumDiffArgs;
-    NumDiffArgs.push_back(targetFuncCall);
+    // build a c style cast around target function call to avoid ambiguity 
+    // for overload and function templates when the generated code is printed.
+    QualType targetQT = targetFuncCall->getType();
+    auto castExp = CStyleCastExpr::Create(
+        m_Context, targetQT, targetFuncCall->getValueKind(), CastKind::CK_NoOp,
+        targetFuncCall, 0 /*FPO=*/CLAD_COMPAT_CLANG12_CastExpr_DefaultFPO,
+        m_Context.getTrivialTypeSourceInfo(targetQT),
+        targetFuncCall->getBeginLoc(), targetFuncCall->getEndLoc());
+    NumDiffArgs.push_back(castExp);
     NumDiffArgs.push_back(targetArg);
     NumDiffArgs.push_back(ConstantFolder::synthesizeLiteral(m_Context.IntTy,
                                                             m_Context,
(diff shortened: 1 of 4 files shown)


lib/Differentiator/VisitorBase.cpp
```diff
@@ -761,7 +769,15 @@ namespace clad {
...
-    NumDiffArgs.push_back(targetFuncCall);
+    // build a c style cast around target function call to avoid ambiguity 
+    // for overload and function templates when the generated code is printed.
+    QualType targetQT = targetFuncCall->getType();
+    auto castExp = CStyleCastExpr::Create(
+        m_Context, targetQT, targetFuncCall->getValueKind(), CastKind::CK_NoOp,
+        targetFuncCall, 0 /*FPO=*/CLAD_COMPAT_CLANG12_CastExpr_DefaultFPO,
+        m_Context.getTrivialTypeSourceInfo(targetQT),
+        targetFuncCall->getBeginLoc(), targetFuncCall->getEndLoc());
+    NumDiffArgs.push_back(castExp);
```
vgvassilev, line 780: I am confused, is that the template instantiation that fails to deduce the correct argument type?

grimmmyshini: Technically yes, the num different function fails to deduce the function type in some cases where the function type is ambiguous (I.e. function overloads, function templates etc). Now please keep in mind that this is relating to running the output from code dumps.

vgvassilev: Okay, I understand. I think we are looking for another solution where we fix the actual template instantiation rather than changing the types after the fact.

grimmmyshini: There is nothing wrong with the template instantiation at all. Everything is instantiated correctly by clang. It can be verified by looking at the AST. The issue is that when we  print the code equivalent of the AST (where the template is correctly instantiated), we end up losing a lot of important information about the template instantiation. So, to overcome this, we have to add a redundant cast over the parameters. 

Having said that, we could possibly do things the way you say but it will obviously be much much more complicated and possibly targeting an issue that doesn't need to be fixed. 

I agree it looks ugly but this is the fastest and least complex way of doing this that I could come up with.

vgvassilev: > There is nothing wrong with the template instantiation at all. Everything is instantiated correctly by clang. It can be verified by looking at the AST. The issue is that when we print the code equivalent of the AST (where the template is correctly instantiated), we end up losing a lot of important information about the template instantiation. So, to overcome this, we have to add a redundant cast over the parameters.

That sounds very surprising. We discussed that over the last months extensively. Could you give a standalone example with code that is written by hand where we lose information about the template instantiation?

> Having said that, we could possibly do things the way you say but it will obviously be much much more complicated and possibly targeting an issue that doesn't need to be fixed.

It is a matter of consistency. Could you provide the code that we synthesize so that I can run it in a standalone clang instance and compare what we do wrong?

PR #343, opened 2021-12-31
Fix repeated generation of same derivatives

### LINKED ISSUE #115: Fix repeated generation of same derivatives
If the same derivative is requested several times (e.g. if several calls to `clad::gradient(f)` are found, or if `f` is called from another differentiated function on several places), Clad will differentiate the function each time. We should instead memorize the result and return it if requested.

parth-07:
This PR fixes repeated generation of the same derivatives by memorizing and reusing previously computed derivatives. 

Implementing this resulted in some minor design changes. Summary of changes are as follows: 

- `clad::DiffRequest` aka Differentiation Request now represents complete information regarding a differentiation request, including the parsed differentiation argument information. This change shifts the responsibility of parsing the differentiation arguments to the entity that creates the differentiation request. 

- `clad::DerivedFnInfo` class is created to store and represent information about a derived function.

- `clad::DerivedFnCollector` class is created to store `DerivedFnInfo` objects of computed derivatives. This class provide functions to check if some already computed derivatives satisfy the requirement of the current differentiation request. 

- Behaviour of `-fdump-derived-fn`, `-fdump-derived-fn-ast` and `-fgenerate-source-file` flags have been changed to dump the same derived function only once irrespective of the number of times it is requested to be generated. 

- Singature of `DerivativeBuilder::Derive(const clang::FunctionDecl* FD, const DiffRequest& request)` have been changed to `DerivativeBuilder::Derive(const DiffRequest& request)` because `FD` was always equal to `request.Function`. 

Closes #115


vgvassilev:
Thanks for the PR @parth-07. 

I had a question on a design level: why can't we store `DerivedFnInfo` in the `DiffRequest` and remove the need for `DerivedFnCollector`? Note that, at that level performance is not really a concern if we will have to shuffle through several diff plans.

parth-07:
> Thanks for the PR @parth-07.
> 
> I had a question on a design level: why can't we store `DerivedFnInfo` in the `DiffRequest` and remove the need for `DerivedFnCollector`? Note that, at that level performance is not really a concern if we will have to shuffle through several diff plans.

Please correct me if I am wrong somewhere. 
`DiffRequest` conceptually only contains information about a single differentiation request. It is also short-lived, that is, it is discarded when the differentiation request is processed. , Therefore, we cannot store `DerivedFnInfo` inside a `DiffRequest`. We can instead store `DerivedFnInfo` objects in a static variable of the `DiffCollector` class. But that will break the single responsibility principle. As the `DiffCollector` class purpose is to collect differentiation requests. 

You earlier suggested removing `DerivedFnInfo` and `DerivedFnCollector` classes and instead shifting the responsibility of creating derived functions names from `*Visitor` classes to the `DiffPlanner` file, so that we can lookup derived functions if they exist without venturing into `DerivativeBuilder`. Should I go with this approach?

vgvassilev:
> > Thanks for the PR @parth-07.
> > I had a question on a design level: why can't we store `DerivedFnInfo` in the `DiffRequest` and remove the need for `DerivedFnCollector`? Note that, at that level performance is not really a concern if we will have to shuffle through several diff plans.
> 
> Please correct me if I am wrong somewhere. `DiffRequest` conceptually only contains information about a single differentiation request. It is also short-lived, that is, it is discarded when the differentiation request is processed. , Therefore, we cannot store `DerivedFnInfo` inside a `DiffRequest`. We can instead store `DerivedFnInfo` objects in a static variable of the `DiffCollector` class. But that will break the single responsibility principle. As the `DiffCollector` class purpose is to collect differentiation requests.
> 
> You earlier suggested removing `DerivedFnInfo` and `DerivedFnCollector` classes and instead shifting the responsibility of creating derived functions names from `*Visitor` classes to the `DiffPlanner` file, so that we can lookup derived functions if they exist without venturing into `DerivativeBuilder`. Should I go with this approach?

The lookup is an easy way out. I think we need the information about what was produced in various places anyway. Check `m_GeneratedDerivatives` and `m_Derivatives`. In these places we store the `Decl*`. Instead, we could store more information about the `DiffRequest`, or just some `DeclarationName`-> `Decl*` mapping which we can look up when constructing the diff request...

parth-07:
> The lookup is an easy way out. I think we need the information about what was produced in various places anyway. Check `m_GeneratedDerivatives` and `m_Derivatives`. In these places we store the `Decl*`. Instead, we could store more information about the `DiffRequest`, or just some `DeclarationName`-> `Decl*` mapping which we can look up when constructing the diff request...

Just `DeclarationName` -> `Decl*` mapping will not be enough to uniquely identify a derivative. Because the same function can be differentiated in several different ways, such as different modes, different independent arguments and different requested derivative orders. And we also cannot just use `DeclarationName` due to the same name at different scope issues.

We can store all this required information to uniquely find the derivative that satisfies the current differentiation request. But we also need to maintain 2 maps, one for derivative declaration and the other for overloaded derivative declaration. 

I have tried to solve this problem using `DerivedFnInfo` class and `DerivedFnCollector`. `DerivedFnInfo` stores all the information required to uniquely identify a derivative and also stores the results that we need, derivative declaration and overloaded derivative declaration.
`DerivedFnCollector` stores `DerivedFnInfo` object of all the computed derived functions and provide utilities to easily find a derivative that satisfies the current request.

vgvassilev:
> > The lookup is an easy way out. I think we need the information about what was produced in various places anyway. Check `m_GeneratedDerivatives` and `m_Derivatives`. In these places we store the `Decl*`. Instead, we could store more information about the `DiffRequest`, or just some `DeclarationName`-> `Decl*` mapping which we can look up when constructing the diff request...
> 
> Just `DeclarationName` -> `Decl*` mapping will not be enough to uniquely identify a derivative. Because the same function can be differentiated in several different ways, such as different modes, different independent arguments and different requested derivative orders. And we also cannot just use `DeclarationName` due to the same name at different scope issues.
> 
> We can store all this required information to uniquely find the derivative that satisfies the current differentiation request. But we also need to maintain 2 maps, one for derivative declaration and the other for overloaded derivative declaration.
> 
> I have tried to solve this problem using `DerivedFnInfo` class and `DerivedFnCollector`. `DerivedFnInfo` stores all the information required to uniquely identify a derivative and also stores the results that we need, derivative declaration and overloaded derivative declaration. `DerivedFnCollector` stores `DerivedFnInfo` object of all the computed derived functions and provide utilities to easily find a derivative that satisfies the current request.

Yeah, the `DeclarationName` was an (over)simplification. I do not mind having the `DerivedFnInfo` instead of the `Decl*` of the `m_Derivatives`. If we want to disambiguate between different overloads we can generally speaking store the function signature of the derived function...

parth-07:
> Yeah, the DeclarationName was an (over)simplification. I do not mind having the DerivedFnInfo instead of the Decl* of the m_Derivatives. If we want to disambiguate between different overloads we can generally speaking store the function signature of the derived function...

To disambiguate between different overloads, `DerivedFnInfo` class provides an interface to access the derived function and the overloaded derived function (both are required for reverse mode derivatives). 
`DerivedFnCollector` just contains a mapping from `OriginalFunctionDecl*` -> `DerivedFnInfo` and contains functions to find a particular derivative. Would you prefer a simpler alternate than `DerivedFnCollector` here?

vgvassilev:
> > > The lookup is an easy way out. I think we need the information about what was produced in various places anyway. Check `m_GeneratedDerivatives` and `m_Derivatives`. In these places we store the `Decl*`. Instead, we could store more information about the `DiffRequest`, or just some `DeclarationName`-> `Decl*` mapping which we can look up when constructing the diff request...
> > 
> > 
> > Just `DeclarationName` -> `Decl*` mapping will not be enough to uniquely identify a derivative. Because the same function can be differentiated in several different ways, such as different modes, different independent arguments and different requested derivative orders. And we also cannot just use `DeclarationName` due to the same name at different scope issues.
> > We can store all this required information to uniquely find the derivative that satisfies the current differentiation request. But we also need to maintain 2 maps, one for derivative declaration and the other for overloaded derivative declaration.
> > I have tried to solve this problem using `DerivedFnInfo` class and `DerivedFnCollector`. `DerivedFnInfo` stores all the information required to uniquely identify a derivative and also stores the results that we need, derivative declaration and overloaded derivative declaration. `DerivedFnCollector` stores `DerivedFnInfo` object of all the computed derived functions and provide utilities to easily find a derivative that satisfies the current request.
> 
> Yeah, the `DeclarationName` was an (over)simplification. I do not mind having the `DerivedFnInfo` instead of the `Decl*` of the `m_Derivatives`. If we want to disambiguate between different overloads we can generally speaking store the function signature of the derived function...

Ok, thanks for guiding me through this. I think what we have is good. Can we move the `DerivedFnCollector` in `ClangPlugin.h` and merge it with `m_Derivatives`?

vgvassilev:
> > > The lookup is an easy way out. I think we need the information about what was produced in various places anyway. Check `m_GeneratedDerivatives` and `m_Derivatives`. In these places we store the `Decl*`. Instead, we could store more information about the `DiffRequest`, or just some `DeclarationName`-> `Decl*` mapping which we can look up when constructing the diff request...
> > 
> > 
> > Just `DeclarationName` -> `Decl*` mapping will not be enough to uniquely identify a derivative. Because the same function can be differentiated in several different ways, such as different modes, different independent arguments and different requested derivative orders. And we also cannot just use `DeclarationName` due to the same name at different scope issues.
> > We can store all this required information to uniquely find the derivative that satisfies the current differentiation request. But we also need to maintain 2 maps, one for derivative declaration and the other for overloaded derivative declaration.
> > I have tried to solve this problem using `DerivedFnInfo` class and `DerivedFnCollector`. `DerivedFnInfo` stores all the information required to uniquely identify a derivative and also stores the results that we need, derivative declaration and overloaded derivative declaration. `DerivedFnCollector` stores `DerivedFnInfo` object of all the computed derived functions and provide utilities to easily find a derivative that satisfies the current request.
> 
> Yeah, the `DeclarationName` was an (over)simplification. I do not mind having the `DerivedFnInfo` instead of the `Decl*` of the `m_Derivatives`. If we want to disambiguate between different overloads we can generally speaking store the function signature of the derived function...

Ok, thanks for guiding me through this. I think what we have is good. Can we move the `DerivedFnCollector` in `ClangPlugin.h` and merge it with `m_Derivatives`?

parth-07:
Yes sure, I will move it with `ClangPlugin.h` and merge the functionality with `m_Derivatives`.

parth-07:
Yes sure, I will move it with `ClangPlugin.h` and merge the functionality with `m_Derivatives`.

parth-07:
parth.2916403218@ipu.ac.in

tools/ClangPlugin.h
```diff
@@ -32,6 +38,40 @@ namespace clang {
...
+    /// Adds a derived function to the collection.
+    void Add(const DerivedFnInfo& DFI);
+
+    /// Finds a `DerivedFnInfo` object in the collection that satisfies the
+    /// given differentiation request.
+    DerivedFnInfo Find(const DiffRequest& request) const;
+
+    bool IsDerivative(const clang::FunctionDecl* FD) const;
+
+    const DerivativesSet& Derivatives() const {
```
vgvassilev, line 65:
```suggestion
    const DerivativesSet& getGeneratedDerivatives() const {
```

tools/ClangPlugin.h
```diff
@@ -32,6 +38,40 @@ namespace clang {
...
+  /// making it possible to reuse previously computed derivatives.
+  class DerivedFnCollector {
+    using DerivedFns = llvm::SmallVector<DerivedFnInfo, 16>;
+    /// Mapping to efficiently find out information about all the derivatives of
+    /// a function.
+    std::map<const clang::FunctionDecl*, DerivedFns> m_DerivedFnInfoCollection;
+
+    // Require to efficiently determine if a declaration is of a derivative
+    // computed by clad
+    DerivativesSet m_Derivatives;
```
vgvassilev, line 53: I think we can remove this member. We seem to only use it in the `DiffCollector::DiffCollector` to decide if we should skip over the already generated code. This can be solved at the level of the caller `CladPlugin::HandleTopLevelDecl`.

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -1,5 +1,8 @@
+#ifndef CLAD_DIFF_PLANNER_H
+#define CLAD_DIFF_PLANNER_H
+#include "clad/Differentiator/DiffMode.h"
```
vgvassilev, line 3:
```suggestion

#include "clad/Differentiator/DiffMode.h"
```

include/clad/Differentiator/DerivedFnInfo.h
```diff
@@ -0,0 +1,44 @@
+#ifndef CLAD_DERIVED_FN_H
```
vgvassilev, line 1: Could this header and its corresponding source file go in tools? Maybe they can become local to ClangPlugin.h/cpp.

include/clad/Differentiator/DerivedFnCollector.h
```diff
@@ -0,0 +1,11 @@
...
+#define CLAD_DERIVED_FN_COLLECTOR
+
+
+#include "clad/Differentiator/DiffPlanner.h"
+
+namespace clad {
+  
+} // namespace clad
+
+#endif
```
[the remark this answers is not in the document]
parth-07: ah yes, sorry I forgot to remove this.

include/clad/Differentiator/DerivedFnInfo.h
parth-07: Yes, that sounds good.

include/clad/Differentiator/ParseDiffArgsTypes.h
```diff
@@ -0,0 +1,34 @@
+#ifndef CLAD_PARSE_DIFF_ARGS_TYPES_H
+#define CLAD_PARSE_DIFF_ARGS_TYPES_H
+#include "clang/AST/Decl.h"
+#include "llvm/ADT/SmallVector.h"
+
+#include <cstddef>
+#include <utility>
+
+namespace clad {
+  struct IndexInterval {
```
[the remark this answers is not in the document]
parth-07: Yes, sure!

tools/ClangPlugin.h
```diff
@@ -32,6 +38,40 @@ namespace clang {
...
   class EstimationPlugin;
+
+  /// This class is designed to store collection of `DerivedFnInfo` objects.
+  /// It's purpose is to avoid repeated generation of same derivatives by
+  /// making it possible to reuse previously computed derivatives.
+  class DerivedFnCollector {
+    using DerivedFns = llvm::SmallVector<DerivedFnInfo, 16>;
+    /// Mapping to efficiently find out information about all the derivatives of
+    /// a function.
+    std::map<const clang::FunctionDecl*, DerivedFns> m_DerivedFnInfoCollection;
```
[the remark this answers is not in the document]
parth-07: Yes, I will replace std::map with llvm::DenseMap here.

tools/ClangPlugin.h
parth-07: > This can be solved at the level of the caller CladPlugin::HandleTopLevelDecl

How can this be solved in `CladPlugin::HandleTopLevelDecl`. 

Please tell me if I am wrong here. I am also not sure why we require this in `DiffCollector::DiffCollector`, we don't seem to traverse the same AST again in the `DiffCollector::DiffCollector` functions, and we only add derivatives in the AST that's already traversed.

vgvassilev: We do seem to call HandleTopLevelDecl here https://github.com/vgvassilev/clad/blob/def47c0a0bee649dcf76b554b928821580262506/tools/ClangPlugin.cpp#L141 However, now we flip a flag `m_HandleTopLevelDeclInternal` which we can check and ignore even before we reach the `DiffCollector` so all that seems indeed dead code..

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -8,6 +8,7 @@
 #define CLAD_DERIVATIVE_BUILDER_H
 
 #include "Compatibility.h"
+#include "clad/Differentiator/ParseDiffArgsTypes.h"
```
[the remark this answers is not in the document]
parth-07: We do not, I don't know why I added it here. I have removed this include now.

tools/ClangPlugin.h
parth-07: ah yes, we are calling `HandleTopLevelDecl` for derivatives as well.
I have removed the `DerivativesSet` and the related dead code.

parth-07:
> @parth-07, can you resolve the conflict?

Yes, I have resolved the conflict.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -326,6 +323,177 @@ namespace clad {
...
+    DiffParams params{};
+    auto E = diffArgs->IgnoreParenImpCasts();
+    // Case 1)
+    if (auto SL = dyn_cast<StringLiteral>(E)) {
+      IndexIntervalTable indexes{};
+      llvm::StringRef string = SL->getString().trim();
+      if (string.empty()) {
+        utils::EmitDiag(semaRef, DiagnosticsEngine::Error,
+                        diffArgs->getEndLoc(), "No parameters were provided");
+        return;
```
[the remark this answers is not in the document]
parth-07: Added.

vgvassilev (approved):
Thanks! LGTM!

PR #347, opened 2022-01-10
debug

vgvassilev:


vgvassilev:
MacOS11 builders seem not to make a difference between VERSION and version. In one of our error estimation framework demos we add an include path to the source folder or clad, exposing clad/VERSION to clang. Then somewhere deep in the system libraries we have `#include <version>` which is now being resolved to clad/VERSION.

@parth-07 kindly agreed to submit a fix.

PR #363, opened 2022-01-29
Make clang-format more llvm-compliant

vgvassilev:
Hi @parth-07 and @sudo-panda, maybe we should be just use the formatting close to llvm. What do you think?

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,23 +1,8 @@
-
 BasedOnStyle: LLVM
-IndentWidth: 2
-NamespaceIndentation: All
-ColumnLimit: 80
 
 Language: Cpp
 Standard: Cpp11
-DerivePointerAlignment: false
 PointerAlignment: Left
-
-AlignAfterOpenBracket: Align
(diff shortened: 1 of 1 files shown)


parth-07:
I think it's a good idea. 
I also think we should run clang-format on the entire codebase once we finalise the `clang-format` script.

vgvassilev:
> I think it's a good idea. I also think we should run clang-format on the entire codebase once we finalise the `clang-format` script.

If we run it then `git blame` becomes harder. Ideally I want this to get in before we start moving headers in lib as part of hiding the redundant includes from `include`.

parth-07:
> If we run it then git blame becomes harder.

Oh, I didn't consider this.

sudo-panda:
> Hi @parth-07 and @sudo-panda, maybe we should be just use the formatting close to llvm. What do you think?

Yep that would be best.

vgvassilev:
Ok, then I will merge it now..

PR #365, opened 2022-01-31
Basic virtual fn support in forward mode.

alexander-penev:
We can differentiate virtual class functions with several caveats:
```
struct A {
  virtual double f(double x) {}
  virtual double f_darg0(double x); // clad will provide a definition
};
```

### DIFF
--- a/demos/ComputerGraphics/smallpt/SmallPT.cpp
+++ b/demos/ComputerGraphics/smallpt/SmallPT.cpp
@@ -1,7 +1,7 @@
 //--------------------------------------------------------------------*- C++ -*-
 // clad - The C++ Clang-based Automatic Differentiator
 //
-// A demo, describing how to use clad in simple Path tracer.
+// A demo, describing how to use clad in simple AD-aware Path tracer.
 //
 // author:  Alexander Penev <alexander_penev-at-yahoo.com>
 //          Based on smallpt, a Path Tracer by Kevin Beason, 2008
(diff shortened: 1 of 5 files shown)


demos/ComputerGraphics/SmallPT.cpp
```diff
@@ -431,26 +387,26 @@ Vec radiance(const Ray &ray, int depth, unsigned short *Xi) {
...
+    double p = f.x>f.y && f.x>f.z ? f.x : f.y>f.z ? f.y : f.z; // max refl
 
     // object intersection id map test
     //return obj.c;
 
     cl = cl + cf.mult(obj.e);
 
-    if (++depth>5) {
-      if (erand48(Xi)<p) f=f*(1/p); else return cl;
+    if (++depth>5 || !p) {
```
vgvassilev, line 397: Why do we need this condition now?

demos/ComputerGraphics/SmallPT.cpp
```diff
@@ -431,26 +387,26 @@ Vec radiance(const Ray &ray, int depth, unsigned short *Xi) {
...
+      if (erand48(Xi)<p) f=f*(1/p); else return cl*E;
```
vgvassilev, line 398: Is that change intended?

vgvassilev (commented):
This is really great! Thanks for working on this!!

demos/ComputerGraphics/SmallPT.cpp
alexander-penev: This in not needed anymore. Removed.

demos/ComputerGraphics/SmallPT.cpp
alexander-penev: Not needed. Source is changed now.

vgvassilev (approved):
LGTM! Thanks!

PR #369, opened 2022-02-01
Fix missing cases for repeated generation of derivatives

parth-07:
Missing cases were as follows:
- We were not generating index interval table for default arguments
- We were not considering that diff params specified in different
  order will produce different DiffParamsInfo object

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -370,6 +370,23 @@ namespace clad {
           candidates.emplace_back(PVD->getName(), PVD);
       }
 
+      // Ensure that diff params are always considered in the same order.
+      std::sort(
+          names.begin(), names.end(),
+          [&candidates](llvm::StringRef a, llvm::StringRef b) {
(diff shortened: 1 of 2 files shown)


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -370,6 +370,23 @@ namespace clad {
           candidates.emplace_back(PVD->getName(), PVD);
       }
 
+      // Ensure that diff params are always considered in the same order.
+      std::sort(
+          names.begin(), names.end(),
+          [&candidates](llvm::StringRef a, llvm::StringRef b) {
+            auto a_it = std::find_if(
```
vgvassilev, line 377: I think we can simplify these loops if we made `names` of type `llvm::SetVector`.

vgvassilev: We plan bigger refactoring here, let's skip this and merge the PR if the CI agrees.

vgvassilev (approved):
Thanks for the quick fix! LGTM!

PR #370, opened 2022-02-02
Add ability to print FP errors in error estimation.

### LINKED ISSUE #295: [Feature Request] Floating Point Error Estimation Framework to do checklist
This issue is to keep track of features to implement in the floating-point error estimation framework.

**Good things to do:**

 - [x] Add a demo to demonstrate how clad can be used to detect numerical instability of algorithms.
*The above is blocked due to the fact that we have no way to automatically print error stats yet.*
 - [x] Add more tests for loops and arrays
*Blocked because of #245 and #246 .*
 - [x] Fix links in the custom model README.md 
*The above is blocked because we do not have tutorials hosted anywhere as of yet.*
- [x] ~~Optimize stores
*There are places where we can optimize our store, however, it introduces a certain amount of complexity into the code which might not be a good tradeoff. This is blocked because some discussion should be done relating to this.*~~
 - [x] Update clad README.md
*Blocked because some discussion is needed here.*

**Future work:**

 - Addition of code to implement forward propagation of errors would be an interesting project. This will help us generalize the error estimation framework.
 - Lossy compression and precision tuning. This can be implemented fairly easily, however, we would need to make another pass of the function to actually compress variables.

grimmmyshini:
Allow users to print intermediate floating-point errors to a file. This PR will allow users to use the "-fprint-error-ests-to-file" and enable printing of whatever they specify in the custom model's ```Print``` function. This will not only enable users to get access to intermediate error and derivative values (something they would not have otherwise) automatically but also allow them a deeper insight into the error analysis. One can perform analysis like error sensitivity of variables through this feature, all they have to do is implement the ```Print``` function as such it prints the sensitivity of each variable of interest.

 A possible pitfall of this implementation is that we have no way (at least not a non-complex way) to attach source information to the "Print" values of each variable, i.e. currently we print to the file something like this: "var_name : <--expr specified by Print-->". Now, this is very terse and can easily become confusing for larger codes where the same variable appears multiple times as the variable of interest. A better implementation would see the printing of something like: "var_name : var_src_line_no : <--expr specified by Print-->" where the ```var_src_line_no``` specifies the line number of the occurrence of that specific variable, tying each "Print" value to an expression in the code.

- [ ] ~Add a demo for sensitivity analysis. (closes #295)~
- [ ] Update the website tutorial to add the ```Print``` function there.

### DIFF
--- a/demos/ErrorEstimation/CustomModel/CustomModel.h
+++ b/demos/ErrorEstimation/CustomModel/CustomModel.h
@@ -11,13 +11,13 @@
 #include "clad/Differentiator/EstimationModel.h"
 
 /// This is our dummy estimation model class.
-// We will be using this to override the virtual function in the
-// FPErrorEstimationModel class.
+/// We will be using this to override the virtual function in the
+/// FPErrorEstimationModel class.
 class CustomModel : public clad::FPErrorEstimationModel {
 public:
   CustomModel(clad::DerivativeBuilder& builder)
       : FPErrorEstimationModel(builder) {}
-  // Return an expression of the following kind:
-  //  dfdx * delta_x
+  /// Return an expression of the following kind:
+  ///  dfdx * delta_x
   clang::Expr* AssignError(clad::StmtDiff refExpr);
 };
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -137,6 +137,11 @@ namespace clad {
     /// an in-built one (TaylorApprox) or one provided by the user.
     void
     SetErrorEstimationModel(std::unique_ptr<FPErrorEstimationModel> estModel);
+    /// Function to initialize the \c ostream object if error printing was
+    /// requested.
+    ///
+    /// \returns the equivalent QualType of \c ostream objects.
+    clang::QualType GetErrorFileType();
     /// Fuction to set the error diagnostic printing value for numerical
     /// differentiation.
     ///
--- a/include/clad/Differentiator/ErrorEstimator.h
+++ b/include/clad/Differentiator/ErrorEstimator.h
@@ -44,11 +44,16 @@ class ErrorEstimationHandler : public ExternalRMVSource {
   clang::Expr* m_IdxExpr;
   /// A set of declRefExprs for parameter value replacements.
   std::unordered_map<const clang::VarDecl*, clang::Expr*> m_ParamRepls;
+  /// A variable to hold the type of the "ofstream" object.
+  clang::QualType m_ErrorFileType;
+  /// A variable to keep track of if error printing is enabled.
+  bool m_PrintErrors = false;
+  /// A variable to store the "ofstream" object.
+  clang::DeclRefExpr* m_ErrorFile = nullptr;
 
   std::stack<bool> m_ShouldEmit;
   ReverseModeVisitor* m_RMV;
   clang::Expr* m_DeltaVar = nullptr;
-  llvm::SmallVectorImpl<clang::QualType>* m_ParamTypes = nullptr;
   llvm::SmallVectorImpl<clang::ParmVarDecl*>* m_Params = nullptr;
 
 public:
@@ -58,14 +63,17 @@ public:
         m_EstModel(nullptr), m_IdxExpr(nullptr) {}
   ~ErrorEstimationHandler() {}
 
+  /// A function to set the \c ofstream file type.
+  void SetErrorFileType(clang::QualType qt) { m_ErrorFileType = qt; }
+
+  /// A function to enable error printing.
+  void EnableErrorPrinting() { m_PrintErrors = true; }
+
   /// Function to set the error estimation model currently in use.
   ///
   /// \param[in] estModel The error estimation model, can be either
   /// an in-built one (TaylorApprox) or one provided by the user.
   void SetErrorEstimationModel(FPErrorEstimationModel* estModel);
-
-  /// \param[in] finErrExpr The final error expression.
-  void SetFinalErrorExpr(clang::Expr* finErrExpr) { m_FinalError = finErrExpr; }
 
   /// Shorthand to get array subscript expressions.
   ///
@@ -81,6 +89,20 @@ public:
   /// \returns The final error expression so far.
   clang::Expr* GetFinalErrorExpr() { return m_FinalError; }
 
+  /// Function to generate an expression of type:
+  /// '<--ostream-object--> << "<--source-info-->" << <--expression-->;'
+  ///
+  /// \param[in] var_name The name of the variable whose error contribution we
+  /// are printing.
+  /// \param[in] var The variable whose error contribution we
+  /// are printing.
+  /// \param[in] var_dx The derivative of 'var'.
+  /// \param[in] var_err The error associated with 'var'.
+  ///
+  /// \returns The print expression to be added to code.
+  clang::Expr* GetPrintExpr(std::string var_name, clang::Expr* var,
+                            clang::Expr* var_dx, clang::Expr* var_err);
+
   /// Function to build the final error statemnt of the function. This is the
   /// last statement of any target function in error estimation and
   /// aggregates the error in all the registered variables.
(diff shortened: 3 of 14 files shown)


grimmmyshini:
Once I rebase against #304, I will switch the ```SetErrorObject()``` method to use ```clad::utils::CreateStringLiteral``` instead of ```ASTContext::getStringLiteralArrayType``` (which probably is not present everywhere).

demos/ErrorEstimation/CustomModel/CustomModel.cpp
```diff
@@ -10,6 +10,10 @@ clang::Expr* CustomModel::AssignError(clad::StmtDiff refExpr) {
 
 clang::Expr* CustomModel::SetError(clang::VarDecl* decl) { return nullptr; }
 
+clang::Expr* CustomModel::Print(clad::StmtDiff refExpr, clang::Expr* errExpr) {
+  return nullptr;
```
vgvassilev, line 14: We were bitten by this type of test coverage reports. Could you develop a test with an empty custom model? That should appease these reports.

demos/ErrorEstimation/CustomModel/CustomModel.h
```diff
@@ -23,4 +23,7 @@ class CustomModel : public clad::FPErrorEstimationModel {
 
   // Return a nullptr here, clad will interpret that as a 0 literal.
   clang::Expr* SetError(clang::VarDecl* decl);
+
+  // For now, we can return a nullptr here.
```
vgvassilev, line 27: Can you switch this line and the few lines above (including the CustomModel docs) to doxygen-style documentation? (Eg add /// before each comment)

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -132,19 +135,35 @@ namespace clad {
     /// an in-built one (TaylorApprox) or one provided by the user.
     void
     SetErrorEstimationModel(std::unique_ptr<FPErrorEstimationModel> estModel);
+    /// Function to initialize the "ofstream" object if error printing was
```
vgvassilev, line 138:
```suggestion
    /// Function to initialize the \c ofstream object if error printing was
```

test/ErrorEstimation/PrintErrors.C
```diff
@@ -0,0 +1,77 @@
...
+#include <fstream> // Necessary for printing data to file.
+
+float func1(float x, float y) {
+  x = x - y - y * y;
+  float t = x * y * y;
+  return t;
+}
+
+//CHECK: void func1_grad(float x, float y, clad::array_ref<float> _d_x, clad::array_ref<float> _d_y, double &_final_error) {
+//CHECK-NEXT:     {{.+}}ofstream _error_file0("error_file");
```
vgvassilev, line 16: Can we take a `func1_grad` take `ostream &out` and accumulate the output there. The advantage is that we can invoke it with a stream of choice including streamstream.

vgvassilev (commented):
Thanks for working on this. I think we should improve the test coverage, and run clang-format if you have not.

test/ErrorEstimation/PrintErrors.C
grimmmyshini: Yeah sure, we can do that.

include/clad/Differentiator/ErrorEstimator.h
```diff
@@ -44,29 +44,37 @@ class ErrorEstimationHandler : public ExternalRMVSource {
   clang::Expr* m_IdxExpr;
   /// A set of declRefExprs for parameter value replacements.
   std::unordered_map<const clang::VarDecl*, clang::Expr*> m_ParamRepls;
+  /// A variable to hold the type of the "ofstream" object.
+  clang::QualType m_ErrorFileType;
+  /// A variable to keep track of if error printing is enabled.
+  bool m_PrintErrors = false;
+  /// A variable to store the "ofstream" object.
+  clang::DeclRefExpr* m_ErrorFile;
```
vgvassilev, line 52:
```suggestion
  clang::DeclRefExpr* m_ErrorFile = nullptr;
```

include/clad/Differentiator/ErrorEstimator.h
```diff
@@ -81,6 +89,20 @@ class ErrorEstimationHandler : public ExternalRMVSource {
...
+  /// \param[in] var_name The name of the variable whose error contribution we
+  /// are printing.
+  /// \param[in] var The variable whose error contribution we
+  /// are printing.
+  /// \param[in] var_dx The derivative of 'var'.
+  /// \param[in] var_err The error associated with 'var'.
+  ///
+  /// \returns The print expression to be added to code.
+  clang::Expr* GetPrintExpr(std::string var_name, clang::Expr* var,
+                              clang::Expr* var_dx, clang::Expr* var_err);
```
vgvassilev, line 104: Did you run clang-format on this?

lib/Differentiator/ErrorEstimator.cpp
```diff
@@ -496,13 +543,17 @@ void ErrorEstimationHandler::EmitDeclErrorStmts(VarDeclDiff VDDiff,
...
+      Expr* errorExpr = GetError(savedDecl.getExpr_dx(), varDeclExpr);
+      Expr* prntExpr =
+          GetPrintExpr(VD->getNameAsString(), savedDecl.getExpr_dx(),
+                       varDeclExpr, errorExpr);
+      AddErrorStmtToBlock(VDRef, EstVD, errorExpr, isInsideLoop, prntExpr);
     }
   }
 }
 
+
```
vgvassilev, line 556: Stray line?

lib/Differentiator/EstimationModel.cpp
```diff
@@ -50,6 +55,12 @@ namespace clad {
     return nullptr; 
   }
 
+  llvm::SmallVector<Expr*, 8> FPErrorEstimationModel::Print(std::string varName,
```
vgvassilev, line 58: We should not be returning an `std::vector` or `SmallVector` due to lower performance. Can we take it as a by-ref argument to the function?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -421,6 +421,7 @@ namespace clad {
       m_Variables[param] = BuildDeclRef(VDDerived);
       addToBlock(BuildDeclStmt(VDDerived), m_Globals);
     }
+
```
vgvassilev, line 424: Another stray line. Maybe run clang-format?

vgvassilev (commented):
Thanks, this is really looking good!

grimmmyshini:
I will squash and format the changes once the changes get a green light.

grimmmyshini:
@vgvassilev Can you greenlight the changes before I squash and format? Or would you rather review after I do those things?

vgvassilev:
I had some suggestions but this is good to go. Please squash and format.

include/clad/Differentiator/ErrorEstimator.h
vgvassilev: ping.

vgvassilev (commented):
LGTM, modulo the comments.

grimmmyshini:
@vgvassilev, we have to pause this, test failures due to #459

include/clad/Differentiator/ErrorEstimator.h
grimmmyshini: It is being initialized in the c'tor.

grimmmyshini:
As per @parth-07's comments, added the same workaround for #459

parth-07:
> As per @parth-07's comments, added the same workaround for #459

Which workaround did you use? And can you please briefly explain why was any workaround required for error estimation here?

include/clad/Differentiator/ErrorEstimator.h
vgvassilev: It does not hurt to initialize it in place.

grimmmyshini:
> > As per @parth-07's comments, added the same workaround for #459
> 
> Which workaround did you use? And can you please briefly explain why was any workaround required for error estimation here?

By workaround I mean using what is done everywhere else, i.e. `OutputParamType_t<Args, void>`.

parth-07:
> > Which workaround did you use? And can you please briefly explain why was any workaround required for error estimation here?
> 
> By workaround I mean using what is done everywhere else, i.e. `OutputParamType_t<Args, void>`.

Oh okay, now I understand. Thanks!

include/clad/Differentiator/ErrorEstimator.h
grimmmyshini: Yes, sure. I was just trying to be consistent with the other initializations.

vgvassilev:
@grimmmyshini apologies for the delay, can you rebase?

vgvassilev:
I was wondering what is the fate of this PR?

vgvassilev:
Is #512 strictly better?

grimmmyshini:
I don't know if it's strictly better  but i don't see any reason for not using the print model... This version is very rigid.

vgvassilev:
Okay, then please close this PR.

PR #376, opened 2022-02-14
[ci] Fix the benchmark comparisons.

vgvassilev:
This patch tries to fix the PR benchmark results comparisons as we currently compare the baseline against the baseline. It adds more robust logic for after-merge checks where we fail to deduce the hash and now we explicitly branch the baseline.


vgvassilev:
This was resolved by @vaithak in the master.

PR #379, opened 2022-02-18
Modify clad function to throw error on mismatch of arguments

### LINKED ISSUE #378: Miss-match in args number should throw error
It seems that the execute function accepts less arguments due to a set of defaults. In the case shown bellow, an error should be raised, rather than returning 0.  

image

@sudo-panda @davidlange6

sudo-panda:
Fixes #378 

The padding of supplied args with default args has been removed from all modes other than reverse mode. So they will throw error if less number of args are supplied.

For reverse mode the plugin needs to provide the `CladFunction` with the number of arguments so that it can check if the correct number of arguments are provided. This is yet to be implemented.

### DIFF
--- a/test/ForwardMode/NotEnoughArgError.C
+++ b/test/ForwardMode/NotEnoughArgError.C
@@ -1,0 +1,18 @@
+// RUN: %cladclang %s -I%S/../../include -fsyntax-only -Xclang -verify 2>&1
+// CHECK-NOT: {{.*warning|note:.*}}
+
+#include "clad/Differentiator/Differentiator.h"
+
+double func1(double x, double y) {
+  return x * x + y * y;
+}
+
+int main () {
+  auto func1_dx = clad::differentiate(func1, "x");
+
+  func1_dx.execute(5);
+  // expected-error@clad/Differentiator/Differentiator.h:113 {{too few arguments to function call, expected 2, have 1}}
+  // expected-note@clad/Differentiator/Differentiator.h:245 {{in instantiation of function template specialization 'clad::execute_with_default_args<false, double, double (*)(double, double), int, true>' requested here}}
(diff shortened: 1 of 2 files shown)


test/ForwardMode/NotEnoughArgError.C
```diff
@@ -0,0 +1,14 @@
...
+
+#include "clad/Differentiator/Differentiator.h"
+
+double func1(double x, double y) {
+  return x * x + y * y;
+}
+
+int main () {
+  auto func1_dx = clad::differentiate(func1, "x");
+  func1_dx.execute(5);
```
vgvassilev, line 13:
```suggestion
  func1_dx.execute(5); // expected-error {{some text here}}
```

Please remove the XFAIL.

sudo-panda: The error is not thrown there:
```cpp
/media/sudo-panda/D/workspaces/clad/src/tools/clad/include/clad/Differentiator/Differentiator.h:113:40: error: too few arguments to function call, expected 2, have 1
           ~                           ^
/media/sudo-panda/D/workspaces/clad/src/tools/clad/include/clad/Differentiator/Differentiator.h:245:16: note: in instantiation of function template specialization 'clad::execute_with_default_args<false, double, double (*)(double, double), int, true>' requested here
        return execute_with_default_args<EnablePadding>(
               ^
/media/sudo-panda/D/workspaces/clad/src/tools/clad/include/clad/Differentiator/Differentiator.h:196:14: note: in instantiation of function template specialization 'clad::CladFunction<double (*)(double, double), clad::NoObject, false>::execute_helper<double (*)(double, double), int>' requested here
      return execute_helper(m_Function, static_cast<Args>(args)...);
             ^
../src/tools/clad/test/ForwardMode/Errors.C:13:12: note: in instantiation of function template specialization 'clad::CladFunction<double (*)(double, double), clad::NoObject, false>::execute<int, double (*)(double, double)>' requested here
  func1_dx.execute(5); //expected-error {{too few arguments to function call, expected 2, have 1}}
           ^
```

sudo-panda: Works!

PR #383, opened 2022-02-23
Add initial support for pullback functions

### LINKED ISSUE #281: Clad does not generate derivatives of functions with reference inputs correctly.
**[The thread of interest](https://github.com/vgvassilev/clad/pull/247#discussion_r668181712)**

**Problem:**

No, for the following example:
```cpp
void helper(float& a){
    a = a*a;
}

float func(float x){
    helper(x);
    return x;
}
```
the following gradient is generated:
```cpp
void func_grad(float x, clad::array_ref<float> _d_x, double &_final_error) {
    helper(x);
    float func_return = x;
    goto _label0;
  _label0:
    * _d_x += 1;
   // double _delta_x = 0;
   // _delta_x += * _d_x * x * 1.1920928955078125E-7;
   //  _final_error += _delta_x;
}
```
The gradient is obviously wrong.

**Proposed solution:**

we should somehow propagate the changes forward if any parameter is an L value reference type. I think we could do this by just forwarding the ```array_ref``` for the parameter passed as reference instead of creating a temporary ```_grad*```.

Then the gradient would look something like this:
```cpp
void helper_grad(float &a, clad::array_ref<float> _d_a) {
    float _t0;
    float _t1;
    _t1 = a;
    _t0 = a;
    a = _t1 * _t0;
    {
        float _r_d0 = * _d_a;
        float _r0 = _r_d0 * _t0;
        * _d_a += _r0;
        float _r1 = _t1 * _r_d0;
        * _d_a += _r1;
        * _d_a -= _r_d0;
        * _d_a;
    }
}
void func_grad(float x, clad::array_ref<float> _d_x, double &_final_error) {
    helper(x);
    float func_return = x;
    goto _label0;
  _label0:
    * _d_x += 1;
    helper_grad(x, _d_x);
```
**A generalized poc:**

I feel like what I proposed *should* work. Again, I suppose this requires some debate and some input from some more experienced folks but here is my argument:

Any function call can essentially be swapped into the original function code (or inlined to be more precise), so different combinations and their possible inlines may look like the following:

This is the target function we will use, it will contain a call to a void function or a call-assign to a non-void function
```cpp
float func(float x, float y){
    // some ops
    y = helper_*(x, y);
    //OR
    helper_ *(x, y);
    // some ops
}
```
---------------------------------------------
*Only one reference variable + no return.*
```cpp
void helper_void_ref(float& x){
    x = x * x;
}
```
results in the following conversion:
```cpp
float func(float x, float y){
    // some ops
    helper_void_ref(x);
    // some ops
}
```
TO: 
```cpp
float func(float x, float y){
    // some ops
    x = x * x;
    // some ops
}
```
---------------------------------------------
*One/pure reference variable(s) with non-void return.*
```cpp
float helper_ret_ref(float& x){
    return (x = x * x);
}
```
results in the following conversion:
```cpp
float func(float x, float y){
    // some ops
    y = helper_ret_ref(x);
    // some ops
}
```
TO: 
```cpp
float func(float x, float y){
    // some ops
    x = x * x;
    y = x;
    // some ops
}
```
---------------------------------------------
*Pass by value with non void return (usual case).*
```cpp
float helper_ret_val(float x){
    return (x = x * x);
}
```
results in the following conversion:
```cpp
float func(float x, float y){
    // some ops
    y = helper_ret_val(x);
    // some ops
}
```
TO: 
```cpp
float func(float x, float y){
  // some ops
  // pass by val so make a copy for x, use the temp value from this point on.
  float _temp_x = x;
  _temp_x = _temp_x * _temp_x;
  y = _temp_x;
  // some ops; use x as usual beyond this point, destruct temp.
}
```
---------------------------------------------
*Mixed call with void return.*
```cpp
void helper_void_mixed(float& x,  float y){
    x = x * y;
}
```
results in the following conversion:
```cpp
float func(float x, float y){
    // some ops
    helper_void_mixed(x, y);
    // some ops
}
```
TO: 
```cpp
float func(float x, float y){
  // some ops
  // pass by val so make a copy for y, use the temp value from this point on for y.
  // pass by ref for x so use x as is.
  float _temp_y = y;
  x = x * _temp_y;
  // some ops; use y as usual beyond this point, destruct temp_y.
}
```
---------------------------------------------
*Mixed call with non-void return.*
```cpp
float helper_ret_mixed(float& x,  float y){
    x = x * x;
    y = x;
    return y;
}
```
results in the following conversion:
```cpp
float func(float x, float y){
    // some ops
    y = helper_ret_mixed(x, y);
    // some ops
}
```
TO: 

```cpp
float func(float x, float y){
  // some ops
  // pass by val so make a copy for y, use the temp value from this point on for y.
  // pass by ref for x so use x as is.
  float _temp_y = y;
  x = x * x;
  _temp_y = x;
  // return assign here:
  y = _temp_y;
  // some ops; use y as usual beyond this point, destruct temp_y.
}
```
These are good comparisons to test my logic. I feel like we can prove here by how passing the same array ref (and a new temp one for pass by value) results in correct results.

Another "peculiar" case is functions with multiple references and a void return. One way is to actually inline these functions into the code and then continue the differentiation as is. One way to achieve this is to retain the tape types and then differentiate the function normally. To do this, we just have to assume that the void return function actually has a non void type (return type of the target function) and returns a random constant of that type. For example, consider the following:

```cpp
void helperR(float& x, float& y){
    x = x*x;
    y += x;
}

float randomR(float x, float y){
    helperR(x, y);
    return x + y;
}
```

Now, lets take two cases: 

1. If we inline the function directly:

*Transformed functions:*
```cpp
float randomR(float x, float y){
    x = x*x;
    y += x;
    return x + y;
}
```

*Derivative Produced:*
```cpp
void randomR_grad(float x, float y, clad::array_ref<float> _d_x, clad::array_ref<float> _d_y) {
    float _t0;
    float _t1;
    _t1 = x;
    _t0 = x;
    x = _t1 * _t0;
    y += x;
    float randomR_return = x + y;
    goto _label0;
  _label0:
    {
        * _d_x += 1;
        * _d_y += 1;
    }
    {                                      // Block of interest starts here
        float _r_d1 = * _d_y;
        * _d_y += _r_d1;
        * _d_x += _r_d1;
        * _d_y -= _r_d1;
        * _d_y;
    }
    {
        float _r_d0 = * _d_x;
        float _r0 = _r_d0 * _t0;
        * _d_x += _r0;
        float _r1 = _t1 * _r_d0;
        * _d_x += _r1;
        * _d_x -= _r_d0;
        * _d_x;
    }                                         // Block of interest ends here.
}
```
2. By using the dummy return:

*Transformed Functions:*
```cpp
// Here we make sure the "dummy" return type is the same as the target function return type.
float helperR(float& x, float& y){
    x = x*x;
    y += x;
    // Here we return 0, or zero equivalent of the return type.
    // Or just return nothing! that works too.
    return 0;
}

float randomR(float x, float y){
    // dummy temp value to make sure clad differentiates this right now...
    float temp = helperR(x, y);
    return x + y;
}
```

*Derivative produced:*

```cpp
void helperR_grad(float &x, float &y, clad::array_ref<float> _d_x, clad::array_ref<float> _d_y) {
    float _t0;
    float _t1;
    _t1 = x;
    _t0 = x;
    x = _t1 * _t0;
    y += x;
    int helperR_return = 0;
    goto _label0;
  _label0:
    ;
    {                                    // See how this block is completely identical to the inlined one we saw before?
        float _r_d1 = * _d_y;
        * _d_y += _r_d1;
        * _d_x += _r_d1;
        * _d_y -= _r_d1;
        * _d_y;
    }
    {
        float _r_d0 = * _d_x;
        float _r0 = _r_d0 * _t0;
        * _d_x += _r0;
        float _r1 = _t1 * _r_d0;
        * _d_x += _r1;
        * _d_x -= _r_d0;
        * _d_x;
    } 
}
void randomR_grad(float x, float y, clad::array_ref<float> _d_x, clad::array_ref<float> _d_y) {
    float _t0;
    float _t1;
    float _d_temp = 0;
    _t0 = x;
    _t1 = y;
    float temp = helperR(_t0, _t1);
    float randomR_return = x + y;
    goto _label0;
  _label0:
    {
        * _d_x += 1;
        * _d_y += 1;
    }
    {
        // float _grad0 = 0.F; -- No need
        // float _grad1 = 0.F; -- No need
        // here we need to make copies x and y again because these temps have been changed before.
        helperR_grad(_t0, _t1, _d_x, _d_y);
        float _r0 = _d_temp * _grad0; //d_temp unused so = 0; noop
        * _d_x += _r0; // Addition with 0; noop
        float _r1 = _d_temp * _grad1; // also 0; noop
        * _d_y += _r1; // also addition with 0; noop
    }
}
```

Now, you can probably see that the above two cases are absolutely identical. I suppose now it is just a choice of how we want to implement this.

Again, I have no exhaustively thought of this or tested it out so we probably need some more debate here...but that is my two cents!

**P.S. Excuse any typos!**

_Originally posted by @grimmmyshini in https://github.com/vgvassilev/clad/pull/247#discussion_r684677355_

### LINKED ISSUE #386: Program crash when calling a function with single array argument
Reproducible example:
```cpp
double modify(double* arr) {
    arr[0] = 5*arr[0] + arr[1];
    return 1;
}

double fn(double* arr) {
    double temp = modify(arr);
    return arr[0];
}

```

Running this code with Clad results in a compile-time crash.
<details>
<summary> Stack Trace </summary>

```
Stack dump:
0.	Program arguments: /usr/lib/llvm-10/bin/clang -cc1 -triple x86_64-pc-linux-gnu -emit-obj -mrelax-all -disable-free -disable-llvm-verifier -discard-value-names -main-file-name bug-rev-fn-call-array.cpp -mrelocation-model static -mthread-model posix -mframe-pointer=all -fmath-errno -fno-rounding-math -masm-verbose -mconstructor-aliases -munwind-tables -target-cpu x86-64 -dwarf-column-info -fno-split-dwarf-inlining -debug-info-kind=limited -dwarf-version=4 -debugger-tuning=gdb -resource-dir /usr/lib/llvm-10/lib/clang/10.0.0 -I /home/parth/clubs/open-source/clad/pristine/builds/build-10/inst/include -internal-isystem /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9 -internal-isystem /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/x86_64-linux-gnu/c++/9 -internal-isystem /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/x86_64-linux-gnu/c++/9 -internal-isystem /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/backward -internal-isystem /usr/local/include -internal-isystem /usr/lib/llvm-10/lib/clang/10.0.0/include -internal-externc-isystem /usr/include/x86_64-linux-gnu -internal-externc-isystem /include -internal-externc-isystem /usr/include -std=c++11 -fdeprecated-macro -fdebug-compilation-dir /home/parth/clubs/open-source/clad/tutorial -ferror-limit 19 -fmessage-length 0 -fgnuc-version=4.2.1 -fobjc-runtime=gcc -fcxx-exceptions -fexceptions -fdiagnostics-show-option -add-plugin clad -load /home/parth/clubs/open-source/clad/pristine/builds/build-10/inst/lib/clad.so -plugin-arg-clad -fdump-derived-fn -faddrsig -o /tmp/bug-rev-fn-call-array-e55a8c.o -x c++ bug-rev-fn-call-array.cpp 
#0 0x00007f06970be4ff llvm::sys::PrintStackTrace(llvm::raw_ostream&) (/lib/x86_64-linux-gnu/libLLVM-10.so.1+0x9814ff)
#1 0x00007f06970bc7b0 llvm::sys::RunSignalHandlers() (/lib/x86_64-linux-gnu/libLLVM-10.so.1+0x97f7b0)
#2 0x00007f06970beac5 (/lib/x86_64-linux-gnu/libLLVM-10.so.1+0x981ac5)
#3 0x00007f069d88e3c0 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x153c0)
#4 0x00007f069b6e94f1 clang::ASTContext::getTypeInfoImpl(clang::Type const*) const (/lib/x86_64-linux-gnu/libclang-cpp.so.10+0x9184f1)
#5 0x00007ffcf00c8960 
clang: error: unable to execute command: Segmentation fault (core dumped)
clang: error: clang frontend command failed due to signal (use -v to see invocation)
clang version 10.0.0-4ubuntu1 
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm-10/bin
clang: note: diagnostic msg: PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace, preprocessed source, and associated run script.
clang: note: diagnostic msg: 
********************

PLEASE ATTACH THE FOLLOWING FILES TO THE BUG REPORT:
Preprocessed source(s) and associated run script(s) are located at:
clang: note: diagnostic msg: /tmp/bug-rev-fn-call-array-847715.cpp
clang: note: diagnostic msg: /tmp/bug-rev-fn-call-array-847715.sh
clang: note: diagnostic msg: 

********************
/usr/lib/llvm-10/bin/clang++ -g -Xclang -add-plugin -Xclang clad -Xclang -load -Xclang /home/parth/clubs/open-source/clad/pristine/builds/build-10/inst/lib/clad.so -Xclang -plugin-arg-clad -Xclang -fdump-derived-fn -I/home/parth/clubs/open-source/clad/pristine/builds/build-10/inst/include -x c++ -std=c++11 bug-rev-fn-call-array.cpp -o build/bug-rev-fn-call-array && build/bug-rev-fn-call-array
```
</details>

The root cause of the error is that we are trying to differentiate the function call using the forward mode AD since there is only one argument. We cannot use the forward mode AD on array arguments because array has multiple elements, thus we should not try to use forward mode AD for single array argument functions.

### LINKED ISSUE #387: Incorrect differentiation of function calls containing array arguments
Reproducible example: 

```cpp
double modify(double* arr, int n) {
    arr[0] += 5*arr[n-1];   
    return 1; 
}

double fn(double* arr, int n) {
    double temp = modify(arr, n);
    return arr[0];
}

int main() {
    double arr[2] = {1, 2};
    double d_arr[2] = {};
    clad::array_ref<double> d_arr_ref(d_arr, 2);
    double d_n = 0;
    fn_grad.execute(arr, 2, d_arr_ref, &d_n);

    for (int i=0; i<2; ++i)
        std::cout<<d_arr[i]<<" ";
}
```
**Expected output:** 1 5
**Actual output:** 1 0

We also do not get any warning/error for unsupported differentiation operations.

parth-07:
This PR adds initial support for pullback functions. The pullback function approach allows to "continue" the reverse mode automatic derivation when required. This allows correctly computing derivatives when arguments are passed by reference or pointers.

This PR also modifies custom gradient functions to custom pullback functions. 

Closes #281, Closes #386, Closes #387


lib/Differentiator/ErrorEstimator.cpp
```diff
@@ -149,6 +149,11 @@ void ErrorEstimationHandler::EmitNestedFunctionParamError(
   for (size_t i = 0; i < numArgs; i++) {
     if (!fnDecl->getParamDecl(0)->getType()->isLValueReferenceType())
       continue;
+    // FIXME: Argument passed by reference do not have any corresponding
+    // `ArgResultDecl`. Handle arguments passed by reference in error
+    // estimation.
+    if (utils::IsReferenceOrPointerType(fnDecl->getParamDecl(i)->getType()))
```
parth-07, line 155: @grimmmyshini Can you please see how we should handle arguments passed by reference for error estimation? 
Arguments passed by reference are handled using the pullback function approach. To see the difference between the earlier and the new approach (pullback functions), please consider the following code snippets: 

Functions that will be used for examples:
```cpp
double someFn(double i, double j);
double update(double& i, double j);
```
**Old approach:** 
```cpp
// original code
double res = someFn(i, j);

// forward pass
_t0 = i;
_t1 = j;
double res = someFn(_t0, _t1);

// reverse pass
double _grad0 = 0;
double _grad1 = 0;
someFn_grad(_t0, _t1, &_grad0, &_grad1);
double _r0 = _grad0 * _d_res;
_d_i += _r0;
double _r1 = _grad1 * _d_res;
_d_j += _r1;
```

**New approach:**
```cpp
double _grad0 = 0;
double _grad1 = 0;
someFn_pullback(_t0, _t1, _d_res, &_grad0, &_grad1);
double _r0 = _grad0;
_d_i += _r0;
double _r1 = _grad1;
_d_j += _r1;
```

Here, notice that we did not need to explicitly multiply `_d_res` with `_grad` variables.

We are passing arg result (`_r0` and `_r1` here) variable declarations for error estimation. For non-reference arguments, arg result variable will be same for both the old approach and the new pullback function approach. 

**New approach with reference arguments involved:**
```cpp
// original code
double res = update(i, j);  // i and j are passed by reference

// forward pass
_t0 = i;
_t1 = j;
double res = update(i, j);

// reverse pass
update_pullback(_t0, _t1, _d_res, &_d_i, &_d_j);  // We are directly passing derivatives to the pullback function
```

Here, please note that there are no arg result variables for arguments passed by reference because they need to be modified directly inside the callee function. What should we use in-place of arg results for reference argument cases?

test/Hessian/BuiltinDerivatives.C
```diff
@@ -309,28 +309,28 @@ float f5(float x) {
...
 // CHECK: void f5_darg0_grad(float x, clad::array_ref<float> _d_x) {
-// CHECK-NEXT:   float _d__d_x = 0;
-// CHECK-NEXT:   float _t0;
-// CHECK-NEXT:   float _t1;
-// CHECK-NEXT:   float _d_x0 = 1;
-// CHECK-NEXT:   _t0 = x;
-// CHECK-NEXT:   _t1 = _d_x0;
-// CHECK-NEXT:   decltype(::std::pow(float(), float())) f5_darg0_return = clad::custom_derivatives{{(::std)?}}::pow_pushforward(2.F, _t0, 0.F, _t1);
-// CHECK-NEXT:   goto _label0;
+// CHECK-NEXT:     float _d__d_x = 0;
```
vgvassilev, line 312: Can you undo the spacing fixes in the tests?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1159,51 +1311,130 @@ namespace clad {
...
+      // a[clad::push(_t0, i)];
+      // a[clad::pop(_t0)];
+      // ```
+      // The expression `a[clad::pop(_t0)]` might already be used in the AST if
+      // visit was called with a dfdx() present.
+      // And thus using this expression in the AST explicitly may lead to size
+      // assertion failed.
+      //
+      // We should modify the design so that the behaviour of obtained StmtDiff
+      // expression is consistent both inside and outside loops.
```
vgvassilev, line 1365: Can you open an issue for this?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1159,51 +1311,130 @@ namespace clad {
...
-    // If the function has a single arg, we look for a derivative w.r.t. to
-    // this arg (it is unlikely that we need gradient of a one-dimensional'
+    // If the function has a single arg and does not returns a reference or take
+    // arg by reference, we look for a derivative w.r.t. to this arg using the
+    // forward mode(it is unlikely that we need gradient of a one-dimensional'
     // function).
     bool asGrad = true;
-    if (NArgs == 1) {
+
+    if (NArgs == 1 && !utils::HasAnyReferenceOrPointerArgument(FD)) {
```
vgvassilev, line 1437: Why this is hardcoded to 1? Worth a comment?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1234,66 +1473,129 @@ namespace clad {
...
+      for (auto argDerivative : CallArgDx) {
+        gradVarDecl = nullptr;
+        gradVarExpr = nullptr;
+        gradArgExpr = nullptr;
+        gradVarII = CreateUniqueIdentifier(funcPostfix());
+
+        auto PVD = FD->getParamDecl(idx);
+        bool passByRef = utils::IsReferenceOrPointerType(PVD->getType());
+        if (passByRef) {
+          // FIXME: Add support for array/pointer derivative types.
```
vgvassilev, line 1490: Should we track that as an issue?

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: Yes, I will open an issue for this.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: We have a comment regarding this just above the `bool asGrad = true;` line.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: Currently, we do not have good support for differentiating pointer types. This is the reason why it couldn't be handled here. Regarding array types, we ideally, should not have array derivative types, derivatives of array types are of `clad::array` type.
We should open issues for tracking differentiation support of pointers.

test/Hessian/BuiltinDerivatives.C
parth-07: That would be difficult to do in all the tests. Is there some efficient way of doing this?

parth-07: I have fixed the tests.

lib/Differentiator/ErrorEstimator.cpp
grimmmyshini: Okay so, I do not think this changes anything to be fair, except probably having to save the parameters before the reference arg function call. I think the rest of the code-gen from error estimation remains the same. Ideally, we should be able to generate error code for the reference args as well, but that requires significant changes (this would involve passing each argument's error to the next function). But as for how things are now, the only thing needed is to save the parameters before the function call in the forward function block.

parth-07: Yes, error estimation seems to be working fine when only non-reference parameters are involved. Since arg result variables (`_r0`,`_r1` etc) remains the same for both the new and the old approach. I am concerned about the handling of reference arguments. 

> but that requires significant changes (this would involve passing each argument's error to the next function)

Should I add an `assert` that fails if we try to estimate the error of a function that contains a function call having arguments passed by reference? This would also involve modifying/removing a test of error estimation that passes an argument by reference (but do not modify it). I have modified this test for now so that it passes (You can see the current changes made in the test by seeing changes made in `test/ErrorEstimation/BasicOps.C` in this PR)

grimmmyshini: I think adding a warning would be better maybe instead of an assert? because the codegen should remain the same regardless of the reference inputs given we assume those are constant in that function call. So maybe a warning like "All reference types are assumed to be constant in error estimation" or something like that?

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: Yeah, please do.

lib/Differentiator/ErrorEstimator.cpp
grimmmyshini: They should not be assumed to be constant, yes. But more often than not you'd actually see constant references being used so it's not a big deal if we ignore them. Especially in this case because doing anything without introducing some form of pass through the called function would just be incorrect.

parth-07: > But more often than not you'd actually see constant references being used so it's not a big deal if we ignore them. 
Okay, sounds right. 

> Especially in this case because doing anything without introducing some form of pass through the called function would just be incorrect.
Can you please explain this more?

grimmmyshini: Okay, let's assume 2 cases:

1.  Function parameters are not changed in the function

In this case, everything is good. We do not have to do anything extra to make this work. Ideally/theoretically it might even make sense to assign the final erro from the called function to the variable that stores it's result in the original code but i haven't tested this out to say much about it. 

2. Function parameters are changed

In this case, we'd probably want to pass the error estimates of the variables so far to the function (i.e. create more parameters in the resulting grad function) so that the called function can add the errors onto the already existing _delta_* values. As you can tell this requires a bit of changes.

parth-07: Oh, okay. Thank you so much for the detailed explanation. I understand the situation much better now.

parth-07: Now error estimation work as expected if arguments passed by reference are not modified. I have also added a `FIXME` comment for issuing a warning if arguments are being passed by reference during error estimation.

parth-07:
> LGTM! Can you squash and put the PR description as a commit message?

Done :).

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: I will open issues for tracking pointer progress soon.

PR #389, opened 2022-02-28
Add initial support for diff class types in the forw mode

parth-07:
This PR adds initial support for differentiating class types with respect to scalar numerical types using the forward mode AD.

Given a function:
Screenshot from 2022-02-28 20-10-43

The forward mode automatic differentiation allows us to compute derivative of _all_ the output variables with respect to a single input variable (independent variable) in a single pass. Mathematically, the forward mode AD allows us to efficiently compute columns of the jacobian matrix. Up till now, we are not utilising this ability of the forward mode AD -- since we only supported differentiating scalar type with respect to a scalar type. 
Thus, differentiation of aggregate types in the forward mode AD will allow us to utilize this ability and obtain multiple derivatives in a single call to the derived function. 

Clad requires the following assumptions to be true when differentiating a class type:
- Class should represent a real vector space. 
- Class should have a default constructor. Class variables initialised by the default constructor should represent 0 tangent vector -- that is, all real data members should be equal to 0. 

Differentiation of class types in the forward mode AD currently have the following limitations:
- Calls to member functions are not supported.
- Calls to overloaded operators are not supported.
- Initialising class type variables using non-default constructor are not supported.
- Class cannot have pointer or reference data members.

Most of these limitations will be removed soon.

### DIFF
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -177,5 +177,35 @@ namespace clad {
     bool SameCanonicalType(clang::QualType T1, clang::QualType T2) {
       return T1.getCanonicalType() == T2.getCanonicalType();
     }
+    MemberExpr* BuildMemberExpr(Sema& semaRef, Expr* base, ValueDecl* member) {
+      auto& C = semaRef.getASTContext();
+      auto DAP = DeclAccessPair::make(member, member->getAccess());
+      CXXScopeSpec CSS;
+      DeclarationName DN = member->getDeclName();
+      DeclarationNameInfo DNI(DN, noLoc);
+      bool isArrow = base->getType()->isPointerType();
+      // Expressions should never have a reference type.
+      return clad_compat::BuildMemberExpr(
+          semaRef, base, isArrow, noLoc, &CSS, noLoc, member, DAP, false, DNI,
+          member->getType().getNonReferenceType(), ExprValueKind::VK_LValue,
+          ExprObjectKind::OK_Ordinary);
+    }
+
+    bool isDifferentiableType(QualType T) {
+      // FIXME: Handle arbitrary depth pointer types and arbitrary dimension
+      // array type as well.
+      if (isArrayOrPointerType(T))
+        T = T->getPointeeOrArrayElementType()->getCanonicalTypeInternal();
+      T = T.getNonReferenceType();
+      if (T->isRealType() || T->isRecordType())
+        return true;
+      return false;
+    }
+
+    SourceLocation GetValidSLoc(Sema& semaRef) {
+      auto& SM = semaRef.getSourceManager();
+      return SM.getLocForStartOfFile(SM.getMainFileID());
+    }
+
   } // namespace utils
 } // namespace clad
--- a/lib/Differentiator/VisitorBase.cpp
+++ b/lib/Differentiator/VisitorBase.cpp
@@ -71,7 +71,10 @@ namespace clad {
     if (Init) {
       m_Sema.AddInitializerToDecl(VD, Init, DirectInit);
       VD->setInitStyle(IS);
+    } else {
+      m_Sema.ActOnUninitializedDecl(VD);
     }
+    m_Sema.FinalizeDeclaration(VD);
     // Add the identifier to the scope and IdResolver
     m_Sema.PushOnScopeChains(VD, getCurrentScope(), /*AddToContext*/ false);
     return VD;
@@ -609,7 +612,9 @@ namespace clad {
     // build the clad::tape<clad::array_ref>> = {};
     QualType RefType = GetCladArrayRefOfType(retType);
     QualType TapeType = GetCladTapeOfType(RefType);
-    auto VD = BuildVarDecl(TapeType);
+    auto VD = BuildVarDecl(
+        TapeType, "_t", getZeroInit(TapeType), /*DirectInit=*/false,
+        /*TSI=*/nullptr, VarDecl::InitializationStyle::CInit);
     NumericalDiffMultiArg.push_back(BuildDeclStmt(VD));
     Expr* TapeRef = BuildDeclRef(VD);
     NumDiffArgs.push_back(TapeRef);
--- a/test/FirstDerivative/CallArguments.C
+++ b/test/FirstDerivative/CallArguments.C
@@ -65,7 +65,7 @@ float f_const_args_func_3(const float x, float y) {
 // CHECK-NEXT: return _d_x * x + x * _d_x + _d_y * y + y * _d_y - 0.F;
 // CHECK-NEXT: }
 
-struct Vec { float x,y,z; };
+struct Vec { float x=0, y=0, z=0; };
 float f_const_args_func_4(float x, float y, const Vec v) {
   return x * x + y * y - v.x;
 }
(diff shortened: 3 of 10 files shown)


lib/Differentiator/VisitorBase.cpp
```diff
@@ -69,9 +69,15 @@ namespace clad {
                         SC_None);
 
     if (Init) {
+      // llvm::errs()<<"Dumping INIT:\n";
+      // Init->dumpColor();
+      // llvm::errs()<<"\n";
```
vgvassilev, line 74: This seems leftover?

test/FirstDerivative/CallArguments.C
```diff
@@ -65,35 +65,38 @@ float f_const_args_func_3(const float x, float y) {
...
+// CHECK: float f_const_args_func_5_darg0(float x, float y, Vec v) {
 // CHECK-NEXT: float _d_x = 1;
 // CHECK-NEXT: float _d_y = 0;
-// CHECK-NEXT: return _d_x * x + x * _d_x + _d_y * y + y * _d_y - 0.F;
+// CHECK-NEXT:     Vec _d_v;
+// CHECK-NEXT:     return _d_x * x + x * _d_x + _d_y * y + y * _d_y - _d_v.x;
 // CHECK-NEXT: }
 
-float f_const_args_func_6(const float x, const float y, const Vec &v) {
+float f_const_args_func_6(const float x, const float y, Vec v) {
```
vgvassilev, line 91: Why we needed this change?

test/ErrorEstimation/LoopsAndArraysExec.C
```diff
@@ -146,11 +146,11 @@ double divSum(float* a, float* b, int n) {
 //CHECK-NEXT:     double _EERepl_sum0;
 //CHECK-NEXT:     unsigned long _t0;
 //CHECK-NEXT:     int _d_i = 0;
-//CHECK-NEXT:     clad::tape<float> _t1 = {};
-//CHECK-NEXT:     clad::tape<int> _t2 = {};
-//CHECK-NEXT:     clad::tape<float> _t4 = {};
-//CHECK-NEXT:     clad::tape<int> _t5 = {};
-//CHECK-NEXT:     clad::tape<double> _EERepl_sum1 = {};
+//CHECK-NEXT:     clad::tape<float> _t1({});
```
vgvassilev, line 149: Why do we need these changes? I can imagine we can tweak the constructor of the clad::tape to support brace-inits...

vgvassilev (commented):
That's awesome! Thanks for working on this!

test/FirstDerivative/CallArguments.C
parth-07: We don't support reference arguments in the original forward mode. Because for a reference argument, we cannot create a derivative variable such as `float& _d_x = 0`, `const Vec& _d_v`. They are both invalid C++ expressions because references cannot bind to rvalue expressions and reference object cannot be uninitialised. 

Reference arguments are only supported in the pushforward design. Currently, when a user differentiates a function `fn` using `clad::differentiate`, we differentiate the function `fn` using the original forward mode. All the function calls inside the function `fn` are differentiated using the pushforward design. 
This imposes the limitation that the original function arguments cannot be of reference type.

test/ErrorEstimation/LoopsAndArraysExec.C
parth-07: This change is required due to this:
```cpp
if (Init) {
    m_Sema.AddInitializerToDecl(VD, Init, DirectInit);
    VD->setInitStyle(IS);
} else {
    m_Sema.ActOnUninitializedDecl(VD);
}
```

Earlier we did not have the `else` block containing the call to `ActOnUninitializedDecl`. Call to `ActOnUninitializedDecl` is essential due to two reasons:
- To properly emulate clang, we need to call `ActOnUninitializedDecl` on variables of class types that have no explicit initializer. `ActOnUninitializedDecl` creates the `CXXConstructExpr` for a call to the default constructor. 
- This allows us to be more consistent with clang.

But yes, most probably we should be able to tweak the constructor of `clad::tape` to avoid this change. I will try doing this.

lib/Differentiator/VisitorBase.cpp
parth-07: Yes, sorry I forgot to remove this. I have now removed this.

test/ErrorEstimation/LoopsAndArraysExec.C
parth-07: Done. 
Now in reverse mode, we are creating all the temporary variables including clad tape using C style init method. Thus clad tape initialisation invocation will remain the same.

test/NumericalDiff/GradientMultiArg.C
```diff
@@ -23,7 +23,7 @@ double test_1(double x, double y){
 // CHECK-NEXT:     {
 // CHECK-NEXT:         double _grad0 = 0.;
 // CHECK-NEXT:         double _grad1 = 0.;
-// CHECK-NEXT:         clad::tape<clad::array_ref<double> > _t2 = {};
+// CHECK-NEXT:         clad::tape<clad::array_ref<double> > _t2({});
```
vgvassilev, line 26: Why we still need this?

test/ForwardMode/UserDefinedTypes.C
```diff
@@ -0,0 +1,88 @@
+// RUN: %cladclang %s -I%S/../../include -oUserDefinedTypes.out 2>&1 | FileCheck %s
+// RUN: ./UserDefinedTypes.out | FileCheck -check-prefix=CHECK-EXEC %s
+
```
vgvassilev, line 3:
```suggestion

//CHECK-NOT: {{.*error|warning|note:.*}}

```

lib/Differentiator/CladUtils.cpp
```diff
@@ -153,5 +153,38 @@ namespace clad {
...
+
+    MemberExpr* BuildMemberExpr(Sema& semaRef, Expr* base, ValueDecl* member) {
+      auto& C = semaRef.getASTContext();
+      auto DAP = DeclAccessPair::make(member, member->getAccess());
+      CXXScopeSpec CSS;
+      DeclarationName DN = member->getDeclName();
+      DeclarationNameInfo DNI(DN, noLoc);
+      bool isArrow = false;
+      if (base->getType()->isPointerType())
+        isArrow = true;
```
vgvassilev, line 165:
```suggestion
      bool isArrow = base->getType()->isPointerType();
```

lib/Differentiator/CladUtils.cpp
```diff
@@ -153,5 +153,38 @@ namespace clad {
...
+      // array type as well.
+      if (isArrayOrPointerType(T))
+        T = T->getPointeeOrArrayElementType()->getCanonicalTypeInternal();
+      T = T.getNonReferenceType();
+      if (T->isRealType() || T->isRecordType())
+        return true;
+      return false;
+    }
+
+    SourceLocation GetValidSL(Sema& semaRef) {
```
vgvassilev, line 184:
```suggestion
    SourceLocation GetValidSLoc(Sema& semaRef) {
```

test/NumericalDiff/GradientMultiArg.C
parth-07: This must be being created at a different place.

parth-07: Done. I have fixed this as well.

test/FirstDerivative/CallArguments.C
parth-07: We can remove the reference from the derivative types. For example, if the original parameter type is `float&`, then the derivative type can be just `float`. Should we handle this in a different PR?

vgvassilev:
Can you rebase?

parth-07:
> Can you rebase?

Done.

vgvassilev (approved):
LGTM!

PR #391, opened 2022-03-02
Create derivatives of reference arguments in the forward mode.

parth-07:
This PR modifies the forward mode such that now we also consider reference real type arguments as differentiable and thus creates derivatives for them of the corresponding non-reference types.

### DIFF
--- a/lib/Differentiator/ForwardModeVisitor.cpp
+++ b/lib/Differentiator/ForwardModeVisitor.cpp
@@ -185,7 +185,9 @@ namespace clad {
       }
       m_IndependentVarIndex = indexIntervalTable[0].Start;
       derivativeSuffix = "_" + std::to_string(m_IndependentVarIndex);
-    } else if (!m_IndependentVar->getType()->isRealType()) {
+    } else if (!m_IndependentVar->getType()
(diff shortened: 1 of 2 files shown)


lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -185,7 +185,9 @@ namespace clad {
       }
       m_IndependentVarIndex = indexIntervalTable[0].Start;
       derivativeSuffix = "_" + std::to_string(m_IndependentVarIndex);
-    } else if (!m_IndependentVar->getType()->isRealType()) {
+    } else if (!m_IndependentVar->getType()
+                    .getNonReferenceType()
+                    ->isRealType()) {
```
vgvassilev, line 190: Maybe extract that as a file static function called `IsRealNonReferenceType`?

parth-07: That will be a very short-lived function. In the upcoming PR for adding support for class types, we will be using `util::IsDifferentiableType` here. Should I create that function just for this pull request?

vgvassilev (approved):
LGTM!

PR #392, opened 2022-03-06
Add initial support for differentiating user-defined types in the reverse mode.

parth-07:
This PR adds initial support for differentiating a scalar built-in numerical type with respect to user-defined types using the reverse-mode AD.

The reverse mode automatic differentiation allows to efficiently compute the derivative of a single output variable with respect to all the input variables. Mathematically, the reverse mode AD allows to efficiently compute rows of the Jacobian matrix. 

Clad requires the following assumptions to be true while differentiating with respect to class types in the reverse mode AD:

- Class should represent a real vector space.
- Class should have a default constructor that zero initialises all the data members of the class. That is, a newly default constructed class object should represent 0 tangent vector. 

Differentiation with respect to class types in the reverse mode AD currently has the following limitations:

- Calls to member functions are not supported
- Calls to overloaded operators are not supported
- Initialising class type variables using non-default constructors that take non-literal arguments (non-zero derivatives) are not supported.
- Class cannot have pointer or reference data members.

Most of these limitations will be removed soon.

## Overloaded derived function

This PR also modifies the overloaded derivative function mechanism as described in the issue #394 . Two most important changes are as follows:

- Overloaded derived function is now created for all the derived functions that are generated using the reverse mode AD.
  Earlier, we were not creating the overloaded derived function if gradient was requested with respect to all the original
  function parameters.

- Type of all the parameters derivatives is: `clad::array_ref<void>`. It gets converted to (by `static_cast`) 
  to correct type before being passed as an argument to the actual derived function call.

## Test Utilities

Apart from the class type support this PR also adds test utilities for writing gradient tests. 

Gradient test utility interface is as follows: 

Two overloads are provided for `INIT_GRADIENT` Macro -- initialise gradient function using `clad::gradient`
```
INIT_GRADIENT(fn);  // Initialises derived function that computes gradient with respect to all function parameters
INIT_GRADIENT(fn, args);  // Initialises derived function that computes gradient with respect to all elements described by 'args'
```

To perform a test on the initialised gradient functions, `TEST_GRADIENT` Macro is provided.

```cpp
std::pair<double, long double> p(3, 7), d_p;
Tangent t, d_t;
double d_i, d_j;
// Reset (to zero) all the derivative args, executes gradient function and then prints 
// all the derivatives separated by commas and enclosed within curly braces
TEST_GRADIENT(fn1, /*NumOfDerivativeArgs=*/4, 1.00, 2, p, t, &d_i, &d_j, &d_p, &d_t);
```

References #394 , #395

### DIFF
--- a/include/clad/Differentiator/ArrayRef.h
+++ b/include/clad/Differentiator/ArrayRef.h
@@ -40,7 +40,8 @@ namespace clad {
       return *this;
     }
     /// Returns the size of the underlying array
-    CUDA_HOST_DEVICE std::size_t size() { return m_size; }
+    CUDA_HOST_DEVICE std::size_t size() const { return m_size; }
+    CUDA_HOST_DEVICE T* ptr() const { return m_arr; }
     /// Returns an array_ref to a part of the underlying array starting at
     /// offset and having the specified size
     CUDA_HOST_DEVICE array_ref<T> slice(std::size_t offset, std::size_t size) {
@@ -88,6 +89,36 @@ namespace clad {
         m_arr[i] -= Ar[i];
       return *this;
     }
+  };
+
+  /// `array_ref<void>` specialisation is created to be used as a placeholder
+  /// type in the overloaded derived function. All `array_ref<T>` types are
+  /// implicitly convertible to `array_ref<void>` type.
+  ///
+  /// `array_ref<void>` variables should be converted to the correct
+  /// `array_ref<T>` type before being used. To make this process easier and
+  /// more convenient, `array_ref<void>` provides implicit converter operators
+  /// that facilitates convertion to `array_ref<T>` type using `static_cast`.
+  template <> class array_ref<void> {
+  private:
+    /// The pointer to the underlying array
+    void* m_arr = nullptr;
+    /// The size of the array
+    std::size_t m_size = 0;
+
+  public:
+    // delete the default constructor
+    array_ref() = delete;
+    CUDA_HOST_DEVICE array_ref(void* arr, std::size_t size = 1)
+        : m_arr(arr), m_size(size) {}
+    template <typename T>
+    CUDA_HOST_DEVICE array_ref(const array_ref<T>& other)
+        : m_arr(other.ptr()), m_size(other.size()) {}
+    template <typename T> CUDA_HOST_DEVICE operator array_ref<T>() {
+      return array_ref<T>(static_cast<T*>(m_arr), m_size);
+    }
+    CUDA_HOST_DEVICE void* ptr() const { return m_arr; }
+    CUDA_HOST_DEVICE std::size_t size() const { return m_size; }
   };
 } // namespace clad
 
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -101,6 +101,28 @@ namespace clad {
       return result;
     }
 
+    /// This visit method explicitly sets `dfdx` to `nullptr` for this visit.
+    ///
+    /// This method is helpful when we need derivative of some expression but we
+    /// do not want `_d_expression += dfdx` statments to be (automatically)
+    /// added.
+    ///
+    /// FIXME: Think of a better way for handling this situation. Maybe we
+    /// should improve the overall dfdx design and approach. One other way of
+    /// designing `VisitWithExplicitNoDfDx` in a more general way is
+    /// to develop a function that takes an expression E and returns the
+    /// corresponding derivative without any side effects. The difference
+    /// between this function and the current `VisitWithExplicitNoDfDx` will be
+    /// 1) better intent through the function name 2) We will also get
+    /// derivatives of expressions other than `DeclRefExpr` and `MemberExpr`.
+    StmtDiff VisitWithExplicitNoDfDx(const clang::Stmt* stmt) {
+      m_Stack.push(nullptr);
+      auto result =
+          clang::ConstStmtVisitor<ReverseModeVisitor, StmtDiff>::Visit(stmt);
+      m_Stack.pop();
+      return result;
+    }
+
     /// Get the latest block of code (i.e. place for statements output).
     Stmts& getCurrentBlock(direction d = direction::forward) {
       if (d == direction::forward)
@@ -312,6 +334,10 @@ namespace clad {
     StmtDiff VisitDoStmt(const clang::DoStmt* DS);
     StmtDiff VisitContinueStmt(const clang::ContinueStmt* CS);
     StmtDiff VisitBreakStmt(const clang::BreakStmt* BS);
+    StmtDiff VisitCXXThisExpr(const clang::CXXThisExpr* CTE);
+    StmtDiff VisitCXXConstructExpr(const clang::CXXConstructExpr* CE);
+    StmtDiff
+    VisitMaterializeTemporaryExpr(const clang::MaterializeTemporaryExpr* MTE);
     VarDeclDiff DifferentiateVarDecl(const clang::VarDecl* VD);
 
     /// A helper method to differentiate a single Stmt in the reverse mode.
@@ -343,11 +369,18 @@ namespace clad {
     /// Builds an overload for the gradient function that has derived params for
     /// all the arguments of the requested function and it calls the original
     /// gradient function internally
-    clang::FunctionDecl* CreateGradientOverload(
-        llvm::SmallVectorImpl<clang::QualType>& GradientParamTypes,
-        llvm::SmallVectorImpl<clang::ParmVarDecl*>& GradientParams,
-        clang::DeclarationNameInfo& GradientName,
-        clang::FunctionDecl* GradientFD);
+    clang::FunctionDecl* CreateGradientOverload();
+
+    /// Returns the type that should be used to represent the derivative of a
+    /// variable of type `yType` with respect to a parameter variable of type
+    /// `xType`.
+    ///
+    /// FIXME: Parameter derivative type rules are different from the derivative
+    /// type rules for local variables. We should remove this inconsistency.
+    /// See the following issue for more details:
+    /// https://github.com/vgvassilev/clad/issues/385
+    clang::QualType GetParameterDerivativeType(clang::QualType yType,
+                                               clang::QualType xType);
 
     /// Allows to easily create and manage a counter for counting the number of
     /// executed iterations of a loop. 
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -1,5 +1,7 @@
 #include "clad/Differentiator/CladUtils.h"
 
+#include "ConstantFolder.h"
+
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/Decl.h"
 #include "clang/AST/Expr.h"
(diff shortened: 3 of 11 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2181,9 +2217,36 @@ namespace clad {
...
+      // `VDDerivedType` is the corresponding non-reference type and the initial
+      // value is set to 0.
+      // Otherwise, for non-reference types, the initial value is set to 0.
       VDDerivedInit = getZeroInit(VD->getType());
-      VDDerived = BuildVarDecl(VDDerivedType, "_d_" + VD->getNameAsString(),
-                               VDDerivedInit);
+      if (isVDRefType) {
+        VDDerivedType = getNonConstType(VDDerivedType, m_Context, m_Sema);
+        initDiff = Visit(VD->getInit());
+        if (initDiff.getExpr_dx()) {
```
vgvassilev, line 2232: No braces here and in the else section.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2181,9 +2217,36 @@ namespace clad {
...
+      }
+      // Here separate behaviour for record and non-record types is only
+      // necessary to preserve the old tests.
+      if (VDDerivedType->isRecordType()) {
+        VDDerived =
+            BuildVarDecl(VDDerivedType, "_d_" + VD->getNameAsString(),
+                         VDDerivedInit, VD->isDirectInit(),
+                         m_Context.getTrivialTypeSourceInfo(VDDerivedType),
+                         VD->getInitStyle());
+      } else {
```
vgvassilev, line 2246: Likewise.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2214,10 +2277,17 @@ namespace clad {
...
-                                    initDiff.getExpr(),
-                                    VD->isDirectInit());
+    VarDecl* VDClone = nullptr;
+    // Here separate behaviour for record and non-record types is only
+    // necessary to preserve the old tests.
+    if (VD->getType()->isRecordType()) {
+      VDClone = BuildVarDecl(VD->getType(), VD->getNameAsString(),
+                             initDiff.getExpr(), VD->isDirectInit(),
+                             VD->getTypeSourceInfo(), VD->getInitStyle());
+    } else {
```
vgvassilev, line 2287: Likewise.

lib/Differentiator/VisitorBase.cpp
```diff
@@ -461,15 +461,16 @@ namespace clad {
...
     auto RDType = RD->getTypeForDecl();
     auto thisObjectQType = C.getQualifiedType(
         RDType, clad_compat::CXXMethodDecl_getMethodQualifiers(MD));        
-    if (MD->getRefQualifier() == RefQualifierKind::RQ_LValue)
-      thisObjectQType = C.getLValueReferenceType(thisObjectQType);
-    else if (MD->getRefQualifier() == RefQualifierKind::RQ_RValue)
+    if (MD->getRefQualifier() == RefQualifierKind::RQ_RValue)
       thisObjectQType = C.getRValueReferenceType(thisObjectQType);
+    else
+      thisObjectQType = C.getLValueReferenceType(thisObjectQType);
```
vgvassilev, line 473: Can we add a test for this case?

lib/Differentiator/VisitorBase.cpp
```diff
@@ -479,8 +480,15 @@ namespace clad {
     Expr* thisExpr = clad_compat::Sema_BuildCXXThisExpr(m_Sema, FD);
     bool isArrow = true;
 
-    if (useRefQualifiedThisObj) {
-      auto thisQType = getRefQualifiedThisType(m_Context, FD);
+    // C++ do not supports perfect forwarding of `*this` object inside
```
vgvassilev, line 483:
```suggestion
    // C++ does not supports perfect forwarding of `*this` object inside
```

lib/Differentiator/VisitorBase.cpp
```diff
@@ -479,8 +480,15 @@ namespace clad {
...
+    // a member function.
+    // Cast `*this` to rvalue if the current function has rvalue qualifier so
```
vgvassilev, line 485:
```suggestion
    // Cast `*this` to an rvalue if the current function has rvalue qualifier so
```

lib/Differentiator/VisitorBase.cpp
```diff
@@ -479,8 +480,15 @@ namespace clad {
...
+    // that correct method overload is resolved. We do not need to cast to
+    // lvalue because without any cast `*this` will always be considered an
```
vgvassilev, line 487:
```suggestion
    // an lvalue because without any cast `*this` will always be considered an
```

test/TestUtils.h
```diff
@@ -0,0 +1,114 @@
+#ifndef TEST_TEST_UTILS_H
+#define TEST_TEST_UTILS_H
+#include "clad/Differentiator/Differentiator.h"
```
vgvassilev, line 3: I find it surprising to need this header file in a testing utility. Can we include the relevant array_ref header if it was the intent?

vgvassilev (commented):
Overall looks pretty good. I added a few minor comments.

lib/Differentiator/CladUtils.cpp
```diff
@@ -1,5 +1,7 @@
 #include "clad/Differentiator/CladUtils.h"
 
+#include "ConstantFolder.h"
```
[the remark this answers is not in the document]
parth-07: Yes. 

Can we also have a different class for AST related functions? For example, an `ASTHelper` class that can be initialised with `clang::Sema` object and the `clang::Scope`. This will allow us to have a separate dedicated location for AST related utility functions. It will also facilitate easier usage of AST related utility functions, as we will not need to pass `clang::Sema` (and in some cases, `clang::Scope` as well) to every function call. 

Over some time period, slowly and steadily we should also be able to migrate AST helper functions present in `Visitor` classes to this `ASTHelper` class. This will reduce the responsibilities from the `Visitor` classes and will allow them to focus more and more only on automatic differentiation. 

If we do decide to go with this approach, then we should also provide 2 alternatives for each function:
- One, a member function, that should be called without passing `clang::Sema` and `clang::Scope` objects.
- Second, a static function, that should be called with explicit `clang::Sema` and if necessary, `clang::Scope` objects. This will be helpful in situations where we only require one or two of the AST utility functions. 

These two functions can be implemented as follows: 
```cpp
static Expr* ASTHelper::BuildCallToMemFn(Sema&, ...) { }
Expr* ASTHelper::BuildCallToMemFn(...) {
  return ASTHelper::BuildCallToMemFn(m_Sema, ...);
}
```

test/TestUtils.h
parth-07: Yes, including `array_ref` was the main intent. I have changed the include header file to `ArrayRef.h`.

lib/Differentiator/VisitorBase.cpp
parth-07: Yes, but I am not sure if we should use this. 

Currently, in the overloaded function we are doing as follows: 

```cpp
void memFn_grad(double i, double j, clad::array_ref<void> _temp_d_i, clad::array_ref<void> _temp_d_j) && {
  clad::array_ref<double> _d_i = _temp_d_i;
  clad::array_ref<double> _d_j = _temp_d_j;
  static_cast<Class&&>(*this).memFn_grad(i, j, _d_i, _d_j);
}
```

If a member function has an lvalue reference qualifier or no reference qualifier, then the `static_cast` does not provide us with any benefits. On the other hand, if we use `static_cast` for a member function call that originally has no reference qualifier, then we may accidentally cast an originally rvalue `this` expression to an `lvalue`. 

I just realized we can safely use `static_cast` when the member function has either an lvalue qualifier or an rvalue qualifier. This will allow us to improve the test coverage.

parth-07: I have modified the code such that now we are using `static_cast` when the original member function has an lvalue qualifier as well. Thus this line should be covered by the tests now.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -532,28 +561,33 @@ namespace clad {
...
 
-    QualType arrayRefValueType = m_Function->getReturnType();
+
+    QualType effectiveReturnType = m_Function->getReturnType();
     // Generally, we use the function's return type as the argument's derivative
     // type. We cannot follow this strategy for `void` function return type.
     // Thus, temporarily use `double` type as the placeholder type for argument
     // derivatives. We should think of a more uniform and consistent solution to
-    // this problem. Check this related issue for more details:
+    // this problem. One effective strategy that may hold well: If we are
```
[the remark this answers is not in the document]
parth-07: Yes! I have now prefixed the message with `FIXME:`.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -101,6 +101,19 @@ namespace clad {
       return result;
     }
 
+    /// This visit method explicitly set `dfdx` to nullptr for this visit.
+    ///
+    /// This method is helpful when we need derivative of some expression but we
+    /// do not want `_d_expression += dfdx` statments to be (automatically)
+    /// added.
+    StmtDiff VisitWithExplicitNoDfDx(const clang::Stmt* stmt) {
```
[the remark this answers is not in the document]
parth-07: I have added a `FIXME` for this. But dropping the entire `dfdx` approach would require immense planning and effort.

include/clad/Differentiator/ArrayRef.h
```diff
@@ -89,6 +92,28 @@ namespace clad {
       return *this;
     }
   };
+
+  template <> class array_ref<void> {
```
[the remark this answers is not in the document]
parth-07: I have added documentation here.

include/clad/Differentiator/ArrayRef.h
```diff
@@ -41,6 +41,9 @@ namespace clad {
     }
     /// Returns the size of the underlying array
     CUDA_HOST_DEVICE std::size_t size() { return m_size; }
+    CUDA_HOST_DEVICE std::size_t size() const { return m_size; }
```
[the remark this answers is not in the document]
parth-07: I just realized that in our case `const` and non-`const` member functions were exactly the same. Thus we could just combine them into one. But yes, I will keep the suggestion provided in the StackOverflow answer in mind for future codes :).

include/clad/Differentiator/ReverseModeVisitor.h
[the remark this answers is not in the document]
vgvassilev: Yep, I know...

lib/Differentiator/CladUtils.cpp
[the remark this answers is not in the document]
vgvassilev: We should cache the objects that do not change like `Sema` however, for `Scope` I prefer to pass it as an argument each time. I think that's mostly syntactic sugar and can be done outside of this PR.

lib/Differentiator/VisitorBase.cpp
vgvassilev: Nice!

vgvassilev (approved):
lgtm!

PR #397, opened 2022-03-13
Add support for differentiating calls to member functions in the reverse mode

parth-07:
This PR adds support for differentiating calls to member functions in the reverse mode AD.

This PR introduces a breaking change in the signature of derived member functions. Earlier we were considering data members to be constant while differentiating member functions. This is no longer the case, and thus now we need a separate variable to store derivative of the function value with respect to implicit `this` pointer. To allow users to obtain derivatives with respect to the implicit `this` object, derived member function prototype has been changed to include a derivative parameter for the implicit `this` pointer.

For example: 

```cpp
struct SomeStruct {
  double memFn(double i, long double j) {...}
};
```

This function will have the gradient function as follows:

```cpp
void memFn(double i, long double j, clad::array_ref<SomeStruct> _d_this, clad::array_ref<double> _d_i, clad::array_ref<double> _d_j) {...}
```

Class type support hasn't been fully-added to the hessian differentiation mode. It is because there is no easy way to represent the hessian matrix when class types are involved. Nonetheless, when differentiating member functions using the hessian differentiation mode, data members are considered differentiable. This is made feasible by creating and passing a dummy `_d_this` variable when calling gradient routine inside the hessian function.

This PR also adds support for C99 language extension of `__real` and `__imag` unary operators in the reverse mode AD. This in turn allows us to perform basic differentiation of `std::complex` types.

### DIFF
--- a/include/clad/Differentiator/CladConfig.h
+++ b/include/clad/Differentiator/CladConfig.h
@@ -31,16 +31,16 @@ void trap(int code) {
 
 #ifdef  __CUDACC__
 template<typename T>
-__device__ T* addressof(T& r) {
+__device__ T* clad_addressof(T& r) {
   return __builtin_addressof(r);
 }
 template<typename T>
-__host__ T* addressof(T& r) {
+__host__ T* clad_addressof(T& r) {
   return std::addressof(r);
 }
 #else
 template<typename T>
-T* addressof(T& r) {
+T* clad_addressof(T& r) {
   return std::addressof(r);
 }
 #endif
--- a/include/clad/Differentiator/CladUtils.h
+++ b/include/clad/Differentiator/CladUtils.h
@@ -148,6 +148,8 @@ namespace clad {
     clang::DeclarationNameInfo BuildDeclarationNameInfo(clang::Sema& S,
                                                         llvm::StringRef name);
 
+    /// Returns true if the function has any reference or pointer parameter;
+    /// otherwise returns false.
     bool HasAnyReferenceOrPointerArgument(const clang::FunctionDecl* FD);
 
     /// Returns true if `T` is a reference, pointer or array type.
@@ -196,8 +198,9 @@ namespace clad {
                      clang::TypeSourceInfo* TSI = nullptr);
 
     /// If `T` represents an array or a pointer type then returns the
-    /// corresponding array element or the pointee type. Otherwise, if `T` is
-    /// neither an array nor a pointer type, then simply returns `T`.
+    /// corresponding array element or the pointee type. If `T` is a reference
+    /// type then return the corresponding non-reference type. Otherwise, if `T`
+    /// is neither an array nor a pointer type, then simply returns `T`.
     clang::QualType GetValueType(clang::QualType T);
 
     /// Builds and returns the init expression to initialise `clad::array` and
@@ -207,6 +210,10 @@ namespace clad {
     /// `{arr, arrSize}`
     clang::Expr* BuildCladArrayInitByConstArray(clang::Sema& semaRef,
                                                 clang::Expr* constArrE);
+
+    /// Returns true if `FD` is a class static method; otherwise returns
+    /// false.
+    bool IsStaticMethod(const clang::FunctionDecl* FD);
   } // namespace utils
 }
 
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -89,9 +89,31 @@ namespace clad {
           DC2 = DC2->getParent();  
           continue;
         }
+        // We don't want to 'extend' the DC1 context with class declarations.
+        // There are 2 main reasons for this:
+        // - Class declaration context cannot be extended the way namespace
+        // declaration contexts can.
+        //
+        // - Primary usage of `FindDeclarationContext` is to create the correct
+        // declaration context for searching some particular custom derivative.
+        // But for class functions, we would 'need' to create custom derivative
+        // in the original declaration context only (We don't support custom
+        // derivatives for class functions yet). We cannot use the default
+        // context that starts from `clad::custom_derivatives::`. This is
+        // because custom derivatives of class functions need to have same
+        // access permissions as the original member function.
+        //
+        // We may need to change this behaviour if in the future
+        // `FindDeclarationContext` function is being used for some place other
+        // than finding custom derivative declaration context as well.
+        //
+        // Silently return nullptr if DC2 contains any CXXRecord declaration
+        // context.
+        if (isa<CXXRecordDecl>(DC2))
+          return nullptr;
         assert(isa<NamespaceDecl>(DC2) &&
-               "DC2 should only contain namespace (and "
-               "translation unit) declaration.");
+               "DC2 should only consists of namespace, CXXRecord and "
+               "translation unit declaration context.");
         contexts.push_back(DC2);
         DC2 = DC2->getParent();
       }
@@ -237,10 +259,14 @@ namespace clad {
 
     clang::QualType GetValueType(clang::QualType T) {
       QualType valueType = T;
-      if (isArrayOrPointerType(T)) {
+      if (T->isPointerType())
+        valueType = T->getPointeeType();
+      else if (T->isReferenceType()) 
+        valueType = T.getNonReferenceType();
+      // FIXME: `QualType::getPointeeOrArrayElementType` loses type qualifiers.
+      else if (T->isArrayType()) 
         valueType =
             T->getPointeeOrArrayElementType()->getCanonicalTypeInternal();
-      }
       return valueType;
     }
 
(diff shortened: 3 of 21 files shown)


lib/Differentiator/VisitorBase.cpp
```diff
@@ -444,19 +444,36 @@ namespace clad {
...
+    // Ideally, if `MemberFunctionName` corresponds to an overloaded member
+    // function name, then `Sema::ActOnMemberAccessExpr` should return an
+    // unresolved expression and `Sema::ActOnCallExpr` should automatically
+    // resolve this unresolved expression on the basis of the call arguments
+    // provided. But currently, since only one of the member function overload
+    // is found by the lookup, unresolved expression is not created and thus, we
+    // explicitly should assign the correct member function whenever we can.
+    if (memberDecl)
+      ME->setMemberDecl(memberDecl);
+    // llvm::errs()<<"ME Dump:\n";
```
vgvassilev, line 474: It seems we have some debug leftovers.

parth-07: Yes, I totally missed them. I have removed them now.

vgvassilev:
> This PR also refactors ReverseModeVisitor::Derive and ReverseModeVisitor::DerivePullback such that the derived function parameter types and the parameters declarations are computed and built in separate subroutines. This facilitates the reuse of the common code between different differentiation modes supported by the reverse mode (DiffMode::reverse, DiffMode::experimental_pullback, and DiffMode::jacobian).

Can you move this in a separate PR? That will help speed up the review.

parth-07:
> Can you move this in a separate PR? That will help speed up the review.

Yes. I have moved that into a separate PR.

vgvassilev:
All commits except for “ Add support for differentiating calls to member functions in the reverse mode” (which I need more time to review) lgtm. If you move them in a separate pr we can merge them straight away.

parth-07:
> If you move them in a separate pr we can merge them straight away.

I have moved the first 2 commits to the separate PR here: https://github.com/vgvassilev/clad/pull/401

vgvassilev:
Can you rebase?

parth-07:
> Can you rebase?

Yes. I have rebased the pull request.

include/clad/Differentiator/CladConfig.h
```diff
@@ -31,16 +31,16 @@ void trap(int code) {
 
 #ifdef  __CUDACC__
 template<typename T>
-__device__ T* addressof(T& r) {
+__device__ T* clad_addressof(T& r) {
```
vgvassilev, line 34: Why did we change that name here?

include/clad/Differentiator/CladUtils.h
```diff
@@ -207,6 +208,10 @@ namespace clad {
     /// `{arr, arrSize}`
     clang::Expr* BuildCladArrayInitByConstArray(clang::Sema& semaRef,
                                                 clang::Expr* constArrE);
+
+    /// Returns true if `FD` is a class instance method; otherwise returns
+    /// false.
```
vgvassilev, line 213:
```suggestion
    /// Returns true if `FD` is a class member function; otherwise returns
    /// false.
```

Perhaps adjust the invert the name of the function to `IsStaticMethod` as well to somewhat clearer to the casual reader.

lib/Differentiator/CladUtils.cpp
```diff
@@ -89,9 +89,13 @@ namespace clad {
           DC2 = DC2->getParent();  
           continue;
         }
+        // Silently return nullptr if DC2 contains any CXXRecord declaration
+        // context.
```
vgvassilev, line 93: Instead of saying what we do (as it is somewhat visible from the code), can we say why we do it.

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -310,6 +310,31 @@ namespace clad {
...
+      ///
+      // FIXME: Add support for class type in the hessian matrix. For this, we
+      // need to add a way to represent hessian matrix when class type objects
+      // are involved.
+      if (auto MD = dyn_cast<CXXMethodDecl>(m_Function)) {
+        const CXXRecordDecl* RD = MD->getParent();
+        if (MD->isInstance() && !RD->isLambda()) {
+          QualType thisObjectType =
+              clad_compat::CXXMethodDecl_GetThisObjectType(m_Sema, MD);
+          thisObjectType.removeLocalConst();
```
vgvassilev, line 327: This is the case of `struct S{void f() const; }` right, and here we drop the const of the implicit this pointer?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1127,9 +1132,10 @@ namespace clad {
     }
 
     auto NArgs = FD->getNumParams();
-    // If the function has no args then we assume that it is not related
-    // to independent variables and does not contribute to gradient.
-    if (!NArgs)
+    // If the function has no args and is not a member function call then we
+    // assume that it is not related to independent variables and does not
+    // contribute to gradient.
+    if (!NArgs && !dyn_cast<CXXMemberCallExpr>(CE))
```
vgvassilev, line 1138:
```suggestion
    if (!NArgs && !isa<CXXMemberCallExpr>(CE))
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1146,8 +1152,10 @@ namespace clad {
     // f(x = y))
     // If the callee function takes arguments by reference then it can affect
     // derivatives even if there is no `dfdx()` and thus we should call the
-    // derived function.
-    if (!dfdx() && !utils::HasAnyReferenceOrPointerArgument(FD)) {
+    // derived function. In the case of member functions, `implicit`
+    // this object is always passed by reference.
+    if (!dfdx() && !utils::HasAnyReferenceOrPointerArgument(FD) &&
+        !dyn_cast<CXXMemberCallExpr>(CE)) {
```
vgvassilev, line 1158:
```suggestion
        !isa<CXXMemberCallExpr>(CE)) {
```

vgvassilev: Can we add a test for this branch here?

vgvassilev (commented):
This looks really good. I've left some comments. Also, do we know why the tests for clang-5 fail?

include/clad/Differentiator/CladConfig.h
parth-07: Because `std` namespace also has a function with the same name (https://en.cppreference.com/w/cpp/memory/addressof). If we were passing an object of `std` namespace to the unqualified `addressof` function. Then 2 functions were being found: `clad::addressof` and `std::addressof` (due to argument dependent lookup).

lib/Differentiator/HessianModeVisitor.cpp
parth-07: Yes! 
We only drop the `const` qualifier from the derivative type. This behaviour is consistent with the derivative types of non-class types. For example, a derivative of `const double` variable is of type `double`. 
I will add details on why we are removing `const` qualifier as a comment in the codebase.

include/clad/Differentiator/CladConfig.h
parth-07: If we don't want to change the name, then we can always make a qualified call to `addressof` (such as, `clad::addressof`).

lib/Differentiator/CladUtils.cpp
parth-07: I have added more details there.

lib/Differentiator/HessianModeVisitor.cpp
vgvassilev: Maybe add a comment to capture that clarification?

parth-07: I have added the clarification that this behaviour is consistent with non-class types as well.

lib/Differentiator/CladUtils.cpp
```diff
@@ -255,5 +281,12 @@ namespace clad {
       llvm::SmallVector<Expr*, 2> args = {constArrE, sizeE};
       return semaRef.ActOnInitList(noLoc, args, noLoc).get();
     }
+
+    bool IsStaticMethod(const clang::FunctionDecl* FD) {
+      if (auto MD = dyn_cast<CXXMethodDecl>(FD)) {
```
vgvassilev, line 286: We don't need braces here.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -464,8 +466,11 @@ namespace clad {
     DiffParams args{};
     std::copy(FD->param_begin(), FD->param_end(), std::back_inserter(args));
 
-    assert(!args.empty() && "Cannot generate pullback function of a function "
-                            "with no differentiable arguments");
+    bool isInstanceMethod = !utils::IsStaticMethod(FD);
```
vgvassilev, line 469: I think the point of the `IsStaticMethod` rename was to make the code paths refer to the non-static functions, which I thought would make the code clearer.

vgvassilev: I guess I imagine the code to look something like this:
```
...
if (!IsStaticMethod)
  // handle this ptr, etc.
...
```

IsInstanceMethod also works but it makes it a little harder to read it because one needs to figure out that's the opposite of static...

parth-07: Oh okay, now I understand. I will make the necessary changes.

parth-07: I have made the necessary changes.

vgvassilev (approved):
LGTM!

PR #398, opened 2022-03-14
Removed strcpy and strlen Definitions from Differentiator.h and inlin…

### LINKED ISSUE #388: Including cstring in a clad program throws errors
Minimum reproducible example:
```cpp
#include <iostream>
#include "clad/Differentiator/Differentiator.h" // order of 'Differentiator.h` and `cstring` is important
#include <cstring>                              

int main() {}
```

Result:
```
In file included from bug-cstring-include.cpp:3:
In file included from /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/cstring:42:
/usr/include/string.h:122:14: error: exception specification in declaration does not match previous declaration
extern char *strcpy (char *__restrict __dest, const char *__restrict __src)
             ^
/home/parth/clubs/open-source/clad/pristine/builds/build-10/inst/include/clad/Differentiator/Differentiator.h:23:9: note: previous declaration is here
  char* strcpy (char* destination, const char* source);
        ^
In file included from bug-cstring-include.cpp:3:
In file included from /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/cstring:42:
/usr/include/string.h:385:15: error: exception specification in declaration does not match previous declaration
extern size_t strlen (const char *__s)
              ^
/home/parth/clubs/open-source/clad/pristine/builds/build-10/inst/include/clad/Differentiator/Differentiator.h:24:10: note: previous declaration is here
  size_t strlen(const char*);
         ^
2 errors generated.
/usr/lib/llvm-10/bin/clang++ -g -Xclang -add-plugin -Xclang clad -Xclang -load -Xclang /home/parth/clubs/open-source/clad/pristine/builds/build-10/inst/lib/clad.so -I/home/parth/clubs/open-source/clad/pristine/builds/build-10/inst/include -x c++ -std=c++11 bug-cstring-include.cpp -o build/bug-cstring-include && build/bug-cstring-include
```

Nirhar:
strcpy function was not used anywhere else in the codebase and hence I removed its definition from Differentiator.h. strlen was used once in Differentiator.h, which i replaced it with the code of strlen.

Closes #388

### DIFF
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -20,12 +20,12 @@
 
 extern "C" {
   int printf(const char* fmt, ...);
-  char* strcpy (char* destination, const char* source);
-  size_t strlen(const char*);
 #if defined(__APPLE__) || defined(_MSC_VER)
+  extern size_t strlen(const char*);
   void* malloc(size_t);
   void free(void *ptr);
 #else
+  extern size_t strlen(const char*) __THROW __attribute_pure__ __nonnull ((1));
   void* malloc(size_t) __THROW __attribute_malloc__ __wur;
   void free(void *ptr) __THROW;
 #endif


include/clad/Differentiator/Differentiator.h
```diff
@@ -36,15 +34,11 @@ namespace clad {
...
     const char* code_copy = code;
-    #ifdef __CUDACC__
-      count = 0;
+    count = 0;
       while (*code_copy != '\0') {
         count++;
         code_copy++;
       }
-    #else
-      count = strlen(code_copy);
```
vgvassilev, line 46: I think we want to use `strlen` if it was provided by the system.

Nirhar: is it fine if I #include <cstring> in this file and use strlen from cstring?

Nirhar: #include cstring i meant

vgvassilev: Instead of including cstring, I suspect that we should find the "right" way to forward declare `strlen`.

Nirhar:
I have resolved the issue by including the exact declaration of strlen(with the extern) from the string standard library. Can you check once if that's alright?

vgvassilev:
@parth-07, can you take a look why the bots fail?

vgvassilev:
Can you squash the two commits into one?

parth-07:
> I have resolved the issue by including the exact declaration of strlen(with the extern) from the string standard library

@Nirhar Different compilers, and also perhaps different versions of the same compiler may have a different declaration of `strlen` function. Can this cause the original error to happen again?

Nirhar:
@vgvassilev I have squashed the commits into 1

@parth-07 In my opinion this error was not due to a compiler issue, It was because of conflict in headers arising from the definition in the standard string library(which is installed in a person's computer and linked with our code during preprocessing phase) and our strlen definition. Since the definition of strlen has been standardised across platforms and has not changed for years, this error should not occur again.

vgvassilev:
@Nirhar, the builds seem unhappy with this change. Probably the signature is different on the different platforms.

Nirhar:
Okay, I was wrong with my previous comment that these libraries are common across platforms. I have looked into the Mac version for string.h and have updated the code accordingly to take into account both Linux as well as Mac OS. Hope all tests pass this time.

vgvassilev (approved):
LGTM, thanks a lot!

PR #401, opened 2022-03-19
Refactor building params routine in the reverse mode AD.

### LINKED ISSUE #393: Cannot compute gradient of functions that have default arguments
Minimum reproducible example: 

```cpp
double fn(double i, double j = 0) {
  return i+j;
}

int main() {
}
```

**Output:**
```
error: missing default argument on parameter '_d_i'
error: missing default argument on parameter '_d_j'
2 errors generated. 
```

The root cause of the error is that we are adding default arguments in the derived function for the parameters that had default arguments in the original function. This does not work well for the derived functions because we append derivative parameters at the end, and C++ requires all parameters on the right-hand side of a parameter with default argument to have default argument as well. 
This can be easily fixed by removing default arguments in the derived function.

parth-07:
This PR refactors `ReverseModeVisitor::Derive` and `ReverseModeVisitor::DerivePullback` such that the derived function parameter types and the parameters declarations are computed and built in separate subroutines. This facilitates the reuse of the common code between different differentiation modes supported by the reverse mode (DiffMode::reverse, DiffMode::experimental_pullback, and DiffMode::jacobian).

This PR also fixes the differentiation of functions consisting of default args in the reverse mode AD. 

Closes #393

### DIFF
--- a/include/clad/Differentiator/CladUtils.h
+++ b/include/clad/Differentiator/CladUtils.h
@@ -185,11 +185,15 @@ namespace clad {
     ///
     /// This function is just a convenient routine that internally calls
     /// `clang::ParmVarDecl::Create`.
+    ///
+    /// \note `TSI` parameter only needs to be provided if the type should be
+    /// represented exactly how it was represented in the source code.
     clang::ParmVarDecl*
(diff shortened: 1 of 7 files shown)


vgvassilev (approved):
LGTM!

PR #405, opened 2022-03-22
Restructure Documentation

parth-07:
This PR restructures readthedocs layout to the design discussed recently. It adds 2 main files (`docs/userDocs/source/user/UsingClad.rst` and `docs/userDocs/source/user/CoreConcepts.rst`). Most of the new information should be added to these files. 

readthedocs with the updated layout can be visited here: https://trying-read-the-docs.readthedocs.io/en/latest/


docs/userDocs/source/conf.py
```diff
@@ -72,7 +72,19 @@
...
 
-
+# latex_elements = {
+#     "preamble": r"""
+#     \usepackage{physics}
+#     """,
+#     "extrapackages": r"""
+#     \usepackage{physics}
+#     """
+# }
```
vgvassilev, line 82: Do we need these comments?

docs/userDocs/source/user/UsingClad.rst
```diff
@@ -0,0 +1,59 @@
+Using Clad
+***********
```
vgvassilev, line 2: We should decide either to use `***` or `===` I think they both mean the same.

docs/userDocs/source/user/UsingClad.rst
```diff
@@ -0,0 +1,59 @@
...
+You may want to skim some sections on the first read. 
+
+In case you haven't installed Clad already, then please do before proceeding 
+with this guide. Visit :doc:`Clad installation and usage <InstallationAndUsage>` 
+to know more about installing clad.
+
+Let's get started.
+
+Automatic Differentiation
+===========================
```
vgvassilev, line 15: We should just use `===` otherwise depending on the editor the formatting of the raw text looks weird.

docs/userDocs/source/user/UsingClad.rst
parth-07: I was following the style guide from here: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections

It basically assigns different heading character (`*`, `=`, `-` etc) depending on the heading level.

docs/userDocs/source/user/UsingClad.rst
parth-07: Okay.

docs/userDocs/source/conf.py
parth-07: No, we don't. 
I was experimenting with different math display formats. I forgot to remove the comments. I am removing them.

docs/userDocs/source/user/UsingClad.rst
parth-07: It seems that we cannot just use `===` (3 characters) for heading. 
[RestructuredText Primer quickstart](https://docutils.sourceforge.io/docs/user/rst/quickstart.html) states "The underline/overline must be at least as long as the title text."

vgvassilev: Ok, can you open a new PR?

parth-07: I had not made the change of shortening the heading characters in this PR yet.

PR #417, opened 2022-03-25
Add support for diff overloaded operators in the forw mode

parth-07:
This PR adds support for differentiating overloaded operators in the forward mode. 

It also fixes the cloning of `CXXOperatorCallExpr` node.

### DIFF
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -14,12 +14,91 @@ using namespace clang;
 namespace clad {
   namespace utils {
     static SourceLocation noLoc{};
-    
+
     std::string ComputeEffectiveFnName(const FunctionDecl* FD) {
-      // TODO: Add cases for more operators
       switch (FD->getOverloadedOperator()) {
-        case OverloadedOperatorKind::OO_Call: return "operator_call";
-        default: return FD->getNameAsString();
+      case OverloadedOperatorKind::OO_Plus:
+        return "operator_plus";
+      case OverloadedOperatorKind::OO_Minus:
+        return "operator_minus";
+      case OverloadedOperatorKind::OO_Star:
+        return "operator_star";
+      case OverloadedOperatorKind::OO_Slash:
+        return "operator_slash";
+      case OverloadedOperatorKind::OO_Percent:
+        return "operator_percent";
+      case OverloadedOperatorKind::OO_Caret:
+        return "operator_caret";
+      case OverloadedOperatorKind::OO_Amp:
+        return "operator_amp";
+      case OverloadedOperatorKind::OO_Pipe:
+        return "operator_pipe";
+      case OverloadedOperatorKind::OO_Tilde:
+        return "operator_tilde";
+      case OverloadedOperatorKind::OO_Exclaim:
+        return "operator_exclaim";
+      case OverloadedOperatorKind::OO_Equal:
+        return "operator_equal";
+      case OverloadedOperatorKind::OO_Less:
+        return "operator_less";
+      case OverloadedOperatorKind::OO_Greater:
+        return "operator_greater";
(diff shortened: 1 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -710,7 +710,7 @@ jobs:
       if: ${{ failure() }}
       uses: mxschmitt/action-tmate@v3
       # When debugging increase to a suitable value!
-      timeout-minutes: ${{ github.event.pull_request && 1 || 20 }}
+      timeout-minutes: 35
```
vgvassilev, line 713: Can you undo this?

lib/Differentiator/CladUtils.cpp
```diff
@@ -14,12 +14,87 @@ using namespace clang;
...
+        return "operator_plus_plus";
+      case OverloadedOperatorKind::OO_MinusMinus:
+        return "operator_minus_minus";
+      case OverloadedOperatorKind::OO_Comma:
+        return "operator_comma";
+      case OverloadedOperatorKind::OO_ArrowStar:
+        return "operator_arrow_star";
+      case OverloadedOperatorKind::OO_Arrow:
+        return "operator_arrow";
+      case OverloadedOperatorKind::OO_Call:
```
vgvassilev, line 92: I'd add test cases for these operators to appease the coverage bots.

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -132,17 +134,28 @@ namespace clad {
...
+        m_DeclReplacements[PVD] = newPVD;
+
       QualType nonRefParamType = PVD->getType().getNonReferenceType();
       // FIXME: Add support for pointer and array parameters in the
       // forward mode.
       if (!utils::isDifferentiableType(PVD->getType()) ||
           utils::isArrayOrPointerType(PVD->getType()))
         continue;
-      auto derivedPVDName = "_d_" + PVD->getNameAsString();
+      auto derivedPVDName = "_d_" + std::string(PVDII->getName());
```
vgvassilev, line 158: Why did we need this change?

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -1057,26 +1072,68 @@ namespace clad {
...
       StmtDiff argDiff = Visit(arg);
       if (!Multiplier)
         Multiplier = argDiff.getExpr_dx();
       else {
         Multiplier = BuildOp(BO_Add, Multiplier, argDiff.getExpr_dx());
       }
+
+      /// If original argument is an RValue and function expects an RValue
+      /// parameter, then convert the cloned argument and the corresponding
+      /// derivative to RValue if they are not RValue.
```
vgvassilev, line 1123: Decide if we will use doxygen style comments in function bodies or not. My feeling is that the rest does not use doxygen style comments.

lib/Differentiator/ForwardModeVisitor.cpp
parth-07: If the parameter originally did not have a name, then we are creating a unique identifier for the parameter. 
Thus, PVDII is not always the same as `PVD->getIdentifier()`. 

Some of the special member functions created by Clang have parameters that do not have any associated parameters.

lib/Differentiator/ForwardModeVisitor.cpp
parth-07: Yes, we do not use doxygen style comments inside a function body. I must have used them here by mistake.

lib/Differentiator/CladUtils.cpp
parth-07: I have added test cases for most of the operators now.

vgvassilev (approved):
LGTM!

Issue #382, opened 2022-02-21
Incorrect handling of reference variables in the forward mode AD.


parth-07:
Reproducible example:
```cpp
void modify(double& i, double& j) {
    j *= 2;
}

double fn(double i, double j) {
    modify(i, j);
    modify(i, j);
    i *= j;
    return i;
}

int main() {
    auto d_fn = clad::differentiate(fn, "i");
    std::cout<<d_fn.execute(3, 5)<<"\n";
}
```

**Expected output:** 20
**Actual output:** 80

# The root problem

The root cause of the problem is that we compute the same code of the function `modify` twice each time it is called. To understand the problem, please consider the portion of the derived function generated for `fn` in the above example:

```cpp
    modify_pushforward(i, j, _d_i, _d_j);
    modify(i, j);
    modify_pushforward(i, j, _d_i, _d_j);
    modify(i, j);
```

Here for each call to `modify`, we have a corresponding call to `modify_pushforward`. And the way the forward mode AD works, it runs all the statements of the original function (`modify` function here). In the case of reference arguments, the effect of computing function statement twice is "seen" as well, and thus leads to producing incorrect results. We will have this exact same problem for variables passed as pointers. 

# Viable solutions

## 1) Pass copy of original variables to the pushforward functions.

We can pass a copy of the original variables which needs to be passed by reference to the pushforward functions. This way the pushforward functions will not be able to modify the original variable, and the effect of computing function code twice will not be "visible". 

### Pros
Easy to implement

### Cons
- Creation of a temporary variable for each reference argument.
- We are still computing function code twice. 

## 2) Generate a function that computes both the function result and its corresponding pushforward. 

We can generate pushforward functions in a way that they return the original function result as well as the corresponding pushforward.
It is possible because we process all the statements of the original function in the pushforward function already. 

### Pros
- Function code will only be computed once, thus we will reduce overall computation time.

Sample resulting code snippet if we use this solution:

```cpp
res = update(i, j);
```
_Original statement_

```cpp
auto resWithDerivative = update_pushforward(i, j, _d_i, _d_j);
_d_res  = resWithDerivative.derivative;
res = resWithDerivative.result;
``` 
_Corresponding derived statement_

parth-07:
Resolved by #420

PR #420, opened 2022-03-29
Modify pushforward fn mechanism to compute both fn value and derivative.

parth-07:
This PR modifies pushforward function mechanism to compute both the function value as well as the pushforward. 

**Major breaking change from user's perspective:** Already existing custom derivative pushforward functions will need to be modified to follow the design and rules of the updated pushforward function mechanism. 

## Advantages

This change in the pushfoward function mechanism has the following advantages:

- It leads to less computation in the derived function and thus produces more optimal derived function code.

```cpp
double c = someFn(a, b, c);
```

Earlier, we were differentiating this call expression as follows: 
```cpp
double _d_c = someFn_pushforward(a, b, c, _d_a, _d_b, _d_c);
double c = someFn(a, b, c);
```
Here all the statements of `someFn` are processed twice -- first, in the call to `someFn_pushforward` and then again in the call to `someFn`. This is because in the forward mode derived function we run all the statements of the original function.

With the updated pushforward mechanism, differentiating call expressions will not involve any repeated computation and both the function value and it's pushforward will be computed in a single pass of the pushfoward function. Please consider the following code for a concrete example of differentiation using the updated pushforward mechanism.  

```cpp
double c = someFn(a, b, c);
```

Using the updated pushforward mechanism, the above code will be differentiated as follows:

```cpp
auto valueWithPushforward = someFn_pushforward(a, b, c, _d_a, _d_b, _d_c);
double _d_c = valueWithPushforward.pushforward;
double c = valueWithPushforward.value;
```
- Multiple processing of the same statements of the original function is especially troublesome when the call expression passes arguments by reference. This is because modifications to arguments passed by reference are reflected in the original variables that are passed. And making same modification twice will result in incorrect values and derivatives. Issue #382 describes this problem in more detail. The updated pushforward mechanism solves this problem by removing multiple processing of same statements.

- This PR also modifes custom derivatives to use the updated pushforward mechanism as well, and thus custom derivatives provided by users can also have all the benefits of the updated pushforward mechanism. In particular, users can design custom derivative pushforward functions in such a way that allows for maximum reusability of computations by the original function section and the pushforward computation section.

## Disadvantages

The updated pushforward design has one major disadvantage that should be noted. The complexity of higher-order pushforward functions increases considerably. This wouldn't be a big of an issue if only Clad automatically generates the pushforward function. But we need to use the updated pushforward design for custom derivatives as well. Handwritting higher-order custom derivatives using the new design is more difficult and error-prone. For example, let's write 2nd order pushforward function for `std::sin`:

**1st order pushforward**
```cpp
template<typename T>
clad::ValueAndPushforward<T, T> sin_pushforward(T i, T d_i) {
  return {::std::sin(i), ::std::cos(i) * d_i};
}
```

The functioning of 1st order pushforward function `sin_pushforward` is very intuitive. Various terms used in its
definition can be understood as follows: 

consider 'x' as the independent variable with respect to which differentiation is being taken place.

- d_i: derivative of 'i' w.r.t 'x'
- FunctionResult.value: value of the function call: `sin(i)`
- FunctionResult.pushforward: pushforward of the function call: `sin(i)`. Mathematically, it is, derivative of the function w.r.t 'i' if the seed value 'd_i' is 1.

**2nd order pushforward**
```cpp
template<typename T>
clad::ValueAndPushforward<clad::ValueAndPushforward<T, T>, clad::ValueAndPushforward<T, T>>
sin_pushforward_pushforward(T i, T d_i, T d_i0, T d_d_i) {
  return {sin_pushforward(x, d_i),  {sin_pushforward(i, d_i0).pushforward, -1 * ::std::sin(i) * d_i0 * d_i + ::std::cos(i) * d_d_i}};
}
```

The functioning and the result of this 2nd order pushforward function is not very intuitive at all. We should also note that with each increase in the differentiation order, the complexity of the pushforward function will increase significantly. Various terms used in its definition can be understood as follows:

consider 'x' as the independent variable with respect to which differentiation is being taken place.

-  `d_i`: derivative of 'i' w.r.t 'x'.
- `d_i0`: derivative of 'i' w.r.t 'x'.
- `d_d_i`: 2nd order derivative of 'i' w.r.t 'x'.
- `FunctionResult.value.value`: value of the function call `sin(i)`.
- `FunctionResult.value.pushforward`: 1st order pushforward of `sin(i)`. 
- `FunctionResult.pushforward.value`: 1st order pushforward of `sin(i)`.
- `FunctionResult.pushforward.pushforward`: 2nd order pushforward of `sin(i)`. 

### why do we have `d_i` and `d_i0` representing the same thing?

They will not always represent the same thing. For 2nd order derivatives, they will represent the same thing only if we are differentiating with respect to the same variable twice. For example, consider the function:

```cpp
double fn(double x1, double x2, double x3) {...} 
```
If we want to compute:

Screenshot from 2022-03-29 17-07-51

Then, we will need to make the following call to the pushforward function:

```cpp
double res = fn_pushforward(x1, x2, x3, /*dx1/dx1=*/1, /*dx2/dx1=*/0, /*dx3/dx1=*/0, /*dx1/dx2=*/0, /*dx2/dx2=*/1, /*dx3/dx2=*/0, /*ddx1/ddx2=*/0, /*ddx2/ddx2=*/0, /*ddx3/ddx2=*/0);
```

Here, none of the seed variable represent the same thing. Effectively, this call is first computing derivative of `fn` w.r.t
`x1`, and then computing the derivative of previously computed derivative w.r.t `x2`.

We can use the same `fn_pushforward` to compute any of the following derivatives by modifying the seed values: 

Screenshot from 2022-03-29 17-15-48

Users cannot use this functionality currently because there is no user interface function for directly using pushforwards.

-----

This PR also marks 2 hessian tests as `XFAIL`. We need to improve class type support in the reverse mode to fix these tests.

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -6,7 +6,6 @@
 
 #ifndef CLAD_BUILTIN_DERIVATIVES
 #define CLAD_BUILTIN_DERIVATIVES
-
 // Avoid assertion custom_derivative namespace not found. FIXME: This in future
 // should go.
 namespace custom_derivatives{}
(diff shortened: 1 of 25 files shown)


test/ROOT/TFormula.C
```diff
@@ -66,233 +56,32 @@ void TFormula_example_grad_1(Double_t* x, Double_t* p, Double_t* _d_p);
...
+//CHECK-NEXT:       clad::ValueAndPushforward<Double_t, Double_t> _t2 = clad::custom_derivatives{{(::std)?}}::TMath::Abs_pushforward(p[1], 0);
+//CHECK-NEXT:       return 0 * _t0 + x[0] * (1 + 0 + 0) + _t1.pushforward + _t2.pushforward;
 //CHECK-NEXT:   }
 
 //CHECK:   Double_t TFormula_example_darg1_1(Double_t *x, Double_t *p) {
 //CHECK-NEXT:       double _t0 = (p[0] + p[1] + p[2]);
-//CHECK-NEXT:       return 0 * _t0 + x[0] * (0 + 1 + 0) + clad::custom_derivatives::TMath::Exp_pushforward(-p[0], -0) + clad::custom_derivatives::TMath::Abs_pushforward(p[1], 1);
-//CHECK-NEXT:   }
-
-//CHECK:   void TFormula_example_darg1_1_grad_1(Double_t *x, Double_t *p, clad::array_ref<Double_t> _d_p) {
```
vgvassilev, line 203: Why did we drop these?

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -18,85 +18,77 @@ namespace custom_derivatives{}
...
+  return {::std::log(x), static_cast<T>((1.0 / x) * d_x)};
 }
 
 template <typename T1, typename T2>
 CUDA_HOST_DEVICE void
 pow_pullback(T1 x, T2 exponent, decltype(::std::pow(T1(), T2())) d_y,
              clad::array_ref<decltype(::std::pow(T1(), T2()))> d_x,
              clad::array_ref<decltype(::std::pow(T1(), T2()))> d_exponent) {
-  *d_x += pow_pushforward(x, exponent, static_cast<T1>(1), static_cast<T2>(0)) *
+  *d_x += pow_pushforward(x, exponent, static_cast<T1>(1), static_cast<T2>(0))
```
vgvassilev, line 75: Is this formatting suggested by clang-format? Let's try to use temporaries to make this more readable.

include/clad/Differentiator/DiffReturnTypes.h
```diff
@@ -0,0 +1,11 @@
+#ifndef CLAD_DIFF_RETURN_TYPES_H
+#define CLAD_DIFF_RETURN_TYPES_H
+
+namespace clad {
+template <typename T, typename U> struct ValueAndPushforward {
+  T value;
+  U pushforward;
+};
+} // namespace clad
```
vgvassilev, line 9: This definition should go in `BuiltinDerivatives.h`

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -690,12 +704,21 @@ namespace clad {
...
-            .ActOnReturnStmt(noLoc,
-                             retValDiff.getExpr_dx(), // return the derivative
-                             m_CurScope)
-            .get();
+    Stmt* returnStmt = nullptr;
+    if (m_Mode == DiffMode::forward) {
+      returnStmt =
+          m_Sema
+              .ActOnReturnStmt(noLoc,
+                               retValDiff.getExpr_dx(), // return the derivative
```
vgvassilev, line 712: Can you move this into a temporary and reformat?

vgvassilev (commented):
Overall looks good to me.

test/ROOT/TFormula.C
parth-07: `TFormula.C` file has three tests -- one for forward mode AD, one for reverse-mode AD and one for hessian computation. Currently, the hessian computation part fails, and I have thus moved that to a separate file `test/ROOT/Hessian.C` with `XFAIL`. The dropped `CHECK` directives belong to the hessian computation part and have thus been moved to `test/ROOT/Hessian.C` test file.

include/clad/Differentiator/BuiltinDerivatives.h
parth-07: I have created temporaries here to make this more readable.

lib/Differentiator/ForwardModeVisitor.cpp
parth-07: Done.

vgvassilev (approved):
LGTM!

PR #422, opened 2022-03-30
Modified UpdateDiffParamsInfo to not add result as a parameter in jacobian mode

### LINKED ISSUE #408: Incorrect result when calling clad::jacobian wrt given parameter
Incorrect result when calling clad::jacobian wrt given parameter, the issue can be seen in the code below. The derivatives matrix should have contained [0, 10, 3]. 

image

The behaviour is correct when calling clad::jacobian wrt all parameters:

image

For reference, the generated code is:

image

Nirhar:
I have modified ```DiffRequest::UpdateDiffParamsInfo``` to ignore the last parameter(which is the result parameter), if the Mode is jacobian. Accordingly I have changed the ```ReverseModeVisitor::Derive``` to not pop the last element in jacobian mode.

As far as I understand the codebase, the changes I have made must only affect Jacobian mode of Differentiation. Please let me know if you feel that there are some other dependencies that I must consider.

closes #408

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -262,17 +262,25 @@ namespace clad {
     std::string gradientName = derivativeBaseName + funcPostfix();
     // To be consistent with older tests, nothing is appended to 'f_grad' if
     // we differentiate w.r.t. all the parameters at once.
-    if (!(args.size() == FD->getNumParams() &&
-          std::equal(FD->param_begin(), FD->param_end(), std::begin(args)))) {
-      for (auto arg : args) {
-        auto it = std::find(FD->param_begin(), FD->param_end(), arg);
-        auto idx = std::distance(FD->param_begin(), it);
-        gradientName += ('_' + std::to_string(idx));
+    if(isVectorValued){
+      // If Jacobian is asked, the last parameter is the result parameter
+      // and should be ignored
+      if (args.size() != FD->getNumParams()-1){
+        for (auto arg : args) {
+          auto it = std::find(FD->param_begin(), FD->param_end()-1, arg);
+          auto idx = std::distance(FD->param_begin(), it);
+          gradientName += ('_' + std::to_string(idx));
+        }
+      }
+    }else{
+      if (args.size() != FD->getNumParams()){
+        for (auto arg : args) {
+          auto it = std::find(FD->param_begin(), FD->param_end(), arg);
+          auto idx = std::distance(FD->param_begin(), it);
+          gradientName += ('_' + std::to_string(idx));
+        }
       }
     }
-
-    if (isVectorValued)
-      args.pop_back();
 
     IdentifierInfo* II = &m_Context.Idents.get(gradientName);
     DeclarationNameInfo name(II, noLoc);
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -432,14 +432,24 @@ namespace clad {
     // function's arguments were requested.
     if (isa<CXXDefaultArgExpr>(E)) {
       std::copy(FD->param_begin(), FD->param_end(), std::back_inserter(params));
+      
       // If the function has no parameters, then we cannot differentiate it."
-      if (params.empty()) {
+      // and if the DiffMode is Jacobian, we must have atleast 2 parameters.
+      if (params.empty() || (params.size()==1 && this->Mode == DiffMode::jacobian)) {
         utils::EmitDiag(semaRef, DiagnosticsEngine::Error,
                         CallContext->getEndLoc(),
                         "Attempted to differentiate a function without "
                         "parameters");
         return;
       }
+
+      // If it is a Vector valued function, the last parameter is to store the output vector
+      // and hence is not a differentiable parameter, so we must pop it out
+      if (this->Mode == DiffMode::jacobian){
+        params.pop_back();
+      }
+
+
       IndexIntervalTable indexes{};
       // insert an empty index for each parameter.
       for (unsigned i=0; i<params.size(); ++i)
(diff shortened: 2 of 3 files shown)


Nirhar:
Codecov is failing because I have introduced a new branch in the code, which is not covered by the existing test cases. On checking the test-set, I found out that the test-set doesn't contain any test cases for Jacobian calculation with respect to a select few input parameters. In the absence of such testcases, the new branch that I have introduced is not taken.

@parth-07 and @vgvassilev can you give me some references on how to write test cases and what libraries are used for the same, so that I can add some new test cases.

vgvassilev:
@Nirhar, can you remove the merge commits by rebasing on top of master?

@parth-07, can you take a look?

parth-07:
> can you give me some references on how to write test cases

@Nirhar We use python LIT testing framework combined with LLVM FileCheck as the pattern matching tool. Please see existing test files to get an idea on how to write the tests. Briefly, the process of writing and running test is as follows: 

Each test file has commands specified in comments at the top of the file. An Example of such a command is as follows: 

```cpp
// RUN: %cladclang %s -lstdc++ -I%S/../../include -oPointer.out 2>&1 | FileCheck %s
// RUN: ./Pointer.out | FileCheck -check-prefix=CHECK-EXEC %s
```

These commands are passed to LIT. Make sure no two test file in the same folder has same test output name. In this particular example, `Pointer` is the test output name. 

A test gets automatically registered if it's under the `/test/` directory and has extension `.C`. 

To run all the present tests, run command `lit test` at the root of the build directory. To run tests under a particular folder, run `lit test/FolderName`. 

Please tell us if you are having any other problems while creating tests.

Nirhar:
Thanks for the help @parth-07 !
The tests have passed now. @vgvassilev are there any other changes that I need to do?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -262,18 +262,28 @@ namespace clad {
     std::string gradientName = derivativeBaseName + funcPostfix();
     // To be consistent with older tests, nothing is appended to 'f_grad' if
     // we differentiate w.r.t. all the parameters at once.
-    if (!(args.size() == FD->getNumParams() &&
+    if(isVectorValued){
+      // If Jacobian is asked, the last parameter is the result parameter
+      // and should be ignored
+      if (!(args.size() == FD->getNumParams()-1 &&
+          std::equal(FD->param_begin(), FD->param_end()-1, std::begin(args)))) {
```
parth-07, line 269: Can the second condition that consists of a call to `std::equal` ever be false if the first condition is true?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -262,18 +262,28 @@ namespace clad {
...
+        for (auto arg : args) {
+          auto it = std::find(FD->param_begin(), FD->param_end()-1, arg);
+          auto idx = std::distance(FD->param_begin(), it);
+          gradientName += ('_' + std::to_string(idx));
+        }
+      }
+    }else{
+      if (!(args.size() == FD->getNumParams() &&
```
parth-07, line 277: can `args.size()` ever be equal to `FD->getNumParams()` for jacobian mode?

parth-07: It seems that we are now popping the last parameter for jacobian mode in the `DiffPlanner` section.

parth-07 (changes requested):

lib/Differentiator/ReverseModeVisitor.cpp
Nirhar: I wrote this if block based on an already existing code snippet. The else part of this if(isVectorValued) statement has a if statement inside it, which was already a part of the codebase. I wrote the if condition(line 268) based on that if statement(line 277). Logically both conditions mean the same, but I am not sure if there are any edge cases where the conditions are not equivalent. @vgvassilev what's your opinion?

lib/Differentiator/ReverseModeVisitor.cpp
Nirhar: Perhaps args.size() might be equal to FD->getNumParams() if we don't consider a vector valued function(ie consider a function with non-void return type and no input result parameter). Finding the jacobian of such a function is the same as 
ReverseMode AD. So I thought maybe that usecase is not supported by the Jacobian mode. Let me once check if this usecase is supported.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: We should not add 2 conditions inside an `if` statement if both the conditions say the same thing. Each condition should have a clear purpose to avoid any ambiguity. If you want to be extra sure, then add the second condition as an assert statement inside the `if` block. 

> I wrote the if condition(line 268) based on that if statement(line 277).

Then we should fix that `if` statement as well.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: > Finding the jacobian of such a function is the same as ReverseMode AD.

Using `jacobian` mode on such a function should ideally give an error. Currently, we require the last parameter to be the output parameter.

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: > 

What is the intent of the std::equal here? Maybe it can become an assert?

Nirhar: I've made the requested changes. @parth-07 can you have a look if its fine?

parth-07 (commented):
I think we should remove both the added asserts. In these asserts we are just ensuring that two sequences of different sizes are indeed different. Otherwise looks good to me.

Nirhar:
@parth-07 I have removed the assert statements now. Can you check if everything is fine?

parth-07 (approved):
Looks good to me.

vgvassilev (approved):

PR #423, opened 2022-04-01
Add initial support for differentiating pointers in forw mode

parth-07:
This PR adds initial support for differentiating pointers in the forward mode AD. 

This PR adds support for the following features:

- Address of operator (`&`) 
- Dereference operator (`*`)
- `new` operator
- `delete` operator

This PR does not add support for differentiating functions with pointers as arguments, because in that case, there is no way of determining if the pointer points to a single element or an array of elements.


vgvassilev (approved):
LGTM!

PR #424, opened 2022-04-06
Fix/clad params types

### LINKED ISSUE #385: Improve derivative type consistency.
```cpp
double fn(float f, float* arr) {
  float another_f;
  float another_arr[5];
}
```

This function will have the derived function signature as follows:

```cpp
double fn_grad(float f, float* arr, clad::array_ref<double> d_f, clad::array_ref<double>(arr)) {
  float d_another_f = 0;
  float another_f;
  clad::array<float> d_another_arr(5UL);
  float another_arr[5];
}
```

Please notice that `f` argument has the effective derivative type of `double` whereas `another_f` local variable has the effective derivative type of `float`.
And `arr` argument has the effective derivative type of `double*` whereas `another_arr` local variable has the effective derivative type of `float*`. 
Even though `f` and `another_f` are of the same type and `arr` and `another_arr` are of the same type. 

These type inconsistencies can make it difficult to reason behaviour of the program and also makes it more confusing for the users to estimate the derivative type of some variable. We should strive to keep such inconsistencies to as minimum as possible. 
One way to remove these inconsistencies is as follows:
Derivative obtained by differentiation of a variable of type `A` with respect to a variable of type `B` should always be of type `B` when `A` and `B` are both scalar numerical types.

ioanaif:
This aims to fix #385 :

- [x]   Preserving param types [1]:
```
int someFn(int i, double j) { return 7*i + 9*j; }
int fn(int i, double j) { return someFn(i, j) + 309; }
auto f_dx =  clad::gradient(fn);
void fn_grad(int i, double j, clad::array_ref<int> d_i, clad::array_ref<double> d_j)
```

         
- [x]  Preserving param types [2]:
```
void someFn(float i, int j, double &out) {
   out = 7*i + 9*j;
}
double fn(float i, int j, double &out) {
   someFn(i, j, out);
   return out + 309;
}
```

- [x]  Preserving param types - array type:
```
double addArr(int *arr, double n) {
   double ret = 0;
   for (int i = 0; i < n; i++) 
      ret += arr[i];
      return ret;
   }
double f(int* arr) { return addArr(arr, 4)};
auto f_dx = clad::gradient(f);
```

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1190,15 +1190,7 @@ namespace clad {
     // Save current index in the current block, to potentially put some
     // statements there later.
     std::size_t insertionPoint = getCurrentBlock(direction::reverse).size();
-    // Store the type to reduce call overhead that would occur if used in the
-    // loop
-    auto CEType = getNonConstType(CE->getType(), m_Context, m_Sema);
-    // Use `double` as the placeholder type for the derivatives when the funtion
-    // return type is void. We need to do this because we are using function
-    // return type to compute the derivative type of the arguments. See
-    // https://github.com/vgvassilev/clad/issues/385 for more details.
-    if (CEType->isVoidType())
-      CEType = m_Context.DoubleTy;
+
     // FIXME: We should add instructions for handling non-differentiable
     // arguments. Currently we are implicitly assuming function call only
     // contains differentiable arguments.
@@ -1215,8 +1207,6 @@ namespace clad {
         QualType argResultValueType =
             utils::GetValueType(argDiff.getExpr()->getType())
                 .getNonReferenceType();
-        if (argResultValueType->isRealType())
-          argResultValueType = CEType;
         // Create ArgResult variable for each reference argument because it is
         // required by error estimator. For automatic differentiation, we do not need
         // to create ArgResult variable for arguments passed by reference.
@@ -1253,7 +1243,7 @@ namespace clad {
         // same as the call expression as it is the type used to declare the
         // _gradX array
         Expr* dArg;
-        dArg = StoreAndRef(/*E=*/nullptr, CEType, direction::reverse, "_r",
+        dArg = StoreAndRef(/*E=*/nullptr, arg->getType(), direction::reverse, "_r",
                            /*forceDeclCreation=*/true);
         ArgResultDecls.push_back(
             cast<VarDecl>(cast<DeclRefExpr>(dArg)->getDecl()));
@@ -1396,12 +1386,6 @@ namespace clad {
 
       size_t idx = 0;
 
-      auto ArrayDiffArgType = GetCladArrayOfType(CEType.getCanonicalType());
-      QualType requiredValueType = CEType;
-      if (CEType->isVoidType()) {
-        requiredValueType = m_Context.DoubleTy;
-        ArrayDiffArgType = GetCladArrayOfType(requiredValueType);
-      }
       /// Add base derivative expression in the derived call output args list if
       /// `CE` is a call to an instance member function.
       if (auto MCE = dyn_cast<CXXMemberCallExpr>(CE)) {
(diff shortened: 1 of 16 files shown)


parth-07:
Regarding the failing MacOS tests for `real_pullback` and other `std::complex` functions, since the platforms are using different implementations of the standard library, we are currently using a complicated regex expression in LIT `CHECK`s to match both of them, this isn't a good solution. We have two ways to proceed: 

1) Have separate `CHECK`s directives for MacOS and Linux Github runners when the source function is different for the platforms. This can be done as follows: 

```cpp
#ifdef __APPLE__
// CHECK directives for macos
#else
// CHECK directives for linux
#endif
```

Alternatively, we can also check if `libc++` is being used as a standard library or `libstdc++` using the preprocessor flag `_LIBCPP_VERSION`. 

2) We can avoid checking LIT `CHECK`s for standard library functions if the above-mentioned approach becomes difficult to maintain.

ioanaif:
> 1. Have separate `CHECK`s directives for MacOS and Linux Github runners when the source function is different for the platforms. This can be done as follows:
> 
> ```c++
> #ifdef __APPLE__
> // CHECK directives for macos
> #else
> // CHECK directives for linux
> #endif
> ```

It seems that it skips these conditions, so I made the regex more lenient in that special case, but this might raise other issues in the future.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2883,12 +2820,18 @@ namespace clad {
     // derivative variables should always be of non-const type.
     xValueType.removeLocalConst();
     QualType nonRefXValueType = xValueType.getNonReferenceType();
-    if (nonRefXValueType->isRealType())
-      return GetCladArrayRefOfType(yType);
-    else
+  //   if (nonRefXValueType->isRealType())
+  //     return GetCladArrayRefOfType(yType);
+  //   else
```
vgvassilev, line 2825: Do we need this commented code here?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2883,12 +2820,18 @@ namespace clad {
...
       return GetCladArrayRefOfType(nonRefXValueType);
   }
 
+  clang::QualType ReverseModeVisitor::ComputeAdjointType(clang::QualType T) {
```
vgvassilev, line 2829: Can we add documentation describing why we remove the `const` qualification here?

test/Hessian/ArrayErrors.C
```diff
@@ -3,7 +3,7 @@
 #include "clad/Differentiator/Differentiator.h"
 
 //CHECK-NOT: {{.*error|warning|note:.*}}
-
+//XFAIL:*
```
vgvassilev, line 6: Why this starts to fail now? Worth a comment?

vgvassilev (commented):
Good work. Should we add a test for CUDA if that patch improves the current state?

test/Hessian/ArrayErrors.C
parth-07: This PR has made the change that the adjoint type will only depend on `x` value (`x` from `dy/dx` terminology). 

The hessian mode requires that all derivative parameter types are the same irrespective of the original parameter types. This worked well with the original approach, which used to compute adjoint types solely based on the type of the return value. An example of hessian mode derived function is as follows: 

```cpp
double fn(int a, double b, long double c) { ... }
```

Hessian derived function of `fn` will have the following signature:

```cpp
void fn_hessian(int a, double b, long double c, clad::array_ref<double> hessianMatrix) { ... }
```

Here please note that derived types need to be the same for all the parameters. The change in this PR breaks this requirement, and I have thus marked the test as `XFAIL`. We can easily fix this by adding special behaviour for hessian mode, but I think that change should be in a separate PR.

vgvassilev: I will merge this PR now, but please open an issue for this to be able to keep track of it.

vgvassilev (approved):
LGTM, modulo the comment.

PR #425, opened 2022-04-10
Add initial support for diff of ref return types in rev mode

parth-07:
This PR adds initial support for correctly differentiating function calls with reference return types. The main motivation for adding this support is many of the class operator overloads, such as, `=`, `+=`, `-=`, `*=` etc, naturally return a reference to the class object. 

The work done in this PR develops a base that would be extended upon for adding many other functionalities as well -- most notably, differentiation of function calls with pointer return types and maintaining the correct stack of pointer derivatives when pointers are passed to a call expression. 

## What does this PR solves?

C++ has the functionality to declare a variable as a reference (an alias) to an already existing variable, object or function. 

```cpp
double a = 11;
double& a_ref = a;
```

From the mathematical point of view, reference declaration is a no-operation. We are just defining a new name for an already existing variable. What I am trying to emphasize here is, unlike a normal variable declaration, a reference declaration should not have any corresponding reverse pass statements. With that being said, a reference variable declaration does impact the differentiation of further operations. 

For example: 

1) 

```cpp
someVar += a;
```

2) 

```cpp
someVar += a_ref;
```

Statements described in (1) and (2) are mathematically identical. Thus they should produce derived statements that have the same behaviour as well. In Clad, we solve this by effectively using the same derivative variable for both the original variable and the reference variable. To put things more concretely, please consider the following code snippet: 

```cpp
double& a_ref = a;
```

This statement produces the following statements in the derived function:

```cpp
// derivative declarations
double _d_a = 0;
double& _d_a_ref = _d_a;

// forward pass
double& a_ref = a;
````

Please note here that the derivative of `a_ref` is a reference variable pointing to the derivative of `a`. 

In the example that we just discussed we can easily point `_d_a_ref` to `_d_a` because the derivative of `a` is known at compile time. This is not always the case, for example, consider the following code:

```cpp
double& someFn(double& i, double&j, double& k) { ... }

double fn(double i, double j, double k) {
  double& ref = someFn(i, j, k);
}
```

We cannot determine which variable `ref` is referencing at compile time. Thus, we also cannot determine which derivative should `_d_ref` refer to.

This PR provides functionality to correctly point `_d_ref` to the derivative of the variable to which `ref` refers when this variable is not known at compile time.

## How does this PR solves this problem?

This PR solves the problem of correctly setting the derivative of a reference variable when the reference variable is being assigned to the result of a call expression by modifying the primal call expression such that it returns both the primal computation and the adjoint information. 

When Clad is differentiating a call to the function `someFn` that returns a reference then Clad generates a new function `someFn_forw` by transforming the original function `someFn` such that it takes adjoint information as input parameters and returns both the primal value and the adjoint information. For the remainder of this discussion, I will refer to this transformation mode as Reverse Mode Forward Pass mode. Please suggest a better and more intuitive name for this transformation mode. For example, consider the following function:

```cpp
double& someFn(double& i, double& j) {
  double& k = i;
  double& l = j;
  if (...)
    return k;
  else
    return l;
}
```

The corresponding `someFn_forw` will be as follows: 

```cpp
clad::ValueAndAdjoint<double&, double&>
someFn_forw(double& i, double& j, clad::array_ref<double> _d_i,
            clad::array_ref<double> _d_j) {
  double* _d_k = nullptr;
  double* _d_l = nullptr;

  // forward pass
  _d_k = &*_d_i;
  double& k = i;

  _d_l = &*_d_j;
  double& l = j;

  if (...)
    return {k, *_d_k};
  else
    return {l, *_d_l};
} 
```
Therefore, the following statement: 

```cpp
double& ref = someFn(i, j);
```

will produce the following statements in the derived function:

```cpp
// derivative declarations
double* _d_ref = nullptr;

// forward pass
double t0 = i;
double t1 = j;
clad::ValueAndAdjoint<double&, double&> t = someFn_forw(i, j, &_d_i, &_d_j);
_d_ref = &t.adjoint;
double& ref = t.value;

// reverse pass
someFn_pullback(t0, t1, /*pullback=*/double(), &_d_i, &_d_j);
```
Please note that pullback or `dfdx()` should be zero-tangent when the return type is a reference value. This will be discussed later in the next section.

### `ReverseModeForwPassVisitor`

Currently, I have created a new visitor class `ReverseModeForwPassVisitor` that inherits from `ReverseModeVisitor` and is responsible for creating forward pass functions (`_forw` functions). Please note that the transformation to generate a forward pass function only requires a tiny subset of functionalities of Reverse Mode. To be precise, required functionality of reverse mode are: correctly initialised variable declaration for derivative of each local variable, and forward pass should effectively trace the primal function. 

#### Benefits of inheriting from `ReverseModeVisitor`:

- Reverse mode forward pass includes all the functionalities that we require in the Reverse Mode Forward Pass transformed function. Sorry for the confusing terminology here. Thus, if we use the functionality of `ReverseModeVisitor`, discard the reverse pass and only provide the implementation of `Visit*` functions that are different in `ReverseModeForwPass` mode such as `VisitReturnStmt` then we can avoid too much code duplication. If we don't inherit from `ReverseModeVisitor` then we would need to provide the implementation for all of the `Visit*` functions. In most of these functions, we would simply be cloning the AST nodes.

#### Disadvantages of inheriting from `ReverseModeVisitor`:

- If we are reusing the functionality of `ReverseModeVisitor`, then we would not be able to generate Reverse Mode Forward Pass transformed functions of any function that cannot be differentiated in Reverse Mode, for example, due to the lack of support of some C++ construct. This is problematic because it is a lot easier to generate Reverse Mode Forward Pass transformed functions then it is to perform complete Reverse Mode differentiation, as most of the C++ features and constructs simply need to be cloned in Reverse Mode Forward Pass. 
- Reuse of `ReverseModeVisitor` implicitly assumes that forward pass of Reverse Mode derived function has effectively the same behaviour as the primal function. In the future, if we add active variable and data flow analysis in Reverse Mode, and we would be adding it soon, then this assumption would not be true. 
- As I earlier mentioned, in the future, we would require `_forw` functions for adding support of other functionalities as well. Many of these functionalities may require transformations that diverge from the functionalities included in Reverse Mode forward pass, and thus we may need to provide separate implementation of more and more `Visit*` functions.

### Derivative expressions to be used in Reverse Mode forward pass

Until now, we only needed to use derivative variables in the reverse pass. With the introduction of `_forw` functions, we need to use derivative variables in forward pass too. Since, `_forw` function takes adjoint information as well.
The problem here is, `Visit` function is designed to return `StmtDiff` that consists of a clone of the original expression and corresponding derivative if it exists. Sometimes, the clone contains `clad::push(...)` expression and derivative contain `clad::pop(...)` expression. In these cases, the derivative is designed to be used in the reverse pass only -- a notable example being array expressions in loops. Therefore, we need a routine that allows us to conveniently build derivative expressions that can be used in the forward pass. The building of such a routine will be non-trivial because `Visit*` function is tightly coupled with how derivative expressions are obtained.

### Pullback function signature of functions with reference return types

For a function `fn`: 
```cpp
double fn(double i, double j);
```

The pullback function should have the signature as follows: 

```cpp
void fn_pullback(double i, double j, double _d_y, clad::array_ref<double> _d_i, clad::array_ref<double> _d_j);
```

For functions with reference return types, the situation is slightly more complicated. 

`_d_y` or the pullback is used to initialise/correctly set the return value derivative. For example:

```cpp
return val;
```
Inside a pullback function, the following code gets differentiated to:

```cpp
// reverse pass
_d_val += _d_y;
```

Intuitively, this behaviour can be reasoned as follows: 

```cpp
double someFn(double& i, double& j) {
  ... statement 1 ...
  ... statement 2 ...
  ... and so on ... 
  return val;
}
  ...
  ...
  y = someFn(i, j);
  ...
  ...
}
```

If the return value of the function `someFn` is `val`, then the statement `y = someFn(i, j)` can effectively be visualised as follows: 

```cpp
  ...
  ...
  ... statement 1 ...
  ... statement 2 ...
  ... and so on ... 
  double y = val;
  ...
  ...
}
```

Now, if `y` is a reference variable, then `double& y = val` becomes a no-operation and should have no corresponding reverse mode derived statement. Therefore, ideally, pullback functions with reference return types do not have any corresponding pullback value. Again, sorry for the confusing terminology. We have two ways to proceed from here:

- Pullback function signature of functions with reference return types be same as the pullback function signature of functions with the corresponding non-reference return types. In this, a function call to pullback function of the functions with reference return types would need to be called with a dummy pullback (`_d_y`) value. The dummy value should be equal to zero-tangent vector of the return type. 

Or

- Pullback function signature of functions with reference return types be same as the pullback function signature of functions with `void` return types. This way, we would not need to pass any dummy value while calling pullback functions. On the downside, this makes rules of pullback function signature slightly more complicated. 

Please give your reviews and suggestions on the approach used, the problems and the design decisions discussed here.

### DIFF
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -70,7 +70,7 @@ namespace clad {
     friend class ReverseModeVisitor;
     friend class HessianModeVisitor;
     friend class JacobianModeVisitor;
-    friend class TransformSourceFnVisitor;
+    friend class ReverseModeForwPassVisitor;
     
     clang::Sema& m_Sema;
     plugin::CladPlugin& m_CladPlugin;
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -70,7 +70,8 @@ namespace clad {
     friend class ReverseModeVisitor;
     friend class HessianModeVisitor;
     friend class JacobianModeVisitor;
-
+    friend class TransformSourceFnVisitor;
+    
     clang::Sema& m_Sema;
     plugin::CladPlugin& m_CladPlugin;
     clang::ASTContext& m_Context;
--- a/include/clad/Differentiator/DiffMode.h
+++ b/include/clad/Differentiator/DiffMode.h
@@ -8,7 +8,7 @@ namespace clad {
     experimental_pushforward,
     experimental_pullback,
     reverse,
-    reverse_source_fn,
+    reverse_mode_forward_pass,
     hessian,
     jacobian,
     error_estimation
--- a/include/clad/Differentiator/DiffMode.h
+++ b/include/clad/Differentiator/DiffMode.h
@@ -8,6 +8,7 @@ namespace clad {
     experimental_pushforward,
     experimental_pullback,
     reverse,
+    reverse_source_fn,
     hessian,
     jacobian,
     error_estimation
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -32,6 +32,12 @@ extern "C" {
 }
 
 namespace clad {
+  template<typename T, typename U>
+  struct ValueAndAdjoint {
+    T value;
+    U adjoint;
+  };
+
   /// \returns the size of a c-style string
   CUDA_HOST_DEVICE unsigned int GetLength(const char* code) {
     unsigned int count;
--- a/include/clad/Differentiator/ReverseModeForwPassVisitor.h
+++ b/include/clad/Differentiator/ReverseModeForwPassVisitor.h
@@ -1,8 +1,8 @@
 #ifndef CLAD_TRANSFORM_SOURCE_FN_VISITOR_H
 #define CLAD_TRANSFORM_SOURCE_FN_VISITOR_H
 
-#include "clad/Differentiator/ReverseModeVisitor.h"
 #include "clad/Differentiator/ParseDiffArgsTypes.h"
+#include "clad/Differentiator/ReverseModeVisitor.h"
 
 #include "clang/AST/StmtVisitor.h"
 #include "clang/Sema/Sema.h"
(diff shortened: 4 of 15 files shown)


parth-07:
The work in this PR was rebased on top of master and merged as part of #601

PR #433, opened 2022-04-22
Add default definition for SetError.

grimmmyshini:
This is helpful as most people will not require explicitly overloading SetError with a definition different than the default.


vgvassilev (approved):
LGTM!

PR #434, opened 2022-04-23
Add abs call over default error estimates.

grimmmyshini:
This PR introduces the following changes:
- Build a `std::abs` call over the error estimates from the Taylor approx model.
- Two helper functions for users to build their own function calls.

### DIFF
--- a/include/clad/Differentiator/EstimationModel.h
+++ b/include/clad/Differentiator/EstimationModel.h
@@ -48,6 +48,39 @@ namespace clad {
     /// \param[in] VD The declaration to track.
     void AddVarToEstimate(clang::VarDecl* VD, clang::Expr* VDRef);
 
+    /// Build a function lookup expression that can be later resolved to build
+    /// an actual function call expression with arguments. Implemented as a
+    /// separate function to save multiple calls to function/namespace lookup.
+    ///
+    /// \param[in] funcName The name of the function to build the expression
+    /// for.
+    /// \param[in] nmspace The name of the namespace for the function,
+    /// currently does not support nested namespaces.
+    ///
+    /// \return The unresolved lookup expression that can be used to build
+    /// actual function call. This cannot be used directly to emit code, use
+    /// \c BuildFucntionCallExpr instead.
+    clang::Expr* GetFunctionCall(std::string funcName,
+                                 std::string nmspace = "");
+
+    /// Helper to build a function call expression.
+    /// \note This function does not check if the parameters result in correct
+    /// overload resolution, this is done to avoid the overhead of checking
+    /// if the correct overload exists given this function will be called
+    /// multiple times for the same parameters types. The user has to ensure
+    /// the parameters correspond to a valid overload of the function.
+    ///
+    /// \param[in] lookupRes The result from \c GetFunctionCall , this param
+    /// is used to build the final function call expression.
+    /// \param[in] callArgs A vector of \c clang::Expr of all the parameters
+    /// to the function call.
+    ///
+    /// \return The function call expression that can be used to emit into
+    /// code.
+    clang::Expr*
+    BuildFunctionCallExpr(clang::Expr* lookupRes,
+                          llvm::SmallVectorImpl<clang::Expr*>& callArgs);
+
     /// User overridden function to return the error expression of a
     /// specific estimation model. The error expression is returned in the form
     /// of a clang::Expr, the user may use BuildOp() to build the final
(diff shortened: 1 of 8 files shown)


include/clad/Differentiator/EstimationModel.h
```diff
@@ -48,6 +48,39 @@ namespace clad {
...
+    /// currently does not support nested namespaces.
+    ///
+    /// \return The unresolved lookup expression that can be used to build
+    /// actual function call. This cannot be used directly to emit code, use
+    /// \c BuildFucntionCallExpr instead.
+    clang::Expr* GetFunctionCall(std::string funcName,
+                                 std::string nmspace = "");
+
+    /// Helper to build a function call expression.
+    /// \note This function does not check if the parameters result in correct
```
vgvassilev, line 67: Did you actually try that? The code you wrote on `GetFunctionCall` seems to be pretty generic to handle overloads. Calling `ActOnCallExpr` with `UnresolvedLookupExpr` does the overload resolution. If the current code doesn't do that, I'd prefer to allow this behavior as I am not very concerned about the performance overhead of overload resolution.

grimmmyshini: I am not sure if I fully understand your question/comment. I just removed the explicit `noOverloadExists` check because it is very very unlikely that the user is passing incorrect parameters for overload resolution. And anyway it will be verified when we build the call expression/during compilation, so no need for an extra step that will be done anyway at a later point in time.

vgvassilev: @grimmmyshini, ok, I think we need to adjust the `\note` in the comment. My claim is that the implementation will do the overload resolution. Can you check?

grimmmyshini: Oh yes, it does, why do we have the 'NoOverloadExists' function then?

parth-07: `NoOverloadExists` function is used to ensure that an overload exists for a given unresolved expression and an argument list. It does not determine the overload itself.

grimmmyshini: It is not really required but I suppose you need it so that you can fail gracefully. In all other cases using an ActOnCallExpr is sufficient.

parth-07: > It is not really required but I suppose you need it so that you can fail gracefully.

Yes, exactly.

lib/Differentiator/EstimationModel.cpp
```diff
@@ -49,20 +52,64 @@ namespace clad {
   Expr* FPErrorEstimationModel::SetError(VarDecl* declStmt) { 
     return nullptr; 
   }
+  
+  Expr* FPErrorEstimationModel::GetFunctionCall(std::string funcName,
```
vgvassilev, line 56: Why not merging this function with `BuildFunctionCallExpr`. Is there a case in the codebase where we need to build an `UnresolvedLookup` without then calling `BuildFunctionCallExpr`?

lib/Differentiator/EstimationModel.cpp
```diff
@@ -49,20 +52,64 @@ namespace clad {
...
+    Expr* UnresolvedLookup = nullptr;
+    if (!R.empty())
+      UnresolvedLookup =
+          m_Sema.BuildDeclarationNameExpr(SS, R, /*ADL=*/false).get();
+    return UnresolvedLookup;
+  }
+
+  Expr* FPErrorEstimationModel::BuildFunctionCallExpr(
+      Expr* lookupRes, llvm::SmallVectorImpl<Expr*>& callArgs) {
+    if (!lookupRes)
```
vgvassilev, line 80: I'd prefer to turn this into an `assert` as this is a programmatic error if we don't reach here.

include/clad/Differentiator/EstimationModel.h
```diff
@@ -129,11 +157,15 @@ namespace clad {
 
   /// Example class for taylor series approximation based error estimation.
   class TaylorApprox : public FPErrorEstimationModel {
+    /// A variable to keep track of the function lookup expression for the
+    /// \c std::abs call
+    clang::Expr* absCallExpr = nullptr;
```
vgvassilev, line 162:
```suggestion
    clang::Expr* m_absCallExpr = nullptr;
```

vgvassilev (commented):
Thanks for working on this. Here are some suggestions.

lib/Differentiator/EstimationModel.cpp
grimmmyshini: Can you rephrase this? I don't think I understand...

vgvassilev: IIUC, the optimization you have in mind is reusing the `UnresolvedLookupExpr` built by this function. I am saying that we should not do this, as we should have an `UnresolvedLookupExpr` per CallExpr.

grimmmyshini: Okay, so each call expr should have its own unresolved lookup expr?

vgvassilev: Correct!

vgvassilev (approved):
LGTM!

PR #435, opened 2022-04-27
Fix differentiation of pointer arithmetic in forward mode

parth-07:
Integers used in pointer arithmetic should be considered
non-differentiable entities. For example:

```
p + i;
```

Derived statement should be:

```
_d_p + i;
```

instead of:

```
_d_p + _d_i;
```

Therefore, effective derived expression of `i` is `i` instead of `_d_i`.

### DIFF
--- a/lib/Differentiator/ForwardModeVisitor.cpp
+++ b/lib/Differentiator/ForwardModeVisitor.cpp
@@ -1214,6 +1214,52 @@ namespace clad {
     }
   }
 
+  /// Computes effective derivative operands. It should be used when operands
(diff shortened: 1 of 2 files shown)


vgvassilev:
@parth-07, can you clarify what `p` and `i` are? If they are a pointer and an offset, then I guess this would make sense.

vgvassilev:
> @parth-07, can you clarify what `p` and `i` are? If they are a pointer and an offset, then I guess this would make sense.

That was clarified over IM.

vgvassilev (approved):
LGTM!

PR #436, opened 2022-04-30
Add support for specifying custom derivatives of class functions

parth-07:
This PR adds support for specifying custom derivatives of class functions. It also solves a number of other issues (each in separate PR) that were required to enable differentiation of `std::accumulate`, `std::begin` and `std::end` without any custom derivative.

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -14,6 +14,8 @@ namespace custom_derivatives{}
 #include "clad/Differentiator/CladConfig.h"
 
 #include <cmath>
+#include <vector>
+
 namespace clad {
 template <typename T, typename U> struct ValueAndPushforward {
   T value;
@@ -80,6 +82,7 @@ pow_pullback(T1 x, T2 exponent, decltype(::std::pow(T1(), T2())) d_y,
   t = pow_pushforward(x, exponent, static_cast<T1>(0), static_cast<T2>(1));
   *d_exponent += t.pushforward * d_y;
 }
+
 } // namespace std
 // These are required because C variants of mathematical functions are
 // defined in global namespace.
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -14,7 +14,6 @@ namespace custom_derivatives{}
 #include "clad/Differentiator/CladConfig.h"
 
 #include <cmath>
-#include <vector>
 
 namespace clad {
 template <typename T, typename U> struct ValueAndPushforward {
--- a/include/clad/Differentiator/STLBuiltins.h
+++ b/include/clad/Differentiator/STLBuiltins.h
@@ -1,5 +1,7 @@
-#ifndef CLAD_STL_BUILTIN_DERIVATIVES_H
-#define CLAD_STL_BUILTIN_DERIVATIVES_H
+#ifndef CLAD_STL_BUILTINS_H
+#define CLAD_STL_BUILTINS_H
+
+#include <vector>
 
 namespace clad {
 namespace custom_derivatives {
@@ -28,4 +30,4 @@ void resize_pushforward(::std::vector<T>* v, unsigned sz, U val,
 } // namespace custom_derivatives
 } // namespace clad
 
-#endif
+#endif // CLAD_STL_BUILTINS_H
--- a/lib/Differentiator/ForwardModeVisitor.cpp
+++ b/lib/Differentiator/ForwardModeVisitor.cpp
@@ -791,9 +791,10 @@ namespace clad {
   StmtDiff
   ForwardModeVisitor::VisitArraySubscriptExpr(const ArraySubscriptExpr* ASE) {
     auto ASI = SplitArraySubscript(ASE);
-    const Expr* Base = ASI.first;
+    const Expr* base = ASI.first;
     const auto& Indices = ASI.second;
-    Expr* clonedBase = Clone(Base);
+    StmtDiff cloneDiff = Visit(base);
+    Expr* clonedBase = cloneDiff.getExpr();
     llvm::SmallVector<Expr*, 4> clonedIndices(Indices.size());
     std::transform(std::begin(Indices),
                    std::end(Indices),
@@ -820,7 +821,7 @@ namespace clad {
         VD = decl;
       }
     } else if (isa<MemberExpr>(clonedBase->IgnoreParenImpCasts())) {
-      auto derivedME = Visit(clonedBase).getExpr_dx();
+      auto derivedME = cloneDiff.getExpr_dx();
       if (!isa<MemberExpr>(derivedME->IgnoreParenImpCasts())) {
         return {cloned, zero};
       }
@@ -1013,12 +1014,11 @@ namespace clad {
         utils::BuildNNS(m_Sema, originalFnDC, SS);
         DC = originalFnDC;
       } else {
-        if (isa<RecordDecl>(originalFnDC)) {
+        if (isa<RecordDecl>(originalFnDC))
           DC = utils::LookupNSD(m_Sema, "class_functions",
                                 /*shouldExist=*/false, NSD);
-        } else {
+        else
           DC = utils::FindDeclContext(m_Sema, NSD, originalFnDC);
-        }
         if (DC)
           utils::BuildNNS(m_Sema, DC, SS);
       }
--- a/lib/Differentiator/ForwardModeVisitor.cpp
+++ b/lib/Differentiator/ForwardModeVisitor.cpp
@@ -2011,4 +2011,20 @@ namespace clad {
             .get();
     return {clonedE, derivedE};
   }
+
+  StmtDiff ForwardModeVisitor::VisitCXXFunctionalCastExpr(
+      const clang::CXXFunctionalCastExpr* FCE) {
+    StmtDiff castExprDiff = Visit(FCE->getSubExpr());
+    Expr* clonedFCE = m_Sema
+                          .BuildCXXFunctionalCastExpr(
+                              FCE->getTypeInfoAsWritten(), FCE->getType(),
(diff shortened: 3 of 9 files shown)


lib/Differentiator/CladUtils.cpp
```diff
@@ -411,5 +411,17 @@ namespace clad {
     bool IsRValue(const clang::Expr* E) {
       return E->isRValue() || E->isXValue();
     }
+
+    clang::DeclContext* GetInnermostNamespaceOrTUnitDC(clang::Sema& semaRef,
```
vgvassilev, line 415:
```suggestion
    clang::DeclContext* GetEnclosingNamespaceOrTUnitDC(clang::Sema& semaRef,
```

Can't we use [DeclContext::getEnclosingNamespaceContext](https://clang.llvm.org/doxygen/classclang_1_1DeclContext.html#aa838660af381c6967722c6552580c2a2)?

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -14,6 +14,8 @@ namespace custom_derivatives{}
 #include "clad/Differentiator/CladConfig.h"
 
 #include <cmath>
+#include <vector>
```
vgvassilev, line 17: That will be likely problematic. Maybe we should find a way to include this headers somehow on demand...

lib/Differentiator/CladUtils.cpp
parth-07: We can, I totally forgot about this function. I will modify to code to use this function instead.

include/clad/Differentiator/BuiltinDerivatives.h
parth-07: Should we make a different custom derivatives file for custom derivatives of `std::` namespace?

parth-07: This way, users can include this file if they require it.

parth-07: > That way we can detect if the TU has std::vector or other files and then just include their custom derivatives respectively.

What if the user has included `<vector>` and other `std::` libraries but does not want to include the custom derivatives for `std::` library.

lib/Differentiator/CladUtils.cpp
parth-07: I have slightly modified the custom derivative for class function design such that we no longer require this function. 

Now custom derivatives for class functions are always searched in the namespace: `clad::custom_derivatives::class_functions`. It does not take into account the namespace in which the class is defined. 
We can safely ignore the namespace in which class is defined because for each instance member function the first argument is a pointer to a class object that encodes class namespace information as well. For example: 

```cpp
namespace class_functions {

template <typename T>
void clear_pushforward(::std::vector<T>* v, ::std::vector<T>* d_v) {
  d_v->clear();
  v->clear();
}
```
The main motivation behind this change is that in libc++ vector is defined in `::std::__1` namespace, and in libstdc++ vector is defined in `::std` namespace. 

Please give your suggestion on this change.

include/clad/Differentiator/BuiltinDerivatives.h
vgvassilev: I'd think that'd be a rare case and maybe we can control it via a macro outside the header include (inverting the macro from my previous suggestion).

lib/Differentiator/CladUtils.cpp
vgvassilev: Yes, that makes sense to me. Why do we synthesize a pointer (eg `::std::vector<T>* v`) rather than a reference (`::std::vector<T>& v`)?

parth-07: It is to be more consistent with how C++ passes objects implicitly to member functions in the form of `this` pointer. 

We can easily change the requirements to pass by reference as well.

vgvassilev: We can make this change outside of this PR. I'd think it looks more natural to use `.` operators. That'd be expert feature but still probably better.

parth-07: In that case, should we ideally pass derivative object by reference as well?

vgvassilev: Absolutely.

include/clad/Differentiator/BuiltinDerivatives.h
parth-07: > Maybe we should require the Clad custom derivatives header file to be included after all header files?

I am concerned such a condition would be difficult to effectively impose and may lead to confusing behaviour for users. 

> That way we can detect if the TU has std::vector or other files and then just include their custom derivatives respectively

It be really nice if we can automatically include the required derivatives but from what I know, clang plugins do not have much control over preprocessing, and from what I understand, automatically including required custom derivatives would require controlling preprocessing of the source file by Clad.

lib/Differentiator/CladUtils.cpp
parth-07: In that case, I am concerned that users would intuitively assume incorrect type of derivative functions. Since `this` is commonly assumed to be a pointer, people may expect derivative of `this` to be a pointer as well.

include/clad/Differentiator/BuiltinDerivatives.h
vgvassilev: Fair point. Let's defer this discussion as we don't seem to have a reasonable way out. I'd propose to create a file called something like `STLBuiltinDerivatives.h` and include it unconditionally from within `BuiltinDerivatives.h`. We can decide how to optimize this on a later stage.

parth-07: Okay.

parth-07: I have added `STLBuiltinDerivatives.h` header file, it contains custom derivatives of `std::vector`. Should we change the directory design to have header files for custom derivatives as follows: 
```
#include "clad/Differentiator/BuiltinDerivatives/STL.h"
#include "clad/Differentiator/BuiltinDerivatives/Eigen.h"
#include "clad/Differentiator/BuiltinDerivatives/basic.h"
#include "clad/Differentiator/BuiltinDerivatives/all.h"
```

lib/Differentiator/CladUtils.cpp
vgvassilev: Can you provide an example for it?

include/clad/Differentiator/BuiltinDerivatives.h
vgvassilev: Maybe `#include "clad/Differentiator/STLBuiltins.h`, etc?

lib/Differentiator/CladUtils.cpp
parth-07: Yes. 
Please consider the following code: 
```cpp
struct Complex {
  double real, im;
  double product() {
    return real*im;
  }
};
```
The member function `double product()` can also be written as: 

```cpp
double product() {
  return this->real * this->im;
}
```

Intuitively, the object is passed to a member function as an argument `this` -- that is of type pointer to the corresponding class of the member function. Therefore, since we are following that the derivative type is the same as of primal type, users may intuitively assume that the derivative of `this` argument is of the same type as of `this` as well -- pointer to the class.

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -1016,9 +1014,14 @@ namespace clad {
         utils::BuildNNS(m_Sema, originalFnDC, SS);
         DC = originalFnDC;
       } else {
-        utils::BuildNNS(m_Sema, NSD, SS);
-        utils::BuildNNS(m_Sema, originalFnDC, SS);
-        DC = utils::FindDeclContext(m_Sema, NSD, originalFnDC);
+        if (isa<RecordDecl>(originalFnDC)) {
```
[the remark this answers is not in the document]
parth-07: Okay, I have removed the braces from here.

include/clad/Differentiator/BuiltinDerivatives.h
parth-07: Can you please elaborate on where should we put this header file?

vgvassilev: In STLBuiltins.h.

parth-07: Oh, I thought I already moved this include to `STLBuiltins.h`. I have now moved this include to `STLBuiltins.h`.

include/clad/Differentiator/STLBuiltins.h
```diff
@@ -0,0 +1,33 @@
+#ifndef CLAD_STL_BUILTIN_DERIVATIVES_H
+#define CLAD_STL_BUILTIN_DERIVATIVES_H
```
vgvassilev, line 2:
```suggestion
#ifndef CLAD_STL_BUILTINS_H
#define CLAD_STL_BUILTINS_H
```

include/clad/Differentiator/STLBuiltins.h
```diff
@@ -0,0 +1,33 @@
...
+void resize_pushforward(::std::vector<T>* v, unsigned sz, U val,
+                        ::std::vector<T>* d_v, unsigned d_sz, U d_val) {
+  d_v->resize(sz, d_val);
+  v->resize(sz, val);
+}
+} // namespace class_functions
+} // namespace custom_derivatives
+} // namespace clad
+
+#endif
```
vgvassilev, line 33:
```suggestion
#endif // CLAD_STL_BUILTINS_H

```

vgvassilev (approved):
LGTM modulo the suggestions.

lib/Differentiator/CladUtils.cpp
vgvassilev: Thanks for the example. I don't think that should be a problem. I guess the same arguments hold for defining a copy constructor and yet people pass by reference.

vgvassilev:
I am not sure why codecov is unhappy - maybe it was a glitch.

Issue #442, opened 2022-05-01
Calls to overloaded derived functions defined inside a namespace fails while linking


parth-07:
Reproducible example:
```cpp
namespace A {
  double someFn_1(double& i, double j, double k) {
    i = j;
    return 1;
  } 

  double someFn_1(double& i, double j) {
    someFn_1(i, j, j);
    return 2;
  }

  double someFn(double& i, double& j) {
    someFn_1(i, j);
    return 3;
  }
}

double fn(double i, double j) {
  A::someFn(i, j);
  return i + j;
}

int main() {
  auto d_fn = clad::differentiate(fn, "j");
  std::cout<<d_fn.execute(3, 5)<<"\n";
}
```

Output: 

```
/usr/bin/ld: /tmp/bug-forw-overloaded-fns-2f0bef.o: in function `A::someFn_pushforward(double&, double&, double&, double&)':
/home/parth/clubs/open-source/clad/tutorial/bug-forw-overloaded-fns.cpp:16: undefined reference to `A::someFn_1_pushforward(double&, double, double&, double)'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
````

PR #443, opened 2022-05-01
Modify Clad to call HandleTopLevelDecl on fns inside namespaces

parth-07:
This PR modifies Clad such that it calls `m_CI.getASTConsumer().HandleTopLevelDecl(...)` for (generated) declarations that are contained in a namespace. Until now we were only calling it for top-level declarations -- declarations contained directly in `TranslationUnitDecl`. This modification seems to fix the issue #442. Please confirm if the change is correct, as it is kind of counter-intuitive to the `ASTConsumer::HandleTopLevelDecl(...)` function name. 

On a related note, why do we return declaration context with `DerivativeBuilder::Derive`, `ForwardModeVisitor::Derive`, `ReverseModeVisitor::Derive`, ....? If the returned declaration context is the derivative function context then it can easily be obtained by `derivative->getDeclContext()`. The current value of declaration context obtained by these derive functions corresponds to the first namespace in sequence if the derivative function is in a namespace, otherwise, it is nullptr. I am not sure if this is useful, and we do not seem to be depending on it anywhere either.


tools/ClangPlugin.cpp
```diff
@@ -241,14 +242,15 @@ namespace clad {
...
-                          ->isTranslationUnit();
-          if (isTU) {
-            ProcessTopLevelDecl(DerivativeDeclOrEnclosingContext);
+          // Call CodeGen only if the produced Decl is a top-most decl or is
+          // contained in a namespace decl.
+          DeclContext* derivativeDC = DerivativeDecl->getDeclContext();
+          bool isTUorND =
+              derivativeDC->isTranslationUnit() || derivativeDC->isNamespace();
+          // FIXME: Should we call `Sema::MarkFunctionReferenced` on derived
+          // functions? Everything seems to be working fine without calling it.
```
vgvassilev, line 251: What does clang do in that case? Maybe copy the generated code from the modified test case and debug it?

parth-07: Clang calls `Sema::MarkFunctionReferenced` when a function declaration (or any other declaration) `DeclRefExpr` node is created. `Sema::BuildDeclRefExpr` implicilty calls `Sema::MarkAnyDeclReferenced`. Therefore, we do not need to manually call it.

parth-07: I have removed the `FIXME:` comment now.

vgvassilev: > Clang calls `Sema::MarkFunctionReferenced` when a function declaration (or any other declaration) `DeclRefExpr` node is created. `Sema::BuildDeclRefExpr` implicilty calls `Sema::MarkAnyDeclReferenced`. Therefore, we do not need to manually call it.

But then does it call `HandleTopLevelDecl` on a declaration in a namespace?

vgvassilev: I think we should pursue the same behavior. I think we should call MarkReferenced instead?

parth-07: Using `Sema::MarkFunctionReferenced` does not resolve the issue #442 .

I did more research into the matter. Clang calls `clang::BackendConsumer::HandleTopLevelDecl` on each top level declaration, this in turn calls `clang::CodeGeneratorImpl::HandleTopLevelDecl`, this in turn calls `EmitTopLevelDecl`. For namespaces, this function calls `EmitDeclContext` which in turn calls `EmitTopLevelDecl` for each declaration inside the namespace. 

Therefore to pursue the same behaviour as Clang, we need to call `EmitTopLevelDecl` on the derived functions inside a namespace. I don't think this function is publicly accessible, what should we do in this situation?

We cannot call `ASTConsumer::HandleTopLevelDecl` for the namespace because it would emit each original declaration inside the namespace multiple times.

vgvassilev: I think we should document that finding and and keep your solution then.

parth-07: I have added a comment documenting the reason for calling `HandleTopLevel` on declarations inside a namespace. 
I have also called `MarkFunctionReferenced` for derived functions.

tools/ClangPlugin.cpp
```diff
@@ -241,14 +242,29 @@ namespace clad {
...
+          // `BackendConsumer::HandleTopLevelDecl`.
+          // `BackendConsumer::HandleTopLevelDecl` emits LLVM IR of each
+          // declaration inside the namespace using CodeGen. We need to manually
+          // call `HandleTopLevelDecl` for each new declaration added to a
+          // namespace because `HandleTopLevelDecl` has already been called for
+          // a namespace by Clang when the namespace is parsed.
+
+          // Call CodeGen only if
+          // the produced Decl is a top-most decl or is contained in a namespace
+          // decl.
```
vgvassilev, line 262:
```suggestion
          // Call CodeGen only if the produced Decl is a top-most
          // decl or is contained in a namespace decl.
```

vgvassilev (approved):
LGTM!

PR #445, opened 2022-05-06
Enable differentiation wrt data members of parameters in forw mode

parth-07:
This PR adds initial support for differentiation with respect to data members in Forward Mode AD. Functions can be differentiated with respect to a data member by specifying the data member in the differentiation args. For example: 

```cpp
double fn(std::pair<double, double> u, std::pair<double, double> v) {
  return u.first + v.second;
}

auto d_fn = clad::differentiate(fn, "u.first"); 
```

This PR does not add support for differentiating with respect to array data members. 

This PR also fixes the test invocation of `FirstDerivative/Recursive` and `ROOT/Interface` tests. These tests were using clashing names for output file and that was causing failed tests due to race conditions.

### DIFF
--- a/include/clad/Differentiator/ParseDiffArgsTypes.h
+++ b/include/clad/Differentiator/ParseDiffArgsTypes.h
@@ -6,7 +6,9 @@
 #define CLAD_PARSE_DIFF_ARGS_TYPES_H
 
 #include "clang/AST/Decl.h"
+#include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/StringRef.h"
 
 #include <cstddef>
 #include <utility>
@@ -34,10 +36,51 @@ namespace clad {
       return Start == rhs.Start && Finish == rhs.Finish;
     }
   };
+  
+  using IndexIntervalTable = llvm::SmallVector<IndexInterval, 16>;
+
+  /// `DiffInputVarInfo` is designed to store all the essential information about a
+  /// differentiation input variable. Please note that here input variable corresponds
+  /// to mathematical variable, not a programming one. 
+  // FIXME: 'DiffInputVarInfo' name is probably not accurate, since we can have multiple
+  // differentiation input variables for same parameter as well. 'DiffInputVarInfo' 
+  // name implicitly guides that there would be at most one `DiffInputVarInfo` object for
+  // one parameter, but that is not strictly true.
+  struct DiffInputVarInfo {
+    /// Source string specified by user that defines differentiation
+    /// specification for the input variable.
+    /// For example, if complete input string specified by user is:
+    /// 'u, v.first, arr[3]'
+    /// then `source` data member value for 2nd input variable should be
+    /// 'v.first'
+    std::string source;
+    /// Parameter associated with the input variable.
+    const clang::ValueDecl* param = nullptr;
+    /// array index range associated with the parameter.
+    IndexInterval paramIndexInterval;
+    /// Nested field information.
+    llvm::SmallVector<std::string, 4> fields;
+    // FIXME: Add support for differentiating with respect to array fields.
+    // llvm::SmallVector<IndexInterval> fieldIndexIntervals;
+
+    DiffInputVarInfo(const clang::ValueDecl* pParam = nullptr,
+                     IndexInterval pParamIndexInterval = {},
+                     llvm::SmallVector<std::string, 4> pFields = {})
+        : param(pParam), paramIndexInterval(pParamIndexInterval),
+          fields(pFields) {}
+
+    // FIXME: Move function definitions to ParseDiffArgTypes.cpp
+    bool operator==(const DiffInputVarInfo& rhs) const {
+      return param == rhs.param &&
+             paramIndexInterval == rhs.paramIndexInterval &&
+             fields == rhs.fields;
+    }
+  };
+
+  using DiffInputVarsInfo = llvm::SmallVector<DiffInputVarInfo, 16>;
 
   using DiffParams = llvm::SmallVector<const clang::ValueDecl*, 16>;
-  using IndexIntervalTable = llvm::SmallVector<IndexInterval, 16>;
   using DiffParamsWithIndices = std::pair<DiffParams, IndexIntervalTable>;
-} // namespace clad
+  } // namespace clad
 
 #endif
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -420,5 +420,65 @@ namespace clad {
       else if (S)
         block.push_back(S);
     }
+    
+    MemberExpr*
+    BuildMemberExpr(clang::Sema& semaRef, clang::Scope* S, clang::Expr* base,
+                    llvm::ArrayRef<clang::StringRef> fields) {
+      MemberExpr* ME = nullptr;
+      for (auto field : fields) {
+        ME = BuildMemberExpr(semaRef, S, base, field);
+        base = ME;
+      }
+      return ME;
+    }
+
+    clang::FieldDecl* LookupDataMember(clang::Sema& semaRef, clang::RecordDecl* RD,
+                          llvm::StringRef name) {
+      LookupResult R(semaRef, BuildDeclarationNameInfo(semaRef, name),
+                     Sema::LookupNameKind::LookupMemberName);
+      CXXScopeSpec CSS;
+      semaRef.LookupQualifiedName(R, RD, CSS);
+      if (R.empty())
+        return nullptr;
+      assert(R.isSingleResult() && "Lookup in valid classes should always "
+                                   "return a single data member result.");
+      auto D = R.getFoundDecl();
+      // We are looking data members only!
+      if (auto FD = dyn_cast<FieldDecl>(D))
+        return FD;
+      return nullptr;
+    }
+
+    bool IsValidMemExprPath(clang::Sema& semaRef, clang::RecordDecl* RD,
+                     llvm::ArrayRef<llvm::StringRef> fields) {
+      for (std::size_t i = 0; i < fields.size(); ++i) {
+        FieldDecl* FD = LookupDataMember(semaRef, RD, fields[i]);
+        if (!FD)
+          return false;
+        if (FD->getType()->isRecordType())
+          RD = FD->getType()->getAsCXXRecordDecl();
+        // Current Field declaration is not of record type, therefore
+        // it cannot have any field within it. And any member access
+        // ('.') expression would be an invalid path.
+        else if (i != fields.size() - 1)
+          return false;
+      }
+      return true;
+    }
+
+    clang::QualType
+    ComputeMemExprPathType(clang::Sema& semaRef, clang::RecordDecl* RD,
+                           llvm::ArrayRef<llvm::StringRef> fields) {
+      assert(IsValidMemExprPath(semaRef, RD, fields) &&
+             "Invalid field path specified!");
+      QualType T = RD->getTypeForDecl()->getCanonicalTypeInternal();
+      for (auto field : fields) {
+        auto FD = LookupDataMember(semaRef, RD, field);
+        if (FD->getType()->isRecordType())
+          RD = FD->getType()->getAsCXXRecordDecl();
+        T = FD->getType();
+      }
+      return T;
+    }
   } // namespace utils
 } // namespace clad
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -263,12 +263,12 @@ namespace clad {
   }
 
   void DiffRequest::UpdateDiffParamsInfo(Sema& semaRef) {
+    DVI.clear();
     auto& C = semaRef.getASTContext();
     const Expr* diffArgs = Args;
     const FunctionDecl* FD = Function;
     FD = FD->getDefinition();
     if (!diffArgs || !FD) {
-      DiffParamsInfo = {{}, {}};
       return;
     }
     DiffParams params{};
@@ -283,11 +283,11 @@ namespace clad {
         return;
       }
       // Split the string by ',' characters, trim whitespaces.
-      llvm::SmallVector<llvm::StringRef, 16> names{};
-      llvm::StringRef name{};
+      llvm::SmallVector<llvm::StringRef, 16> diffParamsSpec{};
       do {
-        std::tie(name, string) = string.split(',');
-        names.push_back(name.trim());
+        llvm::StringRef pInfo{};
+        std::tie(pInfo, string) = string.split(',');
+        diffParamsSpec.push_back(pInfo.trim());
       } while (!string.empty());
       // Stores parameters and field declarations to be used as candidates for
       // independent arguments.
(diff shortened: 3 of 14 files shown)


include/clad/Differentiator/ParseDiffArgsTypes.h
```diff
@@ -7,6 +7,8 @@
 
 #include "clang/AST/Decl.h"
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/StringRef.h"
```
vgvassilev, line 11:
```suggestion
#include "llvm/ADT/ArrayRef.h"
#include "llvm/ADT/SmallVector.h"
#include "llvm/ADT/StringRef.h"
```

include/clad/Differentiator/ParseDiffArgsTypes.h
```diff
@@ -34,9 +36,55 @@ namespace clad {
...
+    const clang::ValueDecl* param = nullptr;
+    /// array index range associated with the parameter.
+    IndexInterval paramIndexInterval;
+    /// Nested field information. 
+    llvm::SmallVector<std::string, 4> fields;
+    // FIXME: Add support for differentiating with respect to array fields.
+    // llvm::SmallVector<IndexInterval> fieldIndexIntervals;
+
+    DiffInputVarInfo(const clang::ValueDecl* pParam = nullptr,
+                  IndexInterval pParamIndexInterval = {},
```
vgvassilev, line 67: Did you run clang-format?

include/clad/Differentiator/ParseDiffArgsTypes.h
```diff
@@ -34,9 +36,55 @@ namespace clad {
...
+          fields(pFields) {}
+
+    // FIXME: Move function definitions to ParseDiffArgTypes.cpp          
+    bool operator==(const DiffInputVarInfo& rhs) const {
+      return param == rhs.param &&
+             paramIndexInterval == rhs.paramIndexInterval &&
+             fields == rhs.fields;
+    }
+
+    llvm::SmallVector<llvm::StringRef, 4> GetFieldsRef() {
```
vgvassilev, line 79: We should not return vectors, can we move this as an out parameter?

lib/Differentiator/CladUtils.cpp
```diff
@@ -411,5 +411,62 @@ namespace clad {
...
+      // We are looking data members only!
+      if (auto FD = dyn_cast<FieldDecl>(D))
+        return FD;
+      return nullptr;
+    }
+
+    bool IsValidMemExprPath(clang::Sema& semaRef, clang::RecordDecl* RD,
+                     llvm::ArrayRef<llvm::StringRef> fields) {
+      for (std::size_t i = 0; i < fields.size(); ++i) {
+        auto FD = LookupDataMember(semaRef, RD, fields[i]);
```
vgvassilev, line 446: Let's expand the type here instead of using `auto`.

lib/Differentiator/CladUtils.cpp
```diff
@@ -411,5 +411,62 @@ namespace clad {
...
+        if (!FD)
+          return false;
+        if (FD->getType()->isRecordType())
+          RD = FD->getType()->getAsCXXRecordDecl();
+        else if (i != fields.size() - 1)
```
vgvassilev, line 451: We need a comment describing this branch.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -377,23 +389,56 @@ namespace clad {
...
+        }
+
+        std::size_t dotIdx = diffSpec.find(".");
+        while (dotIdx != llvm::StringRef::npos) {
+          std::size_t fieldEndIdx = diffSpec.find(".", dotIdx+1);
+          llvm::StringRef fieldName =
+              diffSpec.substr(dotIdx + 1, fieldEndIdx - dotIdx - 1);
+          dVarInfo.fields.push_back(fieldName.str());
+          dotIdx = fieldEndIdx;
+        }
```
vgvassilev, line 408: Maybe we can simplify this by:

```cpp
while (!dotIdx.empty()) {
    StringRef Arg;
    std::tie(Arg, Env) = dotIdx.split(' ');
    dVarInfo.fields.push_back(Arg.str());
 }
```

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -367,6 +389,16 @@ namespace clad {
                                      dParam);
       addToCurrentBlock(BuildDeclStmt(dParamDecl));
       dParam = BuildDeclRef(dParamDecl);
+      if (dParamType->isRecordType() && param == m_IndependentVar) {;
```
vgvassilev, line 392:
```suggestion
      if (dParamType->isRecordType() && param == m_IndependentVar) {
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -240,8 +240,11 @@ namespace clad {
     assert(m_Function && "Must not be null.");
 
     DiffParams args{};
+    DiffInputVarsInfo DVI;
     if (request.Args) {
-      std::tie(args, std::ignore) = request.DiffParamsInfo;
+      DVI = request.DVI;
+      for (auto dParam : DVI)
+        args.push_back(dParam.param);
```
vgvassilev, line 247: That code seems duplicated in `HessianModeVisitor.cpp` as well. We should add another point to our refactoring plans.

vgvassilev (commented):
This looks very good! Here are some minor comments.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: Okay, I will open an issue that keeps track of the refactoring details.

lib/Differentiator/DiffPlanner.cpp
parth-07: I have replaced the current code with the implementation that uses the `split` function. Final code is this: 

```cpp
std::size_t dotIdx = diffSpec.find(".");
dotIdx += (dotIdx != StringRef::npos);
StringRef fieldsSpec = diffSpec.substr(dotIdx);
while (!fieldsSpec.empty()) {
  StringRef fieldName;
  std::tie(fieldName, fieldsSpec) = fieldsSpec.split('.');
  dVarInfo.fields.push_back(fieldName.str());
}
```

include/clad/Differentiator/ParseDiffArgsTypes.h
parth-07: Yes, we can, but that will defeat the purpose of this function. 

I have created some functions that takes `llvm::ArrayRef<llvm::StringRef>` as an input. With `GetFieldsRef`, we can call those functions as follows: 

```cpp
bool isValid =
              utils::IsValidMemExprPath(semaRef, RD, dVarInfo.GetFieldsRef());
```

Without this function, this same call will be as follows:

```cpp
llvm::ArrayRef<llvm::StringRef> ref(dVarInfo.fields.begin(), dVarInfo.fields.end());
bool isValid = utils::IsValidMemExprPath(semaRef, RD, ref);
```
If we move the return value to an out parameter, then same call will be as follows:

```cpp
llvm::SmallVector<llvm::StringRef> ref;
dVarInfo.GetFieldsRef(ref);
bool isValid = utils::IsValidMemExprPath(semaRef, RD, ref);
```

What should we do in this case?

include/clad/Differentiator/ParseDiffArgsTypes.h
parth-07: Yes, this code was not formatted. I have now formatted this file.

lib/Differentiator/DiffPlanner.cpp
vgvassilev: Why did you need `StringRef fieldsSpec = diffSpec.substr(dotIdx);`?

parth-07: Because if the differentiation specification is `"obj.mem1.mem2.mem3"`, then `fieldsSpec` should be `"mem1.mem2.mem3"`. Parameter information is stored before this code segment.

include/clad/Differentiator/ParseDiffArgsTypes.h
parth-07: I cannot seem to find a constructor of `llvm::MakeArrayRef` that takes an iterator range as an input. There's a constructor that takes pointer range, but `fields` is an array of `std::string` whereas we want `llvm::ArrayRef` of `llvm::StringRef` elements. Therefore, `MakeArrayRef` constructor that takes range of `std::string` iterator builds `llvm::ArrayRef<std::string>` object instead of `llvm::ArrayRef<llvm::StringRef>`

vgvassilev: Ok, then we should probably go with:
```cpp
llvm::SmallVector<llvm::StringRef> ref;
dVarInfo.GetFieldsRef(ref);
bool isValid = utils::IsValidMemExprPath(semaRef, RD, ref);
```

lib/Differentiator/DiffPlanner.cpp
vgvassilev: Hm.. okay..

include/clad/Differentiator/ParseDiffArgsTypes.h
parth-07: Can we please instead go with this: 

```cpp
llvm::ArrayRef<llvm::StringRef> ref(dVarInfo.fields.begin(), dVarInfo.fields.end());
bool isValid = utils::IsValidMemExprPath(semaRef, RD, ref);
```

It uses one line less.

vgvassilev: sure, that works.

parth-07: Done, I have removed the `GetFieldsRef` function now.

lib/Differentiator/CladUtils.cpp
```diff
@@ -411,5 +411,65 @@ namespace clad {
...
+      return ME;
+    }
+
+    clang::FieldDecl* LookupDataMember(clang::Sema& semaRef, clang::RecordDecl* RD,
+                          llvm::StringRef name) {
+      LookupResult R(semaRef, BuildDeclarationNameInfo(semaRef, name),
+                     Sema::LookupNameKind::LookupMemberName);
+      CXXScopeSpec CSS;
+      semaRef.LookupQualifiedName(R, RD, CSS);
+      if (R.empty())
```
vgvassilev, line 432: I am not sure about the semantics/reachability of this function. Can we look up something that does not exist? Eg, isn't it, at that point, a programmer mistake since we have done the validation if a field exists earlier? If that's the case we should turn this into an assert, else we should write a test to appease codecov.

test/FirstDerivative/DiffInterface.C
```diff
@@ -133,5 +149,8 @@ int main () {
   clad::differentiate(f_2, ""); // expected-error {{No parameters were provided}}
   clad::differentiate(fn_with_no_params); // expected-error {{Attempted to differentiate a function without parameters}}
 
+  clad::differentiate(f_2, "x.mem1");                             // expected-error {{Fields can only be provided for class type parameters. Field information is incorrectly specified in 'x.mem1' for a non-class type parameter 'x'}}
```
vgvassilev, line 152:
```suggestion
  clad::differentiate(f_2, "x.mem1");                             // expected-error {{Fields can only be provided for class type parameters. Field information is incorrectly specified in 'x.mem1' for non-class type parameter 'x'}}
```

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -242,14 +250,26 @@ namespace clad {
...
+      if (auto RD = diffVarInfo.param->getType()->getAsCXXRecordDecl()) {
+        llvm::SmallVector<llvm::StringRef, 4> ref(diffVarInfo.fields.begin(),
+                                                  diffVarInfo.fields.end());
+        T = utils::ComputeMemExprPathType(m_Sema, RD, ref);
+        isField = true;
+      }
+      if (!IsRealNonReferenceType(T)) {
+        diag(DiagnosticsEngine::Error, request.Args->getEndLoc(),
+             "Attempted differentiation w.r.t. %0 '%1' which is not "
+             "of a real type.",
```
vgvassilev, line 268:
```suggestion
             "of real type",
```

vgvassilev (commented):
Mostly details.

lib/Differentiator/CladUtils.cpp
parth-07: This case is reachable. User can provide "c.mem1.mem2.mem3" as differentiation specification. `LookupDataMember` function ensures that `mem1`, `mem2`, and `mem3` fields indeed exist, if not, then Clad gives "invalid path specified" error.

lib/Differentiator/CladUtils.cpp
```diff
@@ -411,5 +411,65 @@ namespace clad {
...
+        return nullptr;
+      assert(R.isSingleResult() && "Lookup in valid classes should always "
+                                   "return a single data member result.");
+      auto D = R.getFoundDecl();
+      // We are looking data members only!
+      if (auto FD = dyn_cast<FieldDecl>(D))
+        return FD;
+      return nullptr;
```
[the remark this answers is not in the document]
parth-07: I think we can write a test for this by using a member function name where a data member name is expected. I will try to write a test for this.

lib/Differentiator/CladUtils.cpp
```diff
@@ -420,5 +420,65 @@ namespace clad {
...
+
+    clang::FieldDecl* LookupDataMember(clang::Sema& semaRef, clang::RecordDecl* RD,
+                          llvm::StringRef name) {
+      LookupResult R(semaRef, BuildDeclarationNameInfo(semaRef, name),
+                     Sema::LookupNameKind::LookupMemberName);
+      CXXScopeSpec CSS;
+      semaRef.LookupQualifiedName(R, RD, CSS);
+      if (R.empty())
+        return nullptr;
+      assert(R.isSingleResult() && "Lookup in valid classes should always "
```
vgvassilev, line 443: Can we write a test where we have a base and child class with the same field name:
```cpp
struct A {
    int mem1;
};

struct B: public A {
   int mem1;
};```

lib/Differentiator/CladUtils.cpp
vgvassilev: Then that comment of mine can be resolved by adding a test.

lib/Differentiator/CladUtils.cpp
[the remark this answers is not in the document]
vgvassilev: Great!

lib/Differentiator/CladUtils.cpp
parth-07: Yes! 
I will add a test for this.

lib/Differentiator/CladUtils.cpp
parth-07: I have added a test for this.

lib/Differentiator/CladUtils.cpp
parth-07: I have added a test for this.

lib/Differentiator/CladUtils.cpp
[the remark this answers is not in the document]
parth-07: I have added a test for this.

vgvassilev (approved):
LGTM!

PR #449, opened 2022-05-15
Change return type of `*::Derive` fns to `DerivativeAndOverload`

parth-07:
`*::Derive` functions were returning function declaration context as well that was not being used anywhere. Moreover, we can access the derived function declaration context using `derivative->getDeclContext()`.

### DIFF
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -51,10 +51,15 @@ namespace clad {
   // instead would lead to unnecessarily returning a nullptr in the overloaded
   // FD
   using DeclWithContext = std::pair<clang::FunctionDecl*, clang::Decl*>;
-  /// A tuple which consists of a FunctionDecl, it's potential enclosing context
-  /// and optionally it's overload FunctionDecl
-  using OverloadedDeclWithContext =
-      std::tuple<clang::FunctionDecl*, clang::Decl*, clang::FunctionDecl*>;
+  /// Stores derivative and the corresponding overload. If no overload exist
+  /// then `second` data member should be `nullptr`.
+  struct DerivativeAndOverload {
+    clang::FunctionDecl* derivative = nullptr;
+    clang::FunctionDecl* overload = nullptr;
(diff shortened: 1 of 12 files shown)


include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -51,10 +51,12 @@ namespace clad {
...
   using DeclWithContext = std::pair<clang::FunctionDecl*, clang::Decl*>;
-  /// A tuple which consists of a FunctionDecl, it's potential enclosing context
-  /// and optionally it's overload FunctionDecl
-  using OverloadedDeclWithContext =
-      std::tuple<clang::FunctionDecl*, clang::Decl*, clang::FunctionDecl*>;
+  /// Stores derivative and the corresponding overload. If no overload exist
+  /// then `second` data member should be `nullptr`.
+  struct DerivativeAndOverload {
+    clang::FunctionDecl* derivative;
+    clang::FunctionDecl* overload;
```
vgvassilev, line 58: Do we need default initialization here in case there is no overload?

parth-07: We are always specifically adding `nullptr` value for `overload` data member until now:

```cpp
 return DerivativeAndOverload{cloneFunctionResult.first,
                                 /*OverloadFunctionDecl=*/nullptr};
```

We can add a default value for `overload` data member for these cases. Should I add a default value for `overload` intialisation here?

vgvassilev: Yes, and maybe capitalize the datamembers if we want to use them without getters?

parth-07: We use pascal case (PascalCase) for functions and types. Can we use some other style for public data members?

vgvassilev: Ok, if that's the adopted format elsewhere consistency let's keep it as is.

vgvassilev (approved):
LGTM!

vgvassilev:
Seems a hiccup on codecov side and `-0.01%` lost coverage is not bad either..

PR #451, opened 2022-05-17
Fix diff of ref variable in loops in rev mode

### LINKED ISSUE #440: Reference variable declaration inside loop produces incorrect derivative
Reproducible example: 
```cpp
  double arr[5] = {};
  for (int idx=0; idx<5; ++idx) {
    double& ref = arr[idx];
    ref = i;
  }
  return arr[0] + arr[1] + arr[2] + arr[3] + arr[4];
}

int main() {
  double d_i=0, d_j = 0;
  fn_grad.execute(3, 5, &d_i, &d_j);
  std::cout<<"d_i: "<<d_i<<"\n";
}
```
**Expected Output:** d_i: 5
**Actual Output:**  d_i: 1

Derived function generated by Clad is as follows: 

```cpp
    clad::array<double> _d_arr(5UL);
    unsigned long _t0;
    int _d_idx = 0;
    clad::tape<int> _t1 = {};
    double &_d_ref = _d_arr[_t2];
    double arr[5] = {};
    _t0 = 0;
    for (int idx = 0; idx < 5; ++idx) {
        _t0++;
        double &ref = arr[clad::push(_t1, idx)];
        ref = i;
    }
    double fn_return = arr[0] + arr[1] + arr[2] + arr[3] + arr[4];
    goto _label0;
  _label0:
    {
        _d_arr[0] += 1;
        _d_arr[1] += 1;
        _d_arr[2] += 1;
        _d_arr[3] += 1;
        _d_arr[4] += 1;
    }
    for (; _t0; _t0--) {
        {
            int _t2 = clad::pop(_t1);
            double _r_d0 = _d_ref;
            * _d_i += _r_d0;
            _d_ref -= _r_d0;
        }
    }
}
```

The incorrect derivative is caused by the combination of two bugs: 

1) Local reference variable can be an alias of a different variable in each iteration of the loop, but the corresponding derivative of the reference variable -- a global reference variable can be bound to only one variable in one run of the function.

2) Clad does not have any way to properly access derivatives in the forward sweep. 

Please consider the following code snippet of the derived function:
```cpp
 double &_d_ref = _d_arr[_t2];   <------------- Line of interest
    double arr[5] = {};
    _t0 = 0;
    for (int idx = 0; idx < 5; ++idx) {
        _t0++;
        double &ref = arr[clad::push(_t1, idx)];
        ref = i;
    }

// reverse pass
....
...
...
int _t2 = clad::pop(_t1);  <--------- also of interest
```

Please note both the lines that are marked as lines of interest.  `_t2` is declared and initialised after it is used, this is because derivative expression returned by `Visit(...)` is expected to be used in forward sweep only, so each time Clad is using a derivative expression in forward sweep can potentially be incorrect. There are few ways to move forward from here: 

- We can try to modify `Visit(...)` in such a way that this dependency does not exist.
- We can create a new routine `DerivativeExprForForwSweep(...)`, this routine would return a derivative expression that is only meant to be used in the forward sweep. This method would allow creating derivative expressions efficiently and with fewer resources as we would not need to create a global variables to store part of the derivative expressions. For example, we would be able to use `arr[idx]` instead of `arr[clad::pop(idx)]`.

parth-07:
Closes #440


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2092,6 +2105,17 @@ namespace clad {
         }
       }
 
+      if (isVDRefType) {
+        QualType T = utils::GetValueType(VD->getType());
+        T.removeLocalConst();
```
vgvassilev, line 2110: Didn't/shouldn't we have a helper function that gives us the non-const version of a type?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2092,6 +2105,17 @@ namespace clad {
...
+        VDDerivedType =
+            m_Context.getPointerType(T);
+        initDiff = Visit(VD->getInit());
+        if (!initDiff.getExpr_dx())
+          VDDerivedType = VDDerivedType.getNonReferenceType();
```
vgvassilev, line 2115: Why codecov reports lack of coverage here? Why did we add this condition?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2591,7 +2640,19 @@ namespace clad {
       // for forward-pass loop statement body
       endScope();
     }
-
+    Stmts revLoopBlock;
+    for (auto S : m_LoopBlock.back()) {
```
vgvassilev, line 2644: No curlies.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2591,7 +2640,19 @@ namespace clad {
...
+      revLoopBlock.push_back(S);
+    }
+    if (auto CS = dyn_cast_or_null<CompoundStmt>(bodyDiff.getStmt_dx())) {
```
vgvassilev, line 2647: Likewise.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: I just realized, we have a helper function, `ReverseModeVisitor::ComputeAdjointType`. We can make this function more general, such that it returns correct derivative types for reference variables as well.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: This condition is for reference variables initialised with global variables. Global variables are considered constant and thus have no derivatives. I will try to add a test for this branch.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: We are now using `ReverseModeVisitor::ComputeAdjointType` function here.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: I have added a branch for this code.

parth-07: I have added a code for this branch.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2591,7 +2635,18 @@ namespace clad {
...
-
+    Stmts revLoopBlock;
+    for (auto S : m_LoopBlock.back())
+      revLoopBlock.push_back(S);
+    if (auto CS = dyn_cast_or_null<CompoundStmt>(bodyDiff.getStmt_dx()))
+      for (auto S : CS->body())
+        revLoopBlock.push_back(S);
+    else if (bodyDiff.getStmt_dx()) {
+      revLoopBlock.push_back(bodyDiff.getStmt_dx());
+    }
```
vgvassilev, line 2646:
```suggestion
    else if (bodyDiff.getStmt_dx())
      revLoopBlock.push_back(bodyDiff.getStmt_dx());
```

vgvassilev (approved):
LGTM, modulo the comment. Please put the PR description as a commit message.

parth-07:
@vgvassilev Done, I have resolved the commit and added PR description to the commit message.

PR #453, opened 2022-05-24
Fix cuda test invocation

parth-07:
This PR makes the following changes to correctly run CUDA programs and tests: 
1) Remove explicit `extern "C" { ... }` definitions, thus making the code more portable and easier to maintain. 
2) Modify test invocation to use `CLING_TEST_CUDA_SM_LEVEL` environment variable value as defined in lit configuration file. 
3) Add `RUN` invocations to actually compile the test.

### DIFF
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -17,19 +17,7 @@
 
 #include <assert.h>
 #include <stddef.h>
-
-extern "C" {
-  int printf(const char* fmt, ...);
-#if defined(__APPLE__) || defined(_MSC_VER)
-  extern size_t strlen(const char*);
-  void* malloc(size_t);
-  void free(void *ptr);
-#else
-  extern size_t strlen(const char*) __THROW __attribute_pure__ __nonnull ((1));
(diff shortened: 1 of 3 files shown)


include/clad/Differentiator/Differentiator.h
```diff
@@ -17,19 +17,7 @@
...
-  extern size_t strlen(const char*);
-  void* malloc(size_t);
-  void free(void *ptr);
-#else
-  extern size_t strlen(const char*) __THROW __attribute_pure__ __nonnull ((1));
-  void* malloc(size_t) __THROW __attribute_malloc__ __wur;
-  void free(void *ptr) __THROW;
-#endif
-}
+#include <cstring>
```
vgvassilev, line 20: Is rather keep the old and minimal version. What problem we solve here by including cstring?

parth-07: We do not need to keep track of separate declarations for Apple, MSVC, GNU, and CUDA. And also, different versions of CUDA most probably require different declarations as well.

vgvassilev: Ok, fair point...

vgvassilev (approved):
LGTM!

PR #454, opened 2022-05-25
Modified the dump function to print a more readable output.

### LINKED ISSUE #377: Readability suggestion
https://github.com/vgvassilev/clad/blob/4f1c10e7d9237d4929822fa6e1074dfe377e4ec6/include/clad/Differentiator/Differentiator.h#L201

would be more readable with another line break

```
printf("The code is: \n%s\n", getCode());
```

Nirhar:
Previously the dump function printed the code immediately after "The code is"
which made the actual code less readable. Now a '\n' character has been added
after "The code is" so that the actual code starts from a new line, making it more
readable.

closes #377

### DIFF
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -219,7 +219,7 @@ namespace clad {
     }
  
     void dump() const {
-      printf("The code is: %s\n", getCode());
+      printf("The code is: \n%s\n", getCode());
     }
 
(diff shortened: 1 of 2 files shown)


vgvassilev (approved):
LGTM!

vgvassilev:
You can add in the commit message “This patch fixes #issue_number” and GitHub will connect the commit with the issue and then upon merge it will automatically close it.

PR #457, opened 2022-06-05
Patch for the failure of functions with void return type and with return statement

### LINKED ISSUE #432: Creating pushforward fails for function with `void` return type that contains statement `return;`
Minimum reproducible example: 
```cpp
void some_fn() {
  return;
}

double fn(double i) {
  some_fn();
  return i;
} 

int main() {
  auto d_fn = clad::differentiate(fn);
}
```

**Root cause of the error**
In pushforward differentiation mode, `ForwardModeVisitor::VisitReturnStmt` tries to transform `return res` to `return {res, _d_res}`, but this transformation fails if `return` statement does not has any associated return value.

Nirhar:
Prior to this patch, clad would throw an assertion error when attempting to differentiate void functions that contain a return statement.
The assertion error would indicate that there was an attempt to differentiate a nullptr.
This Patch checks if the there is an expression associated with the null pointer and only if there is one, we differentiate it.
Added tests for the same

This patch fixes #432

### DIFF
--- a/lib/Differentiator/ForwardModeVisitor.cpp
+++ b/lib/Differentiator/ForwardModeVisitor.cpp
@@ -747,6 +747,10 @@ namespace clad {
   }
 
   StmtDiff ForwardModeVisitor::VisitReturnStmt(const ReturnStmt* RS) {
+    //If there is no return value, we must not attempt to differentiate
+    if (!RS->getRetValue())
+      return nullptr;
+    
     StmtDiff retValDiff = Visit(RS->getRetValue());
     Stmt* returnStmt = nullptr;
     if (m_Mode == DiffMode::forward) {
--- a/test/ForwardMode/MemberFunctions.C
+++ b/test/ForwardMode/MemberFunctions.C
@@ -28,6 +28,27 @@ public:
   // CHECK-NEXT:     return (_d_this->x + _d_this->y) * i + _t0 * _d_i + (_d_i * j + i * _d_j) * j + _t1 * _d_j;
   // CHECK-NEXT: }
 
+  void mem_fn_with_void_return() {
+	  return;
+  }
+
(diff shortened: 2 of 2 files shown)


lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -747,6 +747,10 @@ namespace clad {
   }
 
   StmtDiff ForwardModeVisitor::VisitReturnStmt(const ReturnStmt* RS) {
+    //If there is no return value, we must not attempt to differentiate
+    if(RS->getRetValue()==nullptr){
+      return nullptr;
```
parth-07, line 752: We do not use braces with one-liner `if` statements.

Nirhar: have corrected it

parth-07 (commented):
Looks good.

I'd suggest improving the commit message style. Maybe [this](https://chris.beams.io/posts/git-commit/) would be a useful read.

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -747,6 +747,10 @@ namespace clad {
   }
 
   StmtDiff ForwardModeVisitor::VisitReturnStmt(const ReturnStmt* RS) {
+    //If there is no return value, we must not attempt to differentiate
+    if(RS->getRetValue()==nullptr)
```
vgvassilev, line 751: @Nirhar, can you run clang-format?

@parth-07, can you add a section in the docs about running clang-format?

parth-07: Yes, I will add a section for this in the docs.

vgvassilev:
```suggestion
    if (!RS->getRetValue())
```

vgvassilev (approved):
Thanks for the fix! LGTM!

PR #458, opened 2022-06-08
Add Support for calculating FP error estimates of nested function calls.

grimmmyshini:
This PR covers the following cases:

1) Nested function calls with a return value (with and without pass-by-reference inputs). Currently, reference inputs will work correctly in the case they are not being assigned to. Otherwise, the error in those inputs is added to the final error twice. Needs a bit of discussion to resolve.
2) Functions with no return, no reference inputs.
3) Functions with no return but reference inputs.

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -34,7 +34,7 @@ namespace clad {
   DerivativeBuilder::DerivativeBuilder(clang::Sema& S, plugin::CladPlugin& P)
     : m_Sema(S), m_CladPlugin(P), m_Context(S.getASTContext()),
       m_NodeCloner(new utils::StmtClone(m_Sema, m_Context)),
-      m_BuiltinDerivativesNSD(nullptr), m_NumericalDiffNSD(nullptr), m_ErrorEstHandler(nullptr) {}
+      m_BuiltinDerivativesNSD(nullptr), m_NumericalDiffNSD(nullptr) {}
 
   DerivativeBuilder::~DerivativeBuilder() {}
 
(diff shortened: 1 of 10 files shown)


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -165,7 +190,13 @@ namespace clad {
       result = V.Derive(FD, request);
     } else if (request.Mode == DiffMode::experimental_pullback) {
       ReverseModeVisitor V(*this);
+      if(!m_ErrorEstHandler.empty()) {
```
vgvassilev, line 193: Can you run clang-format?

lib/Differentiator/ErrorEstimator.cpp
```diff
@@ -496,8 +502,14 @@ void ErrorEstimationHandler::EmitDeclErrorStmts(VarDeclDiff VDDiff,
     StmtDiff savedDecl = SaveValue(VDRef, isInsideLoop);
     // If the VarDecl has an init, we should assign it with an error.
     if (VD->getInit() && !GetUnderlyingDeclRefOrNull(VD->getInit())) {
-      Expr* errorExpr = GetError(savedDecl.getExpr_dx(),
-                                 m_RMV->BuildDeclRef(VDDiff.getDecl_dx()));
+      Expr* errorExpr;
+      if (!m_NestedFuncError)
```
vgvassilev, line 506: Can we move that duplicate code in a file static function?

lib/Differentiator/MultiplexExternalRMVSource.cpp
```diff
@@ -192,6 +192,15 @@ void MultiplexExternalRMVSource::ActBeforeFinalizingDifferentiateSingleExpr(
   }
 }
 
+void MultiplexExternalRMVSource::ActBeforeDifferentiatingCallExpr(
+    llvm::SmallVectorImpl<clang::Expr*>& pullbackArgs,
+    llvm::SmallVectorImpl<clang::DeclStmt*>& ArgDecls, bool hasAssignee) {
+  for (auto source : m_Sources) {
```
vgvassilev, line 198: We don't add curlies on a single-stmt bodies.

test/ErrorEstimation/BasicOps.C
```diff
@@ -412,6 +501,7 @@ float func8(float x, float y) {
 //CHECK-NEXT:     _final_error += _delta_{{x|y|z}} + _delta_{{x|y|z}} + _delta_{{x|y|z}};
 //CHECK-NEXT: }
 
+
```
vgvassilev, line 504: Stray line.

vgvassilev (commented):
Could you squash all changes into a single commit?

vgvassilev (approved):
LGTM!

PR #460, opened 2022-06-10
Detect Request for use of Enzyme as Backend by user

Nirhar:
This patch is the first step towards integrating enzyme AD with Clad. The user can now mention that he/she wants to use enzyme for AD instead of clad by using the API calls:
           ```clad::differentiate<clad::opts::use_enzyme>(func,"...");```
and     ```clad::gradient<clad::opts::use_enzyme>(func);```
Via this patch one can successfully recognise the request for Enzyme AD backend. It is implemented via Overloading the function templates for differentiate and gradient. Currently after successfully recognising a Enzyme request, an empty function is created to show that we have identified the request for enzyme. That is,
    
```clad::gradient<clad::opts::use_enzyme>(func);```
generates the function:
```
        void func_grad_enzyme(...function arguments... , ...arguments to store derivatives...){
        }
```
Currently generation of dummy function for clad::differentiate<clad::opts::use_enzyme>(...) is not implemented.
These functions must be modified in future patches to insert valid code for enzyme to work on.

### DIFF
--- a/include/clad/Differentiator/DiffPlanner.h
+++ b/include/clad/Differentiator/DiffPlanner.h
@@ -59,6 +59,9 @@ namespace clad {
     /// member information for record (class) type parameters.
     DiffInputVarsInfo DVI;
 
+    // A flag to enable the use of enzyme for backend instead of clad
+    bool use_enzyme = false;
+
     /// Recomputes `DiffInputVarsInfo` using the current values of data members.
     ///
     /// Differentiation parameters info is computed by parsing the argument
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -35,7 +35,9 @@ namespace clad {
     #endif
     return count;
   }
-  
+
+  enum opts { use_enzyme = -1 }; // enum opts
+
   /// Tape type used for storing values in reverse-mode AD inside loops.
   template <typename T>
   using tape = tape_impl<T>;
@@ -286,16 +288,13 @@ namespace clad {
   /// \param[in] fn function to differentiate 
   /// \param[in] args independent parameter information 
   /// \returns `CladFunction` object to access the corresponding derived function.
-  template <unsigned N = 1,
-            typename ArgSpec = const char*,
-            typename F,
+  template <signed N = 1, typename ArgSpec = const char*, typename F,
             typename DerivedFnType = ExtractDerivedFnTraitsForwMode_t<F>,
             typename = typename std::enable_if<
                 !std::is_class<remove_reference_and_pointer_t<F>>::value>::type>
   CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>> __attribute__((
       annotate("D")))
-  differentiate(F fn,
-                ArgSpec args = "",
+  differentiate(F fn, ArgSpec args = "",
                 DerivedFnType derivedFn = static_cast<DerivedFnType>(nullptr),
                 const char* code = "") {
     assert(fn && "Must pass in a non-0 argument");
@@ -306,15 +305,13 @@ namespace clad {
   /// Specialization for differentiating functors.
   /// The specialization is needed because objects have to be passed
   /// by reference whereas functions have to be passed by value.
-  template <unsigned N = 1,
-            typename ArgSpec = const char*,
-            typename F,
+  template <signed N = 1, typename ArgSpec = const char*, typename F,
             typename DerivedFnType = ExtractDerivedFnTraitsForwMode_t<F>,
             typename = typename std::enable_if<
                 std::is_class<remove_reference_and_pointer_t<F>>::value>::type>
-  CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>> __attribute__((annotate("D")))
-  differentiate(F&& f,
-                ArgSpec args = "",
+  CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>> __attribute__((
+      annotate("D")))
+  differentiate(F&& f, ArgSpec args = "",
                 DerivedFnType derivedFn = static_cast<DerivedFnType>(nullptr),
                 const char* code = "") {
     return CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>>(derivedFn, code, f);
@@ -327,7 +324,8 @@ namespace clad {
   /// \param[in] args independent parameters information
   /// \returns `CladFunction` object to access the corresponding derived
   /// function.
-  template <typename ArgSpec = const char*, typename F,
+  template <signed E = 1 /*To check for enzyme*/,
+            typename ArgSpec = const char*, typename F,
             typename DerivedFnType = GradientDerivedFnTraits_t<F>,
             typename = typename std::enable_if<
                 !std::is_class<remove_reference_and_pointer_t<F>>::value>::type>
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: include/clad/Differentiator/DiffPlanner.h [hunk 1]
```diff
@@ -59,6 +59,9 @@ namespace clad {
     /// member information for record (class) type parameters.
     DiffInputVarsInfo DVI;
 
+    // A flag to enable the use of enzyme for backend instead of clad
+    bool use_enzyme = false;
```
parth-07, line 63: We use `DerivedFnCollector` to store derived functions and make sure that Clad does not generate the same derived function multiple times. You will need to modify `DerivedFnCollector`, and possibly `DerivedFnInfo`, to keep their expected behaviour once differentiation through enzyme is involved.

include/clad/Differentiator/Differentiator.h
```diff
@@ -320,17 +325,40 @@ namespace clad {
     return CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>>(derivedFn, code, f);
   }
 
+  //Specialization for using Enzyme as Backend
```
parth-07, line 328: It seems like, `clad::gradient<clad::opts::use_clad>` would work but `clad::differentiate<clad::opts::use_clad>` would not work, shouldn't we make this design more consistent?

include/clad/Differentiator/Differentiator.h
```diff
@@ -35,6 +35,11 @@ namespace clad {
     #endif
     return count;
   }
+
+  namespace opts{
+    struct use_enzyme{};
+    struct use_clad{};
```
vgvassilev, line 41: We don’t need this option. That’s the default anyway.

Nirhar:
Are we planning on adding any more backends in the future, other than enzyme?

vgvassilev:
> Are we planning on adding any more backends in the future, other than enzyme?

I am not aware of any other things we can add at that point but let’s keep the design open where reasonable.

include/clad/Differentiator/DiffPlanner.h
Nirhar: I didnt find the need for this as of now, Clad is able to recognise duplicate calls to differentiate a function and derives the ode only once, atleast for whatever tests I have written till now. But would definitely keep this in mind for upcoming patches. Thanks for bringing this to my attention!

include/clad/Differentiator/Differentiator.h
Nirhar: I have removed ```use_clad``` option

Nirhar:
One thing I wanted to bring to your notice in the new code is that, I have used EnzymeVisitor, which is derived from VisitorBase. As of now I dont see any need for "Visiting" any nodes when integrating with enzyme, but used it anyways because it was convenient to do so. Is this okay? The only place where I think we might need to visit nodes is when we recursively differentiate functions(similar to push_forward and pull_back) but from whatever experiments I have conducted on enzyme, it is capable of handling such cases.

cc @vgvassilev @davidlange6 @parth-07

include/clad/Differentiator/Differentiator.h
vgvassilev: It seems still in the diff, did you push?

lib/Differentiator/CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ add_llvm_library(cladDifferentiator
   ConstantFolder.cpp
   DerivativeBuilder.cpp
   DiffPlanner.cpp
+  EnzymeVisitor.cpp
```
vgvassilev, line 36: I don't expect to need this. Upon seeing `clad::differentiate` or `clad::gradient`, we will form a DiffRequest. In there we can check which was the function we wanted to differentiate and send it to enzyme.

include/clad/Differentiator/Differentiator.h
Nirhar: sry, it wasn't removed during a merge conflict, I have updated it now.

lib/Differentiator/CMakeLists.txt
Nirhar: But enzyme's API needs us to make calls to ```enzyme_autodiff``` with our function as a parameter. We will need to create these Function calls when we ```ProcessDiffRequests``` and enclose them in some wrapper functions so that its easy for clad to interface with it. This is how we communicate with enzyme stating which function we want to differentiate. 

We can try to bypass the Enzyme API and try to use Internal functions of Enzyme directly. But I think this would render our integration useless if the Enzyme team decides to change their internal functions. Enzyme has a ```HandleAutoDiff``` function which takes as input llvm ```CallInst``` and performs differentiation of that function. But that would entail us to write a llvm pass for it and force us to work at the llvm IR level. I think we should not take this approach. Or am I missing something?

parth-07:
> One thing I wanted to bring to your notice in the new code is that, I have used EnzymeVisitor, which is derived from VisitorBase. As of now I dont see any need for "Visiting" any nodes when integrating with enzyme, but used it anyways because it was convenient to do so

If you do not find any need for "visiting" nodes, then why is it convenient to derive from `VisitorBase`?

Nirhar:
@parth-07 I used EnzymeVisitor to use some helper functions available in the Visitor class. But anyways, after a discussion with vassil, I have removed the EnzymeVisitor class.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -550,6 +551,12 @@ namespace clad {
         request.Mode = DiffMode::jacobian;
       } else if (A->getAnnotation().equals("G")) {
         request.Mode = DiffMode::reverse;
+      } else if(A->getAnnotation().equals("GE")){
+        request.Mode = DiffMode::reverse;
```
vgvassilev, line 555: Can you run clang-format, you should be able to do something like `git clang-format HEAD~1`

include/clad/Differentiator/Differentiator.h
```diff
@@ -357,6 +387,24 @@ namespace clad {
...
+  //Specialization of Gradient Template for using Enzyme as backend
+  template <typename E/*For the use of enzyme*/,
+            typename F,
+            typename DerivedFnType = GradientDerivedFnTraits_t<F>,
+            typename = typename std::enable_if<
+                !std::is_class<remove_reference_and_pointer_t<F>>::value>::type,
+            /*Check if enzyme has been asked for*/
+            typename = typename std::enable_if<std::is_same<E,clad::opts::use_enzyme>::value>::type>
+  CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>, true> __attribute__((
+      annotate("GE"))) CUDA_HOST_DEVICE
```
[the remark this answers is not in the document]
Nirhar: I have updated it to remove those annotations. Previously I was using it to differentiate between calls to use enzyme and clad.

lib/Differentiator/DiffPlanner.cpp
Nirhar: Done.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -261,6 +261,11 @@ namespace clad {
       outputArrayStr = m_Function->getParamDecl(lastArgN)->getNameAsString();
     }
 
+    // Check if DiffRequest asks for use of enzyme as backend
+    if (request.use_enzyme) {
```
vgvassilev, line 265: Can you avoid using braces for single statement operators? Here and in a few other places in this PR.

vgvassilev (commented):
This seems almost good to go.

lib/Differentiator/ReverseModeVisitor.cpp

vgvassilev (approved):
LGTM!

PR #466, opened 2022-06-29
Generate code for Enzyme autodiff for functions with Pointer/array arguments

Nirhar:
For functions of type
	```double myfunction(double* arr){...};```
OR 	```double myfunction(double arr[3]){...};```

we generate derivative code of type:
```	
        void d_myfunction(double* arr, clad::array_ref<double> _d_arr){
		double* d_arr = _d_arr.ptr();
		__enzyme_autodiff_myfunction(myfunction, arr, d_arr);
	}
```

Further on enzyme will handle the function differentiation

This Commit also Downloads and Installs Enzyme in the Clad Directory, if
the flag "-DENZYME=On" is passed to cmake during project configuration.
    
This Commit also contains tests for the above feature in test/Gradient/Enzyme.C


parth-07:
> @vgvassilev @parth-07 The clad code in this pull request throws the error
> 
> "cannot initialize a parameter of type 'double (double *)' with an lvalue of type 'double (double *)'"
> 
> during compilation of the below code:
> 
> ```c++
> #include "clad/Differentiator/Differentiator.h"
> double f(double *arr) { return arr[0] * arr[1]; }
> int main() {
>     auto f_grad=clad::gradient<clad::opts::use_enzyme>(f);
> }
> ```
> 
> Any idea how to resolve this?

Are you able to print generated derived function? If yes, can you please paste that in a comment here?

Nirhar:
@parth-07 No, I am not able to print the generated function. The same error is thrown when I attempt to print it.

Nirhar:
> > @parth-07 No, I am not able to print the generated function. The same error is thrown when I attempt to print it.
> 
> Have you tried printing the derived function using the flag `-fdump-derived-fn`?

Yes I tried that, the generated function does not contain the __enzyme_autodiff function call

Nirhar:
> > "cannot initialize a parameter of type 'double (double *)' with an lvalue of type 'double (double *)'"
> 
> The error suggests that you are using an incorrect value category (https://en.cppreference.com/w/cpp/language/value_category) for the function argument of the call to `__enzyme_autodiff_`. Correctly generating `clang::DeclRefExpr` node may fix the issue. You can use, `Sema::BuildDeclRefExpr` or `VisitorBase::BuildDeclRef` for correctly generating it.

Okay let me try this out

Nirhar:
This code runs on my machine, but here the tests throw the error ``` fatal: could not create work tree dir 'Enzyme': Permission denied```. How do I deal with this @vgvassilev @parth-07 ?

Nirhar:
Now I am getting the below error in the CI tests:
```
CMake Error: Could not open file for write in copy operation /usr/local/EnzymeAD/tmp/Enzyme-mkdirs.cmake.tmp
```
I am not getting this error on my machine. How do I resolve this @vgvassilev @parth-07?

test/lit.cfg
```diff
@@ -244,11 +244,46 @@ def inferClang(PATH):
...
+config.opt = inferOpt(config.environment['PATH']).replace('\\', '/')
+config.inst,config.version = getInstallDirAndLLVMVersion()
 
 config.cladlib = inferCladLib(config.environment['PATH']).replace('\\', '/')
 if not lit_config.quiet:
     lit_config.note('using clang: %r' % config.clang)
+if not lit_config.quiet:
+    lit_config.note('using opt: %r' % config.opt)
 if not lit_config.quiet:
     lit_config.note('using auto diff lib: %r' % config.cladlib)
```
parth-07, line 288: We should branch these if conditions into one.

CMakeLists.txt
```diff
@@ -308,3 +308,28 @@ if (TARGET clangBasic)
...
+  SOURCE_SUBDIR enzyme
+
+  GIT_REPOSITORY https://github.com/wsmoses/Enzyme
+  GIT_TAG origin/main
+  GIT_PROGRESS TRUE
+  UPDATE_COMMAND ""
+
+  CMAKE_GENERATOR Ninja
+  CMAKE_ARGS -DLLVM_DIR=${LLVM_DIR}
+  BUILD_COMMAND ninja
```
parth-07, line 327: If I recall correctly, ubuntu GitHub runners do not include the Ninja build system by default. Can you make sure if it is installed in a runner or not?

parth-07:
> Now I am getting the below error in the CI tests:
> 
> ```
> CMake Error: Could not open file for write in copy operation /usr/local/EnzymeAD/tmp/Enzyme-mkdirs.cmake.tmp
> ```
> 
> I am not getting this error on my machine. How do I resolve this @vgvassilev @parth-07?

It seems to be a permission issue. Can you try to install it in a non-root directory? And also, check if we are missing the right permissions somewhere.

test/lit.cfg
Nirhar: done

Nirhar:
> > Now I am getting the below error in the CI tests:
> > ```
> > CMake Error: Could not open file for write in copy operation /usr/local/EnzymeAD/tmp/Enzyme-mkdirs.cmake.tmp
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I am not getting this error on my machine. How do I resolve this @vgvassilev @parth-07?
> 
> It seems to be a permission issue. Can you try to install it in a non-root directory? And also, check if we are missing the right permissions somewhere.

Currently I am installing Enzyme in the "CMAKE_INSTALL_PREFIX". I think in the CI machines this is in some root directory. Where do I install it, is the build_dir a good place?

parth-07:
> > > Now I am getting the below error in the CI tests:
> > > ```
> > > CMake Error: Could not open file for write in copy operation /usr/local/EnzymeAD/tmp/Enzyme-mkdirs.cmake.tmp
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I am not getting this error on my machine. How do I resolve this @vgvassilev @parth-07?
> > 
> > 
> > It seems to be a permission issue. Can you try to install it in a non-root directory? And also, check if we are missing the right permissions somewhere.
> 
> Currently I am installing Enzyme in the "CMAKE_INSTALL_PREFIX". I think in the CI machines this is in some root directory. Where do I install it, is the build_dir a good place?

You can just specify a non-root directory for the `CMAKE_INSTALL_PREFIX` option in the CMake configure command.

parth-07:
> Where do I install it, is the build_dir a good place?

Any non-root directory should work for the debugging purpose right now. We would like things to work as expected when the user is installing in root locations as well. Thus the main goal is to find the underlying issue that's causing the CMake configuration to fail.

Nirhar:
Its mentioned in the Enzyme website that Enzyme has been tested against LLVM 7,8,9,10,11,12. I personally tested it against LLVM 13 and 14 and it works. Maybe the lower versions of LLVM/clang are causing a problem? As I have seen, most failures in the CI are because the Enzyme build has failed.

parth-07:
> Its mentioned in the Enzyme website that Enzyme has been tested against LLVM 7,8,9,10,11,12. I personally tested it against LLVM 13 and 14 and it works. Maybe the lower versions of LLVM/clang are causing a problem? As I have seen, most failures in the CI are because the Enzyme build has failed.

Can you test enzyme integration with Clad when lower versions of LLVM are used in your local system?

Nirhar:
@parth-07 Enzyme fails during build in my System when built with llvm 6, with the same errors as thrown by the CI module, for example osx-clang-runtime6

I noticed a few other CI modules fail due to a flag used during compilation with enzyme(after build), an unrecognised flag in lower versions of llvm. I will correct this.

vgvassilev:
Can you rebase your PR, I think we have fixed the bot failures.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -468,8 +468,82 @@ namespace clad {
...
-        arrayOfParams); This is straightforward to deal in enzyme
+        arrayOfParams); This is straightforward to deal in enzyme - DONE
       */
+      unsigned num_params = m_Function->getNumParams();
+      auto orig_params = m_Function->parameters();
+
+      // Case 1: The function to be differentiated is of type double
+      // func(double* arr){...};
+      //        or double func(double arr[n]){...};
+      if (num_params == 1) {
```
parth-07, line 479: Based on the above comment, shouldn't we add a check here to make sure that the first parameter is of array/pointer type?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -468,8 +468,82 @@ namespace clad {
...
+        QualType type = orig_params[0]->getOriginalType();
+        const clang::Type* t_ptr = type.getTypePtr();
+        if (dyn_cast_or_null<ConstantArrayType>(t_ptr) ||
```
parth-07, line 482: In which case can `t_ptr` be null here?
And if you do not need the cast result, then prefer to use `llvm::isa<>`.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -468,8 +468,82 @@ namespace clad {
...
+            dyn_cast_or_null<PointerType>(t_ptr)) {
+          // Extract Pointer from Clad Array Ref
+          auto arrayRefNameExpr = m_Sema.BuildDeclRefExpr(
+              dyn_cast<ValueDecl>(paramsRef[1]), paramsRef[1]->getType(),
+              VK_LValue, noLoc);
```
parth-07, line 487: Prefer to use `VisitorBase::BuildDeclRef` here, that makes the codebase more uniform and easier to read.

include/clad/Differentiator/Compatibility.h
```diff
@@ -553,6 +553,12 @@ static inline Qualifiers CXXMethodDecl_getMethodQualifiers(const CXXMethodDecl*
 #define CLAD_COMPAT_IfStmt_Create_IfStmtKind_Param(Node) Node->getStatementKind()
 #endif
 
+#if LLVM_VERSION_MAJOR <= 8
+#define CLAD_COMPAT_ACTIONRESULT_TO_PTRTY(Res) Res.get()
```
parth-07, line 557: Can we use `clad_compat::GetResult<...>` instead of this?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -468,8 +468,82 @@ namespace clad {
...
+            enz_params_type.push_back(i->getType());
+
+          // Prepare Function call
+          std::string enzyme_call_name =
+              "__enzyme_autodiff_" + FD->getNameAsString();
+          IdentifierInfo* II_Enz = &m_Context.Idents.get(enzyme_call_name);
+          DeclarationName name_enz(II_Enz);
+
+          QualType enzymeFunctionType = m_Sema.BuildFunctionType(
+              m_Context.VoidTy, enz_params_type, noLoc, name_enz,
```
parth-07, line 533: We generally use a camel case for local variable names. But we also do not have any fixed convention. @vgvassilev Should we decide and create some naming convention for the codebase?

test/Enzyme/ReverseMode.C
```diff
@@ -0,0 +1,24 @@
...
+// RUN: %cladopt input.ll -load=%S/../../EnzymeAD/src/Enzyme-build/Enzyme/%enzyme_so_name -enzyme -o output.ll -S %enzyme_opt_flag
+// RUN: %cladopt output.ll -O2 -o output_opt.ll -S
+// RUN: clang output_opt.ll -o Enzyme.out -lstdc++ -lm
+// RUN: ./Enzyme.out | FileCheck -check-prefix=CHECK-EXEC %s
+// CHECK-NOT: {{.*error|warning|note:.*}}
+// REQUIRES: Enzyme
+
+#include "clad/Differentiator/Differentiator.h"
+
+double f(double* arr) { return arr[0] * arr[1]; }
```
parth-07, line 11: Can you add more comprehensive tests? You mentioned in chat that you got enzyme to work for a wide class of function prototypes. Can you also add tests for them?

test/Enzyme/ReverseMode.C
```diff
@@ -0,0 +1,24 @@
...
+// CHECK: void f_grad_enzyme(double *arr, clad::array_ref<double> _d_arr) {
+// CHECK-NEXT:    double *d_arr = _d_arr.ptr();
+// CHECK-NEXT:    __enzyme_autodiff_f(f, arr, d_arr);
+// CHECK-NEXT:}
+
+int main(){
+  auto f_grad = clad::gradient<clad::opts::use_enzyme>(f);
+  double v[2] = {3, 4};
+  double g[2] = {0};
+  f_grad.execute(v, g);
```
parth-07, line 22: Does enzyme require size information of primal or derivative array anywhere?

parth-07 (changes requested):

lib/Differentiator/ReverseModeVisitor.cpp
Nirhar: yes this done in line number 482

test/Enzyme/ReverseMode.C
Nirhar: This PR does not deal with that, it only deals with function prototypes which have a single arraytype or pointer as argument. My next PR will contain more examples and will deal with different types of function prototypes.

test/Enzyme/ReverseMode.C
Nirhar: No it doesnt for this specific example. Examples where size information are required will be given in the next PR.

lib/Differentiator/ReverseModeVisitor.cpp
Nirhar: changed it to llvm::isa<>

lib/Differentiator/ReverseModeVisitor.cpp

include/clad/Differentiator/Compatibility.h
Nirhar: Since I am using ```VisitorBase::BuildDeclRef```, which already contains ```clad_compat::GetResult<...>```, I have removed this code

lib/Differentiator/ReverseModeVisitor.cpp
Nirhar: changed variable names to camel case

vgvassilev: > We generally use a camel case for local variable names. But we also do not have any fixed convention. @vgvassilev Should we decide and create some naming convention for the codebase?

We should, although what we do is consistent to clang where there is nothing strict enforced. However, I agree we should not add more coding convention styles :)

test/Enzyme/ReverseMode.C
parth-07: Okay! 
Can you add this information in the PR description?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -468,8 +468,69 @@ namespace clad {
         structures that can deal with this easily
 
         b. How to deal with functions of the form - double function(double*
-        arrayOfParams); This is straightforward to deal in enzyme
+        arrayOfParams); This is straightforward to deal in enzyme - DONE
       */
+      unsigned numParams = m_Function->getNumParams();
```
parth-07, line 473: This function does a lot of things now. We should try to make the code more modular. Should we divide it into smaller sections using something like this: 

```cpp
if (!use_enzyme)
  DifferentiateUsingClad(...);
else
  DifferentiateUsingEnzyme(...);
gradientBody = endBlock();
```

@vgvassilev What do you think about this?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -468,8 +468,69 @@ namespace clad {
...
+
+          // Prepare Function call
+          std::string enzymeCallName =
+              "__enzyme_autodiff_" + FD->getNameAsString();
+          IdentifierInfo* IIEnzyme = &m_Context.Idents.get(enzymeCallName);
+          DeclarationName nameEnzyme(IIEnzyme);
+          QualType enzymeFunctionType = m_Sema.BuildFunctionType(
+              m_Context.VoidTy, enzymeParamsType, noLoc, nameEnzyme,
+              originalFnType->getExtProtoInfo());
+          FunctionDecl* enzymeCallFD = FunctionDecl::Create(
```
parth-07, line 523: Where are you creating and pushing scopes necessary for the `enzymeCallFD` ? For example, codes like these:
```cpp
beginScope(Scope::FunctionPrototypeScope | Scope::FunctionDeclarationScope |
               Scope::DeclScope);
m_Sema.PushFunctionScope();
m_Sema.PushDeclContext(getCurrentScope(), m_Derivative);
```

parth-07: Since `enzymeCallFD` represents an extern function declaration, we do not scopes here.

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: That sounds reasonable to me.

test/Enzyme/ReverseMode.C
Nirhar: Done

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -372,142 +378,145 @@ namespace clad {
...
-        addToCurrentBlock(S);
-    else
-      addToCurrentBlock(BodyDiff);
+      Stmt* BodyDiff = Visit(FD->getBody()).getStmt();
+      if (auto CS = dyn_cast<CompoundStmt>(BodyDiff))
+        for (Stmt* S : CS->body())
+          addToCurrentBlock(S);
+      else
+        addToCurrentBlock(BodyDiff);
+    } else {
```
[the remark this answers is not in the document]
Nirhar: I have removed Fwd mode code from here for now, would add it later when I make a PR for forward mode

lib/Differentiator/ReverseModeVisitor.cpp
Nirhar: enzymeCallFD is an extern function, and hence it would not need a Scope

lib/Differentiator/ReverseModeVisitor.cpp
Nirhar: I think this is a good idea, but I would want to attempt this after a few more PRs .

.github/workflows/ci.yml
```diff
@@ -16,17 +16,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-
         include:
           - name: osx-clang-runtime5
             os: macos-10.15
             compiler: clang
             clang-runtime: '5.0'
+            no_enzyme: true
```
vgvassilev, line 24: We do not need `no_enzyme`, we can write `enable_enzyme` for the platform we care about and for the rest it will be undefined. Then we can check `if enable_enzyme` and that should be enough.

.github/workflows/ci.yml
```diff
@@ -638,6 +653,7 @@ jobs:
           -DCMAKE_BUILD_TYPE=$([[ -z "$BUILD_TYPE" ]] && echo RelWithDebInfo || echo $BUILD_TYPE) \
           -DCODE_COVERAGE=${CODE_COVERAGE}  \
           -DLLVM_EXTERNAL_LIT="`which lit`" \
+          $ENZYME_FLAG  \
           $GITHUB_WORKSPACE                 \
           ${{ matrix.extra_cmake_options }}
```
vgvassilev, line 658: We should use the `extra_cmake_options` to add `-DENZYME=On`. I think we should call this option `ENABLE_ENZYME_BACKEND`.

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: That would bloat the history for no good reason I fear.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -468,8 +468,69 @@ namespace clad {
...
+      auto origParams = m_Function->parameters();
+
+      // Case 1: The function to be differentiated is of type double
+      // func(double* arr){...};
+      //        or double func(double arr[n]){...};
+      if (numParams == 1) {
+        QualType type = origParams[0]->getOriginalType();
+        const clang::Type* tPtr = type.getTypePtr();
+        if (llvm::isa<ConstantArrayType>(tPtr) ||
+            llvm::isa<PointerType>(tPtr)) {
```
vgvassilev, line 483:
```suggestion
        QualType OrigTy = origParams[0]->getOriginalType();
        if (OrigTy->isConstantArrayType() || OrigTy->isPointerType()) {
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -468,8 +468,69 @@ namespace clad {
         structures that can deal with this easily
 
         b. How to deal with functions of the form - double function(double*
-        arrayOfParams); This is straightforward to deal in enzyme
+        arrayOfParams); This is straightforward to deal in enzyme - DONE
```
vgvassilev, line 471: Stray comments?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -468,8 +468,69 @@ namespace clad {
...
+          // Prepare Arguments to enzyme_autodiff
+          llvm::SmallVector<Expr*, 16> enzymeArgs;
+          enzymeArgs.push_back(BuildDeclRef(const_cast<FunctionDecl*>(FD)));
+          enzymeArgs.push_back(BuildDeclRef(paramsRef[0]));
+          enzymeArgs.push_back(BuildDeclRef(arrayRefToArrayStmt));
+
+          // Prepare Parameters for Function Signature
+          llvm::SmallVector<ParmVarDecl*, 16> enzymeParams;
+          enzymeParams.push_back(m_Sema.BuildParmVarDeclForTypedef(
+              const_cast<DeclContext*>(FD->getDeclContext()), noLoc,
```
vgvassilev, line 501: We should use a local variable for the DeclContext here.

vgvassilev (commented):
Looks like we are making progress but we still have a few rough edges. Codecov reports quite a lot of untested branches. Can you also add the relevant tests?

lib/Differentiator/ReverseModeVisitor.cpp
Nirhar: Okay, I have put both those sections into 2 different functions

test/Enzyme/ReverseMode.C
```diff
@@ -0,0 +1,24 @@
+// RUN: %cladclang %s -I%S/../../include -S -emit-llvm -o input.ll -O2 -fno-vectorize -fno-slp-vectorize -fno-unroll-loops | FileCheck %s 
+// RUN: %cladopt input.ll -load=%S/../../EnzymeAD/src/Enzyme-build/Enzyme/%enzyme_so_name -enzyme -o output.ll -S %enzyme_opt_flag
+// RUN: %cladopt output.ll -O2 -o output_opt.ll -S
+// RUN: clang output_opt.ll -o Enzyme.out -lstdc++ -lm
```
vgvassilev, line 4: We should construct enzyme in clad. That is, we should call `%cladclang ... -fenzyme-backend ...`, then clad should construct the expected optimization pipeline by enzyme, compile it and trigger enzyme.

test/lit.cfg
```diff
@@ -244,12 +244,45 @@ def inferClang(PATH):
...
+        inst_dir = "../../../inst/"
+    
+    llvm_config = inferClang(config.environment['PATH']).replace('\\', '/')
+    temp = re.search("llvm[-@].*/b",llvm_config)
+    llvm_version= llvm_config[temp.start()+5:temp.end()-2]
+    return inst_dir,llvm_version
+
+
 config.clang = inferClang(config.environment['PATH']).replace('\\', '/')
+config.opt = inferOpt(config.environment['PATH']).replace('\\', '/')
```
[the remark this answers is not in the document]
Nirhar: We need opt because, when enzyme generates code, it generates derivative code which is unoptimised. It is recommended that this code is passed through various compiler optimisations like constant folding, expression simplifications etc to increase performance. Moreover the enzyme pass is linked during an opt execution.

test/lit.cfg
```diff
@@ -244,12 +244,45 @@ def inferClang(PATH):
...
+    opt = lit.util.which('opt', PATH)
+
+    if not opt:
+        lit_config.fatal("couldn't find 'opt' program, try setting "
+                  "OPT in your environment")
+
+    return opt
+
+def getInstallDirAndLLVMVersion():
+    #FIX-THIS: CMAKE_INSTALL_PREFIX is not recognised
```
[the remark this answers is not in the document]
Nirhar: yes enzyme is a shared library, and it is located in the build folder configured by external project add in cmake. This build folder is somewhere inside the EnzymeAD folder(which is inside the clad folder). The existence of the EnzymeAD folder will show that enzyme has been included in the project

test/Enzyme/ReverseMode.C
Nirhar: I managed to remove dependency on opt, and have integrated the enzyme pass into clang command itself. Do I need to create this special flag called "-fenzyme-backend"? I think creation of this flag is out of the scope of clad, as clad only has a frontend plugin, but enzyme has to act in the backend right?

test/Enzyme/ForwardMode.C
```diff
@@ -1,13 +1,14 @@
 // RUN: %cladclang %s -lstdc++ -I%S/../../include -oEnzyme.out 2>&1 | FileCheck %s
 // RUN: ./Enzyme.out
 // CHECK-NOT: {{.*error|warning|note:.*}}
+// REQUIRES: Enzyme
 // XFAIL:*
```
vgvassilev, line 5: Why this test is expected to fail?

Nirhar: Differentiation of functions in Forward mode AD with enzyme is not yet implemented.

vgvassilev: Ah, ok, let's add that as a comment before the xfail.

.github/workflows/ci.yml
```diff
@@ -20,7 +20,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-
```
vgvassilev, line 23: Maybe this change here is significant. Can you undo it?

tools/ClangPlugin.cpp
```diff
@@ -346,10 +363,82 @@ namespace clad {
...
+llvmGetPassPluginInfo() {
+  return {LLVM_PLUGIN_API_VERSION, BACKEND_PLUGIN_NAME, BACKEND_PLUGIN_VERSION,
+          clad::ClangBackendPluginPass::registerCallbacks};
+}
+
+static void loadEnzymePass(const llvm::PassManagerBuilder& Builder,
+                           llvm::legacy::PassManagerBase& PM) {
+  llvm::PassRegistry* PR = llvm::PassRegistry::getPassRegistry();
+  clad::plugin::EnzymePassRegistrationListener er;
+  PR->enumerateWith(&er);
```
Nirhar, line 397: I could have used `PR->getPassInfo("enzyme")` and removed the dependence on `PassRegistrationListener`, but for some reason, the `getPassInfo` function returned `nullptr` for the above code. hence I had to use the `PassRegistrationListener` to iterate through all the passes and get the enzyme pass

vgvassilev: Can you investigate why that happens?

tools/ClangPlugin.cpp
```diff
@@ -346,10 +363,82 @@ namespace clad {
...
+    enzymePassLoader_Ox(llvm::PassManagerBuilder::EP_VectorizerStart,
+                        loadEnzymePass);
+static llvm::RegisterStandardPasses
+    enzymePassLoader_O0(llvm::PassManagerBuilder::EP_EnabledOnOptLevel0,
+                        loadEnzymePass);
+static llvm::RegisterStandardPasses
+    nvvmPassLoader_OEarly(llvm::PassManagerBuilder::EP_EarlyAsPossible,
+                          loadNVVMPass);
+
+#if LLVM_VERSION_MAJOR >= 9
```
vgvassilev, line 427: Do we have test coverage for clang later than 9? If not we should add a new entry to the build matrix.

tools/ClangPlugin.cpp
```diff
@@ -346,10 +363,82 @@ namespace clad {
...
+  loadEnzymePass(Builder, PM);
+  llvm::PassManagerBuilder Builder2 = Builder;
+  Builder2.Inliner = nullptr;
+  Builder2.LibraryInfo = nullptr;
+  Builder2.ExportSummary = nullptr;
+  Builder2.ImportSummary = nullptr;
+  const_cast<llvm::PassManagerBuilder&>(Builder2).populateModulePassManager(PM);
+}
+
+static llvm::RegisterStandardPasses enzymePassLoader_LTO(
```
vgvassilev, line 440: These should probably go in `ClangBackendPlugin.cpp`.

tools/ClangPlugin.h
```diff
@@ -76,9 +84,23 @@ namespace clad {
       bool ValidateClangVersion : 1;
       bool CustomEstimationModel : 1;
       bool PrintNumDiffErrorInfo : 1;
+      bool UseEnzymeBackend : 1;
       std::string CustomModelName;
     };
 
+    llvm::PassInfo* enzymePassInfo;
+    llvm::PassInfo* nvvmPassInfo;
+    struct EnzymePassRegistrationListener : llvm::PassRegistrationListener {
```
vgvassilev, line 93: Why this code is in the header file and not in the source file?

tools/CMakeLists.txt
```diff
@@ -2,21 +2,96 @@ if( NOT CLAD_BUILT_STANDALONE AND NOT LLVM_BUILD_TOOLS )
...
+  ClangBackendPlugin.cpp
   DerivedFnInfo.cpp
   RequiredSymbols.cpp
-  )
+)
+add_llvm_library(cladPlugin ${CLAD_PLUGIN_SRC} )
+
+if (TARGET clangBasic)
+  add_dependencies(cladPlugin clangBasic)
+endif()
```
vgvassilev, line 15:
```suggestion

```
We do not need that anymore.

tools/ClangPlugin.cpp
Nirhar: currently the tests are run in clang-runtime 11, so adding a new entry into build matrix must not be needed

tools/ClangPlugin.h
Nirhar: Okay I will shift it to the source file

tools/ClangPlugin.cpp
Nirhar: Okay, upon some debugging I found that there are two overloaded functions for getPassInfo, one of which takes a StringRef argument, which gives its name, and the other void* type argument which is supposed to give its identifier. I was giving a char* as input hoping that it would be casted as a StringRef, but it was interpreting it as a void* argument and hence the wrong function was being called. Explicitly creating a StringRef object and passing it to getPassInfo solved the problem.

tools/ClangPlugin.cpp
```diff
@@ -346,10 +363,80 @@ namespace clad {
...
+          clad::ClangBackendPluginPass::registerCallbacks};
+}
+
+static void loadEnzymePass(const llvm::PassManagerBuilder& Builder,
+                           llvm::legacy::PassManagerBase& PM) {
+  llvm::PassRegistry* PR = llvm::PassRegistry::getPassRegistry();
+  const llvm::PassInfo* enzymePassInfo = PR->getPassInfo(StringRef("enzyme"));
+  //if enzyme pass is not found, then return
+  if (!enzymePassInfo)
+    return;
```
Nirhar, line 399: The bot shows this part of the code as not covered because, this part of the code will be executed only when cmake is not configured to enable enzyme, and hence the enzyme pass is not found. In the bot where tests are enabled, this will not be covered, and hence the error. The fact that the other bots run successfully mean that this portion of code is executed when enzymePass is not found saving us from some seg faults

test/Enzyme/ForwardMode.C
```diff
@@ -1,13 +1,15 @@
 // RUN: %cladclang %s -lstdc++ -I%S/../../include -oEnzyme.out 2>&1 | FileCheck %s
 // RUN: ./Enzyme.out
 // CHECK-NOT: {{.*error|warning|note:.*}}
+// REQUIRES: Enzyme
```
vgvassilev, line 4: Can you check if this actually runs?

tools/ClangPlugin.cpp
```diff
@@ -346,10 +368,80 @@ namespace clad {
...
+  llvm::PassManagerBuilder Builder2 = Builder;
+  Builder2.Inliner = nullptr;
+  Builder2.LibraryInfo = nullptr;
+  Builder2.ExportSummary = nullptr;
+  Builder2.ImportSummary = nullptr;
+  const_cast<llvm::PassManagerBuilder&>(Builder2).populateModulePassManager(PM);
+}
+
+static llvm::RegisterStandardPasses enzymePassLoader_LTO(
+    llvm::PassManagerBuilder::EP_FullLinkTimeOptimizationEarly, loadLTOPass);
```
vgvassilev, line 444: Can you move this in `ClangBackendPlugin.cpp`?

tools/ClangBackendPlugin.cpp
```diff
@@ -0,0 +1,84 @@
...
+      });
+}
+} // namespace clad
+
+static void loadEnzymePass(const llvm::PassManagerBuilder& Builder,
+                           llvm::legacy::PassManagerBase& PM) {
+  llvm::PassRegistry* PR = llvm::PassRegistry::getPassRegistry();
+  const llvm::PassInfo* enzymePassInfo = PR->getPassInfo(llvm::StringRef("enzyme"));
+  //if enzyme pass is not found, then return
+  if (!enzymePassInfo)
```
vgvassilev, line 40: I do not see how that would be the case if we have linked enzyme in clad. Perhaps turn this into an assertion.

tools/ClangBackendPlugin.cpp
```diff
@@ -0,0 +1,84 @@
...
+    return;
+  llvm::Pass* enzymePassInstance = enzymePassInfo->createPass();
+  PM.add(enzymePassInstance);
+}
+static void loadNVVMPass(const llvm::PassManagerBuilder& Builder,
+                         llvm::legacy::PassManagerBase& PM) {
+  llvm::PassRegistry* PR = llvm::PassRegistry::getPassRegistry();
+  const llvm::PassInfo* nvvmPassInfo = PR->getPassInfo(llvm::StringRef("preserve-nvvm"));
+  //if nvvm pass is not found, then return
+  if (!nvvmPassInfo)
```
vgvassilev, line 50: Likewise.

tools/ClangBackendPlugin.cpp
```diff
@@ -0,0 +1,84 @@
...
+  llvm::PassManagerBuilder Builder2 = Builder;
+  Builder2.Inliner = nullptr;
+  Builder2.LibraryInfo = nullptr;
+  Builder2.ExportSummary = nullptr;
+  Builder2.ImportSummary = nullptr;
+  const_cast<llvm::PassManagerBuilder&>(Builder2).populateModulePassManager(PM);
+}
+
+static llvm::RegisterStandardPasses enzymePassLoader_LTO(
+    llvm::PassManagerBuilder::EP_FullLinkTimeOptimizationEarly, loadLTOPass);
```
vgvassilev, line 81: I don't think we use clad in the context of link-time optimizations. I'd suggest to leave that our for now, including the `loadLTOPass` function.

tools/ClangBackendPlugin.cpp
Nirhar: The linking happens only if the ENABLE_ENZYME_BACKEND flag is set during cmake configuration. This is not set in the usual case, and hence the enzyme pass will not be found in the regular use case

vgvassilev: Ah... fair point, this code is still triggered. Perhaps we can adjust that as well to be conditional outside of this PR.

Nirhar: Hmm.. for me codecov shows that this code is not covered by tests. Are you looking at line 40 or 41?

vgvassilev: Both. I wanted to avoid that report from codecov.

tools/ClangBackendPlugin.cpp
```diff
@@ -0,0 +1,67 @@
...
+// These constructors add our pass to a list of global extensions.
+static llvm::RegisterStandardPasses
+    enzymePassLoader_Ox(llvm::PassManagerBuilder::EP_VectorizerStart,
+                        loadEnzymePass);
+static llvm::RegisterStandardPasses
+    enzymePassLoader_O0(llvm::PassManagerBuilder::EP_EnabledOnOptLevel0,
+                        loadEnzymePass);
+static llvm::RegisterStandardPasses
+    nvvmPassLoader_OEarly(llvm::PassManagerBuilder::EP_EarlyAsPossible,
+                          loadNVVMPass);
```
Nirhar, line 65: The osx-clang-9 bot seems to fail when the destructors of these objects are called by the runtime, during program termination. Will I have to build clang in debug mode on this bot to understand what's happening?

Nirhar: [This](https://github.com/sampsyo/llvm-pass-skeleton/issues/7#issuecomment-461988004) seems like a reasonable explanation to what's happening

Nirhar: Basically, this issue seems to be a standard problem in llvm-9 on Mac OS, and has been fixed in llvm-10. The only work around for this problem is [this](https://github.com/sampsyo/llvm-pass-skeleton/issues/7#issuecomment-576786718). As per the instructions in the link, I am setting the environment variable, and running clang as below:
```
export DYLD_INSERT_LIBRARIES="./obj/lib/clad.dylib"
/usr/local/opt/llvm@9/bin/clang -I ./include -x c++ -std=c++11 -fplugin=./obj/lib/clad.dylib trial.cpp -o trial -lstdc++ -lm
```

but I get the error:
```
dyld: could not load inserted library './obj/lib/clad.dylib' because no suitable image found.  Did find:
        ./obj/lib/clad.dylib: not a dylib
        ./obj/lib/clad.dylib: stat() failed with errno=1
```
@vgvassilev any leads on how this can be resolved? Clearly the .dylib object is found, but is not recognised as a .dylib object. I did try the top answer in [this](https://stackoverflow.com/questions/32730312/reason-no-suitable-image-found), but that too didnt work.

vgvassilev: Can you provide the stacktrace for the early dlclose call?

Nirhar: The nvvm pass is not crucial for enzyme to operate.  Currently I have commented the early nvvm pass registration, and still the error occurs

vgvassilev (approved):
LGTM! Great work @Nirhar!

The codecov coverage reports are mostly due to code using the new pass manager. Unfortunately enzyme has not yet moved to the new pass manager. It's registration happens via the old way. However, I'd like to keep the new way around and they move we will do little work to reuse the already put in place infrastructure.

PR #475, opened 2022-07-26
Recover the testsuite run with clang built in debug mode

vgvassilev:
This PR fixes several problems with clad's testsuite when ran with clang built in debug mode.


vgvassilev:
@grimmmyshini, can you take a look at this PR?

grimmmyshini (commented):
I think it looks good. However, some of the tests are canceled on the CI. Is that intentional?

vgvassilev:
> I think it looks good. However, some of the tests are canceled on the CI. Is that intentional?

Yes, see #476

vgvassilev:
The negative codecov report is expected as I have added a stack protection assertion in `getUnderlyingArrayType`.

PR #485, opened 2022-08-14
Add implicit casts to the dx arguments potentially used in the pushforward

vgvassilev:
The call argument and its differentiated call argument should have the same type. That is not the case where we synthesize 0 or 1 as an integer literals when applying the differentiation rules.

This type mismatch is problematic for the analytical derivatives via pushforwards where we have `ValueAndPushforward<T, T> X_pushforward(T x, T d_x)`. In that case if we pass arguments of type `double` and an `int` we fail to instantiate and thus fail to use the relevant pushforward.

This patch adds an implicit cast to the right type if the type of the call argument differs from the type of the differentiated one.

@sudo-panda, for some reason it seems we do not propagate the pushforward results to the input array in hessian mode. Do you have any clue what might be wrong?


vgvassilev:
@parth-07, I'd like to move forward with this. However, this PR seems to not synthesize the right casts for non-templated, non-overloaded pushforwards (although this does not seem to be problematic) https://github.com/vgvassilev/clad/blob/f3bbf774c1e0c43b249af88139e17eb810b4e584/test/ROOT/TFormula.C#L81-L84

Do you know where that place is?

PR #486, opened 2022-08-16
Add Support for Differentiating functions with both pointer/array type and primitive type parameters with Enzyme

Nirhar:
This commit adds support to differentiate many more function types with enzyme.
Some example supported function types are:
```cpp
double f(double* arr)
double f(double x, double y, double z)
double f(double* arr, int n)
double f(double* arr1, int n,  double* arr2, int m)
double f(double arr[], double x,int n,double y)
```

Tests for these have been written in tests/Enzyme/ReverseMode.C

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -584,68 +584,127 @@ namespace clad {
   }
 
   void ReverseModeVisitor::DifferentiateWithEnzyme() {
-    // FIXME: Generalize this function to differentiate other kinds
-    // of function prototypes
     unsigned numParams = m_Function->getNumParams();
     auto origParams = m_Function->parameters();
     llvm::ArrayRef<ParmVarDecl*> paramsRef = m_Derivative->parameters();
     auto originalFnType = dyn_cast<FunctionProtoType>(m_Function->getType());
 
-    // Case 1: The function to be differentiated is of type double
-    // func(double* arr){...};
-    //        or double func(double arr[n]){...};
-    if (numParams == 1) {
-      QualType origTy = origParams[0]->getOriginalType();
-      if (origTy->isConstantArrayType() || origTy->isPointerType()) {
-        // Extract Pointer from Clad Array Ref
-        auto arrayRefNameExpr = BuildDeclRef(paramsRef[1]);
-        auto getPointerExpr = BuildCallExprToMemFn(arrayRefNameExpr, "ptr", {});
-        auto arrayRefToArrayStmt = BuildVarDecl(
-            origTy, "d_" + paramsRef[0]->getNameAsString(), getPointerExpr);
-        addToCurrentBlock(BuildDeclStmt(arrayRefToArrayStmt),
-                          direction::forward);
-
-        // Prepare Arguments to enzyme_autodiff
-        llvm::SmallVector<Expr*, 16> enzymeArgs;
-        enzymeArgs.push_back(
-            BuildDeclRef(const_cast<FunctionDecl*>(m_Function)));
-        enzymeArgs.push_back(BuildDeclRef(paramsRef[0]));
-        enzymeArgs.push_back(BuildDeclRef(arrayRefToArrayStmt));
-
-        // Prepare Parameters for Function Signature
-        llvm::SmallVector<ParmVarDecl*, 16> enzymeParams;
-        DeclContext* fdDeclContext =
-            const_cast<DeclContext*>(m_Function->getDeclContext());
-        enzymeParams.push_back(m_Sema.BuildParmVarDeclForTypedef(
-            fdDeclContext, noLoc, m_Function->getType()));
-        enzymeParams.push_back(m_Sema.BuildParmVarDeclForTypedef(
-            fdDeclContext, noLoc, paramsRef[0]->getType()));
-        enzymeParams.push_back(m_Sema.BuildParmVarDeclForTypedef(
-            fdDeclContext, noLoc, arrayRefToArrayStmt->getType()));
-
-        // Get the Parameter Types in Function Signature
-        llvm::SmallVector<QualType, 8> enzymeParamsType;
-        for (auto i : enzymeParams)
-          enzymeParamsType.push_back(i->getType());
-
-        // Prepare Function call
-        std::string enzymeCallName =
-            "__enzyme_autodiff_" + m_Function->getNameAsString();
-        IdentifierInfo* IIEnzyme = &m_Context.Idents.get(enzymeCallName);
-        DeclarationName nameEnzyme(IIEnzyme);
-        QualType enzymeFunctionType = m_Sema.BuildFunctionType(
-            m_Context.VoidTy, enzymeParamsType, noLoc, nameEnzyme,
-            originalFnType->getExtProtoInfo());
-        FunctionDecl* enzymeCallFD = FunctionDecl::Create(
-            m_Context, const_cast<DeclContext*>(m_Function->getDeclContext()),
-            noLoc, noLoc, nameEnzyme, enzymeFunctionType,
-            m_Function->getTypeSourceInfo(), SC_Extern);
-        enzymeCallFD->setParams(enzymeParams);
-
-        // Add Function call to block
-        Expr* enzymeCall = BuildCallExprToFunction(enzymeCallFD, enzymeArgs);
-        addToCurrentBlock(enzymeCall);
+    // Extract Pointer from Clad Array Ref
+    llvm::SmallVector<VarDecl*, 8> cladRefParams;
+    for (int i = 0; i < numParams; i++) {
+      QualType paramType = origParams[i]->getOriginalType();
+      if (paramType->isRealType()) {
+        cladRefParams.push_back(nullptr);
+        continue;
       }
+
+      paramType = m_Context.getPointerType(
+          QualType(paramType->getPointeeOrArrayElementType(), 0));
+      auto arrayRefNameExpr = BuildDeclRef(paramsRef[numParams + i]);
+      auto getPointerExpr = BuildCallExprToMemFn(arrayRefNameExpr, "ptr", {});
+      auto arrayRefToArrayStmt = BuildVarDecl(
+          paramType, "d_" + paramsRef[i]->getNameAsString(), getPointerExpr);
+      addToCurrentBlock(BuildDeclStmt(arrayRefToArrayStmt), direction::forward);
+      cladRefParams.push_back(arrayRefToArrayStmt);
+    }
+    // Prepare Arguments and Parameters to enzyme_autodiff
+    llvm::SmallVector<Expr*, 16> enzymeArgs;
+    llvm::SmallVector<ParmVarDecl*, 16> enzymeParams;
+    llvm::SmallVector<ParmVarDecl*, 16> enzymeRealParams;
+    llvm::SmallVector<ParmVarDecl*, 16> enzymeRealParamsRef;
+
+    // First add the function itself as a parameter/argument
+    enzymeArgs.push_back(BuildDeclRef(const_cast<FunctionDecl*>(m_Function)));
+    DeclContext* fdDeclContext =
+        const_cast<DeclContext*>(m_Function->getDeclContext());
+    enzymeParams.push_back(m_Sema.BuildParmVarDeclForTypedef(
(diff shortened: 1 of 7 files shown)


Nirhar:
cc @vgvassilev for review

test/Enzyme/ReverseMode.C
```diff
@@ -5,17 +5,112 @@
...
+  double f4_res1[3]={0};
+  double f4_res2[2]={0};
+  int f4_dn1=0,f4_dn2=0;
+  f4_grad.execute(f4_list1,3,f4_list2,2,f4_res1,&f4_dn1,f4_res2,&f4_dn2);
+  printf("d_x1 = %.2f, d_x2 = %.2f, d_x3 = %.2f, d_n1 = %d\n",f4_res1[0],f4_res1[1],f4_res1[2],f4_dn1); 
+  //CHECK-EXEC: d_x1 = 6.00, d_x2 = 8.00, d_x3 = 10.00, d_n1 = 0
+  printf("d_y1 = %.2f, d_y2 = %.2f, d_n2 = %d\n",f4_res2[0],f4_res2[1],f4_dn2); 
+  //CHECK-EXEC: d_y1 = 2.00, d_y2 = 4.00, d_n2 = 0
+
+  auto f5_grad=clad::gradient<clad::opts::use_enzyme>(f5);
```
vgvassilev, line 107: I am wondering if we can cross check both implementation by running all clad tests with `clad::opts::use_enzyme` (specified somehow by default) and check if we get the same results from tests.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -584,68 +584,144 @@ namespace clad {
...
-        auto arrayRefToArrayStmt = BuildVarDecl(
-            origTy, "d_" + paramsRef[0]->getNameAsString(), getPointerExpr);
-        addToCurrentBlock(BuildDeclStmt(arrayRefToArrayStmt),
-                          direction::forward);
+    // Extract Pointer from Clad Array Ref
+    llvm::SmallVector<VarDecl*, 8> cladRefParams;
+    for (int i = 0; i < numParams; i++) {
+      QualType paramType = origParams[i]->getOriginalType();
+      QualType q2(paramType->getAs<PointerType>(), 0);
+      if (!(paramType->isArrayType() || paramType->isPointerType())) {
```
vgvassilev, line 597:
```suggestion
      if (!isArrayOrPointerType(origParams[i]->getOriginalType())) {
```
We might want to leverage `isDifferentiableType`?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -584,68 +584,144 @@ namespace clad {
...
+        cladRefParams.push_back(nullptr);
+        continue;
+      }
+      if (paramType->isArrayType()) {
+        QualType base(paramType->getArrayElementTypeNoTypeQual(), 0);
+        paramType = m_Context.getPointerType(base);
+      }
```
vgvassilev, line 604:
```suggestion
  paramType = paramType->getPointeeOrArrayElementType();
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -584,68 +584,144 @@ namespace clad {
...
+    for (int i = 0; i < numParams; i++) {
+      // First Add the original parameter
+      enzymeArgs.push_back(BuildDeclRef(paramsRef[i]));
+      enzymeParams.push_back(m_Sema.BuildParmVarDeclForTypedef(
+          fdDeclContext, noLoc, paramsRef[i]->getType()));
+
+      // If the Original parameter does not have a clad array ref
+      // associated with it, then skip the addition of the array ref
+      // pointer.
+      if (cladRefParams[i] == nullptr) {
```
vgvassilev, line 635:
```suggestion
      if (!cladRefParams[i]) {
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -584,68 +584,144 @@ namespace clad {
...
+    // Add rest of the parameters/arguments
+    for (int i = 0; i < numParams; i++) {
+      // First Add the original parameter
+      enzymeArgs.push_back(BuildDeclRef(paramsRef[i]));
+      enzymeParams.push_back(m_Sema.BuildParmVarDeclForTypedef(
+          fdDeclContext, noLoc, paramsRef[i]->getType()));
+
+      // If the Original parameter does not have a clad array ref
+      // associated with it, then skip the addition of the array ref
+      // pointer.
```
vgvassilev, line 634: Can you mention the motivation for this? Is it the case where we have `double*` as a parameter without the corresponding clad::array_ref? If so why we make this shortcut?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -584,68 +584,144 @@ namespace clad {
...
+      enzymeParams.push_back(m_Sema.BuildParmVarDeclForTypedef(
+          fdDeclContext, noLoc, cladRefParams[i]->getType()));
+    }
+
+    llvm::SmallVector<QualType, 16> enzymeParamsType;
+    for (auto i : enzymeParams)
+      enzymeParamsType.push_back(i->getType());
+
+    QualType TT;
+    if (enzymeRealParams.size() != 0) {
```
vgvassilev, line 653:
```suggestion
    if (enzymeRealParams.size()) {
```

vgvassilev (commented):
Thanks for working on this. Here is my first round of review.

lib/Differentiator/ReverseModeVisitor.cpp
Nirhar: No, `isDifferentiableType()` will return true if the QualType is real or pointer/array types. Here I explicitly want to check if it is realType.  Maybe I should rephrase this to `isRealType(origParams[i]->getOriginalType())`

lib/Differentiator/ReverseModeVisitor.cpp
Nirhar: I think this comment is slightly misleading. i will correct this. What I meant was that, if the original parameter was of real type(float/double) then we don't have to obtain the pointer to the arrays from the `clad_array_ref` and neither do we have to add it to the arguments to the enzyme call. Thus we can skip the rest of the code in the loop.

lib/Differentiator/ReverseModeVisitor.cpp
Nirhar: also I cant get rid of the paramType variable, because I am using in many more places from here, but I have removed q2

test/Enzyme/ReverseMode.C
vgvassilev: We need to design this to be able to be used for verification. We also have some past work in this area in #290

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: I think using `isArrayOrPointerType` is fine and keeping `paramType` is also fine.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -584,68 +584,145 @@ namespace clad {
...
+             "cannot find clad::Gradient");
+
+      auto gradDecl = cast<TemplateDecl>(gradR.getFoundDecl());
+      TemplateArgumentListInfo TLI{};
+      llvm::APSInt argValue(std::to_string(enzymeRealParams.size()));
+      TemplateArgument TA(m_Context, argValue, m_Context.UnsignedIntTy);
+      TLI.addArgument(TemplateArgumentLoc(TA, TemplateArgumentLocInfo()));
+      TT = m_Context.getElaboratedType(
+          ETK_None, NS,
+          m_Sema.CheckTemplateIdType(TemplateName(gradDecl), noLoc, TLI));
```
vgvassilev, line 674: Can we use `VisitorBase::GetCladClassOfType`? We should probably rename this function to `MaybeInstantiateAndGetTypeOf` or similar.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -584,68 +584,145 @@ namespace clad {
...
+
+    QualType TT;
+    if (enzymeRealParams.size()) {
+      // Find the Gradient datastructure
+      NamespaceDecl* cladNS = GetCladNamespace();
+      CXXScopeSpec CSS;
+      CSS.Extend(m_Context, cladNS, noLoc, noLoc);
+      NestedNameSpecifier* NS = CSS.getScopeRep();
+      DeclarationName gradIdent = &m_Context.Idents.get("EnzymeGradient");
+      LookupResult gradR(m_Sema, gradIdent, noLoc, Sema::LookupUsingDeclName,
```
vgvassilev, line 661: We can use `VisitorBase::GetCladClassDecl`, and maybe rename the function to `LookupInNamespaceClad` or something.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -584,68 +584,145 @@ namespace clad {
...
+
+      for (int i = 0; i < enzymeRealParams.size(); i++) {
+        auto LHSExpr = BuildOp(UO_Deref, BuildDeclRef(enzymeRealParamsRef[i]));
+
+        UnqualifiedId member;
+        member.setIdentifier(&m_Context.Idents.get("d_arr"), noLoc);
+        CXXScopeSpec SS;
+        auto gradDeclExpr = BuildDeclRef(gradDeclStmt);
+        auto ME = m_Sema
+                      .ActOnMemberAccessExpr(getCurrentScope(), gradDeclExpr,
```
vgvassilev, line 707: Why not `VisitorBase::BuildCallExprToMemFn`?

lib/Differentiator/ReverseModeVisitor.cpp
Nirhar: No, I cant use `CladClassOfType`, because this function inherently assumes that the template arguments will be of `QualType`, but I am using Integral constant template arguments, which wont fit into what the function was designed to do(which is to instantiate templates that use typenames)

lib/Differentiator/ReverseModeVisitor.cpp
Nirhar: No, I cant use `BuildCallExprToMemFn` because I am building an expression that access member variables and am not calling any member function. Thus the `m_Sema.ActOnCallExpr(...)` at the end of `BuildCallExprToMemFn` is problematic to my cause and so I should not be using it.

lib/Differentiator/ReverseModeVisitor.cpp
Nirhar: Changed it to `LookupTemplateDeclInCladNamespace`

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: Can't the function take optionally a `TemplateArgumentListInfo` parameter?

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: We can have an overload of this function which does not call `ActOnCallExpr`.

Nirhar: Created a new function called `BuildExprToMemVariable`

vgvassilev: How is that different from `BuildMemberExpr` in `CladUtils.cpp`?

Nirhar: Oh yes, sorry, I overlooked this

lib/Differentiator/VisitorBase.cpp
```diff
@@ -360,21 +361,14 @@ namespace clad {
                        Sema::LookupUsingDeclName,
                        clad_compat::Sema_ForVisibleRedeclaration);
     m_Sema.LookupQualifiedName(TapeR, CladNS, CSS);
+    std::string errorMsg = "cannot find clad::" + ClassName.str();
     assert(!TapeR.empty() && isa<TemplateDecl>(TapeR.getFoundDecl()) &&
-           "cannot find clad::tape");
+           errorMsg.c_str());
```
vgvassilev, line 366: That's dead code it will not execute. I'd return the old version.

vgvassilev (approved):
LGTM!

PR #491, opened 2022-09-01
Benchmark Clad and Enzyme

Nirhar:
We compare the performance of Reverse Mode in Clad and Enzyme for 4 different kinds of functions:
1. Sum of elements of an array
2. Product of elements of an array
3. Sum of elements of an array multiplied with 2 scalars
4. The Gaussian density function

### DIFF
--- a/benchmark/EnzymeCladComparison.cpp
+++ b/benchmark/EnzymeCladComparison.cpp
@@ -1,0 +1,152 @@
+#include "benchmark/benchmark.h"
+#include "clad/Differentiator/Differentiator.h"
+
+#include "BenchmarkedFunctions.h"
+
+// Compare the execution of Reverse Mode Enzyme and Clad.
+
+static void
+BM_ReverseModeAddArrayAndMultiplyWithScalarsExecute(benchmark::State& state) {
+  auto grad = clad::gradient(addArrayAndMultiplyWithScalars);
+  double x = 5, y = 6;
+  double dx = 0, dy = 0;
+  int n = 5;
+  int dn = 0;
+  double arr[5] = {1, 2, 3, 4, 5};
+  double darr[5] = {0};
+  for (auto _ : state) {
+    grad.execute(arr, x, y, 5, darr, &dx, &dy, &dn);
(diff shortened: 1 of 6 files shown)


benchmark/EnzymeCladComparison.cpp
```diff
@@ -0,0 +1,200 @@
+#include "benchmark/benchmark.h"
+
+// FIXME: If we move this before benchmark.h we have tons of errors due to a bug
+#include "clad/Differentiator/Differentiator.h"
+
+#include "BenchmarkedFunctions.h"
+
+// Compare the execution of Reverse Mode Enzyme and Clad.
+
+// FIXME: Copy made because when we run clad::gradient on the same function
```
vgvassilev, line 10: Wasn't that addressed by #492

Nirhar: yes, have removed the copy functions now

benchmark/EnzymeCladComparison.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+  double dx[5];
+  double dp[5];
+  double ds;
+  int ddim;
+
+  for (auto _ : state) {
+    dfdp_grad.execute(x, p, /*sigma*/ 2, dim, dx, dp, &ds, &ddim);
+    for (int i = 0; i < dim; i++) {
+      // benchmark::DoNotOptimize(sum += result_ref[i]);
```
vgvassilev, line 138: Does uncommenting this lines change anything for the benchmarks?

Nirhar: No, the overall performance difference of the gauss function looks the same to me.

Nirhar: So should I remove these `benchmark::DoNotOptimize` statements from everywhere?

vgvassilev: Yes.

vgvassilev (approved):
LGTM! Thanks!

PR #492, opened 2022-09-01
Treat Request to use Enzyme for Differentiating a Function to be different from a Clad differentiation

Nirhar:
Prior to this commit, If a user differentiates the same function with both Clad and Enzyme, only one derivative is generated, because both Derivatives get stored in the same `DerivedFnInfo` object.
    
Hence if the derivative of a function was already found by using Enzyme or Clad, using the other method would not succeed. This is because the new `DiffRequest` would be taken as a repetition of the old one, and hence the old derivative would be returned.
    
This commit adds a new field `m_UsesEnzyme` to `DerivedFnInfo` that helps the program differentiate between a Function derived with Enzyme and Clad respectively.


vgvassilev:
Can we add a test case for this?

Nirhar:
I have now added a test case demonstrating the use case

Nirhar:
Yes sure, I have updated it now

PR #495, opened 2022-09-03
Add tests to verify enzyme results with clad

Nirhar:
This commit takes a few test cases from `test/Gradient/Gradients.C` and `test/Gradient/Loops.C` and generates gradients for the same using Enzyme as the backend. It also generates clad derivatives for the same and verifies if the results of both are same.

This commit also restructures the test/Enzyme folder with more meaningful test names.


vgvassilev:
Should the names of the tests have something hinting that we are doing comparisons?

Nirhar:
Have updated the names

PR #497, opened 2022-09-07
Add support for Pullback of a Pushforward Function

Nirhar:
requesting @parth-07 for a code review

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1139,17 +1139,25 @@ namespace clad {
     addToCurrentBlock(LS, direction::reverse);
     for (Stmt* S : cast<CompoundStmt>(ReturnDiff.getStmt())->body())
       addToCurrentBlock(S, direction::forward);
-    // Since returned expression may have some side effects affecting reverse
-    // computation (e.g. assignments), we also have to emit it to execute it.
-    Expr* retDeclRefExpr = StoreAndRef(ExprDiff.getExpr(), direction::forward,
-                                       utils::ComputeEffectiveFnName(
-                                           m_Function) +
-                                           "_return",
-                                       /*forceDeclCreation=*/true);
-    
-    if (m_ExternalSource)
-      m_ExternalSource->ActBeforeFinalisingVisitReturnStmt(ExprDiff, retDeclRefExpr);
-    
+
+    // FIXME: When the return type of a function is a class, ExprDiff.getExpr()
+    // returns nullptr, which is a bug. For the time being, the only use case of
+    // a return type being class is in pushforwards. Hence a special case has
+    // been made to to not do the StoreAndRef operation when return type is
+    // ValueAndPushforward.
+    if (!isCladValueAndPushforwardType(type)) {
+      // Since returned expression may have some side effects affecting reverse
+      // computation (e.g. assignments), we also have to emit it to execute it.
+      Expr* retDeclRefExpr =
+          StoreAndRef(ExprDiff.getExpr(), direction::forward,
+                      utils::ComputeEffectiveFnName(m_Function) + "_return",
+                      /*forceDeclCreation=*/true);
+
+      if (m_ExternalSource)
+        m_ExternalSource->ActBeforeFinalisingVisitReturnStmt(ExprDiff,
+                                                             retDeclRefExpr);
+    }
+
     // Create goto to the label.
     return m_Sema.ActOnGotoStmt(noLoc, noLoc, LD).get();
   }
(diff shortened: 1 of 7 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1161,20 +1162,38 @@ namespace clad {
...
+        Expr* clonedEI = Visit(ILE->getInit(i), array_at_i).getExpr();
+        clonedExprs[i] = clonedEI;
+      }
+
+      Expr* clonedILE = m_Sema.ActOnInitList(noLoc, clonedExprs, noLoc).get();
+      return StmtDiff(clonedILE);
+    }else{
+      llvm::SmallVector<Expr*, 16> clonedExprs(ILE->getNumInits());
+      std::string s = ILEType.getAsString().substr(0,25);
+
```
parth-07, line 1184: Can you add an assert here that ensures that the init list expression is of `clad::PushforwardAndValue` type?

Nirhar:
requesting @vgvassilev for review

vgvassilev:
Can you remove the XFAILs introduced in this commit https://github.com/vgvassilev/clad/commit/fac880fcd5a047b9be034d963e1d306a0f5c01d6

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1139,17 +1139,29 @@ namespace clad {
...
-      m_ExternalSource->ActBeforeFinalisingVisitReturnStmt(ExprDiff, retDeclRefExpr);
-    
+
+    // FIXME: When the return type of a function is a class, ExprDiff.getExpr()
+    // returns nullptr, which is a bug. For the time being, the only use case of
+    // a return type being class is in pushforwards. Hence a special case has
+    // been made to to not do the StoreAndRef operation when return type is
+    // ValueAndPushforward.
+    std::string retTypeString = type.getAsString().substr(0, 25);
+    bool isValueAndPushforward =
```
vgvassilev, line 1149: We could model that similarly to `VisitorBase::isCladArrayType`?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1161,20 +1173,49 @@ namespace clad {
...
+        clonedExprs[i] = clonedEI;
+      }
+
+      Expr* clonedILE = m_Sema.ActOnInitList(noLoc, clonedExprs, noLoc).get();
+      return StmtDiff(clonedILE);
+    } else {
+      // FIXME: This is a makeshift arrangement to differentiate an InitListExpr
+      // that represents a ValueAndPushforward type. Ideally this must be
+      // differentiated at VisitCXXConstructExpr
+      llvm::SmallVector<Expr*, 16> clonedExprs(ILE->getNumInits());
```
vgvassilev, line 1196: Can we move out this invariant code outside the if-stmt?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1139,17 +1139,29 @@ namespace clad {
...
+    // FIXME: When the return type of a function is a class, ExprDiff.getExpr()
+    // returns nullptr, which is a bug. For the time being, the only use case of
+    // a return type being class is in pushforwards. Hence a special case has
+    // been made to to not do the StoreAndRef operation when return type is
+    // ValueAndPushforward.
+    std::string retTypeString = type.getAsString().substr(0, 25);
+    bool isValueAndPushforward =
+        retTypeString == "clad::ValueAndPushforward" ||
+        retTypeString.substr(0, 19) == "ValueAndPushforward";
+    if (!isValueAndPushforward) {
```
parth-07, line 1152: Can we create `_return` variable even when the return type is a class type? It will keep things consistent. 
If not, then can you please add the reason why we are not creating `_return` variable when the return type is a class type?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1139,17 +1139,29 @@ namespace clad {
...
+      // Since returned expression may have some side effects affecting reverse
+      // computation (e.g. assignments), we also have to emit it to execute it.
+      Expr* retDeclRefExpr =
+          StoreAndRef(ExprDiff.getExpr(), direction::forward,
+                      utils::ComputeEffectiveFnName(m_Function) + "_return",
+                      /*forceDeclCreation=*/true);
+
+      if (m_ExternalSource)
+        m_ExternalSource->ActBeforeFinalisingVisitReturnStmt(ExprDiff,
```
parth-07, line 1161: Why don't we call it outside the `if (!isValueAndPushforward)` block?

test/Hessian/BuiltinDerivatives.C
```diff
@@ -1,8 +1,8 @@
-// RUN: %cladclang %s -lm -I%S/../../include -oHessianBuiltinDerivatives.out 2>&1 | FileCheck %s
+// RUN: %cladclang %s -lm -I%S/../../include -oHessianBuiltinDerivatives.out 2>&1
```
parth-07, line 1: Why did you remove 'FileCheck' from here?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2970,7 +3011,7 @@ namespace clad {
   ReverseModeVisitor::VisitCXXConstructExpr(const CXXConstructExpr* CE) {
     llvm::SmallVector<Expr*, 4> clonedArgs;
     for (auto arg : CE->arguments()) {
-      auto argDiff = Visit(arg);
+      auto argDiff = Visit(arg, dfdx());
```
parth-07, line 3014: Can you please add a FIXME comment here explaining the limitations of constructor differentiation here?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1161,20 +1173,49 @@ namespace clad {
...
+      bool isValueAndPushforward =
+          initListType == "clad::ValueAndPushforward" ||
+          initListType.substr(0, 19) == "ValueAndPushforward";
+      assert(isValueAndPushforward &&
+             "Only InitListExpr that represents arrays or ValueAndPushforward "
+             "Object initialization is supported");
+
+      Expr* dValueExpr =
+          utils::BuildMemberExpr(m_Sema, getCurrentScope(), dfdx(), "value");
+      StmtDiff clonedValueEI = Visit(ILE->getInit(0), dValueExpr).getExpr();
```
parth-07, line 1207: Can you please add a comment here why don't we need derivative expression here of `ILE->getInit(0)`, and `ILE->getInit(1)`.

parth-07 (changes requested):

lib/Differentiator/ReverseModeVisitor.cpp
Nirhar: This cannot be done because `retDeclRefExpr` sent as parameter in function call is declared in the previous line of the same block

test/Hessian/BuiltinDerivatives.C
Nirhar: This was a temporary change. It will be added in the final version of the PR

lib/Differentiator/ReverseModeVisitor.cpp
Nirhar: Yes, the explanation for this has been added just before the if block begins

lib/Differentiator/ReverseModeVisitor.cpp

test/ROOT/Hessian.C
```diff
@@ -2,7 +2,7 @@
 // RUN: ./Hessian.out | FileCheck -check-prefix=CHECK-EXEC %s
 
 //CHECK-NOT: {{.*error|warning|note:.*}}
-// XFAIL:*
```
vgvassilev, line 5: Awesome!

vgvassilev (approved):
LGTM but let's wait on @parth-07's review.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: We can open an issue later on regarding construction differentiation limitations. We are not differentiating constructors right now at all. `Visit(arg, dfdx())` is also not correct for a general case because relationship between `dfdx()` and `arg` depends on the constructor. We should ideally create a pullback function for the constructor, and the pullback function should update argument derivatives based on the `dfdx`.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: We most probably do not need derivative expression here.

parth-07 (approved):
Looks good.

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: @Nirhar, please open an issue about that so that we do not forget.

vgvassilev:
Let's merge it - I do not understand the codecov report - probably some false positive...

PR #508, opened 2022-11-12
Make pow pushforward fn numerically stable

### LINKED ISSUE #507: Issues with the pushforward/pullback pow function
```cpp

double f1(double x, double y)  {
   return (1.-x)*(1.-x) + 100.*(y-x*x)*(y-x*x);
}

double f2(double x, double y) {
  return std::pow((1.-x),2) + 100. * std::pow(y-std::pow(x,2),2);
   // return TMath::Power((1.-x),2)+100.*TMath::Power((y-TMath::Power(x,2)),2) ;
}

void testClad() {

   auto f = f1;

   auto fgrad  = clad::gradient(f2);

   double dx,dy = 0;

   double x = -1;
   double y = 1.2;

   fgrad.execute(x,y,&dx,&dy);

   std::cout << "gradient is " << dx << "  " << dy << std::endl;

   std::cout << "function value is " << f(x,y) << std::endl;

   double mat_f[4] = {0,0,0,0};
   clad::array_ref<double> h(mat_f, 4);

   auto fhess = clad::hessian(f);

   fhess.execute(x,y, h);

   std::cout << "hessian is \n " << h[0] << "  " << h[1] << "\n "
             << h[2] << "  " << h[3] << std::endl;
}
```

Shows `nan`-s for f2 whereas it works as expected for f1 which does not have uses of `std::pow`.

parth-07:
This commit makes `clad::custom_derivatives::std::pow_pushforward` fn numerically stable by only adding directional derivative w.r.t exponent if the directional seed of exponent is non-zero. This will make the function give the correct directional derivative for the cases where log(base) is undefined but directional derivative w.r.t exponent is not requested -- thus log(base) should ideally never be used.

Closes #507

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -83,9 +83,16 @@ template <typename T1, typename T2>
 CUDA_HOST_DEVICE ValueAndPushforward<decltype(::std::pow(T1(), T2())),
                                      decltype(::std::pow(T1(), T2()))>
 pow_pushforward(T1 x, T2 exponent, T1 d_x, T2 d_exponent) {
-  return {::std::pow(x, exponent),
-          (exponent * ::std::pow(x, exponent - 1)) * d_x +
-              (::std::pow(x, exponent) * ::std::log(x)) * d_exponent};
+  auto val = ::std::pow(x, exponent);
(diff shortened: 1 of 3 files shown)


vgvassilev (approved):
Awesome. Looks good to me!

PR #493, opened 2022-09-01
Add changes necessary for the error estimation benchmark

sudo-panda:

PR #512, opened 2022-12-05
[EE] Add variable name parameter in GetError

sudo-panda:
Provides the user with the name of the variable on which GetError is being called.

This PR is made from splitting #493 into two PRs.


vgvassilev (commented):
Can you extend the demos (eg test.cpp) or testcases to demonstrate the purpose of this change?

vgvassilev (approved):
LGTM!

PR #513, opened 2022-12-07
Fix incompatible array pushes to tape.

grimmmyshini:
Previously, if we see an array that is passed to a function, we would try to insert it into the tape. However, this is wrong because best case, we would end up only cloning the array pointer and not the array itself. This commit introduces a new tape push function that captures a snapshot (clones the array contents) when a pointer is pushed to the tape.

This method is very inefficient given that most function implicity treat arrays as read-only data structures, so we end up cloning the unchanged array several times. One possible fix is to identify if the function the array is passed to expects a const array and then do not clone the array.

This one PR, by no means, is a complete solution to the problem with arrays in loops. For example, for input arrays, we still end up pushing the pointer instead of the whole array.

### DIFF
--- a/include/clad/Differentiator/Array.h
+++ b/include/clad/Differentiator/Array.h
@@ -37,6 +37,14 @@ public:
     for (std::size_t i = 0; i < size; ++i)
       m_arr[i] = a[i];
   }
+
+  CUDA_HOST_DEVICE array(const array<T>& arr) : array(arr.m_arr, arr.m_size) {}
+
+  CUDA_HOST_DEVICE array<T>& operator=(const array<T>& arr) {
+    (*this) = arr.m_arr;
+    return *this;
+  }
+
   /// Destructor to delete the array if it was created by array_ref
   CUDA_HOST_DEVICE ~array() { delete[] m_arr; }
 
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -43,10 +43,10 @@ namespace clad {
   using tape = tape_impl<T>;
 
   /// Add value to the end of the tape, return the same value.
-  template <typename T>
-  CUDA_HOST_DEVICE T push(tape<T>& to, T val) {
-    to.emplace_back(val);
-    return val;
+  template <typename T, typename... ArgsT>
+  CUDA_HOST_DEVICE T push(tape<T>& to, ArgsT... val) {
+    to.emplace_back(std::forward<ArgsT>(val)...);
+    return to.back();
   }
 
   /// Add value to the end of the tape, return the same value.
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -36,6 +36,17 @@ using namespace clang;
 
 namespace clad {
 
+Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
+                       ReverseModeVisitor& rvm) {
+  if (auto CAT = dyn_cast<ConstantArrayType>(AT))
+    return ConstantFolder::synthesizeLiteral(context.getSizeType(), context,
+                                             CAT->getSize().getZExtValue());
+  else if (auto VSAT = dyn_cast<VariableArrayType>(AT))
+    return rvm.Clone(VSAT->getSizeExpr());
+
+  return nullptr;
+}
+
   Expr* ReverseModeVisitor::CladTapeResult::Last() {
     LookupResult& Back = V.GetCladTapeBack();
     CXXScopeSpec CSS;
@@ -53,8 +64,11 @@ namespace clad {
   ReverseModeVisitor::CladTapeResult
   ReverseModeVisitor::MakeCladTapeFor(Expr* E, llvm::StringRef prefix) {
     assert(E && "must be provided");
+    QualType EQt = E->getType();
+    if (dyn_cast<ArrayType>(EQt))
+      EQt = GetCladArrayOfType(utils::GetValueType(EQt));
     QualType TapeType =
-        GetCladTapeOfType(getNonConstType(E->getType(), m_Context, m_Sema));
+        GetCladTapeOfType(getNonConstType(EQt, m_Context, m_Sema));
     LookupResult& Push = GetCladTapePush();
     LookupResult& Pop = GetCladTapePop();
     Expr* TapeRef =
@@ -75,7 +89,12 @@ namespace clad {
     Expr* PopExpr =
         m_Sema.ActOnCallExpr(getCurrentScope(), PopDRE, noLoc, TapeRef, noLoc)
             .get();
-    Expr* CallArgs[] = {TapeRef, E};
+    Expr* exprToPush = E;
+    if (auto AT = dyn_cast<ArrayType>(E->getType())) {
+      Expr* init = getArraySizeExpr(AT, m_Context, *this);
+      exprToPush = BuildOp(BO_Comma, E, init);
+    }
+    Expr* CallArgs[] = {TapeRef, exprToPush};
     Expr* PushExpr =
         m_Sema.ActOnCallExpr(getCurrentScope(), PushDRE, noLoc, CallArgs, noLoc)
             .get();
@@ -228,7 +247,7 @@ namespace clad {
   ReverseModeVisitor::Derive(const FunctionDecl* FD,
                              const DiffRequest& request) {
     if (m_ExternalSource)
-      m_ExternalSource->ActOnStartOfDerive();                               
+      m_ExternalSource->ActOnStartOfDerive();
     silenceDiags = !request.VerboseDiags;
     m_Function = FD;
 
@@ -2258,10 +2277,12 @@ namespace clad {
     auto VDDerivedType = ComputeAdjointType(VD->getType());
     bool isDerivativeOfRefType = VD->getType()->isReferenceType();
     VarDecl* VDDerived = nullptr;
-    
-    if (auto VDCAT = dyn_cast<ConstantArrayType>(VD->getType())) {
-      VDDerivedInit = ConstantFolder::synthesizeLiteral(
-          m_Context.getSizeType(), m_Context, VDCAT->getSize().getZExtValue());
+
+    // VDDerivedInit now serves two purposes -- as the initial derivative value
+    // or the size of the derivative array -- depending on the primal type.
+    if (auto AT = dyn_cast<ArrayType>(VD->getType())) {
+      Expr* init = getArraySizeExpr(AT, m_Context, *this);
+      VDDerivedInit = init;
       VDDerived = BuildVarDecl(VDDerivedType, "_d_" + VD->getNameAsString(),
                                VDDerivedInit, false, nullptr,
                                clang::VarDecl::InitializationStyle::CallInit);
@@ -2434,6 +2455,8 @@ namespace clad {
   StmtDiff ReverseModeVisitor::VisitDeclStmt(const DeclStmt* DS) {
     llvm::SmallVector<Decl*, 4> decls;
     llvm::SmallVector<Decl*, 4> declsDiff;
+    // Need to put array decls inlined.
+    llvm::SmallVector<Decl*, 4> localDeclsDiff;
     // For each variable declaration v, create another declaration _d_v to
     // store derivatives for potential reassignments. E.g.
     // double y = x;
@@ -2464,7 +2487,10 @@ namespace clad {
         if (VDDiff.getDecl()->getDeclName() != VD->getDeclName())
           m_DeclReplacements[VD] = VDDiff.getDecl();
         decls.push_back(VDDiff.getDecl());
-        declsDiff.push_back(VDDiff.getDecl_dx());
+        if (isa<VariableArrayType>(VD->getType()))
+          localDeclsDiff.push_back(VDDiff.getDecl_dx());
+        else
+          declsDiff.push_back(VDDiff.getDecl_dx());
       } else {
         diag(DiagnosticsEngine::Warning,
              D->getEndLoc(),
@@ -2473,11 +2499,22 @@ namespace clad {
     }
 
     Stmt* DSClone = BuildDeclStmt(decls);
-    Stmt* DSDiff = BuildDeclStmt(declsDiff);
-    addToBlock(DSDiff, m_Globals);
 
-    if (m_ExternalSource)
+    if (!localDeclsDiff.empty()) {
+      Stmt* localDSDIff = BuildDeclStmt(localDeclsDiff);
+      addToCurrentBlock(
+          localDSDIff,
+          clad::rmv::forward); // Doesnt work for arrays decl'd in loops.
+    }
+    if (!declsDiff.empty()) {
+      Stmt* DSDiff = BuildDeclStmt(declsDiff);
+      addToBlock(DSDiff, m_Globals);
+    }
+
+    if (m_ExternalSource) {
+      declsDiff.append(localDeclsDiff.begin(), localDeclsDiff.end());
       m_ExternalSource->ActBeforeFinalizingVisitDeclStmt(decls, declsDiff);
+    }
     return StmtDiff(DSClone);
   }
 
@@ -2554,11 +2591,8 @@ namespace clad {
     m_CurScope = m_DerivativeFnScope;
 
     VarDecl* Var = nullptr;
-    if (auto CAT = dyn_cast<ConstantArrayType>(Type)) {
-      Type = GetCladArrayOfType(QualType(CAT->getPointeeOrArrayElementType(),
-                                         Type.getCVRQualifiers()));
-      init = ConstantFolder::synthesizeLiteral(
-          m_Context.getSizeType(), m_Context, CAT->getSize().getZExtValue());
+    if (isa<ArrayType>(Type)) {
+      Type = GetCladArrayOfType(m_Context.getBaseElementType(Type));
       Var = BuildVarDecl(Type, identifier, init, false, nullptr,
                          clang::VarDecl::InitializationStyle::CallInit);
     } else {
@@ -2584,14 +2618,19 @@ namespace clad {
       Expr* Push = CladTape.Push;
       Expr* Pop = CladTape.Pop;
       return {Push, Pop};
-    } else {
-      Expr* Ref = BuildDeclRef(GlobalStoreImpl(Type, prefix));
-      if (E) {
-        Expr* Set = BuildOp(BO_Assign, Ref, E);
-        addToCurrentBlock(Set, direction::forward);
-      }
-      return {Ref, Ref};
     }
+
+    Expr* init = nullptr;
+    if (auto AT = dyn_cast<ArrayType>(Type)) {
+      init = getArraySizeExpr(AT, m_Context, *this);
+    }
+
+    Expr* Ref = BuildDeclRef(GlobalStoreImpl(Type, prefix, init));
+    if (E) {
+      Expr* Set = BuildOp(BO_Assign, Ref, E);
+      addToCurrentBlock(Set, direction::forward);
+    }
+    return {Ref, Ref};
   }
 
   StmtDiff ReverseModeVisitor::GlobalStoreAndRef(Expr* E,
@@ -3074,10 +3113,9 @@ namespace clad {
       TValueType.removeLocalConst();
       return m_Context.getPointerType(TValueType);
     }
-    if (auto CAT = dyn_cast<ConstantArrayType>(T)) {
+    if (isa<ArrayType>(T) && !isa<IncompleteArrayType>(T)) {
       QualType adjointType =
-          GetCladArrayOfType(QualType(CAT->getPointeeOrArrayElementType(),
-                                      CAT->getIndexTypeCVRQualifiers()));
+          GetCladArrayOfType(m_Context.getBaseElementType(T));
       return adjointType;
     }
     T.removeLocalConst();
--- a/test/Arrays/ArrayInputsReverseMode.C
+++ b/test/Arrays/ArrayInputsReverseMode.C
@@ -1,4 +1,4 @@
-// RUN: %cladclang %s -lm -lstdc++ -I%S/../../include -oArrayInputsReverseMode.out 2>&1 | FileCheck %s
+// RUN: %cladclang %s -lm -lstdc++ -I%S/../../include -Wno-unused-value -oArrayInputsReverseMode.out 2>&1 | FileCheck %s
 // RUN: ./ArrayInputsReverseMode.out | FileCheck -check-prefix=CHECK-EXEC %s
 
 //CHECK-NOT: {{.*error|warning|note:.*}}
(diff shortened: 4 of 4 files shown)


vgvassilev (commented):
Can you undo the trailing white space fixes unless the code was touched on purpose. Also, please add more tests at least where codecov suggested.

grimmmyshini:
done.

include/clad/Differentiator/Array.h
```diff
@@ -37,6 +37,13 @@ template <typename T> class array {
     for (std::size_t i = 0; i < size; ++i)
       m_arr[i] = a[i];
   }
+
+  CUDA_HOST_DEVICE array(const array<T>& arr) : array(arr.m_arr, arr.m_size) {}
```
parth-07, line 41: This will lead to a segmentation fault. clad::array owns the array memory it contains and deletes the memory when it is destroyed. With this copy constructor, multiple references of clad::array will point to the same memory, and all of them will try to delete the memory when they are being destroyed.

We should delete the copy constructor of clad::array, and make it a move-only type.

include/clad/Differentiator/Array.h
```diff
@@ -37,6 +37,13 @@ template <typename T> class array {
...
+
+  CUDA_HOST_DEVICE array<T>& operator=(const array<T>& arr) {
+    this = arr.m_arr;
```
parth-07, line 44: Why do we assign `this` to a pointer of type `T*`?
Why don't we have any return value here?

include/clad/Differentiator/Differentiator.h
```diff
@@ -43,10 +43,10 @@ namespace clad {
   using tape = tape_impl<T>;
 
   /// Add value to the end of the tape, return the same value.
-  template <typename T>
-  CUDA_HOST_DEVICE T push(tape<T>& to, T val) {
-    to.emplace_back(val);
-    return val;
+  template <typename T, typename... ArgsT>
+  CUDA_HOST_DEVICE T push(tape<T>& to, ArgsT... val) {
+    to.emplace_back(val...);
```
parth-07, line 48: We should do perfect-forwarding here using `std::forward`

parth-07 (changes requested):

vgvassilev:
Ping on the comments. Did you resolve the cloning issue?

grimmmyshini:
Hi, Vassil! 

Yes, after uselessly debugging this, I realized the issue was quite simple: 

In the generated code, the derivative was declared before the size could be declared (as all derivatives are put into the global block by default):  

```cpp
clad::array<double> _d_arr(n);
// end of Global block
// start of forward block
int n = k;
```

I have fixed this by putting variable-sized array decls in the forward block, like so: 

```cpp
// start of forward block
int n = k;
clad::array<double> _d_arr(n);
```

However, you can tell this just won't work for variable-sized arrays declared in loops and would lead to something like this: 

```cpp
for (int i = 0; i < 3; i++) {
    _t0++;
    int n = k;
    clad::array<double> _d_arr(n);
    double arr[n];
    \\ ...
}
```

So, this is a proposed fix: 

Modify the `clad::array` class to contain a dummy memory address (or nullptr) and then actually allocate the space later. So something like this...

```cpp
// array is now decl'd in global scope with some dummy size.
clad::array<double> _d_arr(1);
for (int i = 0; i < 3; i++) {
    _t0++;
    int n = k;
   // The following function will delete the previous data and reallocate new memory depending on n. 
   // Since only we want to use this if the size of the previous was the same, we do not allocate new space.
    _d_arr.reallocateSpaceForNewSize(n);
    double arr[n];
    \\ ...
}
```

However, I am not sure if this approach is sane AD-wise. Maybe we want to save every copy of the diff array declared in the previous example, and then we would probably need something more sophisticated. Although in my opinion, this should not be necessary... 

Anyway, this is a big change and needs more discussion, so I don't know if we want to include it in this PR.

include/clad/Differentiator/Differentiator.h
grimmmyshini: yes sure

include/clad/Differentiator/Array.h
grimmmyshini: First point - because I expect the overloaded operator `CUDA_HOST_DEVICE array<T>& operator=(T* arr)` to be called there. 

Second point - that is a fair point, we should return `*this`

vgvassilev:
I am not sure if ré-(allocating) in a loop is a good idea. I fail to understand why we can’t debug the original issue. This has been a repeating problem with your setup. Is there anything unique to your setup that prevents you to debug why the reference is not properly updated and more globally, what clang does in the same setup when processing textually written code.

grimmmyshini:
Um, I think you misunderstood my message above. I did debug it, but then I realized that the cloning was okay rather the size was being declared later than the actual array declaration causing a bad-alloc. More details on the previous message.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2472,11 +2501,22 @@ namespace clad {
     }
 
     Stmt* DSClone = BuildDeclStmt(decls);
-    Stmt* DSDiff = BuildDeclStmt(declsDiff);
-    addToBlock(DSDiff, m_Globals);
 
-    if (m_ExternalSource)
+    if (!localDeclsDiff.empty()) {
```
vgvassilev, line 2505: What is the reason to have a new structure here? Why it does not work for arrays declared in loops? It is very unlikely somebody defines/allocates arrays within a loop. Is there a use case in particular? Otherwise we can just transform this into a diagnostic.

@parth-07, can we somehow support this without having to special-case it?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2552,11 +2592,9 @@ namespace clad {
     m_CurScope = m_DerivativeFnScope;
 
     VarDecl* Var = nullptr;
-    if (auto CAT = dyn_cast<ConstantArrayType>(Type)) {
+    if (auto CAT = dyn_cast<ArrayType>(Type)) {
```
vgvassilev, line 2595:
```suggestion
    if (isa<ArrayType>(Type)) {
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3072,11 +3115,13 @@ namespace clad {
...
-          GetCladArrayOfType(QualType(CAT->getPointeeOrArrayElementType(),
-                                      CAT->getIndexTypeCVRQualifiers()));
-      return adjointType;
+    if (auto CAT = dyn_cast<ArrayType>(T)) {
+      if (!dyn_cast<IncompleteArrayType>(T)) {
+        QualType adjointType =
+            GetCladArrayOfType(QualType(CAT->getPointeeOrArrayElementType(),
+                                        CAT->getIndexTypeCVRQualifiers()));
+        return adjointType;
+      }
```
vgvassilev, line 3124:
```suggestion
    if (isa<ArrayType>(T) && !isa<IncompleteArrayType>(T)) {
        QualType adjointType =
            GetCladArrayOfType(QualType(T->getPointeeOrArrayElementType(),
                                        T->getIndexTypeCVRQualifiers()));
        return adjointType;
      }
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -75,7 +89,12 @@ namespace clad {
     Expr* PopExpr =
         m_Sema.ActOnCallExpr(getCurrentScope(), PopDRE, noLoc, TapeRef, noLoc)
             .get();
-    Expr* CallArgs[] = {TapeRef, E};
+    Expr* exprToPush = E;
+    if (auto AT = dyn_cast<ArrayType>(E->getType())) {
```
vgvassilev, line 93:
```suggestion
    if (isa<ArrayType>(E->getType())) {
```
We will need to add an assertion in `getArraySizeExpr` if that was not an array type; and cast the type there.

vgvassilev:
@grimmmyshini ping.

include/clad/Differentiator/Array.h
parth-07: > First point - because I expect the overloaded operator CUDA_HOST_DEVICE array<T>& operator=(T* arr) to 
> be called there.

Then shouldn't the code be:
```cpp
*this = arr.m_arr;
```

parth-07:
Hi @vgvassilev, Hi @grimmmyshini 

Thank you Garima for the detailed problem description and an initial solution to move forward with.

```cpp
// start of forward block
int n = k;
clad::array<double> _d_arr(n);
```

We perhaps should not rely on this technique even outside the loops because `k` here can be an arbitrarily complex expression and may also depend on runtime conditions. It may be infeasible to compute all of those before the forward block starts.

I do like the proposed fix involving `clad::array::reallocateSpaceForNewSize`. 

> Maybe we want to save every copy of the diff array declared in the previous example, and then we would probably 
> need something more sophisticated. Although in my opinion, this should not be necessary...

Yes, that should not be necessary. 

One drawback of this approach is that the derivative programs will consume more than required memory. If we can create local derivative variables in the derivative loops then that can solve this issue.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2257,10 +2276,15 @@ namespace clad {
...
     bool isDerivativeOfRefType = VD->getType()->isReferenceType();
     VarDecl* VDDerived = nullptr;
-    
-    if (auto VDCAT = dyn_cast<ConstantArrayType>(VD->getType())) {
-      VDDerivedInit = ConstantFolder::synthesizeLiteral(
-          m_Context.getSizeType(), m_Context, VDCAT->getSize().getZExtValue());
+
+    if (auto AT = dyn_cast<ArrayType>(VD->getType())) {
+      Expr* init = nullptr;
+      if (auto CAT = dyn_cast<ConstantArrayType>(AT))
```
parth-07, line 2282: Can you please refactor this into a utility function `getArrayInit`?

parth-07: Can you use, `getArraySizeExpr` function here?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2257,10 +2276,15 @@ namespace clad {
     auto VDDerivedType = ComputeAdjointType(VD->getType());
     bool isDerivativeOfRefType = VD->getType()->isReferenceType();
     VarDecl* VDDerived = nullptr;
-    
-    if (auto VDCAT = dyn_cast<ConstantArrayType>(VD->getType())) {
-      VDDerivedInit = ConstantFolder::synthesizeLiteral(
-          m_Context.getSizeType(), m_Context, VDCAT->getSize().getZExtValue());
+
```
parth-07, line 2279: Can you please add a comment here detailing that `VDDerivedInit` now serves two purposes -- as the initial derivative value or the size of the derivative array -- depending on the primal type.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: Hi @vgvassilev
I am not sure at the moment. I am trying to see how we can improve the existing design to handle the missing cases.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2472,11 +2501,22 @@ namespace clad {
     }
 
     Stmt* DSClone = BuildDeclStmt(decls);
-    Stmt* DSDiff = BuildDeclStmt(declsDiff);
-    addToBlock(DSDiff, m_Globals);
 
-    if (m_ExternalSource)
+    if (!localDeclsDiff.empty()) {
+      Stmt* localDSDIff = BuildDeclStmt(localDeclsDiff);
+      addToCurrentBlock(
```
parth-07, line 2507: Why do we add derivative arrays in the current block? Can it cause issues in codes such as below:

```cpp
{
  int n = /*some random expression*/;
  double arr[n] = {....};
  res += arr[0] + arr[1];
}
```

vgvassilev:
@grimmmyshini ping!

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2582,14 +2620,19 @@ namespace clad {
...
-      Expr* Ref = BuildDeclRef(GlobalStoreImpl(Type, prefix));
-      if (E) {
-        Expr* Set = BuildOp(BO_Assign, Ref, E);
-        addToCurrentBlock(Set, direction::forward);
-      }
-      return {Ref, Ref};
     }
+
+    Expr* init = nullptr;
+    if (auto AT = dyn_cast<ArrayType>(Type)) {
```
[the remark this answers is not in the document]
grimmmyshini: This is not a valid suggestion on its own, we need the array type for the `getSizeExpr` function.

lib/Differentiator/ReverseModeVisitor.cpp
grimmmyshini: This is not a valid suggestion on its own. We need the array type to get the underlying element type of the array.

lib/Differentiator/ReverseModeVisitor.cpp
grimmmyshini: I can use the ```!isa<IncompleteArrayType>(T)``` instead of ```dyn_cast``` but I don't see how I can call ```ArrayType``` size expression functions through just an ```expr*``` without ever casting it.

lib/Differentiator/ReverseModeVisitor.cpp
grimmmyshini: Why? It adds two unnecessary checks instead of just one. Secondly, even if I do what you suggest I see no scenario in which ```getArraySizeExpr``` will see a non ```ArrayType```.

include/clad/Differentiator/Array.h
grimmmyshini: true!

lib/Differentiator/ReverseModeVisitor.cpp
grimmmyshini: If the `{}` signify a different scope, then yes. I know this is a problem, but I don't know how to solve this because even currently, we have problems with variables declared in a different scope (having the lifetime of that scope) but are still used in the reverse pass. 

Maybe treat is the same as you would a loop? With something like ```isInsideDifferentScope```.

grimmmyshini:
> Hi @vgvassilev, Hi @grimmmyshini
> 
> Thank you Garima for the detailed problem description and an initial solution to move forward with.
> 
> ```c++
> // start of forward block
> int n = k;
> clad::array<double> _d_arr(n);
> ```
> 
> We perhaps should not rely on this technique even outside the loops because `k` here can be an arbitrarily complex expression and may also depend on runtime conditions. It may be infeasible to compute all of those before the forward block starts.

yeah sure you could have something weird like 

```
int k = 10;
double arr[++k];
```
but in my experience, these special conditions are super rare and not deterministic of the usual case where ```k``` would be a fairly constant variable. We could make a case to support these "weird" cases but I see absolutely no point in doing that when we don't have any use cases for that.

Secondly, coupling the size declaration and array derivative declaration was the most "natural" coupling I found. Maybe in the future, we want to exactly inline all derivative decls with actual decls but even then you will not be covered in weird cases (like the one above) :smiley:  

Any better ideas for implementing this are welcomed though. 

> I do like the proposed fix involving `clad::array::reallocateSpaceForNewSize`.
> 
> > Maybe we want to save every copy of the diff array declared in the previous example, and then we would probably
> > need something more sophisticated. Although in my opinion, this should not be necessary...
> 
> Yes, that should not be necessary.
> 
> One drawback of this approach is that the derivative programs will consume more than the required memory. If we can create local derivative variables in the derivative loops then that will reduce memory consumption.

True but then we need to remove these derivatives' usage in the reverse mode i.e. outside the scope of the original loop. Maybe it is worth considering this in a separate independent issue.

lib/Differentiator/ReverseModeVisitor.cpp
grimmmyshini: Yep, I missed converting this somehow.

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: This way we reduce the cyclomatic code complexity and as for the size you can see the suggestion in https://github.com/vgvassilev/clad/pull/513/files#r1057265867

lib/Differentiator/ReverseModeVisitor.cpp
grimmmyshini: okay

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: `getPointeeOrArrayElementType` exists on clang::Type as well according to the documentation of clang. Is that not sufficient?

lib/Differentiator/ReverseModeVisitor.cpp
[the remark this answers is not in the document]
vgvassilev: Fair enough, I thought the code might read better if `getArraySizeExpr` takes a type and asserts if it were not an array type. I don't have a strong opinion though.

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: Both functions seem available in `clang::Type`.

lib/Differentiator/ReverseModeVisitor.cpp
grimmmyshini: Ah, I did not realize that, sorry. Okay, I can make this change.

lib/Differentiator/ReverseModeVisitor.cpp
grimmmyshini: Yes, okay.

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: Where are these two checks? If you mean that we will have a check outside of the function and then inside the function implementation then sure. However, the way I'd read this code is "hey we have a special case for arrays" and it's being taken care of in this utility function. Leaking a variable also makes the reader track what goes on in the utility function and outside - that's not to worry today but will have to worry tomorrow.

lib/Differentiator/ReverseModeVisitor.cpp
grimmmyshini: um, could you point me to ```Type::getIndexTypeCVRQualifiers```, I don't think it is covered by ```Type```.

vgvassilev: Ah, yes, sorry, you are right. I will be happy if we use `!isa<IncompleteArrayType>(T)` at that point.

lib/Differentiator/ReverseModeVisitor.cpp
grimmmyshini: I disagree here. I think the dynamic cast provides a clear-er message and avoids redundant checks. However, we can move the cast inside if you like.

vgvassilev: If that's an unambiguous code readability improvement for you then let's keep it as it is.

grimmmyshini: Okay, thanks!

lib/Differentiator/ReverseModeVisitor.cpp
grimmmyshini: hmm, what do you suggest? Unfortunately, I do not remember the reason for using this exactly.

vgvassilev: I am not sure if the goal was to add back the qualifiers or something else. Let's hope this is covered by tests - I'd suggest to remove this and add `0` see if we have a failing test and take it from there. I suspect we might be able to use `ASTContext::getBaseElementType` to bring back the qualifiers.

grimmmyshini: ```ASTContext::getBaseElementType``` works well. I have added this and other changes.

vgvassilev: Maybe change the rest of the occurrences if any.

grimmmyshini: Yes, I did that. I also squashed the changes from the round of reviews into one commit.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: Yes, I meant when `{ }` signify a different scope. This is a considerable restriction. Variable-sized arrays can basically only be used in function global scope.

> even currently, we have problems with variables declared in a different scope (having the lifetime of that scope) but are still used in the reverse pass.

Yes, but these cases are rare within clad. We should ideally try to remove all such cases.

Can you please open an issue to track that variable-sized array cannot be used within a local scope?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2464,7 +2487,10 @@ namespace clad {
         if (VDDiff.getDecl()->getDeclName() != VD->getDeclName())
           m_DeclReplacements[VD] = VDDiff.getDecl();
         decls.push_back(VDDiff.getDecl());
-        declsDiff.push_back(VDDiff.getDecl_dx());
+        if (dyn_cast<VariableArrayType>(VD->getType()))
```
vgvassilev, line 2490:
```suggestion
        if (isa<VariableArrayType>(VD->getType()))
```

parth-07:
> but in my experience, these special conditions are super rare and not deterministic of the usual case where k would be a fairly constant variable. We could make a case to support these "weird" cases but I see absolutely no point in doing that when we don't have any use cases for that.

Yeah, makes sense. I misunderstood a few things before, and the approach seemed a lot more restrictive to me than it actually was.

> Secondly, coupling the size declaration and array derivative declaration was the most "natural" coupling I found.

Yes, it is definitely the most "natural" coupling. But there was a reason why we always declared derivative variables in the function's global scope. For instance, having size declaration and array derivative declaration in the same scope restricts the usage of variable-sized arrays to a function's global scope.

> Maybe in the future, we want to exactly inline all derivative decls with actual decls but even then you will not be covered in weird cases (like the one above) 😃

I am not sure if that's a good idea. I believe the example you showed can be easily made to work by using temporaries if the array size expression contains side effects. This would not be a big change. But yes, I don't think it's required right now.

> Any better ideas for implementing this are welcomed though.

I liked your idea of `clad::Array::reallocateSpaceForNewSize`. This solution is robust and can be uniformly used for functions' global and local scopes. It also does not have flaws present in the current solution. I do understand that
`clad::Array::reallocateSpaceForNewSize` solution might be more time-consuming, but we should definitely keep it in mind if we cannot find a better solution in the future.

> True but then we need to remove these derivatives' usage in the reverse mode i.e. outside the scope of the original loop.

In the reverse-pass, the derivative will only be used in the corresponding adjoint code in which it can ideally be declared to reduce memory consumption. Thus, we should be able to limit the scope in the reverse-pass.
One problem here is that we may need to access derivatives in the forward-pass, in some obscure cases such as when which variable a value is assigned to is only known at runtime. For example:

```cpp
fn(i, j, k) = val;
```

`fn(i, j, k)` can return any of `i`, `j` or `k`.

> Maybe it is worth considering this in a separate independent issue.

Yes, definitely.

parth-07 (approved):
Looks good.

vgvassilev:
@grimmmyshini, can you incorporate the last part of the suggestions? I think we are ready to go.

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: Do we have opened an issue for this?

vgvassilev (approved):
LGTM! Modulo the comment I left.

grimmmyshini:
> LGTM! Modulo the comment I left.

which one? I addressed the last one in the new commit.

vgvassilev:
> > LGTM! Modulo the comment I left.
> 
> which one? I addressed the last one in the new commit.

https://github.com/vgvassilev/clad/pull/513/files#r1062206538

PR #516, opened 2022-12-19
Fix return value of `ReverseMoveVisitor::VisitImplicitCastExpr`

### LINKED ISSUE #515: In reverse mode AD, constant reference variable gives segmentation fault
Differentiating the program with constant reference variable in Reverse mode AD gives no compilation error but gives segmentation fault at runtime. 

Example: 

```C++

  double a = i;
  const double& ar = a;
  double res = ar;
  return res;
}

int main() {
  double d_i, d_j;
  fn_grad.execute(3, 5, &d_i, &d_j);
}
```
This program compiles without error but the output file gives a segmentation fault: 
```
[1]    31889 segmentation fault  ./clad/demos/ConstIssue.out
```

By using the  "-fdump-derived-fn" flag, the problematic derivative function was found as follow:
```
    double _d_a = 0;
    double *_d_ar = 0;
    double _d_res = 0;
    double a = i;
    const double &ar = a;
    double res = ar;
    double fn_return = res;
    goto _label0;
  _label0:
    _d_res += 1;
    *_d_ar += _d_res;
    * _d_i += _d_a;
}
```

The issue seems to be: `_d_ar` is not being set to `_d_a`. So, there is no derivative variable associated with `ar`. Clad does set `_d_ar` to `_d_a` when `ar` is  non-constant reference.

arora-vidushi:
This commit fixes `ReverseModeVisitor::VisitImplicitCastExpr` function return value. `VisitImplicitCastExpr` function was not correctly forwarding the value of `StmtDiff::m_DerivativeForForwSweep`.

Closes #515

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2485,7 +2485,8 @@ namespace clad {
     StmtDiff subExprDiff = Visit(ICE->getSubExpr(), dfdx());
     // Casts should be handled automatically when the result is used by
     // Sema::ActOn.../Build...
-    return StmtDiff(subExprDiff.getExpr(), subExprDiff.getExpr_dx());
+    return StmtDiff(subExprDiff.getExpr(), subExprDiff.getExpr_dx(),
+                    subExprDiff.getForwSweepStmt_dx());
   }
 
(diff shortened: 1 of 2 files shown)


vgvassilev (approved):
Awesome first patch! LGTM!

PR #517, opened 2022-12-22
Add test for constant variables in Reverse mode AD 

### LINKED ISSUE #430: Derived variables should never be `const` in Reverse Mode AD
A reproducible example: 

```cpp
  const double a = i;
  double res = a;
  return res;
}

int main() {
}
```

Compiling this program gives the following error: 
```
error: cannot assign to variable '_d_a' with const-qualified type 'const double'
issue.cpp:3:8: note: variable '_d_a' declared const here
       ^~
1 error generated.
```

The root cause of the error is that we are creating the derived variable of a `const` type -- because the corresponding primal variable is of `const` type. But even if the primal variable is of `const` type and cannot be modified in the primal function, the derived variable can still be modified and thus shouldn't be `const`.

arora-vidushi:
Closes #430

### DIFF
--- a/test/Gradient/Gradients.C
+++ b/test/Gradient/Gradients.C
@@ -216,7 +216,7 @@ double f_div2(double x, double y) {
 void f_div2_grad(double x, double y, clad::array_ref<double> _d_x, clad::array_ref<double> _d_y);
 
 double f_c(double x, double y) {
-  return -x*y + (x + y)*(x/y) - x*x; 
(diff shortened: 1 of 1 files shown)


vgvassilev:
If there is no particular reason to have the tests in two separate commits, let’s squash them. Otherwise looks good to me.

arora-vidushi:
Yes I have squashed the commits now.

vgvassilev (approved):
LGTM!

PR #522, opened 2023-01-29
[cmake] Disable version checks when we build against the monorepo

vgvassilev:


vgvassilev:
Ah, I see we have `CLANG_VERSION_UPPER_BOUND` which can be bumped to support my use-case. Let's close this then.

PR #534, opened 2023-02-24
Remove dependency on LastKnowGood*.txt

### LINKED ISSUE #523: Drop LastKnownGood*.txt
Now we have more accurate infrastructure to track what's supported and what not. We should clean up also the dependency on these files from https://github.com/vgvassilev/clad/blob/2bc2b4b5bcede9debc27f4aaa8a4ea52af0d0ba8/lib/Differentiator/CMakeLists.txt#L8

Ris-Bali:
This PR aims to remove dependency on LastKnowGoodClang/LLVM.txt files.

closes #523


vgvassilev (commented):
Thanks for working on this! We should probably get rid of all `getClangCompatRevision` and friends as this becomes unused.

Could you update the commit message to include "Fixes #534" which would automatically close the relevant issue.

include/clad/Differentiator/Version.h
```diff
@@ -7,11 +7,4 @@
 #ifndef CLAD_DIFFERENTIATOR_VERSION
 #define CLAD_DIFFERENTIATOR_VERSION
 
-#include <string>
-
-namespace clad {
-  std::string getCladRevision();
```
parth-07, line 13: We still need to keep getCladRevision.

parth-07 (changes requested):

include/clad/Differentiator/Version.h
Ris-Bali: Noted I will revert this

tools/ClangPlugin.h
```diff
@@ -146,8 +130,6 @@ namespace clad {
             m_DO.GenerateSourceFile = true;
           } else if (args[i] == "-fvalidate-clang-version") {
             m_DO.ValidateClangVersion = true;
-            if (!IsRunningOnExpectedClangVersion())
```
parth-07, line 149: @vgvassilev Instead of removing this function call altogether, should we instead check if the supported Clang version is being used? I think Clad supports Clang 5 to Clang 15.

vgvassilev: Sure, but what happens if you compile from git? This has been a problem before but nowadays we can handle this indeed. So maybe we can check against the clang major version if it is between 5 and 15… 

Although these types of things have caused major troubles throughout years.

parth-07: > Although these types of things have caused major troubles throughout years.

Yes, but I think now we have a good infrastructure to handle this. And also, this is only used if `-fvalidate-clang-version` is specified.

@Ris-Bali Can you please make this change? Max and minimum clang version supported values can be taken 
from the cmake variables `CLANG_MIN_SUPPORTED` and `CLANG_VERSION_UPPER_BOUND`. We will need to make them available to cpp files. Please let us know if you have any questions.

Ris-Bali: Apologies for the late reply,I have made both the variables available to the cpp files from where can I get the clang_version being used ? I couldn't find a variable for that . Thanks

parth-07: Hi @Ris-Bali, you can use the macros `__clang__`, `__clang_major__` and `__clang__minor__` to determine
the clang version being used at runtime. These macros are documented here: [https://clang.llvm.org/docs/LanguageExtensions.html#feature_check](https://clang.llvm.org/docs/LanguageExtensions.html#feature_check)

Ris-Bali:
I have made the changes. Pls review them

tools/ClangPlugin.h
```diff
@@ -146,8 +130,16 @@ namespace clad {
...
-              return false; // Tells clang not to create the plugin.
+            #if defined(__clang__) &&                                                      \
+     (__clang_major__ > CLANG_MINIMUM_MAJOR ||                                  \
+     (__clang_major__ == CLANG_MINIMUM_MAJOR &&                                \
+      __clang_minor__ >= CLANG_MINIMUM_MINOR)) &&                              \
+     (__clang_major__ < CLANG_UPPER_MAJOR ||                                    \
+     (__clang_major__ == CLANG_UPPER_MAJOR &&                                  \
+      __clang_minor__ <= CLANG_UPPER_MINOR))
```
parth-07, line 139: Versions starting from `CLANG_VERSION_UPPER_BOUND` are not supported.
As per your conditions, a clang version matching the `CLANG_VERSION_UPPER_BOUND` is supported, when it's actually not supported.

parth-07 (changes requested):

tools/ClangPlugin.h
```diff
@@ -146,8 +130,16 @@ namespace clad {
             m_DO.GenerateSourceFile = true;
           } else if (args[i] == "-fvalidate-clang-version") {
```
vgvassilev, line 131: Now that we have a good check we should probably make this a default behavior that could be disabled with say `fno-validate-clang-version`?

Ris-Bali: Will work on this

Ris-Bali:
I have made the changes. Pls Review

tools/CMakeLists.txt
```diff
@@ -92,3 +92,14 @@ if (NOT CLAD_BUILD_STATIC_ONLY)
...
+string(REPLACE "." ";" VERSION_LIST_UPPER "${CLANG_VERSION_UPPER_BOUND}")
+string(REPLACE "." ";" VERSION_LIST_MINIMUM "${CLANG_MIN_SUPPORTED}")
+list(GET VERSION_LIST_UPPER 0 CLANG_UPPER_MAJOR)
+list(GET VERSION_LIST_UPPER 1 CLANG_UPPER_MINOR)
+list(GET VERSION_LIST_MINIMUM 0 CLANG_MINIMUM_MAJOR)
+list(GET VERSION_LIST_MINIMUM 1 CLANG_MINIMUM_MINOR)
+add_definitions( -DCLANG_UPPER_MAJOR=${CLANG_UPPER_MAJOR} )
+add_definitions( -DCLANG_UPPER_MINOR=${CLANG_UPPER_MINOR} )
+add_definitions( -DCLANG_MINIMUM_MAJOR=${CLANG_MINIMUM_MAJOR} )
+add_definitions( -DCLANG_MINIMUM_MINOR=${CLANG_MINIMUM_MINOR} )
```
vgvassilev, line 105: Aren't these defined in `clang/Basic/Version.h`?

Ris-Bali: Doesn't clang/Basic/Version.h give the CLANG_VERSION_MAJOR etc of the clang being used for the project ? The variables I have defined here are taken from CLANG_VERSION_UPPER_BOUND and CLANG_MIN variables in clad's CmakeLists.txt.

tools/ClangPlugin.h
```diff
@@ -184,6 +166,19 @@ namespace clad {
             return false; // Tells clang not to create the plugin.
           }
         }
+        if (m_DO.ValidateClangVersion != false) {
+#if defined(__clang__)
```
parth-07, line 170: Hi @Ris-Bali, 
Thank you very much for working on this.

It needs some modifications. I think we were wrong about a few things before.
@vgvassilev Please correct me if I am wrong somewhere.

The built `clad.so` does not support the complete `CLANG_MINIMUM_MAJOR` to `CLANG_UPPER_MAJOR`
range. This is because Clad contains many codes that are Clang version-specific. You can see the clang version-specific codes in `include/clad/Differentiator/Compatibility.h`. To fix this, validate clang version should check if the
currently running clang version is the same as the clang version that was used for building `clad.so`.

Another issue is how to get the clang version of the currently running clang? I know I previously suggested using `__clang__*` macros. I was wrong. These macros are evaluated to the version of the Clang that is used to build Clad.
We need to determine the version of Clang at runtime. The following functions may help to determine the value of the
runtime clang version being used to run Clad:
- `clang::getClangRevision`
- `clang::getClangFullVersion`
- `clang::getClangFullCPPVersion`.

@Ris-Bali Can you please verify that these functions indeed give the runtime Clang version, and not the clang used
to build Clad?

Ris-Bali: Yes the getclangFullCPPVersion() and getClangFullVersion() give similar outputs and give the runtime version of clang , getclangRev

image

So next should I compare them with __clang__version_ macro ? It is a combination of clang_major, clang_minor and clang_patchlevel_ macros and should give the version with which clad was built?

Ris-Bali: I have made the changes. Pls review. I think the tests are failing because of using a different version runtime ?

Ris-Bali:
I have reverted the version check so that it is not default behavior anymore; If I make it default behavior half the tests (with different clang version at runtime) fail. So should I keep it default or is this fine ?

vgvassilev:
Can you squash the commits into one? We should open a new issue which suggests enabling `-fvalidate-clang-version` by default. I am not sure I understand why the builders are failing if we do it.

Ris-Bali:
> Can you squash the commits into one? We should open a new issue which suggests enabling `-fvalidate-clang-version` by default. I am not sure I understand why the builders are failing if we do it.

Sure I'll squash the commits.

Ris-Bali:
Correct me if I am wrong; I think the builders are failing when we keep this as default behavior because we use different clang version for building clad and different clang version at runtime. (eg : clang8-runtime 10 ) will fail as the clang version used to build clad and the clang version at runtime are different. Meanwhile builders which use the same version are not failing (eg clang 9-runtime 9)

vgvassilev:
Yes, you are right. In that case, however, the clad version macros should capture not which compiler has built the binary but what was the target runtime. That is, what is the version of clang used as a library for clad to link against. We have these values in cmake already.

parth-07 (approved):
Thank you @Ris-Bali for working on this. LGTM.

vgvassilev:
> Yes, you are right. In that case, however, the clad version macros should capture not which compiler has built the binary but what was the target runtime. That is, what is the version of clang used as a library for clad to link against. We have these values in cmake already.

I think this is what we want. We should get the version of the runtime clad was build for and not who built it. For example we build with gcc, what would we do there if we pursue this approach?

Ris-Bali:
> > Yes, you are right. In that case, however, the clad version macros should capture not which compiler has built the binary but what was the target runtime. That is, what is the version of clang used as a library for clad to link against. We have these values in cmake already.
> 
> I think this is what we want. We should get the version of the runtime clad was build for and not who built it. For example we build with gcc, what would we do there if we pursue this approach?

Understood I'll use the variables defined in cmake

tools/CMakeLists.txt
```diff
@@ -92,3 +92,4 @@ if (NOT CLAD_BUILD_STATIC_ONLY)
     add_dependencies(clad LLVMEnzyme)
   endif(ENABLE_ENZYME_BACKEND)
 endif()
+add_definitions( -DCLANG_FULL_VERSION="${CLANG_VERSION_MAJOR}.${CLANG_VERSION_MINOR}.${CLANG_VERSION_PATCH}" )
```
vgvassilev, line 95: We should define this only for the cpp file and move the check into a routine that’s defined there.

Ris-Bali: I've made the changes now CLANG_FULL_VERSION can only be accessed by clangplugin.h. Could you please clarify what do you mean by moving the check into a routine. I couldn't understand that part of your request

vgvassilev: I meant something like `set_source_files_properties(ClangPlugin.cpp PROPERTIES COMPILE_DEFINITIONS "CLANG_FULL_VERSION=\"${CLANG_FULL_VERSION}\"")`

Ris-Bali: Thanks for the reply. I think there is no other way other than creating a separate header file through configure_file() if we need to access a cmake variable in a header file (here ClangPlugin.h) As mentioned [here](https://discourse.cmake.org/t/how-to-get-access-cmake-variable-in-a-header-file/7125) .  I have tested this personally too and got an error.

tools/ClangPlugin.h
```diff
@@ -184,6 +167,13 @@ namespace clad {
             return false; // Tells clang not to create the plugin.
           }
         }
+        if (m_DO.ValidateClangVersion != false) {
+          std::string runtimeVersion = clang::getClangFullCPPVersion();
+          std::string builtVersion = CLANG_FULL_VERSION;
+          if (runtimeVersion.find(builtVersion.substr(0, 6)) ==
+              std::string::npos)
+            return false;
```
vgvassilev, line 175: I'd propose to move this code in a routine called `CheckCladVersion` or similar in ClangPlugin.cpp and call it from here. There you need to add in the CMakeLists.txt `set_source_files_properties(ClangPlugin.cpp PROPERTIES COMPILE_DEFINITIONS "CLANG_VERSION_MAJOR=\"${CLANG_VERSION_MAJOR}\"")` to expose the necessary macro.

Ris-Bali: Okay thanks understood.

Ris-Bali:
I've made the requested changes and squashed the commit. Pls review.

tools/CMakeLists.txt
```diff
@@ -92,3 +92,4 @@ if (NOT CLAD_BUILD_STATIC_ONLY)
     add_dependencies(clad LLVMEnzyme)
   endif(ENABLE_ENZYME_BACKEND)
 endif()
+set_source_files_properties(ClangPlugin.cpp PROPERTIES COMPILE_DEFINITIONS CLANG_MAJOR_VERSION="${CLANG_VERSION_MAJOR}.")
```
vgvassilev, line 95:
```suggestion
set_source_files_properties(ClangPlugin.cpp PROPERTIES COMPILE_DEFINITIONS CLANG_MAJOR_VERSION="${CLANG_VERSION_MAJOR}")
```

tools/ClangPlugin.h
```diff
@@ -184,6 +169,10 @@ namespace clad {
             return false; // Tells clang not to create the plugin.
           }
         }
+      if (m_DO.ValidateClangVersion != false) {
+        if(checkClangVersion() == 0)
```
vgvassilev, line 173:
```suggestion
        if (!checkClangVersion()) 
```

vgvassilev (approved):
LGTM! Please run git-clang-format on this patch and I can merge it in.

PR #543, opened 2023-03-15
Fix isUnusedResult to work with clad::array

### LINKED ISSUE #350: Fix extra lines generated when using clad::array.
Clad uses the [`isUnusedResult`](https://github.com/vgvassilev/clad/blob/edb192554fb3b45a5ebaa786227847779e16b4a0/lib/Differentiator/VisitorBase.cpp#L197) function to check for unneccessary statements and remove them from the generated code.

Since the array subscript expression of `clad::array` calls it's operator overload the function `isUnusedResult` is not able to determine that statement is not required.

```
double arr(3);
arr[1]; // <- isUnusedResult returns true for this statement

clad::array a(3);
a[1]; // <- isUnusedResult returns false for this statement but it should return true
```

To fix this `isUnusedResult` needs to be modified to return true for such statements.

Daksh1603:
Fixes #350 

This pull request addresses an issue with the isUnusedResult function in CLAD, which currently does not correctly identify unused results for clad::array expressions. 

To fix this issue , I have made some minor changes to the isUnusedResult function from  the clad/Differentiator/VisitorBase.cpp file which explicity checks whether the passed  expression is an instance of 'clad::array'.

### DIFF
--- a/lib/Differentiator/VisitorBase.cpp
+++ b/lib/Differentiator/VisitorBase.cpp
@@ -39,8 +39,8 @@ namespace clad {
     const Expr* ignoreExpr;
     SourceLocation ignoreLoc;
     SourceRange ignoreRange;
-    if(auto array = dyn_cast<clad::array>(E))
-    	return false;
+    if (auto array = dyn_cast<ArraySubscriptExpr>(E))
+    	return !array->getBase()->isLValue();
     return E->isUnusedResultAWarning(
         ignoreExpr, ignoreLoc, ignoreRange, ignoreRange, m_Context);
   }
--- a/lib/Differentiator/VisitorBase.cpp
+++ b/lib/Differentiator/VisitorBase.cpp
@@ -39,6 +39,8 @@ namespace clad {
     const Expr* ignoreExpr;
     SourceLocation ignoreLoc;
     SourceRange ignoreRange;
+    if(auto array = dyn_cast<clad::array>(E))
+    	return false;
     return E->isUnusedResultAWarning(
         ignoreExpr, ignoreLoc, ignoreRange, ignoreRange, m_Context);
   }


ShounakDas101:
what is this  "  Expr   " type?

parth-07:
@Daksh1603 I was expecting some test changes as well that I don't see.

Can you also please answer to @ShounakDas101 questions?

Daksh1603:
Hi @ShounakDas101 ,

Apologies for the late reply , as I had my mid semester test today .
I don't think the code snippet you mentioned is the right way to replicate this error . In my opinion , for you to replicate this error in it's entirety , try running the following code :
```
#include <iostream>
#include "clad/Differentiator/Array.h"

double fn(double* a) {
  	clad::array<int> b(3);
	b[1];
	double arr1[3];
	arr1[1];
        return a[1];
}

int main() {
  clad::array<double> a(10); 
  a[2];
  double x = 5.0;
  auto darray_dx = clad::differentiate(fn,"a[0]");
  std::cout<<darray_dx.execute(a)<<"\n"; 
  return 0;
}
```
However , this is what I think is right and may not be absolutely correct .

Daksh1603:
Hi @parth-07 @vgvassilev ,
As I dwell into this problem more , I am discovering more vulnerabilities in my commit . The earlier solution which I implemented doesn't seem to work for me , but I couldn't find the flaw . I have 2 possible questions and wanted to ask you for help ->

- I have two separate build directories for the CLAD plugin . One is the standard CLAD build and the other is my own forked CLAD build , to where I make changes . Now in order for me to inspect the changes I have made and whether the issue is really solved , I need to understand how can I swap between these two builds at ease . Usually , I just end up running the `make && make install` command for respective build directory . But perhaps , this might be the wrong way to go about it .
- Secondly , I tried implementing an alternate solution for this , please have a look ->
``` 
if (auto array = dyn_cast<ArraySubscriptExpr>(E)) {
        if (auto variable = dyn_cast<VarDecl>(array->getBase()->getReferencedDeclOfCallee())) {
                if (variable->getNameAsString()== "clad::array") {
                    return !array->getBase()->isLValue();
                }
            }
        }
```
This seems like a reasonable approach but what concerns me is the fact that getReferencedDeclOfCallee() might end up pointing to the operator which is being overloaded instead . A little more guidance will be appreciated on this and I am confident that Ill end up fixing it at the earliest .

parth-07:
> * I have two separate build directories for the CLAD plugin . One is the standard CLAD build and the other is my own forked CLAD build , to where I make changes . Now in order for me to inspect the changes I have made and whether the issue is really solved , I need to understand how can I swap between these two builds at ease . Usually , I just end up running the `make && make install` command for respective build directory . But perhaps , this might be the wrong way to go about it .

Why do you need two builds?
Ideally, you do not need two builds at all. You can have one build with multiple remotes.
And did you globally install Clad in your system? I would recommend locally installing Clad. That generally makes
things more clear. And this way, if you have multiple builds, then you can choose which Clad to use when running a program with Clad plugin, by passing the correct path to the Clad plugin library when running the clang command.

> * Secondly , I tried implementing an alternate solution for this , please have a look ->
> 
> ```
> if (auto array = dyn_cast<ArraySubscriptExpr>(E)) {
>         if (auto variable = dyn_cast<VarDecl>(array->getBase()->getReferencedDeclOfCallee())) {
>                 if (variable->getNameAsString()== "clad::array") {
>                     return !array->getBase()->isLValue();
>                 }
>             }
>         }
> ```
> 
> This seems like a reasonable approach but what concerns me is the fact that getReferencedDeclOfCallee() might end up pointing to the operator which is being overloaded instead . A little more guidance will be appreciated on this and I am confident that Ill end up fixing it at the earliest .

This solution has some major flaws. It seems it would only work for `clad::array`. Do pure array types have the same issue as of `clad::array`? If yes, then we should fix that as well. Here the important thing to check is, if an expression has some side effects, it does then we want to keep it. If not, then we can skip outputting them. I suspect there may be some Clang API to check if an expression has side-effects, but I am not sure at the moment.

vgvassilev:
Let's close this PR as the issue was fixed in #797.

PR #544, opened 2023-03-16
For issue#526

ShounakDas101:
changing m_Code so it no longer has goto redundancy

### DIFF
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -163,6 +163,11 @@ namespace clad {
         m_Function = nullptr;
         m_Code = nullptr;
       }
+      std::string str(m_Code);
+      std::regex pattern("goto _label[0-9]+;\\n\\s*_label[0-9]+:");
+      str = std::regex_replace(str, pattern, "_label$1");
+      strcpy(m_Code, str.c_str());
+      printf("%s",m_Code);
     }
     /// Constructor overload for initializing `m_Functor` when functor
     /// is passed by reference.


ShounakDas101:
Basically looking for "goto _labelx;\n_labelx:" and changing it to _labelx

ShounakDas101:
Apparently it seems that the derivedFn is changed but I am still not sure

include/clad/Differentiator/Differentiator.h
```diff
@@ -163,6 +163,11 @@ namespace clad {
         m_Function = nullptr;
         m_Code = nullptr;
       }
+      std::string str(m_Code);
+      std::regex pattern("goto _label[0-9]+;\\n\\s*_label[0-9]+:");
+      str = std::regex_replace(str, pattern, "_label$1");
+      strcpy(m_Code, str.c_str());
+      printf("%s",m_Code);
```
vgvassilev, line 170: The point is to not fix the string but the underlying issue which is here: https://github.com/vgvassilev/clad/blob/7c801113c53dabf95aed9bf993e2501c86cf6616/lib/Differentiator/ReverseModeVisitor.cpp#L1182

Basically, when we have no statements we should not create that AST node.

ShounakDas101:
Whenever we encounter return statement we enter ReverseModeVisitor::VisitReturnStmtReverseModeVisitor::VisitReturnStmt()
 function.


For example:


double f_decls3(double x, double y) {
  double a = 3 * x;
  double c = 333 * y;
  if (x > 1)
    return 2 * a;     <--(conditional return)----goto- label is necessary
  else if (x < -1)
    return -2 * a;    <--(conditional return)----goto- label is necessary
  double b = a * a;
  return b;           <--(non-conditional return) --goto-label is NOT necessary 
}



if (!isCladValueAndPushforwardType(type)) {
      // Since returned expression may have some side effects affecting reverse
      // computation (e.g. assignments), we also have to emit it to execute it.
      Expr* retDeclRefExpr =
          StoreAndRef(ExprDiff.getExpr(), direction::forward,
                      utils::ComputeEffectiveFnName(m_Function) + "_return",
                      /*forceDeclCreation=*/true);



this part creates the f_decls3_returnX variable which is to be differentiated.

!isCladValueAndPushforwardType(type) detects when there is "return" statement
in fn code to be differentiated

We have to detect either there is conditional return OR non conditional return as explained in the above example.
Based on that we should generate "goto-label".

ShounakDas101:
@parth-07 @vgvassilev 

As I inspected the codes under: StmtDiff ReverseModeVisitor::VisitReturnStmt(const ReturnStmt* RS); I wrote a small test code as below:

double f_decls3(double x, double y) {
double b;
double a = 3 * x;
double c = 333 * y;
return = a * a; <--(non-conditional return) --goto-label is necessary

if (x > 1)
return 2 * a; <--(conditional return)----goto- label is necessary

else if (x < -1)
return -2 * a; <--(conditional return)----goto- label is necessary
}


 
If someone writes the code above, the conditional statement will be unreachable. This mistake generates extra redundant code (for conditional return statements) when compiled, but it will never be used. Additionally, a "goto" label cannot be removed for non-conditional statements. It seems better to insert "gradient_code" inside the "if" block itself, in order to completely eradicate the "goto _labelX;... _labelX" issue.

parth-07:
@ShounakDas101  Thank you very much for the analysis.

> If someone writes the code above, the conditional statement will be unreachable. This mistake generates extra redundant code (for conditional return statements) when compiled, but it will never be used. 

That's fine. Clad should not perform such kinds of optimizations. These kinds of optimizations are left up to the compiler.

> Additionally, a "goto" label cannot be removed for non-conditional statements. It seems better to insert "gradient_code" inside the "if" block itself, in order to completely eradicate the "goto _labelX;... _labelX" issue.

We cannot insert the gradient code inside the `if` block because of the nature of how the reverse mode works.
The important thing to remember is, the reverse-pass statements should only run if the corresponding forward-pass counterparts were run. Now how does this relate to `goto` and the problem at hand? Let's explore this with a slightly modified example of what you presented:

```cpp
double f_decls3(double x, double y) {
  double b;
  double a = 3 * x;
  double c = 333 * y;

  if ( x > 1)
    return a * a; 

  if (x > 2)
    return 2 * a; 

  if (x > 3)
    return -2 * a;
  
  return a;
}
```

Let's see it's derivative code:

```cpp
double f_decl3_grad(double x, double y, double *_d_x, double *_d_y) {
  // Zero-derivatives and global temporary variables declarations 
  double _d_b = 0;
  double _d_a = 0;
  double _d_c = 0;
  bool _t0 = 0;
  bool _t1 = 0;
  bool _t2 = 0;
  double f_decl3_return;

  // forward-pass
  double a = 3 * x;
  double c = 333 * y;

  if ( x > 1) {
  _t0 = 1;
   f_decl3_return = a * a; // 1
   goto label1;
  }
  
  if ( x > 2) {
    _t1 = 1;
    f_decl3_return = 2 * a; // 2
    goto label2;
  }

  if ( x > 3) {
    _t2 = 1;
    f_decl3_return = -2 * a; // 3
    goto label3;
  }
  
  f_decl3_return = a; // 4
  goto label4;

  // reverse pass
  label4:
  // ... reverse-pass counterpart statements for (4)
  
  if (_t2) {
    label3:
    // ... reverse-pass counterpart statements for (3)
  }
  
  if (_t1) {
    label2:
    // ... reverse-pass counterpart statements for (2)
  }

  if (_t0) {
    label1:
    // ... reverse-pass counterpart statements for (1)
  }
  
  *_d_y += _d_c * 333;
  _d_c = 0;
  
  *_d_x += _d_a * 3;
  _d_a = 0;
}
```

Please note that the function has to skip some reverse-pass statements depending on which `if` condition is hit.
`goto` is used to transfer the control to the correct location in the reverse-pass code.  We cannot have the forward-pass and the reverse-pass code in the same `if` block.

In the code shown above, only one goto, `goto label4;`, is redundant. We do need to keep the other `goto`s unless we change the design significantly to replace `goto`s with some other mechanism to manage the control-flow. One such mechanism would be to use switch statements. Switch statements can work since they can act as a runtime goto statement, but they have their own issues.

PR #549, opened 2023-03-21
Fixes #526 STEP 1: In clad/include/clad/Differentiator/ReverseModeVis…

### LINKED ISSUE #526: Optimize away redundant goto statements
In many cases we have:
```cpp
...
goto _label0;
_label0:
...
```

We should optimize away this redundancy in the output and that would simplify taking the higher order derivative of the produced code.

ShounakDas101:
…itor.h (changes in line number 34 as given below)

class ReverseModeVisitor
private:
bool if_encounter = false; // this line added

STEP 2: In ReverseModeVisitor.cpp ( Changes in Line number 772 as given below)

StmtDiff ReverseModeVisitor::VisitIfStmt(const clang::IfStmt* If) { if_encounter = true; // this line added

STEP 3: In ReverseModeVisitor.cpp (Changes in line number 1181 as given below)

In StmtDiff ReverseModeVisitor::VisitReturnStmt(const ReturnStmt* RS)

if(if_encounter==true)
{
if_encounter=false;
return m_Sema.ActOnGotoStmt(noLoc, noLoc, LD).get(); }
else
{
return m_Sema.ActOnNullStmt(noLoc).get();
}

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -61,6 +61,7 @@ namespace clad {
     Stmts m_PopIdxValues;
     std::vector<Stmts> m_LoopBlock;
     unsigned outputArrayCursor = 0;
+    bool if_encounter = false;//Used to Remove redundant goto
     unsigned numParams = 0;
     bool isVectorValued = false;
     bool use_enzyme = false;
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -774,6 +774,8 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     // to this scope.
     beginScope(Scope::DeclScope | Scope::ControlScope);
 
+    if_encounter = true;
+
     StmtDiff cond = Clone(If->getCond());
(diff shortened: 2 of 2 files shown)


ShounakDas101:
@parth-07 @vgvassilev 
I was going through the errors given by codecov. For Ex:-
Screenshot from 2023-03-23 07-23-19
 but when I ran the file on my computer it did not give any error and all the outputs matched

parth-07:
Hi @ShounakDas101, is it possible that you modified the tests locally but forgot to push the test changes?

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -61,6 +61,7 @@ namespace clad {
     Stmts m_PopIdxValues;
     std::vector<Stmts> m_LoopBlock;
     unsigned outputArrayCursor = 0;
+    bool if_encounter = false;//Used to Remove redundant goto
```
parth-07, line 64: What if we have a return statement in a loop or switch body? Would we need similar member variables for loop and switch statements as well?

parth-07 (commented):
I would like to hold this pull request as-of now until we have a better solution.

vgvassilev:
Let's close this PR now and re-open when we have a better solution.

Issue #218, opened 2021-03-24
Incorrect type of clad tape in reverse mode


grimmmyshini:
Minimum reproducible example:

```cpp
float func(float z, int a) { 
  for (int i = 1; i < a; i++){
    z = z * a;
  }
  return z;
}
```
results in the following failed assertion 

```Assertion `(isGenericMethod || ((*I)->isVariablyModifiedType() || (*I).getNonReferenceType()->isObjCRetainableType() || getContext() .getCanonicalType((*I).getNonReferenceType()) .getTypePtr() == getContext() .getCanonicalType((*Arg)->getType()) .getTypePtr())) && "type mismatch in call argument!"' failed.```

The possible culprit here is revealed while looking at the code dump

```cpp
void func_grad(float z, int a, float *_result) {
    unsigned long _t0;
    int _d_i = 0;
    clad::tape<float> _t1 = {};     <--- the type here should be int instead of float as it stores an int 
    clad::tape<float> _t2 = {};
    _t0 = 0;
    for (int i = 1; i < a; i++) {
        _t0++;
        z = clad::push(_t2, z) * clad::push(_t1, a);   <-- _t1 storing an int here
    }
    float func_return = z;
    goto _label0;
  _label0:
    _result[0UL] += 1;
    for (; _t0; _t0--) {
        {
            float _r_d0 = _result[0UL];
            float _r0 = _r_d0 * clad::pop(_t1);
            _result[0UL] += _r0;
            float _r1 = clad::pop(_t2) * _r_d0;
            _result[1UL] += _r1;
            _result[0UL] -= _r_d0;
        }
    }
}
``` 
The problem here is the fact that the tape is initialized with a different type that is being pushed to it. This happens because we initialize the tape with a variable wrapped under implicit casts (the cast type is dependent on the expression result and type promotion) hence we miss the actual type of the variable. There are two possible solutions to this problem:

1) Strip all implicit casts before deciding the type of the tape

2) Build another implicit cast while pushing to the tape

Out of which, the first options seems the suitable solution to me.

May be related to #214

sudo-panda:
@grimmmyshini yeah the culprit is an `ImplicitCastExpr` which we can verify when we call dump on the expression on which `MakeCladTapeFor` is used:
```
ImplicitCastExpr 0x5555629ce030 'float' <IntegralToFloating>
`-ImplicitCastExpr 0x5555629ce018 'int' <LValueToRValue>
[...]

PR #552, opened 2023-04-01
Fix incorrect type of clad tape in reverse mode

PhrygianGates:
#218 
The solution is that, when we create the clad tape, we extract from ImplicitCastExpr to get the real type of DeclRefExpr.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -64,6 +64,9 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
   ReverseModeVisitor::CladTapeResult
   ReverseModeVisitor::MakeCladTapeFor(Expr* E, llvm::StringRef prefix) {
     assert(E && "must be provided");
+    if (auto IE = dyn_cast<ImplicitCastExpr>(E)) {
+      E = IE->getSubExpr()->IgnoreImplicit();
+    }
     QualType EQt = E->getType();
     if (dyn_cast<ArrayType>(EQt))
       EQt = GetCladArrayOfType(utils::GetValueType(EQt));
--- a/test/Gradient/TestTypeConversion.C
+++ b/test/Gradient/TestTypeConversion.C
@@ -1,0 +1,60 @@
+// RUN: %cladnumdiffclang -lm -lstdc++ %s  -I%S/../../include -oGradients.out 2>&1 | FileCheck %s
+// RUN: ./Gradients.out | FileCheck -check-prefix=CHECK-EXEC %s
+
+//CHECK-NOT: {{.*error|warning|note:.*}}
+
+#include "clad/Differentiator/Differentiator.h"
+#include <cmath>
+
+#include "../TestUtils.h"
+
+float fn_type_conversion(float z, int a) { 
+  for (int i = 1; i < a; i++){
+    z = z * a;
+  }
+  return z;
+}
+
+void fn_type_conversion_grad(float z, int a, clad::array_ref<float> _d_z, clad::array_ref<int> _d_a);
+// CHECK: void fn_type_conversion_grad(float z, int a, clad::array_ref<float> _d_z, clad::array_ref<int> _d_a) {
+// CHECK-NEXT:     unsigned long _t0;
+// CHECK-NEXT:     int _d_i = 0;
+// CHECK-NEXT:     clad::tape<int> _t1 = {};
+// CHECK-NEXT:     clad::tape<float> _t2 = {};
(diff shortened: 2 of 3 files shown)


vgvassilev (commented):
The approach looks good to me. Can you squash the commits, improve the commit message and add "Fixes #218"?

PhrygianGates:
> The approach looks good to me. Can you squash the commits, improve the commit message and add "Fixes #218"?

Yeah, of course

.gitignore
```diff
@@ -1,2 +1,4 @@
 /build
 /.vscode
+/build_dir
```
parth-07, line 3: I think this should be in a separate change.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -64,6 +64,9 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
   ReverseModeVisitor::CladTapeResult
   ReverseModeVisitor::MakeCladTapeFor(Expr* E, llvm::StringRef prefix) {
     assert(E && "must be provided");
+    if (dyn_cast<ImplicitCastExpr>(E)) {
```
parth-07, line 67:
```suggestion
    if (auto IE = dyn_cast<ImplicitCastExpr>(E)) {
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -64,6 +64,9 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+      E = dyn_cast<ImplicitCastExpr>(E)->getSubExpr()->IgnoreImplicit();
```
parth-07, line 68: E = IE->getSubExpr()->IgnoreImplicit();

test/Gradient/TestTypeConversion.C
```diff
@@ -0,0 +1,59 @@
...
+    result_1 = 0;                                                             \
+    clad::gradient(F);                                                         \
+    F##_grad(x, y, &result_0, &result_1);                                    \
+    printf("Result is = {%.2f, %.2f}\n", result_0, (float)result_1);                \
+  }
+
+int main() {
+  float result_0;
+  int result_1;
+  TEST(fn_type_conversion, 4, 3); // CHECK-EXEC: Result is = {9.00, 24.00}
```
parth-07, line 58:
```suggestion
  TEST_GRADIENT(fn_type_conversion, /*numOfDerivativeArgs=*/2, 4, 3, &result_0, &result_1); // CHECK-EXEC: Result is = {9.00, 24.00}
```

`TEST_GRADIENT` is defined in `TestUtils.h`.

.gitignore
PhrygianGates: > I think this should be a separate change.

okay!

lib/Differentiator/ReverseModeVisitor.cpp
PhrygianGates: > 

I see

test/Gradient/TestTypeConversion.C
PhrygianGates: > 



> `TEST_GRADIENT` is defined in `TestUtils.h`.

I tried but it shows some error message like:
```
/home/phrygiangates/gsoc/clad/test/Gradient/../TestUtils.h:92:5: error: member reference base type 'void (*)(float, int, clad::array_ref<float>, clad::array_ref<int>)' is not a structure or union
  cf.execute(args...);
```

parth-07: Yes.

Change:
```cpp
TEST(fn_type_conversion, 4, 3);
```
this to:

```cpp
INIT_GRADIENT(fn_type_conversion);
TEST_GRADIENT(fn_type_conversion, /*numOfDerivativeArgs=*/2, 4, 3, &result_0, &result_1); // CHECK-EXEC: {9.00, 24}
```

And also, add an overload `void print(float f)` in `TestUtils.h`. You just have to add the below code in TestUtils.h

```cpp
void print(float f) { printf("%.2f", f); }
```

parth-07 (approved):
LGTM, other than changing `TEST` to `TEST_GRADIENT`. It's best to use a single testing method instead of creating test macros in each file.

vgvassilev:
Can you also squash the commits into one?

test/Gradient/TestTypeConversion.C
PhrygianGates: Thank you so much!

PhrygianGates:
> Can you also squash the commits into one?

Yeah!

vgvassilev (approved):
LGTM!

PR #557, opened 2023-04-19
Fill clad::array with 0s when assigned an empty brace-init list.

grimmmyshini:
Previously, assigning an empty brace-init list to a ```clad::array``` would cause a segfault due to the overloaded ```operator=(T* arr)``` (as ```{}``` is interpreted as a ```nullptr```). This PR modifies the overloaded operator to fill the array with 0s when it is assigned an empty init list.


parth-07 (approved):
Looks good.

Issue #441, opened 2022-05-01
Incorrect derivative produced when array is passed to call expression inside a loop


parth-07:
Minimal reproducible example:

```cpp
void modify(double& elem, double val) {
  elem = val;
}

double fn(double i, double j) {
  double arr[5] = {};
  for (int idx = 0; idx < 5; ++idx) {
    modify(arr[idx], i);
  }
  return arr[0] + arr[1] + arr[2] + arr[3] + arr[4];
}

int main() {
  auto fn_grad = clad::gradient(fn);
  double d_i, d_j;
  d_i = d_j = 0;
  fn_grad.execute(3, 5, &d_i, &d_j);
  std::cout << "d_i: " << d_i << "\n";
}
```

**Expected output:** d_i: 5
**Actual output:** d_i: 1

Derived functions generated by Clad: 
```cpp
void modify_pullback(double &elem, double val, clad::array_ref<double> _d_elem, clad::array_ref<double> _d_val) {
    elem = val;
    {
        double _r_d0 = * _d_elem;
        * _d_val += _r_d0;
        * _d_elem -= _r_d0;
        * _d_elem;
    }
}
void fn_grad(double i, double j, clad::array_ref<double> _d_i, clad::array_ref<double> _d_j) {
    clad::array<double> _d_arr(5UL);
    unsigned long _t0;
    int _d_idx = 0;
    clad::tape<int> _t1 = {};
    clad::tape<double> _t3 = {};
    clad::tape<int> _t4 = {};
    clad::tape<double> _t6 = {};
    double arr[5] = {};
    _t0 = 0;
    for (int idx = 0; idx < 5; ++idx) {
        _t0++;
        clad::push(_t3, arr[clad::push(_t1, idx)]);
        modify(arr[clad::push(_t4, idx)], clad::push(_t6, i));
    }
    double fn_return = arr[0] + arr[1] + arr[2] + arr[3] + arr[4];
    goto _label0;
  _label0:
    {
        _d_arr[0] += 1;
        _d_arr[1] += 1;
        _d_arr[2] += 1;
        _d_arr[3] += 1;
        _d_arr[4] += 1;
    }
    for (; _t0; _t0--) {
        {
            double _r1 = clad::pop(_t3);
            double _grad1 = 0.;
            modify_pullback(_r1, clad::pop(_t6), &_d_arr[_t2], &_grad1); <------------- Interesting line
            double _r0 = _d_arr[_t2];
            double _r2 = _grad1;
[...]

Issue #465, opened 2022-06-27
Incorrect derivative produced when array derivative elements needs to be accessed in expressions without any corresponding dfdx


parth-07:
Reproducible example:
```cpp
double sq(double& elem) {
  elem = elem * elem;
  return elem;
}

double fn(double *arr, int n) {
  double res = 0;
  for (int i=0; i<n; ++i) {
    res += sq(arr[i]);
  }
  return res;
}

int main() {
auto fn_grad = clad::gradient(fn, "arr");
  double arr[5] = {1, 2, 3, 4, 5};
  double darr[5] = {};
  clad::array_ref<double> ref(darr, 5);
  fn_grad.execute(arr, 5, ref);
  for (int i=0; i < 5; ++i)
    std::cout<<darr[I]<<" ";
}
```

Actual Output: `4282 0 0 0 0`
Expected Output: `2 4 6 8 10`

ro4i7:
Hello @parth-07 @vgvassilev 

The issue seems to be with how Clad handles array derivatives when the derivative elements need to be accessed in expressions without any corresponding `dfdx`. In this case, the array elements are modified within the `sq` function, and the derivative with respect to them needs to be computed correctly.

We can use the `CLAD_DERIVE` macro to explicitly define the derivative of the `sq` function. This can be done as follows:
```
CLAD_DEFINE_DERIVATIVE(sq, (double& elem), {
  return 2 * elem;
});
```
Here, we define the derivative of `sq` to be `2 * elem`. This corresponds to the derivative of `elem * elem` with respect to `elem`.

With this change, the expected output is obtained.
```
double sq(double& elem) {
  elem = elem * elem;
  return elem;
}

CLAD_DEFINE_DERIVATIVE(sq, (double& elem), {
  return 2 * elem;
});

double fn(double *arr, int n) {
  double res = 0;
  for (int i=0; i<n; ++i) {
    res += sq(arr[i]);
  }
  return res;
}

int main() {
  auto fn_grad = clad::gradient(fn, "arr");
  double arr[5] = {1, 2, 3, 4, 5};
  double darr[5] = {};
  clad::array_ref<double> ref(darr, 5);
  fn_grad.execute(arr, 5, ref);
  for (int i=0; i < 5; ++i)
    std::cout<<darr[i]<<" "; // outputs "2 4 6 8 10"
}
```
By defining the derivative of `sq`, we ensure that Clad computes the correct derivative with respect to `arr[i]`

[...]

PR #560, opened 2023-04-25
Fixes incorrect derivative produced when array is passed to call expression inside a loop

ShounakDas101:
Added support for passing by reference in a for loop. In VisitArraySubscriptExpr(), for loops , added condition to check if it has reference variable. Fixes issue #441 and #465

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -59,6 +59,7 @@ namespace clad {
     /// to maintain the correct statement order when the current block has
     /// delayed emission i.e. assignment LHS.
     Stmts m_PopIdxValues;
+    Stmts m_RPopIdxValues;//for reference variables as input in function argumentlist
     std::vector<Stmts> m_LoopBlock;
     unsigned outputArrayCursor = 0;
     unsigned numParams = 0;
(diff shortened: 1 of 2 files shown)


ShounakDas101:
@vgvassilev 
If this process is incorrect then please close it. Recently I started working on both issues. Please let me know what are the errors generated by this PR.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -392,6 +393,33 @@ namespace clad {
     /// https://github.com/vgvassilev/clad/issues/385
     clang::QualType GetParameterDerivativeType(clang::QualType yType,
                                                clang::QualType xType);
+                                               
+                                               
+    bool Ref=false;
+	  bool hasReferenceType(const clang::QualType& type) {
+  		return type->isReferenceType();
+	  }
+	  bool printArgTypes(const clang::CallExpr* CE) {
```
parth-07, line 402: Why is this function called `printArgTypes`?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1547,10 +1550,15 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
           block.insert(block.begin() + insertionPoint,
                        BuildDeclStmt(argDiffLocalVD));
           Expr* argDiffLocalE = BuildDeclRef(argDiffLocalVD);
-
+          int Numref=0;
+          while(!m_RPopIdxValues.empty())
```
parth-07, line 1554: What does `m_RPopIdxValues` represent?

parth-07 (commented):
Can you please briefly explain your strategy in the commit message?

lib/Differentiator/ReverseModeVisitor.cpp
ShounakDas101: @parth-07 It keeps the DeclStmt of the variables which were being generated because of having a Reference Variable in the argument list but  were not being added to the derived function. These variables were being generated in  VisitArraySubscriptExpr()

include/clad/Differentiator/ReverseModeVisitor.h
ShounakDas101: It was a prototype function which did print the argument types. I had done this on last day of GSoC application so it is a complete mess and forgot to change the name. This part can be shifted to CladUtils. This function can be simplified a lot. All it is trying to find if there is any ReferenceType. I am very sorry for this.

vgvassilev:
The issues that this pull request intended to fix are already fixed in the master.

Issue #509, opened 2022-11-23
Compute the hessian matrix diagonal only


vgvassilev:
In many cases we need only the diagonal for the optimizer seeding. We can avoid computing the full matrix as in problems with many parameters will be costly.

cc: @lmoneta.

vgvassilev:
@vaithak perhaps we can implement an option to the clad::jacobian and clad::hessian allowing the user to specify that they want only the diagonal. In turn we can change the differentiation plan to include only these instead of computing all and throwing them away.

PR #565, opened 2023-05-29
Add initial basic support for vectorized forward mode AD

vaithak:
This PR adds initial support for vectorized forward mode for differentiating functions w.r.t all parameters in forward mode.

Minimal working example: 
```cpp
  return x*y;
}

int main() {
  auto f_dx = clad::vector_forward_differentiate(f);
  double dx = 0, dy = 0;
  f_dx.execute(/*x=*/3, /*y=*/4, &dx, &dy);
  std::cout << "dx: " << dx << ' ' << "dy: " << dy << std::endl;
  
  // Dump the generated derivative code to standard output.
  f_dx.dump();
}
``` 

Corresponding output:
```bash
dx: 4 dy: 3
The code is: 
void f_d_all_args(double x, double y, double *_d_x, double *_d_y) {
    clad::array<double> _d_vector_x = {1., 0.};
    clad::array<double> _d_vector_y = {0., 1.};
    {
        clad::array<double> _d_vector_return = _d_vector_x * y + x * _d_vector_y;
        *_d_x = _d_vector_return[0];
        *_d_y = _d_vector_return[1];
        return;
    }
}
```


vaithak:
Hi, I need some help with the falling test cases; the test is failing for Clang 15 because it is generating extra pragmas related to floating point arithmetic with the code:
```cpp
void f1_d_all_args(double x, double y, clad::array_ref<double> _d_x, clad::array_ref<double> _d_y) { 
   clad::array<double> _d_vector_x = {1, 0}; 
   clad::array<double> _d_vector_y = {0, 1}; 
   { 
        #pragma clang fp exceptions() 
        #pragma STDC FENV_ROUND FE_DYNAMIC 
        clad::array<double> _d_vector_return = _d_vector_x * y + x * _d_vector_y; 
        * _d_x = _d_vector_return[0]; 
        * _d_y = _d_vector_return[1]; 
    } 
} 
```
Do you have any ideas or suggestions as to what I am missing here and how this can be fixed?

vgvassilev:
I remember we needed some adjustments in the recent clang versions for floating point arithmetic in our compatibility layer. Could you build llvm in debug and set a breakpoint at the time the pragma is produced. 

Cc: @alexander-penev

vaithak:
Thanks for the help, @vgvassilev, using [the custom defined function](https://github.com/vgvassilev/clad/blob/master/include/clad/Differentiator/Compatibility.h#L64) fixed the tests 👍🏼

include/clad/Differentiator/Array.h
```diff
@@ -192,6 +201,114 @@ template <typename T> class array {
       m_arr[i] /= arr[i];
     return *this;
   }
+  /// Multiplies the number to every element in the array and returns a new
+  /// array
+  CUDA_HOST_DEVICE array<T> operator*(T n) {
```
parth-07, line 206: Can it be a `const` member function?

include/clad/Differentiator/Array.h
```diff
@@ -192,6 +201,114 @@ template <typename T> class array {
...
+    array<T> arr(m_size);
+    for (std::size_t i = 0; i < m_size; i++)
+      arr[i] = m_arr[i] * n;
+    return arr;
+  }
+  /// Multiplies the number to every element in the array and returns a new
+  /// array, when the number is on the left side
+  CUDA_HOST_DEVICE friend array<T> operator*(T n, array<T>& arr) {
```
parth-07, line 214: Can we pass `arr` as a const reference instead? If yes, then please make changes in all such operator overloads.

include/clad/Differentiator/Array.h
```diff
@@ -192,6 +201,114 @@ template <typename T> class array {
...
+  CUDA_HOST_DEVICE friend array<T> operator+(array<T>& arr1, array<T>& arr2) {
+    assert(arr1.size() == arr2.size());
+    array<T> arr(arr1.size());
+    for (std::size_t i = 0; i < arr1.size(); i++)
+      arr[i] = arr1[i] + arr2[i];
+    return arr;
+  }
+  /// Friend function to define element wise adding of two arrays, one is
+  /// lvalue and other is rvalue.
+  CUDA_HOST_DEVICE friend array<T> operator+(array<T>& arr1, array<T>&& arr2) {
```
parth-07, line 251: Why do you have overloads that differ in the parameters `array<T> &arr2` and `array<T> &&arr2`?

include/clad/Differentiator/Array.h
```diff
@@ -192,6 +201,114 @@ template <typename T> class array {
...
+  CUDA_HOST_DEVICE friend array<T> operator+(array<T>&& arr1, array<T>& arr2) {
+    assert(arr1.size() == arr2.size());
+    array<T> arr(arr1.size());
+    for (std::size_t i = 0; i < arr1.size(); i++)
+      arr[i] = arr1[i] + arr2[i];
+    return arr;
+  }
+  /// Friend function to define element wise adding of two arrays, both are
+  /// rvalues.
+  CUDA_HOST_DEVICE friend array<T> operator+(array<T>&& arr1, array<T>&& arr2) {
```
parth-07, line 269: Can you please explain why overloads for rvalue references are beneficial here? You are not moving any value
within the function body.

include/clad/Differentiator/Differentiator.h
```diff
@@ -317,6 +320,22 @@ namespace clad {
...
+  /// Specialization for differentiating using vector mode.
+  /// The specialization is needed because the derivative function generated
+  /// by clad for vector mode is different from the one generated for forward
+  /// mode.
+  template <signed N = 1, typename ArgSpec = const char*, typename F,
+            typename DerivedFnType = ExtractDerivedFnTraitsVecForwMode_t<F>,
+            typename = typename std::enable_if<N==clad::opts::vector_fwd_mode &&
+                !std::is_class<remove_reference_and_pointer_t<F>>::value>::type>
+  CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>, true> __attribute__((
+      annotate("D")))
```
parth-07, line 332: `annotate("D")` is used to recognize the differentiation mode. We should have different differentiation mode for ordinary forward and vector forward mode.

parth-07 (changes requested):
Initial review.

parth-07:
@vaithak How can a user request for 4th order differentiation using vector forward mode? It seems that
`clad::opts::vector_fwd_mode` and differentiation order shares the same template parameter.

include/clad/Differentiator/Array.h
vaithak: fixed 👍🏼

include/clad/Differentiator/Array.h
vaithak: changed in all such operator overloads 👍🏼

include/clad/Differentiator/Array.h
vaithak: Yeah, my bad. I have fixed this by making changes in the rvalue reference itself.

include/clad/Differentiator/Array.h
vaithak: This is mainly used for cascaded operations. For ex. `clad::array<double> d = a+b+c`, where `a, b, c` are also clad arrays.

vaithak:
Hi @parth-07, vector forward mode will calculate derivative w.r.t all parameters. Ideally, this should mean calculating the Hessian matrix for n=2 and similarly for other higher-order derivatives. So, I think N>1 won't make much sense for vector fwd mode.

include/clad/Differentiator/Differentiator.h
vaithak: commented below about the reasoning for this.

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -203,6 +204,152 @@ namespace clad {
...
+
+    // Generate the function type for the derivative.
+    llvm::SmallVector<clang::QualType, 8> paramTypes;
+    paramTypes.reserve(m_Function->getNumParams() * 2);
+    for (auto PVD : m_Function->parameters()) {
+      paramTypes.push_back(PVD->getType());
+    }
+    for (auto PVD : m_Function->parameters()) {
+      QualType ValueType = utils::GetValueType(PVD->getType());
+      ValueType.removeLocalConst();
```
parth-07, line 270: What if the original type is a reference type?

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -957,10 +1146,21 @@ namespace clad {
     if (auto VD = dyn_cast<VarDecl>(clonedDRE->getDecl())) {
       // If DRE references a variable, try to find if we know something about
       // how it is related to the independent variable.
-      auto it = m_Variables.find(VD);
-      if (it != std::end(m_Variables)) {
+      clang::Expr* dExpr = nullptr;
+      if (m_VectorMode) {
+        // If we are in vector mode, use the vector of derivatives.
+        auto it = m_VectorVariables.find(VD);
```
parth-07, line 1152: Is there any use of `m_Variables` in vector forward mode?

parth-07 (changes requested):

parth-07:
> Hi @parth-07, vector forward mode will calculate derivative w.r.t all parameters. This should mean calculating the Hessian matrix for n=2 and similarly for other higher-order derivatives. So, I think N>1 won't make much sense for vector fwd mode. 

Hi @vaithak
Yes, vector forward mode will compute derivatives w.r.t all the parameters. For a second-order differentiation, wouldn't this mean computing the diagonal of the Hessian matrix instead of the matrix itself? We already have a requirement for such a case: issue #509. 

We shouldn't use the same function template parameter for entirely different use cases, even if we do not wish to support higher-order differentials for vector forward mode. We can use a different function altogether for vector forward mode, `clad::VectorForwardDifferentiation`. @vgvassilev @vaithak  What do you think about this?

include/clad/Differentiator/Array.h
```diff
@@ -192,6 +201,117 @@ template <typename T> class array {
...
+    for (std::size_t i = 0; i < m_size; i++)
+      arr[i] = m_arr[i] * n;
+    return arr;
+  }
+  /// Multiplies the number to every element in the array and returns a new
+  /// array, when the number is on the left side
+  CUDA_HOST_DEVICE friend array<T> operator*(T n, const array<T>& arr) {
+    return arr * n;
+  }
+  CUDA_HOST_DEVICE friend array<T> operator*(T n, array<T>&& arr) {
```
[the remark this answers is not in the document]
vgvassilev: Yes, please run `git-clang-format` to this patch.

include/clad/Differentiator/Array.h
vaithak: removed all rvalue reference functions for now to keep this simple, will add them later if required.

include/clad/Differentiator/Array.h
vaithak: removed rvalue reference function 👍🏼

include/clad/Differentiator/Array.h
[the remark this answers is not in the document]
vaithak: ran `git-clang-format` and added spaces between functions 👍🏼

include/clad/Differentiator/Array.h
```diff
@@ -192,6 +201,117 @@ template <typename T> class array {
...
+  CUDA_HOST_DEVICE array<T> operator+(T n) const {
+    array<T> arr(m_size);
+    for (std::size_t i = 0; i < m_size; i++)
+      arr[i] = m_arr[i] + n;
+    return arr;
+  }
+  CUDA_HOST_DEVICE friend array<T> operator+(T n, const array<T>& arr) {
+    return arr + n;
+  }
+  CUDA_HOST_DEVICE friend array<T> operator+(T n, array<T>&& arr) {
```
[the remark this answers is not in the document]
vaithak: removed 👍🏼

include/clad/Differentiator/ForwardModeVisitor.h
```diff
@@ -96,6 +102,20 @@ namespace clad {
...
+    //  vectorized forward mode.
+    ///
+    /// Information about the original function, derived function, derived
+    /// function parameter types and the differentiation mode are implicitly
+    /// taken from the data member variables. In particular, `m_Function`,
+    /// `m_Mode` and `m_Derivative` should be correctly set before using this
+    /// function.
+    llvm::SmallVector<clang::ParmVarDecl*, 8> BuildVectorModeParams(DiffParams& diffParams);
+
+    /// Get an expression used to initialize the one-hot vector for the
```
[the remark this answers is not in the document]
vaithak: explained in comments 👍🏼

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -203,6 +204,152 @@ namespace clad {
...
+                                              PVD->getStorageClass());
+      paramDerivatives.push_back(dPVD);
+      ++dParamTypesIdx;
+
+      if (dPVD->getIdentifier())
+            m_Sema.PushOnScopeChains(dPVD, getCurrentScope(),
+                                     /*AddToContext=*/false);
+
+      m_Variables[newPVD] = BuildOp(UO_Deref, BuildDeclRef(dPVD),
+                                       m_Function->getLocation());
```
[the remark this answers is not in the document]
vaithak: passed `noLoc` 👍🏼

lib/Differentiator/ForwardModeVisitor.cpp
vaithak: wouldn't `utils::GetValueType` take care of that?

lib/Differentiator/ForwardModeVisitor.cpp
vaithak: yup, `m_Variables` is for storing the mapping from original function parameters to new parameters for derivatives.

vaithak:
> > Hi @parth-07, vector forward mode will calculate derivative w.r.t all parameters. This should mean calculating the Hessian matrix for n=2 and similarly for other higher-order derivatives. So, I think N>1 won't make much sense for vector fwd mode.
> 
> Hi @vaithak Yes, vector forward mode will compute derivatives w.r.t all the parameters. For a second-order differentiation, wouldn't this mean computing the diagonal of the Hessian matrix instead of the matrix itself? We already have a requirement for such a case: issue #509.
> 
> We shouldn't use the same function template parameter for entirely different use cases, even if we do not wish to support higher-order differentials for vector forward mode. We can use a different function altogether for vector forward mode, `clad::VectorForwardDifferentiation`. @vgvassilev @vaithak What do you think about this?

I think this makes sense, will think more about this and let you know if I find some edge cases otherwise will add the new method for vector mode as you suggested.

vaithak:
@parth-07 I have created a new function for `clad::vector_forward_differentiate` instead of having an `opts` 👍🏼

lib/Differentiator/ForwardModeVisitor.cpp
```diff
@@ -221,6 +368,13 @@ namespace clad {
     if (DVI.empty())
       return {};
 
+    if (request.vector_fwd_mode) {
+      assert (DVI.size() == FD->getNumParams() &&
+              "Vector mode supports differentiating w.r.t. all parameters only.");
```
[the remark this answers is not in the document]
vaithak: created an issue: https://github.com/vgvassilev/clad/issues/566

lib/Differentiator/ForwardModeVisitor.cpp
parth-07: Oh, yes, It would.

include/clad/Differentiator/Array.h
```diff
@@ -192,6 +200,78 @@ template <typename T> class array {
...
+  }
+  CUDA_HOST_DEVICE friend array<T> operator-(T n, const array<T>& arr) {
+    array<T> arr2(arr.size());
+    for (std::size_t i = 0; i < arr.size(); i++)
+      arr2[i] = n - arr[i];
+    return arr2;
+  }
+
+  /// Friend function to define element wise adding of two arrays.
+  CUDA_HOST_DEVICE friend array<T> operator+(const array<T>& arr1,
```
parth-07, line 258: Why does it need to be a friend function? It does not seem to be accessing any private/protected members.
Also, what do you think about moving the operators outside the class?

include/clad/Differentiator/FunctionTraits.h
```diff
@@ -719,6 +719,25 @@ namespace clad {
   using ExtractDerivedFnTraitsForwMode_t =
       typename ExtractDerivedFnTraitsForwMode<F>::type;
 
+  template <class T, class R> struct OutputVecParamType {
+    using type = typename std::add_pointer<R>::type;
```
parth-07, line 723: Will it be correct if the return type is a `const` type?

parth-07 (changes requested):

include/clad/Differentiator/Array.h
```diff
@@ -40,6 +41,13 @@ template <typename T> class array {
 
   CUDA_HOST_DEVICE array(const array<T>& arr) : array(arr.m_arr, arr.m_size) {}
 
+  CUDA_HOST_DEVICE array(std::initializer_list<T> arr)
+      : m_arr(new T[arr.size()]{static_cast<T>(T())}), m_size(arr.size()) {
```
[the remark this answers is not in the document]
vaithak: I followed the convention as it is done for other constructors of this class.

include/clad/Differentiator/Array.h
vaithak: shifted most of the overloaded operators outside the class.

include/clad/Differentiator/FunctionTraits.h
vaithak: yup, it works even if return type is `const double` for example

test/ForwardMode/VectorMode.C
```diff
@@ -0,0 +1,55 @@
...
+//CHECK-NOT: {{.*error|warning|note:.*}}
+
+#include "clad/Differentiator/Differentiator.h"
+#include <cmath>
+
+double f1(double x, double y) {
+  return x*y;
+}
+
+void f1_d_all_args(double x, double y, double *_d_x, double *_d_y);
```
vgvassilev, line 13: Could we extend the test cases? We should probably add a demo (in the demos folder). We should also document the new feature and interface in the documentation under `docs`.

vgvassilev, line 13: Could we extend the test cases? We should probably add a demo (in the demos folder). We should also document the new feature and interface in the documentation under `docs`.

vaithak: Sure, I will add it in follow-up PRs when the implementation is a little stable for some decently complex programs.

vgvassilev: According to codecov this PR needs more tests to reach 100% coverage. I don’t think we need complex examples but examples good enough to test all new code that we have.

vaithak: added more test cases 👍🏼

include/clad/Differentiator/Array.h
```diff
@@ -192,9 +200,100 @@ template <typename T> class array {
...
-};
+}; // class array
+
+/// Overloaded operators for clad::array which return a new array.
+
+/// Multiplies the number to every element in the array and returns a new
+/// array.
+template <typename T>
+CUDA_HOST_DEVICE array<T> operator*(const array<T>& arr, T n) {
+  array<T> arr2(arr);
```
vgvassilev, line 232: Do we need an extra allocation here? Can we use `array_ref`?

include/clad/Differentiator/Array.h
```diff
@@ -192,9 +200,100 @@ template <typename T> class array {
...
+CUDA_HOST_DEVICE array<T> operator/(const array<T>& arr, T n) {
+  array<T> arr2(arr);
+  arr2 /= n;
+  return arr2;
+}
+
+/// Adds the number to every element in the array and returns a new array
+template <typename T>
+CUDA_HOST_DEVICE array<T> operator+(const array<T>& arr, T n) {
+  array<T> arr2(arr);
```
vgvassilev, line 256: Likewise.

include/clad/Differentiator/Array.h
vaithak: The objective of this function is to return a new array (with multiplied entries),  if we use array_ref, won't `arr`'s underlying array also get modified? 
Also, if a new array is needed to be returned, won't it anyway require allocating a new array?

parth-07 (approved):
Overall looks good. Can you please squash all the commits into one?

vaithak:
created a single commit 👍🏼

include/clad/Differentiator/Array.h
vaithak: This overloaded operator is outside the class definition, so we can't use `m_arr` here.

vaithak: @vgvassilev, can we mark this as resolved?

vgvassilev: Yep, thanks for clarifying!

PR #569, opened 2023-06-14
Fixes for overloaded arithmetic operations on clad::array

vaithak:
A minimal example which failed earlier:
```cpp
clad::array<double> arr(5UL);
arr += 0;
```
This resulted in: 
```sh
error: use of overloaded operator '+=' is ambiguous (with operand types 'clad::array<double>' and 'int')
```


vgvassilev (commented):
Can you add tests in `test/Misc/CladArray.C`?

PR #571, opened 2023-06-15
improve clad array overloaded operations

vaithak:

PR #570, opened 2023-06-15
Vector initialization improvements for vector forward mode AD

vaithak:
Improved initialization of vectors while differentiating variable declarations and added test cases to demonstrate vector mode support for programs with loops.

New ways of initializing `clad::array` is added by passing `(size, expression...)`. 
This is done to allow initializing the array by all constant values or expressions which evaluate to a single value; 
- for ex: `clad::array<double> arr(5UL, 1)` can be used to initialize a vector of size 5 with all entries as 1.

PR #572, opened 2023-06-16
Split ForwardMode into separate classes

vaithak:


vaithak:
will add more test cases to make codecov happy, but some of these will require changes in this PR: https://github.com/vgvassilev/clad/pull/571

vgvassilev:
@vaithak do you think you can write some tests for the missed branches that codecov reports. They are not due to this PR but we can right a historical wrong :)

parth-07 (commented):
@vaithak This looks great!

I will need a bit more time to review this pull request.

vaithak:
> @vaithak do you think you can write some tests for the missed branches that codecov reports. They are not due to this PR but we can right a historical wrong :)

Sure, I am working on this, that's why I was able to find issues with comma expressions and the ternary operator case 😅 , these didn't had tests till now for forward mode AD.

lib/Differentiator/CMakeLists.txt
```diff
@@ -24,7 +24,9 @@ add_llvm_library(cladDifferentiator
   ConstantFolder.cpp
   DerivativeBuilder.cpp
   DiffPlanner.cpp
+  BaseForwardModeVisitor.cpp
   ForwardModeVisitor.cpp
+  VectorForwardModeVisitor.cpp
```
parth-07, line 29: Minor: Please sort the filenames here.

parth-07 (changes requested):
It looks great overall.

lib/Differentiator/VectorForwardModeVisitor.cpp
```diff
@@ -0,0 +1,220 @@
...
+  // Add an empty return statement to the array of statements.
+  returnStmts.push_back(
+      m_Sema.ActOnReturnStmt(noLoc, nullptr, getCurrentScope()).get());
+
+  // Create a return statement from the compound statement.
+  Stmt* returnStmt = MakeCompoundStmt(returnStmts);
+  return StmtDiff(returnStmt);
+}
+
+VarDeclDiff VectorForwardModeVisitor::DifferentiateVarDecl(const VarDecl* VD) {
```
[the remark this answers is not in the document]
vaithak: the difference was in prefix assigned tot he derived variable, so created an extra overloaded function for this which takes a prefix as a parameter.

lib/Differentiator/CMakeLists.txt
vaithak: done 👍🏼

lib/Differentiator/VectorForwardModeVisitor.cpp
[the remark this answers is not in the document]
vaithak: Please ignore the previous comment, currently the difference is in just prefix and type of derived variable, but it will be more visible in [this PR](https://github.com/vgvassilev/clad/pull/570), where the initialisation style will also change for vector mode.
So, I think it is better to override this.

vaithak:
added test for most of the uncovered cases to increase coverage (ignoring some failure edge cases)

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -0,0 +1,1901 @@
...
+  } else if (incDiff.getExpr_dx() && incDiff.getExpr()) {
+    // If no declarations are required and only two Expressions are produced,
+    // join them with comma expression.
+    if (!isUnusedResult(incDiff.getExpr_dx()))
+      incResult = BuildOp(BO_Comma, BuildParens(incDiff.getExpr_dx()),
+                          BuildParens(incDiff.getExpr()));
+    else
+      incResult = incDiff.getExpr();
+  } else if (incDiff.getExpr()) {
+    incResult = incDiff.getExpr();
```
vgvassilev, line 548: Don't we want to write a test for this branch?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -0,0 +1,1901 @@
...
+
+  StmtDiff Result =
+      (Block->size() == 1) ? StmtDiff(forStmtDiff) : StmtDiff(Block);
+  return Result;
+}
+
+StmtDiff BaseForwardModeVisitor::VisitReturnStmt(const ReturnStmt* RS) {
+  // If there is no return value, we must not attempt to differentiate
+  if (!RS->getRetValue())
+    return nullptr;
```
vgvassilev, line 585: That should be easy to test with having a `void f(double d) { return; }`?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -0,0 +1,1901 @@
...
+        ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
+    return StmtDiff(clonedME, zero);
+  } else {
+    auto zero =
+        ConstantFolder::synthesizeLiteral(m_Context.DoubleTy, m_Context, 0);
+    auto baseDiff = Visit(ME->getBase());
+    // No derivative found for base. Therefore, derivative is 0.
+    if (isa<IntegerLiteral>(baseDiff.getExpr_dx()) ||
+        isa<FloatingLiteral>(baseDiff.getExpr_dx()))
+      return {clonedME, zero};
```
vgvassilev, line 623: @parth-07 do you have an idea how to test this case?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -0,0 +1,1901 @@
...
+        auto result_at_i = BuildArraySubscript(it->second, clonedIndices);
+        return StmtDiff{cloned, result_at_i};
+      }
+
+      VD = decl;
+    }
+  } else if (isa<MemberExpr>(clonedBase->IgnoreParenImpCasts())) {
+    auto derivedME = cloneDiff.getExpr_dx();
+    if (!isa<MemberExpr>(derivedME->IgnoreParenImpCasts())) {
+      return {cloned, zero};
```
vgvassilev, line 689: @parth-07 and here?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -0,0 +1,1901 @@
...
+    }
+    auto derivedAS = BuildArraySubscript(derivedME, clonedIndices);
+    return {cloned, derivedAS};
+  } else {
+    if (!isa<DeclRefExpr>(clonedBase->IgnoreParenImpCasts()))
+      return StmtDiff(cloned, zero);
```
vgvassilev, line 695: @parth-07 and probably here?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -0,0 +1,1901 @@
...
+  // numerical diff to use correct declaration context.
+  if (forCustomDerv) {
+    DeclContext* outermostDC = utils::GetOutermostDC(m_Sema, originalFnDC);
+    // FIXME: We should ideally construct nested name specifier from the
+    // found custom derivative function. Current way will compute incorrect
+    // nested name specifier in some cases.
+    if (outermostDC &&
+        outermostDC->getPrimaryContext() == NSD->getPrimaryContext()) {
+      utils::BuildNNS(m_Sema, originalFnDC, SS);
+      DC = originalFnDC;
```
vgvassilev, line 873: @grimmmyshini do you have an idea how to test this branch?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -0,0 +1,1901 @@
...
+    return StmtDiff(op, BuildOp(opKind, diff.getExpr_dx()));
+  } else {
+    // This only adds support for differentiating the following expression:
+    // ```
+    // *this;
+    // ```
+    if (opKind == UnaryOperatorKind::UO_Deref) {
+      if (isa<CXXThisExpr>(UnOp->getSubExpr()->IgnoreParenImpCasts())) {
+        Expr* derivedE = BuildOp(opKind, diff.getExpr_dx());
+        return {op, derivedE};
```
vgvassilev, line 1136: @parth-07 and here?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -0,0 +1,1901 @@
...
+  return nullptr;
+}
+
+static void setSwitchCaseSubStmt(SwitchCase* SC, Stmt* subStmt) {
+  if (auto caseStmt = dyn_cast<CaseStmt>(SC)) {
+    caseStmt->setSubStmt(subStmt);
+  } else if (auto defaultStmt = dyn_cast<DefaultStmt>(SC)) {
+    defaultStmt->setSubStmt(subStmt);
+  } else {
+    assert(0 && "Unsupported switch case statement");
```
vgvassilev, line 1522: Is that reachable at all?

vgvassilev:
@vaithak did you run clang-format on this patch?

lib/Differentiator/BaseForwardModeVisitor.cpp
parth-07: This example reaches this code:

```cpp

struct Array {
    double data[5];
};

Array array;

    return array.data[0] * i + array.data[1] * j;
}

int main() {
    auto d_fn = clad::differentiate(fn, "i");
}
```

lib/Differentiator/BaseForwardModeVisitor.cpp
parth-07: It seems that this `if` block can be safely removed. We later added proper support for dereference operator, which made this special case redundant.

lib/Differentiator/BaseForwardModeVisitor.cpp
parth-07: It's an assert. Hopefully, it's not reachable :).

lib/Differentiator/BaseForwardModeVisitor.cpp
parth-07: I cannot think of a case for this right now. But it might be possible in some obscure cases.

lib/Differentiator/BaseForwardModeVisitor.cpp
vaithak: I couldn't find any expression such that the original expression is not null, but it's derivative expression is null.
Can this branch be removed ? cc: @parth-07

vaithak:
> @vaithak did you run clang-format on this patch?

yup 👍🏼

lib/Differentiator/BaseForwardModeVisitor.cpp
vaithak: Thanks for the example, added this as a test.

lib/Differentiator/BaseForwardModeVisitor.cpp
vaithak: removed this condition

lib/Differentiator/BaseForwardModeVisitor.cpp
parth-07: No, this cannot be removed. This example reaches this branch:

```cpp

void increment(int &i) {
    ++i;
}

    double res = 0;
    for (int i=0; i < 5; increment(i))
        res += i*j;
    return res;
}

int main() {
    auto d_fn = clad::differentiate(fn, "i");
}
```

lib/Differentiator/VectorForwardModeVisitor.cpp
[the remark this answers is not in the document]
parth-07: Okay, got it.

parth-07 (approved):
Looks good overall. Please add the remaining test cases if any and squash the commits into one.

lib/Differentiator/BaseForwardModeVisitor.cpp
vgvassilev: Perhaps also issue an unsupported diagnostics?

lib/Differentiator/BaseForwardModeVisitor.cpp
vaithak: Oh, nice! I wrongly tried using return directly as it has a derivative expression as null.
Thanks for this 👍🏼 , added.

lib/Differentiator/BaseForwardModeVisitor.cpp
vaithak: This is inside a static function, so `diag` can't be used here.

parth-07 (approved):
Thank you for the great work. LGTM.

vgvassilev (approved):
Great work!

PR #574, opened 2023-06-18
Fix derivative code for comma expressions in forward mode

### LINKED ISSUE #573: Incorrect results for comma expressions in forward mode AD
Minimal example
```cpp
double f(double x) {
   // returns (x+1)^2
   return (++x , x * x);
}
```

Theoretically expected derivative: `2*(x+1)`

Generated code results in calculation of `2*x` (i.e `++x` is missing before returning the derivative):
```cpp
double f_darg0(double x) {
    double _d_x = 1;
    return (_d_x * x + x * _d_x);
}
```

Similar, is the case with a slightly modified code for the same function:
```cpp
double f(double x) {
   // returns (x+1)^2
   double temp = (++x , x * x);
   return temp;
}
```

Generated code again results in calculation of `2*x` (this time `++x` is misplaced)
```cpp
double f_darg0(double x) {
    double _d_x = 1;
    double _d_temp = (_d_x * x + x * _d_x);
    double temp = (++x , x * x);
    return _d_temp;
}
```

vaithak:
Earlier for an expression like `(E1, E2)`, the derivative code produced `(dE1, dE2)`, 
thus a statement like `double temp = (++x, x*x)` produced:
```cpp
double _d_temp = (_d_x, _d_x * x + x * _d_x);
double temp = (++x, x*x);    <--- original cloned statement
```
This meant that the execution order is `dE1 -> dE2 -> E1 -> E2` (where -> means followed by).

But this doesn't seem right, as `dE2` can depend on `E1`; hence `E1` must be computed before `dE2`.
So, the execution order should be `dE1 -> E1 -> dE2 -> E2`.

This PR generates the derivative code as `(dE1, E1, dE2)`; the original statement is changed to just `E2`. 
Note that the result of both expressions is still the same as earlier (i.e. the derivative code will still have the result of `dE2`)

So, the example statement of computing `(x+1)^2`, will produce:
```cpp
double _d_temp = (_d_x, ++x , (_d_x * x + x * _d_x));
double temp = x*x;
```

closes #573

### DIFF
--- a/lib/Differentiator/ForwardModeVisitor.cpp
+++ b/lib/Differentiator/ForwardModeVisitor.cpp
@@ -1704,12 +1704,16 @@ namespace clad {
                            deriveDiv(Ldiff, Rdiff));
       }
     } else if (opCode == BO_Comma) {
-      if (!isUnusedResult(Ldiff.getExpr_dx()))
-        opDiff = BuildOp(BO_Comma,
-                         BuildParens(Ldiff.getExpr_dx()),
+      // if expression is (E1, E2) then derivative is (E1', E1, E2')
+      // because E1 may change some variables that E2 depends on.
+      if (!isUnusedResult(Ldiff.getExpr_dx())) {
+        opDiff = BuildOp(BO_Comma, BuildParens(Ldiff.getExpr_dx()),
+                         BuildParens(Ldiff.getExpr()));
+        opDiff = BuildOp(BO_Comma, BuildParens(opDiff),
                          BuildParens(Rdiff.getExpr_dx()));
-      else
-        opDiff = Rdiff.getExpr_dx();
+      } else
+        opDiff = BuildOp(BO_Comma, BuildParens(Ldiff.getExpr()),
+                         BuildParens(Rdiff.getExpr_dx()));
     }
     if (!opDiff) {
       // FIXME: add support for other binary operators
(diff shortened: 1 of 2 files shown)


vaithak:
closes #574

parth-07:
@vaithak This is a great find!

Your solution makes the derivative function structure more complicated. Can we make
the derivative function structure simpler? What do you think about:

```cpp
double temp = (++x, x*x);
```

Transforming this to:

```cpp
(++x, _d_x); // We can remove this altogether as the derivative expression can have a side-effect too.
// If there are more arguments then we can have something like this: (arg1, darg1, arg2, darg2, arg3, darg3, ...)
double _d_temp = _d_x * x + x * _d_x; // It will only have the final derivative
double temp  = x * x; // It will only have the final value
```

vaithak:
Hi @parth-07, I understand that my solution makes the derivative a bit long and complicated, but adding an extra expression for `(dE1, E1)` can cause problems; for example, consider this function:
```cpp
double f(double x, bool flag) {
  // if flag is true, return 4x^2, else return (x+1)^2
  return flag ? (x*=2, x * x) : (x+=1, x * x);
}
```

Having an extra expression on top will result in side effects of both conditions taking place, for example, I tried implementing it, and this is the code produced:
```cpp
double f_darg0(double x, bool flag) {
    double _d_x = 1;
    bool _d_flag = 0;
    (_d_x = _d_x * 2 + x * 0) , (x *= 2);
    (_d_x += 0) , (x += 1);
    return flag ? (_d_x * x + x * _d_x) : (_d_x * x + x * _d_x);
}
```

vgvassilev:
> So, I think it is essential that we keep one single expression for the derivative and one single expression for the original cloned statement.

I think we should probably transform expressions into expressions when we can to avoid problems like the one you described. Otherwise we will have to put a lot more infrastructure to make sure the semantics is right. @parth-07 do you think we can prioritize making this work and then possibly figuring out ways to make the code nicer as a separate effort?

parth-07:
@vaithak Thank you for providing the ternary operator case. 

> So, I think it is essential that we keep one single expression for the derivative and one single expression for the original cloned statement.

Yes, you are right.

>  @parth-07 do you think we can prioritize making this work and then possibly figuring out ways to make the code nicer as a separate effort?

Yes, that sounds good to me.

parth-07 (approved):
Looks good.

vaithak:
@vgvassilev, do you think this is good to go?

vgvassilev (approved):
LGTM!

PR #577, opened 2023-07-03
Enhance vector fwd mode to differentiate w.r.t selected params

### LINKED ISSUE #566: Supporting differentiated w.r.t selected parameters in forward vector mode differentiation
Initial PR for vectorized forward mode (https://github.com/vgvassilev/clad/pull/565) assumes that the requirement is to differentiate w.r.t all function parameters. This issue asks to make it flexible so that the user can specify a partial list of parameters for calculating the derivative.

vaithak:
After this commit, the user can specify which params it wants to differentiate against.  

For ex: if the function is of the form: 
```cpp
double f(double x, double y, double z) {
  ...
}
```
Then `clad::vector_forward_differentiate(f, "x,z")` will produce:  
```cpp
void f_dvec_0_2(double x, double y, double z, double *_d_x, double *_d_z) {
  ...
  ...
  {
     *_d_x = ...;
     *_d_z = ...;
     return;
  }
} 
```

closes #566

### DIFF
--- a/include/clad/Differentiator/VectorForwardModeVisitor.h
+++ b/include/clad/Differentiator/VectorForwardModeVisitor.h
@@ -10,6 +10,7 @@ namespace clad {
 /// Used to compute derivatives by clad::vector_forward_differentiate.
 class VectorForwardModeVisitor : public BaseForwardModeVisitor {
 private:
+  llvm::SmallVector<const clang::ValueDecl*, 16> m_IndependentVars;
   /// Map used to keep track of parameter variables w.r.t which the
   /// the derivative is being computed. This is separate from the
   /// m_Variables map because all other intermediate variables will have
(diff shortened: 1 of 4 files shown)


parth-07 (approved):
This looks great!

vgvassilev:
This looks good to me. However I’d like add a demo and a set of benchmarks. In addition perhaps we should discuss the interface. Maybe we could add an option to differentiate such as differentiate<opts::vector,…>?

vaithak:
For the interface, we had a discussion sometime back regarding whether to have a `clad::opts` option or a specialised function, where it was finalized to have a special function, `clad::vector_forward_differentiatie`. 
Are you suggesting that we go away with having a specific function and have an option rather? If so, any specific reasons for this?

Will add a demo along with documentation after this. For benchmarking, we need to have support for arrays inv vector mode because of the most current benchmark functions work on arrays, or I can try to add a benchmark for a non array-input function if that is ok?

vgvassilev:
> For the interface, we had a discussion sometime back regarding whether to have a `clad::opts` option or a specialised function, where it was finalized to have a special function, `clad::vector_forward_differentiate`. Are you suggesting we go away with having a specific function and have an option? If so, any specific reasons for this?

The `vector_forward_differentiate` hints at a particular implementation strategy. In addition it is inconsistent with `clad::differentiate` which should be `forward_differentiate`?

> 
> I will add a demo along with documentation after this. For benchmarking, we need to have support for arrays in vector mode because most of the current benchmark functions work on arrays, or I can try to add a benchmark for a non array-input function if that is ok?

I am fine delaying that until we get the array support.

vaithak:
I agree with you that this interface seems slightly inconsistent with Clad's other methods. For the interface, we can discuss it on discord to finalize it; the changes in this PR are relevant irrespective of whichever interface we choose. Can we merge this if the changes seem okay to you?
Will create a separate PR for whichever interface is finalized along with the demo example and documentation.
WDYT?

vgvassilev (approved):
LGTM!

PR #583, opened 2023-07-12
Improve AD function interfaces with bitmasked options

vaithak:
This will extend existing functions to have support for options like vector mode.

A current working example is:
- `clad::differentiate <clad::opts::vector_mode> (fn)` - first-order differentiation of the function `fn` using vector forward mode

Another example which is not yet supported but may be helpful for the future:
- `clad::gradient <clad::opts::use_enzyme, clad::opts::vector_mode> (fn)` - first-order gradient computation (reverse mode AD) using enzyme's vector mode.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -42,3 +42,5 @@ CheckOptions:
     value:           aNy_CasE
   - key:             readability-identifier-naming.IgnoreMainLikeFunctions
     value:           1
+  - key:             readability-implicit-bool-conversion.AllowPointerConditions
+    value:           1
--- a/docs/userDocs/source/user/UsingEnzymeWithinClad.rst
+++ b/docs/userDocs/source/user/UsingEnzymeWithinClad.rst
@@ -1,52 +1,75 @@
 Using Enzyme within Clad
 *************************
 
-Like Clad, `Enzyme <https://enzyme.mit.edu/>`_ is also a library for Automatic Differentiation(AD). A major difference is that, Enzyme works at the LLVM Intermediate Representation(IR) level, while Clad works at the Clang Abstract Syntax Tree(AST) level.
-Clad analyses the Clang AST to generate the derivative of a function, while Enzyme analyses the LLVM IR of a function to generate the derivative of the same.
+Like Clad, `Enzyme <https://enzyme.mit.edu/>`_ is also a library for Automatic
+Differentiation(AD). A major difference is that, Enzyme works at the LLVM
+Intermediate Representation(IR) level, while Clad works at the Clang Abstract
+Syntax Tree(AST) level. Clad analyses the Clang AST to generate the derivative
+of a function, while Enzyme analyses the LLVM IR of a function to generate the
+derivative of the same.
 
-Languages such as Rust, C++, Julia and Swift output LLVM IR which makes Enzyme more interoperable as it can differentiate functions originally written in multiple languages.
-However, in Enzyme, interoperability comes at a cost. For example, while basic functions are supported across various languages; containers, classes and other language specific constructs require extra scaffolding to be supported by Enzyme.
-Clad is more tightly coupled with the Clang frontend and the C++ language. It has access to the high-level program structure and compile-only constructs such as `consteval`. That allows more coherency when differentiating C++.
-This includes first class support of Object Oriented Programs written in C++.
+Languages such as Rust, C++, Julia and Swift output LLVM IR which makes Enzyme
+more interoperable as it can differentiate functions originally written in
+multiple languages. However, in Enzyme, interoperability comes at a cost. For
+example, while basic functions are supported across various languages;
+containers, classes and other language specific constructs require extra
+scaffolding to be supported by Enzyme. Clad is more tightly coupled with the
+Clang frontend and the C++ language. It has access to the high-level program
+structure and compile-only constructs such as `consteval`. That allows more
+coherency when differentiating C++. This includes first class support of Object
+Oriented Programs written in C++.
 
-The initial integration of Enzyme in Clad is to enable cross validation of the produced code.
+The initial integration of Enzyme in Clad is to enable cross validation of the
+produced code.
 
-Currently, use of Enzyme for Reverse Mode AD is supported within Clad. This section describes how Enzyme can be used within Clad. 
+Currently, use of Enzyme for Reverse Mode AD is supported within Clad. This
+section describes how Enzyme can be used within Clad.
 
 
 Configuring Clad to use Enzyme
 =================================
-To enable the use of enzyme within Clad, one needs to configure Clad to use Enzyme. This can be done by adding the flag ``-DENABLE_ENZYME_BACKEND=On`` to cmake while configuring Clad build.
-Thus the overall cmake command should look something like this
+To enable the use of enzyme within Clad, one needs to configure Clad to use
+Enzyme. This can be done by adding the flag ``-DENABLE_ENZYME_BACKEND=On`` to
+cmake while configuring Clad build. Thus the overall cmake command should look
+something like this
 
 .. code-block:: bash
 
-    cmake ../clad -DClang_DIR=/usr/lib/llvm-11 -DLLVM_DIR=/usr/lib/llvm-11 -DCMAKE_INSTALL_PREFIX=../inst -DLLVM_EXTERNAL_LIT="``which lit``" -DENABLE_ENZYME_BACKEND=On
+    cmake ../clad -DClang_DIR=/usr/lib/llvm-11 -DLLVM_DIR=/usr/lib/llvm-11
+    -DCMAKE_INSTALL_PREFIX=../inst -DLLVM_EXTERNAL_LIT="``which lit``"
+    -DENABLE_ENZYME_BACKEND=On
 
-This flag instructs the build system to download and build Enzyme. Then it is linked as a static library to Clad.
+This flag instructs the build system to download and build Enzyme. Then it is
+linked as a static library to Clad.
 
 Asking Clad to generate gradients with Enzyme
 ================================================
 
-The following code snippet shows how one can request Clad to generate gradients with Enzyme::
+The following code snippet shows how one can request Clad to generate gradients
+with Enzyme::
 
     #include "clad/Differentiator/Differentiator.h"
 
-    double foo(double* arr) { return arr[0] * arr[1]; }
+    double array_product(double* arr) { return arr[0] * arr[1]; }
 
     int main(){
-        auto grad = clad::gradient<clad::opts::use_enzyme>(f1);
+        auto grad = clad::gradient<clad::opts::use_enzyme>(array_product);
         double v[2] = {3, 4};
         double g[2] = {0};
         grad.execute(v, g);
-        printf("d_x = %.2f, d_y = %.2f\n", g[0], g[1]); 
+        printf("d_x = %.2f, d_y = %.2f\n", g[0], g[1]);
     }
 
-Thus, the calling convention is to use ``clad::gradient<clad::opts::use_enzyme>(...)`` instead of the usual calling convention, ``clad::gradient(...)``.
-Calling ``execute``, ``dump`` and other functionalities remain same as that of Clad.
+Thus, the calling convention is to use
+``clad::gradient<clad::opts::use_enzyme>(...)`` instead of the usual calling
+convention, ``clad::gradient(...)``. Calling ``execute``, ``dump`` and other
+functionalities remain same as that of Clad.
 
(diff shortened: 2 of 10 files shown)


include/clad/Differentiator/Differentiator.h
```diff
@@ -291,9 +292,11 @@ namespace clad {
   /// \param[in] fn function to differentiate 
   /// \param[in] args independent parameter information 
   /// \returns `CladFunction` object to access the corresponding derived function.
-  template <signed N = 1, typename ArgSpec = const char*, typename F,
+  template <unsigned N = 1, unsigned BitMaskedOpts = 0,
+            typename ArgSpec = const char*, typename F,
             typename DerivedFnType = ExtractDerivedFnTraitsForwMode_t<F>,
             typename = typename std::enable_if<
+                !(BitMaskedOpts & clad::opts::vector_mode) &&
```
vgvassilev, line 299: Can we hide the bit operation behind an interface. Eg something like HasOption or IsEnabled…

vaithak: added the option and corresponding helper in `CladConfig.h` so that it can be accessed from `Differentiator.h` as well as other files.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -537,51 +538,68 @@ namespace clad {
...
         request.RequestedDerivativeOrder = derivativeOrder;
+
+        unsigned const bitmasked_opts_value =
+            FD->getTemplateSpecializationArgs()
+                ->get(1)
+                .getAsIntegral()
+                .getZExtValue();
+        if (clad::HasOption(bitmasked_opts_value, clad::opts::use_enzyme)) {
+          request.use_enzyme = true;
+        }
```
vgvassilev, line 564:
```suggestion
        if (clad::HasOption(bitmasked_opts_value, clad::opts::use_enzyme))
          request.use_enzyme = true;
```

I believe we could configure clang-format or clang-tidy to find these.

test/Enzyme/ForwardMode.C
```diff
@@ -13,5 +13,5 @@ double f(double x, double y) { return x * y; }
 // CHECK-NEXT:}
 
 int main(){
-     auto f_dx = clad::differentiate<clad::opts::use_enzyme>(f, "x");
+     auto f_dx = clad::differentiate<1, clad::opts::use_enzyme>(f, "x");
```
vgvassilev, line 16: That's a bit annoying. I am not sure if we could elide this if we provide something like: `clad::opts::default` as a default argument to `differentiate`...

parth-07 (approved):
Looks good.

.clang-tidy
```diff
@@ -0,0 +1,32 @@
+Checks: >
```
parth-07, line 1: clang tidy changes should not be part of this pull request.

vaithak: yeah, running those to get an idea of what constraints to add in clang tidy, will remove once this is finalised: https://github.com/vgvassilev/clad/pull/584/files

include/clad/Differentiator/CladConfig.h
```diff
@@ -6,6 +6,17 @@
 #include <cstdlib>
 #include <memory>
 
+namespace clad {
+enum opts {
+  use_enzyme = 1 << 0,
+  vector_mode = 1 << 1,
+}; // enum opts
+
+constexpr bool HasOption(const unsigned bitmasked_opts, const unsigned opt) {
```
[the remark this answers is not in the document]
vgvassilev: Let's disable this one... we are not that modern :)

include/clad/Differentiator/Differentiator.h
```diff
@@ -327,19 +328,20 @@ namespace clad {
...
-  vector_forward_differentiate(
-      F fn, ArgSpec args = "",
-      DerivedFnType derivedFn = static_cast<DerivedFnType>(nullptr),
-      const char* code = "") {
-    assert(fn && "Must pass in a non-0 argument");
-    return CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>, true>(
-        derivedFn, code);
+      annotate("D")))
+  differentiate(F fn, ArgSpec args = "",
+                DerivedFnType derivedFn = static_cast<DerivedFnType>(nullptr),
```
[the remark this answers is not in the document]
vgvassilev: Nope, I'd be in favor of CamelCase...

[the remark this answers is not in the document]
vaithak: kept this as `aNy_Case` because there can be cases like taking function decl as parameter, which is many times `FD` or `F`

.clang-tidy
parth-07: Oh, okay. Sounds good.

test/Enzyme/ForwardMode.C
vaithak: Differentiate has template `<unsigned N=1, unsigned BItMaskedOpts = 0 ....>`, if we want to give non-default value to the second template parameter. I don't see any other way without providing value to the first template parameter.
I think C++-20 allows providing named values to template params, for ex.: `BitMaskedOpts = clad::opts::use_enzyme` but that can be added in the long future.

parth-07: Yes it is slightly annoying, but I don't think we can get away with this if we want to use the same interface, `clad::differentiate`, for both vector and non-vector forward mode.

vgvassilev: Can we somehow mangle the value in the options? Like using the first 3 bits to store the value derivative order.. I don't expect a value greater than 3 to be honest..

vaithak: That is a good point. Will try to make this work and simplify the interface.

vaithak: added the derivative order in the bitmask options 👍🏼

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -537,51 +538,72 @@ namespace clad {
...
       DiffRequest request{};
 
-      // Check if enzyme is requested in case of Forward Mode or reverse Mode
-      if (A->getAnnotation().equals("D") || A->getAnnotation().equals("G")) {
-        signed opts_value = FD->getTemplateSpecializationArgs()
-                                ->get(0)
-                                .getAsIntegral()
-                                .getZExtValue();
+      if (A->getAnnotation().equals("D")) {
+        request.Mode = DiffMode::forward;
```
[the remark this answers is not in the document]
vaithak: Should we suppress this type of warning in clang-tidy?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -537,51 +538,72 @@
...
 
-        if (opts_value == -1) {
+        // bitmask_opts is a template pack of unsigned integers, so we need to
+        // do bitwise or of all the values to get the final value.
+        unsigned bitmasked_opts_value = 0;
+        for (auto const& arg :
+             FD->getTemplateSpecializationArgs()->get(0).pack_elements()) {
```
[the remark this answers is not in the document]
vaithak: same here?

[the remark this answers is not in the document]
vgvassilev: Yes, there seems to be a way to ignore pointer conversions only?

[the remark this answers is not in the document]
vaithak: I don't think so, couldn't find anything here: https://clang.llvm.org/extra/clang-tidy/checks/list.html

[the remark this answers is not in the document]
vaithak: yup, it worked 👍🏼

include/clad/Differentiator/Differentiator.h
```diff
@@ -483,3 +488,14 @@ namespace clad {
 // Enable clad after the header was included.
 // FIXME: The header inclusion should be made automatic if the pragma is seen.
 #pragma clad ON
+
+// // create a function fn which always return single integer,
```
vgvassilev, line 492: Why the double comments here?

vaithak: my bad, I was testing something earlier

include/clad/Differentiator/CladConfig.h
```diff
@@ -6,6 +6,37 @@
 #include <cstdlib>
 #include <memory>
 
+namespace clad {
+// The aim is to have a single unsigned integer for storing both, the
+// differentiation order and the options. The differentiation order is
+// stored in the lower 8 bits, and the options are stored in the upper
+// 24 bits. The differentiation order is stored in the lower 8 bits
+// thus allowing for a maximum differentiation order of 2^8 - 1 = 255.
+const unsigned ORDER_BITS = 8;
```
parth-07, line 15: Minor: this can be constexpr too.

vgvassilev:
Can you squash the second and the last commit into the first and capitalize the clang-tidy commit?

include/clad/Differentiator/CladConfig.h
```diff
@@ -6,6 +6,37 @@
...
+constexpr unsigned GetDerivativeOrder(unsigned const bitmasked_opts) {
+  return bitmasked_opts & ORDER_MASK;
+}
+
+constexpr bool HasOption(unsigned const bitmasked_opts, unsigned const option) {
+  return (bitmasked_opts & option) == option;
+}
+
+constexpr unsigned GetBitmaskedOpts() { return 0; }
+constexpr unsigned GetBitmaskedOpts(unsigned const first) { return first; }
```
parth-07, line 32: Why do we need this overload?

vaithak: This is needed because we unfold this recursively at compile time, as defined in the function below it:
```cpp
template <typename... Opts>
constexpr unsigned GetBitmaskedOpts(unsigned const first, Opts... opts) {
  return first | GetBitmaskedOpts(opts...);
}
```

vaithak:
> Can you squash the second and the last commit into the first and capitalize the clang-tidy commit?

done 👍🏼

include/clad/Differentiator/CladConfig.h
vgvassilev: I am wondering if we can provide some enum for these magic numbers like `1, 2 ...` to make the code more readable... Maybe up to 3rd order should be reasonable..

vaithak: I think it won't be a good experience while implementing because it introduces too many characters into the interface, like `clad::differentiate<clad::opts::order_2>`.  

Whereas, `clad::differentiate<2>` can simply be read as differentiate 2 times.

vgvassilev: People would be still able to write it that assuming we get the enum values right.

vaithak: Sounds good then, I will add the enums, any preferences for the naming? Or, does `clad::opts::order_1/2/3` work?

vgvassilev: Maybe: `clad::opts::order::{first, second, third}`?

I realize as part of this PR we should update the documentation to show the new functionality...

vaithak: Sure, will add it 👍🏼

docs/userDocs/source/user/UsingEnzymeWithinClad.rst
```diff
@@ -35,7 +35,7 @@ The following code snippet shows how one can request Clad to generate gradients
     double foo(double* arr) { return arr[0] * arr[1]; }
 
     int main(){
-        auto grad = clad::gradient<clad::opts::use_enzyme>(f1);
+        auto grad = clad::gradient<clad::opts::use_enzyme>(foo);
```
vgvassilev, line 38: Let's change the identifier here instead of `foo` to something else. Eg. `sq_array`?

docs/userDocs/source/user/UsingVectorMode.rst
```diff
@@ -0,0 +1,39 @@
+Using Vector Mode for Differentiation
+**************************************
+
+For forward mode AD, the restriction is that the function can be only be differentiated with respect to a single input variable.
```
vgvassilev, line 4: Can you configure your editor to break at 80 columns and trim trailing whitespace?

vgvassilev (commented):
Apologies to take so long to merge this PR. I believe that's the last round.

vaithak:
> Apologies to take so long to merge this PR. I believe that's the last round.

No problem, these are user-facing changes so the details matter.

vgvassilev (approved):
Excellent job!

PR #584, opened 2023-07-12
Add .clang-tidy file to limit checks run on static analysis

vaithak:
Checks are similar to the ones defined in LLVM repo: https://github.com/llvm/llvm-project/blob/main/.clang-tidy

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,44 @@
+Checks: >
+  -*,
+  bugprone-*,
+  clang-diagnostic-*,
+  clang-analyzer-*,
+  cppcoreguidelines-*,
+  llvm-*,
+  misc-*,
+  modernize-*,
+  performance-*,
+  portability-*,
+  readability-*,
+  -bugprone-narrowing-conversions,
+  -bugprone-easily-swappable-parameters,
+  -bugprone-implicit-widening-of-multiplication-result,
+  -bugprone-unchecked-optional-access,
+  -misc-const-correctness,
+  -misc-unused-parameters,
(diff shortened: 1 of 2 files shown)


.clang-tidy
```diff
@@ -0,0 +1,27 @@
+Checks: >
+  -*,
+  clang-diagnostic-*,
+  llvm-*,
+  misc-*,
```
vgvassilev, line 5: Don't we want to add `clang-analyzer-`?

vaithak: added 👍🏼

vgvassilev: I am going through the checks and maybe we could benefit from `modernize-` and `performance-`. Maybe `cppcoreguidelines-` but I am not so sure?

vaithak: modernize and performance are used by google cloud's repo also: https://github.com/googleapis/google-cloud-cpp/blob/main/.clang-tidy

vgvassilev: Let's add the three of these and we can further tweak..

vaithak: added `portability` and `readability` too, for readability will try to add some constraints given so many comments on https://github.com/vgvassilev/clad/pull/583

vgvassilev: Makes sense.

vaithak: I think we have a good baseline of checks and constraints, wdyt? It is not producing any false positives on #583; we can add more constraints in case if produces false positives in any future PRs.

vgvassilev: Yep, looks good!

vgvassilev (approved):

Issue #582, opened 2023-07-11
No need to store return values


PetroZarytskyi:
At the end of the forward pass for every non-void function, we store the return value in <function_name>_return. However, we only need to store those variables that could be used in control flow or be present as arguments in functions or operators (which the return value never could be). Therefore, storing it is unnecessary.

PR #585, opened 2023-07-13
Remove the redundant store of primal return result into a variable

PetroZarytskyi:
At the end of the forward pass for every non-void function, we store the return value in <function_name>_return. However, we only need to store those variables that could be used in control flow or be present as arguments in functions or operators (which the return value never could be). Therefore, storing it is unnecessary. In this pull request, all the <function_name>_return variables are removed.

Note: when we need _ret_value in error estimation mode we now directly assign `_ret_value = <return_expression>` instead of `_ret_value = <function_name>_return`. `MultiplexExternalRMVSource::ActBeforeFinalisingVisitReturnStmt()`, `ErrorEstimationHandler::ActBeforeFinalisingVisitReturnStmt()` and `ErrorEstimationHandler::SaveReturnExpr()` now take the return expression as their only argument.

Fixes vgvassilev/clad#582.


vgvassilev:
Could you rebase this PR, remove all commits but the last one?

vgvassilev:
@PetroZarytskyi, we should probably do `git grep _return` and fix all relevant occurrences such as the ones in `test/CUDA/GradientCuda.cu` and `test/Enzyme/DifferentCladEnzymeDerivatives.C` but also in documentation. Could you do that?

PetroZarytskyi:
> @PetroZarytskyi, we should probably do `git grep _return` and fix all relevant occurrences such as the ones in `test/CUDA/GradientCuda.cu` and `test/Enzyme/DifferentCladEnzymeDerivatives.C` but also in documentation. Could you do that?

Sure.

vgvassilev:
I believe codecov is confusing itself about the diff test coverage.

PR #587, opened 2023-07-14
test: Use LLVM_LIBRARY_OUTPUT_INTDIR

hahnjo:
This automatically finds `clad.so` inside `lib64/`, for example on RHEL based distributions.


vgvassilev (approved):
Thats great! Thanks!

PR #589, opened 2023-07-14
Adapt clad for LLVM16

hahnjo:


vgvassilev (approved):

vgvassilev:
The patch coverage and clang-tidy suggestions should be ignored since they are irrelevant for the compatibility layer. 

@alexander-penev, it seems we got the llvm16 support.  Could you check if that’s enough?

PR #590, opened 2023-07-16
add pushforwards for floor() and ceil()

PetroZarytskyi:


PetroZarytskyi:
This pull request still ignores the fact that ceil and floor are not differentiable at integer values. I'm not quite sure how to signal the user about the possible discontinuity (throwing warnings every time such functions are being differentiated or using assert inside the pushforwards themselves doesn't seem right). The same applies to differentiating abs at 0 (right now, the derivative at 0 is 1).

parth-07:
> This pull request still ignores the fact that ceil and floor are not differentiable at integer values. I'm not quite sure how to signal the user about the possible discontinuity (throwing warnings every time such functions are being differentiated or using assert inside the pushforwards themselves doesn't seem right).

Yes, I don't think we should report warnings/errors for such cases.

> The same applies to differentiating abs at 0 (right now, the derivative at 0 is 1).

Yes :').

@vgvassilev What should we do here? Should we not add custom derivatives for such cases, or should we let the
derivatives be 0 for invalid cases, or perhaps error out if the derivative is being computed when there's a discontinuity (and thus, the derivative can not be computed).

vgvassilev:
> > This pull request still ignores the fact that ceil and floor are not differentiable at integer values. I'm not quite sure how to signal the user about the possible discontinuity (throwing warnings every time such functions are being differentiated or using assert inside the pushforwards themselves doesn't seem right).
> 
> Yes, I don't think we should report warnings/errors for such cases.
> 
> > The same applies to differentiating abs at 0 (right now, the derivative at 0 is 1).
> 
> Yes :').
> 
> @vgvassilev What should we do here? Should we not add custom derivatives for such cases, or should we let the derivatives be 0 for invalid cases, or perhaps error out if the derivative is being computed when there's a discontinuity (and thus, the derivative can not be computed).

Isn't that a common problem for other pullbacks, too?

vgvassilev:
> > This pull request still ignores the fact that ceil and floor are not differentiable at integer values. I'm not quite sure how to signal the user about the possible discontinuity (throwing warnings every time such functions are being differentiated or using assert inside the pushforwards themselves doesn't seem right).
> 
> Yes, I don't think we should report warnings/errors for such cases.
> 
> > The same applies to differentiating abs at 0 (right now, the derivative at 0 is 1).
> 
> Yes :').
> 
> @vgvassilev What should we do here? Should we not add custom derivatives for such cases, or should we let the derivatives be 0 for invalid cases, or perhaps error out if the derivative is being computed when there's a discontinuity (and thus, the derivative can not be computed).

Isn't that a common problem for other pullbacks, too?

parth-07:
> Isn't that a common problem for other pullbacks, too?

Control flow statements often lead to discontinuity as well. Therefore it is a common problem in AD in general. But that being said, in those cases, the user explicitly requests derivatives of such functions. Thus, we can say that it is user's job to use the derivatives correctly. But here we are providing the derivatives without the user explicitly requesting it.

That being said, since floating point numbers can not be represented precisely anyways, it should be fine if we compute and return the derivative of float(x = 3.000000000001) when the derivative of float(x=3.0) was requested.

PR #591, opened 2023-07-17
Rename executable in TestTypeConversion

hahnjo:
It clashed with `test/Gradient/Gradients.C` which resulted in frequent failures during parallel testing.

*Note:* Ideally the executable would be placed in a temporary directory, maybe based on `%t`?


vgvassilev:
> It clashed with `test/Gradient/Gradients.C` which resulted in frequent failures during parallel testing.
> 
> _Note:_ Ideally the executable would be placed in a temporary directory, maybe based on `%t`?

That would make sense to me!

vgvassilev (approved):
LGTM!

PR #592, opened 2023-07-17
Use proper clang++ for testing

hahnjo:
This solves a problem on EL8 where Clang uses `libstdc++` from the Developer Toolset and linking requires some extra care, which only `clang++` does but not `clang -x c++`.

@vgvassilev was there an explicit reason to use `clang -x c++`?


vgvassilev:
> This solves a problem on EL8 where Clang uses `libstdc++` from the Developer Toolset and linking requires some extra care, which only `clang++` does but not `clang -x c++`.
> 
> @vgvassilev was there an explicit reason to use `clang -x c++`?

Not that I am aware of. We should probably update all of the readme/docs that suggest using clang that way.

PR #593, opened 2023-07-17
demos: Use C++ std::abs for floating point values

hahnjo:
Clang warns that `abs` from C takes an integer argument. `std::abs` has overloads for floating point arguments that work as expected.


vgvassilev (approved):
LGTM!

PR #594, opened 2023-07-17
Fix two asserts in Clang

hahnjo:
This reduces the number of tests failures when using a LLVM build with `assert`s enabled from 21 to 8:
```
Failed Tests (8):
  clad :: Arrays/ArrayInputsReverseMode.C
  clad :: FirstDerivative/Overloads.C
  clad :: FirstDerivative/VirtualMethodsCall.C
  clad :: ForwardMode/UserDefinedTypes.C
  clad :: ForwardMode/VectorMode.C
  clad :: ForwardMode/VectorModeInterface.C
  clad :: Gradient/UserDefinedTypes.C
  clad :: Misc/RunDemos.C
```

Of these:
 * 4x "point of instantiation must be valid"
 * 2x `VTableBuilder.cpp`, "Should not have method info for this method yet!"
 * 1x `CGCall.cpp`, "type mismatch in call argument!"
 * 1x `SourceManager.cpp`, "Can't get file characteristic of invalid loc!"

### DIFF
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -395,7 +395,8 @@ namespace clad {
                   SourceRange(), false, noLoc, MultiExprArg(), noLoc,
                   SourceRange(), qType, TSI,
                   (arraySize ? arraySize : clad_compat::ArraySize_None()),
-                  GetValidSRange(semaRef), initializer)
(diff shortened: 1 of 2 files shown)


vgvassilev:
@vaithak perhaps you could take a look at the asserts coming from vector mode. They should be new.

vgvassilev (approved):
LGTM!

PR #596, opened 2023-07-17
Restore args passed by reference and arg of increment/decrement

PetroZarytskyi:
If an argument is passed to a function by reference it might be changed. Therefore, we have to restore it on the backward pass. The same applies to increments/decrements.
Note: 
1) Restoring reference type parameters still doesn't work with pointer-array type variables with indices that have to be stored (or without indices).
2) Storing variables before they are incremented/decremented may seem like an overkill since we can insert decrement/increment respectively in the backward pass. However, this approach can easily cause bugs: For `x = x++', we would restore x first and then decrement it by one (due to the order of assignment and increment), which is not correct. Moreover, it would be preferable to have a central point for storing variables (which GlobalStoreAndRef() already is).

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1884,6 +1884,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
   StmtDiff ReverseModeVisitor::VisitUnaryOperator(const UnaryOperator* UnOp) {
     auto opCode = UnOp->getOpcode();
     StmtDiff diff{};
+    Expr* E = UnOp->getSubExpr();
     // If it is a post-increment/decrement operator, its result is a reference
     // and we should return it.
     Expr* ResultRef = nullptr;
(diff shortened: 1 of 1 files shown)


parth-07 (commented):
This looks great. I will need some more time to review this change.

Can you please add a test for this as well?

PetroZarytskyi:
@parth-07 Surely I could do that but what test do we need to add? We seem to have tests both for passing arguments by reference and increments/decrements. The only problem is those tests are quite outdated.

vgvassilev:
We can resolve the bot issues by moving `E->EvaluateAsConstantExpr` into a separate routine similar to: https://github.com/vgvassilev/clad/blob/3230086ff9bb8c9a471ba4ea5b773da4292fd656/include/clad/Differentiator/Compatibility.h#L120-L134

We probably need to pass one less argument depending on the version of llvm.

parth-07:
> Surely I could do that but what test do we need to add? We seem to have tests both for passing arguments by reference and increments/decrements. 

Oh, in that case, we do not need to add any more tests.

parth-07 (approved):
Looks good.

vgvassilev:
This PR does not decrease the test failures but make the tests fail further down.

PR #599, opened 2023-07-19
Fix git-clang-format run when PR not on top of master

vaithak:


vaithak:
This should work fine now 👍🏼

vgvassilev (approved):
LGTM! Can you change that in the rest of the repos?

PR #607, opened 2023-07-27
Fix LLVM assertion errors for vector mode

vaithak:
took me a reasonable amount of time to figure out that vector mode also requires creating an overloaded function similar to gradient computation.

### DIFF
--- a/include/clad/Differentiator/Array.h
+++ b/include/clad/Differentiator/Array.h
@@ -246,6 +246,22 @@ public:
   CUDA_HOST_DEVICE operator T*() const { return m_arr; }
 }; // class array
 
+// Function to instantiate a one-hot array of size n with 1 at index i.
+// A one-hot vector is a vector with all elements set to 0 except for one
+// element which is set to 1.
+// For example, if n=4 and i=2, the returned array is: {0, 0, 1, 0}
+template <typename T>
+CUDA_HOST_DEVICE array<T> one_hot_vector(std::size_t n, std::size_t i) {
+  array<T> arr(n);
+  arr[i] = 1;
+  return arr;
+}
+
+// Function to instantiate a zero vector of size n
+template <typename T> CUDA_HOST_DEVICE array<T> zero_vector(std::size_t n) {
+  return array<T>(n);
+}
+
 /// Overloaded operators for clad::array which return a new array.
 
 /// Multiplies the number to every element in the array and returns a new
--- a/include/clad/Differentiator/FunctionTraits.h
+++ b/include/clad/Differentiator/FunctionTraits.h
@@ -735,7 +735,7 @@ namespace clad {
 
   template <class ReturnType, class... Args>
   struct ExtractDerivedFnTraitsVecForwMode<ReturnType (*)(Args...)> {
-    using type = void (*)(Args..., OutputVecParamType_t<Args, ReturnType>...);
+    using type = void (*)(Args..., OutputVecParamType_t<Args, void>...);
   };
 
   /// Specialization for free function pointer type
--- a/include/clad/Differentiator/VectorForwardModeVisitor.h
+++ b/include/clad/Differentiator/VectorForwardModeVisitor.h
@@ -32,6 +32,17 @@ public:
   DerivativeAndOverload DeriveVectorMode(const clang::FunctionDecl* FD,
                                          const DiffRequest& request);
 
+  /// Builds an overload for the vector mode function that has derived params
+  /// for all the arguments of the requested function and it calls the original
+  /// gradient function internally.
+  /// For ex.: if the original function is: double foo(double x, double y)
+  /// , then the generated vector mode overload will be:
+  /// double foo(double x, double y, void*, void*), irrespective of the
+  /// what parameters are requested to be differentiated w.r.t.
+  /// Inside it, we will call the original vector mode function with the
+  /// original parameters and the derived parameters.
+  clang::FunctionDecl* CreateVectorModeOverload();
+
   /// Builds and returns the sequence of derived function parameters for
   //  vectorized forward mode.
   ///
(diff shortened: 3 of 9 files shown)


vaithak:
clang-tidy is unable to post comments right now on any PR because of failing docker build, I have raised a PR to fix this in the original repo for the github action: https://github.com/ZedThree/clang-tidy-review/pull/90

test/ForwardMode/VectorMode.C
```diff
@@ -12,8 +12,8 @@ double f1(double x, double y) {
 void f1_dvec(double x, double y, double *_d_x, double *_d_y);
 
 // CHECK: void f1_dvec(double x, double y, double *_d_x, double *_d_y) {
-// CHECK-NEXT:   clad::array<double> _d_vector_x = {1., 0.};
-// CHECK-NEXT:   clad::array<double> _d_vector_y = {0., 1.};
+// CHECK-NEXT:   clad::array<double> _d_vector_x = clad::one_hot_vector(2UL, 0UL);
```
parth-07, line 15: `clad::one_hot_vector` and `clad::zero_vector` is a great addition. They have made the code
much more readable.

include/clad/Differentiator/Array.h
```diff
@@ -246,6 +246,19 @@ template <typename T> class array {
   CUDA_HOST_DEVICE operator T*() const { return m_arr; }
 }; // class array
 
+// Function to instantiate a one_hot array of size n with 1 at index i
```
vgvassilev, line 249: What does `one_hot` mean?

vaithak: added a comment for this 👍🏼

vgvassilev: Why do we name it as a "hot" array?

include/clad/Differentiator/FunctionTraits.h
```diff
@@ -37,130 +37,128 @@ namespace clad {
...
-  template <class ReturnType, class C, class... Args> 
-  struct return_type<ReturnType (C::*)(Args..., ...)> { 
-    using type = ReturnType; 
+  template <class ReturnType, class C, class... Args>
+  struct return_type<ReturnType (C::*)(Args..., ...)> {
+    using type = ReturnType;
   };
 
   // specializations for member functions pointer type with only cv-qualifiers
-  template <class ReturnType, class C, class... Args>
```
vgvassilev, line 74: Can you configure your editor to not trim trailing whitespace and revert these changes?

include/clad/Differentiator/Array.h
vaithak: This is a standard term: https://en.wikipedia.org/wiki/One-hot

include/clad/Differentiator/FunctionTraits.h
vaithak: reverted space changes from this file, a couple of minor changes remain in other files but I guess that shouldn't be an issue.

include/clad/Differentiator/Array.h
vgvassilev: Ha, ok. There is always something new that I learn every day.

lib/Differentiator/VectorForwardModeVisitor.cpp
```diff
@@ -138,28 +141,166 @@ VectorForwardModeVisitor::DeriveVectorMode(const FunctionDecl* FD,
...
+  DeclWithContext result =
+      m_Builder.cloneFunction(m_Function, *this, DC, m_Sema, m_Context, noLoc,
+                              vectorModeNameInfo, vectorModeFuncOverloadType);
+  FunctionDecl* vectorModeOverloadFD = result.first;
+
+  // Function declaration scope
+  beginScope(Scope::FunctionPrototypeScope | Scope::FunctionDeclarationScope |
+             Scope::DeclScope);
+  m_Sema.PushFunctionScope();
+  m_Sema.PushDeclContext(getCurrentScope(), vectorModeOverloadFD);
```
vgvassilev, line 192: We should probably extract these into some sort of a RAII object as it seems to be a common pattern throughout the code base. Perhaps we could do it in a separate PR.

include/clad/Differentiator/VectorForwardModeVisitor.h
```diff
@@ -32,6 +32,11 @@ class VectorForwardModeVisitor : public BaseForwardModeVisitor {
   DerivativeAndOverload DeriveVectorMode(const clang::FunctionDecl* FD,
                                          const DiffRequest& request);
 
+  /// Builds an overload for the gradient function that has derived params for
+  /// all the arguments of the requested function and it calls the original
+  /// gradient function internally
```
vgvassilev, line 37: We should probably add an example to clarify what we mean here.

lib/Differentiator/VisitorBase.cpp
```diff
@@ -179,11 +179,12 @@ namespace clad {
     return new (m_Context) DeclStmt(DGR, noLoc, noLoc);
   }
 
-  DeclRefExpr* VisitorBase::BuildDeclRef(DeclaratorDecl* D) {
+  DeclRefExpr* VisitorBase::BuildDeclRef(DeclaratorDecl* D,
+                                         const CXXScopeSpec* SS) {
```
vgvassilev, line 183:
```suggestion
                                         const CXXScopeSpec* SS/*=nullptr*/) {
```

That's something useful we have been doing in the codebase. I thought clang-tidy would suggest it..

lib/Differentiator/VisitorBase.cpp
```diff
@@ -564,6 +564,30 @@ namespace clad {
...
+    clang::LookupResult R(m_Sema, declName, noLoc, Sema::LookupOrdinaryName);
+
+    // Find function declaration
+    NamespaceDecl* CladNS = GetCladNamespace();
+    CXXScopeSpec CSS;
+    CSS.Extend(m_Context, CladNS, loc, loc);
+    m_Sema.LookupQualifiedName(R, CladNS, CSS);
+
+    // Build the template specialization expression.
+    // TODO: currently this doesn't print func<templates>(args...) while dumping
```
vgvassilev, line 581:
```suggestion
    // FIXME: currently this doesn't print func<templates>(args...) while dumping
```

include/clad/Differentiator/VectorForwardModeVisitor.h
vaithak: added example 👍🏼

vgvassilev (approved):
LGTM!

PR #608, opened 2023-07-27
Remove Redundant Goto Statements

ShounakDas101:
Solves issue https://github.com/vgvassilev/clad/issues/526. We can only remove the 'goto' label when the parent of the parent of the return statement is null.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -13,7 +13,7 @@
 #include "clad/Differentiator/StmtClone.h"
 #include "clad/Differentiator/ExternalRMVSource.h"
 #include "clad/Differentiator/MultiplexExternalRMVSource.h"
-
+#include "clang/AST/ParentMapContext.h"  
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/Expr.h"
 #include "clang/AST/TemplateBase.h"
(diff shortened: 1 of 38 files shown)


parth-07:
@ShounakDas101 I suggest we wait on this pull request until we all agree that it's useful.

I am not convinced that this pull request is useful. It's solving a very specific case. I believe it will only
be beneficial if we can remove `goto`s in other general cases as well. @ShounakDas101 Please let me know your
thoughts on this.

ShounakDas101:
@parth-07  In this pull request (PR), I have decided to remove 'goto statements' only if the 'return statement' is the direct child node of the 'Compound statement' that contains the entire function body(This is the only case where we should remove 'goto statement'). The other 'return statements' will be within some control flow block. I believe that 'goto statements' for such return statements are essential and should not be removed.

Could you please suggest a test case where this approach fails? This PR has removed all redundant goto statements in the test cases available in CLAD.

parth-07:
Hi @ShounakDas101 

> Could you please suggest a test case where this approach fails? 

I did not say that your approach fails in any case. I just do not see any advantage of adding extra complexity to
remove `goto`'s for a special case of the `return` statement. For this to be beneficial, we also need to remove other
`goto`'s.

ShounakDas101:
Hi @parth-07 I have tried to solve issue #526. I believe the issue specifically only wants to remove redundant goto statements like this
...
goto _label0;
_label0:
...

> For this to be beneficial, we also need to remove other goto's.
By other goto's do you mean all goto's should be removed?

parth-07:
> I believe the issue specifically only wants to remove redundant goto statements like this

Yes, the issue specifically wants to remove `goto`s from the special case only. But I am still not sure if we
benefit from this functionality. The compiler will optimize the redundant `goto` statements anyways. 

The main concern is that `goto` statement prohibits further differentiation. Solving a special case wouldn't lift this restriction. 

@vgvassilev What do you think about removing the redundant `goto` statements?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -744,6 +744,20 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+        if (!parents.empty()){
+          const Stmt* parentStmt =  parents[0].get<Stmt>();
+          if(parentStmt==nullptr)
+            OnlyReturn=true;
+          else
+            OnlyReturn=false;
+        }else{
+          OnlyReturn=true;
+        }
+      }
```
vgvassilev, line 760: We should be able to detect in a smarter way if we need to emit a goto-statement given a forward sweep and then conditionally emit it.

vgvassilev (commented):
I am still concerned by the amount of test failures we have. I think if we manage to solve that case with sound implementation would be acceptable to me.

vgvassilev:
Fixed in #938

PR #609, opened 2023-07-31
Add matrix class in clad

vaithak:


include/clad/Differentiator/Array.h
```diff
@@ -32,13 +34,23 @@ template <typename T> class array {
...
   CUDA_HOST_DEVICE array(U* a, std::size_t size)
       : m_arr(new T[size]{static_cast<T>(T())}), m_size(size) {
     for (std::size_t i = 0; i < size; ++i)
       m_arr[i] = a[i];
   }
 
+  CUDA_HOST_DEVICE array(T* a, std::size_t size, bool copy = true)
+      : m_arr(copy ? new T[size]{static_cast<T>(T())} : a), m_size(size),
+        m_owned(copy) {
+    if (copy)
```
vgvassilev, line 49: I could not find where we use this branch and if we test it.

vaithak: Added more tests in `Misc/CladArray.c` 👍🏼

include/clad/Differentiator/Matrix.h
```diff
@@ -0,0 +1,277 @@
...
+  template <typename U>
+  CUDA_HOST_DEVICE matrix(size_t rows, size_t cols, U val)
+      : m_data(rows * cols, val), m_rows(rows), m_cols(cols) {}
+
+  /// Copy constructor.
+  CUDA_HOST_DEVICE matrix(const matrix& other)
+      : m_data(other.m_data), m_rows(other.m_rows), m_cols(other.m_cols) {}
+
+  /// Move constructor.
+  CUDA_HOST_DEVICE matrix(matrix&& other)
```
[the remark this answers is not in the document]
vaithak: added 👍🏼

benchmark/Matrix.cpp
```diff
@@ -0,0 +1,66 @@
...
+    for (auto _ : state) {
+        benchmark::DoNotOptimize(
+            m[mid] *= 2
+        );
+    }
+}
+
+BENCHMARK(BM_MatrixRowOp)
+    ->RangeMultiplier(2)
+    ->Range(4, 1024);
```
vgvassilev, line 31: We probably want to run to something that is less (eg 256) since we are using this as part of the CI.

include/clad/Differentiator/Array.h
```diff
@@ -185,39 +208,40 @@ template <typename T> class array {
...
+  CUDA_HOST_DEVICE array<T>& operator*=(const array<U>& arr) {
     assert(arr.size() == m_size);
     for (std::size_t i = 0; i < m_size; i++)
       m_arr[i] *= static_cast<T>(arr[i]);
     return *this;
   }
   /// Performs element wise division
   template <typename U>
-  CUDA_HOST_DEVICE array<T>& operator/=(array<U>& arr) {
+  CUDA_HOST_DEVICE array<T>& operator/=(const array<U>& arr) {
```
vgvassilev, line 244: I am wondering why clang-tidy did not detect this const-correctness issue...

include/clad/Differentiator/Matrix.h
```diff
@@ -0,0 +1,277 @@
...
+
+#include <assert.h>
+#include <initializer_list>
+#include <type_traits>
+
+// Implementing a matrix class using clad::array for storing the data.
+namespace clad {
+
+// This class is not meant to be used by the user. It is used by clad internally
+// only.
```
vgvassilev, line 14: Can you use doxygen-style comments (///) to appear in our documentation?

vgvassilev (approved):
Modulo minor comments this looks good to me! Let's address them and merge it.

include/clad/Differentiator/Array.h
vaithak: Marked whole clad array class for `NOLINT` 👍🏼

include/clad/Differentiator/Matrix.h
```diff
@@ -0,0 +1,281 @@
+#ifndef CLAD_MATRIX_H
+#define CLAD_MATRIX_H
```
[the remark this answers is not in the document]
vaithak: should we change this?

include/clad/Differentiator/Matrix.h
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  /// Returns the reference to the row at the given index.
+  CUDA_HOST_DEVICE clad::array<T> operator[](size_t row_idx) {
+    assert(row_idx < m_rows);
+    return m_data.slice(row_idx * m_cols, m_cols);
+  }
+
+  /// Returns the reference to the row at the given index.
+  CUDA_HOST_DEVICE const clad::array<T> operator[](size_t row_idx) const {
```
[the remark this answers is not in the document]
vaithak: should I change the function to return const reference instead?

include/clad/Differentiator/Matrix.h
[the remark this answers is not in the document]
vgvassilev: That would make sense to me.

include/clad/Differentiator/Matrix.h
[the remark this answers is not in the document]
vgvassilev: Probably that’s a good idea…

vgvassilev:
> I have removed the special handling of creating clad::arrays without copying and instead used array_ref in such cases because it was leading to more confusion.

Agreed. I was about to ask that question in the review but was carried away...

Another generic thing to consider is that eventually we will need to add C support.

vaithak:
> Another generic thing to consider is that eventually we will need to add C support.

Can you elaborate on what you mean by adding C support?

PR #610, opened 2023-08-01
Update readthedocs configuration

Krishna-13-cyber:
As we have support up-to Clang-16, I have updated the build setup for readthedocs configuration.
The builder image are not used nowdays, I have updated it with latest build os versions.


vgvassilev (approved):
LGTM!

PR #611, opened 2023-08-01
Fixes in github actions for clang-tidy and clang-format

vaithak:


.clang-tidy
```diff
@@ -26,14 +26,15 @@ Checks: >
   -readability-magic-numbers,
   -readability-named-parameter,
   -readability-function-cognitive-complexity,
+  -cppcoreguidelines-pro-bounds-pointer-arithmetic,
```
vaithak, line 29: Removed this check because clad::array deals with pointer arithmetic

vaithak: I don't think so; there is a similar discussion here: https://github.com/isocpp/CppCoreGuidelines/issues/862 for implementing a vector-like container.

vgvassilev: How about using that approach: https://clang.llvm.org/extra/clang-tidy/#suppressing-undesired-diagnostics

vaithak: Yup, it worked. I had no idea about this 😅

vgvassilev (approved):
There seem to be some unrelated failure with ubuntu22 and llvm 13. Otherwise, lgtm!

PR #613, opened 2023-08-04
Make EvaluateAsConstantExpr work across versions.

vgvassilev:


vgvassilev:
I have rebased the `tape-push` PR. It compiles fine on all supported platforms. The expected failures went up from last time I remember to 24. Let's merge this now as that's an improvement overall and we can try to understanding the reasons why things fail more.

PR #615, opened 2023-08-10
Remove Redundant Goto Statements

ShounakDas101:
Solves issue #526. We can only remove the goto stmt only when return statement is not inside any control flow statement


vgvassilev:
Please stop opening new pull requests. I strongly encourage you to work on https://github.com/vgvassilev/clad/pull/608

in addition as part of our meeting we have proposed an entry point tasks to get to know the code base and git. We strongly encourage you to do that rather than keeping bombarding the repository with such PRs.

PR #619, opened 2023-08-20
Operator overload in reverse mode

PhrygianGates:
Add support for operator overload in reverse mode

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -369,7 +369,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
 
     if (m_ExternalSource)
       m_ExternalSource->ActAfterCreatingDerivedFnScope();
-    
+
     auto params = BuildParams(args);
 
     if (m_ExternalSource)
@@ -411,7 +411,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         m_IndependentVars.push_back(arg);
       }
     }
-    
+
     if (m_ExternalSource)
       m_ExternalSource->ActBeforeCreatingDerivedFnBodyScope();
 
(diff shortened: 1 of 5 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1993,6 +2016,18 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
           }
         }
       }
+
+      if (opCode == UnaryOperatorKind::UO_Deref &&
+          m_Mode == DiffMode::reverse_mode_forward_pass) {
```
parth-07, line 2021: Why do we have reverse mode forward pass specific code in ReverseModeVisitor? This defeats the whole purpose of having a separate class for ReverseModeVisitor.

parth-07 (changes requested):
We need more tests. 
1) Member operator overloads -- operator overloads that are defined inside the class.
2) Operator overload that does not return a reference type.
3) Unary operator overload.

PhrygianGates:
> We need more tests.
> 
> 1. Member operator overloads -- operator overloads that are defined inside the class.
> 2. Operator overload that does not return a reference type.
> 3. Unary operator overload.

For 1, our tests already cover the case.
For 2, because we have this prior restriction in the existing code we can not support the case
```cpp
      if (!FD->getReturnType()->isVoidType()) {
        assert((pullback && !FD->getReturnType()->isVoidType()) &&
               "Call to function returning non-void type with no dfdx() is not "
               "supported!");
      }
```
 For 3, I have added tests for unary op.

parth-07:
> For 2, because we have this prior restriction in the existing code base, we can not support the case yet.

Why does the restriction that you mentioned prevent non-reference return types? Also, why do you think it only affects non-reference return type operator overloads, but not non-reference return type ordinary member functions?

benchmark/VectorModeComparison.cpp
```diff
@@ -0,0 +1,122 @@
+#include "benchmark/benchmark.h"
+
+#include "clad/Differentiator/Differentiator.h"
+
+#include "BenchmarkedFunctions.h"
+
+// Benchmark forward mode for weighted sum.
+static void BM_ForwardModeWeightedSum(benchmark::State& state) {
```
parth-07, line 8: These benchmarks shouldn't be part of this pull request. Can you please rebase the pull request properly?

vgvassilev (approved):
LGTM!

PR #621, opened 2023-08-23
Fix benchmark comparison in ci

vaithak:


vgvassilev (approved):
Thank you! LGTM!

vgvassilev:
@vaithak, looks like when it went in the master it failed with: https://github.com/vgvassilev/clad/actions/runs/5960281839/job/16167325061

vaithak:
yup, I tried to ssh into tmate for that action, but it didn't allow me to, maybe because of access related to master branch. 
Will try to ssh into a PR branch and see what's going wrong. 

I have found the issue, but need to figure out why that happened.
The problem is that `git rev-parse HEAD` is outputting `HEAD` instead of the commit id, this only happens when HEAD is in a detached state.

vgvassilev:
If you tried to do ssh via the web browser it does not work for me either. If you ssh via the terminal you need to type `ctrl+c` to get to a prompt. I can restart the same build if you want to retry.

vaithak:
I tried with terminal only, it gave me `Permission denied (publickey)` error.

vgvassilev:
I logged in, could you retry?

vaithak:
> I logged in, could you retry

trying

vaithak:
still got the same error of `Permission denied (publickey)`.
I am able to access `ssh -T git@github.com` so there should be no error with the keys setup.

PR #623, opened 2023-08-28
Add benchmark for comparing vector mode with existing methods

vaithak:


benchmark/VectorModeComparison.cpp
```diff
@@ -0,0 +1,122 @@
+#include "benchmark/benchmark.h"
```
vgvassilev, line 1: Don't we need to register that file in cmake?

vaithak: updated 👍🏼

vaithak:
> @vaithak can you interpret the benchmark results?

@vgvassilev By "interpret",  do you mean the reason why vector mode is slower compared to other approaches? If so, the main reason is because of many memory allocations; this includes vector creation for all parameters and local variables, but also the temporary vector creation, which can be eliminated if one unrolls the operations manually. 

For example, for this operation -> `res_a += a1*a2 + a3*a4`, where `res_a, a1, a2, a3, a4` are all `clad::array`s, this will result in the creation of multiple temporary array instances, which can be removed if one writes a loop to do these operations by accessing the array elements. The example in this benchmark also contains such an instance.

An ideal solution for this would be to improve this by modifying the array and matrix class but keeping the generated code intact, as it provides good readability. Do you have a suggestion for how this can be fixed?

vgvassilev:
> > @vaithak can you interpret the benchmark results?
> 
> @vgvassilev By "interpret", do you mean the reason why vector mode is slower compared to other approaches? If so, the main reason is because of many memory allocations; this includes vector creation for all parameters and local variables, but also the temporary vector creation, which can be eliminated if one unrolls the operations manually.
> 
> For example, for this operation -> `res_a += a1*a2 + a3*a4`, where `res_a, a1, a2, a3, a4` are all `clad::array`s, this will result in the creation of multiple temporary array instances, which can be removed if one writes a loop to do these operations by accessing the array elements. The example in this benchmark also contains such an instance.
> 
> An ideal solution for this would be to improve this by modifying the array and matrix class but keeping the generated code intact, as it provides good readability. Do you have a suggestion for how this can be improved?

I think improving the `array` and `matrix` classes would be a very good start, indeed.

PR #624, opened 2023-08-30
Add TBR analysis to Clad

PetroZarytskyi:
This pull request optimizes storing and restoring in the reverse mode of Clad and introduces TBR (To-Be-Recorded) analysis to determine when variables should be stored.


include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -41,6 +41,7 @@ namespace clad {
     /// the reverse mode we also accumulate Stmts for the reverse pass which
     /// will be executed on return.
     std::vector<Stmts> m_Reverse;
+    std::vector<Stmts> m_EssentialReverse;
```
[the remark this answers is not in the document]
PetroZarytskyi: @parth-07 This was introduced by you on the tape-push branch. Do we need this?

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -50,14 +51,14 @@ namespace clad {
     Stmts m_Globals;
     //// A reference to the output parameter of the gradient function.
     clang::Expr* m_Result;
+    /// Based on To-Be-Recorded analysis performed before differentiation,
+    /// tells UsefulToStoreGlobal whether a variable with a given
+    /// SourceLocation has to be stored before being changed or not.
+    std::map<clang::SourceLocation, bool> m_ToBeRecorded;
```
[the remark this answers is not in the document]
PetroZarytskyi: @vgvassilev I fixed this.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1522,7 +1549,9 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       // _t1 = a;
       // modify(a);
       // ```
-      if (passByRef) {
+      // FIXME: We cannot use GlobalStoreAndRef to store a whole array so now
+      // arrays are not stored.
+      if (passByRef && !argDiff.getExpr()->getType()->isArrayType()) {
```
vgvassilev, line 1554:
```suggestion
      if (passByRef && !isArrayOrPointerType(argDiff.getExpr()->getType())) {
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2225,12 +2301,29 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+        addToCurrentBlock(S, direction::reverse);
+
+      if (dfdx() && Lblock_begin != Lblock_end) {
         addToCurrentBlock(*Lblock_begin, direction::reverse);
         Lblock_begin = std::next(Lblock_begin);
       }
-      while(!m_PopIdxValues.empty())
-        addToCurrentBlock(m_PopIdxValues.pop_back_val(), direction::reverse);
+
+      for (auto E : return_exprs) {
```
vgvassilev, line 2317: In these cases I'd spell out the actual type.

lib/Differentiator/StmtClone.cpp
```diff
@@ -63,19 +63,19 @@ Stmt* StmtClone::Visit ## CLASS(CLASS *Node)            \
   return result;                                        \
```
vgvassilev, line 63: Please open a new PR with these changes - we can easily write tests for it by adding an array that takes a `ParmVarDecl` as an index and we will hit the type cloning limitation...

vgvassilev (commented):
This looks like it is going in a good direction. We will need a couple of more iterations to converge. @parth-07, can you also take a look.

include/clad/Differentiator/TBRAnalyzer.h
```diff
@@ -0,0 +1,339 @@
...
+      return eqAPInt(x, y);
+    }
+  };
+
+  /// Just a helper struct serving as a wrapper for IdxOrMemberValue union.
+  /// Used to unwrap expressions like a[6].x.t[3]. Only used in
+  /// TBRAnalyzer::overlay().
+  struct IdxOrMember {
+    enum IdxOrMemberType { FIELD, INDEX };
+    union IdxOrMemberValue {
```
parth-07, line 44: Can we use std::variant instead of union?

include/clad/Differentiator/TBRAnalyzer.h
```diff
@@ -0,0 +1,339 @@
...
+    }
+    IdxOrMember(llvm::APInt&& index) : type(IdxOrMemberType::INDEX) {
+      new (&val.index) llvm::APInt(index);
+    }
+    IdxOrMember(const IdxOrMember& other) {
+      new (&val.index) llvm::APInt();
+      *this = other;
+    }
+    IdxOrMember(const IdxOrMember&& other) noexcept {
+      new (&val.index) llvm::APInt();
```
parth-07, line 67: Why do we need to have different copy constructors and move constructors here?
Also, why do we need to explicitly set `val.index` before calling the copy assignment operator?

include/clad/Differentiator/TBRAnalyzer.h
```diff
@@ -0,0 +1,339 @@
...
+      *this = other;
+    }
+    IdxOrMember& operator=(const IdxOrMember& other) {
```
parth-07, line 70: The copy assignment operator may lead to undefined behavior if `type` and `other.type` have different values.

include/clad/Differentiator/TBRAnalyzer.h
```diff
@@ -0,0 +1,339 @@
...
+  /// whole structures instead (just one VarData for the whole array).
+
+  struct VarData;
+  using ObjMap = std::unordered_map<const clang::FieldDecl*, VarData*>;
+  using ArrMap =
+      std::unordered_map<const llvm::APInt, VarData*, APIntHash, APIntComp>;
+
+  struct VarData {
+    enum VarDataType { UNDEFINED, FUND_TYPE, OBJ_TYPE, ARR_TYPE, REF_TYPE };
+    union VarDataValue {
```
parth-07, line 113: Using std::variant would definitely be better if we can use C++17 features.

lib/Differentiator/TBRAnalyzer.cpp
```diff
@@ -0,0 +1,815 @@
...
+}
+
+void TBRAnalyzer::VisitDeclRefExpr(const DeclRefExpr* DRE) {
+  if (const auto* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
+    auto& curBranch = getCurBranch();
+    if (curBranch.find(VD) == curBranch.end())
+      addVar(VD);
+  }
+
+  if (const auto* E = dyn_cast<clang::Expr>(DRE))
```
vgvassilev, line 560: This seems to be always true.

include/clad/Differentiator/TBRAnalyzer.h
```diff
@@ -0,0 +1,318 @@
...
+#include "clad/Differentiator/Compatibility.h"
+
+#include <map>
+#include <unordered_map>
+
+using namespace clang;
+
+namespace clad {
+
+class TBRAnalyzer : public clang::ConstStmtVisitor<TBRAnalyzer> {
```
vgvassilev, line 17: We can probably remove a lot of the boilerplate VisitXYZ routines by deriving from the RecursiveASTVisitor which has the rules of visitation in the base class.

lib/Differentiator/CladUtils.cpp
```diff
@@ -543,5 +543,74 @@ namespace clad {
       // non-differentiable attribute
       return false;
     }
+
+    std::vector<clang::Expr*> GetInnermostReturnExpr(const clang::Expr* E) {
```
vgvassilev, line 547:
```suggestion
    void GetInnermostReturnExpr(const clang::Expr* E, std::vector<clang::Expr*> Exprs) {
```

include/clad/Differentiator/TBRAnalyzer.h
```diff
@@ -0,0 +1,318 @@
...
+
+  /// FIXME: Add support for references to call expression results.
+  /// 'double& x = f(b);' is not supported.
+
+  struct VarData;
+  using ArrMap =
+      std::unordered_map<const ProfileID, VarData, ProfileIDHash>;
+
+  struct VarData {
+    enum VarDataType { UNDEFINED, FUND_TYPE, OBJ_TYPE, ARR_TYPE, REF_TYPE };
```
vgvassilev, line 58: Why don't we store the `QualType` here?

test/Arrays/ArrayInputsReverseMode.C
```diff
@@ -1,4 +1,4 @@
-// RUN: %cladclang %s -I%S/../../include -Wno-unused-value -oArrayInputsReverseMode.out 2>&1 | FileCheck %s
+// RUN: %cladclang %s -I%S/../../include -Wno-unused-value -oArrayInputsReverseMode.out
 // RUN: ./ArrayInputsReverseMode.out | FileCheck -check-prefix=CHECK-EXEC %s
```
vgvassilev, line 2:
```suggestion
// RUN: %cladclang %s -I%S/../../include -Wno-unused-value -oArrayInputsReverseMode.out 2>&1 | FileCheck %s
// RUN: ./ArrayInputsReverseMode.out | FileCheck -check-prefix=CHECK-EXEC %s
// RUN: %cladclang --enable-tbr %s -I%S/../../include -Wno-unused-value -oArrayInputsReverseModeTBR.out
// RUN: ./ArrayInputsReverseModeTBR.out | FileCheck -check-prefix=CHECK-EXEC %s
```

lib/Differentiator/TBRAnalyzer.cpp
```diff
@@ -0,0 +1,814 @@
...
+    /// For example, in ``(x*=y)=u`` x's location will first be marked as
+    /// required to be stored (when passing *= operator) but then marked as not
+    /// required to be stored (when passing = operator). Current method of
+    /// marking locations does not allow to differentiate between these two.
+    TBRLocs.insert(E->getBeginLoc());
+  }
+}
+
+void TBRAnalyzer::setIsRequired(const clang::Expr* E, bool isReq) {
+  if (!isReq ||
```
parth-07, line 302: Just to verify, shouldn't this be `if (isReq || ...)` instead of `if (!isReq || ...)`?

PetroZarytskyi: @parth-07 It should not. The logic is as follows: We mark ``E`` as required only in a specific mode. When marking ``E`` as not required, we don't care about modes.

include/clad/Differentiator/TBRAnalyzer.h
```diff
@@ -0,0 +1,329 @@
...
+  std::vector<std::unique_ptr<VarsData>> blockData;
+
+  /// Stores the number of performed passes for a given CFG block index.
+  std::vector<short> blockPassCounter;
+
+  /// ID of the CFG block being visited.
+  unsigned curBlockID{};
+
+  /// The set of IDs of the CFG blocks that should be visited.
+  std::set<unsigned> CFGQueue;
```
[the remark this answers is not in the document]
PetroZarytskyi: @parth-07  CFG blocks should be analyzed in the order they go in the code and not in the order we add them to the queue. This can be done by picking the biggest index in ``CFGQueue`` (in ``clang::CFG``. blocks are numerated from the exit to the entry). In other words, we always take the last index in ``CFGQueue`` since ``std::set`` sorts its elements.

lib/Differentiator/TBRAnalyzer.cpp
parth-07: @PetroZarytskyi Oh, okay. Thank you for the information.

test/Arrays/ArrayInputsReverseMode.C
```diff
@@ -133,25 +136,26 @@ float func2(float* a) {
...
 //CHECK-NEXT:         _t0++;
-//CHECK-NEXT:         sum += helper(clad::push(_t3, a[clad::push(_t1, i)]));
+//CHECK-NEXT:         clad::push(_t1, sum);
+//CHECK-NEXT:         sum += helper(a[i]);
 //CHECK-NEXT:     }
 //CHECK-NEXT:     goto _label0;
 //CHECK-NEXT:   _label0:
 //CHECK-NEXT:     _d_sum += 1;
 //CHECK-NEXT:     for (; _t0; _t0--) {
+//CHECK-NEXT:         i--;
```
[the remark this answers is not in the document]
PetroZarytskyi: ``i--`` is done to restore the value of ``i`` and instead of using clad::tape. This has nothing to do with the loop.

include/clad/Differentiator/TBRAnalyzer.h
```diff
@@ -0,0 +1,329 @@
...
+    VarDataType type = UNDEFINED;
+    VarDataValue val;
+
+    VarData() = default;
+    VarData(const VarData& other) = delete;
+    VarData(VarData&& other) noexcept: type(other.type) {
+      *this = std::move(other);
+    }
+    VarData& operator=(const VarData& other) = delete;
+    VarData& operator=(VarData&& other) noexcept {
```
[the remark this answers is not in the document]
vgvassilev: I believe I have addressed that comment. Please check if what I implemented is right.

vgvassilev:
We are closing this PR because it became too big for github to render. We continue with https://github.com/vgvassilev/clad/pull/655

PR #626, opened 2023-09-13
Fix call expr to functor inside a function

### LINKED ISSUE #625: Failure in differentiating call expression to a function object
Results in the following assertion failure:
```
ReverseModeVisitor.cpp:1463: clad::StmtDiff clad::ReverseModeVisitor::VisitCallExpr(const clang::CallExpr*): Assertion `!utils::isArrayOrPointerType(arg->getType()) && "Arguments passed by pointers should be covered in pass by " "reference calls"' failed.
```

Minimal reproducible example (taken from https://github.com/vgvassilev/clad/discussions/618) :
```cpp
# include "clad/Differentiator/Differentiator.h"
# include <iostream>

struct FunObj {
   double operator()(double *x, size_t n)
   {  double ret = 0;
      for (size_t i = 0; i < n; i++) {
         ret += (i+1) * x[i];
      }
      return ret;
   }
};

double f(double *x, size_t n){
  FunObj f_obj;
  return f_obj(x, n);
}

int main() {
   size_t n = 3;
   double x[] = {1.0, 2.0, 3.0};
   double fx = f(x, n);
   std::cout << "f(x, n) = " << fx << "\n";

   double dx[] = {0.0, 0.0, 0.0};
   clad::array_ref<double> dx_ref(dx, n);
   f_dx.execute(x, n, dx_ref);

   for(size_t i = 0; i < n; ++i)
      std::cout << "df/dx[" << i << "] = " << dx[i] << "\n";
   return 0;
}
```

vaithak:
This fixes #625 for the case when a functor is indirectly differentiated because of a call expression inside a function.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -470,7 +470,8 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     if (m_ExternalSource)
       m_ExternalSource->ActAfterParsingDiffArgs(request, args);
 
-    auto derivativeName = request.BaseFunctionName + "_pullback";
+    auto derivativeName =
+        utils::ComputeEffectiveFnName(m_Function) + "_pullback";
     auto DNI = utils::BuildDeclarationNameInfo(m_Sema, derivativeName);
 
     auto paramTypes = ComputeParamTypes(args);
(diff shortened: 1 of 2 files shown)


vgvassilev (approved):
LGTM!

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1412,12 +1413,20 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     // statements there later.
     std::size_t insertionPoint = getCurrentBlock(direction::reverse).size();
 
+    // `CXXOperatorCallExpr` have the `base` expression as the first argument.
+    size_t skipFirstArg = 0;
```
parth-07, line 1417: Apologies for the late review.

What is the main issue if we do not skip the first arg? I suppose it might be required if we modify the functor's member variables in the call operator.

vaithak: The first arg in the operator call is the base expression, I used the logic of skipping the first arg from this in forward mode: https://github.com/vgvassilev/clad/blob/master/lib/Differentiator/BaseForwardModeVisitor.cpp#L980

PR #628, opened 2023-09-18
Initial support for expression templates in array and array_ref class

vaithak:
This is to speed up the operations for vector mode by ensuring that temporary array objects are not created when performing arithmetic operations of vectors.
Reference used: https://indico.cern.ch/event/450743/contributions/1116430/attachments/1224060/1790963/out-iCSC2016-L1.pdf

### DIFF
--- a/include/clad/Differentiator/Array.h
+++ b/include/clad/Differentiator/Array.h
@@ -1,6 +1,7 @@
 #ifndef CLAD_ARRAY_H
 #define CLAD_ARRAY_H
 
+#include "clad/Differentiator/ArrayExpression.h"
 #include "clad/Differentiator/CladConfig.h"
 
 #include <assert.h>
@@ -36,7 +37,7 @@ public:
 
   template <typename U>
   CUDA_HOST_DEVICE array(U* a, std::size_t size)
-      : m_arr(new T[size]{static_cast<T>(T())}), m_size(size) {
+      : m_arr(new T[size]), m_size(size) {
     for (std::size_t i = 0; i < size; ++i)
       m_arr[i] = static_cast<T>(a[i]);
   }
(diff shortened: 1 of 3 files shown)


vaithak:
This improves the timing for the weighted sum benchmark by a factor of 3 (from around 700ns to about 240ns), but this can be improved further. For the benchmark of the simple case, this doesn't add any improvement mainly because it consists of a loop of the following form:
```cpp
for (int i = 0; i < dim; i++)
    r += p[i];
```
The reason is that this current implementation converts from `array_expression` back to `array` on the usage of the `+=` operator. 
A possibly better solution can be only to evaluate the expression when using indexing functions, i.e. `[i]` and `slice()` operations.

I will send a separate PR for this improvement.

include/clad/Differentiator/Array.h
```diff
@@ -49,6 +50,14 @@ template <typename T> class array {
       m_arr[i] = arr[i];
   }
 
+  template <typename L, typename BinaryOp, typename R>
+  CUDA_HOST_DEVICE array(std::size_t size,
+                         const array_expression<L, BinaryOp, R>& expression)
+      : m_arr(new T[size]{static_cast<T>(T())}), m_size(size) {
```
vgvassilev, line 56: Do we need to zero the array?

vaithak: I don't think it is required, given that we are refilling all the entries in the loop anyway. This is the same as in the other constructors.

vgvassilev: Perhaps we should get rid of them as well...

vaithak: removed 👍🏼

include/clad/Differentiator/Array.h
```diff
@@ -229,6 +238,15 @@ template <typename T> class array {
       m_arr[i] *= static_cast<T>(arr[i]);
     return *this;
   }
+  /// Initializes the clad::array from the given clad::array_expression
+  template <typename L, typename BinaryOp, typename R>
+  CUDA_HOST_DEVICE array<T>&
+  operator=(const array_expression<L, BinaryOp, R>& arr_exp) {
+    assert(arr_exp.size() == m_size);
+    for (std::size_t i = 0; i < m_size; i++)
+      m_arr[i] = static_cast<T>(arr_exp[i]);
```
vgvassilev, line 247: Why do we need the `static_cast` here. Doesn't that signal a potential issue due to type incompatibility?

vaithak: removed 👍🏼

vgvassilev (commented):
This mostly looks good to me. Are we covered with tests? Perhaps we should write a benchmark making sure we do not regress the implementation?

vgvassilev (approved):
LGTM!

Issue #627, opened 2023-09-17
Failure in differentiating class methods with non default constructor


vaithak:
Minimal example:
```cpp
# include "clad/Differentiator/Differentiator.h"
# include <iostream>

class Experiment {
private:
  size_t n_;
public:
   Experiment(size_t n) : n_(n) {}

   double multiply(double x) {
     return n_*x;
   }
};

double f(double x, size_t n){
  Experiment E(n);
  return E.multiply(x); // (x*n)
}

int main() {
   size_t n = 3;
   double x = 4;
   double fx = f(x, n);
   std::cout << "f(x, n) = " << fx << "\n";
   
   auto f_dx = clad::gradient(f, "x");
   double dx = 0;
   f_dx.execute(x, n, &dx);
   std::cout << "df/dx = " << dx << "\n";
   return 0;
}
```

The starting of the generated function of `f_grad_0` contains the following snippet:
```cpp
void f_grad_0(double x, size_t n, clad::array_ref<double> _d_x) {
    size_t _d_n = 0;
    Experiment _d_E(<recovery-expr>({}));
    double _t0;
    Experiment _t1 = <recovery-expr>();
    Experiment E(n);
    _t0 = x;
    _t1 = E;
    ...
}
```

Where the recovery expressions are because of the following error messages:
```
error: invalid operands to binary expression ('size_t' (aka 'unsigned long') and 'Experiment')
error: no matching constructor for initialization of 'Experiment'
```

vaithak:
One of the root causes for the second error is that some user-defined classes may not have a default constructor, and in reverse mode, we create some global variables at the start of the function and are assigned values later, for ex., in the above example, `t1` is declared in the start and assigned later to E. 
I can't think of any way other than to explicitly tell the user to make sure that custom classes have a default constructor.

For the first error, still debugging it.

PR #630, opened 2023-09-21
Fix DifferentiateVarDecl for constructors in reverse mode

vaithak:
This fixes the error in #627 related to invalid operands in a binary expression.

A very minimal example that was failing earlier:
```cpp
class Experiment {
   
public:
    Experiment(int n=0){}
};

double f(double x, int n) {
  Experiment E(n);
  return 0.0; 
}

int main() {
}
```


vgvassilev (approved):
LGTM!

vgvassilev:
The codecov report seems invalid. Let’s merge it.

PR #632, opened 2023-10-02
Add fix for const parameters

vaithak:
This was troublesome when we created `clad::array<const double>` type arrays when the parameter was `const double` type, for example.

This was reported in the [discussion here](https://github.com/vgvassilev/clad/discussions/618).


vgvassilev (approved):
LGTM!

PR #634, opened 2023-10-03
Add benchmark for expression templates

vaithak:
- also added `const&` qualifiers while creating `array_expression` objects to avoid the copying of arrays.

### DIFF
--- a/include/clad/Differentiator/Array.h
+++ b/include/clad/Differentiator/Array.h
@@ -58,6 +58,13 @@ public:
       m_arr[i] = expression[i];
   }
 
+  template <typename L, typename BinaryOp, typename R>
+  CUDA_HOST_DEVICE array(const array_expression<L, BinaryOp, R>& expression)
+      : m_arr(new T[expression.size()]), m_size(expression.size()) {
+    for (std::size_t i = 0; i < expression.size(); ++i)
+      m_arr[i] = expression[i];
+  }
+
   // initializing all entries using the same value
   template <typename U>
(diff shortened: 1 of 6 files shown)


benchmark/mybenchmark
vgvassilev, line 1: We probably do not need this file.

vaithak: added it by mistake, removed 👍🏼

include/clad/Differentiator/Array.h
```diff
@@ -293,17 +300,19 @@ template <typename T> class array {
   }
 
   /// Negate the array and return a new array.
-  CUDA_HOST_DEVICE array_expression<T, BinarySub, array<T>> operator-() const {
-    return array_expression<T, BinarySub, array<T>>(static_cast<T>(0), *this);
+  CUDA_HOST_DEVICE array_expression<T, BinarySub, array<T> const&>
+  operator-() const {
+    return array_expression<T, BinarySub, array<T> const&>(static_cast<T>(0),
```
vgvassilev, line 305:
```suggestion
    return array_expression<T, BinarySub, const array<T>&>(static_cast<T>(0),
```

We spell const in clad this way. Can you update the rest of the occurrences?

vaithak: fixed and also added to .clang-format

vgvassilev: nice!

vgvassilev (approved):
LGTM!

PR #638, opened 2023-10-08
Add support for call expressions in vector forward mode AD

vaithak:
Two main changes in this PR:
- Renamed `ForwardModeVisitor` to a more apt name `PushForwardModeVisitor`.
- Made changes to make the `VisitCallExpr` of BaseForwardModeVistor and `the PushForwardModeVisitor` class generic enough to add support for vector mode with very little code repetition,
  - This led to the following inheritance structure, 
     ```
                                          BaseForwardModeVisitor
                                        /                        \
                                       /                          \
                VectorForwardModeVisitor                          PushForwardModeVisitor
                                       \                           /
                                        \                         /
                                        VectorPushForwardModeVisitor
     ```
     To avoid the diamond problem, added virtual inheritance between classes.

Also added appropriate tests for the new functionality.


parth-07:
This change looks good overall. I would like to avoid diamond hierarchy if possible, but I am not sure if it would be. So many things can easily go wrong when diamon hierarchy is used.

vaithak:
Fixed the inheritance structure to remove the diamond problem, now, the structure is pretty simple:
```
                                     BaseForwardModeVisitor
                                   /                        \
                                  /                          \
           VectorForwardModeVisitor                          PushForwardModeVisitor
                       |               
                       |                         
         VectorPushForwardModeVisitor
```
The common code in both PushForwardMode modes is pushed into `BaseForwardMode` class with the differences being a separate

vgvassilev:
@vaithak, can you rebase this PR on top of master?

PR #640, opened 2023-10-14
Fix tape push expr for clad array in reverse mode

### LINKED ISSUE #635: Bug: Adding dummy loop breaks gradient in special circumstances
Simply adding a dummy loop with one iteration breaks certain gradients.

These minimal conditions are required to reproduce the problem:

1. A loop
2. Some "repacking" parameters from an array in a new array (commonly done in RooFit when forwarding subsets of parameters)
3. Doing some calculations with the repacked parameters in another function

My setup: Arch Linux with `Clad` master and clang 16.06.

I hope this problem can be fixed, as right now it is blocking me from implementing numeric integration routines in RooFit (similar to the reproducer, but the loop is not trivial).

Reproducer:
```c++
// Compile with clang++ -std=c++11 -fplugin=/usr/lib/clad.so cladIssue.cxx -o cladIssue
// Add the -DFIXED to see the fixed version

#include <iostream>
#include <vector>

double func(double *params)
{
   return 1.0 / (params[0] * params[0]);
}

{
   double out = 0.0;

#ifndef FIXED
   for (std::size_t i = 0; i < 1; ++i)
#else
#endif
   {
      double paramsPrime[] = {params[0]};
      out = out + func(paramsPrime);
   }

   return out;
}

int main()
{

   double x = 2.0;
   double result = 0.0;
   grad.execute(&x, &result);

}
```

vaithak:
Using comma expressions for pushing a clad array onto a tape generated a warning and also resulted in invalid results in some cases.

Fixes #635

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -95,7 +95,12 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     Expr* exprToPush = E;
     if (auto AT = dyn_cast<ArrayType>(E->getType())) {
       Expr* init = getArraySizeExpr(AT, m_Context, *this);
-      exprToPush = BuildOp(BO_Comma, E, init);
+      llvm::SmallVector<Expr*, 2> pushArgs{E, init};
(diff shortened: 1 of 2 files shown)


vgvassilev (approved):
LGTM!

PR #644, opened 2023-10-20
Set derivative to 0 for fxn calls with literal arguments

### LINKED ISSUE #642: Improvement: Avoid unexpected warnings about num-diff fallback
If I compile this example here:
```c++
// Compile with clang++ -std=c++11 -fplugin=/usr/lib/clad.so cladConfusingWarnings.cxx -o cladConfusingWarnings

#include <iostream>

double func(double x)
{
   return x * std::tanh(1.0);
}

int main()
{
   double x = 2.0;
   std::cout << func(x) << std::endl;

   auto grad = clad::gradient(func, "x");

   double result = 0.0;
   grad.execute(x, &result);

}
```

Then I get a confusing warning:
```txt
warning: Falling back to numerical differentiation for 'tanh' since no suitable overload was found and clad could not derive it. To disable this feature, compile your programs with -DCLAD_NO_NUM_DIFF.
```

That is unexpected, because `tanh(1.0)` is a constant that doesn't need to be differentiated. Why is the fallback needed? I would expect no warning to show up in any case.

It would be good to suppress this warning, because it happens in the RooFit HistFactory likelihoods, where the `LnGamma` from ROOT math is used for constant expressions.

vaithak:
Fixes #642.

Example taken from the issue - also added a similar one in the tests:
```cpp
double func(double x)
{
   return x * std::tanh(1.0);  // -----> tanh(1.0) is a const.
}
```

### DIFF
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -1055,6 +1055,36 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
             .get();
   }
 
+  // If all arguments are constant literals, then this does not contribute to
+  // the gradient.
+  // FIXME: revert this when this is integrated in the activity analysis pass.
+  if (!callDiff) {
+    if (!isa<CXXOperatorCallExpr>(CE) && !isa<CXXMemberCallExpr>(CE)) {
+      bool allArgsAreConstantLiterals = true;
+      for (unsigned i = 0, e = CE->getNumArgs(); i < e; ++i) {
+        const Expr* arg = CE->getArg(i);
+        // if it's of type MaterializeTemporaryExpr, then check its
+        // subexpression.
+        if (const auto* MTE = dyn_cast<MaterializeTemporaryExpr>(arg))
+          arg = clad_compat::GetSubExpr(MTE);
+        if (!isa<FloatingLiteral>(arg) && !isa<IntegerLiteral>(arg)) {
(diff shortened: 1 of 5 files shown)


vgvassilev:
@PetroZarytskyi, could we capture such cases where a subexpression does not contribute to the requested derivative with activity analysis?

vaithak:
> @PetroZarytskyi, could we capture such cases where a subexpression does not contribute to the requested derivative with activity analysis?

That's a valid point; this PR won't be able to capture more complex cases, for example, the composition of function: `f(g(h(2.0)`, which again is a constant but requires tree traversal.

PetroZarytskyi:
> @PetroZarytskyi, could we capture such cases where a subexpression does not contribute to the requested derivative with activity analysis?

I think so. This sounds like a partial case of activity analysis.

vgvassilev:
> > @PetroZarytskyi, could we capture such cases where a subexpression does not contribute to the requested derivative with activity analysis?
> 
> I think so. This sounds like a partial case of activity analysis.

Can we keep track of this and remove it when we implement the activity analysis? I’d merge this but we need to remember to revert it when we implement the analysis.

PetroZarytskyi:
> > > @PetroZarytskyi, could we capture such cases where a subexpression does not contribute to the requested derivative with activity analysis?
> > 
> > 
> > I think so. This sounds like a partial case of activity analysis.
> 
> Can we keep track of this and remove it when we implement the activity analysis? I’d merge this but we need to remember to revert it when we implement the analysis.

Sounds reasonable to me.

vgvassilev (approved):
LGTM!

PR #645, opened 2023-10-24
Fix gradient computation of higher order functions

### LINKED ISSUE #637: Feature request: Support for higher order functions
It would be very nice if differentiating through higher order functions would work also in Clad.

It happens often that one has an algorithm that is implemented in a header-only library, where one or the parameters is a function.

This would be relevant for example when using the numeric integration algorithms from ROOT/RooFit inside a function that one wants to differentiate with Clad.

Here is an example of what I such a pattern would look like:

```c++
// Compile with clang++ -std=c++11 -fplugin=/usr/lib/clad.so cladNestedFuncs.cxx -o cladNestedFuncs

#include <iostream>
#include <vector>

// Supporting templates would also be good (in fact even better for my usecase)
// template<class Func>
// double algo(Func func, double x) {}
double algo(double (*func)(double), double x)
{
   double val = func(x);
   return val * val;
}

double func(double x)
{
   return x + 1;
}

{
   return algo(func, params[0]);
}

int main()
{
   double x = 2.0;
   std::cout << wrapper(&x) << std::endl;

   double result = 0.0;
   grad.execute(&x, &result);

}
```

Or maybe I missed that there is already an existing solution for this?

Thank you very much for considering this!

vaithak:
fixes #637

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1345,9 +1345,13 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         }
         // Create the (_d_param[idx] += dfdx) statement.
         if (dfdx()) {
-          Expr* add_assign = BuildOp(BO_AddAssign, it->second, dfdx());
-          // Add it to the body statements.
-          addToCurrentBlock(add_assign, direction::reverse);
+          // FIXME: not sure if this is generic.
+          // Don't update derivatives of non-record types.
+          if (!decl->getType()->isRecordType()) {
+            auto* add_assign = BuildOp(BO_AddAssign, it->second, dfdx());
+            // Add it to the body statements.
+            addToCurrentBlock(add_assign, direction::reverse);
+          }
         }
         return StmtDiff(clonedDRE, it->second, it->second);
       }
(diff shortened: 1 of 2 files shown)


vaithak:
The change seems to be failing with clang-16. I will update the clang locally to test the error as currently I have version 15 installed.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1345,9 +1345,13 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         }
         // Create the (_d_param[idx] += dfdx) statement.
         if (dfdx()) {
-          Expr* add_assign = BuildOp(BO_AddAssign, it->second, dfdx());
-          // Add it to the body statements.
-          addToCurrentBlock(add_assign, direction::reverse);
+          // FIXME: not sure if this is generic.
+          // Don't update derivatives of non-record types.
+          if (!decl->getType()->isRecordType()) {
```
vgvassilev, line 1350: Can we check this with `isa<RecordDecl>(decl)`?

vaithak: I had that earlier but it failed with clang-16. `isRecordType` does the same thing inside it but performs the operation on the CanonicalType instead of the qualified type.
[source code of isRecordType](https://clang.llvm.org/doxygen/Type_8h_source.html#l07075).

vgvassilev: Ah, ok. That means that if we did `isa<RecordDecl>(decl->getCanonicalDecl())` would work?

vaithak: Yup, that would work. I saw `isRecordType` being used at different places in our codebase, so I used that.
Should I change this to use `isa<RecordDecl>`? Any reason to prefer this?

vgvassilev: Nope, let's use your current solution if it is used elsewhere.

PR #647, opened 2023-10-26
Add support for std::min, std::max and std::clamp functions

### LINKED ISSUE #641: Feature request: support `std::max`, `std::min`, and `std::clamp`
You guys are amazing amazing at dealing with issues on GitHub! :clap:

That encourages me to open more :smile: Note that I don't open them just for the hell of it, but because these are actually affecting the affecting the application of Clad to RooFit (a few more issues will follow).

The `std::max`, `std::min`, and to a lesser extent the `std::camp` functions are used a lot in RooFit. It would really help if these get handled by Clad without issues, just like `std::exp`.

Also, note that right now it just crashes without telling you what the problem is, which reinforces the idea of improving the error messages like we talked about briefly in [this issue #636](https://github.com/vgvassilev/clad/issues/636).

Here is a reproducer that shows it doesn't work right now (note that C++17 is required for `std::clamp`):

```c++
// Compile with clang++ -std=c++17 -fplugin=/usr/lib/clad.so cladMax.cxx -o cladMax

#include <algorithm>
#include <iostream>

double func(double x)
{
   double out = x;
   out = std::max(out, 0.0);
   out = std::min(out, 10.0);
   out = std::clamp(out, 3.0, 7.0);
   return out;
}

int main()
{
   double x = 2.0;
   std::cout << func(x) << std::endl;

   auto grad = clad::gradient(func, "x");

   double result = 0.0;
   grad.execute(x, &result);

}
```

vaithak:
fixes #641

### DIFF
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -187,6 +187,10 @@ namespace clad {
           continue;
         }
         if (DC2->isInlineNamespace()) {
+          // Inline namespace can be skipped from context, because its members
+          // are automatically searched from the parent namespace.
+          // This will also help us to deal with intermediate inline namespaces
+          // like std::__1::, as present in std functions for libc++.
           DC2 = DC2->getParent();
           continue;
         }
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -158,8 +158,9 @@ namespace clad {
       auto typePtr = QT.getTypePtr();
       if (typePtr->isRecordType() && !typePtr->getAs<clang::ElaboratedType>()) {
         CXXScopeSpec CSS;
-        clang::CXXRecordDecl const* recordDecl = typePtr->getAsCXXRecordDecl();
-        clang::DeclContext const* declContext = static_cast<clang::DeclContext const*>(recordDecl);
+        const clang::CXXRecordDecl* recordDecl = typePtr->getAsCXXRecordDecl();
+        const auto* declContext =
+            static_cast<const clang::DeclContext*>(recordDecl);
         utils::BuildNNS(semaRef, const_cast<clang::DeclContext*>(declContext), CSS);
         NestedNameSpecifier* NS = CSS.getScopeRep();
         if (auto* Prefix = NS->getPrefix())
@@ -183,6 +184,10 @@ namespace clad {
           break;
         if (isa<LinkageSpecDecl>(DC2)) {
           DC2 = DC2->getParent();  
+          continue;
+        }
+        if (DC2->isInlineNamespace()) {
+          DC2 = DC2->getParent();
           continue;
         }
         // We don't want to 'extend' the DC1 context with class declarations.
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -13,6 +13,7 @@ namespace custom_derivatives{}
 #include "clad/Differentiator/ArrayRef.h"
 #include "clad/Differentiator/CladConfig.h"
 
+#include <algorithm>
 #include <cmath>
 
 namespace clad {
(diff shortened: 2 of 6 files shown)


vaithak:
The `clang-tidy` comments seem to be false positives. The tests seem to be failing on osx, I will look into this.

parth-07 (approved):
Looks good.

vaithak:
Some comments on why it was failing for `osx` earlier (it took me a lot of time, `tmate sessions` and debug statements to figure this out):
- As @vgvassilev hinted in the discussion sync, the crux of the issue turned out to be osx using by default libc++ instead of libstdc++. 
  - The former encloses some of the functions including the ones added in this PR (i.e. min, max and clamp) in the `inline namespace __1` inside `std`, so our logic expected custom derivatives to be found in `custom_derivatives::std::__1` namespace, which was wrong.
- To fix this, I have added the following statement to skip inline namespaces: https://github.com/vgvassilev/clad/pull/647/files#diff-b0fea173a42567ba1b3a2d6209e469d1bd595873f695fe71ad1fa80068fb9a4cR189.

Codecov seems to be reporting an issue because the inline namespace change would only execute with `libc++` (or osx), whereas the runtime for coverage action is ubuntu, so it reports that line as not covered.

parth-07:
I think the inline namespace change is good. However, it can complicate things too. What if a user has defined a function in a user-defined inline namespace, `SomeNS::SomeInlineNS::ComputeSomething`? In this case, a user cannot define a custom derivative in the intuitive namespace `custom_derivatives::SomeNS::SomeInlineNS`. @vaithak  What do you think about this?

vaithak:
> I think the inline namespace change is good. However, it can complicate things too. What if a user has defined a function in a user-defined inline namespace, `SomeNS::SomeInlineNS::ComputeSomething`? In this case, a user cannot define a custom derivative in the intuitive namespace `custom_derivatives::SomeNS::SomeInlineNS`. @vaithak What do you think about this?

Hi @parth-07, that is a valid point, but it will still work, mainly because we can skip the inline namespace while calling the custom derivative function. I have added a test case to demonstrate that this will work.
This also fixes the Codecov issue.

vgvassilev:
I was wondering if we could optionally for libc++ define move the custom derivatives in the same inline namespace. Something like ifdef libc++; in-line namespace ___1; …pullback…

vaithak:
> I was wondering if we could optionally for libc++ define and move the custom derivatives in the same inline namespace. Something like ifdef libc++; in-line namespace ___1; …pullback…

Yeah, we can do that too. 
My understanding is that the purpose of inline namespaces is to hide the internal structure from the APIs. 
If so, I have some queries regarding this:
 - Is the name `__1` for this namespace stable?  By looking at it, it seems something that can change for future versions. 
 - Another case could be that someone may compile and generate the differentiation code on one machine (for ex., a personal OSX laptop) and try to copy and run that code on a different machine (for ex., a running Linux instance), if we generate the code with `std::_1`, it may need modifications before running.
 - Similarly, for user-defined namespace, as @parth-07 mentioned. If a user uses a library function, ex. `NS1::func1`. Let's say the user wants to define the custom derivative of this function manually. Do we expect the user to know the inline namespace for this function? For ex., it took me a good amount of time to figure out that there was a `__1` inline namespace because the documentation doesn't expose this. 
 In case the user doesn't know, they will define the `custom_derivatives::NS1::func1_pullback` without any inline namespace, which will error out if the extra `if-condition` in the PR is not handled.

parth-07:
> Is the name __1 for this namespace stable? By looking at it, it seems something that can change for future versions.

Yes. It seems to be implementation-defined. We should not rely on it if possible.

> In case the user doesn't know, they will define the custom_derivatives::NS1::func1_pullback without any inline namespace, which will error out if the extra if-condition in the PR is not handled.

With the current code, if a user has defined `custom_derivatives::NS1::InlineNS::func1_pullback`, then that will work as well, right?

parth-07:
> With the current code, if a user has defined custom_derivatives::NS1::InlineNS::func1_pullback, then that will work as well, right?

Oh, yes, it will. I missed the test for this before. @vgvassilev with the @vaithak approach users can define custom derivatives with and without the inline namespace. I like the @vaithak idea here to skip the inline namespace when finding the custom derivative. It works because inline namespace members are automatically searched as well when finding members in the enclosing namespace.

lib/Differentiator/CladUtils.cpp
```diff
@@ -185,6 +186,10 @@ namespace clad {
           DC2 = DC2->getParent();  
           continue;
         }
+        if (DC2->isInlineNamespace()) {
```
vgvassilev, line 189: Can we add a clarification comment for this code and what it enables here?

vaithak: added 👍🏼

vgvassilev (approved):
LGTM!

PR #655, opened 2023-11-21
Add TBR analysis to the reverse mode in Clad.

PetroZarytskyi:
This pull request optimizes storing and restoring in the reverse mode of Clad and introduces TBR (To-Be-Recorded) analysis to determine when variables should be stored.


test/Gradient/FunctionCalls.C
```diff
@@ -698,8 +570,4 @@ int main() {
   TEST_ARR5(fn5, arr, 5);       // CHECK-EXEC: {5.00, 1.00, 0.00, 0.00, 0.00}
   TEST2(fn6, 3, 5);             // CHECK-EXEC: {5.00, 3.00}
   TEST2(fn7, 3, 5);             // CHECK-EXEC: {10.00, 71.00}
-  TEST2(fn8, 3, 5);             // CHECK-EXEC: {7.62, 4.57}
-  TEST2(fn9, 3, 5);             // CHECK-EXEC: {5.00, 3.00}
-  TEST2(fn10, 8, 5);            // CHECK-EXEC: {0.00, 7.00}
-  TEST2(fn11, 3, 5);            // CHECK-EXEC: {1.00, 1.00}
```
vgvassilev, line 704: We need to bring these changes back.

test/Gradient/Loops.C
```diff
@@ -1,4 +1,7 @@
-// RUN: %cladclang %s -I%S/../../include -oReverseLoops.out 2>&1 | FileCheck %s
+/// FIXME: Enable FileCheck for this test.
```
vgvassilev, line 1: We should enable this.

test/Gradient/UserDefinedTypes.C
```diff
@@ -1,4 +1,9 @@
-// RUN: %cladclang %s -I%S/../../include -oUserDefinedTypes.out 2>&1 | FileCheck %s
```
vgvassilev, line 1: We should probably re-enable this, too.

test/Gradient/Assignments.C
```diff
@@ -494,26 +574,38 @@ double f12(double x, double y) {
...
-//CHECK-NEXT:       double _ref0 = (_cond0 ? (t = x) : (t = y));
-//CHECK-NEXT:       _t1 = _ref0;
-//CHECK-NEXT:       _t0 = y;
-//CHECK-NEXT:       _ref0 *= _t0;
+//CHECK-NEXT:       if (_cond0)
+//CHECK-NEXT:           _t0 = t;
+//CHECK-NEXT:       else
+//CHECK-NEXT:           _t1 = t;
+//CHECK-NEXT:       double &_t2 = (_cond0 ? (t = x) : (t = y));
+//CHECK-NEXT:       _t3 = t;
```
parth-07, line 587: Why do we have both `_t3` and `_t4` storing `t` value?

test/Gradient/Assignments.C
```diff
@@ -531,25 +623,24 @@ double f13(double x, double y) {
...
 //CHECK-NEXT:       _t0 = (y = x);
-//CHECK-NEXT:       double t = _t1 * _t0;
-//CHECK-NEXT:       _t3 = t;
+//CHECK-NEXT:       double t = x * _t0;
 //CHECK-NEXT:       _t2 = y;
 //CHECK-NEXT:       goto _label0;
-//CHECK-NEXT:     _label0:
+//CHECK-NEXT:       _label0:
 //CHECK-NEXT:       {
 //CHECK-NEXT:           double _r2 = 1 * _t2;
```
parth-07, line 633: Based on the new approach, shouldn't this instead be, `double _r2 = 1 * y`?

test/Gradient/Loops.C
```diff
@@ -24,16 +26,18 @@ double f1(double x) {
 //CHECK-NEXT:       _t0 = 0;
 //CHECK-NEXT:       for (int i = 0; i < 3; i++) {
 //CHECK-NEXT:           _t0++;
-//CHECK-NEXT:           clad::push(_t2, t);
-//CHECK-NEXT:           t *= clad::push(_t1, x);
+//CHECK-NEXT:           clad::push(_t1, t);
+//CHECK-NEXT:           t *= clad::push(_t2, x);
```
parth-07, line 30: Why do we need to store the value of `x` when it is not changing?

test/Gradient/Assignments.C
PetroZarytskyi: @parth-07 I described this problem to you and Vassil earlier. In general, in order to construct ``dl`` (which you need to differentiate ``L``), you need differentiated ``R``, but to construct ``dr`` (which you need to differentiate ``R``), you need differentiated ``L``. So ``_t2`` instead of y is a temporary easy solution to break this loop.

test/Gradient/Loops.C
PetroZarytskyi: @parth-07 You're making a valid point. I got confused that we would have the same issue as with multiplication differentiation. But here, the store is actually unnecessary.

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -43,6 +43,8 @@ namespace clad {
     bool CallUpdateRequired = false;
     /// A flag to enable/disable diag warnings/errors during differentiation.
     bool VerboseDiags = false;
+    /// A flag to enable TBR analysis during reverse-mode differentiation.
+    bool EnableTBRAnalysis = false;
```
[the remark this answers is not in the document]
vgvassilev: We should figure out how to ignore these suggestions for struct with public-only members. @vaithak do we know how to do it?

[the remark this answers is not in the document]
vaithak: I can't find any way to restrict this for structs only. There is way to have [different restrictions for public members](https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-naming.html#cmdoption-arg-PublicMemberPrefix), but it will apply to both classes and structs.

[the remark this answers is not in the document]
vgvassilev: I guess in that case for the public members we should all allow to be prefixed without `_m` if that's possible.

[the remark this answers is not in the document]
vaithak: Sure, we can add it. Does it make sense to add it in this PR or a separate one?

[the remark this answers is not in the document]
vgvassilev: Probably as a separate one and we can rebase this one on top.

test/Gradient/Assignments.C
parth-07: @PetroZarytskyi Oh, yes. I remember this discussion. Can you please tell me if I understand the problem correctly. When generating adjoint code for the below statement:
```
r = u * SomeExprV
```

To generate derivative statement `_d_u += _d_r * ...`, we first need to evaluate the expression `SomeExprV`. We do not need the derivative of `SomeExprV`, however, we do need the expression evaluation result to determine if we would benefit from a temporary variable in this case or not. 

---------------------

Ideally, if `SomeExprV` does not have any side-effects then we can avoid the store-altogether. Can you please open an issue to track this?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -556,6 +563,20 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
   }
 
   void ReverseModeVisitor::DifferentiateWithClad() {
+    TBRAnalyzer analyzer(m_Context);
+    if (enableTBR) {
+      analyzer.Analyze(m_Function);
+      m_ToBeRecorded = analyzer.getResult();
+    }
+
+    // for (auto pair : m_ToBeRecorded) {
```
vgvassilev, line 572: Can you remove this commented section?

vgvassilev (approved):
Thank you @PetroZarytskyi! LGTM, please address my last comment and update the commit message text and this should be good to go.

PR #664, opened 2023-12-03
Fix assertions seen in clang13 due to type mismatches.

vgvassilev:
When we synthesize the 0 constant we do not take into account if the corresponding type matches to IntTy. In case it does not match we need to add the necessary implicit casts.

This patch fixes an issue that became visible after landing PR #655.

### DIFF
--- a/lib/Differentiator/VisitorBase.cpp
+++ b/lib/Differentiator/VisitorBase.cpp
@@ -325,10 +325,14 @@ namespace clad {
   }
 
   Expr* VisitorBase::getZeroInit(QualType T) {
-    if (T->isScalarType())
-      return ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
-    else
-      return m_Sema.ActOnInitList(noLoc, {}, noLoc).get();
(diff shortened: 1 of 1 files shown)


lib/Differentiator/VisitorBase.cpp
```diff
@@ -325,10 +325,14 @@ namespace clad {
...
 
   Expr* VisitorBase::getZeroInit(QualType T) {
-    if (T->isScalarType())
-      return ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
-    else
-      return m_Sema.ActOnInitList(noLoc, {}, noLoc).get();
+    // FIXME: Consolidate other uses of synthesizeLiteral for creation 0 or 1.
+    if (T->isScalarType()) {
+      ExprResult Zero =
+          ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
```
parth-07, line 331: Instead of explicitly adding implicit cast expression, what do you think about directly creating the zero literal to be of the type `T`?

parth-07 (approved):

PR #674, opened 2023-12-08
Update our version of enzyme from 0.0.36 to 0.0.95.

vgvassilev:
Remove the patch we had on top and limit the versions where the enzyme backend is supported. Enzyme does not work on llvm older than 9 and newer than 12.


vgvassilev:
I believe this is an issue with the build infrastructure. Let's move forward.

alexander-penev (approved):

PR #675, opened 2023-12-08
Add LLVM17 support.

### LINKED ISSUE #669: Add support for clang17

vgvassilev:
Fixes #669

### DIFF
--- a/test/FirstDerivative/Loops.C
+++ b/test/FirstDerivative/Loops.C
@@ -378,7 +378,7 @@ double fn10_darg0(double x, size_t n);
 // CHECK-NEXT:     {
 // CHECK-NEXT:         size_t _d_count = 0;
 // CHECK-NEXT:         size_t _d_max_count = _d_n;
-// CHECK-NEXT:         for (size_t count = 0; max_count; ++count) {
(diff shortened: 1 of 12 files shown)


test/FirstDerivative/Loops.C
```diff
@@ -378,7 +378,7 @@ double fn10_darg0(double x, size_t n);
 // CHECK-NEXT:     {
 // CHECK-NEXT:         size_t _d_count = 0;
 // CHECK-NEXT:         size_t _d_max_count = _d_n;
-// CHECK-NEXT:         for (size_t count = 0; max_count; ++count) {
+// CHECK-NEXT:         for (size_t count = 0; size_t max_count = n; ++count) {
```
vgvassilev, line 381: @PetroZarytskyi, can you take a look at this change - I believe it was broken before that patch...

PetroZarytskyi: @vgvassilev Yes. I think this was a bug in clang with dumping for-loop condition inits. It had to do with the fact that technically in such cases the condition is ``max_count`` and ``size_t max_count = n`` is the condition init.

vgvassilev: Ok, I've fixed it by making it conditional in the test...

PR #680, opened 2023-12-11
Make const loop variables global and drop const.

### LINKED ISSUE #667: Regression in RooFit with Clad `master`
This is the standalone reproducer of what is happening when trying to use clad master inside of ROOT.

The gradients created for likelihood functions as defined in RooFit are not correct anymore with Clad `master` (ec76b70).

```c++
// Compile with clang++ -std=c++11 -fplugin=/usr/lib/clad.so gaussianNLL.cxx -o gaussianNLL

#include <iostream>
#include <vector>

double gaussian(double x, double mean, double sigma)
{
   const double arg = x - mean;
   const double sig = sigma;
   return std::exp(-0.5 * arg * arg / (sig * sig));
}

double gaussianIntegral(double xMin, double xMax, double mean, double sigma)
{
   // The normalisation constant 1./sqrt(2*pi*sigma^2) is left out in evaluate().
   // Therefore, the integral is scaled up by that amount to make RooFit normalise
   // correctly.
   double resultScale = std::sqrt(2 * M_PI) * sigma;

   // Here everything is scaled and shifted into a standard normal distribution:
   double xscale = std::sqrt(2.0) * sigma;
   double scaledMin = 0.;
   double scaledMax = 0.;
   scaledMin = (xMin - mean) / xscale;
   scaledMax = (xMax - mean) / xscale;

   // Here we go for maximum precision: We compute all integrals in the UPPER
   // tail of the Gaussian, because erfc has the highest precision there.
   // Therefore, the different cases for range limits in the negative hemisphere are mapped onto
   // the equivalent points in the upper hemisphere using erfc(-x) = 2. - erfc(x)
   double ecmin = std::erfc(std::abs(scaledMin));
   double ecmax = std::erfc(std::abs(scaledMax));

   double cond = 0.0;
   // Don't put this "prd" inside the "if" because clad will not be able to differentiate the code correctly (as of
   // v1.1)!
   double prd = scaledMin * scaledMax;
   if (prd < 0.0)
      cond = 2.0 - (ecmin + ecmax);
   else if (scaledMax <= 0.0)
      cond = ecmax - ecmin;
   else
      cond = ecmin - ecmax;
   return resultScale * 0.5 * cond;
}

double nll(double *params, double const *obs)
{
   constexpr int n = 54;

   double RooNLLVarNewWeightSum = 0.0;
   double result = 0.0;
   const double sigma_scaled = (params[2] * 1.500000);
   const double mu_shifted = (params[0] + params[1]);
   const double model_Int_x_ = gaussianIntegral(-10.000000, 10.000000, mu_shifted, sigma_scaled);
   for (int loopIdx = 0; loopIdx < n; loopIdx++) {
      const double model = gaussian(obs[loopIdx], mu_shifted, sigma_scaled);
      const double model_over_model_Int_x_ = model / model_Int_x_;
      result -= obs[n + loopIdx] * std::log(model_over_model_Int_x_);
   }

   return result;
}

int main()
{
   std::vector<double> params{0.0, 1.0, 3.0};
   std::vector<double> obs{
      -9.1, -8.1, -5.7, -5.5, -5.3, -4.9, -4.5, -4.3, -3.9, -3.7, -3.5, -3.1, -2.9, -2.5, -2.3, -1.7, -1.5, -1.3,
      -1.1, -0.9, -0.7, -0.5, -0.3, -0.1, 0.1,  0.3,  0.5,  0.7,  0.9,  1.1,  1.3,  1.5,  1.9,  2.1,  2.3,  2.7,
      3.1,  3.7,  4.5,  4.7,  4.9,  5.1,  5.5,  6.1,  6.3,  6.5,  7.1,  7.3,  7.7,  7.9,  8.9,  9.1,  9.3,  9.9,
      1.,   1.,   1.,   1.,   1.,   2.,   3.,   3.,   1.,   1.,   1.,   2.,   1.,   1.,   3.,   2.,   1.,   1.,
      3.,   3.,   3.,   2.,   3.,   3.,   3.,   2.,   2.,   4.,   3.,   4.,   1.,   1.,   1.,   2.,   3.,   2.,
      2.,   1.,   2.,   1.,   1.,   2.,   3.,   1.,   2.,   1.,   2.,   2.,   1.,   2.,   1.,   2.,   1.,   1.,
      1.,   1.,   1.,   1.,   1.,   2.,   3.,   3.,   1.,   1.,   1.,   2.,   1.,   1.,   3.,   2.,   1.,   1.,
      3.,   3.,   3.,   2.,   3.,   3.,   3.,   2.,   2.,   4.,   3.,   4.,   1.,   1.,   1.,   2.,   3.,   2.,
      2.,   1.,   2.,   1.,   1.,   2.,   3.,   1.,   2.,   1.,   2.,   2.,   1.,   2.,   1.,   2.,   1.,   1.};

   std::cout << "NLL value" << std::endl;
   const double nllVal = nll(params.data(), obs.data());
   std::cout << std::endl;

   auto grad = clad::gradient(nll, "params");

   std::vector<double> gradResult(3);
   grad.execute(params.data(), obs.data(), gradResult.data());

   std::cout << gradResult[1] << std::endl;
   std::cout << gradResult[2] << std::endl;
   std::cout << std::endl;

   auto numDiff = [&](int i) {
      const double eps = 1e-6;
      p[i] = params[i] - eps;
      double nllValDown = nll(p.data(), obs.data());
      p[i] = params[i] + eps;
      double nllValUp = nll(p.data(), obs.data());
   };

}
```
Output:
```
NLL value
281.704

Clad diff:
6.57277
6.57277
-53.536

Num diff:
-0.578694
-0.578694
-0.365516
```

PetroZarytskyi:
Currently, local variables in loops are moved to function globals (the initialization is replaced with an assignment in the same place). This is done for the variable's value to not get overwritten on the next loop iteration. However, this is currently not done for const variables since we didn't account for the fact they can be re-initialized with different values on different loop iterations. Therefore, we have to do the same process for const local variables but drop constness (they need to be reassigned every loop iteration).
Fixes #667.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2692,9 +2692,15 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         // assignments. This is a temporary measure to avoid the bug that arises
         // from overwriting local variables on different loop passes.
         if (isInsideLoop) {
-          if (VD->getType()->isBuiltinType() &&
-              !VD->getType().isConstQualified()) {
+          if (VD->getType()->isBuiltinType()) {
             auto* decl = VDDiff.getDecl();
+            /// The same variable will be assigned with new values every
+            /// loop iteration so the const qualifier must be dropped.
+            if (decl->getType().isConstQualified()) {
+              QualType nonConstType =
+                  getNonConstType(decl->getType(), m_Context, m_Sema);
+              decl->setType(nonConstType);
+            }
             if (decl->getInit()) {
               auto* declRef = BuildDeclRef(decl);
               auto pushPop =
@@ -2744,8 +2750,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     /// overwriting local variables on different loop passes.
     if (isInsideLoop) {
       if (auto* VD = dyn_cast<VarDecl>(decls[0])) {
-        if (VD->getType()->isBuiltinType() &&
-            !VD->getType().isConstQualified()) {
+        if (VD->getType()->isBuiltinType()) {
           addToBlock(DSClone, m_Globals);
           Stmt* initAssignments = MakeCompoundStmt(inits);
           initAssignments = unwrapIfSingleStmt(initAssignments);
--- a/test/Gradient/Loops.C
+++ b/test/Gradient/Loops.C
@@ -198,6 +198,55 @@ double f5(double x){
 //CHECK-NEXT:       }
 //CHECK-NEXT:   }
 
+double f_const_local(double x) {
+  double res = 0;
+  for (int i = 0; i < 3; ++i) {
+    const double n = x + i;
+    res += x * n;
+  }
+  return res;
+} // == 3x^2 + 3x
+
+//CHECK:   void f_const_local_grad(double x, clad::array_ref<double> _d_x) {
+//CHECK-NEXT:    double _d_res = 0;
+//CHECK-NEXT:    unsigned long _t0;
+//CHECK-NEXT:    int _d_i = 0;
+//CHECK-NEXT:    clad::tape<double> _t1 = {};
+//CHECK-NEXT:    double _d_n = 0;
+//CHECK-NEXT:    double n = 0;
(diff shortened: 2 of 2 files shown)


vgvassilev (approved):
Very well! Thank you!

PR #688, opened 2023-12-22
Add declarations for variables reused in the reverse pass

### LINKED ISSUE #659: Reusing declarations in the reverse pass
This issue is going to arise after landing #655.
   Reusing the same declarations for the reverse pass as the ones used in the forward pass does not cause any errors since scopes are currently reused as well. For example, if we differentiate this statement
```
if (1) {
    double n;
    n *= x;
}
```
We will get this statement for the forward pass 
```
if (1) {
    double n;
    _t0 = n;
    n *= x;
}
```
And this one for the reverse pass
```
if (1) {
    {
        n = _t0;
        double _r_d0 = _d_n;
        _d_n += _r_d0 * x;
        double _r0 = n * _r_d0;
        * _d_x += _r0;
        _d_n -= _r_d0;
        _d_n;
    }
}
```
   Both statements have the same scope and, therefore, using ``n`` in the reverse pass does not cause an error. However, compiling the dumped code by hand obviously doesn't work.
   Ideally, this should be solved by placing a separate declaration statement in the reverse pass. This would cause a couple of complications: 
1. The declaration should be inserted before the variable is used and that might be tricky because the order of statements in the reverse pass is reversed.
2. We should store the last value of a variable in the forward pass before it goes out of scope and then assign the corresponding variable in the reverse pass with that value. Also, it would be better to do this only when the value is useful (use TBR analysis).

   There is also an easier temporary solution to move the declaration to function globals and place an assignment instead of an initialization. For example this code
```
if (int k = 1) {
...
```
Can be replaced with
```
int k;
if (k = 1) {
...
```
Obviously, this will harm performance and decrease readability.

### LINKED ISSUE #681: Regression with Clad `master` in case of num-diff fallback in specific conditions
Some of the RooFit tests are still failing with Clad `master`, because of code that hits very specific conditions.

To reproduce the underlying problem, first create a library with any function:
```c++
// clang++ -fPIC mylib.cxx -c -o mylib.o && ar r libmylib.a mylib.o

double foo(double x, double y, double z)
{
   return x * y * z;
}
```
Then use that library in a function the is differentiated:
```c++
// Compile with clang++ -O2 -std=c++20 -fplugin=/usr/lib/clad.so repro_regression.cxx -o repro_regression -lmylib -L.

#include <iostream>
#include <vector>

double foo(double x, double y, double z);

{
   if (true) {
      double constant = 11.;
      return foo(params[0], constant, 1.);
   }

   return 0.0;
}

int main()
{
   std::vector<double> params{4};

   std::cout << "Value" << std::endl;
   const double nllVal = wrapper(params.data());
   std::cout << std::endl;

   std::vector<double> gradResult(1);
   grad.execute(params.data(), gradResult.data());

   std::cout << std::endl;

   auto numDiff = [&](int i) {
      const double eps = 1e-6;
      p[i] = params[i] - eps;
      p[i] = params[i] + eps;
   };

}
```

Many conditions need to be met to get this Clad error:
* You need to compile with `-O2`
* You need to call a function in a library such that Clad falls back to num. diff.
* The code needs to be in a branch, like this `if (true)` in the reproducer
* One of the function variables must be a named variable (i.e., doing `foo(params[0], 11., 1.)` would not hit the bug)

The output will be:
```txt
Value
44

Clad diff:
0

Num diff:
11
```

My compiler setup:
```txt
clang version 16.0.6
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /bin
```

PetroZarytskyi:
The problem this PR attempts to fix is described in #659 (which in turn caused #681). Also, this PR introduces placeholders to ``DelayedGlobalStoreAndRef`` to rewrite expressions after they are added to the current block if there is no point in storing them. This is done to avoid cloning while handling multiplication differentiation.

Fixes #659. Fixes #681.

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -41,6 +41,11 @@ namespace clad {
     /// the reverse mode we also accumulate Stmts for the reverse pass which
     /// will be executed on return.
     std::vector<Stmts> m_Reverse;
+    /// Accumulates local variables for all visited blocks.
+    std::vector<std::map<clang::VarDecl*, clang::VarDecl*>> m_Locals;
+    /// Stores all expressions used as placeholders which have to be
+    /// reset later.
+    std::set<const clang::Expr*> m_Placeholders;
     /// Stack is used to pass the arguments (dfdx) to further nodes
     /// in the Visit method.
     std::stack<clang::Expr*> m_Stack;
@@ -51,7 +56,7 @@ namespace clad {
     //// A reference to the output parameter of the gradient function.
     clang::Expr* m_Result;
     /// Based on To-Be-Recorded analysis performed before differentiation,
-    /// tells UsefulToStoreGlobal whether a variable with a given
+    /// tells isToBeRecorded whether a variable with a given
     /// SourceLocation has to be stored before being changed or not.
     std::set<clang::SourceLocation> m_ToBeRecorded;
     /// A flag indicating if the Stmt we are currently visiting is inside loop.
@@ -142,15 +147,18 @@ namespace clad {
     }
     /// Create new block.
     Stmts& beginBlock(direction d = direction::forward) {
-      if (d == direction::forward)
+      if (d == direction::forward) {
         m_Blocks.emplace_back();
-      else
+        m_Locals.emplace_back();
+      } else {
         m_Reverse.emplace_back();
+      }
       return getCurrentBlock(d);
     }
     /// Remove the block from the stack, wrap it in CompoundStmt and return it.
     clang::CompoundStmt* endBlock(direction d = direction::forward) {
       if (d == direction::forward) {
+        EmitReverseSweepDeclarations();
         auto* CS = MakeCompoundStmt(getCurrentBlock(direction::forward));
         m_Blocks.pop_back();
         return CS;
@@ -164,12 +172,31 @@ namespace clad {
 
     Stmts EndBlockWithoutCreatingCS(direction d = direction::forward) {
       auto blk = getCurrentBlock(d);
-      if (d == direction::forward)
+      if (d == direction::forward) {
         m_Blocks.pop_back();
-      else
+        EmitReverseSweepDeclarations();
+      } else
         m_Reverse.pop_back();
       return blk;
     }
+
+    clang::Expr* Clone(const clang::Expr* E) {
+      // Placeholders should not be cloned since otherwise we will not be able
+      // to find and replace them later.
+      if (m_Placeholders.find(E) != m_Placeholders.end())
+        return const_cast<clang::Expr*>(E);
+      return VisitorBase::Clone(E);
+    }
+
+    clang::Stmt* Clone(const clang::Stmt* E) { return VisitorBase::Clone(E); }
+
+    /// Emits declarations in the reverse sweep of variables that were declared
+    /// locally inside some blocks in the forward sweep. Needs to be called
+    /// after all of the usages of the variable in the reverse sweep since
+    /// the declaration should come before the usages when the statement order
+    /// is reversed.
+    void EmitReverseSweepDeclarations(bool endCurBlock = true);
+
     /// Output a statement to the current block. If Stmt is null or is an unused
     /// expression, it is not output and false is returned.
     bool addToCurrentBlock(clang::Stmt* S, direction d = direction::forward) {
@@ -222,6 +249,10 @@ namespace clad {
     /// This is the central point for checkpointing.
     bool ShouldRecompute(const clang::Expr* E);
 
+    /// For a given location, tells if the corresponding expression should
+    /// be stored in the forward pass and then restored in the reverse pass.
+    bool isToBeRecorded(clang::SourceLocation loc);
+
     /// Builds a variable declaration and stores it in the function
     /// global scope.
     ///
@@ -253,8 +284,11 @@ namespace clad {
                                bool force = false);
     StmtDiff BuildPushPop(clang::Expr* E, clang::QualType Type,
                           llvm::StringRef prefix = "_t", bool force = false);
-    StmtDiff StoreAndRestore(clang::Expr* E, llvm::StringRef prefix = "_t",
-                             bool force = false);
+    StmtDiff StoreAndRestore(StmtDiff SD, llvm::StringRef prefix = "_t",
+                             clang::SourceLocation loc = {});
+    StmtDiff StoreAndRestore(StmtDiff SD, clang::QualType Type,
+                             llvm::StringRef prefix = "_t",
+                             clang::SourceLocation loc = {});
 
     //// A type returned by DelayedGlobalStoreAndRef
     /// .Result is a reference to the created (yet uninitialized) global
@@ -266,15 +300,15 @@ namespace clad {
     struct DelayedStoreResult {
       ReverseModeVisitor& V;
       StmtDiff Result;
-      bool isConstant;
       bool isInsideLoop;
       bool needsUpdate;
+      clang::Expr* Placeholder;
       DelayedStoreResult(ReverseModeVisitor& pV, StmtDiff pResult,
-                         bool pIsConstant, bool pIsInsideLoop,
-                         bool pNeedsUpdate = false)
-          : V(pV), Result(pResult), isConstant(pIsConstant),
-            isInsideLoop(pIsInsideLoop), needsUpdate(pNeedsUpdate) {}
-      void Finalize(clang::Expr* New);
+                         bool pIsInsideLoop, bool pNeedsUpdate = false,
+                         clang::Expr* pPlaceholder = nullptr)
+          : V(pV), Result(pResult), isInsideLoop(pIsInsideLoop),
+            needsUpdate(pNeedsUpdate), Placeholder(pPlaceholder) {}
+      void Finalize(StmtDiff New);
     };
 
     /// Sometimes (e.g. when visiting multiplication/division operator), we
@@ -286,7 +320,8 @@ namespace clad {
     /// This is what DelayedGlobalStoreAndRef does. E is expected to be the
     /// original (uncloned) expression.
     DelayedStoreResult DelayedGlobalStoreAndRef(clang::Expr* E,
-                                                llvm::StringRef prefix = "_t");
+                                                llvm::StringRef prefix = "_t",
+                                                bool forceNoRecompute = false);
 
     struct CladTapeResult {
       ReverseModeVisitor& V;
@@ -311,6 +346,9 @@ namespace clad {
     /// \returns A struct containg necessary call expressions for the built
     /// tape
     CladTapeResult MakeCladTapeFor(clang::Expr* E,
+                                   llvm::StringRef prefix = "_t");
+
+    CladTapeResult MakeCladTapeFor(clang::Expr* E, clang::QualType type,
                                    llvm::StringRef prefix = "_t");
 
   public:
@@ -468,20 +506,18 @@ namespace clad {
     /// Helper function to differentiate a loop body.
     ///
     ///\param[in] body body of the loop
+    ///\param[in] cond condition of the loop
     ///\param[in] loopCounter associated `LoopCounter` object of the loop.
-    ///\param[in] condVarDiff derived statements of the condition
-    /// variable, if any.
     ///\param[in] forLoopIncDiff derived statements of the `for` loop
     /// increment statement, if any.
     ///\param[in] isForLoop should be true if we are differentiating a `for`
     /// loop body; otherwise false.
-    ///\returns {forward pass statements, reverse pass statements} for the loop
-    /// body.
-    StmtDiff DifferentiateLoopBody(const clang::Stmt* body,
-                                   LoopCounter& loopCounter,
-                                   clang::Stmt* condVarDifff = nullptr,
-                                   clang::Stmt* forLoopIncDiff = nullptr,
-                                   bool isForLoop = false);
+    ///\returns {differentiated loop body, differentiated loop condition}
+    std::pair<StmtDiff, StmtDiff>
+    DifferentiateLoopBody(const clang::Stmt* body, const clang::Stmt* cond,
+                          LoopCounter& loopCounter,
+                          clang::Stmt* forLoopIncDiff = nullptr,
+                          bool isForLoop = false);
 
     /// This class modifies forward and reverse blocks of the loop
     /// body so that `break` and `continue` statements are correctly
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -119,11 +119,11 @@ namespace clad {
     CompoundStmt* AppendAndCreateCompoundStmt(ASTContext& C, Stmt* initial,
                                               Stmt* S) {
       llvm::SmallVector<Stmt*, 16> block;
-      assert(isa<CompoundStmt>(initial) &&
-             "initial should be of type `clang::CompoundStmt`");
       CompoundStmt* CS = dyn_cast<CompoundStmt>(initial);
       if (CS)
         block.append(CS->body_begin(), CS->body_end());
+      else
+        block.push_back(initial);
       block.push_back(S);
       auto stmtsRef = clad_compat::makeArrayRef(block.begin(), block.end());
       return clad_compat::CompoundStmt_Create(C, stmtsRef /**/ CLAD_COMPAT_CLANG15_CompoundStmt_Create_ExtraParam1(CS), noLoc, noLoc);
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -62,12 +62,53 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     return Call;
   }
 
+  void ReverseModeVisitor::EmitReverseSweepDeclarations(bool endCurBlock) {
+    for (auto iter = m_Locals.back().rbegin(), e = m_Locals.back().rend();
+         iter != e; ++iter) {
+      clang::VarDecl* VDForw = iter->first;
+      clang::VarDecl* VDRev = iter->second;
+
+      clang::QualType type = VDRev->getType();
+      bool isRefType = type->isReferenceType();
+      clang::Expr* ref = BuildDeclRef(VDForw);
+      // For reference-type variables store the pointer, not the value.
+      if (isRefType) {
+        ref = BuildOp(UO_AddrOf, ref);
+        type = ref->getType();
+      }
+      StmtDiff pushPop = StoreAndRestore(ref, type);
+      addToCurrentBlock(pushPop.getStmt(), direction::forward);
+      clang::Expr* init = nullptr;
+      if (auto* BO =
+              llvm::dyn_cast<clang::BinaryOperator>(pushPop.getStmt_dx()))
+        init = BO->getRHS();
+      else
+        init = llvm::cast<clang::Expr>(pushPop.getStmt_dx());
+      if (isRefType)
+        init = BuildOp(UO_Deref, init);
+      VDRev->setInit(init);
+
+      clang::Stmt* declStmt = BuildDeclStmt(VDRev);
+      addToCurrentBlock(declStmt, direction::reverse);
+    }
+    if (endCurBlock)
+      m_Locals.pop_back();
+    else
+      m_Locals.back().clear();
+  }
+
   ReverseModeVisitor::CladTapeResult
   ReverseModeVisitor::MakeCladTapeFor(Expr* E, llvm::StringRef prefix) {
+    return MakeCladTapeFor(E, E->getType(), prefix);
+  }
+
+  ReverseModeVisitor::CladTapeResult
+  ReverseModeVisitor::MakeCladTapeFor(Expr* E, QualType type,
+                                      llvm::StringRef prefix) {
     assert(E && "must be provided");
     E = E->IgnoreImplicit();
     QualType TapeType =
-        GetCladTapeOfType(getNonConstType(E->getType(), m_Context, m_Sema));
+        GetCladTapeOfType(getNonConstType(type, m_Context, m_Sema));
     LookupResult& Push = GetCladTapePush();
     LookupResult& Pop = GetCladTapePop();
     Expr* TapeRef =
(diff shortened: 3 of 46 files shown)


parth-07:
@PetroZarytskyi Can you please describe the solution in the commit message / pull request description?

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -164,12 +174,28 @@ namespace clad {
 
     Stmts EndBlockWithoutCreatingCS(direction d = direction::forward) {
       auto blk = getCurrentBlock(d);
-      if (d == direction::forward)
+      if (d == direction::forward) {
         m_Blocks.pop_back();
-      else
+        EmitRevSweepDecls();
+        m_Locals.pop_back();
```
vgvassilev, line 180: We should move the pop operation in `EmitRevSweepDecls`.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -164,12 +174,28 @@ namespace clad {
...
+      if (m_Placeholders.find(E)!=m_Placeholders.end())
+        return const_cast<clang::Expr*>(E);
+      return VisitorBase::Clone(E);
+    }
+
+    clang::Stmt* Clone(const clang::Stmt* E) {
+      return VisitorBase::Clone(E);
+    }
+
+    /// Add more comments.
```
vgvassilev, line 196: We should address this comment.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -755,7 +784,16 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         m_ExternalSource->ActBeforeDifferentiatingStmtInVisitCompoundStmt();
       StmtDiff SDiff = DifferentiateSingleStmt(S);
       addToCurrentBlock(SDiff.getStmt(), direction::forward);
-      addToCurrentBlock(SDiff.getStmt_dx(), direction::reverse);
+
+      /// FIXME: Explain this.
```
vgvassilev, line 788: Maybe we should address that fixme.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1026,28 +1064,25 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
       m_ExternalSource->ActBeforeDifferentiatingLoopInitStmt();
-    StmtDiff initResult = init ? DifferentiateSingleStmt(init) : StmtDiff{};
+    StmtDiff initResult = nullptr;
+    if (init) {
+      initResult = DifferentiateSingleStmt(init);
+      if (auto initDeclStmt = dyn_cast<DeclStmt>(initResult.getStmt())) {
+        auto* VD = cast<VarDecl>(initDeclStmt->getSingleDecl());
+        Expr* initValue = VD->getInit();
+        initResult.updateStmt(BuildOp(BO_Assign, BuildDeclRef(VD), initValue));
+        VD->setInit(nullptr);
```
vgvassilev, line 1074: Why do we remove the initializer here?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2500,6 +2522,8 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       return BuildOp(opCode, LExpr, RExpr);
     }
     Expr* op = BuildOp(opCode, Ldiff.getExpr(), Rdiff.getExpr());
+    if (!isPointerOp && valueForRevPass==nullptr)
```
vgvassilev, line 2525:
```suggestion
    if (!isPointerOp && !valueForRevPass)
```

vgvassilev (commented):
Here is some partial review. Could you go over your FIXME notes and try to address them as it seems most of them need to be addressed within this PR.

parth-07 (changes requested):

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1026,28 +1064,25 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     const Stmt* init = FS->getInit();
     if (m_ExternalSource)
       m_ExternalSource->ActBeforeDifferentiatingLoopInitStmt();
-    StmtDiff initResult = init ? DifferentiateSingleStmt(init) : StmtDiff{};
+    StmtDiff initResult = nullptr;
```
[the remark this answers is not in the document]
PetroZarytskyi: I agree. Probably did this by mistake.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1026,28 +1064,25 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+        initResult.updateStmt(BuildOp(BO_Assign, BuildDeclRef(VD), initValue));
+        VD->setInit(nullptr);
+        addToCurrentBlock(initDeclStmt);
+      }
+    }
+
 
     // Save the isInsideLoop value (we may be inside another loop).
-    llvm::SaveAndRestore<bool> SaveIsInsideLoop(isInsideLoop);
+    bool old_isInsideLoop = isInsideLoop;
```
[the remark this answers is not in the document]
PetroZarytskyi: Because ``EmitRevSweepDecls`` decides on whether to create a tape based on ``isInsideLoop``. We want the original ``isInsideLoop`` value when emitting reverse sweep decls from the loop init (which happens when ending the forward block).

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1026,28 +1064,25 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
-    }
-
-    // FIXME: for now we assume that cond has no differentiable effects,
-    // but it is not generally true, e.g. for (...; (x = y); ...)...
-    StmtDiff cond;
-    if (FS->getCond())
-      cond = Visit(FS->getCond());
-    const auto* IDRE = dyn_cast<DeclRefExpr>(FS->getInc());
+    bool IDRE = isa<DeclRefExpr>(FS->getInc());
+    /// FIXME: Why do we need to differentiate DeclRefExpr twice?
```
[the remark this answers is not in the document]
PetroZarytskyi: From what I get, ``inc`` is supposed to be the original increment (it's always differentiated later). Why is it differentiated specifically when it is a DeclRefExpr?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1258,18 +1295,17 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     llvm::SmallVector<Expr*, 4> reverseIndices(Indices.size());
     llvm::SmallVector<Expr*, 4> forwSweepDerivativeIndices(Indices.size());
     for (std::size_t i = 0; i < Indices.size(); i++) {
-      // FIXME: Remove redundant indices vectors.
       StmtDiff IdxDiff = Visit(Indices[i]);
-      clonedIndices[i] = Clone(IdxDiff.getExpr());
```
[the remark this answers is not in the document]
PetroZarytskyi: Well, it's necessary to replace ``getExpr`` with ``getRevSweepAsExpr`` for ``reverseIndices``. I could move ``Clone`` to where they were previously if that matters.

test/Gradient/Loops.C
```diff
@@ -21,23 +21,32 @@ double f1(double x) {
...
 //CHECK-NEXT:       clad::tape<double> _t1 = {};
+//CHECK-NEXT:       int _t2;
 //CHECK-NEXT:       double t = 1;
 //CHECK-NEXT:       _t0 = 0;
-//CHECK-NEXT:       for (int i = 0; i < 3; i++) {
-//CHECK-NEXT:           _t0++;
-//CHECK-NEXT:           clad::push(_t1, t);
-//CHECK-NEXT:           t *= x;
+//CHECK-NEXT:       {
+//CHECK-NEXT:           int i;
```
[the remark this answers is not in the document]
PetroZarytskyi: The issue here is that we want to store the last value of the loop init. So we need to access ``i`` once right after the last condition. But we can't do that outside the loop since the variable is not declared there. My solution doesn't too look nice though so if you have a better idea feel free to share it.

include/clad/Differentiator/ReverseModeVisitor.h
PetroZarytskyi: Sometimes we want to emit reverse sweep decls before ending the block (when handling return/break/continue statements). There is an option to add an argument to ``EmitRevSweepDecls`` to decide whether it should do the pop operation.

lib/Differentiator/CladUtils.cpp
```diff
@@ -119,11 +119,11 @@ namespace clad {
     CompoundStmt* AppendAndCreateCompoundStmt(ASTContext& C, Stmt* initial,
                                               Stmt* S) {
       llvm::SmallVector<Stmt*, 16> block;
-      assert(isa<CompoundStmt>(initial) &&
-             "initial should be of type `clang::CompoundStmt`");
```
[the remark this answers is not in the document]
PetroZarytskyi: I changed it to be consistent with ``PrependAndCreateCompoundStmt``. The difference is that when we pass an individual stmt as ``initial``, we wrap it in a compound stmt with ``S`` instead of considering this an error. I needed this at some point but not anymore. I can revert this but I think this change makes sense at least for consistency.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: Because I replaced the declaration inside the loop init with an assignment to move the declaration outside. Parth asked about why this is done. Alternatively, I could move the whole declaration outside and leave the loop init empty.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: The only reason I used ``==nullptr`` was because I think clang-tidy argues about implicit conversion from pointers to bool.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -142,15 +147,20 @@ namespace clad {
...
+      } else {
         m_Reverse.emplace_back();
+      }
       return getCurrentBlock(d);
     }
     /// Remove the block from the stack, wrap it in CompoundStmt and return it.
     clang::CompoundStmt* endBlock(direction d = direction::forward) {
       if (d == direction::forward) {
+        EmitRevSweepDecls();
+        m_Locals.pop_back();
```
vgvassilev, line 162: I had a comment that probably went away to move the pop_back operation in the call to Emit.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -164,12 +174,28 @@
...
+      } else
         m_Reverse.pop_back();
       return blk;
     }
+
+    clang::Expr* Clone(const clang::Expr* E) {
+      if (m_Placeholders.find(E)!=m_Placeholders.end())
+        return const_cast<clang::Expr*>(E);
```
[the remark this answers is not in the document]
PetroZarytskyi: Because to track that as a placeholder, we want the original expression. If it's cloned, we will not be able to replace it and it will remain in the AST.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1127,11 +1166,12 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
                                             noLoc,
                                             noLoc,
                                             noLoc);
+    isInsideLoop = old_isInsideLoop;
```
vgvassilev, line 1169:
```suggestion
    isInsideLoop = SaveIsInsideLoop.get();
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1598,22 +1648,22 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
           // consistent with `StoreAndRef`. This way we will not need to handle
           // inside loop and outside loop cases separately.
           Expr* newArgE = Visit(arg).getExpr();
-          argDiffStore = {newArgE, argDiffLocalE};
+          argDiffStore = {newArgE, nullptr, nullptr, argDiffLocalE};
```
vgvassilev, line 1651: Maybe add the argument kinds as comments.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2056,7 +2106,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         diff_dx = BuildArrayRefPtrRefExpr(diff_dx);
       if (isPointerOp)
         addToCurrentBlock(BuildOp(opCode, diff_dx), direction::forward);
-      if (UsefulToStoreGlobal(diff.getRevSweepAsExpr())) {
+      if (isPointerOp || isToBeRecorded(E->getBeginLoc())) {
```
vgvassilev, line 2109: We should sink `isPointerOp` in `isToBeRecorded`.

Add a fixme note pointing out that we can remove this once we support pointers in TBR.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2994,92 +2993,135 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
-      if (E->isModifiableLvalue(m_Context) == Expr::MLV_Valid) {
-        auto* r = Clone(E);
-        Restore = BuildOp(BO_Assign, r, Ref);
+    Expr* Store = BuildOp(BO_Assign, Ref, SD.getExpr());
+    Expr* Restore = nullptr;
+
+    if (SD.getExpr()->isModifiableLvalue(m_Context) == Expr::MLV_Valid)
+      Restore = BuildOp(BO_Assign, SD.getRevSweepAsExpr(), Ref);
+    else
+      Restore = Ref;
```
vgvassilev, line 3042: Can we add a test here?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2994,92 +2993,135 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
-    return {};
-  }
+    };
+
+    if (!needsUpdate)
+      return;
 
-  void ReverseModeVisitor::DelayedStoreResult::Finalize(Expr* New) {
-    if (isConstant || !needsUpdate)
+    if (Placeholder!=nullptr) {
```
vgvassilev, line 3069:
```suggestion
    if (!Placeholder) {
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3246,27 +3280,65 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+          m_ExternalSource->ActBeforeDifferentiatingLoopCondition();
+        condDiff = DifferentiateSingleStmt(cond, /*dfdS=*/nullptr);
+      }
+      addToCurrentBlock(condDiff.getStmt_dx(), direction::reverse);
+      bodyDiff = Visit(body);
+      if (auto* bodyCS = dyn_cast<CompoundStmt>(bodyDiff.getStmt()))
+        for (Stmt* S : bodyCS->body())
+          addToCurrentBlock(S, direction::forward);
+      else
+        addToCurrentBlock(bodyDiff.getStmt(), direction::forward);
```
vgvassilev, line 3299: Can we add a test here?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3246,27 +3280,65 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+      if (bodyDiff.getStmt_dx()) {
+        if (auto* bodyDxCS = dyn_cast<CompoundStmt>(bodyDiff.getStmt_dx()))
+          for (auto iter = bodyDxCS->body_rbegin(), e = bodyDxCS->body_rend();
+               iter != e; ++iter)
+            addToCurrentBlock(*iter, direction::reverse);
+        else
+          addToCurrentBlock(bodyDiff.getStmt_dx(), direction::reverse);
```
vgvassilev, line 3306: And here?

include/clad/Differentiator/ReverseModeVisitor.h
vgvassilev: Yes, let's add an argument whose default value is to pop.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2705,7 +2737,28 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       }
     }
     m_Variables.emplace(VDClone, derivedVDE);
-
+    // If the declaration is not located in the function global scope,
+    // we have to redeclare it in the reverse sweep since it won't be
+    // visible otherwise.
+    // FIXME: Is there a better way to determine non-function-global scope
```
parth-07, line 2743: Yes, there must be a better way for this. Have you tried `VD->getDeclContext() != m_Function`?
Ideally, function global scope variables should have the primal `FunctionDecl` as their `DeclContext`

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2705,7 +2737,28 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    if (m_Blocks.size()>2) {
+      clang::QualType type = CloneType(VDClone->getType());
+      if (type->isArrayType()) {
+        type = m_Context.getArrayDecayedType(type);
+      }
+      clang::Expr* init = nullptr;
+      /// Creating default initialized reference type or const variable
+      /// declarations produces an error.
+      if (type->isReferenceType() || type.isConstQualified())
+        init = Clone(VD->getInit());
```
parth-07, line 2754: This seems incorrect. What if the expression returned by `VD-getInit()` cannot be used in the reverse sweep? For example:

```cpp
{ 
  double u = 7;
  const double v = u;
}
```
If I understand things correctly, when redeclaring `const double v` in the reverse sweep, `init` is pointing towards `double u` of the forward sweep instead of the reverse sweep.

parth-07 (changes requested):

lib/Differentiator/ReverseModeVisitor.cpp

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: I'm not sure how ``DeclContext`` works exactly but ``getDeclContext`` seems to return ``m_Function`` for variables declared inside loops etc.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: The reason for that init is mainly that creating a const / reference-type declaration without an init produces a warning (even if the init is set later).

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: I added such comments here and in two more places but maybe we'll need more of them.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: I didn't manage to come up with such a test. I think the check is excessive. I removed it.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: I didn't manage to come up with such a test. I think the check is excessive. I removed it.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: Same thing here.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3295,19 +3295,12 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
-        addToCurrentBlock(bodyDiff.getStmt(), direction::forward);
-      if (bodyDiff.getStmt_dx()) {
-        if (auto* bodyDxCS = dyn_cast<CompoundStmt>(bodyDiff.getStmt_dx()))
-          for (auto iter = bodyDxCS->body_rbegin(), e = bodyDxCS->body_rend();
-               iter != e; ++iter)
-            addToCurrentBlock(*iter, direction::reverse);
-        else
-          addToCurrentBlock(bodyDiff.getStmt_dx(), direction::reverse);
-      }
+      for (Stmt* S : cast<CompoundStmt>(bodyDiff.getStmt())->body())
```
vgvassilev, line 3298: These changes will break this valid c++ code:
```cpp
int f(int x) 
  try { return x; }
  catch(...) { return 0; }
```

vgvassilev: Maybe we can add this as a test case.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1026,28 +1071,22 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     const Stmt* init = FS->getInit();
     if (m_ExternalSource)
       m_ExternalSource->ActBeforeDifferentiatingLoopInitStmt();
-    StmtDiff initResult = init ? DifferentiateSingleStmt(init) : StmtDiff{};
+    StmtDiff initResult;
+    if (init) {
+      initResult = DifferentiateSingleStmt(init);
+      if (auto* initDeclStmt = dyn_cast<DeclStmt>(initResult.getStmt())) {
```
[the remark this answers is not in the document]
parth-07: Oh, okay. Got it. Thanks for the explanation.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: Oh, okay. Can you please document in the code that this `init` is only a placeholder and the actual value for `init` is updated later?

parth-07:
The produced gradient of the below code has compile-time error in the local variable redeclaration part.

```cpp
    double r;
    {
        double res;
        double v_copy = v;
        double &v_ref = v_copy;
        res = v_ref;
        r = res;
    }
    return r;
}
```

The local variable redeclaration in the reverse sweep is as follows:

```cpp
double &v_ref = v_copy;
double v_copy = _t3;
double res = _t2;
```

`v_ref` should be declared after `v_copy`.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: I changed this bit so the discussion is probably not relevant anymore.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: Earlier in the code, we check if the body is a compound statement. What we could ask is whether it's possible to have the forward sweep of the result not being a compound statement (when the original body is a compound statement). From what I get, this is not possible.

include/clad/Differentiator/ReverseModeVisitor.h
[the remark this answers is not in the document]
vgvassilev: Can we add a comment here explaining that?

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -164,12 +172,31 @@ namespace clad {
 
     Stmts EndBlockWithoutCreatingCS(direction d = direction::forward) {
       auto blk = getCurrentBlock(d);
-      if (d == direction::forward)
+      if (d == direction::forward) {
         m_Blocks.pop_back();
-      else
+        EmitReverseSweepDeclarations();
```
vgvassilev, line 177: Can we add a test case covering this branch?

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: Done.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: Done.

include/clad/Differentiator/ReverseModeVisitor.h
vgvassilev: Is that dead code then?

vgvassilev:
@PetroZarytskyi, is that PR superseded by #738? Can we close it?

PetroZarytskyi:
> @PetroZarytskyi, is that PR superseded by #738? Can we close it?

I think the solution implemented in #738 is better. We should probably close this PR.

PR #690, opened 2023-12-26
Simplify derivative statements in VisitBinaryOperator

PetroZarytskyi:
Every time we visit multiplication or division in the reverse mode, we store expressions for the derivatives of the LHS and the RHS.
```
dr = StoreAndRef(dr, direction::reverse); 
dl = StoreAndRef(dl, direction::reverse); 
```
This makes the derivative code less readable and often produces unused variables. For instance, this code
```
x = 5 * y;
```
will be differentiated as
```
x = 5 * y;
...
double _r_d0 = _d_x;
double _r0 = _r_d0 * y;
double _r1 = 5 * _r_d0;
_d_y += _r1;
_d_x -= _r_d0;
```
while could be just
```
x = 5 * y;
...
double _r_d0 = _d_x;
_d_y += 5 * _r_d0;
_d_x -= _r_d0;
```

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2164,7 +2164,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       StmtDiff RResult;
       // If R has no side effects, it can be just cloned
       // (no need to store it).
-      if (utils::ContainsFunctionCalls(R) || R->HasSideEffects(m_Context)) {
+      if (!ShouldRecompute(R)) {
         RDelayed = std::unique_ptr<DelayedStoreResult>(
             new DelayedStoreResult(DelayedGlobalStoreAndRef(R)));
         RResult = RDelayed->Result;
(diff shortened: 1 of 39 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2173,42 +2173,41 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+      Expr* RStored = StoreAndRef(RResult.getRevSweepAsExpr(), direction::reverse);
       Expr* dl = nullptr;
-      if (dfdx()) {
+      if (dfdx())
         dl = BuildOp(BO_Div, dfdx(), RStored);
-        dl = StoreAndRef(dl, direction::reverse);
-      }
       Ldiff = Visit(L, dl);
+      StmtDiff LStored = Ldiff;
+      if (utils::ContainsFunctionCalls(LStored.getExpr()) || LStored.getExpr()->HasSideEffects(m_Context))
```
vgvassilev, line 2209: Can we move these checks into UsefulToStore? I believe there is even a fixme there.

test/Gradient/Loops.C
```diff
@@ -337,7 +323,7 @@ double f_log_gaus(double* x, double* p /*means*/, double n, double sigma) {
   double gaus = 1./std::sqrt(std::pow(2*M_PI, n) * sigma) * std::exp(power);
   return std::log(gaus);
 }
-//CHECK-NEXT: void f_log_gaus_grad_1(double *x, double *p, double n, double sigma, clad::array_ref<double> _d_p) {
+//CHECK: void f_log_gaus_grad_1(double *x, double *p, double n, double sigma, clad::array_ref<double> _d_p) {
```
vgvassilev, line 326: Why we had to replace the check next clause?

PetroZarytskyi: @vgvassilev I just thought CHECK made more sense since there was a new function check starting there.

vgvassilev (approved):
LGTM! Great work!

Issue #693, opened 2023-12-30
`multiple definition of trap(int)` link error occurs when multiple files contain `clad/Differentiator/Differentiator.h` header file


parth-07:
Users have to include `clad/Differentiator/Differentiator.h` header file to run clad generated code because this header provides clad tape and clad array interface. It's likely that different files in a project may be using clad generated codes and thus will require including `Differentiator.h` header. However, linking two files containing `Differentiator.h` gives `multiple definition of trap(int)` error. This behaviour is unexpected and incorrect. A header file should not contain definitions that may lead to ODR violations. 

Reproducible example:

**reproduce.sh**
```
#!/usr/bin/env

cat >foo.c <<\EOF
#include "clad/Differentiator/Differentiator.h"

int foo() { return 1; }
EOF

cat >main.cpp <<\EOF
#include "clad/Differentiator/Differentiator.h"

int foo();

int main() { return foo(); }
EOF

/usr/lib/llvm-15/bin/clang++ -g -Xclang -add-plugin -Xclang clad \
    -Xclang -load -Xclang \
    /path/to/clad.so \
    -I/path/to/clad/include \
    main.cpp -xc++ -o main foo.cpp
```

```
bash -x ./reproduce.sh
```

maxaehle:
When running your code with Clad at 1bdd261ea71163b3e8074dca4018e94a481a1c50 and Clang-11, I get the same "multiple definition" error as you and a few more errors, all related to symbols from the namespace `numerical_diff`; e.g., a function `printError(double, double, unsigned int, int)` and a variable `bufferManager`. 

I'm not sure whether you already have a plan on how to fix this, but I have encountered similar problems in the past (in other projects) so I'd like to share a couple of ideas how to fix it here.

For function definitions, I'm aware of the following ways to fix this:

**1. inline**: Declare the function as  `inline`. For C++ sources including the header, 
> There may be more than one definition of an inline function [...] in the program as long as each definition appears in a different translation unit and (for non-static inline functions [...]) all definitions are identical. For example, an inline function [...] may be defined in a header file that is included in multiple source files. 

according to [cppreference.com](https://en.cppreference.com/w/cpp/language/inline). For C sources including the header,

> The inline definition that does not use extern is not externally visible and does not prevent other translation units from defining the same function. This makes the inline keyword an alternative to static for defining functions inside header files, which may be included in multiple translation units of the same program. 

according to [cppreference.com](https://en.cppreference.com/w/c/language/inline).

[...]

Issue #314, opened 2021-10-17
Redefinition of function error on importing Differentiator.h in multiple source files


parth-07:
Currently, we have defined several functions in the `Differentiator.h` header file (and also in header files that `Differentiator.h` includes). This causes redefinition errors if more than one source files includes the `Differentiator.h` header. Therefore, Ideally, we should not define functions in the header file.

Consider this example, 

a.cpp
```c++
#include <iostream>
#include "b.h"
#include "clad/Differentiator/Differentiator.h"
double fn(double i, double j) {
  return i*i*j;
}

int main() {
  auto d_fn = clad::differentiate(fn, "i");
  std::cout<<d_fn.execute(3, 5)<<"\n";
}
```

b.h
```c++
double another_function(double a, double b);
```

b.cpp
```c++
#include "clad/Differentiator/Differentiator.h"
double another_function(double i, double j) {
  return i*j;
}
```
Now compiling these files with clad enabled using a command such as, 
```
/usr/bin/clang++-10 -g -Xclang -add-plugin -Xclang clad -Xclang -load -Xclang /mnt/R/clubs/gsoc/clad/fork/builds/build-10/inst/lib/clad.so -I/mnt/R/clubs/gsoc/clad/fork/builds/build-10/inst/include -x c++ -std=c++11 a.cpp b.cpp -o  build/a
```
<details>

<summary>
gives following error:
</summary>

```
/usr/bin/ld: /tmp/b-972fc8.o: in function `_ZN4clad9tape_implIPvE7destroyIPS1_EENSt9enable_ifIXsr3std25is_trivially_destructibleIDTdeclsr3stdE7declvalIT_EEEEE5valueEvE4typeES6_S6_':
/mnt/R/clubs/gsoc/clad/fork/builds/build-10/inst/include/clad/Differentiator/CladConfig.h:27: multiple definition of `trap(int)'; /tmp/a-6c5cde.o:/mnt/R/clubs/gsoc/clad/fork/builds/build-10/inst/include/clad/Differentiator/CladConfig.h:27: first defined here
/usr/bin/ld: /tmp/b-972fc8.o: in function `numerical_diff::printError(double, double, unsigned int, int)':
[...]

PR #694, opened 2024-01-02
Fix strong symbol definitions in Differentiator.h. Add unittests. 

vgvassilev:
Fixes: https://github.com/vgvassilev/clad/issues/693, Fixes https://github.com/vgvassilev/clad/issues/314.


unittests/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
+add_custom_target(CladUnitTests)
+set_target_properties(CladUnitTests PROPERTIES FOLDER "Clang tests")
```
alexander-penev, line 2: Clang or Clad?

alexander-penev (approved):

unittests/CMakeLists.txt
vgvassilev: Good point!

PR #702, opened 2024-01-05
[cmake] Use the compiler clad is built for if compiling benchmarks and tests

vgvassilev:
We need to attach the clad plugin to the compiler invocations in order to either benchmark or test clad's infrastructure. The best way I could find is by setting/resetting the CMAKE_CXX_COMPILER variable which is a hack that works. Note that we do not have a cmake guarantee that will work but since that's in the core of our testing infrastructure we will find out quickly if, when and where it breaks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -295,8 +295,20 @@ if (CLAD_INCLUDE_DOCS)
   add_subdirectory(docs)
 endif()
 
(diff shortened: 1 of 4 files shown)


vgvassilev:
The benchmark failure is because the master is broken and this PR intends to fix it.

PR #707, opened 2024-01-11
Add test for increment operator in return statement

### LINKED ISSUE #687: Wrong outputs in reverse mode with unary operators in return statements
Minimal reproducible example:
```cpp
double f1(double x) {
  double a = x;
  return (++a)*x; // essentially (x+1)*x
}
```

The generated code is:
```cpp
void f1_grad(double x, clad::array_ref<double> _d_x) {
    double _d_a = 0;
    double a = x;
    goto _label0;
  _label0:
    {
        double _r0 = 1 * x;
        _d_a += _r0;
        --a;
        double _r1 = (a + 1) * 1;
        * _d_x += _r1;
    }
    * _d_x += _d_a;
}
```

Though the generated code correctly contains `--a` in the reverse pass, there is no `++a` in the forward pass, thus resulting in the final derivative being less than the expected correct output. 
For example, `x=3` results in the output of 6, but the correct output should be 7.

Similar is the case with post-increment operator (a++).

vaithak:
Closes #687

### DIFF
--- a/test/Gradient/Gradients.C
+++ b/test/Gradient/Gradients.C
@@ -740,6 +740,27 @@ double fn_global_var_use(double i, double j) {
 // CHECK-NEXT:     }
 // CHECK-NEXT: }
 
+double fn_increment_in_return(double i, double j) {
+  double temp = i;
+  return (++i) * temp; // (i+1)*i
+}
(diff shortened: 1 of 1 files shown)


vaithak:
#687 was fixed in #690, adding the test here to ensure it doesn't arise in future. If this is not required, we can anyway close #687.

vgvassilev (approved):
LGTM!

PR #709, opened 2024-01-15
Fix support for non-type template params

### LINKED ISSUE #706: Failed to differentiate conditional operator depending on a non-template arg
```diff
diff --git a/test/FirstDerivative/TemplateFunction.C b/test/FirstDerivative/TemplateFunction.C
index f808d5d..d9186ab 100644
--- a/test/FirstDerivative/TemplateFunction.C
+++ b/test/FirstDerivative/TemplateFunction.C
@@ -18,6 +18,13 @@ T multiplication(T x) {
   return x * x;
 }
 
+template<size_t __bits>
+double generate_canonical(double x) {
+  const size_t _Dt = 53; // result of std::numeric_limits<double>::digits
+  const size_t __b = _Dt < __bits ? _Dt : __bits;
+  return x;
+}
+
 int main () {
   int x;
 
@@ -75,5 +82,7 @@ int main () {
   // CHECK-NEXT: return _d_x * x + x * _d_x;
   // CHECK-NEXT: }
 
+  clad::differentiate(generate_canonical<15>, 0);
+
   return 0;
 }
```

This is related to #700.

vaithak:
fixes #706

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -739,8 +739,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
   }
   StmtDiff ReverseModeVisitor::VisitStmt(const Stmt* S) {
     diag(
-        DiagnosticsEngine::Warning,
-        S->getBeginLoc(),
+        DiagnosticsEngine::Warning, S->getBeginLoc(),
         "attempted to differentiate unsupported statement, no changes applied");
     // Unknown stmt, just clone it.
     return StmtDiff(Clone(S));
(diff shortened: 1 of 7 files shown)


vaithak:
The example with `<random` is still not working. It still has some unsupported stmt warnings, mainly related to bitwise operations.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -738,11 +738,14 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     }
   }
   StmtDiff ReverseModeVisitor::VisitStmt(const Stmt* S) {
+    // if S is a SubstNonTypeTemplateParmExpr, we need to visit its replacement
+    // expression instead.
+    if (const auto* SNTTPE = dyn_cast<SubstNonTypeTemplateParmExpr>(S))
```
vgvassilev, line 743: We should add a method called `VisitSubstNonTypeTemplateParmExpr` instead of extending it here.

vaithak: my bad, updated 👍🏼

vgvassilev (approved):
LGTM!

vgvassilev:
> The example with `<random` is still not working. It still has some unsupported stmt warnings; maybe we can ignore those as they are related to bitwise operations, but it also produces one error related to wrongly referencing a local variable.

Maybe we will need reproducers for these issues.

PR #718, opened 2024-01-18
Fix seg fault for catch(...)

### LINKED ISSUE #86: Cloning try-block with catch(...) causes SegFault in StmtClone::CloneDecl
Can be reproduced with:

```
int f(int x) 
  try { return x; }
  catch(...) { return 0; }
...
clad::differentiate(f, 0);
```

vaithak:
fixes #86

### DIFF
--- a/lib/Differentiator/StmtClone.cpp
+++ b/lib/Differentiator/StmtClone.cpp
@@ -426,7 +426,9 @@ DEFINE_CLONE_STMT(ForStmt, (Ctx, Clone(Node->getInit()), Clone(Node->getCond()),
                             Node->getForLoc(), Node->getLParenLoc(), Node->getRParenLoc()))
 DEFINE_CLONE_STMT(ContinueStmt, (Node->getContinueLoc()))
 DEFINE_CLONE_STMT(BreakStmt, (Node->getBreakLoc()))
-DEFINE_CLONE_STMT(CXXCatchStmt, (Node->getCatchLoc(), static_cast<VarDecl*>(CloneDecl(Node->getExceptionDecl())), Clone(Node->getHandlerBlock())))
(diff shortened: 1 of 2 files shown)


lib/Differentiator/StmtClone.cpp
```diff
@@ -426,7 +426,10 @@ DEFINE_CLONE_STMT(ForStmt, (Ctx, Clone(Node->getInit()), Clone(Node->getCond()),
                             Node->getForLoc(), Node->getLParenLoc(), Node->getRParenLoc()))
 DEFINE_CLONE_STMT(ContinueStmt, (Node->getContinueLoc()))
 DEFINE_CLONE_STMT(BreakStmt, (Node->getBreakLoc()))
-DEFINE_CLONE_STMT(CXXCatchStmt, (Node->getCatchLoc(), static_cast<VarDecl*>(CloneDecl(Node->getExceptionDecl())), Clone(Node->getHandlerBlock())))
+DEFINE_CLONE_STMT(CXXCatchStmt, (Node->getCatchLoc(),
+                                 static_cast<VarDecl*>(
+                                     CloneDeclOrNull(Node->getExceptionDecl())),
+                                 Clone(Node->getHandlerBlock())))
```
vgvassilev, line 432:
```suggestion
DEFINE_CLONE_STMT(CXXCatchStmt, (Node->getCatchLoc(),                                      CloneDeclOrNull(Node->getExceptionDecl()), Clone(Node->getHandlerBlock())))
```

vaithak: removed the `statiic_cast` and ran clang-format again 👍🏼

vgvassilev: Either way the formatting looks ugly... let's move forward.

PR #719, opened 2024-01-19
Improve test for derivatives of Struct Method Calls

### LINKED ISSUE #49: Cloning call to method via clad::utils::StmtClone::VisitCXXMemberCallExpr causes segmentation fault
Can be reproduced by
```
struct S{
  int f0() { return 0; }
  int f1(int x) { return f0() + x; }
};
...
clad::differentiate(&S::f1, 0);
```

vaithak:
closes #49

### DIFF
--- a/test/FirstDerivative/StructMethodCall.C
+++ b/test/FirstDerivative/StructMethodCall.C
@@ -1,4 +1,5 @@
-// RUN: %cladclang %s -I%S/../../include 2>&1 | FileCheck %s
+// RUN: %cladclang %s -I%S/../../include -oStructMethodCall.out 2>&1 | FileCheck %s
+// RUN: ./StructMethodCall.out | FileCheck -check-prefix=CHECK-EXEC %s
 //CHECK-NOT: {{.*error|warning|note:.*}}
 
 #include "clad/Differentiator/Differentiator.h"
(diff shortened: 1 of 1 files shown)


vaithak:
#49 is already fixed; this just ensures that we have proper testing for the examples mentioned in the issue.

test/FirstDerivative/StructMethodCall.C
```diff
@@ -12,66 +13,101 @@ public:
...
   // CHECK-NEXT:     A *_d_this = &_d_this_obj;
-  // CHECK-NEXT: return _d_x + _d_y * y + y * _d_y;
+  // CHECK-NEXT:     return _d_x + _d_y * y + y * _d_y;
   // CHECK-NEXT: }
 
   // CHECK: int g_2_darg1(int x, int y) {
-  // CHECK-NEXT: int _d_x = 0;
-  // CHECK-NEXT: int _d_y = 1;
+  // CHECK-NEXT:     int _d_x = 0;
+  // CHECK-NEXT:     int _d_y = 1;
```
vgvassilev, line 56: Do we need these white space changes?

vaithak: I would say yes, because this makes the spacing uniform.with other lines of the same generated function and also matches the expected spacing.

vgvassilev (approved):
LGTM!

PR #738, opened 2024-01-26
Make built-in type variable declarations global

### LINKED ISSUE #681: Regression with Clad `master` in case of num-diff fallback in specific conditions

PetroZarytskyi:
When we produce a gradient function we generally have a forward and reverse
sweep. In the forward sweep we accumulate the state and in the reverse sweep
we use that state to invert the program execution. The forward sweep generally
follows the sematics of the primal function and when neccessary stores the state
which would be needed but lost for the reverse sweep.

However, to minimize the stores onto the tape we need to reuse some of the
variables between the forward and the reverse sweeps which requires some
variables to be promoted to the enclosing lexical scope of both sweeps.

Fixes https://github.com/vgvassilev/clad/issues/659, fixes https://github.com/vgvassilev/clad/issues/681.

### DIFF
--- a/include/clad/Differentiator/Array.h
+++ b/include/clad/Differentiator/Array.h
@@ -23,8 +23,8 @@ private:
   std::size_t m_size = 0;
 
 public:
-  /// Delete default constructor
-  array() = delete;
+  /// Default constructor
+  array() = default;
   /// Constructor to create an array of the specified size
   CUDA_HOST_DEVICE array(std::size_t size)
       : m_arr(new T[size]{static_cast<T>(0)}), m_size(size) {}
@@ -81,6 +81,12 @@ public:
   }
 
   CUDA_HOST_DEVICE array<T>& operator=(const array<T>& arr) {
+    if (m_size < arr.m_size) {
+      delete[] m_arr;
+      // NOLINTNEXTLINE(cppcoreguidelines-owning-memory)
+      m_arr = new T[arr.m_size];
+      m_size = arr.m_size;
+    }
     (*this) = arr.m_arr;
     return *this;
   }
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -215,8 +215,8 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       auto* gradientParam = gradientParams[i];
 
       auto* gradientVD =
-          BuildVarDecl(gradientParam->getType(), gradientParam->getName(),
-                       BuildDeclRef(overloadParam));
+          BuildGlobalVarDecl(gradientParam->getType(), gradientParam->getName(),
+                             BuildDeclRef(overloadParam));
       callArgs.push_back(BuildDeclRef(gradientVD));
       addToCurrentBlock(BuildDeclStmt(gradientVD));
     }
@@ -581,8 +581,8 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       if (utils::isArrayOrPointerType(VDDerivedType))
         continue;
       auto* VDDerived =
-          BuildVarDecl(VDDerivedType, "_d_" + param->getNameAsString(),
-                       getZeroInit(VDDerivedType));
+          BuildGlobalVarDecl(VDDerivedType, "_d_" + param->getNameAsString(),
+                             getZeroInit(VDDerivedType));
       m_Variables[param] = BuildDeclRef(VDDerived);
       addToBlock(BuildDeclStmt(VDDerived), m_Globals);
     }
@@ -746,7 +746,12 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
   }
 
   StmtDiff ReverseModeVisitor::VisitCompoundStmt(const CompoundStmt* CS) {
-    beginScope(Scope::DeclScope);
+    int scopeFlags = Scope::DeclScope;
+    // If this is the outermost compound statement of the function,
+    // propagate the function scope.
+    if (getCurrentScope() == m_DerivativeFnScope)
+      scopeFlags |= Scope::FnScope;
+    beginScope(scopeFlags);
     beginBlock(direction::forward);
     beginBlock(direction::reverse);
     for (Stmt* S : CS->body()) {
@@ -1271,26 +1276,10 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       return cloned;
     Expr* result = nullptr;
     Expr* forwSweepDerivative = nullptr;
-    if (utils::isArrayOrPointerType(target->getType())) {
-      // Create the target[idx] expression.
-      result = BuildArraySubscript(target, reverseIndices);
-      forwSweepDerivative =
-          BuildArraySubscript(target, forwSweepDerivativeIndices);
-    }
-    else if (isCladArrayType(target->getType())) {
-      result = m_Sema
-                   .ActOnArraySubscriptExpr(getCurrentScope(), target,
-                                            ASE->getExprLoc(),
-                                            reverseIndices.back(), noLoc)
-                   .get();
-      forwSweepDerivative =
-          m_Sema
-              .ActOnArraySubscriptExpr(getCurrentScope(), target,
-                                       ASE->getExprLoc(),
-                                       forwSweepDerivativeIndices.back(), noLoc)
-              .get();
-    } else
-      result = target;
+    // Create the target[idx] expression.
+    result = BuildArraySubscript(target, reverseIndices);
+    forwSweepDerivative =
+        BuildArraySubscript(target, forwSweepDerivativeIndices);
     // Create the (target += dfdx) statement.
     if (dfdx()) {
       auto* add_assign = BuildOp(BO_AddAssign, result, dfdx());
@@ -1301,10 +1290,10 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
   }
 
   StmtDiff ReverseModeVisitor::VisitDeclRefExpr(const DeclRefExpr* DRE) {
-    auto* clonedDRE = cast<DeclRefExpr>(Clone(DRE));
+    Expr* clonedDRE = Clone(DRE);
     // Check if referenced Decl was "replaced" with another identifier inside
     // the derivative
-    if (auto* VD = dyn_cast<VarDecl>(clonedDRE->getDecl())) {
+    if (auto* VD = dyn_cast<VarDecl>(cast<DeclRefExpr>(clonedDRE)->getDecl())) {
       // If current context is different than the context of the original
       // declaration (e.g. we are inside lambda), rebuild the DeclRefExpr
       // with Sema::BuildDeclRefExpr. This is required in some cases, e.g.
@@ -1312,7 +1301,12 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       // to the underlying struct of a lambda.
       if (VD->getDeclContext() != m_Sema.CurContext)
         clonedDRE = cast<DeclRefExpr>(BuildDeclRef(VD));
-
+      // This case happens when ref-type variables have to become function
+      // global. Ref-type declarations cannot be moved to the function global
+      // scope because they can't be separated from their inits.
+      if (DRE->getDecl()->getType()->isReferenceType() &&
+          clonedDRE->getType()->isPointerType())
+        clonedDRE = BuildOp(UO_Deref, clonedDRE);
       if (isVectorValued) {
         if (m_VectorOutput.size() <= outputArrayCursor)
           return StmtDiff(clonedDRE);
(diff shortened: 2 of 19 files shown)


parth-07:
Why do we need both this pull request and #688?

lib/Differentiator/StmtClone.cpp
```diff
@@ -520,8 +520,12 @@ bool ReferencesUpdater::VisitDeclRefExpr(DeclRefExpr* DRE) {
   // Replace the declaration if it is present in `m_DeclReplacements`.
   if (VarDecl* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
     auto it = m_DeclReplacements.find(VD);
-    if (it != std::end(m_DeclReplacements))
+    if (it != std::end(m_DeclReplacements)) {
       DRE->setDecl(it->second);
+      if (it->second->getType().getNonReferenceType()!=DRE->getType()) {
+        DRE->setType(it->second->getType().getNonReferenceType());
+      }
+    }
```
vgvassilev, line 528:
```suggestion
      QualType NonRefQT = it->second->getType().getNonReferenceType();
      if (NonRefQT != DRE->getType())
        DRE->setType(NonRefQT);
    }
```

Why this change was necessary? Can it go in a separate PR with a test?

lib/Differentiator/VisitorBase.cpp
```diff
@@ -143,8 +144,10 @@ namespace clad {
       m_Sema.ActOnUninitializedDecl(VD);
     }
     m_Sema.FinalizeDeclaration(VD);
+    if (!scope)
+        scope = getCurrentScope();
```
vgvassilev, line 148: Why was this necessary?

lib/Differentiator/VisitorBase.cpp
```diff
@@ -407,11 +418,19 @@ namespace clad {
   }
 
   Expr* VisitorBase::BuildArraySubscript(
-      Expr* Base, const llvm::SmallVectorImpl<clang::Expr*>& Indices) {
```
vgvassilev, line 410: Why dropping the const here - we are not changing the content of `Indices`.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2752,6 +2725,9 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     llvm::SmallVector<Decl*, 4> declsDiff;
     // Need to put array decls inlined.
     llvm::SmallVector<Decl*, 4> localDeclsDiff;
+    // FIXME: find a more reliable way to determine if the declaration
+    // is in the function global scope.
+    bool isInFunctionGlobalScope = m_Reverse.size() <= 2;
```
vgvassilev, line 2730: Can't we ask the current scope if it is a function global scope?

lib/Differentiator/VisitorBase.cpp
PetroZarytskyi: Declarations used to always be added to the current ``clang::Scope``, even the decl stmts were added to the function global scope in the derivative. Because of this, we could, for instance, declare multiple local variables ``i`` and subsequently have multiple ``_d_i`` decl stmts in the function global scope of the derivative. Such name conflicts were not detected. With this PR, we have way more global declarations and so this became more noticeable. You can take a look at ``BuildGlobalVarDecl`` to see why I introduced these lines.

lib/Differentiator/StmtClone.cpp
PetroZarytskyi: To make ref-type declarations global, I replaced them with pointers (to allow us to separate the declaration and the initialization). However, ``VisitorBase::Clone`` didn't expect a cloned expression to have a different type than the original expression, which led to broken AST. This could go in a separate PR but it doesn't make much sense by itself.

lib/Differentiator/VisitorBase.cpp
PetroZarytskyi: This caused an issue on some compiler versions but I fixed it without dropping const now.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2530,19 +2516,26 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
   VarDeclDiff ReverseModeVisitor::DifferentiateVarDecl(const VarDecl* VD) {
     StmtDiff initDiff;
     Expr* VDDerivedInit = nullptr;
+    // We take the parent of the current scope because the main compound
```
vgvassilev, line 2519: If `getCurrentScope()->isFunctionScope()` is false for `int f() {` this probably means that when we create the scope for the first compound statement of the function we should also add the `Scope::FnScope` flag to the current scope.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2752,6 +2725,9 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     llvm::SmallVector<Decl*, 4> declsDiff;
     // Need to put array decls inlined.
     llvm::SmallVector<Decl*, 4> localDeclsDiff;
+    // We take the parent of the current scope because the main compound
+    // statement of the function has its own scope as well.
+    bool isInFunctionGlobalScope = getCurrentScope()->getParent()==m_DerivativeFnScope;
```
vgvassilev, line 2730: See past comments.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2779,33 +2755,34 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
-              auto* assignment = BuildOp(BO_Assign, declRef, decl->getInit());
-              inits.push_back(BuildOp(BO_Comma, pushPop.getExpr(), assignment));
+        if (!isInFunctionGlobalScope) {
+          auto* decl = VDDiff.getDecl();
+          // The same variable will be assigned with new values every
+          // loop iteration so the const qualifier must be dropped.
+          if (decl->getType().isConstQualified()) {
+            QualType nonConstType =
+                getNonConstType(decl->getType(), m_Context, m_Sema);
+            decl->setType(nonConstType);
```
vgvassilev, line 2770: Changing the AST invariants that way is quite dangerous. Do we have a more elegant solution for future? We should add at least a bunch of tests for this case.

We should probably include perhaps a peusdo code example in the comments.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2779,33 +2755,34 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
-                  StoreAndRestore(declRef, /*prefix=*/"_t", /*force=*/true);
-              if (pushPop.getExpr() != declRef)
-                addToCurrentBlock(pushPop.getExpr_dx(), direction::reverse);
-              auto* assignment = BuildOp(BO_Assign, declRef, decl->getInit());
-              inits.push_back(BuildOp(BO_Comma, pushPop.getExpr(), assignment));
+        if (!isInFunctionGlobalScope) {
+          auto* decl = VDDiff.getDecl();
+          // The same variable will be assigned with new values every
+          // loop iteration so the const qualifier must be dropped.
+          if (decl->getType().isConstQualified()) {
```
vgvassilev, line 2767:
```suggestion
          QualType DeclTy = decl->getType();
          if (DeclTy.isConstQualified()) {
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2779,33 +2755,34 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+            QualType nonConstType =
+                getNonConstType(decl->getType(), m_Context, m_Sema);
+            decl->setType(nonConstType);
+          }
+          if (VD->getInit()) {
```
vgvassilev, line 2772: This case is also worth commenting including perhaps a peusdo code example.

lib/Differentiator/StmtClone.cpp
vgvassilev: Ok, the code suggestion still makes sense though..

lib/Differentiator/VisitorBase.cpp
vgvassilev: Maybe that's worth a comment in the code. I am still not getting how the scope can be nullptr.

vgvassilev (commented):
Glad to see moving! Here is a round of comments.

include/clad/Differentiator/Array.h
```diff
@@ -81,6 +81,11 @@ template <typename T> class array {
   }
 
   CUDA_HOST_DEVICE array<T>& operator=(const array<T>& arr) {
+    if (m_size < arr.m_size) {
+      delete[] m_arr;
+      m_arr = new T[arr.m_size];
```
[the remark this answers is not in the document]
PetroZarytskyi: Done.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: Done.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1301,18 +1285,20 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
       // If current context is different than the context of the original
       // declaration (e.g. we are inside lambda), rebuild the DeclRefExpr
       // with Sema::BuildDeclRefExpr. This is required in some cases, e.g.
       // Sema::BuildDeclRefExpr is responsible for adding captured fields
       // to the underlying struct of a lambda.
       if (VD->getDeclContext() != m_Sema.CurContext)
         clonedDRE = cast<DeclRefExpr>(BuildDeclRef(VD));
-
+      if (DRE->getDecl()->getType()->isReferenceType()
+          && clonedDRE->getType()->isPointerType())
```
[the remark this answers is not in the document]
PetroZarytskyi: This happens in the cases where we have to make ref-type variables global. Their clones have to become pointer-type to allow us to separate their declarations and initializations. I added a comment explaining this.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2530,19 +2516,26 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
   VarDeclDiff ReverseModeVisitor::DifferentiateVarDecl(const VarDecl* VD) {
     StmtDiff initDiff;
     Expr* VDDerivedInit = nullptr;
+    // We take the parent of the current scope because the main compound
+    // statement of the function has its own scope as well.
+    bool isInFunctionGlobalScope = getCurrentScope()->getParent()==m_DerivativeFnScope;
```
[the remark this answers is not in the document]
PetroZarytskyi: Because the outermost compound statement is not considered function scope.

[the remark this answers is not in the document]
PetroZarytskyi: I made the outermost compound statement FnScope so now this option works. I did this in my last commit.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: What exactly is dangerous? Can you give an example of what would break this?

vgvassilev: We want to only change the properties of the produced by Clad AST nodes. It seems the case here but I got confused.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2779,33 +2763,48 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+        if (promoteToFnScope) {
+          auto* decl = VDDiff.getDecl();
+          // The same variable will be assigned with new values every
+          // loop iteration so the const qualifier must be dropped.
+          // (look at the next comment for more details).
+          QualType DeclTy = decl->getType();
+          if (DeclTy.isConstQualified()) {
+            QualType nonConstType =
+                getNonConstType(decl->getType(), m_Context, m_Sema);
+            decl->setType(nonConstType);
```
vgvassilev, line 2792: Is codecov right that we do not have a test for this?

PetroZarytskyi: I got confused. This procedure is already done earlier. I removed this bit.

vgvassilev (approved):
LGTM! Good job!

PR #743, opened 2024-01-29
Update developer build process

maximusron:


README.md
```diff
@@ -272,18 +272,38 @@ cmake ../clad -DLLVM_DIR=/opt/homebrew/opt/llvm@12/lib/cmake/llvm -DClang_DIR=/o
...
+make -j8 check-clang # this installs llvm-config required by lit
+cd ../..
+```
+
+Clone and build Clad:
+```
+git clone https://github.com/vgvassilev/clad.git
+cd clad
+mkdir build && cd build
+cmake -DLLVM_DIR=PATH/TO/llvm-project/build -DClang_DIR=PATH/TO/llvm-project/build -DCMAKE_BUILD_TYPE=DEBUG -DLLVM_EXTERNAL_LIT="$(which lit)" ../
```
vgvassilev, line 300: Do we need even `-DClang_DIR` when we have `-DLLVM_DIR` set?

maximusron: right, it is not required, resolved

PR #747, opened 2024-02-03
Fix ambiguity in indexing for array_ref

vaithak:
relevant for https://github.com/root-project/root/pull/14165


vgvassilev (approved):
The issue seems to occur on 32 bit platforms where the pointer size is the same as the size of an int. We can't really test 32 bit platforms on the public github actions infrastructure and I cannot see a way how we can reliably provide a test case...

LGTM!

PR #749, opened 2024-02-07
Fix for char and string literals as function args

### LINKED ISSUE #671: Compiler error for inner function call with char argument

```c++

void g(char c){}

double f(double x) {
  g('a');
  return x;
}

int main() {
  auto f_dx = clad::differentiate(f, "x");
  f_dx.execute(3);
}
```

```
charcall.cpp:6:5: warning: attempted to differentiate unsupported statement, no changes applied
  g('a');
    ^
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
...
clang: error: unable to execute command: Segmentation fault (core dumped)
clang: error: clang frontend command failed due to signal (use -v to see invocation)
Ubuntu clang version 11.1.0-6
...
```

**Further information**: The error goes away when the char literal is replaced by an integer literal. The problem also shows up when a string literal is provided for a `char*` argument (which is the setup where I stumbled across this bug, but the above case shows that it's not related to pointers). Many thanks to @vgvassilev and @parth-07 for your quick responses to my earlier bug reports. I'm trying to apply Clad to a particle physics simulation code with as few changes to that code as possible, which seems to break a few things...

**Setup**: Clad built from the GitHub master branch at commit 1bdd261 on Ubuntu 22.04 with Clang 11.

vaithak:
fixes #671

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -562,10 +562,13 @@ namespace clad {
         // bitmask_opts is a template pack of unsigned integers, so we need to
         // do bitwise or of all the values to get the final value.
         unsigned bitmasked_opts_value = 0;
-        for (auto const& arg :
-             FD->getTemplateSpecializationArgs()->get(0).pack_elements()) {
-          bitmasked_opts_value |= arg.getAsIntegral().getExtValue();
-        }
+        const auto template_arg = FD->getTemplateSpecializationArgs()->get(0);
+        if (template_arg.getKind() == TemplateArgument::Pack)
+          for (const auto& arg :
+               FD->getTemplateSpecializationArgs()->get(0).pack_elements())
+            bitmasked_opts_value |= arg.getAsIntegral().getExtValue();
+        else
+          bitmasked_opts_value = template_arg.getAsIntegral().getExtValue();
         unsigned derivative_order =
             clad::GetDerivativeOrder(bitmasked_opts_value);
         if (derivative_order == 0) {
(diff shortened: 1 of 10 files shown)


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -562,9 +562,14 @@ namespace clad {
...
-             FD->getTemplateSpecializationArgs()->get(0).pack_elements()) {
-          bitmasked_opts_value |= arg.getAsIntegral().getExtValue();
+        const auto template_arg = FD->getTemplateSpecializationArgs()->get(0);
+        if (template_arg.getKind() == TemplateArgument::Pack) {
+          for (const auto& arg :
+               FD->getTemplateSpecializationArgs()->get(0).pack_elements()) {
+            bitmasked_opts_value |= arg.getAsIntegral().getExtValue();
+          }
+        } else {
+          bitmasked_opts_value = template_arg.getAsIntegral().getExtValue();
```
vaithak, line 572: codecov reports this as missing coverage, as it was only required for the `build_release` version.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -602,9 +607,14 @@ namespace clad {
...
```
vaithak, line 617: same here.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -562,9 +562,14 @@ namespace clad {
...
         unsigned bitmasked_opts_value = 0;
-        for (auto const& arg :
-             FD->getTemplateSpecializationArgs()->get(0).pack_elements()) {
-          bitmasked_opts_value |= arg.getAsIntegral().getExtValue();
+        const auto template_arg = FD->getTemplateSpecializationArgs()->get(0);
+        if (template_arg.getKind() == TemplateArgument::Pack) {
+          for (const auto& arg :
+               FD->getTemplateSpecializationArgs()->get(0).pack_elements()) {
+            bitmasked_opts_value |= arg.getAsIntegral().getExtValue();
+          }
```
vgvassilev, line 570:
```suggestion
          for (const auto& arg :
               FD->getTemplateSpecializationArgs()->get(0).pack_elements())
            bitmasked_opts_value |= arg.getAsIntegral().getExtValue();
```

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -562,9 +562,14 @@ namespace clad {
...
+        } else {
```
vgvassilev, line 571: We can drop the braces here too...

vaithak:
The coverage failure is because of statements which executes for debug and build_release versions as mentioned above.

vgvassilev (approved):
Codecov seems confusing itself. LGTM!

Issue #762, opened 2024-02-15
Bug: passing around C-stryle array parameters crashes Clad


guitargeek:
Using array parameters that are passed to the differentiated function as function arguments crashes Clad.

This is very important to support for the RooFit usecase, where we want to forward arrays of constant values like this, also passing arrays with offsets like `obs + 2`.

Reproducer:
```c++
// Compile with clang++ -std=c++11 -fplugin=/usr/lib/clad.so cladIssue.cxx -o cladIssue

#include "clad/Differentiator/Differentiator.h"

#include <iostream>

inline double foo(double *params, double *obs)
{
   return params[0] * obs[0];
}

double wrapper(double* params, double* obs)
{
   // return params[0] * obs[0]; // works!

   return foo(params, obs); // doesn't work!
}

int main()
{
   auto grad = clad::gradient(wrapper, "params");

   double param = 2.0;
   double obs = 2.0;
   double result = 0.0;
   grad.execute(&param, &obs, &result);

   std::cout << result << std::endl;
}
```

vgvassilev:
That is probably related to #735.

vaithak:
This issue is related to the initialization of derivatives of pointer arguments.
In the above example, the differentiation is w.r.t `params` only, so the user won't send `clad::array_ref<double> _d_obs` to us.
How do we initialize `_d_obs`? We don't have any size information about it.

The error is removed when we also include `obs` in the independent variables and pass `_d_obs` as an argument in the execute method.
```cpp
   auto grad = clad::gradient(wrapper, "params, obs");

   double param = 2.0;
   double obs = 2.0;
   double _d_params = 0.0;
   double _d_obs = 0.0;
   grad.execute(&param, &obs, &_d_params, &_d_obs);
```

guitargeek:
Hi @vaithak, thanks for the reply! That would not solve the problem in my use case I think, because doesn't differentiating also with respect to `obs` introduce a huge overhead? There might be only a few elements in `params` but thousands in `obs`, and for all of them I'm not interested in the gradient.

Is there a workaround the problem without generating a different gradient?

Thanks again!

vaithak:
Hi @guitargeek, I agree with you that having a gradient of `obs` will cause unnecessary operations. That is why when doing `params[0] * obs[0]` inside the wrapper function itself, we don't produce gradient operations for `obs` because we know the user hasn't requested it.
[...]

Issue #636, opened 2023-10-05
Bug: Passing `nullptr` around crashes clad, even if they are not dereferenced


guitargeek:
It would be nice if passing around null pointers would nor crash the gradient generation, because then one could apply AD to function with optional pointer arguments.

My setup: Arch Linux with `Clad` master and clang 16.06.

I hope you can reproduce this problem, and a fix will be very appreciated. If this would have worked, it would have saved me lots of time debugging :slightly_smiling_face: 

Reproducer:
```c++
// Compile with clang++ -std=c++11 -fplugin=/usr/lib/clad.so cladIssue.cxx -o cladIssue

#include "clad/Differentiator/Differentiator.h"

#include <iostream>
#include <vector>

double func(double *params, double *)
{
   return params[0];
}

double wrapper(double *params)
{
   return func(params, 0);
   // also nullptr will not work (would be nice if it does)
   // At least with nullptr one gets this warning:
   // warning: attempted to differentiate unsupported statement, no changes applied
}

int main()
{
   auto grad = clad::gradient(wrapper, "params");

   double x = 2.0;
   double result = 0.0;
   grad.execute(&x, &result);

   std::cout << result << std::endl;
}
```

guitargeek:
Here is the stack trace I get:
```
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
[...]

PR #763, opened 2024-02-15
Fix passing pointers as call arguments

### LINKED ISSUE #735: Errors on passing pointers as an argument to function calls
Clang reports errors on clad generated code when pointers are passed as arguments to function calls.

Reproducible example:
```cpp

void do_nothing(const double *v, double *u) { }

    double *pv = &v;
    double *pu = &u;
    do_nothing(pv, pu);
    return v;
}

int main() {
}
```

**Errors:**
All errors are similar to:
```
 error: no viable conversion from 'double *' to 'clad::array<double>'
```

vaithak:
fixes #735, #636

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1166,6 +1166,11 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
                                    SL->getType(), utils::GetValidSLoc(m_Sema)));
   }
 
+  StmtDiff ReverseModeVisitor::VisitCXXNullPtrLiteralExpr(
+      const CXXNullPtrLiteralExpr* NPE) {
+    return StmtDiff(Clone(NPE), Clone(NPE));
+  }
+
   StmtDiff ReverseModeVisitor::VisitReturnStmt(const ReturnStmt* RS) {
     // Initially, df/df = 1.
     const Expr* value = RS->getRetValue();
@@ -1360,7 +1365,9 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
   }
 
   StmtDiff ReverseModeVisitor::VisitIntegerLiteral(const IntegerLiteral* IL) {
-    return StmtDiff(Clone(IL));
+    auto* Constant0 =
+        ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
+    return StmtDiff(Clone(IL), Constant0);
   }
 
   StmtDiff ReverseModeVisitor::VisitFloatingLiteral(const FloatingLiteral* FL) {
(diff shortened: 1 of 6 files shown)


vgvassilev:
Can you check if this PR fixes #762 by any chance?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1475,8 +1483,11 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         // ```
         // _r0 = _d_a;
         // ```
-        Expr* dArg = nullptr;
-        if (utils::isArrayOrPointerType(argDiff.getExpr()->getType())) {
+        if (argDiff.getExpr_dx() && utils::IsLiteral(argDiff.getExpr_dx())) {
+          dArg = StoreAndRef(argDiff.getExpr_dx(), arg->getType(),
+                             direction::reverse, "_r",
+                             /*forceDeclCreation=*/true);
+        } else if (argDiff.getExpr()->getType()->isArrayType()) {
```
vgvassilev, line 1490: We should refactor `argDiff.getExpr()->getType()` into a temporary to improve readibility.

vaithak:
> Can you check if this PR fixes #762 by any chance?

That is because of a separate reason, commented on that issue.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1536,7 +1551,8 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       // FIXME: We cannot use GlobalStoreAndRef to store a whole array so now
       // arrays are not stored.
       StmtDiff argDiffStore;
-      if (passByRef && !argDiff.getExpr()->getType()->isArrayType())
+      if (passByRef && !argDiff.getExpr()->getType()->isArrayType() &&
```
vgvassilev, line 1554: Can't we use `argDiffType` here, too?

vaithak: It was inside the if condition earlier, but moved it out now 👍🏼

vgvassilev (approved):
LGTM!

parth-07:
@vaithak Can you please update pull request description / commit message with details how it addresses the issue?

Does this pull request addresses pointer arithmetic expressions such as `p + 1` used as function call arguments? If so, can you please add a test?

Also, wouldn't this pull request https://github.com/vgvassilev/clad/issues/762 fix the issue as well?

vgvassilev:
@parth-07, I did not see your comment before merging. I guess I was too excited about this PR. Apologies!

vaithak:
> @vaithak Can you please update pull request description / commit message with details how it addresses the issue?
> 
> Does this pull request addresses pointer arithmetic expressions such as `p + 1` used as function call arguments? If so, can you please add a test?
> 
> Also, wouldn't this pull request #762 fix the issue as well?

For #762, I have commented more about that issue itself. Regarding pointer arithmetic, I will test it out and make changes if required along with test cases in a separate PR.

PR #764, opened 2024-02-15
Fix pointer arithmetic for array types

vaithak:
required for workaround mentioned in #762


parth-07:
Thank you for the quick fix. However, if we don't need this urgently, then I think we should wait for #758.

parth-07 (approved):
Looks good.

PR #773, opened 2024-02-16
Simplify error estimation by removing _EERepl_ and _delta_

### LINKED ISSUE #757: Remove excessive stores from the error estimation mode
In the error estimation mode, we create ``_EERepl_`` variables, which are used to store the values of variables at every point of the function to then use them in the reverse sweep. However, on the current master, we can simply use the original variables instead since we now restore their values in the reverse sweep. We don't have to store any intermediate values in error estimation now.

PetroZarytskyi:
This PR simplifies error estimation by removing variables with prefixes ``_EERepl_`` and ``_delta_``. This became possible since we started reusing cloned forward pass variables in the reverse pass. Merging this PR will also unlock #758.
Fixes #757.

### DIFF
--- a/include/clad/Differentiator/ExternalRMVSource.h
+++ b/include/clad/Differentiator/ExternalRMVSource.h
@@ -138,8 +138,10 @@ public:
                                             clang::BinaryOperatorKind& opCode) {
   }
 
-  /// This is called just after finaising processing of assignment operator.
-  virtual void ActBeforeFinalisingAssignOp(clang::Expr*&, clang::Expr*&){};
+  /// This is called just after finalising processing of assignment operator.
+  virtual void ActBeforeFinalisingAssignOp(clang::Expr*&, clang::Expr*&,
+                                           clang::Expr*&,
+                                           clang::BinaryOperator::Opcode&){};
 
   /// This is called at that beginning of
   /// `ReverseModeVisitor::DifferentiateSingleStmt`.
--- a/include/clad/Differentiator/ExternalRMVSource.h
+++ b/include/clad/Differentiator/ExternalRMVSource.h
@@ -104,7 +104,7 @@ public:
 
   /// This is called just before finalising processing of Single statement
   /// branch in `VisitBranch` lambda in
-  virtual void ActBeforeFinalisingVisitBranchSingleStmtInIfVisitStmt() {}
+  virtual void ActBeforeFinalizingVisitBranchSingleStmtInIfVisitStmt() {}
 
   /// This is called just before differentiating init statement of loops.
   virtual void ActBeforeDifferentiatingLoopInitStmt() {}
@@ -117,7 +117,7 @@ public:
   virtual void ActAfterProcessingSingleStmtBodyInVisitForLoop() {}
 
   /// This is called just before finalising `VisitReturnStmt`.
-  virtual void ActBeforeFinalisingVisitReturnStmt(StmtDiff& retExprDiff) {}
+  virtual void ActBeforeFinalizingVisitReturnStmt(StmtDiff& retExprDiff) {}
 
   /// This ic called just before finalising `VisitCallExpr`.
   ///
@@ -131,7 +131,7 @@ public:
 
   /// This is called just before finalising processing of post and pre
   /// increment and decrement operations.
-  virtual void ActBeforeFinalisingPostIncDecOp(StmtDiff& diff){};
+  virtual void ActBeforeFinalizingPostIncDecOp(StmtDiff& diff){};
 
   /// This is called just after cloning of LHS assignment operation.
   virtual void ActAfterCloningLHSOfAssignOp(clang::Expr*&, clang::Expr*&,
@@ -139,7 +139,7 @@ public:
   }
 
   /// This is called just after finalising processing of assignment operator.
-  virtual void ActBeforeFinalisingAssignOp(clang::Expr*&, clang::Expr*&,
+  virtual void ActBeforeFinalizingAssignOp(clang::Expr*&, clang::Expr*&,
                                            clang::Expr*&,
                                            clang::BinaryOperator::Opcode&){};
 
--- a/test/ErrorEstimation/Assignments.C
+++ b/test/ErrorEstimation/Assignments.C
@@ -165,6 +165,27 @@ float func7(float x, float y) { return (x * y); }
 //CHECK-NEXT:     _final_error += std::abs(1. * _ret_value0 * {{.+}});
 //CHECK-NEXT: }
 
+float func8(int x, int y) {
+    x = y * y;
+    return x;
+}
+
+//CHECK: void func8_grad(int x, int y, clad::array_ref<int> _d_x, clad::array_ref<int> _d_y, double &_final_error) {
+//CHECK-NEXT:     int _t0;
+//CHECK-NEXT:     _t0 = x;
+//CHECK-NEXT:     x = y * y;
+//CHECK-NEXT:     goto _label0;
+//CHECK-NEXT:   _label0:
+//CHECK-NEXT:     * _d_x += 1;
+//CHECK-NEXT:     {
+//CHECK-NEXT:         x = _t0;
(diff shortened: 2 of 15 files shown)


vgvassilev:
@grimmmyshini, could you take a look?

include/clad/Differentiator/ExternalRMVSource.h
```diff
@@ -138,8 +138,9 @@ class ExternalRMVSource {
                                             clang::BinaryOperatorKind& opCode) {
   }
 
-  /// This is called just after finaising processing of assignment operator.
-  virtual void ActBeforeFinalisingAssignOp(clang::Expr*&, clang::Expr*&){};
+  /// This is called just after finalising processing of assignment operator.
+  virtual void ActBeforeFinalisingAssignOp(clang::Expr*&, clang::Expr*&, clang::Expr*&,
```
vgvassilev, line 142: We should use US English.
```suggestion
  virtual void ActBeforeFinalizingAssignOp(clang::Expr*&, clang::Expr*&, clang::Expr*&,
```

test/ErrorEstimation/Assignments.C
```diff
@@ -108,34 +97,25 @@ float func4(float x, float y) {
...
 //CHECK-NEXT:         x = _t0;
 //CHECK-NEXT:         float _r_d0 = * _d_x;
 //CHECK-NEXT:         * _d_x -= _r_d0;
 //CHECK-NEXT:         _d_z += _r_d0;
 //CHECK-NEXT:         * _d_y += _r_d0;
-//CHECK-NEXT:         _delta_x += std::abs(_r_d0 * _EERepl_x1 * {{.+}});
 //CHECK-NEXT:         * _d_x;
 //CHECK-NEXT:     }
 //CHECK-NEXT:     * _d_y += _d_z;
-//CHECK-NEXT:     _delta_x += std::abs(* _d_x * _EERepl_x0 * {{.+}});
```
grimmmyshini, line 135: I am making a comment here, but this applies to all the other tests that have been changed similarly. 

1.) You cannot get rid of the _EERepl_ vars (at least to my understanding) so easily. In error estimation, we need to "replay" the value of the variable of interest as it was when we started. (Essentially, every variable is a unique instantiation and is never overwritten). For e.g. the following program:

```
x = x + 1;
x = y; 
y = 20 + x;
```
Essentially means this to the error estimator:

```
x_1 = x_0 + 1;
x_2 = y_0; 
y_1 = 20 + x_2;
```
 Then, the final error estimation is essentially this:

```
final_error = dx_0 * x_0 + dx_1 * x_1 + dx_2 * x_2 + ...
```

So, you cannot simply replace the eerepl objects with 'x' because most functions will overwrite values multiple times.

2.) You cannot just get rid of the _delta_ vars as you may require them while printing/logging. Sure, we can change this but it makes it significantly easier for us to identify errors through these names in the code.

grimmmyshini:
I have left a comment in the code addressing one of my major concerns with this PR. That comment applies to every test that has been changed. I think this PR should also be run on the benchmarks we have for the paper. The resulting error values should also be cross-checked, because, as this PR stands, I do not think it is correct and should be merged.

test/ErrorEstimation/Assignments.C
PetroZarytskyi: > 1.) You cannot get rid of the EERepl vars...

Since 1.3, we started reusing the original variables (``x`` and ``y``) in the reverse sweep. We now restore their values in the reverse sweep. So at every point in the reverse sweep, ``x`` has the same value it had in the corresponding point in the forward sweep, not the last value it was equal to (like it was previously). This is why the error estimation statement has been moved to the top of the enclosing reverse-sweep block (because we want the value of ``x`` after the assignment). Hopefully, I understand your concern correctly.

PetroZarytskyi: > 2.) You cannot just get rid of the delta vars...

Can't we instead rely on custom error evaluation here (``getErrorVal``)? I think we can get the same information this way without adding excessive code. Anyway, this seems like something we could discuss in more detail in Slack.

test/ErrorEstimation/LoopsAndArrays.C
```diff
@@ -69,55 +58,43 @@ float func2(float x) {
...
 //CHECK-NEXT:         }
 //CHECK-NEXT:         {
+//CHECK-NEXT:             _final_error += std::abs(_d_m * m * {{.+}});
 //CHECK-NEXT:             * _d_x += _d_m * x;
 //CHECK-NEXT:             * _d_x += x * _d_m;
 //CHECK-NEXT:             _d_m = 0;
 //CHECK-NEXT:             m = clad::pop(_t1);
-//CHECK-NEXT:             float _r0 = clad::pop(_EERepl_m0);
-//CHECK-NEXT:             _delta_m += std::abs(_d_m * _r0 * {{.+}});
 //CHECK-NEXT:         }
```
PetroZarytskyi, line 95: We should notice that before this PR, we used _d_m after it was set to 0. This is clearly not intended because we should use the old derivative value. On the other hand, moving the error statement to the top solves the problem.

PetroZarytskyi:
@grimmmyshini @vgvassilev
This screenshot shows the discrepancy in the benchmark results before (on the right) and after (on the left) this PR:
Screenshot 2024-03-05 at 23 37 25
During our meeting last Sunday, we concluded that the most problematic part is the ``beta`` variable.
This is the part of the derivative code that causes this discrepancy:
Before the PR:
```
{
    double _r57 = clad::pop(_t125);
    double _r58 = _d_beta / _r57;
     _d_rtrans += _r58;
    double _r59 = _d_beta * -clad::pop(_t126) / (_r57 * _r57);
     _d_oldrtrans += _r59;
     _d_beta = 0;              <-- _d_beta is set to 0
     double _r60 = clad::pop(_EERepl_beta0);
     _delta_beta += clad::getErrorVal(_d_beta, _r60, "beta");    <-- the error is 0
}
```
After the PR:
```
{
    _final_error += clad::getErrorVal(_d_beta, beta, "beta");  <-- the old value of _d_beta is used
    _d_rtrans += _d_beta / oldrtrans;
    double _r1 = _d_beta * -rtrans / (oldrtrans * oldrtrans);
    _d_oldrtrans += _r1;
    _d_beta = 0;
    beta = clad::pop(_t48);
}
```
From what I understand, the new behavior is correct here. You can also see my comment above describing the same in one of the tests. Luckily, this doesn't affect the conclusions drawn in the article.

vgvassilev:
@grimmmyshini, ping.

grimmmyshini:
I see! Well, it's weird that we missed it in the first go but no issue. If you have verified all the error results from the tests, then this PR is ready to go. 👍

vgvassilev:
@PetroZarytskyi, any chances to increase the testing coverage of this patch?

PetroZarytskyi:
@vgvassilev 
> @PetroZarytskyi, any chances to increase the testing coverage of this patch?

I looked into the CodeDev report. At first, I was confused how it was possible that the coverage of every single test either increased or didn't change. Then I realized that I made ErrorEstimator.cpp way smaller and that decreased the average project coverage. This happened because the coverage of ErrorEstimator.cpp is higher than average. So I think making the coverage higher again would involve doing that for lines unrelated to the PR.

vgvassilev:
> @vgvassilev
> 
> > @PetroZarytskyi, any chances to increase the testing coverage of this patch?
> 
> I looked into the CodeDev report. At first, I was confused how it was possible that the coverage of every single test either increased or didn't change. Then I realized that I made ErrorEstimator.cpp way smaller and that decreased the average project coverage. This happened because the coverage of ErrorEstimator.cpp is higher than average. So I think making the coverage higher again would involve doing that for lines unrelated to the PR.

That was my take as well. Thanks for the explanation. Let's move forward.

vgvassilev (approved):
Well done! Thank you! We can ignore the codecov as we reduced the code significantly and that's what makes codecov unhappy but makes everybody else happy...

PR #777, opened 2024-02-19
Initial support for memory operations in reverse mode

### LINKED ISSUE #654: Compiler error related to `new` and/or pointers
```c++
#include <iostream>

struct T {};

double f(double x){
  T* y = new T{};
  return x;
}

int main(){
  double d;
  f_dx.execute(3,&d);
}
```

```
heapallocation.cpp:7:10: warning: attempted to differentiate unsupported statement, no changes applied
  T* y = new T{};
         ^
clang: /home/aehle/clad/src/lib/Differentiator/StmtClone.cpp:500: clang::Stmt* clad::utils::StmtClone::VisitStmt(clang::Stmt*): Assertion `0 && "clone not fully implemented"' failed.
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
...
```

**Further information**: An analogous error message shows up if the type `T` is replaced by e.g. `char` or `double`. 

**Setup**: Clad built from the GitHub master branch at commit https://github.com/vgvassilev/clad/commit/1bdd261ea71163b3e8074dca4018e94a481a1c50 on Ubuntu 22.04 with Clang 11.

vaithak:
This does not add complete support for `new` and `delete` operators, there are many examples that can be constructed which will fail or crash. However, it covers the mostly-used case of allocating memory in the beginning and deallocating in the end.

Edit: Also, added C-style memory alloc and `free` operations.

fixes #654

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -196,6 +196,31 @@ CUDA_HOST_DEVICE void clamp_pullback(const T& v, const T& lo, const T& hi,
 #endif
 
 } // namespace std
+
+// NOLINTBEGIN(cppcoreguidelines-no-malloc)
+// NOLINTBEGIN(cppcoreguidelines-owning-memory)
+inline ValueAndPushforward<void*, void*> malloc_pushforward(size_t sz,
+                                                            size_t d_sz) {
+  return {malloc(sz), malloc(sz)};
+}
+
+inline ValueAndPushforward<void*, void*>
+calloc_pushforward(size_t n, size_t sz, size_t d_n, size_t d_sz) {
+  return {calloc(n, sz), calloc(n, sz)};
+}
+
+inline ValueAndPushforward<void*, void*>
+realloc_pushforward(void* ptr, size_t sz, void* d_ptr, size_t d_sz) {
+  return {realloc(ptr, sz), realloc(d_ptr, sz)};
+}
+
+inline void free_pushforward(void* ptr, void* d_ptr) {
+  free(ptr);
+  free(d_ptr);
+}
+// NOLINTEND(cppcoreguidelines-owning-memory)
+// NOLINTEND(cppcoreguidelines-no-malloc)
+
 // These are required because C variants of mathematical functions are
 // defined in global namespace.
 using std::abs_pushforward;
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -641,5 +641,19 @@ namespace clad {
              isa<ObjCBoolLiteralExpr>(E) || isa<CXXBoolLiteralExpr>(E) ||
              isa<GNUNullExpr>(E);
     }
+
+    bool IsMemoryAllocationFunction(const clang::FunctionDecl* FD) {
+      if (FD->getBuiltinID() == Builtin::BImalloc)
+        return true;
+      if (FD->getBuiltinID() == Builtin::BIcalloc)
+        return true;
+      if (FD->getBuiltinID() == Builtin::BIrealloc)
+        return true;
+      return false;
+    }
+
+    bool IsMemoryDeallocationFunction(const clang::FunctionDecl* FD) {
+      return FD->getBuiltinID() == Builtin::BIfree;
+    }
   } // namespace utils
 } // namespace clad
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -6,6 +6,7 @@
 #include "clang/AST/Decl.h"
 #include "clang/AST/Expr.h"
 #include "clang/AST/RecursiveASTVisitor.h"
+#include "clang/Basic/Builtins.h"
 #include "clang/Basic/SourceLocation.h"
 #include "clang/Sema/Lookup.h"
 #include "llvm/ADT/SmallVector.h"
@@ -400,6 +401,12 @@ namespace clad {
       auto& C = semaRef.getASTContext();
       if (!TSI)
         TSI = C.getTrivialTypeSourceInfo(qType);
+      if (clad_compat::isa_and_nonnull<ImplicitValueInitExpr>(initializer))
+        // If the initializer is an implicit value init expression, then
+        // we don't need to pass it explicitly to the CXXNewExpr. As, clang
+        // internally adds it when initializer is ParenListExpr and
+        // DirectInitRange is valid.
+        initializer = semaRef.ActOnParenListExpr(noLoc, noLoc, {}).get();
       auto newExpr =
           semaRef
               .BuildCXXNew(
@@ -642,18 +649,36 @@ namespace clad {
              isa<GNUNullExpr>(E);
     }
 
-    bool IsMemoryAllocationFunction(const clang::FunctionDecl* FD) {
+    bool IsMemoryFunction(const clang::FunctionDecl* FD) {
+
+#if CLANG_VERSION_MAJOR > 12
       if (FD->getBuiltinID() == Builtin::BImalloc)
         return true;
-      if (FD->getBuiltinID() == Builtin::BIcalloc)
+      if (FD->getBuiltinID() == Builtin::ID::BIcalloc)
         return true;
-      if (FD->getBuiltinID() == Builtin::BIrealloc)
+      if (FD->getBuiltinID() == Builtin::ID::BIrealloc)
         return true;
+      if (FD->getBuiltinID() == Builtin::ID::BImemset)
+        return true;
+#else
+      if (FD->getNameAsString() == "malloc")
+        return true;
+      if (FD->getNameAsString() == "calloc")
+        return true;
+      if (FD->getNameAsString() == "realloc")
+        return true;
+      if (FD->getNameAsString() == "memset")
+        return true;
+#endif
       return false;
     }
 
     bool IsMemoryDeallocationFunction(const clang::FunctionDecl* FD) {
-      return FD->getBuiltinID() == Builtin::BIfree;
+#if CLANG_VERSION_MAJOR > 12
+      return FD->getBuiltinID() == Builtin::ID::BIfree;
+#else
+      return FD->getNameAsString() == "free";
+#endif
     }
   } // namespace utils
 } // namespace clad
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1445,7 +1445,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     // For calls to C-style memory allocation functions, we do not need to
     // differentiate the call. We just need to visit the arguments to the
     // function.
-    if (utils::IsMemoryAllocationFunction(FD)) {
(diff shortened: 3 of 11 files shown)


vaithak:
@parth-07 Agreed, this is in no way a perfect solution, but meant as a fix for preventing clad from crashing on `new` and `delete` statements. For supporting all the use cases, like the one you mentioned would require a more complex analysis which can be done incrementally when use cases/issues arrive from users.
For example, in the first case, just storing the name of the pointer variable won't be sufficient. As, we can create a new pointer variable pointing to the same location and delete using that one, instead of the one we used while calling `new` expression.

vgvassilev:
> @parth-07 Agreed, this is in no way a perfect solution, but meant as a fix for preventing clad from crashing on `new` and `delete` statements. For supporting all the use cases, like the one you mentioned would require a more complex analysis which can be done incrementally when use cases/issues arrive from users. For example, in the first case, just storing the name of the pointer variable won't be sufficient. As, we can create a new pointer variable pointing to the same location and delete using that one, instead of the one we used while calling `new` expression.

Is there a way to diagnose the unsupported cases?

parth-07:
> For supporting all the use cases, like the one you mentioned would require a more complex analysis which can be done incrementally when use cases/issues arrive from users.

Iteratively adding features when they are required is of course the preferred way. However, we should have an idea on how to support the missing features, otherwise, later on, we may need to change the initial groundwork.

> For example, in the first case, just storing the name of the pointer variable won't be sufficient. As, we can create a new pointer variable pointing to the same location and delete using that one, instead of the one we used while calling new expression.

Yes, the first case is easier to resolve. But the second case, `delete` calls across function boundaries, can be troublesome. 

> but meant as a fix for preventing clad from crashing on new and delete statements.

If the goal for now is simply to prevent clad from crashing. Then perhaps we can simply ignore the `delete` statements. I don't think doing manual memory allocation in a mathematical function is a common case.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2573,6 +2595,11 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     bool isDerivativeOfRefType = VD->getType()->isReferenceType();
     VarDecl* VDDerived = nullptr;
     bool isPointerType = VD->getType()->isPointerType();
+    bool isInitializedByNewExpr = false;
+    // Check if the variable is pointer type and initialized by new expression
+    if (isPointerType && (VD->getInit() != nullptr) &&
```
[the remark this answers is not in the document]
vaithak: clang-tidy suggested to have explicit `!= nullptr` in one of the resolved comments above.

vaithak:
> If the goal for now is simply to prevent clad from crashing. Then perhaps we can simply ignore the `delete` statements. I don't think doing manual memory allocation in a mathematical function is a common case.

I think it is better to support it minimally than to not support it at all. I don't see a way of how the analysis of such complex cases will look (even for cases like for loops or if conditions), but even if we have to remove the work in this commit entirely for a better solution, I don't see that as an issue.

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -196,6 +196,26 @@ CUDA_HOST_DEVICE void clamp_pullback(const T& v, const T& lo, const T& hi,
 #endif
 
 } // namespace std
+
+ValueAndPushforward<void*, void*> malloc_pushforward(size_t sz, size_t d_sz) {
```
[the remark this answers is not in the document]
vgvassilev: We need to mark this function `inline`?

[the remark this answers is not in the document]
vaithak: yup, clang-tidy seems to be suggesting so.

lib/Differentiator/CladUtils.cpp
```diff
@@ -641,5 +641,19 @@ namespace clad {
              isa<ObjCBoolLiteralExpr>(E) || isa<CXXBoolLiteralExpr>(E) ||
              isa<GNUNullExpr>(E);
     }
+
+    bool IsMemoryAllocationFunction(const clang::FunctionDecl* FD) {
```
vgvassilev, line 645: Can we use `FunctionDecl::getMemoryFunctionKind()`?

lib/Differentiator/ReverseModeVisitor.cpp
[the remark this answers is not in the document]
vgvassilev: We should fix clang-tidy if possible...

lib/Differentiator/CladUtils.cpp
vaithak: In its source code, I don't see `malloc` being checked: https://clang.llvm.org/doxygen/Decl_8cpp_source.html#l04356.

vgvassilev: I see `Builtin::BImalloc`...

vaithak: Tried this, it didn't work.

vgvassilev: It's probably worth trying to understand why as that'd be a design choice that will be extended in future without a lot of thinking ;)

vaithak: Can you point the line number of link to the line where you found `BImalloc`, I still can't find it in the condition.

parth-07:
> I think it is better to support it minimally than to not support it at all. 

We will be able to support the differentiation of functions that use `new` and `delete` if we ignore the `delete` statements. By ignoring `delete` statement, I mean that no statement will be generated in the adjoint code for a `delete` statement in the primal code. Differentiation will produce correct results even if there is a double `delete` or if there's `delete` across function boundaries. We basically have to choose between:

- Ignore `delete` statements and have correct derivatives at the cost of memory leaks in the generated code.
- Generate `delete` statements in the generated derivative code to avoid memory leaks. In this case, only the basic `delete` statement usage will work correctly. Any advance usage may result in Clad crashing, or an undefined behavior.

> I don't see a way of how the analysis of such complex cases will look (even for cases like for loops or if conditions), but even if we have to remove the work in this commit entirely for a better solution in future, I don't see that as an issue.

Yes, it's not a big issue. However, I believe more effort will be required in this way. I was just giving my two cents. Please continue with the approach that you find best.

vaithak:
> We will be able to support the differentiation of functions that use `new` and `delete` if we ignore the `delete` statements. By ignoring `delete` statement, I mean that no statement will be generated in the adjoint code for a `delete` statement in the primal code. Differentiation will produce correct results even if there is a double `delete` or if there's `delete` across function boundaries. We basically have to choose between:
> 
> * Ignore `delete` statements and have correct derivatives at the cost of memory leaks in the generated code.
> * Generate `delete` statements in the generated derivative code to avoid memory leaks. In this case, only the basic `delete` statement usage will work correctly. Any advance usage may result in Clad crashing, or an undefined behavior.
> 

Ah, I misunderstood the ignoring delete statements part, thanks for clarifying. I guess we should let the user have some options to choose from here, some users prefer one over the other.
So, how about we provide a config option to the user? Something like, `-XClang -enable-experimental-dealloc` ?

lib/Differentiator/CladUtils.cpp
vaithak: Changed this to use the `getBuiltinID()` method and compare with defined macros like BImalloc, ...

lib/Differentiator/CladUtils.cpp
```diff
@@ -641,5 +641,19 @@
...
+        return true;
+      if (FD->getBuiltinID() == Builtin::BIcalloc)
+        return true;
+      if (FD->getBuiltinID() == Builtin::BIrealloc)
+        return true;
+      return false;
+    }
+
+    bool IsMemoryDeallocationFunction(const clang::FunctionDecl* FD) {
+      return FD->getBuiltinID() == Builtin::BIfree;
```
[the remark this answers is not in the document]
vgvassilev: Looks like this was introduced in llvm/llvm-project@7119f76c47797391fddcfcb60f29e70a2b424713 which is available on clang 15 onward...

vgvassilev:
> > We will be able to support the differentiation of functions that use `new` and `delete` if we ignore the `delete` statements. By ignoring `delete` statement, I mean that no statement will be generated in the adjoint code for a `delete` statement in the primal code. Differentiation will produce correct results even if there is a double `delete` or if there's `delete` across function boundaries. We basically have to choose between:
> > 
> > * Ignore `delete` statements and have correct derivatives at the cost of memory leaks in the generated code.
> > * Generate `delete` statements in the generated derivative code to avoid memory leaks. In this case, only the basic `delete` statement usage will work correctly. Any advance usage may result in Clad crashing, or an undefined behavior.
> 
> Ah, I misunderstood the ignoring delete statements part, thanks for clarifying. I guess we should let the user have some options to choose from here, some users would prefer one over the other depending on their applications. So, how about we provide a config option to the user? Something like, `-XClang -enable-experimental-dealloc` ?

I was wondering if we could emit the delete statements when the allocation is in the function we differentiate and produce a warning if the allocation/deallocation is across function boundaries?

vaithak:
> I was wondering if we could emit the delete statements when the allocation is in the function we differentiate and produce a warning if the allocation/deallocation is across function boundaries?

But, there are many cases where even if the delete(s) are in the same function, the current implementation will output wrong statements, for ex. - multiple new and delete combinations (as exemplified by Parth), delete(s) inside for/if conditions and maybe many more.

lib/Differentiator/ReverseModeVisitor.cpp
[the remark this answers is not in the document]
vaithak: fixed this 👍🏼

vaithak:
Turns out there was some issue when zero initializing a newly created array using `new double[2]()`, it took a good debugging session to figure out the root cause, but this is fixed now.

vaithak:
All tests are passing finally 🎉

test/ForwardMode/Pointer.C
```diff
@@ -110,16 +110,84 @@ double fn5(double i, double j) {
...
+// CHECK-NEXT:     *_d_p = _d_i;
+// CHECK-NEXT:     *p = i;
+// CHECK-NEXT:     {{(clad::)?}}ValueAndPushforward<void *, void *> _t1 = clad::custom_derivatives::calloc_pushforward(1, sizeof(T), 0, sizeof(T));
+// CHECK-NEXT:     T *_d_t = (T *)_t1.pushforward;
+// CHECK-NEXT:     T *t = (T *)_t1.value;
+// CHECK-NEXT:     _d_t->i = _d_i;
+// CHECK-NEXT:     t->i = i;
+// CHECK-NEXT:     double _d_res = *_d_p + _d_t->i;
+// CHECK-NEXT:     double res = *p + t->i;
+// CHECK-NEXT:     unsigned {{(int|long)}} _t2 = sizeof(double);
```
vgvassilev, line 164: I was wondering why we are not writing out `size_t` here?

vaithak: not sure about this

vgvassilev:
@parth-07 and I chatted a little and we'd like to move forward. Once we get some more experience then we should decide if we want to relax the delete operations in favor of more memory leaks.

Issue #748, opened 2024-02-03
Add a 32bit CI build


vgvassilev:
That will help us catch issues such as #747. Maybe using QEMU will work assuming that we can install the llvm binaries: https://docs.docker.com/build/ci/github-actions/multi-platform/

Krishna-13-cyber:
Hi @vgvassilev,
I will try to get this done. We will need a dockerized image(clad) for the workflow.

vgvassilev:
@Krishna-13-cyber, I suspect you can find on github already systems that use qemu to build in 32 bits. We probably need something like: https://github.com/flintlib/flint/pull/1564/

Krishna-13-cyber:
> @Krishna-13-cyber, I suspect you can find on github already systems that use qemu to build in 32 bits. We probably need something like: [flintlib/flint#1564](https://github.com/flintlib/flint/pull/1564)

Yes thanks, I will refer this.

PR #782, opened 2024-02-22
Add support for 32bit CI build

Krishna-13-cyber:
Addresses #748

### DIFF
--- a/.github/workflows/arch.yml
+++ b/.github/workflows/arch.yml
@@ -1,0 +1,53 @@
+---
+name: 'Arch'
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  architecture:
+    strategy:
+      matrix:
+        target: [x86]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v3
+      - name: "Setup latest Alpine Linux"
+        uses: jirutka/setup-alpine@v1
+        with:
+          arch:  ${{ matrix.target }}
+          branch: edge
+          packages: >
+            llvm-dev
+            clang-dev
+            clang-static
+            llvm17-static
+            llvm17-gtest
+            cmake
+            llvm
+            clang
+            make
+            git
+      - name: "Setup"
+        run: |
+          clang --version
+          make --version
+        shell: alpine.sh {0}
+
+      - name: "Compile library"
+        run: |
+          mkdir build && cd build
+          cmake -DLLVM_EXTERNAL_LIT="$(which lit)" ../../clad
+          make -j 8 check-clad
+        shell: alpine.sh {0}
+
+      - name: Setup tmate session
+        if: ${{ failure() && runner.debug }}
+        uses: mxschmitt/action-tmate@v3
+        # When debugging increase to a suitable value!
+        timeout-minutes: 30
--- a/test/Arrays/ArrayInputsReverseMode.C
+++ b/test/Arrays/ArrayInputsReverseMode.C
@@ -4,6 +4,7 @@
 // RUN: ./ArrayInputsReverseMode.out | FileCheck -check-prefix=CHECK-EXEC %s
 
 //CHECK-NOT: {{.*error|warning|note:.*}}
+// XFAIL: target={{i586.*}}
 
 #include "clad/Differentiator/Differentiator.h"
 
--- a/test/Arrays/Arrays.C
+++ b/test/Arrays/Arrays.C
@@ -4,6 +4,7 @@
 // RUN: ./Arrays.out | FileCheck -check-prefix=CHECK-EXEC %s
 
 //CHECK-NOT: {{.*error|warning|note:.*}}
+// XFAIL: target={{i586.*}}
 
 #include "clad/Differentiator/Differentiator.h"
 #include <cmath>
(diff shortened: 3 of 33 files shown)


vgvassilev:
Does that work for PRs? I don’t see a build.

Krishna-13-cyber:
> Does that work for PRs? I don’t see a build.

No then, I have enabled it now.

.github/workflows/platforms.yml
```diff
@@ -0,0 +1,118 @@
+---
+name: 'Platforms'
```
vgvassilev, line 2:
```suggestion
name: 'Arch'
```

.github/workflows/platforms.yml
```diff
@@ -0,0 +1,118 @@
...
+           - name: "Setup"
+             run: |
+               clang --version
+               make --version
+             shell: alpine.sh {0}
+
+           - name: "Compile library"
+             run: |
+               mkdir build && cd build
+               cmake -DLLVM_DIR=/usr/lib/ -DCMAKE_BUILD_TYPE=DEBUG -DLLVM_EXTERNAL_LIT="$(which lit)" ../../clad
```
vgvassilev, line 43:
```suggestion
               cmake -DLLVM_EXTERNAL_LIT="$(which lit)" ../../clad
```

.github/workflows/platforms.yml
```diff
@@ -0,0 +1,118 @@
...
```
vgvassilev, line 79:
```suggestion
               cmake -DLLVM_EXTERNAL_LIT="$(which lit)" ../../clad
```

.github/workflows/platforms.yml
```diff
@@ -0,0 +1,118 @@
...
```
vgvassilev, line 115:
```suggestion
               cmake -DLLVM_EXTERNAL_LIT="$(which lit)" ../../clad
```

Can we extract the step "Compile library" into one place as it seems to repeat across builds. The same for the other common steps. Maybe [this](https://stackoverflow.com/questions/59757355/reuse-portion-of-github-action-across-jobs) could help?

Krishna-13-cyber: Sorry for the delayed response. Yes, I will look into this and patch it asap.

vgvassilev: Can you take a look how the clad main workflow is organized, we want something similar for this PR.

Krishna-13-cyber: Yes, will do this.
Thanks!

Krishna-13-cyber: Hi, I have updated with respect to clad convention adhering to matrix strategy.
Thanks!

vgvassilev:
Nice! Now we need to figure out how to suppress the test failures. For this PR maybe it would be fine if we allow failures on armv7 and x86.

.github/workflows/arch.yml
```diff
@@ -0,0 +1,48 @@
...
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  architecture:
+    strategy:
+      matrix:
+        target: [x86, armv7, aarch64]
```
vgvassilev, line 15: Looks like `armv7` and `aarch64` are considerable slower. Can we disable them?

Krishna-13-cyber:
> Nice! Now we need to figure out how to suppress the test failures. For this PR maybe it would be fine if we allow failures on armv7 and x86.

Yes, we can solve the failures in a seperate PR.

vgvassilev:
> > Nice! Now we need to figure out how to suppress the test failures. For this PR maybe it would be fine if we allow failures on armv7 and x86.
> 
> Yes, we can solve the failures in a seperate PR.

Looks like this won't work. Maybe in this PR we should add to each failing test: `XFAIL: target={{(i686|i386).*}}`. Can you try this and then we can drop the non-functional `continue-on-error` thing.

.github/workflows/arch.yml
Krishna-13-cyber: Yes, they are slower.
We had a intention to target multiple architectures right to get hold of failures/bugs so why should we disable them (due to time constraints).
Can we try any other workaround to fit in those(slower ones) as well?

Krishna-13-cyber:
> > > Nice! Now we need to figure out how to suppress the test failures. For this PR maybe it would be fine if we allow failures on armv7 and x86.
> > 
> > 
> > Yes, we can solve the failures in a seperate PR.
> 
> Looks like this won't work. Maybe in this PR we should add to each failing test: `XFAIL: target={{(i686|i386).*}}`. Can you try this and then we can drop the non-functional `continue-on-error` thing.


I added this `continue-on-error` for all the parallel jobs to run independently as without this thing other jobs abort due to any failure on any of the jobs present.
Yes, I will try adding this `XFAIL: target={{(i686|i386).*}}` to failing tests.

Krishna-13-cyber:
> Looks like this won't work. Maybe in this PR we should add to each failing test: `XFAIL: target={{(i686|i386).*}}`. Can you try this and then we can drop the non-functional `continue-on-error` thing.

The failing tests get suppressed with ```XFAIL:*``` and not this `XFAIL: target={{(i686|i386).*}}`.
Should I use this?

vgvassilev:
> > Looks like this won't work. Maybe in this PR we should add to each failing test: `XFAIL: target={{(i686|i386).*}}`. Can you try this and then we can drop the non-functional `continue-on-error` thing.
> 
> The failing tests get suppressed with `XFAIL:*` and not this `XFAIL: target={{(i686|i386).*}}`. Should I use this?

No, that will suppress even working tests.

vgvassilev:
> target={{(i686|i386).*}}

@Krishna-13-cyber, you need `// XFAIL: target={{i586.*}}`

vgvassilev (approved):
Awesome! LGTM!

PR #783, opened 2024-02-22
Kokkos-aware Clad

kliegeois:
This PR provides the implementation of a first set of features for the automatic generation of gradients of Kokkos-based code.

The content of this PR supports reverse mode of Kokkos `parallel_for`, `deep_copy`, and view accesses. 

@vgvassilev @brian-kelley


parth-07 (changes requested):

vgvassilev:
@kliegeois, @brian-kelley, is there a point to try to rebase this PR or @gojakuch should continue picking random things from it as it is now?

kliegeois:
@vgvassilev I am not sure what is best.
If we rebase, would @gojakuch still create different PRs that refactor this PR?
If yes, the rebase is not necessarily useful although it will be easier to highlight the remaining pieces.

vgvassilev:
@Fermats-Last-Theorem do you think you can help bring this over the finish line?

Fermats-Last-Theorem:
Sounds good. I'll have a look through the PR.

Fermats-Last-Theorem:
@vgvassilev 
Alright. I have looked through the PR. I am not entirely sure how this PR is meant to be handled. Much of the later Kokkos work currently in master seems to be built on custom derivatives. There is some stuff which was in the scope of this PR which hasn't yet been added to master like gradient parallel_for, reduce and team level parallelism. Should I implement those as custom derivatives and open separate PRs for them as gojakuch did, so this one can be closed? Or should I refactor this PR with the remaining work?

vgvassilev:
I think separate PRs would make sense to me. In addition we should define some demos of end to end differentiated use cases for kokkos.

Fermats-Last-Theorem:
> I think separate PRs would make sense to me. In addition we should define some demos of end to end differentiated use cases for kokkos.

The reverse mode pullbacks (`parallel_for` and `parallel_reduce`) need active lambda capture support from #1859 before they can be implemented. 
I've also translated the end to end CUDA demos to kokkos and will open a PR for that soon.

PR #785, opened 2024-02-23
[cmake] Link Kokkos::kokkos

vgvassilev:


unittests/Kokkos/CMakeLists.txt
```diff
@@ -12,5 +12,6 @@ if (NOT (LLVM_REQUIRES_RTTI OR LLVM_ENABLE_RTTI))
   endif()
 endif()
 
+target_link_libraries(KokkosTests PRIVATE Kokkos::kokkos)
```
vgvassilev, line 15: Hi @kliegeois, that does not play very well with the Kokkos distribution on ubuntu that we use for out bots. I've noticed that and that's why I added the line after this one. Does it not work for you as it was?

kliegeois: Hi @vgvassilev ! Do you know the version number of the Kokkos you used? Without the line of this commit, I was unable to use the latest Kokkos on my Mac for the unit tests. Maybe this is due to a difference in Kokkos version.

vgvassilev: The Kokkos version that we install is `libtrilinos-kokkos-dev` for ubuntu 22.04 which seems to install `libtrilinos-kokkos-13.2`.

vgvassilev: @kliegeois, ping.

vgvassilev:
@gojakuch, you mentioned you have a similar problem where you needed to add locally `Kokkos::kokkos`. Was your build from source?

vgvassilev:
Looks like this is not needed for our bots. Let’s revisit if there is a need to.

PR #797, opened 2024-03-04
Fix extra lines generated when using clad::array or array_ref

### LINKED ISSUE #350: Fix extra lines generated when using clad::array.

vaithak:
fixes #350

### DIFF
--- a/include/clad/Differentiator/Array.h
+++ b/include/clad/Differentiator/Array.h
@@ -11,6 +11,13 @@
 namespace clad {
 template <typename T> class array_ref;
 
+// For MSVC, __attribute__((pure)) is not supported.
+#if defined(_MSC_VER)
+#define PUREFUNC
+#else
+#define PUREFUNC __attribute__((pure))
+#endif
+
 /// This class is not meant to be used by user. It is used by clad internally
 /// only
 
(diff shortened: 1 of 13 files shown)


include/clad/Differentiator/Array.h
```diff
@@ -102,16 +102,19 @@ template <typename T> class array {
   /// Returns the size of the underlying array
   CUDA_HOST_DEVICE std::size_t size() const { return m_size; }
   /// Returns the ptr of the underlying array
-  CUDA_HOST_DEVICE T* ptr() const { return m_arr; }
-  CUDA_HOST_DEVICE T*& ptr_ref() { return m_arr; }
+  CUDA_HOST_DEVICE T* ptr() const __attribute__((pure)) { return m_arr; }
```
vgvassilev, line 105: We probably should put that in a macro in case we produce derivative codes that get compiled with msvc.

vaithak: Ah, I forgot that; added it now 👍🏼

parth-07 (approved):

include/clad/Differentiator/Array.h
```diff
@@ -102,16 +109,16 @@ template <typename T> class array {
   /// Returns the size of the underlying array
   CUDA_HOST_DEVICE std::size_t size() const { return m_size; }
   /// Returns the ptr of the underlying array
-  CUDA_HOST_DEVICE T* ptr() const { return m_arr; }
-  CUDA_HOST_DEVICE T*& ptr_ref() { return m_arr; }
+  CUDA_HOST_DEVICE T* ptr() const PUREFUNC { return m_arr; }
```
vgvassilev, line 112:
```suggestion
  CUDA_HOST_DEVICE PUREFUNC T* ptr() const  { return m_arr; }
```

Can we move the attributes to go first?

vaithak: fixed 👍🏼

PR #802, opened 2024-03-05
Simplify pullback calls in the reverse mode

### LINKED ISSUE #801: Simplify pullback calls in the reverse mode AD
When differentiating call expressions, we introduce lots of useless declarations and assignments. We pass ``_grad`` variables to get the derivatives, and then assign these values to ``_r`` variables. This reassignment is pointless and often introduces unused variables.
For instance,
```
arr = _t1;
int _grad1 = 0;
sum_pullback(_t1, n, _r_d0, _d_arr, &_grad1);
double *_r0 = _d_arr;
int _r1 = _grad1;
* _d_n += _r1;
```
we could achieve the same functionality with
```
int _r0 = 0;
sum_pullback(_arr, n, _r_d0, _d_arr, &_r0);
* _d_n += _r0;
```

PetroZarytskyi:
This PR simplifies both the code of ``RMV::VisitCallExpr`` and the code generated by it. In particular, it replaces the ``_grad``/``_r`` variable pairs with single ``_r`` variables. The PR also removes some dead code from ``RMV::VisitCallExpr``, addresses clang-tidy warnings triggered by it, and improves its test coverage.
Fixes #801.

### DIFF
--- a/include/clad/Differentiator/CladUtils.h
+++ b/include/clad/Differentiator/CladUtils.h
@@ -174,11 +174,12 @@ namespace clad {
     /// otherwise returns false.
     bool HasAnyReferenceOrPointerArgument(const clang::FunctionDecl* FD);
 
-    /// Returns true if `T` is a reference, pointer or array type.
+    /// Returns true if `arg` is an argument passed by reference or is of
+    /// pointer/array type.
     ///
-    /// \note Please note that this function returns true for array types as
-    /// well.
-    bool IsReferenceOrPointerType(clang::QualType T);
+    /// \note Please note that this function returns false for temporary
+    /// expressions.
+    bool IsReferenceOrPointerArg(const clang::Expr* arg);
 
     /// Returns true if `T1` and `T2` have same cononical type; otherwise
     /// returns false.
--- a/include/clad/Differentiator/VisitorBase.h
+++ b/include/clad/Differentiator/VisitorBase.h
@@ -590,16 +590,20 @@ namespace clad {
     ///
     /// \param[in] targetFuncCall The function to get the derivative for.
     /// \param[in] retType The return type of the target call expression.
+    /// \param[in] dfdx The dfdx corresponding to this call expression.
     /// \param[in] numArgs The total number of 'args'.
-    /// \param[in] NumericalDiffMultiArg The built statements to add to block
-    /// later.
+    /// \param[in] PreCallStmts The built statements to add to block
+    /// before the call to the derived function.
+    /// \param[in] PostCallStmts The built statements to add to block
+    /// after the call to the derived function.
     /// \param[in] args All the arguments to the target function.
     /// \param[in] outputArgs The output gradient arguments.
     ///
     /// \returns The derivative function call.
     clang::Expr* GetMultiArgCentralDiffCall(
         clang::Expr* targetFuncCall, clang::QualType retType, unsigned numArgs,
-        llvm::SmallVectorImpl<clang::Stmt*>& NumericalDiffMultiArg,
+        clang::Expr* dfdx, llvm::SmallVectorImpl<clang::Stmt*>& PreCallStmts,
+        llvm::SmallVectorImpl<clang::Stmt*>& PostCallStmts,
         llvm::SmallVectorImpl<clang::Expr*>& args,
         llvm::SmallVectorImpl<clang::Expr*>& outputArgs);
     /// Emits diagnostic messages on differentiation (or lack thereof) for
--- a/include/clad/Differentiator/VisitorBase.h
+++ b/include/clad/Differentiator/VisitorBase.h
@@ -585,27 +585,6 @@ namespace clad {
     clang::Expr* GetSingleArgCentralDiffCall(
         clang::Expr* targetFuncCall, clang::Expr* targetArg, unsigned targetPos,
         unsigned numArgs, llvm::SmallVectorImpl<clang::Expr*>& args);
-    /// A function to get the multi-argument "central_difference"
-    /// call expression for the given arguments.
-    ///
-    /// \param[in] targetFuncCall The function to get the derivative for.
-    /// \param[in] retType The return type of the target call expression.
-    /// \param[in] dfdx The dfdx corresponding to this call expression.
-    /// \param[in] numArgs The total number of 'args'.
-    /// \param[in] PreCallStmts The built statements to add to block
-    /// before the call to the derived function.
-    /// \param[in] PostCallStmts The built statements to add to block
-    /// after the call to the derived function.
-    /// \param[in] args All the arguments to the target function.
-    /// \param[in] outputArgs The output gradient arguments.
-    ///
-    /// \returns The derivative function call.
-    clang::Expr* GetMultiArgCentralDiffCall(
-        clang::Expr* targetFuncCall, clang::QualType retType, unsigned numArgs,
-        clang::Expr* dfdx, llvm::SmallVectorImpl<clang::Stmt*>& PreCallStmts,
-        llvm::SmallVectorImpl<clang::Stmt*>& PostCallStmts,
-        llvm::SmallVectorImpl<clang::Expr*>& args,
-        llvm::SmallVectorImpl<clang::Expr*>& outputArgs);
     /// Emits diagnostic messages on differentiation (or lack thereof) for
     /// call expressions.
     ///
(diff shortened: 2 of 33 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1991,55 +1865,34 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       // Derivative was found.
       FunctionDecl* fnDecl = dyn_cast<CallExpr>(OverloadedDerivedFn)
                                  ->getDirectCallee();
+      // Put Result array declaration in the function body.
+      // Call the gradient, passing Result as the last Arg.
+      auto& block = getCurrentBlock(direction::reverse);
+      auto* it = std::begin(block) + insertionPoint;
```
vgvassilev, line 1871: We should spell out the types here.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1534,101 +1535,38 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
-        // argument by reference.
-        passByRef = false;
-      }
-      QualType argDiffType;
       // We do not need to create result arg for arguments passed by reference
       // because the derivatives of arguments passed by reference are directly
       // modified by the derived callee function.
-      if (passByRef) {
+      if (utils::isArrayOrPointerType(arg->getType()) ||
+      (arg->IgnoreImplicit()->isLValue() && utils::isTriviallyCastableArg(arg))) {
```
vgvassilev, line 1542: Can you write the code in C++ that this function is supposed to detect and based on that make a cast? That is, which types won't be trivially castable?

PetroZarytskyi: For instance, ``float`` is not trivially castable to ``double``. The thing that is important here is that we cannot cast ``float*`` to ``double*`` (at least not in the sense of preserving the underlying value). Therefore, we cannot pass the derivative of the argument to the pullback. We need a helper variable for this (in this case, of type ``double``).

vgvassilev: I guess you are talking about the cases where we have an out parameter (something passed by reference). Otherwise we can just do a c-style cast.

parth-07:
The work in this pull request seems good. Thank you for proactively trying to improve one of the most complicated components in the codebase. However, I think we may need to look at the issue and the solution more closely.

Please correct me if I am wrong anywhere. My autodiff skills are a bit rusty now.

Given a function:

```cpp
double do_something(double p_u, double p_v) { ... }
```

Considering the primal code is `r = do_something(u, v)`, then previously, we used to differentiate a call to `do_something` as follows:

```cpp
double _r_d0 = _d_res;
double _grad0 = 0;
double _grad1 = 0;
do_something_pullback(u, v, _r_d0, _grad0, _grad1);
double _r0 = _grad0;
double _r1 = _grad1;
*_d_u += _r0;
*_d_v += _r1
```

With your patch, the same call will be differentiated as:

```cpp
double _r_d0 = _d_res;
do_something_pullback(u, v, _r_d0, &*_d_u, &*_d_v);
```

The above two differentiated codes look similar and it appears that they should bring the same result mathematically. However, that's not the case. This is because of two reasons: 
 - a function call hides assignment operations: assigning values of arguments to parameters
 - Argument and parameters are different mathematical variables and therefore they should not share adjoint variables.

Let's look at what happens in a function call more closely.

**Primal code**

```cpp
r = do_something(u, v); // u and v are passed by value
```

This call can be made more clear by expanding it as follows:

```cpp
double p_u = u;
double p_v = v;

do_something(p_u, p_v); // p_u, and p_v are passed by reference
```

In the original case, `u` and `v` are passed by value and assignments to `p_u` and `p_v` happens implicitly. We cannot see it but it's very much there. Now, why does this matter? In the second case, can we use the adjoint variables `_d_u` and `_d_v` for representing the adjoints of `p_u` and `p_v`? No, of course not. We need different adjoint variables for `p_u` and `p_v` because `p_u` is not the same variable as `u` and `p_v` is not the same variable as `v`. Using the same logic, we cannot use the adjoints of `u` and `v` when passing grad variables to `do_something_pullback`. 

A practical example where this distinction matters:

```cpp
#include <iostream>
#define show(x) std::cout << #x << ": " << x <<"\n";

double reset(double u) {
    u = 0;
    return u;
}

    double res = u + v;
    res += reset(u);
    res += u;
    return res;
}

int main() {
    double u = 3, v = 5, du = 0, dv = 0;
    show(du);
    show(dv);
}
```

With this patch, the above code (incorrectly) outputs:

```
du: 1
dv: 1
```

The correct output is:

```
du: 2
dv: 1
```

parth-07 (changes requested):
Please see the above comment for details.

PetroZarytskyi:
@parth-07 
Hi Parth. Thank you for your comment. I agree with your reasoning, I got confused. I still think this PR can be fixed to actually simplify things without breaking anything. Even though the changes will not be that drastic. But it needs more work.

parth-07:
> I still think this PR can be fixed to actually simplify things without breaking anything.

That's great! It's important to simplify the call differentiation component. It has grown really complex over the years.

lib/Differentiator/ReverseModeVisitor.cpp

parth-07:
Hi @PetroZarytskyi 

Can you please add more details in the pull-request description / commit-message regarding how the pullback calls are being simplified?

PetroZarytskyi:
Hi @parth-07, I updated the PR message as well as the #801 issue.

include/clad/Differentiator/VisitorBase.h
```diff
@@ -590,16 +590,20 @@ namespace clad {
...
-    /// later.
+    /// \param[in] PreCallStmts The built statements to add to block
+    /// before the call to the derived function.
+    /// \param[in] PostCallStmts The built statements to add to block
+    /// after the call to the derived function.
     /// \param[in] args All the arguments to the target function.
     /// \param[in] outputArgs The output gradient arguments.
     ///
     /// \returns The derivative function call.
     clang::Expr* GetMultiArgCentralDiffCall(
```
parth-07, line 603: Unrelated to the changes in this PR: We should probably move this function to ReverseModeVisitor.

parth-07 (approved):
Looks good apart from the few minor comments.

include/clad/Differentiator/CladUtils.h
```diff
@@ -174,11 +174,12 @@ namespace clad {
...
-    /// Returns true if `T` is a reference, pointer or array type.
+    /// Returns true if `arg` is an argument passed by reference or is of
+    /// pointer/array type.
     ///
-    /// \note Please note that this function returns true for array types as
-    /// well.
-    bool IsReferenceOrPointerType(clang::QualType T);
+    /// \note Please note that this function returns false for temporary
+    /// expressions.
+    bool IsReferenceOrPointerArg(const clang::Expr* arg);
```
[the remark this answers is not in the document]
parth-07: Oh, the function suffix has been changed from `Type` to `Arg`.

lib/Differentiator/CladUtils.cpp
```diff
@@ -301,8 +301,12 @@ namespace clad {
       return false;
     }
 
-    bool IsReferenceOrPointerType(QualType T) {
-      return T->isReferenceType() || isArrayOrPointerType(T);
+    bool IsReferenceOrPointerArg(const Expr* arg) {
+      // The argument is passed by reference if it's passed as an L-value.
+      // However, if arg is a MaterializeTemporaryExpr, then arg is a
+      // temporary variable passed as a const reference.
+      bool isRefType = arg->isLValue() && !isa<MaterializeTemporaryExpr>(arg);
```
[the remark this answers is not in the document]
PetroZarytskyi: ``arg`` is supposed to be the argument expression passed to the function. If the function expects a ref-type argument, then ``arg`` is an l-value (usually a ``DeclRefExpr``). But when it expects a non-ref type argument, it is implicitly converted to an r-value. The AST of ``arg`` will look somewhat like this:
```
ImplicitCastExpr <l-value to r-value>
-DeclRefExpr
```
So arg will be an r-value. At least this is my understanding.

include/clad/Differentiator/VisitorBase.h
PetroZarytskyi: I fixed this.

Issue #746, opened 2024-02-02
Incorrect differentiation of loop conditions in reverse mode


parth-07:
Loop conditions are incorrectly differentiated in the reverse mode.

Reproducible example:

```cpp
#include "clad/Differentiator/Differentiator.h"
#include <iostream>
#define show(x) std::cout << #x << ": " << x << "\n";

double fn(double i, double j) {
    double res = 0;
    for (int c = 0; res = i * j; ++c) {
        if (c == 1)
            break;
    }
    return res;
}

int main() {
    auto fn_grad = clad::gradient(fn);
    double i = 3, j = 5;
    double di = 0, dj = 0;
    fn_grad.execute(i, j, &di, &dj);
    show(di);
    show(dj);
}
```

**Output:**
di: 0
dj: 0

**Expected output:**
di: 5
dj: 3

KronoCode:
Hello, I would like to take this issue as it seem to be a simple one for a first contribution 🙂

vgvassilev:
Please go ahead!

KronoCode:
I could not run Clad on my machine as it crashed when building the project.. As for the issue, I looked into the code and I had a few questions about the topic : 

- Is the derived function generated by the constructor of the CladFunction class?

- From my understanding, this issue is due to the generation of an incorrect derivate function OR MAYBE the function is not taken into account from the execute method and returns 0.

wymux:
Clang-17 fails to compile provided code:
```
DeclRefExpr for Decl not entered in LocalDeclMap?
UNREACHABLE executed at /var/tmp/paludis/build/dev-lang-clang-17.0.6/work/llvm-project/clang/lib/CodeGen/CGExpr.cpp:2867!
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
[...]

PR #806, opened 2024-03-07
Fix CUDA gradient script

### LINKED ISSUE #744: derivative value of 0 when running the CUDA test script
Hi @ioanaif !

I was running the GradientCuda.cu file locally, I see that when printing the gradients all gradients are 0.00. Furthermore, I see that when executed on the global kernel, the second printf statement is never executed, indicating that the function launch has an error and cannot be executed. This is confirmed when I place a standard cuda error message after the kernel launch <<<>>> 

```
if(cudaSuccess != cudaGetLastError()) {
  printf("Error after kernel launch");
}
```
An error is flagged. The kernel execution code is the same as the GradientCuda.cu test script as shown below. 

```

__global__ void compute(decltype(gauss_g) grad, double* d_x, double* d_p, int n, double* d_result) {

  printf("We have entered the Gloabla kernel");

  grad.execute(d_x, d_p, 2.0, n, d_result);

  printf("We have successfully performed a gradient execution on the GPU");

}

```

kchristin22:
Fixes #744.

As described [here](https://github.com/vgvassilev/clad/issues/744#issuecomment-1981495274), the gradient computation of the function of interest was moved inside the kernel. The results returned from the GPU were compared with the ones derived from the CPU execution to establish the correctness of the implementation.

### DIFF
--- a/test/CUDA/GradientCuda.cu
+++ b/test/CUDA/GradientCuda.cu
@@ -15,6 +15,7 @@
 // XFAIL: clang-15
 
 #include "clad/Differentiator/Differentiator.h"
+#include <vector>
 
 #define N 3
 
@@ -26,7 +27,6 @@ __device__ __host__ double gauss(double* x, double* p, double sigma, int dim) {
    return std::pow(2*M_PI, -dim/2.0) * std::pow(sigma, -0.5) * std::exp(t);
 }
 
-auto gauss_g = clad::gradient(gauss, "p");
 
 // CHECK:    void gauss_grad_1(double *x, double *p, double sigma, int dim, clad::array_ref<double> _d_p) __attribute__((device)) __attribute__((host)) {
 //CHECK-NEXT:     double _d_sigma = 0;
@@ -90,8 +90,9 @@ auto gauss_g = clad::gradient(gauss, "p");
 //CHECK-NEXT:     }
 //CHECK-NEXT: }
 
-__global__ void compute(decltype(gauss_g) grad, double* d_x, double* d_p, int n, double* d_result) {
-  grad.execute(d_x, d_p, 2.0, n, d_result);
+__global__ void compute(double* d_x, double* d_p, int n, double* d_result) {
+  auto gauss_g = clad::gradient(gauss, "p");
+  gauss_g.execute(d_x, d_p, 2.0, n, d_result);
 }
 
 int main(void) {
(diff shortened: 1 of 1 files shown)


test/CUDA/GradientCuda.cu
```diff
@@ -15,6 +15,7 @@
 // XFAIL: clang-15
```
vgvassilev, line 15: Can we understand why it fails for clang-15?

test/CUDA/GradientCuda.cu
```diff
@@ -109,14 +110,24 @@ int main(void) {
   cudaMemcpy(d_x, x, N * sizeof(double), cudaMemcpyHostToDevice);
   cudaMalloc(&d_p, N * sizeof(double));
   cudaMemcpy(d_p, p, N * sizeof(double), cudaMemcpyHostToDevice);
-  double *result, *d_result;
+  std::vector<double> result(N, 0);
```
vgvassilev, line 113: We probably do not need a `std::vector` here. Can we use a fixed size array or an `std::array`?

test/CUDA/GradientCuda.cu
kchristin22: My first guess would be that the local CUDA version is not compatible with clang-15. I will have a look.

kchristin22: Unfortunately, I haven't found a CUDA version old enough (prior to 11.5) and compatible with my system (Ubuntu 22.04) to test this with clang-15, but upon examining another issue I came across a similar failure of clang-15 and clang-14 (didn't test with clang-16). Relevant comment [here](https://github.com/vgvassilev/clad/issues/746#issuecomment-1984401702).

vgvassilev: This looks good to me. We probably should open a new issue and a separate PR to continue the `XFAIL: clang-15` discussion there.

vgvassilev (approved):
LGTM. See my comment.

test/CUDA/GradientCuda.cu
kchristin22: I can open one if you'd like.

I was also thinking that if the CI was expanded to include CUDA's set up, we could check the build errors in the PR (harder for development but for older versions of clang it would be beneficial for developers with the latest OS versions). This could be a separate issue.

vgvassilev: Sounds good.

PR #809, opened 2024-03-08
Add vector forward mode in docs

vaithak:
Moved the vector mode doc from the compiler-research webpage. (also, condensed it a lot).


docs/userDocs/source/user/CoreConcepts.rst
```diff
@@ -127,6 +127,52 @@ Substituting `s = z` we will get `sz` = 1
 Thus we don't need to run the program twice for each input. However, as mentioned
 above the only drawback is we need to re-run the program for a different output.
 
+
+Vectorized Forward Mode Automatic Differentiation
```
vgvassilev, line 131: Could we give examples how to use that within clad with the proper api, etc. Maybe that usage documentation needs to go in a different section.

vaithak: We already have a usage section here: https://github.com/vgvassilev/clad/blob/master/docs/userDocs/source/user/UsingVectorMode.rst.

vgvassilev: Okay, let's move forward then!

vgvassilev (approved):
LGTM!

PR #818, opened 2024-03-12
Differentiate for loop condition expression

### LINKED ISSUE #746: Incorrect differentiation of loop conditions in reverse mode

rohanjulka19:
Fixes #746 - Incorrect differentiation of loop conditions in reverse mode

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -61,6 +61,11 @@ namespace clad {
     /// Output variable of vector-valued function
     std::string outputArrayStr;
     std::vector<Stmts> m_LoopBlock;
+    /// This expression checks if the forward pass loop was terminted due to
+    /// break. It is used to determine whether to run the loop cond
+    /// differentiation. One additional time.
+    clang::Expr* m_CurrentBreakFlagExpr;
+
     unsigned outputArrayCursor = 0;
     unsigned numParams = 0;
     // FIXME: Should we make this an object instead of a pointer?
@@ -561,7 +566,6 @@ namespace clad {
 
       ReverseModeVisitor& m_RMV;
 
-      const bool m_IsInvokedBySwitchStmt = false;
       /// Builds and returns a literal expression of type `std::size_t` with
       /// `value` as value.
       clang::Expr* CreateSizeTLiteralExpr(std::size_t value);
@@ -576,6 +580,8 @@ namespace clad {
       clang::Expr* CreateCFTapePushExpr(std::size_t value);
 
     public:
+      bool m_IsInvokedBySwitchStmt = false;
+
       BreakContStmtHandler(ReverseModeVisitor& RMV, bool forSwitchStmt = false)
           : m_RMV(RMV), m_IsInvokedBySwitchStmt(forSwitchStmt) {}
 
@@ -597,6 +603,11 @@ namespace clad {
       /// expression, where `TapeRef` and `m_CurrentCounter` are replaced
       /// by their actual values respectively.
       clang::Stmt* CreateCFTapePushExprToCurrentCase();
+
+      /// Builds and return `clad::back(TapeRef) != m_CaseCounter`
+      /// expression, where `TapeRef` and `m_CaseCounter` are replaced
+      /// by their actual values respectively
+      clang::Expr* CreateCFTapeBackExprForCurrentCase();
 
       /// Does final modifications on forward and reverse blocks
       /// so that `break` and `continue` statements are handled
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -880,7 +880,6 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
 
     StmtDiff thenDiff = VisitBranch(If->getThen());
     StmtDiff elseDiff = VisitBranch(If->getElse());
-
     Stmt* Forward = clad_compat::IfStmt_Create(
         m_Context, noLoc, If->isConstexpr(), /*Init=*/nullptr, /*Var=*/nullptr,
         condDiffStored, noLoc, noLoc, thenDiff.getStmt(), noLoc,
@@ -992,6 +991,9 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
                Scope::ContinueScope);
     beginBlock(direction::reverse);
     LoopCounter loopCounter(*this);
+    llvm::SaveAndRestore<Expr*> SaveCurrentBreakFlagExpr(
+        m_CurrentBreakFlagExpr);
+    m_CurrentBreakFlagExpr = nullptr;
     const Stmt* init = FS->getInit();
     if (m_ExternalSource)
       m_ExternalSource->ActBeforeDifferentiatingLoopInitStmt();
@@ -1000,7 +1002,6 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     // Save the isInsideLoop value (we may be inside another loop).
     llvm::SaveAndRestore<bool> SaveIsInsideLoop(isInsideLoop);
     isInsideLoop = true;
-
     StmtDiff condVarRes;
     VarDecl* condVarClone = nullptr;
     if (FS->getConditionVariable()) {
@@ -1011,11 +1012,12 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       }
     }
 
-    // FIXME: for now we assume that cond has no differentiable effects,
     // but it is not generally true, e.g. for (...; (x = y); ...)...
-    StmtDiff cond;
+    StmtDiff condDiff;
+    StmtDiff condExprDiff;
     if (FS->getCond())
-      cond = Visit(FS->getCond());
+      std::tie(condDiff, condExprDiff) = DifferentiateSingleExpr(FS->getCond());
+
     const auto* IDRE = dyn_cast<DeclRefExpr>(FS->getInc());
     const Expr* inc = IDRE ? Visit(FS->getInc()).getExpr() : FS->getInc();
 
@@ -1063,7 +1065,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     /// with function globals and replace initializations with assignments.
     /// This is a temporary measure to avoid the bug that arises from
     /// overwriting local variables on different loop passes.
-    Expr* forwardCond = cond.getExpr();
+    Expr* forwardCond = condExprDiff.getExpr();
     /// If there is a declaration in the condition, `cond` will be
     /// a DeclRefExpr of the declared variable. There is no point in
     /// inserting it since condVarRes.getExpr() represents an assignment with
@@ -1073,8 +1075,36 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     if (condVarRes.getExpr() != nullptr && isa<Expr>(condVarRes.getExpr()))
       forwardCond = cast<Expr>(condVarRes.getExpr());
 
+    Stmt* breakStmt = m_Sema.ActOnBreakStmt(noLoc, getCurrentScope()).get();
+
+    /// This part adds the forward pass of loop condition stmt in the body
+    /// In this first loop condition diff stmts execute then loop condition
+    /// is checked if and loop is terminated.
+    beginBlock();
+    if (utils::unwrapIfSingleStmt(condDiff.getStmt()))
+      addToCurrentBlock(condDiff.getStmt());
+
+    Stmt* IfStmt = clad_compat::IfStmt_Create(
+        /*Ctx=*/m_Context, /*IL=*/noLoc, /*IsConstexpr=*/false,
+        /*Init=*/nullptr, /*Var=*/nullptr,
+        /*Cond=*/
+        BuildOp(clang::UnaryOperatorKind::UO_LNot, BuildParens(forwardCond)),
+        /*LPL=*/noLoc, /*RPL=*/noLoc,
+        /*Then=*/breakStmt,
+        /*EL=*/noLoc,
+        /*Else=*/nullptr);
+    addToCurrentBlock(IfStmt);
+
+    Stmt* forwardCondStmts = endBlock();
+    if (BodyDiff.getStmt()) {
+      BodyDiff.updateStmt(utils::PrependAndCreateCompoundStmt(
+          m_Context, BodyDiff.getStmt(), forwardCondStmts));
+    } else {
+      BodyDiff.updateStmt(utils::unwrapIfSingleStmt(forwardCondStmts));
+    }
+
     Stmt* Forward = new (m_Context)
-        ForStmt(m_Context, initResult.getStmt(), forwardCond, condVarClone,
+        ForStmt(m_Context, initResult.getStmt(), nullptr, condVarClone,
                 incResult, BodyDiff.getStmt(), noLoc, noLoc, noLoc);
 
     // Create a condition testing counter for being zero, and its decrement.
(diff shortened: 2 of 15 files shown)


vgvassilev:
Can you add tests?

parth-07 (changes requested):
The solution needs to be corrected. The for-loop `condition` should be differentiated in each iteration. 
The solution in this pull-request generates incorrect derivative for the example show below:

```cpp
#include <iostream>

    double res = 0;
    for (int c = 0; (res += i * j); ++c) {
        if (c == 2)
            break;
    }
    return res;
}

int main() {
    double u = 3, v = 5;
    double du = 0, dv = 0;
    show(du);
    show(dv);
}
```

**Output**
du: 5
dv: 3

**Expected Output**
du: 10
dv: 6

rohanjulka19:
Oh right ok, will fix this

rohanjulka19:
hi @parth-07  for c == 2 condition du and dv will be 15 and 9 right ? cause the loop will run 3 times.

parth-07:
>  hi @parth-07 for c == 2 condition du and dv will be 15 and 9 right ? cause the loop will run 3 times.

Oh, yes. You are right.

rohanjulka19:
hi @parth-07, I have added the differentiation of condition inside the loop body and also at the end of the loop. This is done because for the value where condition is false, the condition is exeucted but the loop body is not executed. So we need to add a extra differentiation step after loop body. This does not seem to be working for when we break at 1st or 2nd iteration and I am not able to understand why exactly. Can you just help me out here ?
```
    double _d_res = 0;
    unsigned long _t0;
    int _d_c = 0;
    int c = 0;
    clad::tape<double> _t1 = {};
    clad::tape<bool> _t3 = {};
    clad::tape<unsigned long> _t4 = {};
    double res = 0;
    _t0 = 0;
    for (c = 0; clad::push(_t1, res) , ((res = i * j) , c < 5); ++c) {
        _t0++;
        bool _t2 = c == 0;
        {
            if (_t2) {
                {
                    clad::push(_t4, 1UL);
                    break;
                }
            }
            clad::push(_t3, _t2);
        }
        clad::push(_t4, 2UL);
    }
    goto _label0;
  _label0:
    _d_res += 1;
    {
        for (; _t0; _t0--)
            switch (clad::pop(_t4)) {
              case 2UL:
                ;
                {
                    _d_res += 0;
                    res = clad::pop(_t1);
                    double _r_d0 = _d_res;
                    _d_res -= _r_d0;
                    * _d_i += _r_d0 * j;
                    * _d_j += i * _r_d0;
                }
                --c;
                if (clad::pop(_t3)) {
                  case 1UL:
                    ;
                }
            }
        {
            _d_res += 0;
            res = clad::pop(_t1);
            double _r_d0 = _d_res;
            _d_res -= _r_d0;
            * _d_i += _r_d0 * j;
            * _d_j += i * _r_d0;
        }
    }
}
```
**Output:**
du: 0
dv:0

**Expected:**
du: 5
dv:3

rohanjulka19:
1. I differentiated cond twice adding one inside the loop and one outside the loop
2. Now RHS of comma operator is differentiated and added to block first this is needed because if there are multiple condition expression the last one should decide the value to be differentiated.

parth-07 (changes requested):
> I differentiated cond twice adding one inside the loop and one outside the loop

This is correct.

The below example is still give wrong result:

```cpp
#include <iostream>

    double res = 0;
    for (int c = 0; true && (res += i * j); ++c) {
        if (c == 0)
            break;
    }
    return res;
}

int main() {
    double u = 3, v = 5;
    double du = 0, dv = 0;
    show(du);
    show(dv);
}
```

vgvassilev:
@rohanjulka19, can you take a look at the comment and possibly make progress here?

rohanjulka19:
Hi @vgvassilev , yeah sorry should have acknowledged here, I had seen the comment. I am currently working on adding && operator support in the VisitBinaryOperator function it's just that I am facing a issue that I haven't figured out how to solve. Basically for expression a && b && c it needs to have concentric if conditions i.e if(a) { df/da if(b) {df/db} } but since we start parsing from rightmost expression I am unable to add the right hand side expression's differentiation inside the IF block of the left expression as I am not sure how to pass it to the left hand side expression when visiting the LHS expression. So was just figuring that out plus I had university work so wasn't able to give it much time, I will try to push it today by end of day (GMT+1).

vgvassilev:
Is that work in the context of gsoc and if so, can you send me an email?

rohanjulka19:
@parth-07
 Have added && support in expressions  but right now it doesn't work with other binary operators correctly.
Also added one cleanup - now we are checking whether if condtion has body and only then adding the if stmt.
Is this code fine ? If yes I will add the support for OR also in a similar fashion and fix the issue of using && with other binary operators

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3528,6 +3565,24 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       }
     }
 
+    if (condDiff && unwrapIfSingleStmt(condDiff->getStmt_dx())) {
+      if (bodyDiff.getStmt_dx()) {
+        bodyDiff.updateStmtDx(utils::PrependAndCreateCompoundStmt(
+            m_Context, bodyDiff.getStmt_dx(), unwrapIfSingleStmt(condDiff->getStmt_dx())));
+      } else {
+        bodyDiff.updateStmtDx(unwrapIfSingleStmt(condDiff->getStmt_dx()));
```
vgvassilev, line 3573: This branch seems not tested. Can you add a test case?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3528,6 +3565,24 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+      }
+    }
+    
+    if (condDiff && unwrapIfSingleStmt(condDiff->getStmt())) {
+      if (bodyDiff.getStmt()) {
+        bodyDiff.updateStmt(utils::PrependAndCreateCompoundStmt(
+            m_Context, bodyDiff.getStmt(), unwrapIfSingleStmt(condDiff->getStmt())));
+      } else {
+        bodyDiff.updateStmt(unwrapIfSingleStmt(condDiff->getStmt()));
```
vgvassilev, line 3582: Likewise.

rohanjulka19: have added test case to check both

rohanjulka19: ok found a small bug will fix it in sometimes

vgvassilev: According to codecov, no test takes this branch.

rohanjulka19: ok is there a way to know locally to know if a line is covered or not ? should I just add asserts to know if it reached there ?

vgvassilev: Maybe adding an printf or assert locally is one of the best options.

parth-07 (changes requested):
Thank you for persistently working on this. It was very nice to see the exhaustive set of tests that you have added.

The pull request still needs some changes. Some modifications relate to the differentiation of `if` statements. Are those changes required to fix this issue? If not, then we should remove them.

Additionally, the behaviour of the derived loop does not seem completely correct for the `continue` statement. Loop-condition and iterator-expression derivatives are skipped in the reverse-pass when the `continue` statement is hit in the forward-pass. Loop condition and iterator-expression should only be skipped when `break` statement is hit in the forward-pass. An example where this results in incorrect derivative:

```cpp
#include <iostream>

    double res = 0;
    for (int i = 0; res += u; i++) {
        if (i == 0)
            continue;
        if (i == 1)
            break;
    }
    return res;
}

int main() {
    double u = 3, v = 5;
    double du, dv;
    du = dv = 0;
    d_fn.execute(u, v, &du, &dv);
    show(du);
    show(dv);
}
```

**Output:**
du: 1
dv: 0

**Expected Output**
du: 2
dv: 0

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -790,7 +790,11 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     // the if statement.
     Expr* PushCond = nullptr;
     Expr* PopCond = nullptr;
-    auto condExpr = Visit(cond.getExpr());
+
+    StmtDiff condExpr;
```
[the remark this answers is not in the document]
rohanjulka19: The current approach I am using for implementing the && operator is by creating it's equivalent if condition and then visiting the if conidition hence forwarding all the responsibility of generating the forward and reverse pass code of the expression on the VisitIfStmt code. So these changes are required for correct differentiation of && expression.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1043,9 +1042,16 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
     // but it is not generally true, e.g. for (...; (x = y); ...)...
-    StmtDiff cond;
-    if (FS->getCond())
-      cond = Visit(FS->getCond());
+    StmtDiff condDiff;
+    StmtDiff condExprDiff;
+    StmtDiff condDiffOuter;
+    StmtDiff condExprDiffOuter;
+    if (FS->getCond()) {
+      std::tie(condDiff, condExprDiff) = DifferentiateSingleExpr(FS->getCond());
```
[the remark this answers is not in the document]
rohanjulka19: Yeah I guess that can be used will change this

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2452,10 +2460,21 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     } else if (opCode == BO_Comma) {
       auto* zero =
           ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
-      Ldiff = Visit(L, zero);
       Rdiff = Visit(R, dfdx());
+      Ldiff = Visit(L, zero);
       valueForRevPass = Ldiff.getRevSweepAsExpr();
       ResultRef = Ldiff.getExpr();
+    } else if (opCode == BO_LAnd) {
```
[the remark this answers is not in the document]
rohanjulka19: yeah will do that will move the if stmt related changes also to that PR then ?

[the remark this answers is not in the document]
parth-07: Yes, sounds good.

vgvassilev:
@rohanjulka19, @parth-07 what is the fate of this PR?

rohanjulka19:
hi @vgvassilev , I have some other important work that I had anticipated to be finished by now but is still under progress. So I guess I will be able to work on this and move this to completion in June only.

rohanjulka19:
Hi,
I am trying to get different conditions fixed but facing a issue here for this code 

```
  double res = 0;
  for (int c =0; (res  += i * j), c<3; ++c) {
    if (c == 1) {
      res = i * j ;
      continue;
    }
  }
  return res;
}
```

The output for 
u: 3    v: 5

should be 
 du :15  dv:9 
 
 whereas the it is returning 10 and 6 
 
 the generated code is this 
 
```
void fn_grad(double i, double j, double *_d_i, double *_d_j) {
    double _d_res = 0;
    double _d_x = 0;
    unsigned long _t0;
    int _d_c = 0;
    int c = 0;
    clad::tape<double> _t1 = {};
    clad::tape<double> _t2 = {};
    clad::tape<bool> _cond0 = {};
    clad::tape<double> _t3 = {};
    clad::tape<unsigned long> _t4 = {};
    double res = 0;
    double x = 0;
    _t0 = 0UL;
    clad::push(_t2, res);
    for (c = 0; (res += i * j) , c < 3; ++c) {
        clad::push(_t1, res);
        _t0++;
        {
            clad::push(_cond0, c == 1);
            if (clad::back(_cond0)) {
                clad::push(_t3, res);
                res = i * j;
                {
                    clad::push(_t4, 1UL);
                    continue;
                }
            }
        }
        clad::push(_t4, 2UL);
    }
    goto _label0;
  _label0:
    _d_res += 1;
    {
        for (; _t0; _t0--) {
            switch (clad::pop(_t4)) {
              case 2UL:
                ;
                --c;
                {
                    if (clad::back(_cond0)) {
                      case 1UL:
                        ;
                        {
                            res = clad::pop(_t3);
                            double _r_d2 = _d_res;
                            _d_res -= _r_d2;
                            *_d_i += _r_d2 * j;
                            *_d_j += i * _r_d2;
                        }
                    }
                    clad::pop(_cond0);
                }
            }
            {
                _d_res += 0;
                res = clad::pop(_t1);
                double _r_d0 = _d_res;
                *_d_i += _r_d0 * j;
                *_d_j += i * _r_d0;
            }
        }
        {
            _d_res += 0;
            res = clad::pop(_t2);
            double _r_d1 = _d_res;
            *_d_i += _r_d1 * j;
            *_d_j += i * _r_d1;
        }
    }
}

```

Now in the code of continue block in the swtich stmt 

```
  {
      res = clad::pop(_t3);
      double _r_d2 = _d_res;
      _d_res -= _r_d2;
      *_d_i += _r_d2 * j;
      *_d_j += i * _r_d2;
  }
```

we are removing the old value of _d_res which I am not able to understand why we are doing exactly (other than for case like res = res as mentioned in the comment in the code [here](https://github.com/vgvassilev/clad/blob/master/lib/Differentiator/ReverseModeVisitor.cpp#L2268)) . Additionaly, this might sound stupid but shouldn't  _d_i and _d_j be resetted also, as if we write something like :- res = i * j then doesn't this mean that res's value now only depends on the current equation i.e i * j and partial derivative of i or j should be calculated just based on that only.

vgvassilev:
Can you rebase the pull request first?

@parth-07, what is your take?

rohanjulka19:
ok turns out just the order in which they are executed had to be changed now it seems to be working. 

Separated the related to IF stmt changes and && support to another PR - https://github.com/vgvassilev/clad/pull/923

> Can we use `DifferentiateSingleStmt` instead of `DifferentiateSingleExpr` here? `DifferentiateSingleExpr` is more complicated, and I prefer to use it only when it is strictly required.
 
 We actually need to use DifferentiateSingleStmt here because we need the expression diff as well as the stmts added to the reverse and forward block 

For handling break stmt - Added a variable in break stmt which is set to 1 in forward pass and in reverse pass loop condition code is executed one less time.

parth-07 (commented):
Thank you for the pull-request. Overall, the changes seem good. I will need some more time to review the pull-request in detail.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -61,6 +61,8 @@ namespace clad {
     /// Output variable of vector-valued function
     std::string outputArrayStr;
     std::vector<Stmts> m_LoopBlock;
+    clang::Expr* m_CurrentBreakFlagExpr;
```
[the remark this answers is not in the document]
parth-07: Please add documentation here regarding the use of `m_CurrentBreakFlagExpr`.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -561,9 +563,6 @@ namespace clad {
 
       ReverseModeVisitor& m_RMV;
 
-      const bool m_IsInvokedBySwitchStmt = false;
-      /// Builds and returns a literal expression of type `std::size_t` with
```
parth-07, line 565: Did you remove this documentation for the below function by mistake? If so, can you please add it back?

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -598,6 +599,8 @@ namespace clad {
       /// by their actual values respectively.
       clang::Stmt* CreateCFTapePushExprToCurrentCase();
 
+      clang::Expr* CreateCFTapeBackExprForCurrentCase();
```
parth-07, line 602: Please add function documentation comments.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2391,8 +2450,8 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     } else if (opCode == BO_Comma) {
       auto* zero =
           ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
-      Ldiff = Visit(L, zero);
       Rdiff = Visit(R, dfdx());
+      Ldiff = Visit(L, zero);
```
parth-07, line 2454: Does the order of differentiation of `Ldiff` and `Rdiff` matter for this functionality added by this pull-request?

parth-07 (changes requested):

lib/Differentiator/ReverseModeVisitor.cpp
rohanjulka19: Hi yeah there is a case where it was causing issue which is

```
 for (int c = 0; (res = i * j), (res=2*i*j), c < 3; ++c) {}
```

In this loop condition the right one should get executed first, in reverse pass otherwise it will return di - 5, dj-3 but it should return di-10, dj-6.

parth-07 (commented):
The changes looks good. Can you please update the 'test/CUDA/GradientCuda.cu' test?

rohanjulka19:
> The changes looks good. Can you please update the 'test/CUDA/GradientCuda.cu' test?

I have updated the Cuda test now it is passing. Can't seem to figure out why a test case is not working on x86.

vgvassilev:
@MihailMihov, this failure seems to be similar to what you've been hunting down?

MihailMihov:
> @MihailMihov, this failure seems to be similar to what you've been hunting down?

Seems like it's the same thing, in fact I'm not sure if one of the numbers isn't the exact same that I got somewhere. I'll see whether it happens on my machine too, like the other issue.

MihailMihov:
Well I checked out your branch, but it seems like the issue is different. I had a really stupid error (uninitialized variable in my test). I ran your test with LLVM's MemorySanitizer and there is an uninitialized value here too, but it seems to be coming from one of the tapes.
```
==1626607==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x55a8d77c8230 in main /home/mihail/dev/clad/test/Gradient/Loops.C:2666:3
    #1 0x7f3ac4a5f2df  (/usr/lib64/libc.so.6+0x262df)
    #2 0x7f3ac4a5f398 in __libc_start_main (/usr/lib64/libc.so.6+0x26398)
    #3 0x55a8d772b374 in _start (/home/mihail/dev/clad/build/ReverseLoops.out+0x32374)

  Uninitialized value was stored to memory at
    #0 0x55a8d780e329 in fn30_grad(double, double, double*, double*) /home/mihail/dev/clad/test/Gradient/../../include/clad/Differentiator/Tape.h
    #1 0x55a8d77c7c8f in fn30_grad(double, double, void*, void*) /home/mihail/dev/clad/test/Gradient/Loops.C:1:1
    #2 0x55a8d77c7c8f in _ZN4clad25execute_with_default_argsILb1EJEPFvddPvS1_EJiiRA5_dPdEJddS1_S1_ETnNSt9enable_ifIXT_EbE4typeELb1EEENS_15function_traitsIT1_E11return_typeENS_4listIJDpT0_EEESB_NSE_IJDpT3_EEEDpOT2_ /home/mihail/dev/clad/test/Gradient/../../include/clad/Differentiator/Differentiator.h:123:12
    #3 0x55a8d77c7c8f in void clad::CladFunction<void (*)(double, double, void*, void*), clad::NoObject, true>::execute_helper<void (*)(double, double, void*, void*), int, int, double (&) [5], double*>(void (*)(double, double, void*, void*), int&&, int&&, double (&) [5], double*&&) /home/mihail/dev/clad/test/Gradient/../../include/clad/Differentiator/Differentiator.h:267:14
    #4 0x55a8d77c7c8f in std::enable_if<!std::is_same<void (*)(double, double, void*, void*), clad::NoFunction*>::value, void>::type clad::CladFunction<void (*)(double, double, void*, void*), clad::NoObject, true>::execute<int, int, double (&) [5], double*, void (*)(double, double, void*, void*)>(int&&, int&&, double (&) [5], double*&&) /home/mihail/dev/clad/test/Gradient/../../include/clad/Differentiator/Differentiator.h:218:14
    #5 0x55a8d77c7c8f in main /home/mihail/dev/clad/test/Gradient/Loops.C:2666:3
    #6 0x7f3ac4a5f2df  (/usr/lib64/libc.so.6+0x262df)

  Uninitialized value was created
    <empty stack>

SUMMARY: MemorySanitizer: use-of-uninitialized-value /home/mihail/dev/clad/test/Gradient/Loops.C:2666:3 in main
Exiting
```
In order to see a more detailed stack trace I'd need to compile clad instrumented with MSan, but I couldn't figure it out today, I'll try to get you a better stack trace tomorrow.

rohanjulka19:
ok thanks to @MihailMihov, turns out there was a variable which wasn't initialised and was causing the test to fail. Now that it is initialised to zero, the test is passing. The variable is _d_cond which is added to ensure that while differentiating expression containing && operator, the expression which is then assigned to the cond variable gets differentiated.
```
cond = (c==0) && (res += i *j)
``` 

@MihailMihov Not sure if this is related to the warning you caught.

parth-07 (approved):

PR #819, opened 2024-03-13
[cmake] Provide the `DISABLE_LLVM_LINK_LLVM_DYLIB` flag for unittests

gojakuch:
This tiny change prevents an undesired linkage.

### DIFF
--- a/unittests/CMakeLists.txt
+++ b/unittests/CMakeLists.txt
@@ -12,7 +12,7 @@ endif()
 # Produces a binary named 'basename(test_dirname)'.
 function(add_clad_unittest test_dirname)
 
-  add_unittest(CladUnitTests ${test_dirname} ${ARGN})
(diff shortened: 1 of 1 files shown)


unittests/Kokkos/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 add_clad_unittest(KokkosTests
   parallel_for.cpp
   main.cpp
+  DISABLE_LLVM_LINK_LLVM_DYLIB
```
vgvassilev, line 4: Can we move this in the caller. Other unittests should be also affected by the same problem we are fixing here. In addition we could improve the commit message.

vgvassilev: https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/

vgvassilev (approved):
LGTM!

Issue #734, opened 2024-01-23
Incorrect gradient when a reference variable is initialized with a pointer dereference


parth-07:
Clad generates an incorrect gradient code when a reference variable is initialized with a pointer deference.

Reproducible example:

```cpp
#include "clad/Differentiator/Differentiator.h"
#include <iostream>
#define show(x) std::cout << #x << ": " << x << "\n";

double fn(double u, double v) {
    double *p = &v;
    double &ref = *p;
    ref += u;
    double res = u + v;
    return res;
}

int main() {
    auto fn_grad = clad::gradient(fn);
    double u = 3, v = 5;
    double du = 0, dv = 0;
    fn_grad.execute(u, v, &du, &dv);
    show(du);
    show(dv);
}
```

**Actual output:**
du: 1
dv: 1

**Expected output:**
du: 2
dv: 1

**Root cause:**

```cpp
double &ref = *p
```
`ref` and `*p` should share the same adjoint variable as reference is just an alias name for the same value / memory location. However, currently, `double &ref` has it's own separate adjoint variable `double _d_ref = 0`.
This case should be handled similarly to how we handle a function returning a reference/pointer. Therefore, `double &ref = *p` should be transformed as follows in the forward pass:

```cpp
double *_d_ref = nullptr;
// ...
double &ref = *p;
_d_ref =  _d_p;
```

saras1212:
I can work on this!

kaushal-malpure:
Hey @parth-07,
Can you please guide me through this issue?

kaushal-malpure:
Hi @parth-07 ,
I would like to contribute to this issue, can you please guide me through this?

kaushal-malpure:
Hi @parth-07 ,
According to my understanding the reference variable doesn't share the same adjoint as the pointer resulting in wrong gradient. For this we could change the gradient function to handle the case where a reference variable is initialized with pointer dereference where we can create a adjoint variable corresponding to reference variable and set it's value equal to the adjoint of pointer.
I attempted to solve this by changing the gradient function in `clad/differentiator/differentiator.h` accordingly, but I'm confused in how do I exactly access the adjoint of the pointer so that I can set it equal to the new adjoint variable created. Can you please help me in this?

PetroZarytskyi:
Fixed by #1230.

PR #820, opened 2024-03-13
Updated to get correct gradient when reference variable is initialized with pointer dereference

kaushal-malpure:
@parth-07 
Referring to #734 
According to my understanding when a reference variable is initialized with pointer dereference, the `clad::gradient` function creates a new adjoint variable for the given reference variable whereas it should share the same adjoint variable as the pointer. To solve this issue we can replace the reference variable with the original pointer. This might be a potential solution. I couldn't verify this because my system gives an error when I try to build the program.

### DIFF
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -372,20 +372,63 @@ inline CUDA_HOST_DEVICE unsigned int GetLength(const char* code) {
           derivedFn /* will be replaced by gradient*/, code);
   }
 
+  /// To check if reference is initialized with pointer dereference
+  template <typename T>
+  constexpr bool has_reference_initialized_with_pointer_dereference =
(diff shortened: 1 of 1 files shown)


PetroZarytskyi:
Hi @kaushal-malpure. Thank you for your interest in Clad. I think the solution to this issue should not be in the Clad's front end (like the ``gradient`` function) but rather in the back end. Perhaps the changes should be introduced to ``ReverseModeVisitor.cpp`` (this is where the gradient code is generated). I would suggest getting familiar with how Clang AST is generated there. You can start with the main function ``Derive``. And yes, you need to be able to build Clad. Being able to take a look at the gradient code you produce is crucial for understanding and testing your changes.

guitargeek:
I ran the repro: `double& r = *p; return r*r;`, gives `dx=6.000000`, correct. Issue #734 closed 2025-03-05.

Issue #805, opened 2024-03-07
Add support for clang18


vgvassilev:

mcbarton:
@vgvassilev I will look at this once I am done solving the issue of osx arm builds in the ci.

PR #824, opened 2024-03-15
[clang] Add clang 18 support

mcbarton:
This PR will add support for clang 18
Fixes https://github.com/vgvassilev/clad/issues/805


lib/Differentiator/CladUtils.cpp
```diff
@@ -269,12 +274,21 @@ namespace clad {
...
       QualType
+      #if CLANG_VERSION_MAJOR < 18
           StrTy = clad_compat::getConstantArrayType(C, CharTyConst,
                                                     llvm::APInt(/*numBits=*/32,
                                                                 str.size() + 1),
                                                     /*SizeExpr=*/nullptr,
                                                     /*ASM=*/ArrayType::Normal,
                                                     /*IndexTypeQuals*/ 0);
+      #else
+          StrTy = clad_compat::getConstantArrayType(C, CharTyConst,
```
vgvassilev, line 285: We should probably make this change in `clad_compat::getConstantArrayType`.

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -36,13 +36,23 @@ namespace clad {
...
+      QualType StrTy =
+          clad_compat::getConstantArrayType(C, CharTyConst,
                                           llvm::APInt(/*numBits=*/32,
                                                       str.size() + 1),
                                           /*SizeExpr=*/nullptr,
                                           /*ASM=*/ArrayType::Normal,
                                           /*IndexTypeQuals*/ 0);
+    #else
+      QualType StrTy =
+          clad_compat::getConstantArrayType(C, CharTyConst,
```
vgvassilev, line 49: If we address the previous comment, this change is not going to be needed.

tools/ClangPlugin.cpp
```diff
@@ -79,10 +79,20 @@ namespace clad {
...
       for (llvm::StringRef P : Opts.Plugins)
-        if (llvm::sys::path::stem(P).endswith("clad")) {
-          CladSoPath = P;
-          break;
-        }
+        //FIXME: Define variable based on
+        //llvm::sys::path::stem(P).endswith("clad")/llvm::sys::path::stem(P).ends_with("clad")
+        //to write more compactly before PR review.
+        #if CLANG_VERSION_MAJOR < 18
+          if (llvm::sys::path::stem(P).endswith("clad")) {
```
vgvassilev, line 86: Instead of making this a macro-based code, can we replace the implementation with std::string. Eg.
std::string PEnd = llvm::sys::path::stem(P);
...

CMakeLists.txt
```diff
@@ -39,12 +39,12 @@ if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
...
   ## Define clad supported version of clang and llvm
 
-  set(CLANG_MIN_SUPPORTED 7.0)
-  set(CLANG_MAX_SUPPORTED "17.0.x")
-  set(CLANG_VERSION_UPPER_BOUND 17.1.0)
-  set(LLVM_MIN_SUPPORTED 7.0)
-  set(LLVM_MAX_SUPPORTED "17.0.x")
-  set(LLVM_VERSION_UPPER_BOUND 17.1.0)
+  set(CLANG_MIN_SUPPORTED 8.0)
+  set(CLANG_MAX_SUPPORTED "19.0.0")
```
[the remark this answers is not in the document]
mcbarton: I didn't know what it would be for CLANG_MAX_SUPPORTED using 18 . Would it be 18.x.x ?

lib/Differentiator/StmtClone.cpp
```diff
@@ -235,7 +234,12 @@ DEFINE_CLONE_EXPR(CXXScalarValueInitExpr,
 DEFINE_CLONE_EXPR(ExtVectorElementExpr, (Node->getType(), Node->getValueKind(), Clone(Node->getBase()), Node->getAccessor(), Node->getAccessorLoc()))
 DEFINE_CLONE_EXPR(CXXBoolLiteralExpr, (Node->getValue(), Node->getType(), Node->getSourceRange().getBegin()))
 DEFINE_CLONE_EXPR(CXXNullPtrLiteralExpr, (Node->getType(), Node->getSourceRange().getBegin()))
-DEFINE_CLONE_EXPR(CXXThisExpr, (Node->getSourceRange().getBegin(), Node->getType(), Node->isImplicit()))
+#if CLANG_VERSION_MAJOR < 18
+   DEFINE_CLONE_EXPR(CXXThisExpr, (Node->getSourceRange().getBegin(), Node->getType(), Node->isImplicit()))
+#else
+   DEFINE_CLONE_EXPR(CXXThisExpr, (Node->getSourceRange().getBegin(), Node->getType(), Node->isImplicit(),Node->getValueKind()))
+#endif
```
mcbarton, line 241: @vgvassilev any suggestions on what the solution the issue with CXXThisExpr might be? I thought that an extra argument had been added in clang-18, after it complained about having no CXXThisExpr with 3 arguments.  When I tried this solution though, the CI point out that the CXXThisExpr I had chosen was implicitly declared private.

vgvassilev: It probably got a `CXXThisExpr::Create` method. We should probably use `DEFINE_CREATE_EXPR` for it when the clang version is greater than 17?

lib/Differentiator/StmtClone.cpp
```diff
@@ -235,7 +234,12 @@ DEFINE_CLONE_EXPR(CXXScalarValueInitExpr,
 DEFINE_CLONE_EXPR(ExtVectorElementExpr, (Node->getType(), Node->getValueKind(), Clone(Node->getBase()), Node->getAccessor(), Node->getAccessorLoc()))
 DEFINE_CLONE_EXPR(CXXBoolLiteralExpr, (Node->getValue(), Node->getType(), Node->getSourceRange().getBegin()))
 DEFINE_CLONE_EXPR(CXXNullPtrLiteralExpr, (Node->getType(), Node->getSourceRange().getBegin()))
-DEFINE_CLONE_EXPR(CXXThisExpr, (Node->getSourceRange().getBegin(), Node->getType(), Node->isImplicit()))
+#if CLANG_VERSION_MAJOR < 18
+   DEFINE_CLONE_EXPR(CXXThisExpr, (Node->getSourceRange().getBegin(), Node->getType(), Node->isImplicit()))
+#else
+   DEFINE_CLONE_EXPR(CXXThisExpr, (Node->getSourceRange().getBegin(), Node->getType(), Node->isImplicit(),Node->getValueKind()))
```
vgvassilev, line 240:
```suggestion
   DEFINE_CREATE_EXPR(CXXThisExpr, (Node->getSourceRange().getBegin(), Node->getType(), Node->isImplicit(),Node->getValueKind()))
```

lib/Differentiator/StmtClone.cpp
```diff
@@ -235,7 +234,12 @@ DEFINE_CLONE_EXPR(CXXScalarValueInitExpr,
 DEFINE_CLONE_EXPR(ExtVectorElementExpr, (Node->getType(), Node->getValueKind(), Clone(Node->getBase()), Node->getAccessor(), Node->getAccessorLoc()))
 DEFINE_CLONE_EXPR(CXXBoolLiteralExpr, (Node->getValue(), Node->getType(), Node->getSourceRange().getBegin()))
 DEFINE_CLONE_EXPR(CXXNullPtrLiteralExpr, (Node->getType(), Node->getSourceRange().getBegin()))
-DEFINE_CLONE_EXPR(CXXThisExpr, (Node->getSourceRange().getBegin(), Node->getType(), Node->isImplicit()))
+#if CLANG_VERSION_MAJOR < 18
+   DEFINE_CLONE_EXPR(CXXThisExpr, (Node->getSourceRange().getBegin(), Node->getType(), Node->isImplicit()))
+#else
+   DEFINE_CREATE_EXPR(CXXThisExpr, (Node->getSourceRange().getBegin(), Node->getType(), Node->isImplicit(),Node->getValueKind()))
```
vgvassilev, line 240:
```suggestion
   DEFINE_CREATE_EXPR(CXXThisExpr, (Ctx, Node->getSourceRange().getBegin(), Node->getType(), Node->isImplicit()))
```

mcbarton:
@vgvassilev It is now building clad with clang-18 but it doesn't pass all the tests. It says that it is expected that 2 will fail, but 5 tests fail. Which 2 are expected to fail?

include/clad/Differentiator/Compatibility.h
```diff
@@ -71,29 +81,21 @@ NamespaceDecl_Create(ASTContext& C, DeclContext* DC, bool Inline,
 }
 #endif
 
-// Clang 7 add one extra param in UnaryOperator constructor.
-
-#if CLANG_VERSION_MAJOR >= 7
 #define CLAD_COMPAT_CLANG7_UnaryOperator_ExtraParams , Node->canOverflow()
```
alexander-penev, line 84: Because this substitution is already unconditional, i.e. for all supported versions, not just post 7, it is good to find and replace this in all files where it is used.
We must find "CLAD_COMPAT_CLANG7_UnaryOperator_ExtraParams" everywhere in the Clad files and replace it with ", Node->canOverflow()".

include/clad/Differentiator/Compatibility.h
```diff
@@ -71,29 +81,21 @@ NamespaceDecl_Create(ASTContext& C, DeclContext* DC, bool Inline,
...
-   return E->EvaluateAsInt(IntValue, Ctx, AllowSideEffects);
-#elif CLANG_VERSION_MAJOR >= 8
    Expr::EvalResult res;
    if (E->EvaluateAsInt(res, Ctx, AllowSideEffects)) {
      IntValue = res.Val.getInt();
      return true;
    }
    return false;
-#endif
+
```
alexander-penev, line 98: This is also now unconditional and the change can be carried over from Compatibility.h into the other Clad files.

include/clad/Differentiator/Compatibility.h
```diff
@@ -134,34 +134,14 @@ static inline IfStmt* IfStmt_Create(const ASTContext &Ctx,
...
 }
 
-
-// Clang 8 change Node->getIdentType() ===> Node->getIdentKind()
-
-#if CLANG_VERSION_MAJOR < 8
-   #define getIdentKind() getIdentType()
-#endif
-
-
```
alexander-penev, line 144: Is this ported over to the Clad code?
i.e.
Has "Node->getIdentType()" been changed in the other Clad's files to "Node->getIdentKind()"?

include/clad/Differentiator/Compatibility.h
```diff
@@ -239,35 +216,25 @@ static inline void ExprSetDeps(Expr* result, Expr* Node) {
...
    )
 {
-#if CLANG_VERSION_MAJOR < 8
-   return new (Ctx) CXXOperatorCallExpr(Ctx, OpKind, Fn, Args, Ty, VK, OperatorLoc, FPFeatures);
-#elif CLANG_VERSION_MAJOR >= 8
    return CXXOperatorCallExpr::Create(Ctx, OpKind, Fn, Args, Ty, VK, OperatorLoc, FPFeatures
    CLAD_COMPAT_CLANG11_CXXOperatorCallExpr_Create_ExtraParamsUse
    );
-#endif
+
```
alexander-penev, line 230: Another already unconditional moment. Good readability of the code will be achieved if clad_comapt::CXXOperatorCallExpr_Create calls are replaced accordingly in the other files and removed from Compatibility.h

include/clad/Differentiator/Compatibility.h
```diff
@@ -284,62 +251,36 @@ static inline CXXMemberCallExpr* CXXMemberCallExpr_Create(ASTContext &Ctx,
...
 static inline CaseStmt* CaseStmt_Create(ASTContext &Ctx,
    Expr *lhs, Expr *rhs, SourceLocation caseLoc, SourceLocation ellipsisLoc, SourceLocation colonLoc)
 {
-#if CLANG_VERSION_MAJOR < 8
-   return new (Ctx) CaseStmt(lhs, rhs, caseLoc, ellipsisLoc, colonLoc);
-#elif CLANG_VERSION_MAJOR >= 8
+
    return CaseStmt::Create(const_cast<ASTContext&>(Ctx), lhs, rhs, caseLoc, ellipsisLoc, colonLoc);
-#endif
+
```
alexander-penev, line 261: Another already unconditional moment. Good readability of the code will be achieved if clad_comapt::CaseStmt_Create calls are replaced accordingly in the other files and removed from Compatibility.h

include/clad/Differentiator/Compatibility.h
```diff
@@ -284,62 +251,36 @@ static inline CXXMemberCallExpr* CXXMemberCallExpr_Create(ASTContext &Ctx,
...
-#endif
-
-
-// Clang 8 change result->setNumArgs(Ctx, Num) ===> result->setNumArgsUnsafe(Num)
-
-#if CLANG_VERSION_MAJOR < 8
-   #define setNumArgsUnsafe(NUM) setNumArgs(Ctx, NUM)
-#endif
-
+#define CLAD_COMPAT_CLANG8_Ctx_ExtraParams Ctx,
```
alexander-penev, line 283: Another already unconditional moment. It just needs to be replaced in all other files.
CLAD_COMPAT_CLANG8_Ctx_ExtraParams -> Ctx,

include/clad/Differentiator/Compatibility.h
```diff
@@ -478,11 +431,9 @@ static inline QualType getConstantArrayType(const ASTContext &Ctx,
...
                                                  const CXXMethodDecl* method) {
-#if CLANG_VERSION_MAJOR >= 8
+
   auto thisType = method->getThisType();
-#elif CLANG_VERSION_MAJOR < 8
-  auto thisType = method->getThisType(SemaRef.getASTContext());
-#endif
+
   return thisType;
 }
```
alexander-penev, line 438: Another already unconditional moment.
clad_compat::CXXMethodDecl_getThisType(ref,method) -> method->getThisType()

include/clad/Differentiator/StmtClone.h
```diff
@@ -122,10 +122,7 @@ namespace utils {
     DECLARE_CLONE_FN(PseudoObjectExpr)
     DECLARE_CLONE_FN(SubstNonTypeTemplateParmExpr)
     DECLARE_CLONE_FN(CXXScalarValueInitExpr)
-    // `ConstantExpr` node is only available after clang 7.
-    #if CLANG_VERSION_MAJOR > 7
     DECLARE_CLONE_FN(ConstantExpr)
-    #endif
 #if CLANG_VERSION_MAJOR > 8
     // `ValueStmt` node is only available after clang 8.
     DECLARE_CLONE_FN(ValueStmt)
```
alexander-penev, line 128: The idea of Compatibility.h is to isolate all version-related parts of the code in one place. Thus, there are dozens of places where such a conditional compilation is not to be found.

This code should be properly ported to Compatilbility.h. each this is a bit more difficult than a simple "#ifdef", but it pays off.

include/clad/Differentiator/VisitorBase.h
```diff
@@ -191,8 +191,13 @@ namespace clad {
...
       clang::CompoundStmt* body = V.endBlock();
-      clang::Expr* lambda =
-          S.ActOnLambdaExpr(noLoc, body, V.getCurrentScope()).get();
+      #if CLANG_VERSION_MAJOR > 17
+        clang::Expr* lambda =
+            S.ActOnLambdaExpr(noLoc, body).get();
+      #else
+        clang::Expr* lambda =
+            S.ActOnLambdaExpr(noLoc, body, V.getCurrentScope()).get();
+      #endif // CLANG_VERSION_MAJOR > 17
```
alexander-penev, line 200: The idea of Compatibility.h is to isolate all version-related parts of the code in one place. Thus, there are dozens of places where such a conditional compilation is not to be found.

This code should be properly ported to Compatilbility.h. each this is a bit more difficult than a simple "#ifdef", but it pays off.

lib/Differentiator/CladUtils.cpp
```diff
@@ -166,7 +166,12 @@ namespace clad {
         utils::BuildNNS(semaRef, const_cast<clang::DeclContext*>(declContext), CSS);
         NestedNameSpecifier* NS = CSS.getScopeRep();
         if (auto* Prefix = NS->getPrefix())
-          return C.getElaboratedType(ETK_None, Prefix, QT);
+          //clang-18 use ElaboratedTypeKeyword::None instead of ETK_None
+          #if CLANG_VERSION_MAJOR < 18
+            return C.getElaboratedType(ETK_None, Prefix, QT);
+          #else
+            return C.getElaboratedType(ElaboratedTypeKeyword::None, Prefix, QT);
+          #endif
```
alexander-penev, line 174: The idea of Compatibility.h is to isolate all version-related parts of the code in one place. Thus, there are dozens of places where such a conditional compilation is not to be found.

This code should be properly ported to Compatilbility.h. each this is a bit more difficult than a simple "#ifdef", but it pays off.

lib/Differentiator/CladUtils.cpp
```diff
@@ -269,12 +274,21 @@ namespace clad {
...
                                                     /*ASM=*/ArrayType::Normal,
                                                     /*IndexTypeQuals*/ 0);
+      #else
+          StrTy = clad_compat::getConstantArrayType(C, CharTyConst,
+                                                    llvm::APInt(/*numBits=*/32,
+                                                                str.size() + 1),
+                                                    /*SizeExpr=*/nullptr,
+                                                    /*ASM=*/ArraySizeModifier::Normal,
+                                                    /*IndexTypeQuals*/ 0);
+      #endif
```
alexander-penev, line 291: The idea of Compatibility.h is to isolate all version-related parts of the code in one place. Thus, there are dozens of places where such a conditional compilation is not to be found.

This code should be properly ported to Compatilbility.h. each this is a bit more difficult than a simple "#ifdef", but it pays off.

lib/Differentiator/StmtClone.cpp
```diff
@@ -235,7 +234,12 @@ DEFINE_CLONE_EXPR(CXXScalarValueInitExpr,
 DEFINE_CLONE_EXPR(ExtVectorElementExpr, (Node->getType(), Node->getValueKind(), Clone(Node->getBase()), Node->getAccessor(), Node->getAccessorLoc()))
 DEFINE_CLONE_EXPR(CXXBoolLiteralExpr, (Node->getValue(), Node->getType(), Node->getSourceRange().getBegin()))
 DEFINE_CLONE_EXPR(CXXNullPtrLiteralExpr, (Node->getType(), Node->getSourceRange().getBegin()))
-DEFINE_CLONE_EXPR(CXXThisExpr, (Node->getSourceRange().getBegin(), Node->getType(), Node->isImplicit()))
+#if CLANG_VERSION_MAJOR < 18
+   DEFINE_CLONE_EXPR(CXXThisExpr, (Node->getSourceRange().getBegin(), Node->getType(), Node->isImplicit()))
+#else
+   DEFINE_CREATE_EXPR(CXXThisExpr, (Ctx, Node->getSourceRange().getBegin(), Node->getType(), Node->isImplicit()))
+#endif
+
```
alexander-penev, line 242: The idea of Compatibility.h is to isolate all version-related parts of the code in one place. Thus, there are dozens of places where such a conditional compilation is not to be found.

This code should be properly ported to Compatilbility.h. each this is a bit more difficult than a simple "#ifdef", but it pays off.

lib/Differentiator/StmtClone.cpp
```diff
@@ -333,7 +337,8 @@ Stmt* StmtClone::VisitUnresolvedLookupExpr(UnresolvedLookupExpr* Node) {
   TemplateArgumentListInfo TemplateArgs;
   if (Node->hasExplicitTemplateArgs())
     Node->copyTemplateArgumentsInto(TemplateArgs);
-  Stmt* result = UnresolvedLookupExpr::Create(Ctx,
+  #if CLANG_VERSION_MAJOR < 18
```
alexander-penev, line 340: The idea of Compatibility.h is to isolate all version-related parts of the code in one place. Thus, there are dozens of places where such a conditional compilation is not to be found.

This code should be properly ported to Compatilbility.h. each this is a bit more difficult than a simple "#ifdef", but it pays off.

lib/Differentiator/VisitorBase.cpp
```diff
@@ -452,7 +452,11 @@ namespace clad {
 
     // Create elaborated type with namespace specifier,
     // i.e. class<T> -> clad::class<T>
-    return m_Context.getElaboratedType(ETK_None, NS, TT);
+    #if CLANG_VERSION_MAJOR < 18
+      return m_Context.getElaboratedType(ETK_None, NS, TT);
+    #else
+      return m_Context.getElaboratedType(ElaboratedTypeKeyword::None, NS, TT);
+    #endif
```
alexander-penev, line 459: The idea of Compatibility.h is to isolate all version-related parts of the code in one place. Thus, there are dozens of places where such a conditional compilation is not to be found.

This code should be properly ported to Compatilbility.h. each this is a bit more difficult than a simple "#ifdef", but it pays off.

alexander-penev (changes requested):
I made some comments about Compatibility.h. It is good that some changes are made before we accept this PR.

mcbarton:
> I made some comments about Compatibility.h. It is good that some changes are made before we accept this PR.

@alexander-penev thanks for the feedback. I'll try to get the majority addressed by sometime tomorrow. I'll link the individual commits with your messages, so that you can see what I am trying to address in each, in case its not obvious from the commit message.

include/clad/Differentiator/StmtClone.h
mcbarton: I believe this comment was a mistake. It doesn't appear to be pointing to something that needs changing.

alexander-penev: #if CLANG_VERSION_MAJOR > 8
has no place outside of Compatibility.h

Of course, it has nothing to do with version 18 or the changes made in this PR, but it is something that needs to be fixed anyway, because it is related to the similar changes in this PR. If it is not done here it will have to be done in a separate PR.

mcbarton: @alexander-penev Ok, I'll investigate this after I have resolved the other issues raised.

include/clad/Differentiator/VisitorBase.h
mcbarton: @alexander-penev Could you make a suggestion on how to move this to Compatibility.h ? I made an attempt to move it, but even if it had worked I don't think it was the right way to do it.

lib/Differentiator/StmtClone.cpp
mcbarton: Does this need to be moved? It seems similar to what has already been implemented in StmtClone.cpp here https://github.com/mcbarton/clad/blob/7741c3c5580b06a593bdafd485a955b7a31d7573/lib/Differentiator/StmtClone.cpp#L244C1-L257C7 for SubstNonTypeTemplateParmExpr .

CMakeLists.txt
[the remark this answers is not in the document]
mcbarton: 18.x.x didn't work, so I have reverted back to 19.0.0 . I will leave it like this unless you have another idea of what to try for CLANG_MAX_SUPPORTED, etc.

include/clad/Differentiator/VisitorBase.h
alexander-penev, line 200: We could add a clad_compat::ActOnLambdaExpr(S, loc, body, scope) function, but this may lead to unnecessary additional calculation of the last parameter in cases of clang>17. I recommend another approach:
```suggestion
        clang::Expr* lambda = S.ActOnLambdaExpr(noLoc, body/*,*/
          CLAD_COMPAT_CLANG17_ActOnLambdaExpr_getCurrentScope_ExtraParam(V)).get();
```
In Compatibility.h must add this:
```
      #if CLANG_VERSION_MAJOR > 17
        #define CLAD_COMPAT_CLANG17_ActOnLambdaExpr_getCurrentScope_ExtraParam(V) /**/
      #else
        #define CLAD_COMPAT_CLANG17_ActOnLambdaExpr_getCurrentScope_ExtraParam(V) ,(V).getCurrentScope()
      #endif // CLANG_VERSION_MAJOR > 17
```

alexander-penev: We could add a clad_compat::ActOnLambdaExpr(S, loc, body, scope) function, but this may lead to unnecessary additional calculation of the last parameter in cases of clang>17. I recommend another approach:
```suggestion
        clang::Expr* lambda = S.ActOnLambdaExpr(noLoc, body/*,*/
          CLAD_COMPAT_CLANG17_ActOnLambdaExpr_getCurrentScope_ExtraParam(V)).get();
```
In Compatibility.h must add this:
```
      #if CLANG_VERSION_MAJOR > 17
        #define CLAD_COMPAT_CLANG17_ActOnLambdaExpr_getCurrentScope_ExtraParam(V) /**/
      #else
        #define CLAD_COMPAT_CLANG17_ActOnLambdaExpr_getCurrentScope_ExtraParam(V) ,(V).getCurrentScope()
      #endif // CLANG_VERSION_MAJOR > 17
```

lib/Differentiator/StmtClone.cpp
alexander-penev: Yes, the problem here is the same as in 1-2 other places. When in the past we "let go" of some not very good change, then an avalanche of similar changes occurs.

If we don't fix it here and now (not just for version 18) we'll need another PR to fix what was not done very well before.

include/clad/Differentiator/Compatibility.h
mcbarton: @alexander-penev @vgvassilev can you make a suggestion? This is the last thing I need to change, before I can remove this code, but I have made multiple attempts to do without success. After this change has been made and the code mentioned in this issue removed, the PR is ready for another review.

mcbarton:
> @vgvassilev It is now building clad with clang-18 but it doesn't pass all the tests. It says that it is expected that 2 will fail, but 5 tests fail. Which 2 are expected to fail?

@alexander-penev which are the 2 tests which are expected to fail?

include/clad/Differentiator/Compatibility.h
mcbarton: I have now found a solution.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1875,14 +1871,14 @@ StmtDiff BaseForwardModeVisitor::VisitCXXTemporaryObjectExpr(
                             OpaquePtr<QualType>::make(TOE->getType()),
                             utils::GetValidSLoc(m_Sema), clonedArgs,
                             utils::GetValidSLoc(m_Sema)
-                                CLAD_COMPAT_IS_LIST_INITIALIZATION_PARAM(TOE))
+                                , TOE->isListInitialization())
```
alexander-penev, line 1874:
```suggestion
                                TOE->isListInitialization())
```

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1875,14 +1871,14 @@ StmtDiff BaseForwardModeVisitor::VisitCXXTemporaryObjectExpr(
...
                         .get();
   Expr* derivedTOE = m_Sema
                          .ActOnCXXTypeConstructExpr(
                              OpaquePtr<QualType>::make(TOE->getType()),
                              utils::GetValidSLoc(m_Sema), derivedArgs,
                              utils::GetValidSLoc(m_Sema)
```
alexander-penev, line 1880:
```suggestion
                             utils::GetValidSLoc(m_Sema),
```

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1875,14 +1871,14 @@ StmtDiff BaseForwardModeVisitor::VisitCXXTemporaryObjectExpr(
...
-                                 CLAD_COMPAT_IS_LIST_INITIALIZATION_PARAM(TOE))
+                                 ,TOE->isListInitialization())
```
alexander-penev, line 1881:
```suggestion
                                 TOE->isListInitialization())
```

lib/Differentiator/StmtClone.cpp
```diff
@@ -74,7 +74,7 @@ DEFINE_CLONE_EXPR_CO11(
     (CLAD_COMPAT_CLANG11_Ctx_ExtraParams Clone(Node->getSubExpr()),
      Node->getOpcode(), CloneType(Node->getType()), Node->getValueKind(),
      Node->getObjectKind(),
-     Node->getOperatorLoc() CLAD_COMPAT_CLANG7_UnaryOperator_ExtraParams
+     Node->getOperatorLoc() , Node->canOverflow()
```
alexander-penev, line 77:
```suggestion
     Node->getOperatorLoc(), Node->canOverflow()
```

alexander-penev:
> > @vgvassilev It is now building clad with clang-18 but it doesn't pass all the tests. It says that it is expected that 2 will fail, but 5 tests fail. Which 2 are expected to fail?
> 
> @alexander-penev which are the 2 tests which are expected to fail?

@mcbarton @vgvassilev The two tests that are expected to fall don't matter because they are not part of the 5 that are currently falling.
As far as I've noticed the falling tests are mainly falling because of some change in function info printing in the new version 18 (I don't know if this can be set or if the check needs to be made more complicated). For example:
```
//CHECK: int f_darg0(int x) __attribute__((always_inline)) {
```

but the printed for the generated result is

```
__attribute__((always_inline)) int f_darg0(int x) {
```


One of the other tests seems to fail because (perhaps a new warning):
> ArrayInputsReverseMode.C:228:14: warning: variable length arrays in C++ are a Clang extension [-Wvla-cxx-extension]

or

> ArrayInputsReverseMode.C:228:14: note: read of non-const variable 'n' is not allowed in a constant expression

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1730,8 +1726,8 @@ BaseForwardModeVisitor::DeriveSwitchStmtBodyHelper(const Stmt* stmt,
           (newCaseSC->getLHS() ? Clone(newCaseSC->getLHS()) : nullptr);
       Expr* rhsClone =
           (newCaseSC->getRHS() ? Clone(newCaseSC->getRHS()) : nullptr);
-      newActiveSC = clad_compat::CaseStmt_Create(
-          m_Sema.getASTContext(), lhsClone, rhsClone, noLoc, noLoc, noLoc);
+      newActiveSC = CaseStmt::Create(const_cast<ASTContext&>(m_Sema.getASTContext()), lhsClone, rhsClone, noLoc, noLoc, noLoc);
```
vgvassilev, line 1729:
```suggestion
      newActiveSC = CaseStmt::Create(Ctx, lhsClone, rhsClone, noLoc, noLoc, noLoc);
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3440,8 +3443,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     Expr* lhsClone = (CS->getLHS() ? Clone(CS->getLHS()) : nullptr);
     Expr* rhsClone = (CS->getRHS() ? Clone(CS->getRHS()) : nullptr);
 
-    auto* newSC = clad_compat::CaseStmt_Create(m_Sema.getASTContext(), lhsClone,
-                                               rhsClone, noLoc, noLoc, noLoc);
+    auto* newSC = CaseStmt::Create(const_cast<ASTContext&>(m_Sema.getASTContext()), lhsClone, rhsClone, noLoc, noLoc, noLoc);
```
vgvassilev, line 3446:
```suggestion
    auto* newSC = CaseStmt::Create(Ctx, lhsClone, rhsClone, noLoc, noLoc, noLoc);
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3623,8 +3625,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
   CaseStmt* ReverseModeVisitor::BreakContStmtHandler::GetNextCFCaseStmt() {
     ++m_CaseCounter;
     auto* counterLiteral = CreateSizeTLiteralExpr(m_CaseCounter);
-    CaseStmt* CS = clad_compat::CaseStmt_Create(m_RMV.m_Context, counterLiteral,
-                                                nullptr, noLoc, noLoc, noLoc);
+    CaseStmt* CS = CaseStmt::Create(const_cast<ASTContext&>(m_RMV.m_Context), counterLiteral, nullptr, noLoc, noLoc, noLoc);
```
vgvassilev, line 3628:
```suggestion
    CaseStmt* CS = CaseStmt::Create(Ctx, counterLiteral, nullptr, noLoc, noLoc, noLoc);
```

lib/Differentiator/StmtClone.cpp
```diff
@@ -399,11 +402,7 @@ Stmt* StmtClone::VisitShuffleVectorExpr(ShuffleVectorExpr* Node) {
 }
 
 Stmt* StmtClone::VisitCaseStmt(CaseStmt* Node) {
-  CaseStmt* result = clad_compat::CaseStmt_Create(Ctx, Clone(Node->getLHS()),
-                                        Clone(Node->getRHS()),
-                                        Node->getCaseLoc(),
-                                        Node->getEllipsisLoc(),
-                                        Node->getColonLoc());
+  CaseStmt* result = CaseStmt::Create(const_cast<ASTContext&>(Ctx), Clone(Node->getLHS()), Clone(Node->getRHS()), Node->getCaseLoc(), Node->getEllipsisLoc(), Node->getColonLoc());
```
vgvassilev, line 405:
```suggestion
  CaseStmt* result = CaseStmt::Create(Ctx, Clone(Node->getLHS()), Clone(Node->getRHS()), Node->getCaseLoc(), Node->getEllipsisLoc(), Node->getColonLoc());
```

tools/ClangPlugin.cpp
```diff
@@ -79,10 +79,10 @@ namespace clad {
...
       llvm::StringRef CladSoPath;
       for (llvm::StringRef P : Opts.Plugins)
-        if (llvm::sys::path::stem(P).endswith("clad")) {
-          CladSoPath = P;
-          break;
-        }
+          if (llvm::sys::path::stem(P).ends_with("clad")) {
+            CladSoPath = P;
+            break;
+          }
```
vgvassilev, line 85: Looks like we got some wrong indent here...

alexander-penev (approved):

CMakeLists.txt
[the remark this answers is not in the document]
alexander-penev: https://discourse.llvm.org/t/rfc-name-the-first-release-from-a-branch-n-1-0-instead-of-n-0-0/75384

In my opinion, following the new LLVM release scheme and what we've done so far we should do the following:
```suggestion
  set(CLANG_MIN_SUPPORTED 8.0)
  set(CLANG_MAX_SUPPORTED "18.1.x")
  set(CLANG_VERSION_UPPER_BOUND 18.2.0)
  set(LLVM_MIN_SUPPORTED 8.0)
  set(LLVM_MAX_SUPPORTED "18.1.x")
  set(LLVM_VERSION_UPPER_BOUND 18.2.0)
```

If a new version of LLVM appears i.e. version 18.2.x, then that means we have an LLVM ABI break and we need to upgrade (with or without Clad changes) our last supported version. If of course LLVM 19 comes out, then we should also upgrade our last supported version to 19.1.x, making the necessary changes and compatibility tests (as well as upgrade the minimally supported version to 9.0).

README.md
```diff
@@ -231,7 +231,7 @@ Clad also provides certain flags to save and print the generated derivative code
 - To print the Clad generated derivative: `-Xclang -plugin-arg-clad -Xclang -fdump-derived-fn`
 
 ## How to install
-At the moment, LLVM/Clang 7.0.x - 17.0.x are supported.
+At the moment, LLVM/Clang 8.0.x - 18.x.x are supported.
```
alexander-penev, line 234:
```suggestion
At the moment, LLVM/Clang 8.0.x - 18.1.x are supported.
```

include/clad/Differentiator/Compatibility.h
```diff
@@ -727,12 +704,17 @@ static inline bool IsPRValue(const Expr* E) { return E->isPRValue(); }
 #define CLAD_COMPAT_CLANG16_CXXDefaultArgExpr_getRewrittenExpr_Param(Node) /**/
 #endif
 
-// Clang 15 rename StringKind::Ascii to StringKind::Ordinary
+// Clang 15 renamed StringKind::Ascii to StringKind::Ordinary
+// Clang 18 renamed clang::StringLiteral::StringKind::Ordinary became clang::StringLiteralKind::Ordinary;
 
 #if CLANG_VERSION_MAJOR < 15
    const auto StringKind_Ordinary = clang::StringLiteral::StringKind::Ascii;
```
alexander-penev, line 711: When there are changes we try to stick to the names in the latest version, for greater proximity to the new versions (easier help search, easier and clearer names, etc.).

```suggestion
   const auto StringLiteralKind_Ordinary = clang::StringLiteral::StringKind::Ascii;
```

include/clad/Differentiator/Compatibility.h
```diff
@@ -727,12 +704,17 @@ static inline bool IsPRValue(const Expr* E) { return E->isPRValue(); }
...
 #elif CLANG_VERSION_MAJOR >= 15
-   const auto StringKind_Ordinary = clang::StringLiteral::StringKind::Ordinary;
+   #if CLANG_VERSION_MAJOR < 18
+      const auto StringKind_Ordinary = clang::StringLiteral::StringKind::Ordinary;
```
alexander-penev, line 714:
```suggestion
      const auto StringLiteralKind_Ordinary = clang::StringLiteral::StringKind::Ordinary;
```

include/clad/Differentiator/Compatibility.h
```diff
@@ -727,12 +704,17 @@ static inline bool IsPRValue(const Expr* E) { return E->isPRValue(); }
...
+   #else
+      const auto StringKind_Ordinary = clang::StringLiteralKind::Ordinary;
```
alexander-penev, line 716:
```suggestion
      const auto StringLiteralKind_Ordinary = clang::StringLiteral::StringKind::Ordinary;
```

include/clad/Differentiator/Compatibility.h
```diff
@@ -21,6 +21,97 @@ namespace clad_compat {
...
+   #define CLAD_COMPAT_CLANG17_ActOnLambdaExpr_getCurrentScope_ExtraParam(V) ,(V).getCurrentScope()
+#endif 
+
+//Clang 18 ArrayType::Normal -> ArraySizeModifier::Normal
+
+#if LLVM_VERSION_MAJOR < 18
+#define ArrayType_Normal ArrayType::Normal
+#else
+#define ArrayType_Normal ArraySizeModifier::Normal
+#endif
```
alexander-penev, line 47:
```suggestion
//Clang 18 ArrayType::Normal -> ArraySizeModifier::Normal

#if LLVM_VERSION_MAJOR < 18
const auto ArraySizeModifier_Normal = ArrayType::Normal
#else
const auto ArraySizeModifier_Normal ArraySizeModifier::Normal
#endif
```
It's a good idea to avoid `#define` as much as possible.
I think it's easy here. In addition, in the code we will use `clad_compat::ArraySizeModifier_Normal`, which, apart from being closer to the current version as a name, shows that it is a matter of compatibility.

include/clad/Differentiator/Compatibility.h
```diff
@@ -21,6 +21,97 @@ namespace clad_compat {
...
+
+}
+
+//Clang 18 ETK_None -> ElaboratedTypeKeyword::None
+
+#if LLVM_VERSION_MAJOR < 18
+#define ElaboratedTypeKeyword_None ETK_None
+#else
+#define ElaboratedTypeKeyword_None ElaboratedTypeKeyword::None
+#endif
```
alexander-penev, line 104:
```suggestion
#if LLVM_VERSION_MAJOR < 18
const auto ElaboratedTypeKeyword_None ETK_None
#else
const auto ElaboratedTypeKeyword_None ElaboratedTypeKeyword::None
#endif
```
We use where necessary as `clad_compat::ElaboratedTypeKeyword_None`.

tools/ClangPlugin.cpp
alexander-penev, line 85:
```suggestion
        if (llvm::sys::path::stem(P).ends_with("clad")) {
          CladSoPath = P;
          break;
        }
```

mcbarton:
@vgvassilev @alexander-penev I have addressed your feedback. This PR is ready further review

vgvassilev:
Thank you @mcbarton and @alexander-penev!

vgvassilev:
The codecov and clang-format complaints are bogus.

PR #826, opened 2024-03-18
Add Kokkos unittests

gojakuch:
This adds unit tests for Kokkos' `parallel_for`, `deep_copy`, and `Kokkos::View` accesses. Actual calls to `clad::differentiate` and `clad::gradient` are commented out, as the features are not implemented yet. This expands upon the tests in "Kokkos-aware Clad #783" (more tests for more features), improves readability (separate tests for forward and reverse modes) and matches the gtest guidelines more (no snake_case in test names, as `_` is reserved for other purposes).


unittests/Kokkos/TestUtils.hpp
```diff
@@ -0,0 +1,8 @@
+// Useful things
+
+#pragma once
```
vgvassilev, line 3: Let's use old-school include protectors as they are better because do not cache path to the file.

vgvassilev:
@kliegeois, does that set of test look good? My intent is to land several tests and then gradually incorporate #783.

unittests/Kokkos/TestUtils.hpp
gojakuch: you mean smth like

```
#ifndef KOKKOS_UNITTEST_UTILS
#define KOKKOS_UNITTEST_UTILS
...
#endif
```

right?

vgvassilev: Yes.

kliegeois:
@vgvassilev I will review the tests later this week or early next week.
Thanks for the PR!

unittests/Kokkos/ViewBasics.cpp
```diff
@@ -0,0 +1,165 @@
...
+#include "clad/Differentiator/Differentiator.h"
+#include "gtest/gtest.h"
+#include "TestUtils.hpp"
+#include "parallel_sum.hpp"
+
+double f(double x, double y) {
+  const int N = 2;
+
+  Kokkos::View<double*[N]> a("a", N);
+  Kokkos::View<double*[N]> b("b", N);
```
kliegeois, line 15: Those Kokkos views are accessed just after from outside a parallel region.
Therefore, it would be better here to use
```
Kokkos::View<double*[N], Kokkos::HostSpace>
```

unittests/Kokkos/ParallelFor.cpp
```diff
@@ -0,0 +1,101 @@
...
+  //   f_grad.execute(x, &dx);
+  //   EXPECT_NEAR(dx_f_FD, dx, abs(tau*dx));
+  // }
+}
+
+double parallel_polynomial(double x) {
+  Kokkos::View <double[1]> res("res");
+  res(0) = 0;
+  Kokkos::parallel_for("polycalc", 5, KOKKOS_LAMBDA(const int i) {
+      res(0) += pow(x, i+1)/(i+1);
```
kliegeois, line 56: This line includes data races.
Different threads (associated to different `i`) write to the same memory address `res(0)`. 
This can be rewritten using a `parallel_reduce` or using a `parallel_for` with atomic adds.

unittests/Kokkos/ParallelFor.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+TEST(ParallelFor, HelloWorldLambdaLoopReverse) {
+  // check finite difference and reverse mode similarity
+
+  // const double eps = 1e-5;
+  // const double tau = 1e-6; // tolerance
+
+  std::function<double(double)> _f = [](double x) {
+    double res = 0.;
+    Kokkos::parallel_for("HelloWorld-2", 5, [&res, x](const int i){ res = x*x; });
```
kliegeois, line 37: This line includes data races.
Different threads (associated to different i) write to the same double `res`.
I would suggest to have a different example that does not have data races.
I would suggest to use `KOKKOS_LAMBDA`.

unittests/Kokkos/ParallelFor.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+TEST(ParallelFor, HelloWorldLambdaLoopForward) {
+  // check finite difference and forward mode similarity
+
+  // const double eps = 1e-5;
+  // const double tau = 1e-6; // tolerance
+
+  std::function<double(double)> _f = [](double x) {
+    double res = 0.;
+    Kokkos::parallel_for("HelloWorld-1", 5, [&res, x](const int i){ res = x*x; });
```
kliegeois, line 15: This line includes data races.
Different threads (associated to different i) write to the same double `res`.
I would suggest to have a different example that does not have data races.
I would suggest to use `KOKKOS_LAMBDA`.

kliegeois (commented):
I have some comments on the examples.

unittests/Kokkos/ViewBasics.cpp
gojakuch: I understand that. my intention was to test different spaces, since the tests should not just have only the "sane and relevant" code. a user may not provide this template argument here, so I thought such a test case should be included. there's another test which uses `f_2` function that has views with `Kokkos::HostSpace`

unittests/Kokkos/ParallelFor.cpp
gojakuch: again, I understand that this is not a good way to use Kokkos. it was my intention to test this, because this should not break Clad for Kokkos. but I agree that a test for `parallel_reduce` is needed as well.

unittests/Kokkos/ParallelFor.cpp
gojakuch: okay, I will add more examples without the data races 👍

unittests/Kokkos/ViewBasics.cpp
kliegeois: I am just worried that if someone tries the Clad/Kokkos unit tests with a Kokkos installation that supports some GPUs, then those tests will fail.

gojakuch: ok, my bad I didn't know this. I will fix it then, thanks for the feedback

unittests/Kokkos/ParallelFor.cpp
vgvassilev: We should start with canonical examples where things work as suggested. I think that’s what Kim is saying. Once we get these things under control we can go for more exotic cases or things that are on the borderline in terms of parallel programming good practices.

unittests/Kokkos/parallel_sum.hpp
```diff
@@ -0,0 +1,348 @@
...
+template <class ExecSpace, class DT, class... DP>
+void parallel_sum(const ExecSpace& space, typename Kokkos::ViewTraits<DT, DP...>::value_type &sum, const Kokkos::View<DT, DP...> A) {
+  using ViewtypeA = Kokkos::View<DT, DP...>;
+  space.fence("parallel_sum: pre sum fence");
+  if (A.span_is_contiguous()) {
+
+    using ViewTypeFlat = Kokkos::View<
+        typename ViewtypeA::value_type*, Kokkos::LayoutRight,
+        Kokkos::Device<typename ViewtypeA::execution_space,
+                      std::conditional_t<ViewtypeA::rank == 0,
```
brian-kelley, line 275: same as above

unittests/Kokkos/parallel_sum.hpp
```diff
@@ -0,0 +1,348 @@
...
+    ViewSum<ViewTypeFlat, Kokkos::LayoutRight, 1, int>::template execute<typename ViewTypeFlat::execution_space>(sum, A_flat, space);
+  }
+  else {
+    ViewSum<ViewtypeA, typename ViewtypeA::array_layout, ViewtypeA::rank, int>::template execute<ExecSpace>(sum, A, space);
+  }
+  space.fence("parallel_sum: post sum fence");
+}
+
+template <class DT, class... DP>
+void parallel_sum(Kokkos::View<DT, DP...> A, typename Kokkos::ViewTraits<DT, DP...>::const_value_type b) {
```
brian-kelley, line 290: Going by the other function names this should probably be ``parallel_add``?

brian-kelley: (Unless the derivative code relies on these different functions having the same name)

unittests/Kokkos/parallel_sum.hpp
```diff
@@ -0,0 +1,348 @@
...
+    ViewAdd<ViewTypeFlat, Kokkos::LayoutRight, 1, int>::template execute<typename ViewTypeFlat::execution_space>(A_flat, b);
+  }
+  else {
+    ViewAdd<ViewtypeA, typename ViewtypeA::array_layout, ViewtypeA::rank, int>::template execute<typename ViewtypeA::execution_space>(A, b);
+  }
+  Kokkos::fence("parallel_sum: post add fence");
+}
+
+template <class ExecSpace, class DT, class... DP>
+void parallel_sum(const ExecSpace& space, Kokkos::View<DT, DP...> A, typename Kokkos::ViewTraits<DT, DP...>::const_value_type b) {
```
brian-kelley, line 313: Same as above

unittests/Kokkos/parallel_sum.hpp
```diff
@@ -0,0 +1,348 @@
...
+    ViewAdd<ViewTypeFlat, Kokkos::LayoutRight, 1, int>::template execute<typename ViewTypeFlat::execution_space>(A_flat, b, space);
+  }
+  else {
+    ViewAdd<ViewtypeA, typename ViewtypeA::array_layout, ViewtypeA::rank, int>::template execute<ExecSpace>(A, b, space);
+  }
+  space.fence("parallel_sum: post add fence");
+}
+
+template <class DT, class... DP, class ST, class... SP>
+void parallel_sum(Kokkos::View<DT, DP...> A, const Kokkos::View<ST, SP...> B) {
```
brian-kelley, line 336: Same

unittests/Kokkos/parallel_sum.hpp
kliegeois: The derivative code (as in my PR) uses the same name but that can be easily changed.

unittests/Kokkos/parallel_sum.hpp
```diff
@@ -0,0 +1,348 @@
...
+template <class DT, class... DP>
+void parallel_sum(typename Kokkos::ViewTraits<DT, DP...>::value_type &sum, const Kokkos::View<DT, DP...> A) {
+  using ViewtypeA = Kokkos::View<DT, DP...>;
+  Kokkos::fence("parallel_sum: pre sum fence");
+  if (A.span_is_contiguous()) {
+
+    using ViewTypeFlat = Kokkos::View<
+        typename ViewtypeA::value_type*, Kokkos::LayoutRight,
+        Kokkos::Device<typename ViewtypeA::execution_space,
+                      std::conditional_t<ViewtypeA::rank == 0,
```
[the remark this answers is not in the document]
kliegeois: I am not sure why this is required.

I based that logic on the `deep_copy` https://github.com/kokkos/kokkos/blob/master/core/src/Kokkos_CopyViews.hpp#L1311-L1317 .

unittests/Kokkos/parallel_sum.hpp
kliegeois: After thinking more about it, I think that it would be best to change the name to `parallel_add` because in the future we will need to change the signature of the function to take as argument `(view_type_1, view_type_2)` to be able to deal with scalar stored on the device. When we will do that, the `parallel_sum` and `parallel_add` would have exactly the same signature if they have the same name.

unittests/Kokkos/ParallelFor.cpp
gojakuch: @brian-kelley thanks for the suggestion, this test is better. however, `[&res, x]` is still needed because otherwise it's impossible to assign to any of the elements of res:
```
unittests/Kokkos/ParallelFor.cpp:17:20: error: read-only variable is not assignable
            res[i] = x*x;
            ~~~~~~ ^
```

brian-kelley: @gojakuch ah ok, I thought res would decay to a pointer here (should have tried it first 😄)

unittests/Kokkos/ParallelFor.cpp
```diff
@@ -0,0 +1,95 @@
...
+    //   double dx;
+    //   f_grad.execute(x, &dx);
+    //   EXPECT_NEAR(dx_f_FD, dx, abs(tau*dx));
+    // }
+}
+
+double parallel_polynomial_for(double x) { // data races
+    Kokkos::View <double[1], Kokkos::HostSpace> res("res");
+    res(0) = 0;
+    Kokkos::parallel_for("polycalc", 5, KOKKOS_LAMBDA(const int i) {
```
kliegeois, line 61: Here you need to specify a `Policy` on host as you did above:

```
using Policy = Kokkos::RangePolicy<Kokkos::DefaultHostExecutionSpace>;
Kokkos::parallel_for("HelloWorld-reverse", Policy(0, 5),
```
otherwise, the code will try to access `res(0)` from device due to:
```
res(0) += pow(x, i+1)/(i+1);
```
which will fail as most likely the host memory is not accessible from device.

kliegeois: Alternatively, if you want to set `res(0)` on host and then run the `parallel_for` without specifying the policy, you can do this:

```
Kokkos::View <double[1]> res("res");
auto res_host = Kokkos::create_mirror_view(Kokkos::view_alloc(Kokkos::WithoutInitializing), res);
// or
// auto res_host = Kokkos::create_mirror_view(res);
res_host(0) = 1; //I would suggest not to use 0 as by default Kokkos views are initialized to zero.
Kokkos::deep_copy(res, res_host);
Kokkos::parallel_for("polycalc", 5, KOKKOS_LAMBDA(const int i) {
```

gojakuch: alright, I will do this with a `Policy` then. I do want to keep some tests for `parallel_for` without it though. I think everything should work fine if I replace the policy with 5 in the code above where it's actually used now and make the changes you suggest to this. what do you think?

vgvassilev:
@gojakuch, I was wondering what is the current state of this PR? Are all comments addressed? Are we close to merging?

kliegeois:
@vgvassilev I need to review it a last time. @gojakuch if you could avoid squashing commits and force pushing for the next PRs, that would be great to more easily see what has changed since the last review. Thanks! An alternative is to squash only when merging the PR.

vgvassilev:
> @vgvassilev I need to review it a last time. @gojakuch if you could avoid squashing commits and force pushing for the next PRs, that would be great to more easily see what has changed since the last review. Thanks! An alternative is to squash only when merging the PR.

@kliegeois, thanks a lot, this is really helpful!

gojakuch:
> I was wondering what is the current state of this PR? Are all comments addressed? Are we close to merging?

@vgvassilev I think I addressed everything that has been commented the last time I pushed. I hope for the last review from @kliegeois and I think we could merge after that.

> if you could avoid squashing commits and force pushing for the next PRs, that would be great to more easily see what has changed since the last review.

alright. I honestly didn't think it's a big issue and thought that this is a good practice to just squash such little changes (I've been told so some time ago). sorry for the inconvenience.

unittests/Kokkos/parallel_add.hpp
```diff
@@ -0,0 +1,348 @@
...
+    ViewSum<ViewTypeFlat, Kokkos::LayoutRight, 1, int>::template execute<typename ViewTypeFlat::execution_space>(sum, A_flat);
+  }
+  else {
+    ViewSum<ViewtypeA, typename ViewtypeA::array_layout, ViewtypeA::rank, int>::template execute<typename ViewtypeA::execution_space>(sum, A);
+  }
+  Kokkos::fence("parallel_add: post sum fence");
+}
+
+template <class ExecSpace, class DT, class... DP>
+void parallel_add(const ExecSpace& space, typename Kokkos::ViewTraits<DT, DP...>::value_type &sum, const Kokkos::View<DT, DP...> A) {
```
kliegeois, line 267: This should be named `parallel_sum` as before.

@brian-kelley comments were for the functions on lines 290, 313, and 336.

unittests/Kokkos/parallel_add.hpp
```diff
@@ -0,0 +1,348 @@
...
+            const iType& i1,
+            const iType& i2) {
+                v(i0, i1, i2) += update(i0, i1, i2);
+        });
+  }
+};
+
+
+template <class DT, class... DP>
+void parallel_add(typename Kokkos::ViewTraits<DT, DP...>::value_type &sum, const Kokkos::View<DT, DP...> A) {
```
kliegeois, line 244: This should be named `parallel_sum` as before.

@brian-kelley comments were for the functions on lines 290, 313, and 336.

kliegeois (commented):
Two last comments on my side. Thanks!

unittests/Kokkos/parallel_add.hpp
gojakuch: fixed

unittests/Kokkos/parallel_add.hpp
gojakuch: fixed

vgvassilev:
@brian-kelley, is that good to go from your side, too?

unittests/Kokkos/ParallelFor.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include "parallel_add.hpp"
+
+TEST(ParallelFor, HelloWorldLambdaLoopForward) {
+    // // check finite difference and forward mode similarity
+    // const double eps = 1e-5;
+    // const double tau = 1e-6; // tolerance
+
+    std::function<double(double)> _f = [](double x) {
+        double res[5] = {0};
+        Kokkos::parallel_for("HelloWorld-forward", 5,
```
brian-kelley, line 14: This still has the issue of GPU trying to access host memory (``res``) if a GPU backend is enabled in Kokkos (if one is enabled, it will be the default). ``x`` is OK because it is captured by value.

If this isn't intentional, either ``res`` should be made a Kokkos::View in the default space, like:
```
Kokkos::View<double*> res("res", 5);
Kokkos::parallel_for("HelloWorld-forward", 5,
KOKKOS_LAMBDA(const int i) {
            res[i] = x*x;
        });
        // copy res back to host
        auto res_host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), res);
        // all elements of res should be the same, so return one of them arbitrarily
        return res_host[2];
```
or any examples that access host memory should be made to run on host only, using ``Kokkos::RangePolicy<Kokkos::DefaultHostExecutionSpace>(0, 5)`` like on line 57/60.

unittests/Kokkos/ParallelFor.cpp
```diff
@@ -0,0 +1,94 @@
...
+TEST(ParallelFor, HelloWorldLambdaLoopReverse) {
+    // // check finite difference and reverse mode similarity
+    // const double eps = 1e-5;
+    // const double tau = 1e-6; // tolerance
+
+    std::function<double(double)> _f = [](double x) {
+        double res[5] = {0};
+        Kokkos::parallel_for("HelloWorld-reverse", 5,
+        [&res, x](const int i) {
+            res[i] = x*x;
```
brian-kelley, line 40: Same as above

vgvassilev:
@gojakuch, I noticed that you chose a base branch a branch different than the master. I have corrected that in the PR. Can you rebase this PR against the master branch?

brian-kelley (commented):
Thanks @gojakuch, this looks good to me!

unittests/Kokkos/ParallelFor.cpp
```diff
@@ -0,0 +1,96 @@
+#include <Kokkos_Core.hpp>
+#include "clad/Differentiator/Differentiator.h"
+#include "gtest/gtest.h"
+// #include "TestUtils.hpp"
+#include "parallel_add.hpp"
```
vgvassilev, line 5: Why we have `parallel_add.hpp` but `ParallelFor.cpp`. Let's follow the same convention and rename it to `parallel_for.cpp`.

vgvassilev (commented):
I do not understand why the builds are failing with this PR. Can you rebase it once again?

unittests/Kokkos/CMakeLists.txt
```diff
@@ -12,5 +14,6 @@ if (NOT (LLVM_REQUIRES_RTTI OR LLVM_ENABLE_RTTI))
   endif()
 endif()
 
+target_link_libraries(KokkosTests PRIVATE Kokkos::kokkos)
```
vgvassilev, line 17: We should drop this to make the Kokkos bot work.

unittests/Kokkos/ParallelFor.cpp
gojakuch: I would suggest renaming `parallel_add.hpp` into `ParallelAdd.hpp` because I think CamelCase is preferred by the gtest library. is that okay?

vgvassilev: Sure.

unittests/Kokkos/CMakeLists.txt
gojakuch: removing this breaks something for me though:

```
clad/unittests/Kokkos/main.cpp:1:10: fatal error: 'Kokkos_Core.hpp' file not found
#include <Kokkos_Core.hpp>
         ^~~~~~~~~~~~~~~~~
1 error generated.
make[3]: *** [unittests/Kokkos/CMakeFiles/KokkosTests.dir/build.make:63: unittests/Kokkos/CMakeFiles/KokkosTests.dir/main.cpp.o] Error 1
make[2]: *** [CMakeFiles/Makefile2:670: unittests/Kokkos/CMakeFiles/KokkosTests.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:977: test/CMakeFiles/check-clad.dir/rule] Error 2
make: *** [Makefile:476: check-clad] Error 2
```

vgvassilev: I don’t see how this influences the include paths.

vgvassilev: Is `${Kokkos_INCLUDE_DIRS}` defined for you? Or there is a different variable? Worst case scenario would be to do something like:

if (TARGET Kokkos::kokkos)
  target_link_libraries(KokkosTests PRIVATE Kokkos::kokkos)
else()
  target_link_libraries(KokkosTests PUBLIC ${Kokkos_LIBRARIES})
  target_include_directories(KokkosTests SYSTEM PRIVATE ${Kokkos_INCLUDE_DIRS})
endif()

unittests/Kokkos/TestUtils.hpp
```diff
@@ -0,0 +1,21 @@
+// Useful things
```
vgvassilev, line 1: We name these files .h.

gojakuch: okay, I will rename this

gojakuch: should I also rename ParallelAdd.hpp to ParallelAdd.h?

vgvassilev: Yeah.

unittests/Kokkos/CMakeLists.txt
```diff
@@ -12,5 +14,5 @@ if (NOT (LLVM_REQUIRES_RTTI OR LLVM_ENABLE_RTTI))
   endif()
 endif()
 
-target_link_libraries(KokkosTests PUBLIC ${Kokkos_LIBRARIES})
-target_include_directories(KokkosTests SYSTEM PRIVATE ${Kokkos_INCLUDE_DIRS})
+target_include_directories(KokkosTests SYSTEM PUBLIC ${Kokkos_INCLUDE_DIRS})
+target_link_libraries(KokkosTests PUBLIC ${Kokkos_LIBRARIES})
```
vgvassilev, line 18: Why that needed to change?

vgvassilev:
Can you apply clang-format (as instructed in the output of the bot) and rebase this PR. Overall it looks good, the remaining test failures seem unrelated to it but I will need to investigate further.

unittests/Kokkos/CMakeLists.txt
gojakuch: it was an accident, I changed it back.

vgvassilev:
We need to apply clang-format as suggested by the bot for multicommit setups...

gojakuch:
> We need to apply clang-format as suggested by the bot for multicommit setups...

@vgvassilev I'm not sure if I understood you correctly. However, does this really matter if we're going to squash everything into a single commit "Add Kokkos unittests" when merging (as kliegeois once suggested)? at least, that's what I thought. the PR now passes all the checks except those we have talked about that I don't really understand and the clang-format / precheckin doesn't seem to be complaining.

unittests/Kokkos/CMakeLists.txt
```diff
@@ -13,4 +15,4 @@ if (NOT (LLVM_REQUIRES_RTTI OR LLVM_ENABLE_RTTI))
 endif()
 
 target_link_libraries(KokkosTests PUBLIC ${Kokkos_LIBRARIES})
-target_include_directories(KokkosTests SYSTEM PRIVATE ${Kokkos_INCLUDE_DIRS})
+target_include_directories(KokkosTests SYSTEM PUBLIC ${Kokkos_INCLUDE_DIRS})
```
vgvassilev, line 18: Does PRIVATE/PUBLIC change makes a difference?

gojakuch: not really. you suggested it to fix building for me, but that turned out to be unrelated. we can revert it, if that's preferable.

vgvassilev (approved):
LGTM! The failures in osx13 are quite puzzling to me. The problem came from the fact that we did not want to link libLLVM.so and added `DISABLE_LLVM_LINK_LLVM_DYLIB` to `add_unittest` that apparently somehow breaks osx13...

PR #827, opened 2024-03-18
Fix erroneous TBR option setting

vaithak:
This was removed and moved to `SetRequestOptions` here: [808](https://github.com/vgvassilev/clad/pull/808/files#diff-959ddda1c4fe3a3b7c6560ce25882ad6ef819af216c1f8c0db57d2bced814ee5L116), but somehow got reverted back in [766](https://github.com/vgvassilev/clad/pull/766/files#diff-959ddda1c4fe3a3b7c6560ce25882ad6ef819af216c1f8c0db57d2bced814ee5R397), and I missed it during the review also :(.

This doesn't do any harm for now, but if we want to enable-tbr by default in future, this will nullify that.


vgvassilev (approved):
LGTM!

PR #828, opened 2024-03-19
Fix loop body and its AST Indexing when there's a continue stmt

### LINKED ISSUE #710: Incorrect derivative when loops contains `continue`
Reproducible example:
```cpp
#include <iostream>

    double res = 11 * u;
    for (int i = 0; i < 3; i++) {
        res += u * i;
        continue;
    }
    return res;
}

int main() {
    double u = 3, v = 5;
    double du, dv;
    du = dv = 0;
    d_fn.execute(u, v, &du, &dv);
    show(du);
    show(dv);
}
```

**Output**
du: 20
dv: 0

**Expected output**
du: 14
dv: 0

kchristin22:
This PR fixes #710 and expands to propose a different way of loop's body derivation in reverse mode. As a result, `tests` need to be updated accordingly if the PR is approved.

For more details please see [this comment](https://github.com/vgvassilev/clad/issues/710#issuecomment-2007143504).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /build
 /.vscode
+/inst
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -522,12 +522,6 @@ namespace clad {
     /// PopBreakContStmtHandler();
     /// ```
     class BreakContStmtHandler {
-      /// Keeps track of all the created switch cases. It is required
-      /// because we need to register all the switch cases later with the
-      /// switch statement that will be used to manage the control flow in
-      /// the reverse block.
-      llvm::SmallVector<clang::SwitchCase*, 4> m_SwitchCases;
-
       /// `m_ControlFlowTape` tape keeps track of which `break`/`continue`
       /// statement was hit in which iteration.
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 /build
 /.vscode
+/inst
```
parth-07, line 3: This should be in a separate PR.

parth-07 (changes requested):
Thank you for the pull request and very good analysis in the [comment](https://github.com/vgvassilev/clad/issues/710#issuecomment-2007143504) here.

The issue #710 and the analysis in the above-mentioned comment seems to be solving different things, and thus should be solved separately. Can you please update this pull request such that this only solves the issue #710, and create an additional issue for your additional analysis in the comment? Please also add an example in the issue demonstrating where Clad currently generates an incorrect derivative.

kchristin22:
Yes sorry I got carried away. I will open one with the change for #710, one with the redesign and one for the gitignore change.

I have tested many cases and with the change that fixes 710, clad runs as expected. I will try some more cases just to be sure. I initially started the redesign cause I thought that it didn't run correctly, but it turns out that this little change suffices. However, the other suggested way seems cleaner and less risk-prone, so I thought it was worth discussing. We can continue the discussion in the new PR.

PR #831, opened 2024-03-20
Add a test to prevent a regression in #464

### LINKED ISSUE #464: Incorrect derivative when arrays are used in chained assignment expression in loops
Reproducible example:
```cpp
double fn(double *arr, int n) {
  double res = 0;
  for (int i=0; i<n; ++i) {
    res += (arr[i] *= 5);
  }
  return res;
}
auto fn_grad = clad::gradient(fn, "arr");
clad::array_ref<double> ref(darr, 5);
fn_grad.execute(arr, 5, ref);
for (int i=0; i < 5; ++i)
  std::cout<<darr[i]<<" ";
```

Actual Output: `1 1 1 1 1`
Expected Output: `5 5 5 5 5`

PetroZarytskyi:
The way we differentiate assignment operations has changed a lot recently. At some point, #464 was fixed (as well as similar issues). This PR adds a test to prevent a possible regression.

Fixes #464.

### DIFF
--- a/test/Gradient/Loops.C
+++ b/test/Gradient/Loops.C
@@ -1624,6 +1624,47 @@ double f_loop_init_var(double lower, double upper) {
 // CHECK-NEXT:     }
 // CHECK-NEXT: }
 
+double fn20(double *arr, int n) {
+  double res = 0;
+  for (int i=0; i<n; ++i) {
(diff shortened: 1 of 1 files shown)


vaithak (approved):
lgtm

PR #833, opened 2024-03-20
Fix Incorrect derivative when loops contains continue

### LINKED ISSUE #710: Incorrect derivative when loops contains `continue`

### LINKED ISSUE #851: Don't increment the for loop variable when there's a break statement
Due to PR #833, the step of the for loop variable is performed for every branch case of the for loop. However, when there's a break statement in the for loop, the step is not applied afterwards, which isn't the case when there's a continue statement or neither of them. So we shouldn't increment the variable beforehand in this case.

Example:
```cpp
#include <iostream>

    double res = 0;
    for (int i = 1; i < 3; i++) {
        res += i * u;
        if (i == 1)
            break;
    }
    return res;
}

int main() {
    double u = 3, v = 5;
    double du, dv;
    du = dv = 0;
    d_fn.execute(u, v, &du, &dv);
    show(du);
    show(dv);
}
```
Expected output:
**du** = 3
**dv** = 0

Output:
**du** = 0
**dv** = 0

kchristin22:
For more details please see https://github.com/vgvassilev/clad/issues/710#issuecomment-2007143504. This PR concerns only the fix of #710, so the only change needed is the one regarding the position of  the step of the forward loop's variable in the reverse pass. Tests need to be updated accordingly if the change is approved.

Closes #710, closes #851.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1386,8 +1386,11 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       BodyDiff.updateStmtDx(utils::unwrapIfSingleStmt(revPassCondStmts));
     }
 
+    Stmt* revInit = loopCounter.getNumRevIterations()
+                        ? BuildDeclStmt(loopCounter.getNumRevIterations())
+                        : nullptr;
     Stmt* Reverse = new (m_Context)
-        ForStmt(m_Context, nullptr, nullptr, nullptr, CounterDecrement,
+        ForStmt(m_Context, revInit, nullptr, nullptr, CounterDecrement,
                 BodyDiff.getStmt_dx(), noLoc, noLoc, noLoc);
 
     addToCurrentBlock(initResult.getStmt_dx(), direction::reverse);
@@ -4123,10 +4126,22 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     activeBreakContHandler->UpdateForwAndRevBlocks(bodyDiff);
     PopBreakContStmtHandler();
 
+    Expr* revCounter = loopCounter.getCounterConditionResult().get().second;
+    if (m_CurrentBreakFlagExpr) {
+      VarDecl* numRevIterations = BuildVarDecl(m_Context.getSizeType(),
+                                               "_numRevIterations", revCounter);
+      loopCounter.setNumRevIterations(numRevIterations);
+    }
+
     // Increment statement in the for-loop is executed for every case
     if (forLoopIncDiff) {
       Stmt* forLoopIncDiffExpr = forLoopIncDiff;
       if (m_CurrentBreakFlagExpr) {
+        m_CurrentBreakFlagExpr =
+            BuildOp(BinaryOperatorKind::BO_LOr,
+                    BuildOp(BinaryOperatorKind::BO_NE, revCounter,
+                            BuildDeclRef(loopCounter.getNumRevIterations())),
+                    BuildParens(m_CurrentBreakFlagExpr));
         forLoopIncDiffExpr = clad_compat::IfStmt_Create(
             m_Context, noLoc, false, nullptr, nullptr, m_CurrentBreakFlagExpr,
             noLoc, noLoc, forLoopIncDiff, noLoc, nullptr);
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -71,20 +71,6 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     return Call;
   }
 
-  Expr* ReverseModeVisitor::CladTapeResult::Size() {
-    LookupResult& TapeSize = V.GetCladTapeSize();
-    CXXScopeSpec CSS;
-    CSS.Extend(V.m_Context, V.GetCladNamespace(), noLoc, noLoc);
-    Expr* SizeDRE = V.m_Sema
-                        .BuildDeclarationNameExpr(CSS, TapeSize,
-                                                  /*AcceptInvalidDecl=*/false)
-                        .get();
-    Expr* Call =
-        V.m_Sema.ActOnCallExpr(V.getCurrentScope(), SizeDRE, noLoc, Ref, noLoc)
-            .get();
-    return Call;
-  }
-
   ReverseModeVisitor::CladTapeResult
   ReverseModeVisitor::MakeCladTapeFor(Expr* E, llvm::StringRef prefix) {
     assert(E && "must be provided");
@@ -3808,6 +3794,9 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
 
     llvm::SaveAndRestore<bool> SaveIsInsideLoop(isInsideLoop);
     isInsideLoop = true;
+    llvm::SaveAndRestore<Expr*> SaveCurrentBreakFlagExpr(
+        m_CurrentBreakFlagExpr);
+    m_CurrentBreakFlagExpr = nullptr;
 
     Expr* condClone = (WS->getCond() ? Clone(WS->getCond()) : nullptr);
     const VarDecl* condVarDecl = WS->getConditionVariable();
@@ -3866,6 +3855,9 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
 
     llvm::SaveAndRestore<bool> SaveIsInsideLoop(isInsideLoop);
     isInsideLoop = true;
+    llvm::SaveAndRestore<Expr*> SaveCurrentBreakFlagExpr(
+        m_CurrentBreakFlagExpr);
+    m_CurrentBreakFlagExpr = nullptr;
 
     Expr* clonedCond = (DS->getCond() ? Clone(DS->getCond()) : nullptr);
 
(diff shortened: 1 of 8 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3529,6 +3529,10 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       bodyDiff.updateStmtDx(MakeCompoundStmt(revLoopBlock));
     m_LoopBlock.pop_back();
 
+    activeBreakContHandler->EndCFSwitchStmtScope();
+    activeBreakContHandler->UpdateForwAndRevBlocks(bodyDiff);
+    PopBreakContStmtHandler();
+    
     // Increment statement in the for-loop is only executed if the iteration
```
parth-07, line 3536: Please update this comment.

parth-07 (changes requested):
Thank you for the pull-request. The solution seems good. Can you please add tests and update the pull-request description?

Also, can you please open an issue regarding fixing incorrect derivatives when loops contains `break`? You can use this test case:

```cpp
#include <iostream>

    double res = 0;
    for (int i = 1; i < 3; i++) {
        res += i * u;
        if (i == 1)
            break;
    }
    return res;
}

int main() {
    double u = 3, v = 5;
    double du, dv;
    du = dv = 0;
    d_fn.execute(u, v, &du, &dv);
    show(du);
    show(dv);
}
```

kchristin22:
I will update the tests, the description and the comment.

> Also, can you please open an issue regarding fixing incorrect derivatives when loops contains break? You can use this test case:

Sure thing!

kchristin22:
Upon checking out the above example, the error occurs after the change of this PR right? I noticed that the decrement of i-- should be after the switch case when there's a break stmt as otherwise we "miss" a value of i. I will open a new PR for the new issue.

parth-07:
>  the error occurs after the change of this PR right? 

Yes, however, the `break` statement had incorrect behavior without this PR change as well. It's just that the particular example that I added happen to work for Clad on master.

> I noticed that the decrement of i-- should be after the switch case when there's a break stmt as otherwise we "miss" a value of i.

In the case of a `break` statement, the last increment does not happen. So we need to skip the corresponding decrement as well when the `break` statement is hit.

kchristin22:
I have not forgotten this. I will update it in the upcoming week. Sorry for the delay.

parth-07 (changes requested):
Can you please add a test case where Clad was giving incorrect derivative result before?

kchristin22:
@parth-07 I tried to add a new function in Loops.C, but even though I get the correct result when I compile locally this file and run it, somehow when I test using `make check-clad`, the result is expected to be 15.00 and if I change that to 15.00 it then says that the result of fn21 is expected to be 7.00 (which is correct). The fn21_grad function printed as a log due to the error also doesn't match the one that was checked and succeeded.

parth-07 (changes requested):
Can you please add tests as well?

kchristin22:
> Can we undo the formatting changes which are not related to the changes of this pr?

Yes sorry about that, I was trying to fix it, cause my local version of git-clang-format produces different results than the CI. I think now it includes only the necessary changes more or less.

parth-07:
@kchristin22 Can you please add tests in the PR?

test/Gradient/Loops.C
```diff
@@ -3133,6 +3142,123 @@ double fn39(double x) {
...
+//CHECK-NEXT:        }
+//CHECK-NEXT:        clad::push(_t2, {{2U|2UL}});
+//CHECK-NEXT:    }
+//CHECK-NEXT:    _d_res += 1;
+//CHECK-NEXT:    for (;; _t0--) {
+//CHECK-NEXT:        {
+//CHECK-NEXT:            if (!_t0)
+//CHECK-NEXT:                break;
+//CHECK-NEXT:        }
+//CHECK-NEXT:        if (clad::size(_t2) != 0 && clad::back(_t2) != 1)
```
parth-07, line 3241: Currently, we are repeating `clad::size(_t2)` in the condition if there are multiple `break` statements in the loop. Can you modify the condition-generation code such that `clad::size(...)` does not get repeated?

Also, do we need to check this condition on each iteration? We know that if `break` is hit, then it must be hit in the last iteration of the loop. Can we use that knowledge to only evaluate this condition once?

parth-07 (changes requested):

test/Gradient/Loops.C
kchristin22: I guess the other solution would be to perform the reverse of the inc diff and cond diff in the forward pass as well, and perform the inc and cond diff in the reverse pass without the if check. So, it would look like a `x + a - a = x` situation. I'm not sure if that's more efficient than the if-check tbh. Maybe the compiler would be able to tell that this if condition is only true in the first iteration either way.

parth-07: > I guess the other solution would be to perform the reverse of the inc diff and cond diff in the forward pass as well, and perform the inc and cond diff in the reverse pass without the if check.

I did not mean completely changing the solution.

With this solution, the `if`-condition for decrementing the counter variable when there are 2 break statements in the primal `for`-loop is as:

```cpp
 if (clad::size(_t2) != 0 && clad::back(_t2) != 1 && clad::size(_t2) != 0 && clad::back(_t2) != 2)
    i--;
```
This `if`-statement is evaluated per iteration of the for-loop. All the conditions in the `if`-statement needs to be evaluated for the common case where the iteration does not correspond to hitting `break` statement. That's 4 expression evaluations per iteration. We know that a `break` expression can only be hit once for a given loop, and if it is indeed hit, it is hit in the last iteration of the loop. Can we use this knowledge to optimize the solution? 

One way might be:

```cpp
 for (size numIterations = _t0;; _t0--) {
  {
    if (!_t0)
    break;
  }

  // if (clad::size(_t2) != 0 && clad::back(_t2) != 1 && clad::size(_t2) != 0 && clad::back(_t2) != 2)
  //   i--;
  // Instead of the above if-statement, we can have:
  if (_t0 != numIterations || (clad::back(_t2) != 1 || clad::back(_t2) != 2))
    i--;
  
  // ... remaining loop will be exactly the same as before ...
  // ...
  // ..
}
```
With this solution, `if`-statement will check for `break` statements counter only for the last iteration. For all other iterations, only one condition in the `if`-statement is evaluated.

kchristin22: I understood what you said, it just didn't go with the current formation of the for-loop (there's no `numIterations` currently), so I wasn't sure if you meant adding that or examine another possibility.

I'll see if I can add it easily.

kchristin22: I did it, let me know if there are any new comments @parth-07

test/Gradient/Loops.C
```diff
@@ -3133,6 +3142,123 @@ double fn39(double x) {
...
+//CHECK-NEXT:        }
+//CHECK-NEXT:        clad::push(_t2, {{2U|2UL}});
+//CHECK-NEXT:    }
+//CHECK-NEXT:    _d_res += 1;
+//CHECK-NEXT:    for (unsigned {{int|long}} _numRevIterations0 = _t0; ; _t0--) {
+//CHECK-NEXT:        {
+//CHECK-NEXT:            if (!_t0)
+//CHECK-NEXT:                break;
+//CHECK-NEXT:        }
+//CHECK-NEXT:        if (_t0 != _numRevIterations0 || (clad::size(_t2) != 0 && clad::back(_t2) != 1))
```
parth-07, line 3241: We don't need `clad::size()` expression now in the `if`-statement.

parth-07 (changes requested):

test/Gradient/Loops.C
kchristin22: It is generally needed in order to ensure that cases like fn17 are covered, where the the tape containing the break path is empty

parth-07: Do you mean this function?
```cpp
double fn17(double i, double j) {
  int counter = 5;
  double res = 0;
  for (int ii=0; ii<counter; ++ii) {
    int jj = ii;
    if (ii < 2)
      continue;
    while (jj--) {
      if (jj < 3) {
        res += i*j;
        break;
      } else {
        continue;
      }
      res += i*i*j*j;
    }
  }
  return res;
}
```

> where the the tape containing the break path is empty

What do you mean by the tape is empty for the break path?

kchristin22: You actually pointed me to a bug! I forgot to refresh the current break flag when visiting a while or do-while loop so in fn17 f.i., for the outer loop decrement, I checked if it was the break path of the inside loop which ofc wasn't necessary and also the inner loop had different iterations than the outer loop, so the tape for the inside loop had less entries, causing an error when its size wasn't checked before trying to read its last value. I fixed it now and removed the size expr.

parth-07 (approved):
Looks good.

test/Gradient/Loops.C
parth-07: Okay, sounds good.

PR #835, opened 2024-03-20
Redesign of loop's body in reverse pass 

### LINKED ISSUE #832: Consider a redesign of the for loop's body in reverse pass
The current implementation of the loop's body in the reverse pass of the Reverse mode is not very clear. Consider this example:

```
    double res = 11 * u;
    for (int i = 0; i < 3; i++) {
        if (i==2){
            res += u * i;
            continue;
        }
        else {
            res -= u * i;
        }
        res += u;
    }
    return res;
}
```

The for loop in the reverse pass, after the fix of #710, is derived like so:
```
for (; _t0; _t0--) {
        i--;
        switch (clad::pop(_t4)) {
          case 2UL:
            ;
            {
                res = clad::pop(_t6);
                double _r_d2 = _d_res;
                * _d_u += _r_d2;
            }
            if (clad::pop(_t2)) {
              case 1UL:
                ;
                {
                    res = clad::pop(_t3);
                    double _r_d0 = _d_res;
                    * _d_u += _r_d0 * i;
                    _d_i += u * _r_d0;
                }
            } else {
                {
                    res = clad::pop(_t5);
                    double _r_d1 = _d_res;
                    * _d_u += -_r_d1 * i;
                    _d_i += u * -_r_d1;
                }
            }
        }
```

Even though the results are correct and the cases handled as expected, the indexing of the code is not very traditional.

The proposed change evaluates this snippet as:
```
for (; _t0; _t0--) {
        i--;
        switch (clad::pop(_t4)) {
          case 2UL:
            {
                res = clad::pop(_t6);
                double _r_d2 = _d_res;
                * _d_u += _r_d2;
                if (clad::pop(_t2))
                    ;
                else {
                    {
                        res = clad::pop(_t5);
                        double _r_d1 = _d_res;
                        * _d_u += -_r_d1 * i;
                        _d_i += u * -_r_d1;
                    }
                }
                break;
            }
          case 1UL:
            {
                res = clad::pop(_t3);
                double _r_d0 = _d_res;
                * _d_u += _r_d0 * i;
                _d_i += u * _r_d0;
                break;
            }
        }
```

kchristin22:
This PR closes #832. What these changes establish is a cleaner switch body, where the cases are independent and indexed correctly, AST speaking.

[This comment](https://github.com/vgvassilev/clad/issues/710#issuecomment-2007143504) contains a comparison of the current version and the proposed one. Cases of nested continue statements have also been tested and the derived body and the results were correct. 

The workflow in a nutshell is the following:
* Replace the `then` or `else` stmt of `if` with the `case` one when there's a `continue` stmt in its body
* The branch of `if` that doesn't include a `continue` stmt is added firstly as it belongs to the previous switch case (outer one)
* After visiting the nodes of the loop's body in the reverse pass, the `case` stmts are brought forward, on the first level of the loop's body, by performing a Depth-First search on its compounds
* The compounds/stmts that are on the same level as the `case` stmt and that follow it (belong to this case) are also brought foward, while the rest of the stmts remain to their level and later appended in the beginning of the loop's body (as the main loop's body compound)
* At that point we add the switch case of the main body in the beginning and set the stmts present between two cases as a substmt to the first of them  

Keep in mind that since the compilation ends up in a completely different result compared to the current implementation in cases were there's a continue stmt, the `tests` have to be updated (so they're expected to currently fail).

If the changes are well received, support for the `break` stmt in a similar manner only requires an additional variable.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -944,6 +944,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     } 
     // if only then block contains a continue statement, we need to add
     // the then block to the current block and create an if stmt for the else block
+    // afterwards to ensure that in the reverse pass it will be included in the prior case
     else if (SaveHasContStmtThen.get()) {
       addToCurrentBlock(thenDiff.getStmt_dx(), direction::reverse);
       if (elseDiff.getStmt_dx()){
@@ -956,6 +957,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     } 
     // if only else block contains a continue statement, we need to add
     // the else block to the current block and create an if stmt for the then block
+    // afterwards to ensure that in the reverse pass it will be included in the prior case
     else if (hasContStmt) {
       addToCurrentBlock(elseDiff.getStmt_dx(), direction::reverse);
       Stmt* Reverse = clad_compat::IfStmt_Create(
@@ -3710,7 +3712,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
               betweenCase = true;
             } else {
               curBlockStmts.push_back(new (m_Context)
-                                          BreakStmt(Stmt::EmptyShell()));
+                                          BreakStmt(Stmt::EmptyShell())); // compatible with all clang versions
               curCaseStmt->setSubStmt(MakeCompoundStmt(curBlockStmts));
               curBlockStmts.clear();
             }
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -897,8 +897,13 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       return StmtDiff(Forward, Reverse);
     };
 
+    llvm::SaveAndRestore<bool> SaveHasContStmt(hasContStmt);
+    hasContStmt = false;
     StmtDiff thenDiff = VisitBranch(If->getThen());
+    llvm::SaveAndRestore<bool> SaveHasContStmtThen(hasContStmt);
+    hasContStmt = false;
     StmtDiff elseDiff = VisitBranch(If->getElse());
+
 
     // It is problematic to specify both condVarDecl and cond thorugh
     // Sema::ActOnIfStmt, therefore we directly use the IfStmt constructor.
@@ -931,7 +936,12 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
                                                thenDiff.getStmt_dx(),
                                                noLoc,
                                                elseDiff.getStmt_dx());
-    addToCurrentBlock(Reverse, direction::reverse);
+    if (!SaveHasContStmtThen.get()) 
+      addToCurrentBlock(Reverse, direction::reverse);
+    else{
+      addToCurrentBlock(thenDiff.getStmt_dx(), direction::reverse);
+      addToCurrentBlock(elseDiff.getStmt_dx(), direction::reverse);
+    }
     CompoundStmt* ForwardBlock = endBlock(direction::forward);
     CompoundStmt* ReverseBlock = endBlock(direction::reverse);
     endScope();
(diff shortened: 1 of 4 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -920,18 +925,48 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    // if only then block contains a continue statement, we need to add
+    // the then block to the current block and create an if stmt for the else block
+    // afterwards to ensure that in the reverse pass it will be included in the prior case
+    else if (SaveHasContStmtThen.get()) {
+      addToCurrentBlock(thenDiff.getStmt_dx(), direction::reverse);
+      if (elseDiff.getStmt_dx()){
+        Stmt* Reverse = clad_compat::IfStmt_Create(
+            m_Context, noLoc, If->isConstexpr(), initResult.getStmt_dx(),
+            condVarClone, reverseCond, noLoc, noLoc,
+            m_Sema.ActOnNullStmt(noLoc).get(), noLoc, elseDiff.getStmt_dx());
```
PetroZarytskyi, line 954: Can we invert the condition instead of using a null stmt?
 e.g. 
```
 if(!cond)
   *else stmt*
```
instead of 
```
 if(cond)
   ;
 else
   *else stmt*
```

parth-07 (changes requested):
Thank you for the pull request. We need to verify it on more cases and discuss it further to make this change ready to merge, as I am currently not confident about the correctness of this design change. The below example fails when Clad is built using the changes in this pull request:

```cpp
#include <iostream>

    double res = 0;
    for (int i = 0; i < 3; i++) {
        res += u;
        if (i == 1)
            continue;
    }
    return res;
}

int main() {
    double u = 3, v = 5;
    double du, dv;
    du = dv = 0;
    d_fn.execute(u, v, &du, &dv);
    show(du);
    show(dv);
}
```

**Output:**
du: 1
dv: 0

**Expected output:**
du: 3
dv: 0

Clad on master gives correct result.

-----------------

> where the cases are independent and indexed correctly, AST speaking.

In the current design, is the indexing of case statements invalid as per C++ standard?

lib/Differentiator/ReverseModeVisitor.cpp
kchristin22: I thought about this approach before but the way I believe this would turn out is checking for each binary condition operator (<=, >=, <, >, ==, !=) and replacing it accordingly. I will look into it more though, thanks for the comment.

kchristin22:
> Thank you for the pull request. We need to verify it on more cases and discuss it further to make this change ready to merge, as I am currently not confident about the correctness of this design change. The below example fails when Clad is built using the changes in this pull request:
> 
> ```c++
> #include "clad/Differentiator/Differentiator.h"
> #include <iostream>
> 
> #define show(x) std::cout << #x << ": " << x << "\n";
> 
> double fn(double u, double v) {
>     double res = 0;
>     for (int i = 0; i < 3; i++) {
>         res += u;
>         if (i == 1)
>             continue;
>     }
>     return res;
> }
> 
> int main() {
>     auto d_fn = clad::gradient(fn);
>     double u = 3, v = 5;
>     double du, dv;
>     du = dv = 0;
>     d_fn.execute(u, v, &du, &dv);
>     show(du);
>     show(dv);
> }
> ```
> 
> **Output:** du: 1 dv: 0
> 
> **Expected output:** du: 3 dv: 0
> 
> Clad on master gives correct result.
> 
> > where the cases are independent and indexed correctly, AST speaking.
> 
> In the current design, is the indexing of case statements invalid as per C++ standard?

Yes, I completely agree that more tests need to be executed to verify that all cases are correctly derived. May I suggest creating a table to gather all the possible cases that may occur in order to make sure that we have not missed any? I can start working on it and append whatever else you and @PetroZarytskyi would like.

Thank you for noticing it. I indeed did not cover the case where multiple cases share some stmts. Right now the approach I have in mind is keeping track of the stmts added in each level (this is the curBlock list probably at each level) till we hit a case and append them to the case hit. So it would look sth like this:
image

However I understand that this is not very efficient, and if I tried to emulate a fall-through approach it would look similar to the original approach. I will have to investigate further.

parth-07:
> May I suggest creating a table to gather all the possible cases that may occur in order to make sure that we have not missed any? I can start working on it and append whatever else you and @PetroZarytskyi would like.

I would suggest to wait because this feature is not a priority right now. This feature would be important to resolve if the current indexing of case statements is invalid as per the C++ standard. By current, I mean the implementation in the master branch. I know the indexing of case statements in the current implementation is quirky and complex but we should be okay as long as it does not violate C++ standard. 

> Right now the approach I have in mind is keeping track of the stmts added in each level (this is the curBlock list probably at each level) till we hit a case and append them to the case hit. So it would look sth like this:

Implementing this seems a little complicated and can lead to a significant code-size increase.

kchristin22:
Yes I agree, we can revisit this if I find another way to do this using fall-through. Feel free to add this as a Draft PR or even close it if it helps in scheduling the reviews of PRs.

PR #836, opened 2024-03-21
Complain if signatures of forward declaration and generated gradient code do not match

### LINKED ISSUE #810: Complain if signatures of the forward declaration and generated code mismatch
Clad builds a gradient with a specific signature. It also allows that gradient function signature to be forward-declared. However if both mismatch there are subtle crashes in CodeGen. We should detect such cases and issue an error if they are in the same translation unit:

For example,

https://github.com/vgvassilev/clad/blob/d7e5434cd4aecc9e8a3e61e1660f1543ae5f3954/test/Gradient/Gradients.C#L269

returns `void`, but the forward declaration is:

https://github.com/vgvassilev/clad/blob/d7e5434cd4aecc9e8a3e61e1660f1543ae5f3954/test/Gradient/Gradients.C#L283

which leads to subtle issues.

DeadSpheroid:
This PR aims to fix #810 by crashing compilation whenever a mismatched gradient forward declaration is found.
There are two separate cases to handle the generated gradient code and search for a forward declaration.
If its a class/struct method, a lookup is performed in the class.
While if its a regular function, a lookup is performed in the TU.
Do let me know if any changes are necessary.

### DIFF
--- a/tools/ClangPlugin.cpp
+++ b/tools/ClangPlugin.cpp
@@ -215,6 +215,59 @@ namespace clad {
         }
       }
 
+      // Only verify forward declaration in reverse mode
+      if (request.Mode == DiffMode::reverse) {
+        bool hasForwardDecl = false;
+        FunctionDecl* deriv = nullptr;
+        FunctionDecl* forwdecl = nullptr;
+
+        if (isa<CXXMethodDecl>(DerivativeDecl)) {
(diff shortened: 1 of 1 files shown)


DeadSpheroid:
I would merge the two code blocks for the class method and regular function into one, but for some reason i cant seem to get Sema::LookupName to look inside classes for results

PR #840, opened 2024-03-24
Add x86 local debugging example to documentation

MihailMihov:
Added a paragraph at the end of the developer documentation showing how to use `systemd-nspawn` to debug inside an Alpine x86 namespace container.


parth-07 (approved):
Looks good.

PR #844, opened 2024-03-28
Support call expressions with non-differentiable arguments in the reverse mode

### LINKED ISSUE #765: Functionality to specify/infer differentiable arguments in a function call.
Currently, we have a FIXME comment for this in ReverseModeVisitor.cpp
```
// FIXME: We should add instructions for handling non-differentiable
// arguments. Currently we are implicitly assuming function call only
// contains differentiable arguments.
```

This should allow working with examples like:
```cpp
// Compile with clang++ -std=c++11 -fplugin=/usr/lib/clad.so cladIssue.cxx -o cladIssue

#include <iostream>

inline double foo(double *params, double *obs)
{
   return params[0] * obs[0];
}

double wrapper(double* params, double* obs)
{
   // return params[0] * obs[0]; // works!

   return foo(params, obs); // doesn't work because it requires _d_obs !
}

int main()
{

   double param = 2.0;
   double obs = 2.0;
   double result = 0.0;
   grad.execute(&param, &obs, &result);

}
```

PetroZarytskyi:
In most cases, we support non-differentiable variables (i.e. variables that don't have adjoints). Currently, the major application for them is non-independent array parameters. For instance, for
```
double fn17 (double x, double* y) {
    return x;
}
```
a request ``clad::gradient(fn17, "x");`` will produce
```
void fn17_grad_0(double x, double *y, double *_d_x) {
    goto _label0;
  _label0:
    ;
}
```
In this example, ``y`` does not have an adjoint.

However, calling a function of ``y`` produces an error. In this PR, the non-differentiability of ``y`` is propagated to the pullback.

Fixes #765.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -273,6 +273,10 @@ namespace clad {
   }
 
   void DiffRequest::UpdateDiffParamsInfo(Sema& semaRef) {
+    // Diff info for pullbacks is generated automatically,
+    // its parameters are not provided by the user.
+    if (Mode == DiffMode::experimental_pullback)
+      return;
     DVI.clear();
     auto& C = semaRef.getASTContext();
     const Expr* diffArgs = Args;
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -471,8 +471,11 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     assert(m_Function && "Must not be null.");
 
     DiffParams args{};
-    std::copy(FD->param_begin(), FD->param_end(), std::back_inserter(args));
-
+    if (!request.DVI.empty())
+      for (const auto& dParam : request.DVI)
+        args.push_back(dParam.param);
+    else
+      std::copy(FD->param_begin(), FD->param_end(), std::back_inserter(args));
 #ifndef NDEBUG
     bool isStaticMethod = utils::IsStaticMethod(FD);
     assert((!args.empty() || !isStaticMethod) &&
@@ -1509,9 +1512,6 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     // statements there later.
     std::size_t insertionPoint = getCurrentBlock(direction::reverse).size();
 
-    // FIXME: We should add instructions for handling non-differentiable
-    // arguments. Currently we are implicitly assuming function call only
-    // contains differentiable arguments.
     bool isCXXOperatorCall = isa<CXXOperatorCallExpr>(CE);
 
     for (std::size_t i = static_cast<std::size_t>(isCXXOperatorCall),
@@ -1729,9 +1729,9 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
                "corresponding dfdx().");
       }
 
-      DerivedCallArgs.insert(DerivedCallArgs.end(),
-                             DerivedCallOutputArgs.begin(),
-                             DerivedCallOutputArgs.end());
+      for (Expr* arg : DerivedCallOutputArgs)
+        if (arg)
+          DerivedCallArgs.push_back(arg);
       pullbackCallArgs = DerivedCallArgs;
 
       if (pullback)
@@ -1782,6 +1782,10 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         // Silence diag outputs in nested derivation process.
         pullbackRequest.VerboseDiags = false;
         pullbackRequest.EnableTBRAnalysis = enableTBR;
+        bool isaMethod = isa<CXXMethodDecl>(FD);
+        for (size_t i = 0, e = FD->getNumParams(); i < e; ++i)
+          if (DerivedCallOutputArgs[i + isaMethod])
+            pullbackRequest.DVI.push_back(FD->getParamDecl(i));
         FunctionDecl* pullbackFD = plugin::ProcessDiffRequest(m_CladPlugin, pullbackRequest);
         // Clad failed to derive it.
         // FIXME: Add support for reference arguments to the numerical diff. If
--- a/test/Gradient/FunctionCalls.C
+++ b/test/Gradient/FunctionCalls.C
@@ -758,6 +758,59 @@ double fn16(double x, double y) {
 //CHECK-NEXT:     }
 //CHECK-NEXT: }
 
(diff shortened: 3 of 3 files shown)


vaithak:
@PetroZarytskyi, I am a little confused, but would this also work with redirections?
For ex: defining a local pointer as `double* z = &a`, the passing it to `add(x, z)` ?

PetroZarytskyi:
@vaithak 
> For ex: defining a local pointer as ``double* z = &x``, then passing it to ``add(x, z)`` ?

Yes, this example works. Why shouldn't it? 
At the same time, this will not work if we initialize ``z`` as ``double* z = y;`` since ``z`` will still be considered differentiable. I believe deducing the opposite requires activity analysis, which this PR is not about. This PR is just about supporting function calls. Moreover, even with this example Clad at least doesn't crash anymore.

vaithak (approved):

PetroZarytskyi:
@vgvassilev 
> Can we add a "Fixes #<issue_number>" to the commit so that it closes the related issues?

Done.

parth-07:
I am not sure if this solution / approach is correct.

```cpp
double fn17 (double x, double* y) {
    return x;
}

auto fn17_grad_x = clad::gradient(fn17, "x");
```

Shouldn't `clad::gradient(fn17, "x")` simply mean compute derivative of function value w.r.t `x`? It does not say anything about the `y` parameter. Assuming `y` to be non-differentiable in this case does not seem correct and is inconsistent with other Clad interfaces.

------------------------------------------------------------

The below examples gives wrong / non-intuitive results when this patch is used:

```cpp
double add(double a, double *b) { return a + b[0]; }

double fn17(double x, double *y) {
  y[0] = x;
  x = add(x, y);
  x = add(x, &x);
  return x;
}

int main() {
  auto fn17_grad_0 = clad::gradient(fn17, "x");
  double y[] = {3.0, 2.0}, dx = 0;
  fn17_grad_0.execute(5, y, &dx);
  show(dx);
}
```

**Output**
dx: 2

**Expected Output**
dx: 3

vgvassilev:
@parth-07, apologies for merging prematurely. @PetroZarytskyi can you take a look?

PetroZarytskyi:
Hi @parth-07. Hi @vgvassilev. Let me clarify. I am aware we cannot assume the array is not differentiable. However, we already did that before this PR. This is because we don't know the size of the array, hence we cannot construct the adjoint. All I did was extend support of this system to function calls, which was specifically requested by Jonas. He pointed out they had many use cases with constant non-differentiable arrays and function calls. The only workaround was to copy entire arrays to fixed-size arrays in loops, which became ugly after differentiation. All of the listed concerns apply in general to what we do when the user doesn't want the derivative w.r.t. an array (before this PR as well). e.g.
if we do the same without function calls
```
double fn17(double x, double *y) {
  y[0] = x;
  x = x + y[0];
  x = x + (&x)[0];
  return x;
}
```
The output is still 2. So if we're not satisfied with the solution in this PR, the only other solution is to require the user to differentiate w.r.t. all array parameters. However, I don't think we should take away functionality from the user by producing an error. In my opinion, the best solution would be to warn the user that the array has to be non-differentiable.

davidlange6:
warnings are ignored and crashes are likely preferred over silently producing wrong results...   can one instead "opt in" and require that clad  be told that an array is non-differentiable?

parth-07:
> I am aware we cannot assume the array is not differentiable. However, we already did that before this PR. 

We never considered an array to be non-differentiable. We only supported cases in which we did not need to know the size of the array.

> All I did was extend support of this system to function calls, which was specifically requested by Jonas. He pointed out they had many use cases with constant non-differentiable arrays and function calls. The only workaround was to copy entire arrays to fixed-size arrays in loops, which became ugly after differentiation.

I think adding non-differentiable entities is a good idea in general. However, we should not make Clad APIs interface inconsistent. So far, none of the Clad APIs follow not specifying a parameter to make it non-differentiable. I think it's best to have a different and a new way to specifying non-differentiable parameters.

> All of the listed concerns apply in general to what we do when the user doesn't want the derivative w.r.t. an array (before this PR as well).

How did the user convey, before this patch, that he doesn't want the derivative w.r.t an array?

PetroZarytskyi:
@parth-07 
> We never considered an array to be non-differentiable.

I believe we did. I think the function I gave in the comment above (combined with your ``main`` function) is an example of such a behavior:
```
void fn17_grad_0(double x, double *y, double *_d_x) {
    double _t0;
    double _t1;
    double _t2;
    _t0 = y[0];
    y[0] = x;
    _t1 = x;
    x = x + y[0];
    _t2 = x;
    x = x + &x[0];
    goto _label0;
  _label0:
    *_d_x += 1;
    {
        x = _t2;
        double _r_d1 = *_d_x;
        *_d_x -= _r_d1;
        *_d_x += _r_d1;
        &*_d_x[0] += _r_d1;
    }
    {
        x = _t1;
        double _r_d0 = *_d_x;
        *_d_x -= _r_d0;
        *_d_x += _r_d0;
    }
    y[0] = _t0;
}
```
We don't create adjoints for array parameters that are not specified (this is what I meant by "non-differentiable").
There's even a fixme dedicated to this:
```
// FIXME: fix when adding array inputs, now we are just skipping all
// array/pointer inputs (not treating them as independent variables).
```
And I don't see a way to fix this.
> So far, none of the Clad APIs follow not specifying a parameter to make it non-differentiable.

I agree this is confusing. Probably we have to decide between a warning and an error. But in my understanding, this issue is about the user not specifying an array parameter, not about this PR in particular. This PR only adds support for non-differentiable arguments in function calls. Which can also be used for any other non-differentiable args. By the way, I'm currently working on a different PR where I use the changes made here and it's not related to arrays at all.

parth-07:
> // FIXME: fix when adding array inputs, now we are just skipping all
  // array/pointer inputs (not treating them as independent variables).

Arrays were forced to be considered as non-differentiable in the _unsupported cases_. That's why the FIXME. I think, ideally, Clad should error out and refuse to differentiate the function if Clad cannot create an adjoint variable for an array/pointer variable.

> I agree this is confusing. Probably we have to decide between a warning and an error.

Why can't we decide a different interface for specifying non-differentiable parameters. Would something like this work?

```cpp
auto fn_17_grad_x = clad::gradient(fn17, /*DifferentiableParameters=*/"x", /*NonDifferentiableParameters=*/"y");
```

This way, the interface would not be misleading and inconsistent with other Clad APIs.

PetroZarytskyi:
> Would something like this work?

The solution looks good to me. Perhaps we should discuss adding a clad::gradient argument to list non-differentiable parameters.

Issue #715, opened 2024-01-18
Fix clad in binder and conda


vgvassilev:
Currently the link to binder is broken because clad does not work with xeus-cling.

We can decide to move away from xeus-cling in favor of xeus-clang-repl (and eventually xeus-cpp). We need to investigate why some tests fail in clad for xeus-clang-repl.

PR #846, opened 2024-03-30
[cmake] Do not build the error estimation codes with differentiable clang

vgvassilev:
We do not need to use a clad-compatible clang to build the error estimation demos.

This issue is related to #715.


vgvassilev:
The codecov report seems bogus, we do not touch any of the code but cmake.

PR #478, opened 2022-08-03
Add Support for Functions with Constant Array Type Parameters in Jacobian Mode

Nirhar:
Now One must be able to find the Jacobian of functions with Constant Arrays in the parameter list.
For example, a function of the form:
```cpp
void func(double arr[3], double x, double y, double* output){
	output[0]=arr[2]*x*y;
	.
	.
	.
	output[n-1]=arr[0]*arr[1]*arr[2];
}
```

will generate a Jacobian of size n x 5.

Corresponding tests for the same have been written.

Previously this feature was not implememnted because it was necessary to know the size
of an array to correctly determine the size of the output jacobian matrix. This has now
been achieved for constant arrays, as we know the array sizes for them and hence we can
precisely locate where each jacobian entry for a array parameter is located. This is
achieved with the help of m_IndependentVarsSize, which stores the number of actual parameters
that each function parameter corresponds to.

Prior to this commit a std::map<ValueDecl, Expr> was used to map variable declarations to
their corresponding jacobian expression, so that we can lookup the latter during the reversemode
computations quickly. While this is fine for primitive variables(because each variable will only
correspond to one jacobian expression), an array declaration will correspond to multiple jacobian
expressions, depending on which index of the array is being referred to. Since the ValueDecl can
only refer to the name of the array and not name+index, we must update the map to use the name of
the array + index as the key. This can be done easily by using a string representation of the
ValueDecl name with the index as a suffix as the key of the map. Hence variables like
m_ExprVariables, m_VectorOutputString have been introduced to map array declarations, indexed by
position in array to their corresponding jacobian expressions.

Closes #472

Issue #472, opened 2022-07-21
crash on jacobian with array inputs


FROL256:
Hello! I took an example of jacobian from main page (h). If i make the first parameter array, clad crashes :(

Input:
```cpp

//void h(float a, float b, float output[]) {
//    output[0] = a * a * a;
//    output[1] = a * a * a + b * b * b;
//    output[2] = 2 * (a + b);
//}

void h(float a[3], float b, float output[]) {
    output[0] = a[0] * a[0] * a[0];
    output[1] = a[0] * a[0] * a[0] + b * b * b;
    output[2] = 2 * (a[0] + b);
}
```

Error:
```
Executing task: clang-12 -I /home/frol/PROG/clad_autodiff_examples/inst/include -x c++ -std=c++11 -g -fplugin=/home/frol/PROG/clad_autodiff_examples/inst/lib/clad.so SourceFile.cpp -o z_out -lstdc++ -lm -Xclang -plugin-arg-clad -Xclang -fgenerate-source-file -DCLAD_NO_NUM_DIFF 

clang: /usr/lib/llvm-12/include/llvm/ADT/SmallVector.h:277: T& llvm::SmallVectorTemplateCommon<T, <template-parameter-1-2> >::operator[](llvm::SmallVectorTemplateCommon<T, <template-parameter-1-2> >::size_type) [with T = const clang::ValueDecl*; <template-parameter-1-2> = void; llvm::SmallVectorTemplateCommon<T, <template-parameter-1-2> >::reference = const clang::ValueDecl*&; llvm::SmallVectorTemplateCommon<T, <template-parameter-1-2> >::size_type = long unsigned int]: Assertion `idx < size()' failed.
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
[...]

PR #848, opened 2024-04-01
Restructure differentiation schedule into a breadth first traversal

vaithak:


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -275,8 +275,25 @@ namespace clad {
   void DiffRequest::UpdateDiffParamsInfo(Sema& semaRef) {
     // Diff info for pullbacks is generated automatically,
     // its parameters are not provided by the user.
-    if (Mode == DiffMode::experimental_pullback)
+    if (Mode == DiffMode::experimental_pullback) {
+      if (!Function->getPreviousDecl())
```
vgvassilev, line 279: What's the intent of this early exit here? Can we have a comment?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -386,57 +389,62 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+        // Creates the ArraySubscriptExprs for the independent variables
+        size_t idx = 0;
+        for (const auto* arg : args) {
+          // FIXME: fix when adding array inputs, now we are just skipping all
+          // array/pointer inputs (not treating them as independent variables).
+          if (utils::isArrayOrPointerType(arg->getType())) {
+            if (arg->getName() == "p")
+              m_Variables[arg] = m_Result;
+            idx += 1;
+            continue;
```
vgvassilev, line 407: Can we come up with a test here?

vgvassilev (commented):
Thank you! This is a major improvement. We have discussed that now we can process diff schedules separately from the differentiation process. That is, we first plan and then act. This PR does not get us there just yet, and this is why it relies on a forward declarations because it only appends to the diff plan instead of inserting.

That's an incremental change but we should also try to make the next step maybe in a separate PR.

lib/Differentiator/DiffPlanner.cpp
vaithak: added a comment 👍🏼

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -37,6 +37,7 @@ namespace clad {
     class CladPlugin;
     clang::FunctionDecl* ProcessDiffRequest(CladPlugin& P,
                                             DiffRequest& request);
+    void AddRequestToSchedule(CladPlugin& P, const DiffRequest& request);
```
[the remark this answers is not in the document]
vaithak: added a fixme 👍🏼

lib/Differentiator/ReverseModeVisitor.cpp
vaithak: This is for Jacobian computation and the particular case is for something related to ROOT as captured here: https://github.com/vgvassilev/clad/pull/478#discussion_r944066696.
I will look into this when I work on the Jacobain array-related issue: https://github.com/vgvassilev/clad/issues/472

include/clad/Differentiator/DerivativeBuilder.h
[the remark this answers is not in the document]
vgvassilev: Since this is more of a design issue we should keep track of it in the issue tracker to avoid dropping from the priority list.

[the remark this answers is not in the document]
vaithak: https://github.com/vgvassilev/clad/issues/857

parth-07 (approved):
Looks really good.

.github/workflows/ci.yml
```diff
@@ -911,7 +911,7 @@ jobs:
         cat obj/CMakeCache.txt
         cat obj/CMakeFiles/*.log
     - name: Setup tmate session
-      if: ${{ failure() && runner.debug }}
+      if: ${{ failure() && matrix.debug_build == true }}
```
vgvassilev, line 914: Can we undo this change?

vgvassilev (commented):
Looks like the clang-tidy reports are relevant. Can you address them?

.github/workflows/ci.yml
vaithak: removed 👍🏼

lib/Differentiator/VisitorBase.cpp
```diff
@@ -524,13 +528,17 @@ namespace clad {
...
+    if (Loc.isInvalid())
+      Loc = m_Function->getLocation();
     UnqualifiedId Member;
-    Member.setIdentifier(&m_Context.Idents.get(MemberFunctionName), noLoc);
+    Member.setIdentifier(&m_Context.Idents.get(MemberFunctionName), Loc);
     CXXScopeSpec SS;
     bool isArrow = Base->getType()->isPointerType();
     auto ME = m_Sema
-                  .ActOnMemberAccessExpr(getCurrentScope(), Base, noLoc,
+                  .ActOnMemberAccessExpr(getCurrentScope(), Base, Loc,
```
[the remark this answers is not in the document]
vaithak: This one seems irrelevant, it is using some heuristic to deduce that `noLoc` is more similar to `OpLoc`, maybe because of Levenshtein distance as mentioned here: https://releases.llvm.org/13.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/readability-suspicious-call-argument.html#levenshtein-distance-as-levenshtein.

vgvassilev (approved):
LGTM! Do you want the merge the 6 commits the way they are or you want to squash or something else?

vaithak:
I think the current order of commits is fine.

Issue #444, opened 2022-05-02
[ci] Add a clang/libc++ builders for unix


vgvassilev:

ro4i7:
Hello @parth-07 @vgvassilev 

To build script for a Unix-based system using Clang and libc++:
```
#!/bin/bash
export CC=/usr/bin/clang
export CXX=/usr/bin/clang++
export CXXFLAGS="-stdlib=libc++"
mkdir build && cd build
cmake ..
make -j4
```
This script sets the environment variables for the Clang compiler and the C++ standard library to be used during the build process. It then creates a build directory, runs CMake to configure the project, and builds the project using the make command with 4 parallel jobs.

Once the build script is set up, then configure our CI tool to execute this script on the virtual machine for each new commit or pull request. This will ensure that the project is continuously built and tested with the Clang compiler and libc++ library on a Unix system.

vgvassilev:
@mcbarton, do you mind looking into this?

mcbarton:
@vgvassilev yes, I can look into this for you.

vgvassilev:
> @vgvassilev yes, I can look into this for you.

Thank you. I figured would be probably a quick fix since you dug in that area recently..

mcbarton:
Closing as a result of the outcome of the debugging showing conflicting standard libraries trying to be used (see https://github.com/vgvassilev/clad/pull/850#issuecomment-2067596340 for more details).

@vgvassilev can you close this issue?

vgvassilev:
Yes, closing.

PR #850, opened 2024-04-03
[CI] Add script to build Clad using clang/libc++ on Unix systems

mcbarton:
This PR introduces a build script which can be used to build Clad on Unix systems. CI tests the build script for all supported llvms.
Fixes https://github.com/vgvassilev/clad/issues/444


mcbarton:
@vgvassilev This PR is ready for review. In subsequent PRs i'll add to the script the ability to build documentation, use Cuda, Kokkos, etc.

mcbarton:
@alexander-penev are you able to review this PR for me please?

vgvassilev:
Would it be worth writing the build script in python for windows eventually?

mcbarton:
As the number of options for building expands it will be worth changing from bash. I can investigate using python for both Windows and Unix. I will raise this as an issue and once I have it working locally I will put in a separate PR.

vgvassilev:
I think the issue is about adding libcxx to a ubuntu node, nothing more. I see here we added a lot of other builds that I am not sure if we need them. Maybe one arm build is fine...

mcbarton:
@vgvassilev I think I misunderstood the issue. I will reduce the number of jobs to one Ubuntu and one osx to cover both Unix systems.

mcbarton:
> I think the issue is about adding libcxx to a ubuntu node, nothing more. I see here we added a lot of other builds that I am not sure if we need them. Maybe one arm build is fine...

@vgvassilev The latest commit should fix the issue.

vgvassilev:
How do we tell a ubuntu node to install libcxx instead of libstdc++?

mcbarton:
> How do we tell a ubuntu node to install libcxx instead of libstdc++?

My guess would be to install this package https://packages.ubuntu.com/source/focal/libcxx-serial and do apt purge on anything to do with libstdc++ in case its installed on the node before you install the libcxx package. That way is should be on the only one available.

vgvassilev:
Perhaps it needs to have clang as a compiler too.

mcbarton:
@vgvassilev Assumming the llvm install comes with libcxx on Ubuntu like it does with Homebrew, then we should able to use it by defining in my script
LDFLAGS="-L$path_to_llvm_build/lib/c++ -Wl,-rpath,$path_to_llvm_build/lib/c++"
This allows the script to still be suitable outside the ci.

mcbarton:
The script has been updated to in theory use libcxx on both Unix systems.

vgvassilev:
I think clang has a flag -stdlib or something like that and probably we need to pass it when building clad. That’d mean we will need some sort of cmake support.

mcbarton:
The original issue suggests setting
CXXFLAGS="-stdlib=libc++"

Is it enough just to pass that during the cmake config stage?

vgvassilev:
> The original issue suggests setting CXXFLAGS="-stdlib=libc++"
> 
> Is it enough just to pass that during the cmake config stage?

Probably…

mcbarton:
Looks like more work is needed on the libcxx side. I will troubleshoot it tomorrow

mcbarton:
@vgvassilev after the installation of another package to be libcxx installed on Ubuntu and the extra flag the libcxx issue appears to be fixed I believe. This PR is ready for further review.

alexander-penev (approved):

vgvassilev:
I am still not sure how we get a libcxx build on ubuntu with the code of this PR. Can you tell me which builder I should look at?

build.sh
```diff
@@ -0,0 +1,180 @@
...
+    export CC=$path_to_llvm_build/bin/clang
+    export CXX=$path_to_llvm_build/bin/clang++
+  else
+    echo 'Install llvm '$llvm_version' using apt to build Clad against'
+    os_codename="`cat /etc/os-release | grep UBUNTU_CODENAME | cut -d = -f 2`"
+    sudo apt update
+    curl https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+    echo "deb https://apt.llvm.org/${os_codename}/ llvm-toolchain-${os_codename}-${llvm_version} main" | sudo tee -a /etc/apt/sources.list
+    sudo apt update
+    sudo apt install -y clang-${llvm_version} libclang-${llvm_version}-dev libc++-${llvm_version}-dev
```
mcbarton, line 59: @vgvassilev The jobs I have added run this script. The libc++ is installed on Ubuntu on this line (see libc++-${llvm_version}-dev). I then make use of it with the following line in the cmake configure stage -DCMAKE_CXX_FLAGS="-stdlib=libc++"

vgvassilev: Can we have a special bot that contains libcxx in its name for this?

vgvassilev: We should specify the apt package list as part of the bot configuration, not hardcoding them in the script. We can have a list of default packages and only extend it when it is different.

mcbarton: So you want me to replace 
          - name: ubu22-clang-runtime18    with              - name: ubu22-clang-libcxx-runtime18
and
          - name: osx14-arm-clang-runtime17   with         - name: osx14-arm-clang-libcxx-runtime17
?

mcbarton: > We should specify the apt package list as part of the bot configuration, not hardcoding them in the script. We can have a list of default packages and only extend it when it is different.

With the current set up, this script can be used by developers outside the CI to test there build before commit. Would you rather I don't use a script like the issue suggests, and make the libcxx build as part of the ci yaml file?

mcbarton: > > We should specify the apt package list as part of the bot configuration, not hardcoding them in the script. We can have a list of default packages and only extend it when it is different.
> 
> With the current set up, this script can be used by developers outside the CI to test there build before commit. Would you rather I don't use a script like the issue suggests, and make the libcxx build as part of the ci yaml file?

I can remove the script and integrate it into the other matrix. It will just be a few days until I can get to it.

vgvassilev: I think it'd be more useful for outside developers if it did not touch their system packages and did not make changes to their system. The build steps are probably very useful but I'd be very hesitant to use a script that has sudo in it...

vgvassilev: It is probably fine to have it as part of the CI, however, some things like the used packages and extra cmake variables should be more explicitly shown in the build configuration. This way it would be easier to find where is the problem is when a PR fails.

vgvassilev:
For some reason I cannot ssh into the bot to debug what's wrong...

mcbarton:
> Unsure why the libcxx job failed when I integrated into the rest of the jobs. I'll investigate.

The job was failing when I used the bash script too, but it was passing on the ci, so I didn't notice. I don't really understand the error message enough to know how to troubleshoot this bug.

@vgvassilev I'm not sure why you can't ssh into the bot to debug.

mcbarton:
@vgvassilev maybe if you push a commit where you set debug to true then it will activate the tmate session, so you can log into the bot.

.github/workflows/ci.yml
```diff
@@ -785,8 +792,8 @@ jobs:
       if: ${{ runner.os != 'windows' }}
       run: |
         mkdir obj && cd obj
-        cmake -DClang_DIR=${{ env.PATH_TO_LLVM_BUILD }}  \
-          -DLLVM_DIR=${{ env.PATH_TO_LLVM_BUILD }}       \
+        cmake -DClang_DIR="${{ env.PATH_TO_LLVM_BUILD }}/lib/cmake/"  \
+          -DLLVM_DIR="${{ env.PATH_TO_LLVM_BUILD }}/lib/cmake/"       \
```
vgvassilev, line 796: Why did we need these changes?

mcbarton: We don't . I just haven't got around to removing it yet. It got added when debugging but hasn't been removed yet. It can be removed once a solution has been found to the error.

vgvassilev:
I was able to debug a little bit the bot. I see two problems. The first one is that system clang gets compiled with libstdc++:
```
ldd /usr/bin/clang
...
        libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fee49c00000)
...
```

Clang dlopen's clad and this means that if we compile clad with libc++ that would mean we will end up with two different standard libraries in the same process. That's probably a recipe for a disaster. 

We will need to build clang ourselves with libc++ and only then clad on top. That can be done but I question its usefulness. I wonder if there is a unix platform (different from darwin) which ships clang built with libc++. Probably not.

I see two options in moving forward - continue to struggle and maybe making it work but to what end? Or stop here, add that explanation to the issue and close it...

mcbarton:
> I was able to debug a little bit the bot. I see two problems. The first one is that system clang gets compiled with libstdc++:
> 
> ```
> ldd /usr/bin/clang
> ...
>         libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fee49c00000)
> ...
> ```
> 
> Clang dlopen's clad and this means that if we compile clad with libc++ that would mean we will end up with two different standard libraries in the same process. That's probably a recipe for a disaster.
> 
> We will need to build clang ourselves with libc++ and only then clad on top. That can be done but I question its usefulness. I wonder if there is a unix platform (different from darwin) which ships clang built with libc++. Probably not.
> 
> I see two options in moving forward - continue to struggle and maybe making it work but to what end? Or stop here, add that explanation to the issue and close it...

I suggest we go with closing the PR and issue, adding the explaination as to why it wasn't able to be implemented.

mcbarton:
Closing as a result of the outcome of the debugging showing conflicting standard libraries trying to be used (see https://github.com/vgvassilev/clad/pull/850#issuecomment-2067596340 for more details).

Issue #849, opened 2024-04-03
Support for static_asserts in primal functions.


vaithak:
Minimal example where `static_assert` fails in the generated function:
```cpp
include "clad/Differentiator/Differentiator.h"

template<typename T>
T f2(double x) {
  static_assert(std::is_floating_point<T>::value,
                    "template argument must be a floating point type");
  return x;
}

double f(double x) {
  return f2<double>(x);
}

int main() {
  auto f_d = clad::differentiate(f);
  printf("dx = %.2f", f_d.execute(3));
}
```

This is an example reproducible for #700.

PR #861, opened 2024-04-12
Remove static asserts from generated code

### LINKED ISSUE #700: Clang segfaults for program using header <random>
**Minimal example**:
```c++
#include <random>

double f(double x){
  std::mt19937 gen64;
  std::uniform_real_distribution<double> distribution(0.0,1.0);
  double rand = distribution(gen64);
  return x+rand;
}

int main(){
  auto f_dx = clad::differentiate(f, "x");
  f_dx.execute(3.14);
}
```

**Expected behaviour**: Code compiles. The derivative should evaluate to 1.

**Observed behaviour**: Clang segfaults.

**Setup**: Clad built from the GitHub master branch at commit 7449329197788a4ba523b179479d2483b66b12b8 on Ubuntu 22.04 with Clang 11. The segfault appears both with the GCC 12 libstdc++ and with the LLVM-11 libc++ (`-stdlib=libc++`).

**Further information**: The analogous code used to crash for Enzyme (EnzymeAD/Enzyme#1600), but likely for a different reason (missing support for `long double` in C++ header `<random>`), which has been fixed in EnzymeAD/Enzyme#1604.

vaithak:
fixes #700 (partially - for fwd mode) , #849

### DIFF
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -11,7 +11,6 @@
 #include "clad/Differentiator/CladUtils.h"
 #include "clad/Differentiator/DiffPlanner.h"
 #include "clad/Differentiator/ErrorEstimator.h"
-#include "clad/Differentiator/StmtClone.h"
 
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/Expr.h"
@@ -574,7 +573,7 @@ StmtDiff BaseForwardModeVisitor::VisitIfStmt(const IfStmt* If) {
 
   VarDecl* condVarClone = nullptr;
   if (const VarDecl* condVarDecl = If->getConditionVariable()) {
-    VarDeclDiff condVarDeclDiff = DifferentiateVarDecl(condVarDecl);
+    DeclDiff<VarDecl> condVarDeclDiff = DifferentiateVarDecl(condVarDecl);
     condVarClone = condVarDeclDiff.getDecl();
     if (condVarDeclDiff.getDecl_dx())
       addToCurrentBlock(BuildDeclStmt(condVarDeclDiff.getDecl_dx()));
@@ -672,7 +671,7 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
   VarDecl* condVarDecl = FS->getConditionVariable();
   VarDecl* condVarClone = nullptr;
   if (condVarDecl) {
-    VarDeclDiff condVarResult = DifferentiateVarDecl(condVarDecl);
+    DeclDiff<VarDecl> condVarResult = DifferentiateVarDecl(condVarDecl);
     condVarClone = condVarResult.getDecl();
     if (condVarResult.getDecl_dx())
       addToCurrentBlock(BuildDeclStmt(condVarResult.getDecl_dx()));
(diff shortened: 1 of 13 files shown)


lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1454,6 +1453,8 @@ StmtDiff BaseForwardModeVisitor::VisitDeclStmt(const DeclStmt* DS) {
         m_DeclReplacements[VD] = VDDiff.getDecl();
       decls.push_back(VDDiff.getDecl());
       declsDiff.push_back(VDDiff.getDecl_dx());
+    } else if (auto SAD = dyn_cast<StaticAssertDecl>(D)) {
+      return VisitStaticAssertDecl(SAD);
```
vgvassilev, line 1457: Isn't here the problem that we do not call `Visit` of `D` in the very first line of the loop?

vaithak: fixed and made it more generic 👍🏼

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -2022,4 +2023,9 @@ StmtDiff BaseForwardModeVisitor::VisitSubstNonTypeTemplateParmExpr(
     const clang::SubstNonTypeTemplateParmExpr* NTTP) {
   return Visit(NTTP->getReplacement());
 }
+
+StmtDiff BaseForwardModeVisitor::VisitStaticAssertDecl(
+    const clang::StaticAssertDecl* SAD) {
+  return nullptr;
```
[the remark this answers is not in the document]
vaithak: Because it's evaluated at compile time, I don't see it performing any side effects we may need to capture in the generated code. Maybe, we will see some use-cases with `consteval` or `contexpr` function calls.

lib/Differentiator/ErrorEstimator.cpp
```diff
@@ -481,8 +481,8 @@ void ErrorEstimationHandler::ActBeforeFinalizingVisitDeclStmt(
   // For all dependent variables, we register them for estimation
   // here.
   for (size_t i = 0; i < decls.size(); i++) {
-    VarDeclDiff VDDiff(static_cast<VarDecl*>(decls[0]),
-                       static_cast<VarDecl*>(declsDiff[0]));
+    DeclDiff<VarDecl> VDDiff(dynamic_cast<VarDecl*>(decls[0]),
```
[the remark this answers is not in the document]
vgvassilev: We should just use `cast` here.

test/Gradient/FunctionCalls.C
```diff
@@ -6,6 +6,7 @@
 // CHECK-NOT: {{.*error|warning|note:.*}}
 
 #include "clad/Differentiator/Differentiator.h"
+#include <random>
```
vgvassilev, line 9: Do we use anything from `<random>` in this test?

vaithak: my bad, removed it 👍🏼

vgvassilev (approved):
LGTM!

vgvassilev:
Can you rebase, the codecov should be fixed by #863

vgvassilev:
Can you add tests to appease codecov?

vaithak:
> Can you add tests to appease codecov?

For the locations related to StaticAssertDecl, it is future proofing for the case we start adding some declarations for static asserts in the generated code. Adding a test for them won't be possible now, as we return `nullptrs` for static asserts.

For the remaining case relation If condition with a declaration, turns out it has some bug, I opened an issue for that here: https://github.com/vgvassilev/clad/issues/865.

PR #862, opened 2024-04-13
[ci] Enable clang-format on unittests

vgvassilev:


vaithak:
Will this work? In `.clang-format`, we have explicitly specified language as Cpp, and the test files are `.C` files.
I tested this locally, and it doesn't seem to be working for the test files.

vgvassilev:
> Will this work? In `.clang-format`, we have explicitly specified language as Cpp, and the test files are `.C` files.
> I tested this locally, and it doesn't seem to be working for the test files.

If you mean the files in /test we should not format them however we should format the ones in /unittests. Is that what you mean?

vaithak:
> If you mean the files in /test we should not format them however we should format the ones in /unittests. Is that what you mean?

Yeah, I thought we meant to format `/test`

vaithak (approved):

PR #864, opened 2024-04-15
Don't create adjoints for integral type variables

PetroZarytskyi:
This PR removes adjoints for integral types in the reverse mode. This is done because integers are not differentiable in the traditional mathematical sense. Such behavior is consistent with other AD tools like Tapenade and Enzyme. Let's consider an example to see the difference in behavior
```
double f(double x) {
    int n = x;
    return n;
}
```
This is essentially a floor function so mathematically the derivative should be 0. However, Clad master produces 1. Let's compare the derivatives on master and in this PR.
master:
```
void f_grad(double x, double *_d_x) {
    int _d_n = 0;
    int n = x;
    goto _label0;
  _label0:
    _d_n += 1;
    *_d_x += _d_n;
}
```
this PR:
```
void f_grad(double x, double *_d_x) {
    int n = x;
    goto _label0;
  _label0:
    ;
}
```
Doing this not only gives us the right results but also removes unnecessary variable declarations.

However, this PR has two major problems: 
1) Removing specific adjoints from the gradient signature requires using gradient adjoints that are not supported in error estimation. So without major changes, it does not seem possible to remove integral type adjoints from the gradient signature in error estimation.
2) There were some doubts about whether we want this behavior in Clad. If we decide to implement this, we need to implement this system in forward mode as well.

This PR needs a discussion to move forward.


vgvassilev:
Since this is largely backward incompatible change, can you open this PR against the `ver-2.0` branch?

PetroZarytskyi:
The decision to close this PR is explained in detail in #1551.

PR #868, opened 2024-04-25
Moving DerivedFnCollector out of ClangPlugin

vaithak:
This is part of the initial refactoring required to remove direct access to ClangPlugin methods.


include/clad/Differentiator/DerivedFnCollector.h
```diff
@@ -0,0 +1,40 @@
+#ifndef CLAD_DIFFERENTIATOR_DERIVEDFNCOLLECTOR_H
```
vgvassilev, line 1: Can we move that header file and its friends in `lib`?

vaithak: Any specific reason as to why we want to move this header file into lib? I thought all header files go into include directory

vgvassilev: These header files are "local" files that are related to the build artifact not the user interface. That is, we do not need to have them installed for clad to work...

tools/ClangPlugin.cpp
```diff
@@ -122,7 +122,7 @@ namespace clad {
       Sema& S = m_CI.getSema();
 
       if (!m_DerivativeBuilder)
-        m_DerivativeBuilder.reset(new DerivativeBuilder(S, *this));
+        m_DerivativeBuilder.reset(new DerivativeBuilder(S, *this, m_DFC));
```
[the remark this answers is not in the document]
vaithak: not sure if we want to apply this suggestion?

include/clad/Differentiator/DerivedFnCollector.h
vaithak: I tried it, but the issue is that this file is imported by `DerivativeBuilder.h`, which is further imported by all the Visitors.

vgvassilev: Ok, fair enough.

tools/ClangPlugin.cpp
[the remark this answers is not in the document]
vgvassilev: No, that’s only available in c++14 onward.

vgvassilev (approved):
LGTM!

Issue #745, opened 2024-01-31
Possible memory leak when differentiating call expressions in forward mode


vaithak:
Minimal reproducible example:
```cpp
#include "clad/Differentiator/Differentiator.h"

double g (double x) { return x; }

double f (double x) {
  const int __k = 9;
  return g(x);
}

int main () {
  auto f_dx = clad::differentiate(f);
  return 0;
}
```

error msg: `error: reference to local variable '_d___k' declared in enclosing function 'f_darg0'`

This was found during the debugging of #700, and is related to the issue in the `mersenne_twister.h` file.

Issue #857, opened 2024-04-10
Avoid access of plugin methods from Visitors


vaithak:
We should restructure our code to ensure that methods like `AddRequestToSchedule` and `ProcessDiffRequest` are not required to be called by `Visitors` methods. Instead, we can make the DiffSchedule array accessible to these visitors.

This was discussed briefly here: https://github.com/vgvassilev/clad/pull/848#discussion_r1554655184

vaithak:
After #873, the only remaining method is `ProcessDiffRequest` which is used for forward declaration of functions. This can be removed once we are able to capture the Differentiation requests graph in the first pass itself, which would require assistance from activity analysis to figure out the complete requests object even before starting any differentiation.

PR #873, opened 2024-04-26
Compute and process Differentiation Request graph

vaithak:
## Plan for dynamic graph
- The relations between different differentiation requests can be modelled as a graph.
For example, if `f_a` calls `f_b`, there will be two differentiation requests `df_a` and `df_b`, the edge between them can be understood as `created_because_of`. 
This also means that the functions called by the users to be explicitly differentiated (or `DiffRequests` created because of these) are the source nodes, i.e. no incoming edges. In most cases, this graph aligns with the call graph, but in some cases, the graph depends on the internal implementation, like the Hessian computation, which requires creating multiple `fwd_mode` requests followed by a `rev_mode` request.

- We can use this graph to order the computation of differentiation requests. This was already being done implicitly in the initial recursive implementation. Whenever we encountered a call expression, we started differentiation of the
called function; this was sort of like a depth-first search strategy.
  - This had problems, as `Clang` reported errors when it encountered a new function scope (of the derivative of the called function) in the middle of the old function scope (of the derivative of the callee function). It treated the nested one like a lambda expression. The issue regarding this: https://github.com/vgvassilev/clad/issues/745.

- To fix this, an initial strategy was to eliminate the recursive approach. Hence, a queue-based breadth-first approach was implemented in this PR: https://github.com/vgvassilev/clad/pull/848.
  - Although it fixed the problem, the graph traversal was still implicit. We needed some way to compute/store the complete graph and possibly optimize it, such as converting edges to model the `requires_derivative_of` relation. Using this, we could proceed with differentiation in a topologically sorted ordering.
  - It also required one caveat: although we don't differentiate the called function completely in a recursive way, we still need to declare it so that we can have the call expression completed (i.e. `auto t0 = f_pushforward(...)`).

- To move towards the final stage of having the complete graph computed before starting the differentiation, we need the complete information on how the `DiffRequest` will be formed inside the visitors (including arguments or `DVI` info).
This whole approach will require activity analysis in the first pass.
  - As an incremental improvement, the first requirement was to implement infrastructure to support explicit modelling of the graph and use that to have a breadth-first traversal (and eventually topological ordering).

This is the initial PR for capturing the differentiation plan in a graphical format.

However, the traversal order is still breadth-first, as we don't have the complete graph in the first pass - mainly because of a lack of information about the args required for `pushforward` and `pullbacks`.

This can be improved with the help of activity analysis to capture the complete graph in the first pass, processing the plan in a topologically sorted manner and pruning the graph for user-defined functions. I started this with this approach, and the initial experimental commit is available here for future reference: https://github.com/vaithak/clad/commit/82c0b4250c1f0399b21ebbc71e50b52a218eae27.

### DIFF
--- a/include/clad/Differentiator/DiffPlanner.h
+++ b/include/clad/Differentiator/DiffPlanner.h
@@ -95,9 +95,7 @@ struct DiffRequest {
   /// Define the == operator for DiffRequest.
   bool operator==(const DiffRequest& other) const {
     // either function match or previous declaration match
-    return (Function == other.Function ||
-            Function->getPreviousDecl() == other.Function ||
-            Function == other.Function->getPreviousDecl()) &&
+    return Function == other.Function &&
            BaseFunctionName == other.BaseFunctionName &&
            CurrentDerivativeOrder == other.CurrentDerivativeOrder &&
            RequestedDerivativeOrder == other.RequestedDerivativeOrder &&
--- a/include/clad/Differentiator/DiffPlanner.h
+++ b/include/clad/Differentiator/DiffPlanner.h
@@ -1,10 +1,11 @@
 #ifndef CLAD_DIFF_PLANNER_H
 #define CLAD_DIFF_PLANNER_H
 
-#include "clad/Differentiator/DiffMode.h"
-#include "clad/Differentiator/ParseDiffArgsTypes.h"
 #include "clang/AST/RecursiveASTVisitor.h"
 #include "llvm/ADT/SmallSet.h"
+#include "clad/Differentiator/DiffMode.h"
+#include "clad/Differentiator/DynamicGraph.h"
+#include "clad/Differentiator/ParseDiffArgsTypes.h"
 
 namespace clang {
   class ASTContext;
@@ -90,9 +91,33 @@ struct DiffRequest {
   ///   3) If no argument is provided, a default argument is used. The
   ///      function will be differentiated w.r.t. to its every parameter.
   void UpdateDiffParamsInfo(clang::Sema& semaRef);
+
+  /// Define the == operator for DiffRequest.
+  bool operator==(const DiffRequest& other) const {
+    // either function match or previous declaration match
+    return (Function == other.Function ||
+            Function->getPreviousDecl() == other.Function ||
+            Function == other.Function->getPreviousDecl()) &&
+           BaseFunctionName == other.BaseFunctionName &&
+           CurrentDerivativeOrder == other.CurrentDerivativeOrder &&
+           RequestedDerivativeOrder == other.RequestedDerivativeOrder &&
+           CallContext == other.CallContext && Args == other.Args &&
+           Mode == other.Mode && EnableTBRAnalysis == other.EnableTBRAnalysis &&
+           DVI == other.DVI && use_enzyme == other.use_enzyme &&
+           DeclarationOnly == other.DeclarationOnly;
+  }
+
+  // String operator for printing the node.
+  operator std::string() const {
+    std::string res = BaseFunctionName + "__order_" +
+                      std::to_string(CurrentDerivativeOrder) + "__mode_" +
+                      DiffModeToString(Mode);
+    if (EnableTBRAnalysis)
+      res += "__TBR";
+    return res;
+  }
 };
 
-  using DiffSchedule = llvm::SmallVector<DiffRequest, 16>;
   using DiffInterval = std::vector<clang::SourceRange>;
 
   struct RequestOptions {
@@ -106,9 +131,9 @@ struct DiffRequest {
     ///
     DiffInterval& m_Interval;
 
-    /// The diff step-by-step plan for differentiation.
+    /// Graph to store the dependencies between different requests.
     ///
-    DiffSchedule& m_DiffPlans;
+    clad::DynamicGraph<DiffRequest>& m_DiffRequestGraph;
 
     /// If set it means that we need to find the called functions and
     /// add them for implicit diff.
@@ -120,12 +145,24 @@ struct DiffRequest {
 
   public:
     DiffCollector(clang::DeclGroupRef DGR, DiffInterval& Interval,
-                  DiffSchedule& plans, clang::Sema& S, RequestOptions& opts);
+                  clad::DynamicGraph<DiffRequest>& requestGraph, clang::Sema& S,
+                  RequestOptions& opts);
     bool VisitCallExpr(clang::CallExpr* E);
 
   private:
     bool isInInterval(clang::SourceLocation Loc) const;
   };
 }
+
+// Define the hash function for DiffRequest.
+template <> struct std::hash<clad::DiffRequest> {
+    std::size_t operator()(const clad::DiffRequest& DR) const {
+      // Use the function pointer as the hash of the DiffRequest, it
+      // is sufficient to break a reasonable number of collisions.
+      if (DR.Function->getPreviousDecl())
+        return std::hash<const void*>{}(DR.Function->getPreviousDecl());
+      return std::hash<const void*>{}(DR.Function);
+    }
+};
 
 #endif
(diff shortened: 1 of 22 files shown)


vaithak:
@vgvassilev One improvement that we can make in this PR itself is to fix the printing order of generated functions; this can be done because we have dynamically created the graph during traversals.
This will help us minimize (or in some cases complete remove) th
Do you think it would be a good idea?

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -106,9 +145,9 @@ struct DiffRequest {
     ///
     DiffInterval& m_Interval;
 
-    /// The diff step-by-step plan for differentiation.
+    /// Graph to store the dependencies between different requests.
     ///
-    DiffSchedule& m_DiffPlans;
+    clad::DynamicGraph<DiffRequest>& m_DiffRequestGraph;
```
vgvassilev, line 150: Can we use a forward declaration here and make `DynamicGraph.h` a private header?

vaithak:
> I think so, is there a catch?

The only issue I see is that diagnostic messages won't align with the sequence of generated functions in the code dump. So, maybe it's a little bit tough for debugging.

vgvassilev:
> > I think so, is there a catch?
> 
> The only issue I see is that diagnostic messages won't align with the sequence of generated functions in the code dump. So, maybe it's a little bit tough for debugging.

Hm... I am confused - how is it different from now?

vaithak:
> Hm... I am confused - how is it different from now?

Currently, if the order of differentiation is f1 followed by f2. Then, any diagnostic warning or error in f2 will appear after f1 has been dumped/printed and before f2 is dumped.
This won't be true anymore.

vgvassilev:
> > Hm... I am confused - how is it different from now?
> 
> Currently, if the order of differentiation is f1 followed by f2. Then, any diagnostic warning or error in f2 will appear after f1 has been dumped/printed and before f2 is dumped. This won't be true anymore.

Let's have this improvement in a separate PR and then we can discuss further. What do you think?

vaithak:
> Let's have this improvement in a separate PR and then we can discuss further. What do you think?

Sounds good 👍🏼

include/clad/Differentiator/DiffPlanner.h
vgvassilev: Ok, fair enough...

include/clad/Differentiator/DynamicGraph.h
```diff
@@ -0,0 +1,158 @@
...
+#include <queue>
+#include <set>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+namespace clad {
+template <typename T> class DynamicGraph {
+private:
+  // Storing nodes in the graph. The index of the node in the vector is used as
```
vgvassilev, line 16: It's probably worth keeping doxygen-style comments with `///`.

include/clad/Differentiator/DynamicGraph.h
```diff
@@ -0,0 +1,158 @@
...
+        dfs(source);
+
+    if (reverseOrder)
+      return res;
+    std::reverse(res.begin(), res.end());
+    return res;
+  }
+
+  // Get the next to process node from the queue of nodes to be processed.
+  T getNextToProcessNode() {
```
vgvassilev, line 148: Do we want to make hand over a forward iterator so that it becomes friendlier to client code?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -684,7 +685,7 @@ namespace clad {
       llvm::SaveAndRestore<const FunctionDecl*> saveTopMost = m_TopMostFD;
       m_TopMostFD = FD;
       TraverseDecl(derivedFD);
-      m_DiffPlans.push_back(std::move(request));
+      m_DiffRequestGraph.addNode(request, true /*isSource*/);
```
vgvassilev, line 688:
```suggestion
      m_DiffRequestGraph.addNode(request, /*isSource=*/true);
```
There must be a clang-tidy check, why it did not warn here - can you check?

test/Misc/DynamicGraph.C
```diff
@@ -0,0 +1,70 @@
+// RUN: %cladclang %s -I%S/../../include -oGraph.out 2>&1
```
vgvassilev, line 1: Isn't that more suitable to become a unittest. There we could use all gtest facilities to test the class?

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -59,25 +60,40 @@ namespace clad {
     IndependentArgRequest.UpdateDiffParamsInfo(SemaRef);
     // FIXME: Find a way to do this without accessing plugin namespace functions
     FunctionDecl* firstDerivative =
-        plugin::ProcessDiffRequest(CP, IndependentArgRequest);
+        Builder.FindDerivedFunction(IndependentArgRequest);
+    if (!firstDerivative) {
+      // Derive declaration of the the forward mode derivative.
+      IndependentArgRequest.DeclarationOnly = true;
+      firstDerivative = plugin::ProcessDiffRequest(CP, IndependentArgRequest);
```
vgvassilev, line 67: This PR does not get rid of `plugin::ProcessDiffRequest`, but do we have a plan to remove it. I think that's a layering violation.

test/Misc/DynamicGraph.C
vaithak: Created a unit test and removed this 👍🏼

lib/Differentiator/DiffPlanner.cpp
vaithak: Even after a lot of trials, I can't figure it out. I tried this option with multiple configurations of its options: https://clang.llvm.org/extra/clang-tidy/checks/bugprone/argument-comment.html.
Still doesn't work, changed it manually for now.

lib/Differentiator/HessianModeVisitor.cpp
vaithak: The only reason we need this is for the forward declaration of derivative signatures to complete the call expressions. Once we get the graph in the first pass itself (with the help of activity analysis), we can proceed with topologically ordered differentiation (instead of breadth-first as we are doing now), and then this forward declaration will not be required.

include/clad/Differentiator/DynamicGraph.h
vaithak: The issue with iterator is that given the dynamic nature of the graph, the iterators can become invalid, which will result in segfaults.

vgvassilev: ah, ok, good point.

include/clad/Differentiator/DynamicGraph.h
```diff
@@ -0,0 +1,136 @@
...
+  /// @param dest
+  void addEdge(const T& src, const T& dest) {
+    std::pair<bool, size_t> srcInfo = addNode(src);
+    std::pair<bool, size_t> destInfo = addNode(dest);
+    size_t srcId = srcInfo.second;
+    size_t destId = destInfo.second;
+    m_adjList[srcId].insert(destId);
+  }
+
+  /// @brief Add a node to the graph. If the node is already present, return the
```
vgvassilev, line 53: We should omit `brief` as it is default.

include/clad/Differentiator/DynamicGraph.h
```diff
@@ -0,0 +1,136 @@
...
+  /// id of the node in the graph. If the node is a source node, add it to the
+  /// queue of nodes to be processed.
+  /// @param node
```
vgvassilev, line 56:
```suggestion
  /// \param node
```

That's common elsewhere in the codebase and llvm.

lib/Differentiator/HessianModeVisitor.cpp
vgvassilev: Can we keep track of that discussion as an issue?

vgvassilev (approved):
LGTM!

lib/Differentiator/HessianModeVisitor.cpp
vaithak: We already have a related issue here: https://github.com/vgvassilev/clad/issues/857, added a comment in it referencing this PR.

PR #875, opened 2024-04-28
Add `check-clad-exec-only` target

### LINKED ISSUE #874: Add a target for verifying only `CHECK-EXEC` in tests
`check-clad` target verifies both the derivative code and derivative output. It's difficult to use this target when we are making Clad modifications that are expected to change the derivative code. It would be nice to have a separate target which only checks the derivative outputs.

parth-07:
This commit adds a new verification target, 'check-clad-exec-only'. This target only verifies the 'CHECK-EXEC' filecheck checks. It is helpful during the development phase when derivative code is expected to change frequently.

Fixes #874


test/lit.cfg
```diff
@@ -199,9 +199,19 @@ for pattern in [r"\bbugpoint\b(?!-)",   r"(?<!/)\bclang\b(?!-)",
...
 excludes = []
 
+filecheck_sub = next((sub for sub in config.substitutions if 'FileCheck' in sub[0]), None)
+config.substitutions.append(('%filecheck_exec', 
+    '{filecheck} --check-prefix=CHECK-EXEC'.format(filecheck = filecheck_sub[1])))
+if config.filecheck_check_exec_only:
+    # Ideally, just 'true', instead of 'xargs -0 true', should work as well.
+    # However, due to some lit-specific issue piping to true from Clang always
+    # returns an error.
+    config.substitutions.append(('%filecheck', 'xargs -0 true'))
```
vaithak, line 211: Will this work if some test file contains some options for `FileCheck`? Those will get appended as `xargs` arguments/options, right?

vaithak (approved):

test/CMakeLists.txt
```diff
@@ -24,6 +24,8 @@ if(CLAD_BUILT_STANDALONE)
   set(LLVM_LIBS_DIR ${LLVM_INSTALL_PREFIX}/lib/)
 endif()
 
+set(CLAD_FILECHECK_CHECK_EXEC_ONLY False)
```
vgvassilev, line 27: Why do we need a separate option here? Can't we just dial the target with `make check-clad-exec-only`?

test/CMakeLists.txt
```diff
@@ -36,6 +38,12 @@ configure_lit_site_cfg(
   ${CMAKE_CURRENT_SOURCE_DIR}/Unit/lit.cfg.py
   )
 
+set(CLAD_FILECHECK_CHECK_EXEC_ONLY True)
+configure_lit_site_cfg(
+  ${CMAKE_CURRENT_SOURCE_DIR}/lit.site.cfg.in
+  ${CMAKE_CURRENT_BINARY_DIR}/CheckExecOnly/lit.site.cfg
```
vgvassilev, line 44: Can we get away without creating a new folder? I suspect that you could pass an extra configuration to `lit`. Using `lit --debug` might be able to hint you how to do it...

test/CMakeLists.txt
```diff
@@ -74,6 +82,14 @@ add_lit_testsuite(check-clad "Running the Clad regression tests"
   )
 set_target_properties(check-clad PROPERTIES FOLDER "Clad tests")
 
+add_lit_testsuite(check-clad-exec-only "Running the Clad regression tests"
```
vgvassilev, line 85:
```suggestion
add_lit_testsuite(check-clad-exec-only "Running the Clad execution regression tests"
```

vgvassilev (commented):
That's quite handy. Thank you! I bet @PetroZarytskyi will appreciate that maybe more than anybody else ;)

test/lit.cfg
parth-07: Yes, this should work in this case. The arguments to `FileCheck` will be passed to `true` instead of `xargs`.

test/CMakeLists.txt
parth-07: This option is used for configuring `lit.cfg` for the two cases -- check derivative code and don't check derivative code. Now we have two `lit.site.cfg` files that represents these two cases: `test/lit.site.cfg` and `test/CheckExecOnly`.

This option is used so that we set the correct value of `config.filecheck_check_exec_only` when we call `configure_lit_site_cfg`.

test/CMakeLists.txt
parth-07: This seems interesting. I will look into it.

vgvassilev: @parth-07, I've pushed some changes which make the implementation a bit more flexible. Can you take a look?

parth-07 (commented):
Looks good to me. @vgvassilev Thank you for the changes. Your changes made the functionality much more flexible and clean.

PR #877, opened 2024-05-04
Provide a backup plan if we fail to delay the handling of declarations. 

vgvassilev:
This patch enhances the work done in https://github.com/vgvassilev/clad/pull/766 where we assumed the clients are going to be well-behaved such as clang or clang-repl. However, we can have clients which connect the clang components in a non-standard way and we should provide a reasonable fallback.

This patch detects if the setup is non-standard and teaches clad to handle requests as they come, pretty much the same way as before https://github.com/vgvassilev/clad/pull/766 was implemented. Of course this comes with a cost when we try to differentiate declarations that are defined later in the translation unit. We expect such setups to be rare and mostly in the cases of incremental processing and old repls such as Cling.

Depends on #876.

### DIFF
--- a/tools/ClangPlugin.h
+++ b/tools/ClangPlugin.h
@@ -23,6 +23,7 @@
 
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/Casting.h"
 #include "llvm/Support/Timer.h"
 
 namespace clang {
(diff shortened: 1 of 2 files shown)


vgvassilev:
I am not sure if we can test this patch - it will need to have either cling or some sort of cling-like infrastructure which is probably not worth maintaining in a long run.

tools/ClangPlugin.h
```diff
@@ -258,6 +268,10 @@ class CladTimerGroup {
       DelayedCallInfo DCI{CallKind::HandleTopLevelDecl, D};
       assert(!llvm::is_contained(m_DelayedCalls, DCI) && "Already exists!");
       AppendDelayed(DCI);
+      // We could not delay the process due to some strange way of
+      // initialization, inform the consumers now.
+      if (!m_Multiplexer)
+        m_CI.getASTConsumer().HandleTopLevelDecl(DCI.m_DGR);
```
vaithak, line 274: If we are not delaying the call, do we still want to add it to `AppendDelayed`?

vgvassilev: I’d think so. That’d give us some uniform debugging capabilities for both cases….

vaithak (approved):

PR #882, opened 2024-05-07
Update actions to latest version (part of migration from node 16 to node 20)

mcbarton:
This PR should update all actions which cause warning about the migration from node 16 to node 20), with the exception of branch name.


vgvassilev (approved):
Thank you. LGTM!

PR #883, opened 2024-05-07
Write numerical diff results to c-style arrays in the generated code.

PetroZarytskyi:
Currently, the numerical diff generated inside gradients is unable to handle different type arguments. As an example, let's differentiate a call to  "double ff(double, unsigned int);":
```
x = ff(a, n);
```
-->
```
...
clad::tape<clad::array_ref<double> > _t1 = {};
double _grad0 = 0;
clad::push(_t1, &_grad0);
unsigned int _grad1 = 0;
clad::push(_t1, &_grad1);
numerical_diff::central_difference(ff, _t1, 0, a, n);
...
```
So ``&_grad1``(``unsigned int*``) is pushed to ``_t1`` (``clad::tape<clad::array_ref<double>>``). The obvious solution here would be to only use the type of the tape element. In this case, there's no point in creating separate ``_grad``, it would make more sense to create an array. This is the solution suggested in this PR:
```
x = ff(a, n);
```
-->
```
 double _grad0[2] = {0};
 numerical_diff::central_difference(ff, _grad0, 0, a, n);
```
One could argue this will only work for numerical type arguments but other types are not supported in reverse mode numerical diff anyway.

### DIFF
--- a/include/clad/Differentiator/NumericalDiff.h
+++ b/include/clad/Differentiator/NumericalDiff.h
@@ -326,6 +326,76 @@ namespace numerical_diff {
     }
   }
 
+  /// A helper function to calculate the numerical derivative of a target
+  /// function.
+  ///
+  /// \param[in] \c f The target function to numerically differentiate.
+  /// \param[out] \c _grad The gradient array reference to which the gradients
+  /// will be written.
+  /// \param[in] \c printErrors A flag to decide if we want to print numerical
+  /// diff errors estimates.
+  /// \param[in] \c idxSeq The index sequence associated with
+  /// the input parameter pack.
+  /// \param[in] \c args The arguments to the function to differentiate.
+  template <typename F, std::size_t... Ints,
+            typename RetType = typename clad::function_traits<F>::return_type,
+            typename... Args>
+  void central_difference_helper(F f, RetType* _grad, bool printErrors,
+                                 clad::IndexSequence<Ints...> idxSeq,
+                                 Args&&... args) {
+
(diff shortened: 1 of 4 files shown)


vgvassilev:
@mcbarton, looks like we got `Error: Unsupported version for platform (os=linux, arch=x64, version=18.1.3)!` for the clang-tidy check.

mcbarton:
@vgvassilev The latest commit added support for 18.1.3 https://github.com/KyleMayes/install-llvm-action/commits/master/ . It can also be seen here for v2.0.2 release. Try changing @v2 to @v2.0.2 .

vgvassilev:
> @vgvassilev I'm not sure why that didn't work. It is supported according to the github action release. Try an earlier version of 18, or another version of llvm to see what happens.

These things are a can of worms. Maybe we should move to the earliest version possible?

mcbarton:
https://github.com/KyleMayes/install-llvm-action/releases/tag/v2.0.0 v2 removed support for llvm 7, so any version after that should work @vgvassilev

vgvassilev:
Maybe trying to bump it to 16 if we were using 15?

mcbarton:
@vgvassilev Its probably best for now to revert the changes with regards to clang tidy. I will do some experimention on how to fix this in CppInterOp . I will modify the code to remove the warnings when building, which should activate the clang tidy and allow me to be sure it works before committing it.

vgvassilev (commented):
This looks reasonable to me. Why the PR is a draft?

vgvassilev (approved):
LGTM!

PR #886, opened 2024-05-08
Fix clang tidy ci

mcbarton:
@vgvassilev based on the PR in CppInterOp this should fix the clang tidy issue. Its at least running in CppInterOp again after these changes. It seems the one action now has issues taking multiple commands.


mcbarton:
@vgvassilev To avoid any new merges which don't work ust like CppInterOp and xeus-cpp I have made a change to the code to test that the review workflow now works. I just removed a line so it activates.

mcbarton:
@vgvassilev The clang tidy ci now runs with these changes.

PR #887, opened 2024-05-08
Fix paranthesis for derivative of division operator

### LINKED ISSUE #880: Non-compilable code output for division operation in reverse mode
Minimal code:
```cpp

double f(double x) {
  return -0.5/x;
}

int main() {
  clad::gradient(f);
  return 0;
}
```

The output derivative code is:
```cpp
void f_grad(double x, double *_d_x) {
    goto _label0;
  _label0:
    {
        double _r0 = 1 * --0.5 / (x * x);  // <-------- error - consecutive (--) operators do not compile.
        *_d_x += _r0;
    }
}
```

vaithak:
fixes #880


vgvassilev:
@vaithak, can you fix the cuda test by adding parenthesis, too?

vaithak:
> @vaithak, can you fix the cuda test by adding parenthesis, too?

Yup, will do

PR #889, opened 2024-05-08
Add llvm 18 for osx to ci

mcbarton:
This PR adds llvm 18 to the ci for osx

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,11 +48,11 @@ jobs:
             os: macos-14
             compiler: clang
             clang-runtime: '17'
-          #Turn on once clang-18 available through Homebrew
-          #- name: osx14-arm-clang-runtime18
-          #  os: macos-14
-          #  compiler: clang
-          #  clang-runtime: '18'
+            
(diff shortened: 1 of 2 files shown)


vgvassilev:
@mcbarton, what is the fate of this PR?

mcbarton:
@vgvassilev I'll take another look at this PR in the coming week (got sidetracked with the work on xeus-cpp). I just need to make the ci avoid curling a binary from the llvm project Github page, and it should work.

mcbarton:
@vgvassilev are you able to take a look at this, or have someone else who can finish it off? Clad is building and the tests are running with Clang 18. I'm just a little busy at the moment to work out why they are failing and fix them (I also don't know which are the ones which are expected to fail).

vgvassilev:
Looks like there is a single failure that might be hard to debug. Perhaps we should add a debug build based on clang-18 so that we can debug such issues on the bots.

mcbarton:
@vaithak Any ideas what the issue with this PR here, based on the output from the ci? At least one of the issues seems to have been introduced in the PR you was part of here https://github.com/vgvassilev/clad/pull/873

Issue #865, opened 2024-04-15
Error in gradients for `if conditions` with declarations


vaithak:
Minimal example:
```cpp
#include "clad/Differentiator/Differentiator.h"
#include <iostream>

double f(double x, double y, int flag) {
  if (int cond = flag) 
    return y*y;
  return x*x + y*y;
}

int main()
{
    auto f_dx = clad::gradient(f, "x,y");
    double dx = 0, dy = 0;
    f_dx.execute(3, 4, 1, &dx, &dy);
    std::cout << "(dx, dy): (" << dx << ", " << dy << ")\n";
    return 0;
}
```
error: `reference to local variable 'cond' declared in enclosing function 'f'`

This can be somewhat related to https://github.com/vgvassilev/clad/issues/273.

PR #891, opened 2024-05-14
Handle variable declarations in conditions of if-statements (#865)

gojakuch:
This patch handles variable declarations in conditions of if-statements in reverse mode differentiation. This is done by actually visiting the conditions instead of just cloning them inside the VisitIfStmt and by choosing the condition variable DeclStmt over cond when possible. This also helps to handle conditions that may affect derivatives as a side-effect. Also a couple of blocks have been added to wrap some statements and preserve the correct logic in the constructed derivatives and some unused code has been removed from the method.

Fixes: #865

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -800,14 +800,42 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     // to this scope.
     beginScope(Scope::DeclScope | Scope::ControlScope);
 
-    StmtDiff cond = Clone(If->getCond());
     // Condition has to be stored as a "global" variable, to take the correct
     // branch in the reverse pass.
     // If we are inside loop, the condition has to be stored in a stack after
     // the if statement.
     Expr* PushCond = nullptr;
     Expr* PopCond = nullptr;
-    auto condExpr = Visit(cond.getExpr());
+    // Create a block "around" if statement, e.g:
+    // {
+    //   ...
+    //  if (...) {...}
+    // }
+    beginBlock(direction::forward);
+    beginBlock(direction::reverse);
+    StmtDiff condDiff;
+    // if the statement has an init, we process it
+    if (If->hasInitStorage()) {
+      StmtDiff initDiff = Visit(If->getInit());
+      addToCurrentBlock(initDiff.getStmt(), direction::forward);
+      addToCurrentBlock(initDiff.getStmt_dx(), direction::reverse);
+    }
+    // this ensures we can differentiate conditions that affect the derivatives
+    // as well as declarations inside the condition:
+    beginBlock(direction::reverse);
+    if (const auto* condDeclStmt = If->getConditionVariableDeclStmt())
+      condDiff = Visit(condDeclStmt);
+    else
+      condDiff = Visit(If->getCond());
+    CompoundStmt* RCS = endBlock(direction::reverse);
+    if (!RCS->body_empty()) {
+      std::reverse(
+          RCS->body_begin(),
+          RCS->body_end()); // it is reversed in the endBlock() but we don't
+                            // actually need this, so we reverse it once again
+      addToCurrentBlock(RCS, direction::reverse);
+    }
+
     if (isInsideLoop) {
       // If we are inside for loop, cond will be stored in the following way:
       // forward:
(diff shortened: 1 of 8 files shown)


gojakuch:
Do we need special tests to be added for these features? I can do this in this PR as well

vgvassilev:
How about "Handle variable declarations in conditions of if-statements", then explain what this patch does and finish with fixes:...

We will need to update the PR description and title accordingly.

vgvassilev:
You need to update the commit message accordingly, too.

parth-07 (changes requested):
Can you please add a test where if-condition has a side-effect that affects differentiation result?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -800,51 +800,65 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+
     if (isInsideLoop) {
-      // If we are inside for loop, cond will be stored in the following way:
-      // forward:
-      // _t = cond;
-      // if (_t) { ... }
-      // clad::push(..., _t);
-      // reverse:
+      // If we are inside for loop, condDiff will be stored in the following
+      // way: forward: _t = cond; if (_t) { ... } clad::push(..., _t); reverse:
```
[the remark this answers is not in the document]
gojakuch: it has been changed automatically by my clang format for some reason

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -902,37 +898,21 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     StmtDiff thenDiff = VisitBranch(If->getThen());
     StmtDiff elseDiff = VisitBranch(If->getElse());
 
-    // It is problematic to specify both condVarDecl and cond thorugh
-    // Sema::ActOnIfStmt, therefore we directly use the IfStmt constructor.
```
vaithak, line 906: Do we need to delete this comment?

gojakuch: condVarDecl is always set to nullptr here now, because it doesn't change anything, so I'm not sure if this is relevant. we generally don't need condVarDecl in the generated statement, only cond

vaithak: So, would it make sense to use `ActOnIfStmt` now?

lib/Differentiator/ReverseModeVisitor.cpp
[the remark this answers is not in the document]
vaithak: maybe changing it into a multiline comment style (/*  ... */) should fix it.

vaithak (approved):
Other than some clang formatting issues, this looks good to me.

lib/Differentiator/ReverseModeVisitor.cpp
gojakuch: I think it's possible, but I'm not sure what benefit that would have.

vgvassilev:
Can we test the c++20 syntax:
```cpp
if (int a = 12; a < 0) {
}
```

test/Gradient/Gradients.C
```diff
@@ -850,4 +930,12 @@ int main() {
   INIT_GRADIENT(fn_div);
   dx = 0;
   TEST_GRADIENT(fn_div, /*numOfDerivativeArgs=*/1, 2, &dx); // CHECK-EXEC: 0.12
+
+  INIT_GRADIENT(fn_cond_decl);
+
+  INIT_GRADIENT(fn_cond_side_eff);
+  dx = 0;
```
parth-07, line 937: You don't need to explicitly reset the value to 0. `TEST_GRADIENT` takes care of that.

parth-07 (approved):
Looks good to me.

This patch does not handle if-condition side-effects for all cases. The below example gives incorrect derivative:

```cpp
#include <iostream>

#define show(x) std::cout << #x <<": " << x << "\n";

    double res = 0;
    double foo = v;
    res = foo;
    res += foo;
    if ((foo = u) > 10) {
        res += foo * v;
    }
    return res + foo;
}

int main() {
    double u, v;
    u = 3, v = 5;
    double du, dv;
    du = dv = 0;
    show(du);
    show(dv);
}
```
Output:
du: 0
dv: 3

Expected Output:
du: 1
dv: 2

This case can be handled separately in a different pull-request. Can you please create an issue for this example?

test/Gradient/Gradients.C
gojakuch: oh, I didn't know this, thanks

gojakuch:
> This patch does not handle if-condition side-effects for all cases
yeah I was afraid that it would be the case but the simple things I tested worked out. I'll open up an issue about it. at least some of them work now.

gojakuch:
> Can we test the c++20 syntax:
> 
> ```c++
> if (int a = 12; a < 0) {
> }
> ```

this doesn't work out well unfortunately. but it seems fixable.

gojakuch:
> > ```c++
> > if (int a = 12; a < 0) {
> > }
> > ```

fixed it

vgvassilev:
> > > ```c++
> > > if (int a = 12; a < 0) {
> > > }
> > > ```
> 
> fixed it

I do not see a test case.

gojakuch:
> I do not see a test case.

already added

PR #893, opened 2024-05-16
Only register ClangPlugin if clad is found

hahnjo:


vgvassilev:
Hi @hahnjo, thanks a lot for this. Any chance to figure out how to test it?

hahnjo:
Not really; it happens inside a ROOT build, see the linked PR and issue above. Right now, we would have an empty string in the list of plugins, which leads to a (currently unchecked) `Expected` with an error.

vgvassilev (approved):
LGTM! XPASS-ing tests are unrelated - the clang folks fixed clang-18 and we need to adapt in a subsequent PR I guess.

Issue #890, opened 2024-05-13
Regression in support of functions with unknown definition


guitargeek:
Clad doesn't work with functions where is doesn't know the definition anymore.

Reproducer:
```c++
// clang++ -fPIC mylib.cxx -c -o mylib.o && ar r libmylib.a mylib.o

double foo(double x, double alpha, double theta, double x0 = 0)
{
   return x * alpha * theta * x0;
}
```
```c++
// Compile with clang++ -O2 -std=c++20 -fplugin=/usr/lib/clad.so repro.cxx -o repro -lmylib -L.

#include "clad/Differentiator/Differentiator.h"

#include <iostream>
#include <vector>

double foo(double x, double alpha, double theta, double x0 = 0);

double wrapper(double *params)
{
   const double ix = 1 + params[0];
   return foo(10., ix, 1.0);
}

int main()
{
   std::vector<double> params{4};

   std::cout << "Value" << std::endl;
   const double nllVal = wrapper(params.data());
   std::cout << nllVal << std::endl;
   std::cout << std::endl;

   auto grad = clad::gradient(wrapper, "params");
}
```

The error:
```txt
warning: Falling back to numerical differentiation for 'foo' since no suitable overload was found and clad could not derive it. To disable this feature, compile your programs with -DCLAD_NO_NUM_DIFF.
In file included from repro.cxx:3:
/usr/include/clad/Differentiator/Differentiator.h:62:12: error: no viable conversion from returned value of type 'const double *' to function return type 'clad::array_ref<double>'
   62 |     return val;
      |            ^~~
note: in instantiation of function template specialization 'clad::push<double, const double *>' requested here
/usr/include/clad/Differentiator/ArrayRef.h:16:29: note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'const double *' to 'const array_ref<double> &' for 1st argument
   16 | template <typename T> class array_ref {
      |                             ^~~~~~~~~
/usr/include/clad/Differentiator/ArrayRef.h:16:29: note: candidate constructor (the implicit move constructor) not viable: no known conversion from 'const double *' to 'array_ref<double> &&' for 1st argument
   16 | template <typename T> class array_ref {
      |                             ^~~~~~~~~
/usr/include/clad/Differentiator/ArrayRef.h:32:20: note: candidate constructor not viable: 1st argument ('const double *') would lose const qualifier
[...]

PR #897, opened 2024-05-17
Add custom derivative to DerivativeSet

vaithak:
This is required if the definition of the custom derivative is not found in the current translation unit and is linked in from another. 
Adding it to the set of derivatives ensures that the custom derivative is not differentiated again using numerical differentiation due to an unavailable definition.


vaithak:
To test this locally with Clang, we can declare a custom derivative (for a called function) but cannot offer its definition. Now, computing the hessian should result in a linker error instead of a compile-time error.
A similar scenario was observed in the cling and ROOT setup, where the definition of custom derivative was not found in some cases.

Minimal code:
```cpp

namespace clad{
  clad::ValueAndPushforward<double, double> f_pushforward(double x, double _d_x);
}
}

double f(double x) {
  return x*x;
}

double wrapper(double x) {
  return f(x);
}

int main() {
  clad::hessian(wrapper);
  return 0;
}
```

vgvassilev:
Can we add this test to the PR?

vaithak:
As mentioned above, the crude way to test this locally was seeing a linker error instead of a compile time error because of numerical differentiation. I don't think it would be good to have something like that in the CI check.

We can test this with ROOT infrastructure though.

parth-07:
@vaithak We can test this by adding multiple files for the test, right? For example, we can have two files, `CustomDerivativesDecl.C` and `CustomDerivativesDef.cpp`. `CustomDerivativesDecl.C` will have declarations of custom derivatives and `CustomDerivativesDef.cpp` will have the definitions of custom derivatives. `CustomDerivativesDecl.C` will contain the actual test as well and it will utilize `CustomDerivativesDef.cpp` in the compile commands of the test.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -253,6 +253,17 @@ static void registerDerivative(FunctionDecl* derivedFD, Sema& semaRef) {
...
+
+      // Add the custom derivative to the set of derivatives.
+      // This is required in case the definition of the custom derivative
+      // is not found in the current translation unit and is linked in
+      // from another translation unit.
+      // Adding it to the set of derivatives ensures that the custom
+      // derivative is not differentiated again using numerical
+      // differentiation due to unavailable definition.
+      if (auto* CE = dyn_cast<CallExpr>(OverloadedFn))
+        if (auto* FD = CE->getDirectCallee())
```
vgvassilev, line 265:
```suggestion
        if (FunctionDecl* FD = CE->getDirectCallee())
```

tools/ClangPlugin.cpp
```diff
@@ -407,12 +407,17 @@ namespace clad {
...
-        m_DiffRequestGraph.setCurrentProcessingNode(request);
-        ProcessDiffRequest(request);
-        m_DiffRequestGraph.markCurrentNodeProcessed();
-        request = m_DiffRequestGraph.getNextToProcessNode();
+      if (!m_DiffRequestGraph.isProcessingNode()) {
+        // This check is to avoid recursive processing of the graph, as
+        // HandleTopLevelDecl can be called recursively in non-standard
+        // setup for code generation.
+        DiffRequest request = m_DiffRequestGraph.getNextToProcessNode();
+        while (request.Function != nullptr) {
```
vgvassilev, line 415:
```suggestion
        while (request.Function) {
```

vgvassilev:
@vaithak, is it worth adding the test from #890 here?

vaithak:
> @vaithak, is it worth adding the test from #890 here?

Added tests 👍🏼

vgvassilev:
I am wondering why the coverage was decreased when we added a test… quite strange. Can you fix the failures and squash into a single commit?

vaithak:
> I am wondering why the coverage was decreased when we added a test… quite strange. Can you fix the failures and squash into a single commit?

I need to fix Kokkos test, not sure what is happening there.
For codecov, the lines we added were already covered (although not in the way we wanted). 
I guess adding more lines only increased total lines without changing the total covered lines, hence decreasing the percent by a bit.
This is just a guess though :)

vgvassilev:
> > I am wondering why the coverage was decreased when we added a test… quite strange. Can you fix the failures and squash into a single commit?
> 
> I need to fix Kokkos test, not sure what is happening there. For codecov, the lines we added were already covered (although not in the way we wanted). I guess adding more lines only increased total lines without changing the total covered lines, hence decreasing the percent by a bit. This is just a guess though :)

The kokkos failure looks like concurrency failure. We built this file before building clad.so?

Issue #899, opened 2024-05-20
Add support for differentiating the null (empty) statement `;`


gojakuch:
Currently, Clad throws a warning ("warning: attempted to differentiate unsupported statement, no changes applied") for every additional semicolon put in the source code (both in forward and reverse modes). Furthermore, all the semicolons are also pulled into the derivative code. The following code:
```C++
double f(double x) {;;;;;;;return x;;}
```
produces a bunch of warnings and results into the following derivative (forward mode):
```C++
double f_darg0(double x) {
    double _d_x = 1;
    ;
    ;
    ;
    ;
    ;
    ;
    ;
    return _d_x;
    ;
}
```
same happens in the reverse mode.
expected behaviour:
```C++
double f_darg0(double x) {
    double _d_x = 1;
    return _d_x;
}
```

PR #900, opened 2024-05-20
Add support for null (empty) statements

gojakuch:
This commit removes the "unsupported" warning when differentiating code with ";" in both forward and reverse modes. It also prevents additional semicolons from getting pulled into the derivative code.

Fixes: #899

### DIFF
--- a/include/clad/Differentiator/BaseForwardModeVisitor.h
+++ b/include/clad/Differentiator/BaseForwardModeVisitor.h
@@ -109,6 +109,7 @@ public:
       const clang::SubstNonTypeTemplateParmExpr* NTTP);
   StmtDiff VisitImplicitValueInitExpr(const clang::ImplicitValueInitExpr* IVIE);
   StmtDiff VisitCStyleCastExpr(const clang::CStyleCastExpr* CSCE);
+  StmtDiff VisitNullStmt(const clang::NullStmt* NS) { return StmtDiff{}; };
   static DeclDiff<clang::StaticAssertDecl>
   DifferentiateStaticAssertDecl(const clang::StaticAssertDecl* SAD);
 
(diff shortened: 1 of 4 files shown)


vgvassilev:
Can you break the long commit message lines at 80-columns? This has been somewhat standard.

include/clad/Differentiator/BaseForwardModeVisitor.h
```diff
@@ -109,6 +109,7 @@ class BaseForwardModeVisitor
       const clang::SubstNonTypeTemplateParmExpr* NTTP);
   StmtDiff VisitImplicitValueInitExpr(const clang::ImplicitValueInitExpr* IVIE);
   StmtDiff VisitCStyleCastExpr(const clang::CStyleCastExpr* CSCE);
+  StmtDiff VisitNullStmt(const clang::NullStmt* NS) { return StmtDiff{}; };
```
vgvassilev, line 112: I suspect there is a lot of such cases for the different types of statements. Maybe you can take a look at the statements we do not have a Visit routine for. Some of them could be as trivial and easy to test. That'd increase the coverage of supported things generally speaking.

vgvassilev (approved):
LGTM, assuming the commit message is addressed.

test/Gradient/Gradients.C
```diff
@@ -901,6 +901,18 @@ double fn_cond_init(double x) {
 //CHECK-NEXT:              }
 //CHECK-NEXT:          }
 
+double fn_empty_lines(double x) {
```
vgvassilev, line 904:
```suggestion
double fn_null_stmts(double x) {
```

include/clad/Differentiator/BaseForwardModeVisitor.h
gojakuch: alright, I'll look for things like this

vgvassilev:
Codecov complains because now the NullStmts have their own routine and we do not fall back into `VisitStmt` (which generally should never happen).

PR #904, opened 2024-05-23
Make GlobalStoreAndRef consistent in usage inside and outside of loops

### LINKED ISSUE #431: Improve consistency of `GlobalStoreAndRef`
`GlobalStoreAndRef` creates a variable that can be used for storing and accessing information in the global scope of the function. 
It internally uses `clad::tape`, if and only if, it needs to store the value of an expression contained with a loop. 
A close cousin of `GlobalStoreAndRef` is `StoreAndRef` -- it creates a variable that can be used for storing and accessing information in the local scope. 

`GlobalStoreAndRef` returns a `StmtDiff` object containing expressions that should be used in the forward pass and the reverse pass respectively. 

If `GlobalStoreAndRef` is used to store an expression that is not contained within a loop, then it returns `{ref, ref}`, where `ref` is a `DeclRefExpr` node for the global variable created by `GlobalStoreAndRef`. Please note the following things:

- `ref` can be used any number of times in the generated function. 
- The interface of `StoreAndRef` is consistent with `StoreAndRef` -- increased consistency makes is easier to reason about the behaviour of a program.

If `GlobalStoreAndRef` is used to store an expression that is contained within a loop, then in returns `{clad::push(...), clad::pop(...)}`. Please note the following things: 

- `clad::push` and `clad::pop` should ideally be used once only, or more precisely, they should be used exactly same number of times in the forward and the reverse pass. This doesn't always match with the differentiation requirements.
- The interface is different for the special case in which the stored expression resides in a loop. This creates difficulties in the cases where we need to be agnostic about if the expression is contained in a loop or not. For these cases, currently, we need to have separate branches such as these: 

```cpp
if (isInsideLoop) {
  ...
  ...
} else {
  ...
  ...
}
```
It also makes `GlobalStoreAndRef` behaviour inconsistent, and forces the developer using this functionality to be extra cautious. 

I propose to make the functionality `GlobalStoreAndRef` more consistent and easier to use. If `GlobalStoreAndRef` is used to store an expression that is contained in a loop. Then we can modify `GlobalStoreAndRef` to behave as follows: 

```cpp
// global variable declarations
clad::tape<double> t0;

// forward pass
for (...; ...; ...) {
  double t1 = t0.push(e);  // added implicitly by 'GlobalStoreAndRef' in the current forward block
  ....
}

// reverse pass
for (...; ...; ...) {
  double t2 = t0.pop();  // added implicitly by 'GlobalStoreAndRef' in the current reverse block
}
```

Now, `GlobalStoreAndRef` will return `{t1, t2}`. 

This can help to simplify various implementation that uses `GlobalStoreAndRef`, most notably -- `ReverseModeVisitor::VisitCallExpr`. 

This functionality have a downside that it creates 2 additional variables, thus if we do decide to add this functionality, we should add some flag in `GlobalStoreAndRef` function using which developer that needs to use `GlobalStoreAndRef` functionality can optimally decide if they require this additional consistency or not.

PetroZarytskyi:
The purpose of this PR is to make ``GlobalStoreAndRef`` work more consistently inside and outside of loops. The issue is well described in #431. The solution proposed in this PR is as follows:
When ``GlobalStoreAndRef`` is called inside a loop, it adds a push statement (``clad::push(_t0, x)``) to the forward block and a pop statement (``clad::pop(_t0)``) to the reverse block. The return type of ``GlobalStoreAndRef`` is changed from ``StmtDiff`` to ``clang::Expr*``. The return value is the top element on the tape (``clad::back(_t0)``). When ``GlobalStoreAndRef`` is called outside a loop, the behavior is essentially the same as before except now it returns a single reference to the created variable ``ref`` instead of ``StmtDiff(ref, ref)``.
For example (when used to store the condition of an if-stmt inside a loop):
```
 // forward sweep:
clad::push(_cond1, another_choice > 1);
if (clad::back(_cond1)) {
    ...
}
```
```
 // reverse sweep:
if (clad::back(_cond1)) {
    ...
}
clad::pop(_cond1);
```
With this approach, we automatically ensure safe tape pushes/pops when working with return/break/continue statements, which wasn't the case previously. Because of this a lot of the code used to handle those statements got simplified.
Also, the change of the return type makes ``GlobalStoreAndRef`` consistent in usage with ``StoreAndRef``.

Note: I made ``GlobalStoreAndRef`` return ``clad::back(_t0)`` instead of stored values of ``clad::push(_t0)`` and ``clad::pop(_t0)`` (as proposed by @parth-07 in #431) because in all use cases ``_r0 = clad::pop(_t0)`` was added to the reverse block after ``_r0`` was used (because reverse blocks are reversed). Getting the statement order right would make ``GlobalStoreAndRef`` a lot harder to use.

Fixes #431.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2991,24 +2991,6 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         E, getNonConstType(E->getType(), m_Context, m_Sema), prefix, force);
   }
 
-  StmtDiff ReverseModeVisitor::BuildPushPop(clang::Expr* E,
-                                            clang::QualType Type,
-                                            llvm::StringRef prefix,
-                                            bool force) {
-    if (isInsideLoop) {
-      auto CladTape = MakeCladTapeFor(Clone(E));
-      Expr* Push = CladTape.Push;
-      Expr* Pop = CladTape.Pop;
-      return {Push, Pop};
-    }
-    Expr* init = nullptr;
-    if (const auto* const AT = dyn_cast<ArrayType>(Type))
-      init = getArraySizeExpr(AT, m_Context, *this);
-
-    Expr* Ref = BuildDeclRef(GlobalStoreImpl(Type, prefix, init));
-    return {Ref, Ref};
-  }
-
   StmtDiff ReverseModeVisitor::StoreAndRestore(clang::Expr* E,
                                                llvm::StringRef prefix,
                                                bool force) {
@@ -3018,10 +3000,11 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       return {};
 
     if (isInsideLoop) {
-      auto pushPop = BuildPushPop(E, Type, prefix, true);
-      auto* popAssign = BuildOp(BinaryOperatorKind::BO_Assign, Clone(E),
-                                pushPop.getExpr_dx());
-      return {pushPop.getExpr(), popAssign};
+      auto CladTape = MakeCladTapeFor(Clone(E), prefix);
+      Expr* Push = CladTape.Push;
+      Expr* Pop = CladTape.Pop;
+      auto* popAssign = BuildOp(BinaryOperatorKind::BO_Assign, Clone(E), Pop);
+      return {Push, popAssign};
     }
 
     Expr* init = nullptr;
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -801,12 +801,6 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     // to this scope.
     beginScope(Scope::DeclScope | Scope::ControlScope);
 
-    // Condition has to be stored as a "global" variable, to take the correct
-    // branch in the reverse pass.
-    // If we are inside loop, the condition has to be stored in a stack after
-    // the if statement.
-    Expr* PushCond = nullptr;
-    Expr* PopCond = nullptr;
     // Create a block "around" if statement, e.g:
     // {
     //   ...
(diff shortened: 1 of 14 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -897,18 +871,12 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     StmtDiff elseDiff = VisitBranch(If->getElse());
 
     Stmt* Forward = clad_compat::IfStmt_Create(
-        m_Context, noLoc, If->isConstexpr(), nullptr, nullptr,
-        condDiff.getExpr(), noLoc, noLoc, thenDiff.getStmt(), noLoc,
-        elseDiff.getStmt());
+        m_Context, noLoc, If->isConstexpr(), nullptr, nullptr, condDiffStored,
```
vgvassilev, line 874:
```suggestion
        m_Context, noLoc, If->isConstexpr(), /*Init=*/nullptr, /*Var=*/nullptr, condDiffStored,
```
@vaithak, not sure why clang-tidy refuses to complain but that's quite annoying. Can you take another look?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1111,18 +1072,9 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
-                                            CounterDecrement,
-                                            ReverseResult,
-                                            noLoc,
-                                            noLoc,
-                                            noLoc);
-    addToCurrentBlock(Forward, direction::forward);
-    Forward = endBlock(direction::forward);
-    addToCurrentBlock(loopCounter.getPop(), direction::reverse);
+    Stmt* Reverse = new (m_Context)
+        ForStmt(m_Context, nullptr, CounterCondition, nullptr, CounterDecrement,
```
vgvassilev, line 1076: Likewise.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1572,59 +1524,17 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
-                             argDiffStore.getExpr());
-          block.insert(block.begin() + insertionPoint, op);
-          // We added restoration of the original arg. Thus we need to
-          // correspondingly adjust the insertion point.
-          insertionPoint += 1;
-
-          argDiffStore = {argDiff.getExpr(), argDiffStore.getExpr_dx()};
-        }
+        // Restore args
+        auto& block = getCurrentBlock(direction::reverse);
```
vgvassilev, line 1528: We should spell out the type here.

parth-07 (commented):
This looks amazing. 
I will review it in-depth soon.

vgvassilev (commented):
This looks good to me. Let's wait for @parth-07's review here.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: Fixed it.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: And this.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: Fixed it.

parth-07 (approved):
This pull-request has simplified a lot of complicated codes. Looks good to me.

Issue #273, opened 2021-07-31
Incorrect handling of condition variable of for statement in forward mode


parth-07:
Currently, 

```c++
for (; int c = b;) {
  c += x;
  b -= 1;
  ...
  ...
}
```
This code gets differentiated to, 

```c++
{
  int _d_c = _d_b;
  for (; int c = b;) {
    _d_c += _d_x;
    c += x;
    _d_b -= 0;
    b -= 1;
    ...
    ...
  }
}
```

The condition variable is created and initialized at each iteration, therefore, the derived variable of the condition variable should also be created and initialized (or at least assigned to 0) at each iteration.
The current implementation of creating a derived variable of the condition variable in an outer scope gives incorrect results.

The above code should get ideally differentiated to something like this, 

```c++
{

  for (; int c = b;) {
    int _d_c = _d_b;  // initialized at each iteration
    _d_c += _d_x;
    c += x;
    _d_b -= 0;
    b -= 1;
    ...
    ...
  }
}
```

PR #905, opened 2024-05-23
Fix condition declarations and assignments in for loops

gojakuch:
These changes fix the differentiation of variable declarations in for loop conditions that used to result into wrong derivatives. The commit also tackles the problem of having an assignment operator that affects the derivative in the for-loop condition, as well as adds support for logical operators in for-loops and allows to combine assignments with them.

Fixes: #273

### DIFF
--- a/include/clad/Differentiator/CladUtils.h
+++ b/include/clad/Differentiator/CladUtils.h
@@ -27,6 +27,10 @@ namespace clad {
     /// function `FD`.
     std::string ComputeEffectiveFnName(const clang::FunctionDecl* FD);
 
+    // Unwraps S to a single statement if it's a compound statement only
+    // containing 1 statement.
+    clang::Stmt* unwrapIfSingleStmt(clang::Stmt* S);
+
     /// Creates and returns a compound statement having statements as follows:
     /// {`S`, all the statement of `initial` in sequence}
     clang::CompoundStmt* PrependAndCreateCompoundStmt(clang::ASTContext& C,
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -668,18 +668,60 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
   const Stmt* init = FS->getInit();
   StmtDiff initDiff = init ? Visit(init) : StmtDiff{};
   addToCurrentBlock(initDiff.getStmt_dx());
-  VarDecl* condVarDecl = FS->getConditionVariable();
-  VarDecl* condVarClone = nullptr;
-  if (condVarDecl) {
-    DeclDiff<VarDecl> condVarResult = DifferentiateVarDecl(condVarDecl);
-    condVarClone = condVarResult.getDecl();
+
+  StmtDiff condDiff = Clone(FS->getCond());
+  Expr* cond = condDiff.getExpr();
+
+  // The declaration in the condition needs to be differentiated.
+  if (VarDecl* condVarDecl = FS->getConditionVariable()) {
+    // Here we create a fictional cond that is equal to the assignment used in
+    // the declaration. The declaration itself is thrown before the for-loop
+    // without any init value. The fictional condition is then differentiated as
+    // a normal condition would be (see below). For example, the declaration
+    // inside `for (;double t = x;) {}` will be first processed into the
+    // following code:
+    // ```
+    // {
+    // double t;
+    // for (;t = x;) {}
+    // }
+    // ```
+    // which will then get differentiated normally as a for-loop with a
+    // differentiable condition in the next section.
+    DeclDiff<VarDecl> condVarResult =
+        DifferentiateVarDecl(condVarDecl, /*ignoreInit=*/true);
+    VarDecl* condVarClone = condVarResult.getDecl();
     if (condVarResult.getDecl_dx())
       addToCurrentBlock(BuildDeclStmt(condVarResult.getDecl_dx()));
+    auto condInit = condVarClone->getInit();
+    condVarClone->setInit(nullptr);
+    cond = BuildOp(BO_Assign, BuildDeclRef(condVarClone), condInit);
+    addToCurrentBlock(BuildDeclStmt(condVarClone));
   }
-  Expr* cond = FS->getCond() ? Clone(FS->getCond()) : nullptr;
-  const Expr* inc = FS->getInc();
+
+  // Condition differentiation.
+  // This adds support for assignments in conditions.
+  if (cond) {
+    cond = cond->IgnoreParenImpCasts();
+    // If it's a supported differentiable operator we wrap it back into
+    // parentheses and then visit. To ensure the correctness, a comma operator
+    // expression (cond_dx, cond) is generated and put instead of the condition.
+    // FIXME: Add support for other expressions in cond (unary operators,
+    // comparisons, function calls, etc.). Ideally, we should be able to simply
+    // always call Visit(cond)
+    BinaryOperator* condBO = dyn_cast<BinaryOperator>(cond);
+    if (condBO && (condBO->isLogicalOp() || condBO->isAssignmentOp())) {
+      condDiff = Visit(cond);
+      if (condDiff.getExpr_dx() && !isUnusedResult(condDiff.getExpr_dx()))
+        cond = BuildOp(BO_Comma, BuildParens(condDiff.getExpr_dx()),
+                       BuildParens(condDiff.getExpr()));
+      else
+        cond = condDiff.getExpr();
+    }
+  }
 
   // Differentiate the increment expression of the for loop
+  const Expr* inc = FS->getInc();
   beginBlock();
   StmtDiff incDiff = inc ? Visit(inc) : StmtDiff{};
   CompoundStmt* decls = endBlock();
@@ -714,27 +756,24 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
     incResult = incDiff.getExpr();
   }
 
+  // Build the derived for loop body.
   const Stmt* body = FS->getBody();
   beginScope(Scope::DeclScope);
   Stmt* bodyResult = nullptr;
-  if (isa<CompoundStmt>(body)) {
-    bodyResult = Visit(body).getStmt();
-  } else {
-    beginBlock();
-    StmtDiff Result = Visit(body);
-    for (Stmt* S : Result.getBothStmts())
-      addToCurrentBlock(S);
-    CompoundStmt* Block = endBlock();
-    if (Block->size() == 1)
-      bodyResult = Block->body_front();
-    else
-      bodyResult = Block;
-  }
+  beginBlock();
+  StmtDiff bodyVisited = Visit(body);
+  for (Stmt* S : bodyVisited.getBothStmts())
+    addToCurrentBlock(S);
+  CompoundStmt* bodyResultCmpd = endBlock();
+  if (bodyResultCmpd->size() == 1)
+    bodyResult = bodyResultCmpd->body_front();
+  else
+    bodyResult = bodyResultCmpd;
   endScope();
 
-  Stmt* forStmtDiff =
-      new (m_Context) ForStmt(m_Context, initDiff.getStmt(), cond, condVarClone,
-                              incResult, bodyResult, noLoc, noLoc, noLoc);
+  Stmt* forStmtDiff = new (m_Context)
+      ForStmt(m_Context, initDiff.getStmt(), cond, /*condVar=*/nullptr,
+              incResult, bodyResult, noLoc, noLoc, noLoc);
 
   addToCurrentBlock(forStmtDiff);
   CompoundStmt* Block = endBlock();
@@ -1366,6 +1405,25 @@ BaseForwardModeVisitor::VisitBinaryOperator(const BinaryOperator* BinOp) {
     } else
       opDiff = BuildOp(BO_Comma, BuildParens(Ldiff.getExpr()),
                        BuildParens(Rdiff.getExpr_dx()));
+  } else if (BinOp->isLogicalOp()) {
+    // For (A && B) return ((dA, A) && (dB, B)) to ensure correct evaluation and
+    // correct derivative execution.
+    auto buildOneSide = [this](StmtDiff& Xdiff) {
+      if (Xdiff.getExpr_dx() && !isUnusedResult(Xdiff.getExpr_dx()))
+        return BuildParens(BuildOp(BO_Comma, BuildParens(Xdiff.getExpr_dx()),
+                                   BuildParens(Xdiff.getExpr())));
+      return BuildParens(Xdiff.getExpr());
+    };
+    // dLL = (dL, L)
+    Expr* dLL = buildOneSide(Ldiff);
+    // dRR = (dR, R)
+    Expr* dRR = buildOneSide(Rdiff);
+    opDiff = BuildOp(opCode, dLL, dRR);
+
+    // Since the both parts are included in the opDiff, there's no point in
+    // including it as a Stmt_dx. Moreover, the fact that Stmt_dx is left
+    // nullptr is used for treating expressions like ((A && B) && C) correctly.
+    return StmtDiff(opDiff, nullptr);
   }
   if (!opDiff) {
     // FIXME: add support for other binary operators
@@ -1386,7 +1444,21 @@ BaseForwardModeVisitor::VisitBinaryOperator(const BinaryOperator* BinOp) {
 
 DeclDiff<VarDecl>
 BaseForwardModeVisitor::DifferentiateVarDecl(const VarDecl* VD) {
-  StmtDiff initDiff = VD->getInit() ? Visit(VD->getInit()) : StmtDiff{};
+  return DifferentiateVarDecl(VD, false);
+}
+
+DeclDiff<VarDecl>
+BaseForwardModeVisitor::DifferentiateVarDecl(const VarDecl* VD,
+                                             bool ignoreInit) {
+  StmtDiff initDiff{};
+  const Expr* init = VD->getInit();
+  if (init) {
+    if (!ignoreInit)
+      initDiff = Visit(init);
+    else
+      initDiff = StmtDiff(Clone(init));
+  }
+
   // Here we are assuming that derived type and the original type are same.
   // This may not necessarily be true in the future.
   VarDecl* VDClone =
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -764,11 +764,7 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
   StmtDiff bodyVisited = Visit(body);
   for (Stmt* S : bodyVisited.getBothStmts())
     addToCurrentBlock(S);
-  CompoundStmt* bodyResultCmpd = endBlock();
-  if (bodyResultCmpd->size() == 1)
-    bodyResult = bodyResultCmpd->body_front();
-  else
-    bodyResult = bodyResultCmpd;
+  bodyResult = utils::unwrapIfSingleStmt(endBlock());
   endScope();
 
   Stmt* forStmtDiff = new (m_Context)
--- a/include/clad/Differentiator/BaseForwardModeVisitor.h
+++ b/include/clad/Differentiator/BaseForwardModeVisitor.h
@@ -75,6 +75,8 @@ public:
   // Decl is not Stmt, so it cannot be visited directly.
   virtual DeclDiff<clang::VarDecl>
   DifferentiateVarDecl(const clang::VarDecl* VD);
+  virtual DeclDiff<clang::VarDecl>
+  DifferentiateVarDecl(const clang::VarDecl* VD, bool ignoreInit);
   /// Shorthand for warning on differentiation of unsupported operators
   void unsupportedOpWarn(clang::SourceLocation loc,
                          llvm::ArrayRef<llvm::StringRef> args = {}) {
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -102,6 +102,19 @@ namespace clad {
       }
     }
 
+    Stmt* unwrapIfSingleStmt(Stmt* S) {
+      if (!S)
+        return nullptr;
+      if (!isa<CompoundStmt>(S))
+        return S;
+      auto* CS = cast<CompoundStmt>(S);
+      if (CS->size() == 0)
+        return nullptr;
+      if (CS->size() == 1)
+        return CS->body_front();
+      return CS;
+    }
+
     CompoundStmt* PrependAndCreateCompoundStmt(ASTContext& C, Stmt* initial,
                                                Stmt* S) {
       llvm::SmallVector<Stmt*, 16> block;
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -783,19 +783,6 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     return StmtDiff(Forward, Reverse);
   }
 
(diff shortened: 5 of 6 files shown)


lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -668,18 +668,39 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
...
     if (condVarResult.getDecl_dx())
-      addToCurrentBlock(BuildDeclStmt(condVarResult.getDecl_dx()));
+      condVarDeclStmt_dx = BuildDeclStmt(condVarResult.getDecl_dx());
+  }
+
+  // condition
+  StmtDiff condDiff = Clone(FS->getCond());
+  if (Expr* cond =
+          condDiff
+              .getExpr()) { // this adds support for assignments in conditions
```
vgvassilev, line 688: Can you move the comments on a separate line before the statement that we are commenting on? That will help clang-format avoid the weird indentations...

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -668,18 +668,39 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
...
+    while (CastExpr* condCast = dyn_cast<CastExpr>(cond))
+      cond = condCast->getSubExpr();
+    while (ParenExpr* condParen = dyn_cast<ParenExpr>(cond))
+      cond = condParen->getSubExpr();
+    if (BinaryOperator* condBO = dyn_cast<BinaryOperator>(cond)) {
+      if (condBO->isAssignmentOp())
+        condDiff = Visit(new (m_Context) ParenExpr(
+            noLoc, noLoc,
+            cond)); // if it's an assignment operator we wrap it back into
```
vgvassilev, line 697: Likewise.

parth-07:
The fix does not work for all cases. In the case when the for-loop condition fails, we should run the derivative statement for the for-loop condition, but we do not. This leads to incorrect derivative in some cases. 

Reproducible example:

```cpp
#include <iostream>

    double res = 0;
    for (; (res = u * v) && false ;) {}
    return res;
}

int main() {
    auto fn_diff = clad::differentiate(fn, "u");
    show(fn_diff.execute(3, 5));
}
```
Output:
fn_diff.execute(3, 5): 0

Expected Output:
fn_diff.execute(3, 5): 5

parth-07 (changes requested):
Please see my earlier comment for the case which produces wrong output with this solution.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -668,18 +668,37 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
   const Stmt* init = FS->getInit();
   StmtDiff initDiff = init ? Visit(init) : StmtDiff{};
   addToCurrentBlock(initDiff.getStmt_dx());
+
+  // declaration in the condition (if any) needs to be differentiated
```
vgvassilev, line 672:
```suggestion
  // Declaration in the condition (if any) needs to be differentiated.
```

Our comments are full sentences starting with a capital letter and ending with punctuation. Can you fix the rest of the comments, too?

gojakuch: okay, I will fix this together with what Parth asked for.

parth-07:
It seems that we do not support logical operators in a for-loop conditions. Hence, support for conditions such as this `(res = u * v) && false` can be added in a separate pull-request. However, the below example should work with this pull request:

```cpp
#include <iostream>

    double res = u * v;
    for (; (res = 0) ;) {}
    return res;
}

int main() {
    auto fn_diff = clad::differentiate(fn, "u");
    show(fn_diff.execute(3, 5));
}
```

**Output:** 
fn_diff.execute(3, 5): 5

**Expected output:** 
fn_diff.execute(3, 5): 0

gojakuch:
> It seems that we do not support logical operators in a for-loop conditions.

I have already started working on this because of your initial suggestion. Adding this as a separate PR would require rewriting the changes made in this PR once again. I can just change the commit and PR messages to also include this once I'm done.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -668,18 +668,50 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
...
-    condVarClone = condVarResult.getDecl();
-    if (condVarResult.getDecl_dx())
-      addToCurrentBlock(BuildDeclStmt(condVarResult.getDecl_dx()));
+
+  StmtDiff condDiff = Clone(FS->getCond());
+  Expr* cond = condDiff.getExpr();
+
+  // Declaration in the condition (if any) needs to be differentiated.
+  if (VarDecl* condVarDecl = FS->getConditionVariable()) {
+    // FIXME: here DifferentiateVarDecl is only used to clone the declaration
```
parth-07, line 677: What other ways did you try before DifferentiateVarDecl?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -668,18 +668,50 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
...
+    cond =
+        BuildOp(BO_Assign, BuildDeclRef(condVarClone), condVarClone->getInit());
+    condVarClone->setInit(nullptr);
+    addToCurrentBlock(BuildDeclStmt(condVarClone));
+  }
+
+  // Condition differentiation.
+  // This adds support for assignments in conditions.
+  if (cond) {
+    while (CastExpr* condCast = dyn_cast<CastExpr>(cond))
```
parth-07, line 693: Can these two `while` loops be replaced by: `Expr::IgnoreParenImpCasts` function?

parth-07 (changes requested):
It seems that currently for-loop condition is only differentiated if it's a binary operator. This additional check seems unnecessary and is incorrect. You should try to directly call `Visit(...)` function on the condition expression.

The below code gives an incorrect derivative due to the condition expression only being differentiated if it's a binary operator:

```cpp
#include <iostream>

    double res = 0;
    for (; !(res = u * v) ;) {}
    return res;
}

int main() {
    auto fn_diff = clad::differentiate(fn, "u");
    show(fn(3, 5));
    show(fn_diff.execute(3, 5));
}
```

**Output:**
fn_diff.execute(3, 5): 0

**Expected Output:**
fn_diff.execute(3, 5): 5

lib/Differentiator/BaseForwardModeVisitor.cpp
gojakuch: yes, thank you, I didn't know this exists

lib/Differentiator/BaseForwardModeVisitor.cpp
gojakuch: I have found another way, I'll push it soon.

gojakuch:
> It seems that currently for-loop condition is only differentiated if it's a binary operator. This additional check seems unnecessary and is incorrect. You should try to directly call Visit(...) function on the condition expression.

As for this, I enabled support for unary operators, added support for visiting the logical not in general. I also allowed comparison operators, which are probably the most common type of expression in for-loop conditions, and made the binary operator visitor work with them. I understand that it is a bit ugly to artificially limit the differentiation of cond, however, Clad doesn't support many other operators and if we simply visit anything that is inside cond, some tests start to crash because of complicated reasons related to trying to visit an unsupported statement (that is my guess, I spent some time trying to understand what's going on but I didn't understand much). I don't think we should add general support for other operators in this PR, so I added a fixme message to add support for other expressions in cond. this PR already covers a lot of things that were not possible or were differentiated incorrectly before, so I think visiting only these supported statements could be a common ground and a solution so far.

speaking of crashing checks, as you can see, there are some checks failing for OSX now, specifically the ForwardMode/UserDefinedTypes.C test. which is exactly what started crashing on my machine after I had tried to visit every for-loop condition regardless of its type. these errors here are different though and I don't get them on my machine with the current PR code. I'm not even sure how they are related to for loops. I would appreciate if someone more experienced took a look at these at least, maybe I could make visiting all the conds work with similar fixes then.

parth-07:
@gojakuch You are right. We are enabling a lot of different things in this pull request. It is my bad. I thought some features (binary operator, unary operator) were supported before, but they were not. Can we please limit the scope of this pull request? Let's handle the unary operator and general support for visiting conditions in a separate pull request. 

Can you please revert the latest changes and bring the pull request to the point when you enabled binary operator support and add FIXME in the code regarding general support is missing?

vgvassilev:
In addition can you open issues for all things that this pull request discovered are not supported?

gojakuch:
> Can you please revert the latest changes and bring the pull request to the point when you enabled binary operator support and add FIXME in the code regarding general support is missing?

done in the last force-push

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -751,6 +783,15 @@ StmtDiff BaseForwardModeVisitor::VisitReturnStmt(const ReturnStmt* RS) {
     return nullptr;
 
   StmtDiff retValDiff = Visit(RS->getRetValue());
+  // llvm::errs() << "\n\n\n\n\nretdx: \n";
```
parth-07, line 786: Please remove the debugging statements.

parth-07 (changes requested):

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -668,18 +668,53 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
...
+
+  // Declaration in the condition (if any) needs to be differentiated.
+  if (VarDecl* condVarDecl = FS->getConditionVariable()) {
+    VarDecl* condVarClone =
+        BuildVarDecl(condVarDecl->getType(), condVarDecl->getNameAsString(),
+                     Clone(condVarDecl->getInit()), condVarDecl->isDirectInit(),
+                     nullptr, condVarDecl->getInitStyle());
+    VarDecl* _d_condVarDecl = BuildVarDecl(
+        condVarClone->getType(), "_d_" + condVarClone->getNameAsString(),
+        nullptr, condVarClone->isDirectInit(), nullptr,
```
vgvassilev, line 683:
```suggestion
        /*Init=*/nullptr, condVarClone->isDirectInit(), nullptr,
```
We comment in the the arguments which are unclear what parameters refer to.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -668,18 +668,53 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
...
+        condVarClone->getInitStyle());
+    m_Variables.emplace(condVarClone, BuildDeclRef(_d_condVarDecl));
+    // Here we create a fictional cond that is equal to the assignment used in
+    // the declaration. The declaration itself is thrown before the for-loop
+    // without any init value. The fictional condition is then differentiated as
+    // normal condition (see below).
+    cond =
+        BuildOp(BO_Assign, BuildDeclRef(condVarClone), condVarClone->getInit());
```
vgvassilev, line 691: I'd move the DeclRefExpr from `BuildDeclRef(condVarClone)` into a separate variable. I thin we need to clone the initializer expression of `condVarClone` if we want to use it separately. However, in this situation it is fine because on the next line we remove it. The code needs to become more clear about that. Maybe something like:

CondInit = condVarClone->getInit();
condVarClone->setInit(nullptr);
...

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -668,18 +668,53 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
...
+    condVarClone->setInit(nullptr);
+    addToCurrentBlock(BuildDeclStmt(condVarClone));
+  }
+
+  // Condition differentiation.
+  // This adds support for assignments in conditions.
+  if (cond) {
+    cond = cond->IgnoreParenImpCasts();
```
vgvassilev, line 699: Shouldn't we strip the implict casts of the init expression when forming it in the previous section?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -668,18 +668,53 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
...
+    // FIXME: Add support for other expressions in cond (unary operators,
+    // comparisons, function calls, etc.). Ideally, we should be able to simply
+    // always call Visit(cond)
+    BinaryOperator* condBO = dyn_cast<BinaryOperator>(cond);
+    if (condBO && (condBO->isLogicalOp() || condBO->isAssignmentOp())) {
+      condDiff = Visit(cond);
+      cond = (condDiff.getExpr_dx() && !isUnusedResult(condDiff.getExpr_dx()))
+                 ? BuildOp(BO_Comma, BuildParens(condDiff.getExpr_dx()),
+                           BuildParens(condDiff.getExpr()))
+                 : condDiff.getExpr();
```
vgvassilev, line 712: That's unreadable. We should probably expand the ternary if.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -714,26 +749,23 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
...
-  }
+  beginBlock();
+  StmtDiff bodyVisited = Visit(body);
+  for (Stmt* S : bodyVisited.getBothStmts())
+    addToCurrentBlock(S);
+  CompoundStmt* bodyResultCmpd = endBlock();
+  if (bodyResultCmpd->size() == 1)
+    bodyResult = bodyResultCmpd->body_front();
+  else
+    bodyResult = bodyResultCmpd;
```
vgvassilev, line 764: Can't we use `unwrapIfSingleStmt`?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -714,26 +749,23 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
...
   endScope();
 
   Stmt* forStmtDiff =
-      new (m_Context) ForStmt(m_Context, initDiff.getStmt(), cond, condVarClone,
+      new (m_Context) ForStmt(m_Context, initDiff.getStmt(), cond, nullptr,
```
vgvassilev, line 768:
```suggestion
      new (m_Context) ForStmt(m_Context, initDiff.getStmt(), cond, /*condVar=*/nullptr,
```

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1366,6 +1407,26 @@ BaseForwardModeVisitor::VisitBinaryOperator(const BinaryOperator* BinOp) {
...
       opDiff = BuildOp(BO_Comma, BuildParens(Ldiff.getExpr()),
                        BuildParens(Rdiff.getExpr_dx()));
+  } else if (BinOp->isLogicalOp()) {
+    // For (A && B) return ((dA, A) && (dB, B)) to ensure correct evaluation and
+    // correct derivative execution. dLL = (dL, L)
+    Expr* dLL =
+        BuildParens(((Ldiff.getExpr_dx() && !isUnusedResult(Ldiff.getExpr_dx()))
+                         ? BuildOp(BO_Comma, BuildParens(Ldiff.getExpr_dx()),
+                                   BuildParens(Ldiff.getExpr()))
+                         : Ldiff.getExpr()));
```
vgvassilev, line 1417: That is unreadable.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1366,6 +1407,26 @@ BaseForwardModeVisitor::VisitBinaryOperator(const BinaryOperator* BinOp) {
...
+    // dRR = (dR, R)
+    Expr* dRR =
+        BuildParens(((Rdiff.getExpr_dx() && !isUnusedResult(Rdiff.getExpr_dx()))
+                         ? BuildOp(BO_Comma, BuildParens(Rdiff.getExpr_dx()),
+                                   BuildParens(Rdiff.getExpr()))
+                         : Rdiff.getExpr()));
```
vgvassilev, line 1423: Likewise. You probably can create some lambda function to avoid duplicating this code in a few places...

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -668,18 +668,53 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
...
+                     nullptr, condVarDecl->getInitStyle());
+    VarDecl* _d_condVarDecl = BuildVarDecl(
+        condVarClone->getType(), "_d_" + condVarClone->getNameAsString(),
+        nullptr, condVarClone->isDirectInit(), nullptr,
+        condVarClone->getInitStyle());
+    m_Variables.emplace(condVarClone, BuildDeclRef(_d_condVarDecl));
+    // Here we create a fictional cond that is equal to the assignment used in
+    // the declaration. The declaration itself is thrown before the for-loop
+    // without any init value. The fictional condition is then differentiated as
+    // normal condition (see below).
```
vgvassilev, line 689: Probably better to give an example with code, too.

lib/Differentiator/BaseForwardModeVisitor.cpp
gojakuch: not always, because if there is no declaration in this for this section will just serve the purpose of differentiating the actual cond from the source code, not the one generated above. in this case it won't be unwrapped and could contain impcasts.

lib/Differentiator/BaseForwardModeVisitor.cpp
gojakuch: I agree

lib/Differentiator/BaseForwardModeVisitor.cpp
gojakuch: I think I tried but I don't remember why I didn't put it in the PR. maybe I just accidentally reverted it together with something that didn't work out well. in any case, as far as I remember this part of code was already there, I just removed an unnecessary if condition it was wrapped into. I'll try `unwrapIfSingleStmt` again rn.

vgvassilev: Let's move it in a common place and use it.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1366,6 +1405,26 @@ BaseForwardModeVisitor::VisitBinaryOperator(const BinaryOperator* BinOp) {
...
       opDiff = BuildOp(BO_Comma, BuildParens(Ldiff.getExpr()),
                        BuildParens(Rdiff.getExpr_dx()));
+  } else if (BinOp->isLogicalOp()) {
+    // For (A && B) return ((dA, A) && (dB, B)) to ensure correct evaluation and
+    // correct derivative execution.
+    auto buildOneSide = [this](StmtDiff& Xdiff) {
+      if (Xdiff.getExpr_dx() && !isUnusedResult(Xdiff.getExpr_dx()))
+        return BuildParens(BuildOp(BO_Comma, BuildParens(Xdiff.getExpr_dx()),
+                                   BuildParens(Xdiff.getExpr())));
+      else
```
vgvassilev, line 1415: No else after a return statement.

include/clad/Differentiator/CladUtils.h
```diff
@@ -27,6 +27,10 @@ namespace clad {
     /// function `FD`.
     std::string ComputeEffectiveFnName(const clang::FunctionDecl* FD);
 
+    // Unwraps S to a single statement if it's a compound statement only
+    // containing 1 statement.
+    clang::Stmt* unwrapIfSingleStmt(clang::Stmt* S);
```
vgvassilev, line 32: That operation should be in a separate commit.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1366,6 +1405,25 @@ BaseForwardModeVisitor::VisitBinaryOperator(const BinaryOperator* BinOp) {
...
       opDiff = BuildOp(BO_Comma, BuildParens(Ldiff.getExpr()),
                        BuildParens(Rdiff.getExpr_dx()));
+  } else if (BinOp->isLogicalOp()) {
+    // For (A && B) return ((dA, A) && (dB, B)) to ensure correct evaluation and
+    // correct derivative execution.
+    auto buildOneSide = [this](StmtDiff& Xdiff) {
+      if (Xdiff.getExpr_dx() && !isUnusedResult(Xdiff.getExpr_dx()))
+        return BuildParens(BuildOp(BO_Comma, BuildParens(Xdiff.getExpr_dx()),
+                                   BuildParens(Xdiff.getExpr())));
+      return BuildParens(Xdiff.getExpr());
```
vgvassilev, line 1415: We should figure out how to test this branch.

gojakuch: what do you mean, like add unit tests for this? this code is executed in the already existing tests if that's what you mean

vgvassilev: Codecov claims this code path is not tested. If that’s true (which is almost always the case) we should add a test.

lib/Differentiator/BaseForwardModeVisitor.cpp
gojakuch: sorry, I didn't understand your question, but this line `cond = cond->IgnoreParenImpCasts();` is needed (at least for now). this way we can check if the cond expression is supported or not, specifically, if there's an assignment operator like this:
```C++
double res;
for(;(r = x);){}
```
this cond is wrapped in implicit casts to bool and parentheses. once we can safely call `Visit(cond)` regardless of its type (that is, once there's support for almost most expressions as for-loop conditions), this line can be removed. but since now we need to check if that is an assignment operator, this is needed.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -668,18 +668,64 @@ StmtDiff BaseForwardModeVisitor::VisitForStmt(const ForStmt* FS) {
...
+
+  StmtDiff condDiff = Clone(FS->getCond());
+  Expr* cond = condDiff.getExpr();
+
+  // Declaration in the condition (if any) needs to be differentiated.
+  if (VarDecl* condVarDecl = FS->getConditionVariable()) {
+    VarDecl* condVarClone =
+        BuildVarDecl(condVarDecl->getType(), condVarDecl->getNameAsString(),
+                     Clone(condVarDecl->getInit()), condVarDecl->isDirectInit(),
+                     nullptr, condVarDecl->getInitStyle());
```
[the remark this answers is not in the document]
parth-07: We are repeating the code of `DifferentiateVarDecl` here just because we have to take the init as `null`. Can you reuse `DifferentiateVarDecl` here by adding an extra argument such as `ignoreInit` to `DifferentiateVarDecl` function?

lib/Differentiator/BaseForwardModeVisitor.cpp
vgvassilev: The way that it is written assumes that the line `Expr* cond = condDiff.getExpr();` may have implicit casts. Is that the case? If it is not my proposal was to move `cond->IgnoreParenImpCasts()` operation in the prior if-stmt.

lib/Differentiator/BaseForwardModeVisitor.cpp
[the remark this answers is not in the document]
gojakuch: that seems reasonable. I've implemented this. the thing is though, I needed to create a separate 2-argument method that is called from the 1-argument overload because some other classes specifically override the old single-argument method and this would be messed up by adding an optional second argument.

vgvassilev (approved):
LGTM! Let's wait for @parth-07's review.

parth-07 (approved):
Looks good.

vgvassilev:
@gojakuch, can you fix the compilation error we got here?

Issue #742, opened 2024-01-29
LLVM crashing when nested void functions are used in conjunction with multiple double* input values


stebos100:
I have recently been running an experiment and LLVM crashes. Running the function on its own provides the correct answer, but if I however try and run the script with clad::gradient on this function it crashes. Hopefully you can point me in the right direction. 

I just have a couple of questions.

1.) Can clad accommodate nested functions containing void functions, with passing by reference and generate the derivative function thereof ? ie lets say of the following form : 
---------------------------------------------------------------------------------------------------------------------------
```
__host__ void fnOne(double* &rate) {

    input[0] = input[0]*12;
    input[1] = input[1]*24;
    input[2] = input[2]*36;
}

__host__ void fnTwo(double* &rate) {
    input[0] = input[0]+ 12;
    input[1] = input[1]/24;
    input[2] = input[2] +0.058;
}

double fn (double* rate) {

fnOne (rate); //void
fnTwo(rate); //void 
double ret = fnThree(rate);

}

```

When I call the following function for clad::gradient(), I get an error of : "not viable: no known conversion from 'clad::array_ref<double> to 'const array_ref<double *> &' for first argument. 

---------------------------------------------------------------------------------------------------------------------------

2.) I have a question regarding functions with multiple double* as inputs. Lets say the function has the form of the following : 
---------------------------------------------------------------------------------------------------------------------------
```

double fn(double* IntRates, double* randomNumbers, double* fxRates) {
        
        fnOne (randomNumbers, IntRates ); //void
        fnTwo(fxRates, IntRates); //void 
        double ret = fnThree(IntRates);
}
```

---------------------------------------------------------------------------------------------------------------------------

firstly, can clad accomodate this if we only wanted to derive the function w.r.t the input double* intRates ? And if possible how would one specify this to the clad::gradient function ? I have tried the following and it crashes. 

---------------------------------------------------------------------------------------------------------------------------

```
int main() {
    double* intRates = (double*)malloc(120*sizeof(double));
    double* randomNumbers= (double*)malloc(120*sizeof(double)); 
    double* fxRates= (double*)malloc(120*sizeof(double)); 
    
    //Values are then assigned to the various slots
    
    auto fn_grad = clad::gradient(fn, "IntRates"); //Running this command alone crashes clad so I suspect I am doing something wrong here 
    
    double dx[120] = {0};
    clad::array_ref<double> dx_ref(dx, 120);
    
[...]

PR #906, opened 2024-05-24
Correctly handle pointer reference parameters in the reverse mode

PetroZarytskyi:
When differentiating ``void fn(double* &x)`` we used to get a pullback ``void fn_pullback(double* &x, double** _d_x)``, which causes crashes because in all other places ``_d_x`` is expected to be of type ``double*``. This was the root problem in the first point in #742 (the issue is old so it describes a mismatch between ``clad::array_ref<double>`` and ``clad::array_ref<double*>`` instead). This PR solves this problem.

Fixes the 1st point in #742.


parth-07 (commented):
@PetroZarytskyi Can you please rebase this pull-request on top of master?

vgvassilev:
Can you rebase once again?

vgvassilev:
Can we add a test?

parth-07 (approved):
Looks good, but needs more discussion in the future if we plan to support more complicated pointer cases.

vgvassilev:
@PetroZarytskyi, can you open an issue tracking @parth-07's concern?

PetroZarytskyi:
Hi @parth-07, thank you for the comment. I agree with your concerns. But I think having ``T*&`` as an adjoint type in the pullback signature is a bad solution because perhaps we'll have to do the same in the gradient overloads signature for consistency. But that would break the whole idea of all adjoint parameter types being convertible to ``void*``. 
In my opinion, a better solution is using ``T**`` for the adjoint as you mentioned. I'm now thinking this is what I should do in this PR instead.

vgvassilev:
If that would be a breaking change should go to the 2.0 branch.

PetroZarytskyi:
@vgvassilev Actually this PR changes the signature. The new suggestion was to keep it the way it had been previously. Honestly, I think Parth is right and I was mostly confused. I already implemented what I was talking about. I can push and if you have concerns I'll revert.

PetroZarytskyi:
@parth-07 What do you think now?

parth-07 (approved):
@PetroZarytskyi Thank you for addressing my concerns. The solution looks good to me.

PR #907, opened 2024-05-25
Return type adjusted to depend on called function's return type

ovdiiuv:
This commit changes the return type of 0 that is returned in case all arguments are constant literals. In particular it used to give warnings when differentiating NaN(or __builtin_nanf (const char * str )) which is crucial for the future fmax custom derivative implementation.

Fixes: https://github.com/vgvassilev/clad/issues/908.

### DIFF
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -1175,8 +1175,8 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
                                validLoc, llvm::MutableArrayRef<Expr*>(CallArgs),
                                validLoc)
                 .get();
-        auto* zero =
-            ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
+        auto* zero = ConstantFolder::synthesizeLiteral(CE->getType(), m_Context,
+                                                       /*val=*/0);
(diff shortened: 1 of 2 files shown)


vgvassilev:
Can you open an issue, if do not have one already opened for this?

Can you also put the same text in from the pull request description to the commit log?

parth-07 (commented):
A nitpick: Shouldn't the commit title be: 'Return type adjusted to depend on called function **return** type'?

ovdiiuv:
yeah, thanks for pointing out

> A nitpick: Shouldn't the commit title be: 'Return type adjusted to depend on called function **return** type'?

vgvassilev:
The commit message still differs from the PR description.

vgvassilev:
@ovdiiuv, can you address the clang-tidy recommendation?

vgvassilev:
If this fixes #908 the PR description and the commit message should include `Fixes #908`. Please take a look at our contributors guide: https://github.com/vgvassilev/clad/blob/master/CONTRIBUTING.md#creating-a-successfull-pull-request

ovdiiuv:
> If this fixes #908 the PR description and the commit message should include `Fixes #908`. Please take a look at our contributors guide: https://github.com/vgvassilev/clad/blob/master/CONTRIBUTING.md#creating-a-successfull-pull-request

There are two commits in this PR, would it make sense to add this to a message when squashing?

vgvassilev:
I did this for you.

Issue #830, opened 2024-03-19
Support for std::initializer_list


krathul:
differentiating functions containing std::initializer_list shows the "Not supported" warning and throws error.
```c
int fn(int x) {
  int res = 0;
  auto &&range = {1, 2, 3};
  for (auto i = range.begin(); i != range.end(); i++) {
    res += x * (*i);
  }
  return res;
}
```

Deepanshu9376:
I think because you're trying to use a range-based for loop with an std::initializer_list, which doesn't support non-const iterators.  Instead, you can iterate over the elements directly without using iterators.
Like this :-)
int fn(int x) {
  int res = 0;
  auto range = {1, 2, 3};
  for (auto val : range) {
    res += x * val;
  }
  return res;
}

vgvassilev:
This depends on https://github.com/vgvassilev/clad/issues/724.

PR #909, opened 2024-05-26
Add primitive support for constructor custom pushforward functions

parth-07:
This commit adds primitive support for custom pushforward functions for
constructors. Custom constructor pushforward function support will enable the
below features:
- Class differentiation support for classes whose constructor Clad
  cannot automatically differentiate. Now, We can enable differentiation of
  entire C++ standard library by providing custom derivatives.
- Remove the restriction of default-constructible for class types. This
  was a troublesome restriction. Now, the only restriction for class
  types is to have a sensible copy-constructor. That is, copy
  constructor should copy the class members and after copy-construction,
  both the objects should be equivalent, mathematically speaking.

Constructor pushforward functions differ from ordinary pushforward
functions in two important ways:
- Constructor pushforward functions initialize the primal class object
  and the corresponding derivative object. Ordinary member function
  pushforwards takes an already-existing primal class object and the
  corresponding derivative object as inputs.
- Constructor pushforward functions return a value even though
  constructor do not return anything. Constructor pushforward functions
  return initialized primal object and the derivative object. These are
  then used to initialize primal object and the derivative in the
  derivative function code.

How to write custom constructor pushforward functions
----------------------------------------

Let's see how to write custom pushforward function for a
constructor:

- Custom constructor pushforwards must have the name `constructor_pushforward`
- Custom constructor pushforwards must be defined in
  `::clad::custom_derivatives::class_functions` namespace.
- The parameters of the custom constructor pushforward must be:
  {`::clad::Identify<Class>`, original params..., derivative params...}.

'original parameters...' and 'derivative parameters...' is same as what
we have for other pushforward functions. We will soon see why do we need
`::clad::Identify<T>` for constructor custom pushforwards.

Let's see a basic example of how to write custom constructor
pushforward.

```cpp
class Coordinates {
  Coordinates(double px, double py, double pz) :
    x(px), y(py), z(pz) {}

  public:
  double x, y, z;
}

namespace clad {
namespace class_functions {
// custom constructor pushforward function
clad::ValueAndPushforward<Coordinates, Coordinates>
constructor_pushforward(clad::Identify<Coordinates>, double x, double y,
                        double z, double d_x, double d_y, double d_z) {
  return {Coordinates(x, y, z), Coordinates(d_x, d_y, d_z) };
}
} // namespace class_functions
} // namespace custom_derivatives
} // namespace clad

// custom constructor pushforward is used as follows:
// primal code
Constructor c(u, v, w);

// derivative code
clad::ValueAndPushforward<Coordinates, Coordinates> _t0 =
  constructor_pushforward(clad::Identify<Coordinates>, u, v, w, _d_u,
    _d_v, _d_w);
Coordinates _d_c = _t0.pushforward;
Coordinates c = _t0.value;
```

Now, let's see a bit advanced example based on `std::vector` constructor.

```cpp
namespace clad {
namespace class_functions {
// Custom pushforward for: vector(size_t n, const typename
::std::vector<T>::allocator_type alloc)
template <typename T>
clad::ValueAndPushforward<::std::vector<T>, ::std::vector<T>>
constructor_pushforward(
    Identify<::std::vector<T>>, size_t n,
    const typename ::std::vector<T>::allocator_type alloc, size_t d_n,
    const typename ::std::vector<T>::allocator_type d_alloc) {
  ::std::vector<T> v(n, alloc);
  ::std::vector<T> d_v(n, 0, alloc);
  return {v, d_v};
}

// Custom pushfoward for: vector(size_t n, T val, const typename
::std::vector<T>::allocator_type alloc)
template <typename T>
clad::ValueAndPushforward<::std::vector<T>, ::std::vector<T>>
constructor_pushforward(
    Identify<::std::vector<T>>, size_t n, T val,
    const typename ::std::vector<T>::allocator_type alloc, size_t d_n, T d_val,
    const typename ::std::vector<T>::allocator_type d_alloc) {
  ::std::vector<T> v(n, val, alloc);
  ::std::vector<T> d_v(n, d_val, alloc);
  return {v, d_v};
}
} // namespace class_functions
} // namespace custom_derivatives
} // namespace clad

// The custom constructor pushforwards is used as follows:

// Primal code:
std::vector<double> v(10, u);

// Derivative code:
clad::ValueAndPushforward<std::vector<double>, std::vector<double>> _t0 =
  clad::custom_derivatives::class_functions::constructor_pushforward(
    clad::Identify<std::vector<double> >(), 10, u, allocator_type(),
    0, _d_u, allocator_type());
std::vector<double> d_v = _t0.pushforward;
std::vector<double> v = _t0.value;
```

Why `clad::Identify<T>`?
------------------------

So, why do we need clad::Identify<T>?

For a constructor that takes two parameters of types `size_t` and `double`,
the custom pushforward will have the following signature if we do not
include `::clad::Identify<T>`:

```cpp
clad::ValueAndPushforward<Class, Class> constructor_pushforward(size_t n,
  double val, size_t d_n, double d_val);
```

Now, the question is: How to distinguish custom constructor pushforwards
for different classes?

```cpp
MyClassA a(3, 5.0);
MyClassB b(7, 9.0);
```

There is no way for overload resolution selector to distinguish
constructor_pushforward for classes `MyClassA` and `MyClassB`.

`clad::Identify<T>` is used to identify the class for which
custom constructor pushforward is defined.

Please note that we cannot use the same strategy which we use for custom
member function pushforwards because member function pushforwards
always have parameters of the class type which are used for identifying
the class.

We also cannot simply ask users to define the pushforwards inside the
declaration context of the class because it may not always be feasible
to modify the source code of external libraries.

--------------------------------------------

If anyone has better ideas for interface for defining custom constructor
pushforward function, then definitely share them!

### DIFF
--- a/include/clad/Differentiator/BaseForwardModeVisitor.h
+++ b/include/clad/Differentiator/BaseForwardModeVisitor.h
@@ -3,6 +3,7 @@
 
 #include "Compatibility.h"
 #include "VisitorBase.h"
+#include "clang/AST/ExprCXX.h"
 #include "clang/AST/RecursiveASTVisitor.h"
 #include "clang/AST/StmtVisitor.h"
 #include "clang/Sema/Sema.h"
@@ -140,6 +141,16 @@ protected:
   /// \return active switch case label after processing `stmt`
   clang::SwitchCase* DeriveSwitchStmtBodyHelper(const clang::Stmt* stmt,
                                                 clang::SwitchCase* activeSC);
+
+  /// Tries to build custom derivative constructor pushforward call for the
+  /// given CXXConstructExpr.
+  ///
+  /// \return A call expression if a suitable custom derivative is found;
+  /// Otherwise returns nullptr.
+  clang::Expr* BuildCustomDerivativeConstructorPFCall(
+      const clang::CXXConstructExpr* CE,
+      llvm::SmallVectorImpl<clang::Expr*>& clonedArgs,
+      llvm::SmallVectorImpl<clang::Expr*>& derivedArgs);
 };
 } // end namespace clad
 
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -21,6 +21,26 @@ template <typename T, typename U> struct ValueAndPushforward {
   T value;
   U pushforward;
 };
+
+/// It is used to identify constructor custom pushforwards. For
+/// constructor custom pushforward functions, we cannot use the same
+/// strategy which we use for custom pushforward for member
+/// functions. Member functions custom pushforward have the following
+/// signature:
+///
+/// mem_fn_pushforward(ClassName *c, ..., ClassName *d_c, ...)
+///
+/// We use the first argument 'ClassName *c' to determine the class of member
+/// function for which the pushforward is defined.
+///
+/// In the case of constructor pushforward, there are no objects of the class
+/// type passed to the constructor. Therefore, we cannot simply use arguments
+/// to determine the class. To solve this, 'ConstructorPushforwardTag<T>' is
+/// used. A custom_derivative pushforward for constructor is required to have
+/// 'ConstructorPushforwardTag<T>' as the first argument, where 'T' is the
+/// class for which constructor pushforward is defined.
+template <class T> class ConstructorPushforwardTag {};
+
 namespace custom_derivatives {
 #ifdef __CUDACC__
 template <typename T>
--- a/include/clad/Differentiator/STLBuiltins.h
+++ b/include/clad/Differentiator/STLBuiltins.h
@@ -1,8 +1,9 @@
 #ifndef CLAD_STL_BUILTINS_H
 #define CLAD_STL_BUILTINS_H
 
+#include <clad/Differentiator/BuiltinDerivatives.h>
+#include <initializer_list>
 #include <vector>
-#include "clad/Differentiator/BuiltinDerivatives.h"
 
 namespace clad {
 namespace custom_derivatives {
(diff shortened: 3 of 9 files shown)


lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -2043,4 +2080,9 @@ BaseForwardModeVisitor::DifferentiateStaticAssertDecl(
     const clang::StaticAssertDecl* SAD) {
   return DeclDiff<StaticAssertDecl>();
 }
+
+StmtDiff BaseForwardModeVisitor::VisitCXXStdInitializerListExpr(
+    const clang::CXXStdInitializerListExpr* ILE) {
+  return Visit(ILE->getSubExpr());
```
vgvassilev, line 2086: Would that resolve #830?

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -21,6 +21,25 @@ template <typename T, typename U> struct ValueAndPushforward {
...
+///
+/// We use the first argument 'ClassName *c' to determine the class of member
+/// function for which the pushforward is defined.
+///
+/// In the case of constructor pushforward, there are no objects of the class
+/// type passed to the constructor. Therefore, we cannot simply use arguments to
+/// determine the class. To solve this, Identify<T> is used. A custom_derivative
+/// pushforward for constructor is required to have 'Identify<T>' as the first
+/// argument, where T is the class for which constructor pushforward is defined.
+template <class T> class Identify {};
```
vgvassilev, line 41:
```suggestion
template <class T> class ConstructorPushfowardTag {};
```

It seems that the proposed design uses a technique called "tag dispatching" to help the overload resolution. Let's use that in the name.

include/clad/Differentiator/STLBuiltins.h
```diff
@@ -26,6 +28,87 @@ void resize_pushforward(::std::vector<T>* v, unsigned sz, U val,
   d_v->resize(sz, d_val);
   v->resize(sz, val);
 }
+
+template <typename T>
+clad::ValueAndPushforward<::std::vector<T>, ::std::vector<T>>
+constructor_pushforward(
+    Identify<::std::vector<T>>, typename ::std::vector<T>::size_type n,
```
vgvassilev, line 35: The tag dispatcher generally goes last in the function signature.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1850,6 +1850,45 @@ BaseForwardModeVisitor::VisitCXXConstructExpr(const CXXConstructExpr* CE) {
...
+  QualType identifyClassT =
+      GetCladIdentityOfType(CE->getType().withoutLocalFastQualifiers());
+  // Builds clad::Identify<T> declaration
+  Expr* identifyClassArg =
+      m_Sema
+          .BuildCXXTypeConstructExpr(
+              m_Context.getTrivialTypeSourceInfo(identifyClassT,
+                                                 utils::GetValidSLoc(m_Sema)),
+              noLoc, MultiExprArg{}, noLoc, /*ListInitialization=*/false)
+          .get();
```
vgvassilev, line 1865: Can we move this code in a separate function with appropriate name? This will help the code readibility.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1850,6 +1850,45 @@ BaseForwardModeVisitor::VisitCXXConstructExpr(const CXXConstructExpr* CE) {
...
+  customPushforwardArgs.push_back(identifyClassArg);
+  customPushforwardArgs.append(clonedArgs.begin(), clonedArgs.end());
+  customPushforwardArgs.append(derivedArgs.begin(), derivedArgs.end());
+  std::string customPushforwardName =
+      clad::utils::ComputeEffectiveFnName(CE->getConstructor()) +
+      GetPushForwardFunctionSuffix();
+  Expr* pushforwardCall = m_Builder.BuildCallToCustomDerivativeOrNumericalDiff(
+      customPushforwardName, customPushforwardArgs, getCurrentScope(),
+      const_cast<DeclContext*>(CE->getConstructor()->getDeclContext()));
```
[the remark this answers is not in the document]
vgvassilev: Don't we need to operate on the cloned `CE`?

lib/Differentiator/VisitorBase.cpp
```diff
@@ -812,4 +812,15 @@ namespace clad {
     return m_Sema.ActOnCallExpr(getCurrentScope(), pushDRE, noLoc, args, noLoc)
         .get();
   }
+
+  clang::TemplateDecl* VisitorBase::GetCladIdentifyDecl() {
+    static TemplateDecl* cladIdentifyDecl = nullptr;
```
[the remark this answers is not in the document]
vgvassilev: We cache the these runtime utilities as members generally. Can we do the same here?

vgvassilev (commented):
@PetroZarytskyi, can you take a look?

include/clad/Differentiator/BuiltinDerivatives.h
parth-07: Yes, that's a good idea. I will make this change.

include/clad/Differentiator/STLBuiltins.h
parth-07: I am not too sure about this. If we put the tag dispatcher (`ConstructorPushforwardTag`) at end, then it might be difficult to find constructor pushforwards for a class which has 5 constructors with different number of parameters. I feel that it is easy to spot (dare I say,`Identify` :) ) constructor pushforwards for a particular class by seeing the first parameter of constructor pushforwards as compared to the last parameter. Please share your thoughts on this.

lib/Differentiator/BaseForwardModeVisitor.cpp
parth-07: Yes, definitely. I will make this change.

lib/Differentiator/BaseForwardModeVisitor.cpp
parth-07: Yes, it fixes the issue for forward-mode. However, the issue does not state forward-mode / reverse-mode, so I am not sure if we can close the issue because we still do not support initializer list in the reverse-mode.

parth-07: I have fixed this in a separate pull-request: #935

lib/Differentiator/BaseForwardModeVisitor.cpp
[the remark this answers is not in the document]
parth-07: We do not need to clone `CXXConstructExpr* CE` here because we are not (directly) using the `CXXConstructExpr` in the derivative code. We are using `CE` here to get the class (declaration context) of the constructor.

lib/Differentiator/VisitorBase.cpp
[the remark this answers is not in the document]
parth-07: Done. We use a member variable to cache the result now.

include/clad/Differentiator/BuiltinDerivatives.h
parth-07: Done.

lib/Differentiator/BaseForwardModeVisitor.cpp
parth-07: I have added a helper function `BuildCustomDerivativeConstructorPFCall` for this.

include/clad/Differentiator/STLBuiltins.h
vgvassilev: Okay, I do not have a strong preference here. I believe being a first argument will work out, too because what we do here is slightly different than the canonical tag dispatch practice in C++.

vgvassilev (approved):
LGTM! Please address the clang-tidy reports they seem good. The `const_cast` report can be suppressed with:

```
```

include/clad/Differentiator/BaseForwardModeVisitor.h
```diff
@@ -8,6 +8,7 @@
 #include "clang/Sema/Sema.h"
 
 #include <array>
+#include <clang/AST/ExprCXX.h>
```
vgvassilev, line 11: That include seems not at the right place. Can you fix it?

parth-07: I have fixed it.

vgvassilev (approved):
LGTM!

PR #910, opened 2024-05-27
Add bugprone-argument-comment in clang-tidy

vaithak:


vaithak:
This doesn't work for all the cases; I tried to debug several instances but couldn't find the reason behind why. Still adding this, to catch whatever it can.

PR #916, opened 2024-05-31
Add support for `[[clad::non_differentiable]]` in reverse mode

### LINKED ISSUE #717: Support [[clad::non_differentiable]] in the reverse-mode

MihailMihov:
fixes #717 

I've added the code for handing `[[clad::non_differentiable]]` that is already present in the forward mode visitor to the one for reverse mode. I also modified the tests from forward mode, but `ReverseMode/NonDifferentiable.C` is currently failing because of an issue with differentiating operator overloads in reverse mode.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1420,6 +1420,26 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       return StmtDiff(Clone(CE));
     }
 
+    // If the function is non_differentiable, return zero derivative.
+    if (clad::utils::hasNonDifferentiableAttribute(CE)) {
+      // Calling the function without computing derivatives
+      llvm::SmallVector<Expr*, 4> ClonedArgs;
+      for (unsigned i = 0, e = CE->getNumArgs(); i < e; ++i)
+        ClonedArgs.push_back(Clone(CE->getArg(i)));
+
+      SourceLocation validLoc = clad::utils::GetValidSLoc(m_Sema);
+      Expr* Call = m_Sema
+                       .ActOnCallExpr(getCurrentScope(), Clone(CE->getCallee()),
+                                      validLoc, ClonedArgs, validLoc)
+                       .get();
+      // Creating a zero derivative
+      auto* zero = ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context,
+                                                     /*val=*/0);
+
+      // Returning the function call and zero derivative
+      return StmtDiff(Call, zero);
+    }
+
     auto NArgs = FD->getNumParams();
     // If the function has no args and is not a member function call then we
     // assume that it is not related to independent variables and does not
@@ -2061,6 +2081,13 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     } else if (opCode == UnaryOperatorKind::UO_Deref) {
       diff = Visit(E);
       Expr* cloneE = BuildOp(UnaryOperatorKind::UO_Deref, diff.getExpr());
+
+      // If we have a pointer to a member expression, which is
+      // non-differentiable, we just return a clone of the original expression.
+      if (auto* ME = dyn_cast<MemberExpr>(diff.getExpr()))
+        if (clad::utils::hasNonDifferentiableAttribute(ME->getMemberDecl()))
+          return {cloneE};
+
       Expr* diff_dx = diff.getExpr_dx();
       bool specialDThisCase = false;
       Expr* derivedE = nullptr;
@@ -2655,9 +2682,14 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     // If `VD` is a reference to a non-local variable then also there's no
     // need to call `Visit` since non-local variables are not differentiated.
     if (!isDerivativeOfRefType && (!isPointerType || isInitializedByNewExpr)) {
-      Expr* derivedE = BuildDeclRef(VDDerived);
-      if (isInitializedByNewExpr)
-        derivedE = BuildOp(UnaryOperatorKind::UO_Deref, derivedE);
+      Expr* derivedE = nullptr;
+
+      if (!clad::utils::hasNonDifferentiableAttribute(VD)) {
+        derivedE = BuildDeclRef(VDDerived);
+        if (isInitializedByNewExpr)
+          derivedE = BuildOp(UnaryOperatorKind::UO_Deref, derivedE);
+      }
+
       if (VD->getInit()) {
         if (isa<CXXConstructExpr>(VD->getInit()))
           initDiff = Visit(VD->getInit());
@@ -2689,6 +2721,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         addToCurrentBlock(assignToZero, direction::reverse);
       }
     }
+
     VarDecl* VDClone = nullptr;
     Expr* derivedVDE = nullptr;
     if (VDDerived)
@@ -2815,19 +2848,21 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     if (declsBegin != DS->decls().end() && isa<VarDecl>(*declsBegin)) {
       auto* VD = dyn_cast<VarDecl>(*declsBegin);
       QualType QT = VD->getType();
-      if (!QT->isPointerType()) {
-        auto* typeDecl = QT->getAsCXXRecordDecl();
-        // We should also simply copy the original lambda. The differentiation
-        // of lambdas is happening in the `VisitCallExpr`. For now, only the
-        // declarations with lambda expressions without captures are supported.
-        isLambda = typeDecl && typeDecl->isLambda();
-        if (isLambda) {
-          for (auto* D : DS->decls())
-            if (auto* VD = dyn_cast<VarDecl>(D))
-              decls.push_back(VD);
-          Stmt* DSClone = BuildDeclStmt(decls);
-          return StmtDiff(DSClone, nullptr);
-        }
+      if (QT->isPointerType())
+        QT = QT->getPointeeType();
+
+      auto* typeDecl = QT->getAsCXXRecordDecl();
+      // We should also simply copy the original lambda. The differentiation
+      // of lambdas is happening in the `VisitCallExpr`. For now, only the
+      // declarations with lambda expressions without captures are supported.
+      isLambda = typeDecl && typeDecl->isLambda();
+      if (isLambda ||
+          (typeDecl && clad::utils::hasNonDifferentiableAttribute(typeDecl))) {
+        for (auto* D : DS->decls())
+          if (auto* VD = dyn_cast<VarDecl>(D))
+            decls.push_back(VD);
+        Stmt* DSClone = BuildDeclStmt(decls);
+        return StmtDiff(DSClone, nullptr);
       }
     }
 
@@ -2839,6 +2874,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     for (auto* D : DS->decls()) {
       if (auto* VD = dyn_cast<VarDecl>(D)) {
         DeclDiff<VarDecl> VDDiff;
+
         if (!isLambda)
           VDDiff = DifferentiateVarDecl(VD);
 
@@ -3014,6 +3050,11 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
            "CXXMethodDecl nodes not supported yet!");
     MemberExpr* clonedME = utils::BuildMemberExpr(
         m_Sema, getCurrentScope(), baseDiff.getExpr(), field->getName());
+    if (clad::utils::hasNonDifferentiableAttribute(ME)) {
+      auto* zero = ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context,
+                                                     /*val=*/0);
+      return {clonedME, zero};
+    }
     if (!baseDiff.getExpr_dx())
       return {clonedME, nullptr};
     MemberExpr* derivedME = utils::BuildMemberExpr(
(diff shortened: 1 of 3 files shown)


parth-07:
> but ReverseMode/NonDifferentiable.C is currently failing because of an issue with differentiating operator overloads in reverse mode.

If we mark that operator overload as non-differentiable, then should the issue still happen?

MihailMihov:
> If we mark that operator overload as non-differentiable, then should the issue still happen?

I just tried changing the test and it still errors, but this time it is `too many arguments to function call, expected 1, have 2`, while if it tries to differentiate the operator it is `too few arguments to function call, expected 4, have 3`. I did open an issue for this and it is #917, where I put a more minimal failing test. I can try and look into the case where it fails when non-differentiable, but if I had to guess they are the same issue and fixing one would fix both cases.

parth-07:
> but if I had to guess they are the same issue and fixing one would fix both cases.

Oh, yes, you are right. Thank you for the details.

vgvassilev:
@MihailMihov, can you rebase this pull request?

vgvassilev (commented):
Can you take care of the clang-tidy suggestions and adapt the new test in such a way that we work around the pre-existing issue of #917?

test/Gradient/NonDifferentiable.C
```diff
@@ -0,0 +1,157 @@
...
+#include "clad/Differentiator/Differentiator.h"
+
+extern "C" int printf(const char* fmt, ...);
+
+class SimpleFunctions1 {
+public:
+  SimpleFunctions1() noexcept : x(0), y(0) {}
+  SimpleFunctions1(double p_x, double p_y) noexcept : x(p_x), y(p_y) {}
+  double x;
+  non_differentiable double y;
```
vaithak, line 16: Let's also test with some pointer member types.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2954,6 +2977,10 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
            "CXXMethodDecl nodes not supported yet!");
     MemberExpr* clonedME = utils::BuildMemberExpr(
         m_Sema, getCurrentScope(), baseDiff.getExpr(), field->getName());
+    auto* zero = ConstantFolder::synthesizeLiteral(m_Context.DoubleTy,
+                                                   m_Context, /*val=*/0);
```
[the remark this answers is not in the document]
vaithak: I also saw that this is 'DoubleTy` in forward mode. Can you also test that too and fix that in a separate PR, if possible?

test/Gradient/NonDifferentiable.C
MihailMihov: I just added an `fn_s1_field_pointer` to the test, is that what you wanted me to test?

lib/Differentiator/ReverseModeVisitor.cpp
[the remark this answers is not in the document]
MihailMihov: I did change the `DoubleTy` to an `IntTy`, but that didn't actually do anything to the tests and pointers didn't work either way. I added another check for them when visiting `UO_Deref`, not sure if there isn't a better way to fix them however, but now the new test is passing.

test/Gradient/NonDifferentiable.C
```diff
@@ -0,0 +1,176 @@
+// RUN: %cladclang %s -I%S/../../include -oNonDifferentiable.out 2>&1 | %filecheck %s
+// RUN: ./NonDifferentiable.out | %filecheck_exec %s
+// CHECK-NOT: {{.*error|warning|note:.*}}
+
+#define non_differentiable __attribute__((annotate("another_attribute"), annotate("non_differentiable")))
+
+#include "clad/Differentiator/Differentiator.h"
+
+extern "C" int printf(const char* fmt, ...);
```
parth-07, line 9: Do we need this declaration?

test/Gradient/NonDifferentiableError.C
```diff
@@ -0,0 +1,42 @@
...
+
+#define TEST_FUNC(name, i, j)                                                  \
+  auto d_##name = clad::differentiate(&name, "i");                             \
+  printf("%.2f\n", d_##name.execute(i, j));                                    \
+  printf("\n");
+
+int main() {
+  INIT_EXPR(SimpleFunctions2);
+  TEST_CLASS(SimpleFunctions2, mem_fn, 3, 5);
+  TEST_FUNC(fn_s2_mem_fn, 3, 5);  // expected-error {{attempted differentiation of function 'fn_s2_mem_fn', which is marked as non-differentiable}}
```
parth-07, line 41: What happens if we provide a custom derivative for a function marked as non-differentiable?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1369,6 +1369,27 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
     }
 
+    SourceLocation validLoc{CE->getBeginLoc()};
+
+    // If the function is non_differentiable, return zero derivative.
+    if (clad::utils::hasNonDifferentiableAttribute(CE)) {
+      // Calling the function without computing derivatives
+      llvm::SmallVector<Expr*, 4> ClonedArgs;
+      for (unsigned i = 0, e = CE->getNumArgs(); i < e; ++i)
+        ClonedArgs.push_back(Clone(CE->getArg(i)));
```
parth-07, line 1379: Simply cloning the argument seems incorrect. What if the arguments have side-effect which can affect the derivative computation?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1369,6 +1369,27 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+
+      Expr* Call = m_Sema
+                       .ActOnCallExpr(getCurrentScope(), Clone(CE->getCallee()),
+                                      validLoc, ClonedArgs, validLoc)
```
parth-07, line 1383: Does the actual value of location matter here? If not, then I prefer to use `clad::utils::GetValidSLoc`. `clad::utils::GetValidSLoc` makes it clear that the actual value of the source location is not relevant.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2010,6 +2031,13 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     } else if (opCode == UnaryOperatorKind::UO_Deref) {
       diff = Visit(E);
       Expr* cloneE = BuildOp(UnaryOperatorKind::UO_Deref, diff.getExpr());
+
+      // If we have a pointer to a member expression, which is
+      // non-differentiable, we just return a clone of the original expression.
+      if (auto* ME = dyn_cast<MemberExpr>(diff.getExpr()))
```
parth-07, line 2037: Can it be handled more uniformly in `VisitMemberExpr`?

parth-07: If `diff.getExpr_dx()` is `0`, then we would need to add a special condition here.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2954,6 +2984,10 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
            "CXXMethodDecl nodes not supported yet!");
     MemberExpr* clonedME = utils::BuildMemberExpr(
         m_Sema, getCurrentScope(), baseDiff.getExpr(), field->getName());
+    auto* zero = ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context,
```
parth-07, line 2987: I think it is better to create `zero` inside the if-condition as it is only used there.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2755,19 +2783,21 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
-        }
+      if (QT->isPointerType())
+        QT = QT->getPointeeType();
+
+      auto* typeDecl = QT->getAsCXXRecordDecl();
+      // We should also simply copy the original lambda. The differentiation
+      // of lambdas is happening in the `VisitCallExpr`. For now, only the
+      // declarations with lambda expressions without captures are supported.
+      isLambda = typeDecl && typeDecl->isLambda();
+      if (isLambda ||
```
parth-07, line 2794: Can you please add a test for a local variable declaration with non_differentiable attribute?

parth-07 (changes requested):

test/Gradient/NonDifferentiable.C
MihailMihov: I don't actually know what it's purpose is, but it was is ForwardMode/NonDifferentiable.C, which I copied to make this test. But it's also in 28 other tests. If removed, the test still passes on my machine.

test/Gradient/NonDifferentiableError.C
MihailMihov: It still fails with the error that it's non-differentiable. I think that's better than ignoring the attribute and using the derivative, but what do you expect?

lib/Differentiator/ReverseModeVisitor.cpp
MihailMihov: The value is probably only for diagnostics, but I agree that `GetValidSLoc` would be better, so I have changed it. By the way all of this code is taken from the forward mode and mostly kept as is, so after this is merged should I open a PR for the same things but in forward mode?

lib/Differentiator/ReverseModeVisitor.cpp
MihailMihov: > Can it be handled more uniformly in `VisitMemberExpr`?

It is already handled in `VisitMemberExpr` where we return `{clonedME, zero}`, but what happened without the above check is that it tries to build something along the lines of `*0 += ...`, when visiting the `UO_Deref`.

> If `diff.getExpr_dx()` is `0`, then we would need to add a special condition here.

With the above check I eliminate one of the cases where we could end up with a 0 above, if you can think of anything else, then we should handle those too. Do you have anything in mind?

lib/Differentiator/ReverseModeVisitor.cpp
MihailMihov: I have changed that, also what do you think about this being `IntTy` or `DoubleTy`. In the forward mode code it was `DoubleTy`, but Vaibhav suggested changing it to `IntTy`. It didn't seem to make any difference, but maybe somewhere it will?

lib/Differentiator/ReverseModeVisitor.cpp
MihailMihov: I just added `fn_non_diff_var` to Gradient/NonDifferentiable.C test, but I believe it's not working as expected. The correct output would be 0.00 0.00 right? I'll try to get that fixed now.

lib/Differentiator/ReverseModeVisitor.cpp
MihailMihov: I'm not sure that I understand the issue here. It the arguments do have side effects then those would be kept when we clone them, is that not what is expected? When do you think that this wouldn't work correctly?

parth-07: Consider an example such as this:

```cpp
some_non_differentiable_fn_call(r = u * v, s = u + v); 
```

Now, if we simply clone the arguments then we will not generate adjoint statements for `r = u * v` and `s = u + v`. 

You don't necessarily need to fix this issue in this PR.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: > so after this is merged should I open a PR for the same things but in forward mode?

Yes, please do that.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: Can you please create an issue for this and resolve it in a follow-up pull-request?

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: > if you can think of anything else, then we should handle those too.

I am concerned there might be many such cases... For example, `->` operator. 

It might be better to test if the `diff.getExpr_dx()` is a constant (or `0`) instead of testing if the member has a non-differentiable attribute. This is because it will help us cover more cases. For example, the adjoint of member expressions of global class objects should also be `0` and consequently they should be handled similarly but they do not have `non_differentiable` attribute.

test/Gradient/NonDifferentiableError.C
parth-07: I agree. It is better than ignoring the attribute. The current behavior seems good then.

parth-07 (approved):
Looks good as an initial support for `[[non_differentiable]]`. Thank you for working on this.

Can you please open issues for the comments which need some future work:

- Differentiating argument expressions of non differentiable function calls. This is required because argument expressions can have side-affects.
- Supporting non differentiate attribute on local variables.
- Improve handling of non_differentiable variables in expressions such as `*` (dereference operator), `->` and so on.

vgvassilev:
Can you squash the tests into the other commit?

MihailMihov:
> Can you please open issues for the comments which need some future work:
> 
>     * Differentiating argument expressions of non differentiable function calls. This is required because argument expressions can have side-affects.
> 
>     * Supporting non differentiate attribute on local variables.
> 
>     * Improve handling of non_differentiable variables in expressions such as `*` (dereference operator), `->` and so on.

I opened issues for 1 and 3. For 2 this PR includes a basic test and fix, but more work may be necessary. Do I just create an issue saying that there might be a cleaner fix or something more specific?

vgvassilev:
> > Can you please open issues for the comments which need some future work:
> > ```
> > * Differentiating argument expressions of non differentiable function calls. This is required because argument expressions can have side-affects.
> > 
> > * Supporting non differentiate attribute on local variables.
> > 
> > * Improve handling of non_differentiable variables in expressions such as `*` (dereference operator), `->` and so on.
> > ```
> 
> I opened issues for 1 and 3. For 2 this PR includes a basic test and fix, but more work may be necessary. Do I just create an issue saying that there might be a cleaner fix or something more specific?

Yes, we should create an issue describing what this "more work may be necessary" means in technical and practical terms.

PR #918, opened 2024-05-31
Fix clang tidy run failure in PR checks

vaithak:
With LLVM and Clang >= 16, clang-tidy was returning the following error: 
```
unknown argument: '-fno-lifetime-dse' [clang-diagnostic-error]
```

### DIFF
--- a/.github/workflows/clang-tidy-review.yml
+++ b/.github/workflows/clang-tidy-review.yml
@@ -18,9 +18,9 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install LLVM and Clang
-        uses: KyleMayes/install-llvm-action@v2.0.2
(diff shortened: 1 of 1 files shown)


vgvassilev:
> unknown argument: '-fno-lifetime-dse'

This is a gcc-only flag and when we build the compilation database this becomes a problem. When running the clang-based tools such as clang-tidy we should build clad with clang to avoid propagating such flags.

vaithak:
> This is a gcc-only flag and when we build the compilation database this becomes a problem. When running the clang-based tools such as clang-tidy we should build clad with clang to avoid propagating such flags.

Fixed this 👍🏼

vgvassilev (approved):
Let's give it a shot.

PR #919, opened 2024-06-04
Improve derived variable init for const pointers

vaithak:
This will fix the segmentation fault in roofit benchmarks (when auxiliary arrays are removed).


vaithak:
cc: @guitargeek. I have tested this with the atlas likelihood benchmark code, it now works after removing the auxiliary arrays. The runtime results have also improved because of this.

vaithak:
The issue was that while calling pullbacks, we don't maintain non-differentiable arguments. @PetroZarytskyi  had fixed this for most of the cases here: <https://github.com/vgvassilev/clad/pull/802>, but it still had some more complex cases which required activity analysis, as he mentioned in the PR too.
My PR just tries to improve upon that by handling the case for const pointers, which can be handled without activity analysis too.

vgvassilev (commented):
LGTM but I'd wait for @PetroZarytskyi's review.

PetroZarytskyi:
Looks good to me. Sure, this will probably become obsolete once we have activity analysis but this is a great simplification for now.

Issue #911, opened 2024-05-27
Add support for the logical negation operator in conditions (forward mode)


gojakuch:
There's currently no support for the unary logical not operator in general. In particular, adding support for the unary negation operator should allow correct differentiation of such conditions (originally mentioned by @parth-07 in #905):

```C++
#include "clad/Differentiator/Differentiator.h"
#include <iostream>
#define show(x) std::cout << #x << ": " << x << "\n";

double fn(double u, double v) {
    double res = 0;
    for (; !(res = u * v) ;) {}
    return res;
}

int main() {
    auto fn_diff = clad::differentiate(fn, "u");
    show(fn(3, 5));
    show(fn_diff.execute(3, 5));
}
```

Output:
fn_diff.execute(3, 5): 0

Expected Output:
fn_diff.execute(3, 5): 5

PR #920, opened 2024-06-04
Add support for the logical not operator in for-loop `cond`s (forward mode)

gojakuch:
This commit adds support for logical negation operators with side-effects inside for-loop conditions in the forward mode.

Fixes: #911


gojakuch:
if anything, this PR adds some changes that I first implemented in #905 and then reverted, since it was too far of a stretch for a single PR

vgvassilev (approved):
LGTM!

PR #924, opened 2024-06-07
Simplify the Jacobian visitors. NFC

vgvassilev:


vgvassilev:
Codecov is right to complain, when we remove code we generally reduce coverage but having less code while keeping the same feature set is generally a good thing.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -371,8 +370,8 @@ static void registerDerivative(FunctionDecl* derivedFD, Sema& semaRef) {
       HessianModeVisitor H(*this);
       result = H.Derive(FD, request);
     } else if (request.Mode == DiffMode::jacobian) {
-      JacobianModeVisitor J(*this);
-      result = J.Derive(FD, request);
+      ReverseModeVisitor R(*this);
+      result = R.Derive(FD, request);
```
vaithak, line 374: Looks good.
Maybe we can merge the if conditions for `DiffMode::reverse` and `DiffMode::jacobian` together.
```cpp
else if (request.Mode == DiffMode::reverse || request.Mode == DiffMode::jacobian) {
```

vaithak (approved):

lib/Differentiator/DerivativeBuilder.cpp
vgvassilev: I am planning to rework that part in a subsequent PR.

PR #925, opened 2024-06-07
Remove unnecessary usages of clad::array_ref and update the documentation

PetroZarytskyi:
This PR removes unnecessary usages of clad::array_ref from the codebase (mostly custom pullbacks) and the documentation. Also, it updates the error estimation documentation.


vgvassilev (approved):
LGTM!

PR #928, opened 2024-06-07
A reverse block is never ended in RMV::VisitConditionalStatement

### LINKED ISSUE #922: Regression in `clad` master in dealing with templated functions
This was noticed when running the RooFit unit tests with clad `master`.

Reproducer (ROOT macro, but should be easy to turn into compiled executable):
```c++
template <bool pdfMode>
inline double polynomial(double const *coeffs, int nCoeffs, int lowestOrder, double x)
{
   double retVal = coeffs[nCoeffs - 1];
   for (int i = nCoeffs - 2; i >= 0; i--)
      retVal = coeffs[i] + x * retVal;
   retVal = retVal * std::pow(x, lowestOrder);
   return retVal + (pdfMode && lowestOrder > 0 ? 1.0 : 0.0);
}

double roo_func_wrapper_4(double *params)
{
   double t4[] = {params[0], params[1], 1.};
   const double t5 = polynomial<false>(t4, 3, 0, 1.);
   return t5;
}

#pragma clad ON
void roo_func_wrapper_4_req()
{
   clad::gradient(roo_func_wrapper_4, "params");
}
#pragma clad OFF

void reproducer()
{
   std::vector<double> parametersVec = {-0.5, -0.5, 0.5};

   auto wrapper = [&](double *params) { return roo_func_wrapper_4(params); };

   std::cout << roo_func_wrapper_4(parametersVec.data()) << std::endl;
   roo_func_wrapper_4_grad(parametersVec.data(), gradientVec.data());

   std::cout << gradientVec[0] << std::endl;
   std::cout << gradientVec[1] << std::endl;
   std::cout << gradientVec[2] << std::endl;

   auto numDiff = [&](int i) {
      const double eps = 1e-6;
   };

}
```

Output:
```txt
0
Clad diff:
0
0
0
Num diff:
1
1
0
```

PetroZarytskyi:
This PR fixed a bug when a reverse block was never ended in ``RMV::VisitConditionalStatement``. This is the root cause of the problem mentioned in #922.
Fixes #922.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -945,20 +945,20 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
                          .get();
     // If result is a glvalue, we should keep it as it can potentially be
     // assigned as in (c ? a : b) = x;
+    Expr* ResultRef = nullptr;
     if ((CO->isModifiableLvalue(m_Context) == Expr::MLV_Valid) &&
         ifTrueExprDiff.getExpr_dx() && ifFalseExprDiff.getExpr_dx()) {
-      Expr* ResultRef = m_Sema
-                            .ActOnConditionalOp(noLoc, noLoc, condStored,
-                                                ifTrueExprDiff.getExpr_dx(),
-                                                ifFalseExprDiff.getExpr_dx())
-                            .get();
+      ResultRef = m_Sema
+                      .ActOnConditionalOp(noLoc, noLoc, condStored,
+                                          ifTrueExprDiff.getExpr_dx(),
+                                          ifFalseExprDiff.getExpr_dx())
+                      .get();
       if (ResultRef->isModifiableLvalue(m_Context) != Expr::MLV_Valid)
         ResultRef = nullptr;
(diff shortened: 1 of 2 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -945,20 +945,20 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
                          .get();
     // If result is a glvalue, we should keep it as it can potentially be
     // assigned as in (c ? a : b) = x;
```
vaithak, line 947: Can we add the test mentioned in the issue?

vaithak (approved):
LGTM.
I think we can squash the commits together, and maybe have a better PR message.
Maybe "Ensure proper ending of the reverse block for conditional statements."
Please ensure that the example that Jonas posted is working.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: Done.

PetroZarytskyi:
@vaithak I just ran the example and it works:
```
Clad diff:
1
1
0
Num diff:
1
1
0
```

PR #929, opened 2024-06-07
Revert removing clad::array_ref from InterfaceCompatibility.C

PetroZarytskyi:
``clad::array_ref`` should not have been removed from InterfaceCompatibility.C in #925. The whole point of InterfaceCompatibility.C is to check the compatibility of Clad's interfaces with ``clad::array_ref``.


vgvassilev (approved):
Good catch!  LGTM!

PR #930, opened 2024-06-10
Fix an assertion when building the clad benchmarks with clang18 on osx.

vgvassilev:
Unfortunately, it'd be difficult to come up with a test...

### DIFF
--- a/lib/Differentiator/VisitorBase.cpp
+++ b/lib/Differentiator/VisitorBase.cpp
@@ -280,10 +280,10 @@ namespace clad {
         (isa<CXXOperatorCallExpr>(ENoCasts) &&
          cast<CXXOperatorCallExpr>(ENoCasts)->getNumArgs() == 2) ||
         isa<ConditionalOperator>(ENoCasts) ||
-        isa<CXXBindTemporaryExpr>(ENoCasts))
-      return m_Sema.ActOnParenExpr(noLoc, noLoc, E).get();
-    else
-      return E;
+        isa<CXXBindTemporaryExpr>(ENoCasts)) {
+      return m_Sema.ActOnParenExpr(E->getBeginLoc(), E->getEndLoc(), E).get();
+    }
+    return E;
   }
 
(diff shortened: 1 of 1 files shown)


vaithak (approved):
Looks good.
But, given that we will improve the error or diagnostic messages in the future, maybe we can use `E->getBeginLoc` and `E->getEndLoc` instead of creating fake ones.

vgvassilev:
> Looks good. But, given that we will improve the error or diagnostic messages in the future, maybe we can use `E->getBeginLoc` and `E->getEndLoc` instead of creating fake ones.

Until now we explicitly tried to use the fakeLoc approach to consistently annotate where we need to handle source locations in a non-crashing way. We could but that would point to the original function. Is that what you want here?

vaithak:
> Until now we explicitly tried to use the fakeLoc approach to consistently annotate where we need to handle source locations in a non-crashing way. We could but that would point to the original function. Is that what you want here?

Although it would point to the original function. It will be really helpful for creating minimal reproducers from big functions. As, these location informations in the primal function will point to the line or location of the troublesome expression.

vgvassilev:
> > Until now we explicitly tried to use the fakeLoc approach to consistently annotate where we need to handle source locations in a non-crashing way. We could but that would point to the original function. Is that what you want here?
> 
> Although it would point to the original function, it will be really helpful for creating minimal reproducers from big functions, as these location informations in the primal function will point to the line or location of the troublesome expression.

Ok. Let me change that. However, this comment is questioning our currently strategy here, right? If you agree we should open an issue where we suggest replacing all of the fake locations with ones pointing to the original function.

vaithak:
> Ok. Let me change that. However, this comment is questioning our currently strategy here, right? If you agree we should open an issue where we suggest replacing all of the fake locations with ones pointing to the original function.

I am in full support for the new strategy of pointing to original function locations, so that it provides ease of debugging.

vgvassilev:
I do not think the codecov complaint is relevant for this PR. Let's move forward.

Issue #870, opened 2024-04-26
Custom derivatives are not detected and used correctly when namespaces are present.


alexander-penev:
When implementing custom derivatives for one of the still unsupported std::erf built-ins, I noticed that custom derivatives were not being found and used correctly.

For example:

``` C++
namespace clad { namespace custom_derivatives {
namespace std {
  template <typename T>
  ValueAndPushforward<T, T> erf_pushforward(T x, T d_x) {
    return {::std::erf(x), M_2_SQRTPI*::std::exp(-::std::pow(x,2.0)) * d_x};
  }
}
using std::erf_pushforward; // Required if we use C style function call erf(...) instead C++ std::erf(...)
}}
```

``` C++
double f(double x) {
  return std::erf(x) + 1;
}

int main() {
  auto df = clad::differentiate(f, 0);
  df.dump();
  std::cout << "f(3)=" << f(3.0) << "; df(3)=" << df.execute(3.0) << std::endl;
  return 0;
}
```

Dimps:

```
The code is:
double f_darg0(double x) {
    double _d_x = 1;
    ValueAndPushforward<double, double> _t0 = clad::custom_derivatives::erf_pushforward(x, _d_x);
    return _t0.pushforward + 0;
}
f(3)=1.99998; df(3)=0.000139253
```

but we expect this:

``` C++
... = clad::custom_derivatives::std::erf_pushforward(x, _d_x);
```

In contrast, in this case:

``` C++
double f(double x) {
  return erf(x) + 1;
}
```

we must generate:

``` C++
... = clad::custom_derivatives::erf_pushforward(x, _d_x);
```

The problem is not only in the print/dump. This is easily seen if, instead of `using`, we define non-std pushforward function, which returns, for example, a different result.

vgvassilev:
This issue was addressed in #1171. The test case with the pull requests looks very similar to the one provided in this issue:
https://github.com/vgvassilev/clad/blob/1fcd683bcdbcd5dfdb3a81ada02c083939d2d260/test/FirstDerivative/BuiltinDerivatives.C#L11-L26

@alexander-penev, can you check if it fixes the example for you and if not reopen.

PR #931, opened 2024-06-10
Improve finding of higher order custom derivatives

vaithak:
Before this change, even if the user had provided a custom derivative of a custom derivative function, Clad couldn't find it.


vgvassilev:
Is that trying to solve maybe even partially #870?

vaithak:
> Is that trying to solve maybe even partially #870?

Yes. In a follow up PRs, that will be solved.

Issue #721, opened 2024-01-21
Refactor duplicate code in ::Derive functions in the visitors


vgvassilev:
BaseForwardModeVisitor, ReverseModeForwPassVisitor, HessianModeVisitor and JacobianModeVisitor have practically a common method `Derive` which duplicates a lot of code. We should move the common code into a utility class that's shared between the visitors.

vgvassilev:
This became a lot simplified now. There is a need for more refactoring but at that stage we can mark this issue as complete as the current state captures well the original intent of the issue...

PR #933, opened 2024-06-11
Pass the DiffRequest down to the visitors. NFC.

vgvassilev:
The intent of this patch is to centralize the information about the diff request in one place. This will help upcoming patches to reduce the state of the visitors and refactor some of the implementation in a common place.

That will unblock work on #721.

### DIFF
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -33,8 +33,9 @@
 using namespace clang;
 
 namespace clad {
-BaseForwardModeVisitor::BaseForwardModeVisitor(DerivativeBuilder& builder)
-    : VisitorBase(builder) {}
+BaseForwardModeVisitor::BaseForwardModeVisitor(DerivativeBuilder& builder,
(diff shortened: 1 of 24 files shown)


lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -60,8 +61,8 @@ bool IsRealNonReferenceType(QualType T) {
 DerivativeAndOverload
 BaseForwardModeVisitor::Derive(const FunctionDecl* FD,
                                const DiffRequest& request) {
```
parth-07, line 63: Should we continue to pass `DiffRequest &request` to `Derive` considering that now we instantiate the visitor with the `DiffRequest`? Ideally, now we can remove both the `Derive` parameters, right?

vgvassilev: No, that is my intent. In several subsequent pull requests that will become much cleaner. I cannot do it in one go because it would make bisection quite hard if something breaks for external clients...

parth-07 (approved):
This looks good to me.

PR #934, opened 2024-06-11
Add support for custom derivatives for top level derivatives

### LINKED ISSUE #352: Custom derivatives are not considered for direct differentiation request.
Reproducible example:
```c++
double identity(double i) {
  return i;
}
  void identity_grad(double i, clad::array_ref<double> d_i) {
    d_i += 1;
  }
}
auto d_identity = clad::gradient(identity);
```
**Result:** Clad generates the derivative of `double identity(double)` instead of reusing the custom derivative provided by the user.

**The root cause of the issue**: We only consider custom derivatives when we are differentiating function calls. This is done by finding if a custom derivative is available for the function being called in `VisitCallExpr`. We should instead check for custom derivative availability when processing the differentiation request in `CladPlugin::ProcessDiffRequest`.

vaithak:
fixes #352


vaithak:
This doesn't add support with vector mode, I am trying to debug that and will add a follow up PR for that.

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -59,19 +59,27 @@ namespace clad {
...
     if (!firstDerivative) {
+      alreadyDerived = false;
       // Derive declaration of the the forward mode derivative.
       IndependentArgRequest.DeclarationOnly = true;
       firstDerivative = plugin::ProcessDiffRequest(CP, IndependentArgRequest);
 
+      // It is possible that user has provided a custom derivative for the
+      // derivative function. In that case, we should not derive the definition
+      // again.
+      if (firstDerivative->getDefinition())
```
vgvassilev, line 74: Could the provided function be a declaration whose definition in in a library provided by the user?

vaithak: Yup, that would be ok. We already have a unittest for that here: https://github.com/vgvassilev/clad/blob/master/unittests/Misc/CallDeclOnly.cpp

vgvassilev: Nice!

vaithak: Let me test this once. I suspect that this might fail for hessians for a particular case.

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -133,6 +133,32 @@ namespace clad {
...
+    ///
+    /// \param[in] Name The name of the function to look up.
+    /// \param[in] originalFnDC The original function's DeclContext.
+    /// \param[in] SS The CXXScopeSpec to extend with the namespace of the
+    /// function.
+    /// \param[in] forCustomDerv A flag to keep track of which
+    /// namespace we should look in for the overloads.
+    /// \param[in] namespaceShouldExist A flag to enforce assertion failure
+    /// if the overload function namespace was not found. If false and
+    /// the function containing namespace was not found,
```
vgvassilev, line 147: Seems we lost a line or two here?

test/Gradient/FunctionCalls.C
```diff
@@ -778,7 +823,7 @@ int main() {
   TEST_ARR5(fn4, arr, 5);       // CHECK-EXEC: {23.00, 3.00, 3.00, 3.00, 3.00}
   TEST_ARR5(fn5, arr, 5);       // CHECK-EXEC: {5.00, 1.00, 0.00, 0.00, 0.00}
   TEST2(fn6, 3, 5);             // CHECK-EXEC: {5.00, 3.00}
-  TEST2(fn7, 3, 5);             // CHECK-EXEC: {10.00, 71.00}
+  TEST2(fn7, 3, 5);             // CHECK-EXEC: {11.00, 78.00}
```
vgvassilev, line 826: Are we sure this result is correct?

vaithak: Yup, because `temp` holds the address of `i`, similar to `k`. So, when `k` is updated - `temp` is updated too.

lib/Differentiator/HessianModeVisitor.cpp
vaithak: I added a new test for this. There was an issue with this, but it is solved by separately keeping track of clad-generated derivatives and custom derivatives, added a new commit for this.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -402,7 +467,8 @@ static void registerDerivative(FunctionDecl* derivedFD, Sema& semaRef) {
     return nullptr;
   }
 
-  void DerivativeBuilder::AddEdgeToGraph(const DiffRequest& request) {
-    m_DiffRequestGraph.addEdgeToCurrentNode(request);
+  void DerivativeBuilder::AddEdgeToGraph(const DiffRequest& request,
```
vgvassilev, line 470: Can't we move all the `alreadyDerived` logic here, rather than having it scattered around?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -160,16 +160,20 @@ BaseForwardModeVisitor::Derive(const FunctionDecl* FD,
   for (auto field : diffVarInfo.fields)
     argInfo += "_" + field;
 
-  IdentifierInfo* II = &m_Context.Idents.get(
-      request.BaseFunctionName + "_d" + s + "arg" + argInfo + derivativeSuffix);
+  // Check if the function is already declared as a custom derivative.
+  std::string gradientName =
+      request.BaseFunctionName + "_d" + s + "arg" + argInfo + derivativeSuffix;
+  auto* DC = const_cast<DeclContext*>(m_DiffReq->getDeclContext());
```
vgvassilev, line 166:
```suggestion
  // FIXME: We should not use const_cast to get the decl context here.
  // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
  auto* DC = const_cast<DeclContext*>(m_DiffReq->getDeclContext());
```

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -134,6 +112,26 @@ static const StringLiteral* CreateStringLiteral(ASTContext& C,
...
+    const auto* originalFnProtoType =
+        cast<FunctionProtoType>(m_DiffReq->getType());
+    QualType hessianFunctionType = m_Context.getFunctionType(
+        m_Context.VoidTy,
+        llvm::ArrayRef<QualType>(paramTypes.data(), paramTypes.size()),
+        // Cast to function pointer.
+        originalFnProtoType->getExtProtoInfo());
+
+    // Check if the function is already declared as a custom derivative.
+    auto* DC = const_cast<DeclContext*>(m_DiffReq->getDeclContext());
```
vgvassilev, line 130:
```suggestion
  // FIXME: We should not use const_cast to get the decl context here.
  // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
  auto* DC = const_cast<DeclContext*>(m_DiffReq->getDeclContext());
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -361,13 +361,22 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         // Cast to function pointer.
         originalFnType->getExtProtoInfo());
 
+    // Check if the function is already declared as a custom derivative.
+    auto* DC = const_cast<DeclContext*>(m_DiffReq->getDeclContext());
```
vgvassilev, line 365:
```suggestion
    // FIXME: We should not use const_cast to get the decl context here.
    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
    auto* DC = const_cast<DeclContext*>(m_DiffReq->getDeclContext());
```

lib/Differentiator/DerivativeBuilder.cpp
vaithak: The last commit in this PR related to refactoring does that. 
All that special logic is now centralized inside `DerivativeBuilder::HandleNestedDiffRequest`, all the visitors just call that function.

vgvassilev: Nice!

vaithak:
Added vector mode support in a new commit, it works 👍🏼

vgvassilev (approved):
LGTM! Is this ready to be merged as it looks like you are pushing more code to it?

vaithak:
> LGTM! Is this ready to be merged as it looks like you are pushing more code to it?

Ready to go. The latest commit was just for vector mode, it works in the same way as all other visitors. I earlier thought it would take more effort.
Good to be merged 👍🏼

PR #935, opened 2024-06-11
Add support for initializer_list in forward mode AD

parth-07:
This commit adds primitive support for initializer_list in the forward mode AD.


vgvassilev (approved):
LGTM!

PR #936, opened 2024-06-11
Emit clad::pop after clad::back when in RMV::VisitBinaryOperator

### LINKED ISSUE #927: Regression in Clad when dealing with loops
I tried to create a minimal reproducer from Roofit's failing test cases:
```cpp

inline double gaussian(double x, double mean, double sigma) {
   const double arg = x - mean;
   const double sig = sigma;
   return std::exp(-0.5 * arg * arg / (sig * sig));
}

double wrapper(double* params) {
  double RooNLLVarNewResult = 0.0;
  double t24[] = {params[0],params[1],1.0};
  for(int loopIdx0 = 0; loopIdx0 < 60; loopIdx0++) {
    const double t10 = gaussian(2.0, 5.000000, params[2]);
    const double t11 = t10/3;
    const double t14 = gaussian(2.0, 2.000000, params[3]);
    const double t15 = t14/3;
    const double t18 = gaussian(2.0, 7.000000, params[4]);
    const double t19 = t18/3;
    const double t22 = gaussian(2.0, 6.000000, params[5]);
    const double t23 = t22/3;
    double t7[] = {t11,t15,t19,t23};
    double t27 = 0, t28= 0;
    for(int i_t29 = 0; i_t29 < 3; i_t29++) {
      t27 += t7[i_t29] * t24[i_t29];
      t28 += t24[i_t29];
    }
    t27 += t7[3] * (1 - t28);
    RooNLLVarNewResult += t27;
  }
  return RooNLLVarNewResult;
}

int main() {
    0.25, 0.25, 0.5, 1, 1.5, 2
  };
  double d_params[] = {0.0, 0.0, 0.0, 0.0, 0.0, 0.0};
  auto f_dx = clad::gradient(wrapper);
  f_dx.execute(parametersVec.data(), d_params);
  std::cout << "autodiff: " << std::endl;
  std::cout << d_params[0] << std::endl;
  std::cout << d_params[1] << std::endl;
  std::cout << d_params[2] << std::endl;
  std::cout << d_params[3] << std::endl;
  std::cout << d_params[4] << std::endl;
  std::cout << d_params[5] << std::endl;

  auto numDiff = [&](int i) {
      const double eps = 1e-6;
   };

   std::cout << "num diff: " << std::endl;
   std::cout << numDiff(3) << std::endl;
   std::cout << numDiff(4) << std::endl;
   std::cout << numDiff(5) << std::endl;
  return 0;
```

The error message on Clang setup:
```txt
../include/clad/Differentiator/Tape.h:76: clad::tape_impl::reference clad::tape_impl<double>::back() [T = double]: Assertion `_size' failed.
[1]    2223726 IOT instruction (core dumped)  ./a.out
```

With Cling and Roofit setup, we don't see the error for some reason but still the output is wrong compared to numerical differentiation.
After doing a git-bisect, I found that this was introduced in #904.

PetroZarytskyi:
After #904, operands in multiplication and division inside loops were not correctly stored anymore. The pop expression was emitted before the value was used:
```
// forward sweep (inside a loop)
    ...
    clad::push(_t3, arr[0]);
    out += clad::back(_t3) * params[0];
    ...
        
// reverse sweep (inside a loop)
    ...
    clad::pop(_t3);    <- the value is popped before it's used
    _d_params[0] += clad::back(_t3) * _r_d0;
    ...
```
This is because in RMV::VisitBinaryOperator the order of statements in the corresponding reverse block is not reversed (while in all other use cases of GlobalStoreAndRef, the order of statements is reversed).
This PR uses a simple workaround to fix this bug.
Fixes #927

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2096,9 +2096,17 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       // df/dxr += df/dxi * dxi/xr = df/dxi * xl
       // Store left multiplier and assign it with L.
       StmtDiff LStored = Ldiff;
+      // Catch the pop statement and emit it after
+      // the LStored value is used.
+      // This workaround is necessary because GlobalStoreAndRef
+      // is designed to work with the reversed order of statements
+      // in the reverse sweep and in RMV::VisitBinaryOperator
+      // the order is not reversed.
+      beginBlock(direction::reverse);
       if (!ShouldRecompute(LStored.getExpr()))
         LStored = GlobalStoreAndRef(LStored.getExpr(), /*prefix=*/"_t",
                                     /*force=*/true);
+      Stmt* LPop = endBlock(direction::reverse);
       Expr::EvalResult dummy;
       if (RDelayed ||
           !clad_compat::Expr_EvaluateAsConstantExpr(R, dummy, m_Context)) {
@@ -2110,6 +2118,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         if (RDelayed)
           RDelayed->Finalize(Rdiff.getExpr());
       }
+      addToCurrentBlock(utils::unwrapIfSingleStmt(LPop), direction::reverse);
       std::tie(Ldiff, Rdiff) =
           std::make_pair(LStored.getExpr(), RResult.getExpr());
     } else if (opCode == BO_Div) {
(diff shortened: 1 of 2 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2248,20 +2266,8 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       // For x, AssignedDiff is _d_x, for x[i] its _d_x[i], for reference exprs
       // like (x = y) it propagates recursively, so _d_x is also returned.
       Expr* AssignedDiff = Ldiff.getExpr_dx();
-      if (!AssignedDiff) {
-        // If either LHS or RHS is a declaration reference, visit it to avoid
-        // naming collision
```
vaithak, line 2253: Is the case in this comment not valid anymore?

PetroZarytskyi: First, name collisions are now handled in ``Clone``. Second, I don't think ``LDRE ? Visit(L).getExpr() : L;`` makes sense at all. ``L`` is an expression from the original function, we cannot use it in the derivative.

vaithak (approved):
Looks good to me 👍🏼

PR #937, opened 2024-06-12
Add support for simple lambda expressions in forward mode

gojakuch:
This commit provides support for the simplest lambda expressions, that is, those with no captures, in forward mode. The original lambda function is copied into the derivative, but the corresponding lambda class gets extended to also have a pushforward method for the call operator overload. Essentially, there's no visiting of lambda expressions yet (this will be implemented later to support captures, I'm working on this), but rather we allow treating simple cases of lambda functions as functors.

### DIFF
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -13,6 +13,7 @@
 #include "clad/Differentiator/ErrorEstimator.h"
 
 #include "clang/AST/ASTContext.h"
+#include "clang/AST/ASTLambda.h"
 #include "clang/AST/Expr.h"
 #include "clang/AST/TemplateBase.h"
 #include "clang/Sema/Lookup.h"
@@ -1039,6 +1040,9 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
 
   SourceLocation validLoc{CE->getBeginLoc()};
 
+  // Calls to lambda functions are processed differently
+  bool isLambda = isLambdaCallOperator(FD);
+
   // If the function is non_differentiable, return zero derivative.
   if (clad::utils::hasNonDifferentiableAttribute(CE)) {
     // Calling the function without computing derivatives
@@ -1070,12 +1074,24 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
   // a direct or indirect (operator overload) call to member function.
   StmtDiff baseDiff;
   // Add derivative of the implicit `this` pointer to the `diffArgs`.
-  if (auto MD = dyn_cast<CXXMethodDecl>(FD)) {
+  if (isLambda) {
+    if (const auto* OCE = dyn_cast<CXXOperatorCallExpr>(CE)) {
+      QualType ptrType = m_Context.getPointerType(m_Context.getRecordType(
+          FD->getDeclContext()->getOuterLexicalRecordContext()));
+      // For now, only lambdas with no captures are supported, so we just pass
+      // a nullptr instead of the diff object.
+      baseDiff =
+          StmtDiff(Clone(OCE->getArg(0)),
+                   new (m_Context) CXXNullPtrLiteralExpr(ptrType, validLoc));
+      diffArgs.push_back(baseDiff.getExpr_dx());
+    }
+  } else if (const auto* MD =
+                 dyn_cast<CXXMethodDecl>(FD)) { // isLambda == false
     if (MD->isInstance()) {
       const Expr* baseOriginalE = nullptr;
-      if (auto MCE = dyn_cast<CXXMemberCallExpr>(CE))
+      if (const auto* MCE = dyn_cast<CXXMemberCallExpr>(CE))
         baseOriginalE = MCE->getImplicitObjectArgument();
-      else if (auto OCE = dyn_cast<CXXOperatorCallExpr>(CE))
+      else if (const auto* OCE = dyn_cast<CXXOperatorCallExpr>(CE))
         baseOriginalE = OCE->getArg(0);
       baseDiff = Visit(baseOriginalE);
       Expr* baseDerivative = baseDiff.getExpr_dx();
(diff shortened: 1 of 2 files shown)


lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1064,18 +1068,32 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
...
-        baseOriginalE = MCE->getImplicitObjectArgument();
-      else if (auto OCE = dyn_cast<CXXOperatorCallExpr>(CE))
-        baseOriginalE = OCE->getArg(0);
-      baseDiff = Visit(baseOriginalE);
-      Expr* baseDerivative = baseDiff.getExpr_dx();
-      if (!baseDerivative->getType()->isPointerType())
-        baseDerivative = BuildOp(UnaryOperatorKind::UO_AddrOf, baseDerivative);
-      diffArgs.push_back(baseDerivative);
+  if (const auto* MD = dyn_cast<CXXMethodDecl>(FD)) {
+    if (!isLambda) {
```
vgvassilev, line 1072: I'd invert the condition to reduce the negations. Also, we might just use `clang::isLambdaCallOperator` for this check in the if-stmt.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1499,7 +1519,11 @@ StmtDiff BaseForwardModeVisitor::VisitDeclStmt(const DeclStmt* DS) {
     if (QT->isPointerType())
       QT = QT->getPointeeType();
     auto* typeDecl = QT->getAsCXXRecordDecl();
-    if (typeDecl && clad::utils::hasNonDifferentiableAttribute(typeDecl)) {
+    // For lambda functions, we don't need another declaration for the
+    // derivative, as there's a pushforward method generated for each lambda
+    // type, so they are treated as non-differentiable.
+    if (typeDecl && (clad::utils::hasNonDifferentiableAttribute(typeDecl) ||
+                     typeDecl->isLambda())) {
```
vgvassilev, line 1526: Can we have a test case with a custom pushforward covering that comment?

gojakuch: if I understood you correctly, I don't think it's possible, since the generated pushforward method is put directly into the class describing the lambda function, which is not accessible to modifications when writing user input code. so it's not possible to put a user-defined pushforward in that context I suppose

lib/Differentiator/BaseForwardModeVisitor.cpp
gojakuch: error: ‘isLambdaCallOperator’ was not declared in this scope.

I couldn't manage to use this check unfortunately. but I inverted the condition

lib/Differentiator/BaseForwardModeVisitor.cpp
vgvassilev: I am not sure I understand that comment.

gojakuch: so, I basically don't a user can provide a custom pushforward for a lambda. or did you mean just having a test case that tests a generated pushforward?

vgvassilev: I mean that I cannot understand the comment in the code before the if-stmt.

gojakuch: well I wanted to say that we don't need to create a new declaration for the derivative of the lambda, as the only thing we're doing for now is copying the original lambda and then processing the call to it to create the pushforward as a separate method of the lambda class later. I probably should've formulated it better.

gojakuch: I've updated the comment, please take a look.

lib/Differentiator/BaseForwardModeVisitor.cpp
vgvassilev: Did you include `clang/AST/ASTLambda.h` before?

gojakuch: ok, for some reason I was sure that it's included. it finds the method now but if I set `isLambda` variable to `isLambdaCallOperator(FD->getDeclContext())` on the line 1035 (which is, I assume, is how it's supposed to work) it's just not equivalent to the current check so it actually crashes:
```c++
// current check, equals to 1 in lambda tests
isLambda = (FD->getDeclContext()->isRecord() && FD->getDeclContext()->getOuterLexicalRecordContext()->isLambda());
// new check, equals to 0 in lambda tests
isLambda = isLambdaCallOperator(FD->getDeclContext());
```
I'm might be using `isLambdaCallOperator` wrong. 

btw I suggest that `isLambda` is kept as a variable if you meant that we should call `isLambdaCallOperator` in the if-condition directly because it's used twice already and might be used more as I implement this further.

vgvassilev: My suggestion to use a "standard" check was to make sure we do what the compiler does to check if something is a lambda. Why does it fail?

vgvassilev: `isLambdaCallOperator` takes a `CXXMethodDecl`. You are taking the parent DeclContext which is the class...

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1036,9 +1037,12 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
          "Differentiation of only direct calls is supported. Ignored");
     return StmtDiff(Clone(CE));
   }
-
+  const auto* MD = dyn_cast<CXXMethodDecl>(FD);
   SourceLocation validLoc{CE->getBeginLoc()};
 
+  // Calls to lambda functions are processed differently
+  bool isLambda = MD && isLambdaCallOperator(MD);
```
vgvassilev, line 1044: This should be able to take FD as is. Can you pass that instead without casting it?

vgvassilev, line 1044:
```suggestion
  bool isLambda = isLambdaCallOperator(FD);
```

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1036,9 +1037,12 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
          "Differentiation of only direct calls is supported. Ignored");
     return StmtDiff(Clone(CE));
   }
-
+  const auto* MD = dyn_cast<CXXMethodDecl>(FD);
```
vgvassilev, line 1040:
```suggestion
```

vgvassilev (approved):
LGTM!

PR #938, opened 2024-06-13
Remove useless goto/label statements

### LINKED ISSUE #526: Optimize away redundant goto statements

PetroZarytskyi:
This PR removes all pairs of goto/label statements with nothing in between:
```
goto _label0;
_label0:
```
Obviously, this bit of code doesn't do anything. Such cases make up the majority of goto uses in Clad. 
The idea in this PR is to check whether the corresponding return statement is the last statement in the function's body. If it is, there's no need to create a goto statement.
Before this PR, Clad tests contained 328 goto stmts, now they only contain 21.

Fixes #526.

### DIFF
--- a/demos/ErrorEstimation/CustomModel/README.md
+++ b/demos/ErrorEstimation/CustomModel/README.md
@@ -48,8 +48,6 @@ The code is: void func_grad(float x, float y, float *_d_x, float *_d_y, double &
     float z;
     _t0 = z;
     z = x + y;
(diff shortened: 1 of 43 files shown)


vgvassilev (approved):
LGTM!

Issue #913, opened 2024-05-29
Incorrect derivative for while-loops with declarations and assignments in forward mode


gojakuch:
In the following example, `fn` is basically a zero function. Yet, Clad computes its derivative to be 1.

```C++
double fn(double x) {
    while (double t = (x = 0)) {}
    return x;
} // = 0

int main() {
    auto f_dx = clad::differentiate(fn, 0);

    std::cout << f_dx.execute(5) << '\n';
}
```

**Expected output:** 0

**Output:** 1

Same thing for 

```C++
double fn(double x) {
    while ((x = 0)) {}
    return x;
} // = 0

int main() {
    auto f_dx = clad::differentiate(fn, 0);

    std::cout << f_dx.execute(5) << '\n';
}
```
**Expected output:** 0

**Output:** 1

This could be solved similarly to how #905 tries to solve the same issue for for-loops, that is with comma operators with a possibility to support more complicated conditions like `(y = x*x) || (y = y - 1)`.

PR #941, opened 2024-06-14
While stmt fix

ovdiiuv:
This PR adds support for assignments in while-loops. It also enables to combine multiple assignments in the while condition by adding supprot for some logoical operators.

Fixes: #913

### DIFF
--- a/test/Gradient/FunctionCalls.C
+++ b/test/Gradient/FunctionCalls.C
@@ -962,8 +962,6 @@ double sq_defined_later(double x) {
 // CHECK-NEXT: }
 
 // CHECK: void custom_identity_pullback(double &i, double _d_y, double *_d_i) {
-// CHECK-NEXT:     goto _label0;
-// CHECK-NEXT:   _label0:
 // CHECK-NEXT:     *_d_i += _d_y; 
 // CHECK-NEXT: }
 
--- a/test/Hessian/BuiltinDerivatives.C
+++ b/test/Hessian/BuiltinDerivatives.C
@@ -363,8 +363,6 @@ int main() {
 // CHECK-NEXT: }
 
 // CHECK: void f7_darg0_grad(float x, float y, float *_d_x, float *_d_y) {
-// CHECK-NEXT:     goto _label0;
-// CHECK-NEXT:   _label0:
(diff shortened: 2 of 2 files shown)


lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1633,21 +1633,45 @@ StmtDiff BaseForwardModeVisitor::VisitWhileStmt(const WhileStmt* WS) {
...
+
+    if ((condBO && (condBO->isLogicalOp() || condBO->isAssignmentOp())) ||
+        condUO) {
+      auto condDiff = Visit(cond);
+      if (condDiff.getExpr_dx() &&
+          (!isUnusedResult(condDiff.getExpr_dx()) || condUO))
+        cond = BuildOp(BO_Comma, BuildParens(condDiff.getExpr_dx()),
+                       BuildParens(condDiff.getExpr()));
+      else
+        cond = condDiff.getExpr();
```
vgvassilev, line 1667: Can we add a test case covering that branch? Can you also add a comment before the complex if-stmt condition which are the cases we enter with there?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1662,29 +1686,6 @@ StmtDiff BaseForwardModeVisitor::VisitWhileStmt(const WhileStmt* WS) {
...
-  //   ...
-  // }
-  //
-  // gets differentiated to,
-  //
-  // while (double b = a) {
-  //   double _d_b = _d_a;
-  //   ...
-  //   ...
-  // }
```
vgvassilev, line 1682: This comment was rather good. Why did we drop it?

ovdiiuv: Now what it does to the `while (double b = a){}` is it creates variables `b` and `_d_b` before the while statement and assigns them the right values inside of the while condition, so the comment is no longer relevant.

vgvassilev: Why not updating the comment rather than deleting it?

PR #942, opened 2024-06-15
Fix test failure on master due to rebasing

vaithak:
This was caused because I didn't rebase #934, sorry for that.


vgvassilev (approved):
Ah, good catch.

PR #944, opened 2024-06-17
Remove redundant data member in favor of centralizing in DiffRequest. NFC

vgvassilev:
Partially addresses #721.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -258,12 +258,15 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     silenceDiags = !request.VerboseDiags;
     assert(m_DiffReq == request);
 
-    // reverse mode plugins may have request mode other than
+    // FIXME: reverse mode plugins may have request mode other than
     // `DiffMode::reverse`, but they still need the `DiffMode::reverse` mode
     // specific behaviour, because they are "reverse" mode plugins.
-    m_Mode = DiffMode::reverse;
-    if (request.Mode == DiffMode::jacobian)
-      m_Mode = DiffMode::jacobian;
+    // assert(m_DiffReq.Mode == DiffMode::reverse ||
+    //        m_DiffReq.Mode == DiffMode::jacobian && "Unexpected Mode.");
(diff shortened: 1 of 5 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -258,12 +258,15 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
     assert(m_DiffReq == request);
 
-    // reverse mode plugins may have request mode other than
+    // FIXME: reverse mode plugins may have request mode other than
     // `DiffMode::reverse`, but they still need the `DiffMode::reverse` mode
     // specific behaviour, because they are "reverse" mode plugins.
-    m_Mode = DiffMode::reverse;
-    if (request.Mode == DiffMode::jacobian)
-      m_Mode = DiffMode::jacobian;
+    // assert(m_DiffReq.Mode == DiffMode::reverse ||
```
vaithak, line 264: Do we need these comments?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -258,12 +258,15 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    //        m_DiffReq.Mode == DiffMode::jacobian && "Unexpected Mode.");
+    if (m_DiffReq.Mode == DiffMode::error_estimation)
+      // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
+      const_cast<DiffRequest&>(m_DiffReq).Mode = DiffMode::reverse;
```
vaithak, line 268: I am not sure error estimation works, so can't judge if this is necessary.

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: Yes, that's a reminder to enable this once we address the fact that we should not change the mode of the DiffRequest.

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: Unfortunately without that line the error estimation test suite starts to fail.

PR #947, opened 2024-06-17
Remove the enableTBR state from reverse mode visitor. NFC

vgvassilev:
Partially addresses #721.


PetroZarytskyi:
We don't need the ``enableTBR`` field in the RMV at all, right? I suggest we remove it from ReverseModeVisitor.h as well. Other than that, the PR looks good to me.

vgvassilev:
> We don't need the `enableTBR` field in the RMV at all, right? I suggest we remove it from ReverseModeVisitor.h as well. Other than that, the PR looks good to me.

Good point, I screwed up with the rebase. Now it should be better.

PetroZarytskyi:
LGTM!

PetroZarytskyi (approved):

PR #950, opened 2024-06-18
Add support for computing only the diagonal hessian entries

### LINKED ISSUE #509: Compute the hessian matrix diagonal only

vaithak:
fixes #509

### DIFF
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -364,8 +364,9 @@ BaseForwardModeVisitor::Derive(const FunctionDecl* FD,
 
     endScope(); // Function body scope
 
-    if (request.DerivedFDPrototype)
-      m_Derivative->setPreviousDeclaration(request.DerivedFDPrototype);
+    if (request.DerivedFDPrototypes.size() >= request.CurrentDerivativeOrder)
+      m_Derivative->setPreviousDeclaration(
+          request.DerivedFDPrototypes[request.CurrentDerivativeOrder - 1]);
   }
   m_Sema.PopFunctionScopeInfo();
   m_Sema.PopDeclContext();
@@ -529,8 +530,9 @@ BaseForwardModeVisitor::DerivePushforward(const FunctionDecl* FD,
 
     endScope(); // Function body scope
 
-    if (request.DerivedFDPrototype)
-      m_Derivative->setPreviousDeclaration(request.DerivedFDPrototype);
+    if (request.DerivedFDPrototypes.size() >= request.CurrentDerivativeOrder)
+      m_Derivative->setPreviousDeclaration(
+          request.DerivedFDPrototypes[request.CurrentDerivativeOrder - 1]);
   }
 
   m_Sema.PopFunctionScopeInfo();
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -364,6 +364,8 @@ BaseForwardModeVisitor::Derive(const FunctionDecl* FD,
 
     endScope(); // Function body scope
 
+    // Size >= current derivative order means that there exists a declaration
+    // or prototype for the currently derived function.
     if (request.DerivedFDPrototypes.size() >= request.CurrentDerivativeOrder)
       m_Derivative->setPreviousDeclaration(
           request.DerivedFDPrototypes[request.CurrentDerivativeOrder - 1]);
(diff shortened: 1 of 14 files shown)


tools/ClangPlugin.cpp
```diff
@@ -137,7 +137,7 @@ namespace clad {
         FinalizeTranslationUnit();
     }
 
-    FunctionDecl* CladPlugin::ProcessDiffRequest(DiffRequest& request) {
+    FunctionDecl* CladPlugin::ProcessDiffRequest(DiffRequest request) {
```
vgvassilev, line 140: Why do we copy here?

vaithak: For higher-order derivatives, it changes the request object, which results in the changed object being inserted in the differentiation plan graph. Thus causing issues in order of differentiation.

vaithak: This is the place where the request object is modified: https://github.com/vgvassilev/clad/blob/master/tools/ClangPlugin.cpp#L295

vgvassilev: Can we think of a way how to avoid this? Probably not without a lot of restructuring but maybe at least add a comment about the reasoning why we copy.

vaithak: Removed this change, we can get away with this for now.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -364,8 +364,9 @@ BaseForwardModeVisitor::Derive(const FunctionDecl* FD,
 
     endScope(); // Function body scope
 
-    if (request.DerivedFDPrototype)
-      m_Derivative->setPreviousDeclaration(request.DerivedFDPrototype);
+    if (request.DerivedFDPrototypes.size() >= request.CurrentDerivativeOrder)
```
vgvassilev, line 367: Can we add a comment here explaining why `DerivedFDPrototypes` needs to be `>=` `request.CurrentDerivativeOrder`?

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -295,6 +295,11 @@ static void registerDerivative(FunctionDecl* derivedFD, Sema& semaRef) {
     FunctionDecl* derivative = this->FindDerivedFunction(request);
     if (!derivative) {
       alreadyDerived = false;
+
+      // Store the function and its order before processing the nested request.
+      const FunctionDecl* origFn = request.Function;
+      unsigned origFnOrder = request.CurrentDerivativeOrder;
```
vgvassilev, line 301: Is it worth using `llvm::SaveAndRestore`?

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -423,7 +431,8 @@ static void registerDerivative(FunctionDecl* derivedFD, Sema& semaRef) {
     } else if (request.Mode == DiffMode::reverse_mode_forward_pass) {
       ReverseModeForwPassVisitor V(*this, request);
       result = V.Derive(FD, request);
-    } else if (request.Mode == DiffMode::hessian) {
+    } else if (request.Mode == DiffMode::hessian ||
```
vgvassilev, line 434: Not for this request but we need to somehow model that `hessian_diagonal` is a flavor of `hessian`. That will help the diagnostics bellow avoid relying on the attribute. We have a similar problem with `error_estimation` and `reverse`...

vgvassilev (commented):
I think this is heading in a good direction. I am a bit worried that we switched from running forward and reverse to running forward twice for the diagonal only. I suspect the reason is our plan to try vector mode but is there something else? Is there any point in making how we create the hessian configurable?

lib/Differentiator/BaseForwardModeVisitor.cpp
vaithak: Done 👍🏼

lib/Differentiator/DerivativeBuilder.cpp
vaithak: I don't think we need to restore the entire request object; we can reuse some modifications made to the new request object.

vgvassilev: I think `llvm::SaveAndRestore` works for object members, too.

vaithak: Updated 👍🏼

vgvassilev (approved):
LGTM! I will squash on merge.

PR #952, opened 2024-06-18
Add validation to prevent error on empty if block

rohanjulka19:


parth-07 (changes requested):
Please add a test case.

rohanjulka19:
Added test case. 
Actually this issue fixes this case 
```
 if (x > 0);
```

parth-07 (changes requested):
We should keep the if-block with `NullStmt` as then-statement instead of completely removing the if-block when there is `NullStmt` as then-statement. The below example fails if we completely remove the if-block when there is no then-statement:

```cpp
    double res = 0;
    if ((res = 0))
    ; else {
        res = 5 * j;
    }
    return res;
}
```

To resolve this issue, we need to add `VisitNullStmt` function to `ReverseModeVisitor`.

rohanjulka19:
Screenshot 2024-06-18 at 8 03 05 PM

Is this fine like this is working and I don't see any issue in returning null stmt for all the null stmts in general ? VisitNullStmt was already implemented in the header file i just added the clone stmts in StmtDiff

rohanjulka19:
That null stmt fix was giving an issue with the below test case 

```

double fn_null_stmts(double x) {
    ;;;;;;;;;;;;;;;;;
    ;;;;;return x;;;;
    ;;;;;;;;;;;;;;;;;
} // = x

```

So now in case the then block is empty  I just add a empty block in place of that  for both forward and reverse pass

vgvassilev:
> That null stmt fix was giving an issue with the below test case
> 
> ```
> 
> double fn_null_stmts(double x) {
>     ;;;;;;;;;;;;;;;;;
>     ;;;;;return x;;;;
>     ;;;;;;;;;;;;;;;;;
> } // = x
> ```
> 
> So now in case the then block is empty I just add a empty block in place of that for both forward and reverse pass

@gojakuch could probably comment on the intent of this test.

parth-07:
> Is this fine like this is working and I don't see any issue in returning null stmt for all the null stmts in general ? VisitNullStmt was already implemented in the header file i just added the clone stmts in StmtDiff

It does not seem like the pull-request solution is based on this idea. Why did you diverge from this?

rohanjulka19:
```
double fn_null_stmts(double x) {
    ;;;;;;;;;;;;;;;;;
    ;;;;;return x;;;;
    ;;;;;;;;;;;;;;;;;
} // = x
```
If we follow that idea for this test case it considers each semicolon as a statement so we get a bunch lines with semicolons

gojakuch:
> > That null stmt fix was giving an issue with the below test case
> > ```
> > 
> > double fn_null_stmts(double x) {
> >     ;;;;;;;;;;;;;;;;;
> >     ;;;;;return x;;;;
> >     ;;;;;;;;;;;;;;;;;
> > } // = x
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >      
> >   
> > So now in case the then block is empty I just add a empty block in place of that for both forward and reverse pass
> 
> @gojakuch could probably comment on the intent of this test.

right, when I implemented the NullStmt support it seemed an ugly idea to copy all the meaningless `;;` into the derivative code, so it doesn't. this test is there to test an arbitrary number of NullStmt repetitions that should be removed.

gojakuch:
> So now in case the then block is empty I just add a empty block in place of that for both forward and reverse pass

idk that seems better in my opinion because, in general, copying a bunch of nullstmts into the produced code where they are placed on a new line each does sound a bit awkward and makes the code less readable (which is what that test case tries to check). however, since the behaviour of the produced code doesn't change I don't think it's a big deal. but (if I understood this correctly) I think creating `if (cond) {}` instead of an `if (cond) ;` is kinda a fair trade-off for not creating 
```
;
;
```
out of `;;` or even more semicolons (a case that one could worsen infinitely by adding more and more `;` as opposed to the if statement that you only add an empty block to once). but I don't claim this to be the only right opinion and it depends on the priorities of Clad.

parth-07:
> If we follow that idea for this test case it considers each semicolon as a statement so we get a bunch of lines with semicolons.

I think that is fine. We are staying true to the primal code. I definitely prefer a general solution that reduces code branches over a restricted solution that require more checks.

> If you want I can just revert to older idea and push it, this test case can be ignored as it still works for this, just gives a very long result

Yes, I would prefer that.

gojakuch:
> I think that is fine. We are staying true to the primal code. I definitely prefer a general solution that reduces code branches over a restricted solution that requires more checks. Also, note that `if` condition is not the only place that can have `NullStmt` as the body. In a more restrictive solution, we will need more checks in other areas as well, such as loops.

@parth-07 ok I see your point. that makes sense as well. maybe let's just make a pair of `check-exec`s for this test case, so that it at least tests the unchanged behaviour of visiting a bunch of `NullStmt`s.

rohanjulka19:
can we just change the test case to have one or two semicolons it will yeild the same outcome anyways

gojakuch:
> can we just change the test case to have one or two semicolons it will yield the same outcome anyways

yeah, whatever works out for you then. just keep some sort of a meaningful check for this, in case someone touches the nullstmt again later.

vgvassilev:
> > If we follow that idea for this test case it considers each semicolon as a statement so we get a bunch of lines with semicolons.
> 
> I think that is fine. We are staying true to the primal code. I definitely prefer a general solution that reduces code branches over a restricted solution that requires more checks. Also, note that `if` condition is not the only place that can have `NullStmt` as the body. In a more restrictive solution, we will need more checks in other areas as well, such as loops.
> 
> > If you want I can just revert to older idea and push it, this test case can be ignored as it still works for this, just gives a very long result
> 
> Yes, I would prefer that. @vgvassilev What do you think about this?
 
I do not have a strong opinion. This test case is rather artificial and I do not expect to have such cases in real code. If we go with what @parth-07 suggest we can adjust that test to use a little less semicolons ;)

rohanjulka19:
ok since now we are handling null stmts we do not need to check for empty if blocks in the VisitIfStmt code so removed it.

parth-07 (approved):
Looks good to me.

parth-07:
Please update the commit message to reflect the final solution.

Issue #793, opened 2024-03-03
Do not relocate the tape elements


vgvassilev:
In the current implementation of `clad::tape` we ensure the elements are contiguous. The `grow` operation relocates the elements when the size exceeds the tape capacity:

https://github.com/vgvassilev/clad/blob/f242077c9b386a9711cdb257b1126a07b52c5006/include/clad/Differentiator/Tape.h#L119

Instead we should allocate slabs of 32 or 64 elements and the last element should point to the next slab.

vgvassilev:
@aditimjoshi, can you look into this issue?

PR #954, opened 2024-06-20
Use connected lists for tape data structure

rohanjulka19:
Currently I have added a basic implementation of tape data structure with connected lists/slabs. It is more or less similar to current tape data strucutre I copied a lot of code just added blocks and only added the functionality which I feel is needed for reverse mode. 

Right now NumericalDiff also uses tape so I defined two types one is "tape" which points to the new tape implementation and "old_tape" which points to current tape data strcuture being used. 

Tests are running fine for the new implementation. 

These are the benchmarks for old_tape and new_tape. 

[LastTest-new-tape.log](https://github.com/user-attachments/files/15916789/LastTest-new-tape.log)
[LastTest-tape.log](https://github.com/user-attachments/files/15916791/LastTest-tape-1.log)

Fixes - #793

### DIFF
--- a/include/clad/Differentiator/NewTape.h
+++ b/include/clad/Differentiator/NewTape.h
@@ -1,0 +1,119 @@
+#ifndef CLAD_DIFFERENTIATOR_NEWTAPE_H
+#define CLAD_DIFFERENTIATOR_NEWTAPE_H
+
+#include <cassert>
+#include <cstdio>
+#include <type_traits>
+#include <utility>
+
+#include "clad/Differentiator/CladConfig.h"
+
+namespace clad {
+
+static const int capacity = 32;
+
+template <typename T> class Block {
+public:
+  T data[capacity];
+  Block<T>* next;
+  Block<T>* prev;
+  using pointer = T*;
+  using iterator = pointer;
+
+  CUDA_HOST_DEVICE Block() {
+  }
+
+  CUDA_HOST_DEVICE ~Block() { destroy(block_begin(), block_end()); }
+
+  Block(const Block& other) = delete;
+  Block& operator=(const Block& other) = delete;
+
+  Block(Block&& other) = delete;
+  Block& operator=(const Block&& other) = delete;
+
+  CUDA_HOST_DEVICE iterator block_begin() { return data; }
+
+  CUDA_HOST_DEVICE iterator block_end() { return data + capacity; }
+
+  template <typename It> using value_type_of = decltype(*std::declval<It>());
+
+  template <typename It>
+  static typename std::enable_if<
+      !std::is_trivially_destructible<value_type_of<It>>::value>::type
+  destroy(It B, It E) {
+    for (It I = E - 1; I >= B; --I)
+      I->~value_type_of<It>();
+  }
+
+  template <typename It>
+  static typename std::enable_if<
+      std::is_trivially_destructible<value_type_of<It>>::value>::type
+      CUDA_HOST_DEVICE
+      destroy(It B, It E) {}
+};
+
+template <typename T> class new_tape_impl {
+  using NonConstT = typename std::remove_cv<T>::type;
+
+  Block<NonConstT>* m_cur_block = nullptr;
+  std::size_t m_size = 0;
+
+public:
+  new_tape_impl() = default;
+
+  ~new_tape_impl() { }
+
+  new_tape_impl(new_tape_impl& other) = delete;
+  new_tape_impl operator=(new_tape_impl& other) = delete;
+
+  new_tape_impl(new_tape_impl&& other) = delete;
+  new_tape_impl& operator=(new_tape_impl&& other) = delete;
+
+  template <typename... ArgsT>
+
+  CUDA_HOST_DEVICE void emplace_back(ArgsT&&... args) {
+    if (!m_cur_block || m_size >= capacity) {
+      Block<NonConstT>* prev_block = m_cur_block;
+      m_cur_block =  static_cast<Block<NonConstT>*>(::operator new(sizeof(Block<NonConstT>)));
+      if (prev_block != nullptr) {
+        prev_block->next = m_cur_block;
+        m_cur_block->prev = prev_block;
+      }
+      m_size = 0;
+    }
+    m_size += 1;
+    ::new (const_cast<void*>(static_cast<const volatile void*>(end())))
+        T(std::forward<ArgsT>(args)...);
+  }
+
+  [[nodiscard]] CUDA_HOST_DEVICE std::size_t size() const { return m_size; }
+
+  CUDA_HOST_DEVICE T* end() { return m_cur_block->data + (m_size - 1); }
+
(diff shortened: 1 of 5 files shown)


vgvassilev:
@rohanjulka19, do you mind addressing the clang-tidy complaints where relevant?

rohanjulka19:
hi, yeah will address these clang-tidy issues, actually i got stuck on the loop condition diffirentiation issue (once again) https://github.com/vgvassilev/clad/pull/818. But now I am fixing that one, once I push that fix will address these then.

rohanjulka19:
@vgvassilev I have fixed most of the suggestion, the above four I don't think needs to be addressed.

vgvassilev:
> @vgvassilev I have fixed most of the suggestion, the above four I don't think needs to be addressed.

The include protector suggestions seem good. Looks like the new tape has a problem with cuda.

rohanjulka19:
> 
> The include protector suggestions seem good. Looks like the new tape has a problem with cuda.

Have changed the include protector to align with the suggestion. The cuda test is passing but I am not sure if it will work when this runs on more than one thread in parallel

vgvassilev:
I am a little puzzled since the new tape implementation is supposed to be faster than the existing one but it seems consistently slower: https://github.com/vgvassilev/clad/actions/runs/9798455926/job/27056951548?pr=954#step:31:596

Do you have any clue what might be going wrong?

rohanjulka19:
I thought it was due to the small block size leading to more memory allocation but increasing the block size also didn't worked so will investigate this further

vgvassilev:
@rohanjulka19, do we have a clue what might be still wrong for the benchmarks? Maybe we should profile it with something like heaptrack, kcachegrind or similar...

rohanjulka19:
oh yeah I tried with timer but i couldn't figure out but I had a intution that because I am assinging memory twice once for the block and next for the array inside the block maybe that's whats causing the issue. So I am trying to just assign memory once by taking a fixed array of size 32 that should allocate all the memory once, but I am not very good with cpp so there is some issue with either memory allocation or something else. I will investigate the code this weekend, what do you think can this be the performance issue  ?

include/clad/Differentiator/NewTape.h
```diff
@@ -0,0 +1,122 @@
...
+
+  Block<NonConstT>* m_cur_block = nullptr;
+  std::size_t m_size = 0;
+
+public:
+  new_tape_impl() = default;
+
+  ~new_tape_impl() {
+    if (m_cur_block) {
+      //       destroy();
```
rohanjulka19, line 65: Hi  @vgvassilev  so when I comment this line the time reduces. This function calls the destructors for all the blocks. This should never be called as we are destructing the block when we pop the last element from it, but this is to be safe in case tape is terminated without popping all the values due to some error.  Now even though I have put this function inside the if stmt still this function is affecting the performance even when this is not being called. I believe this is because of function inlining. Is there way to ensure this doesn't happen ?

rohanjulka19:
Additionally on my local for BM_ReverseModeProductExecute I am getting this time 

BM_ReverseModeProductExecute      0.151 ns        0.150 ns   1000000000

But the benchmark returns this result 

BM_ReverseModeProductExecute                                     114 ns          114 ns      5956940


What must be the reason for this ?  Also I am on M2 Mac (ARM) so like what tool should I use to actually understand what's the issue ?

include/clad/Differentiator/NewTape.h
```diff
@@ -0,0 +1,122 @@
...
+    return *end();
+  }
+
+  CUDA_HOST_DEVICE void pop_back() {
+    assert(m_size || m_cur_block->prev);
+    m_size -= 1;
+    if (m_size == 0) {
+      Block<NonConstT>* temp = m_cur_block;
+      m_cur_block = m_cur_block->prev;
+      temp->~Block<NonConstT>();
```
vgvassilev, line 107: Why we do not `delete` the `m_cur_block` here?

include/clad/Differentiator/NewTape.h
vgvassilev: Probably that's right. If you compile the code with clang you can add `-Rpass-missed=.*` to get more information about the missed optimization opportunities.

vgvassilev:
> Additionally on my local for BM_ReverseModeProductExecute I am getting this time
> 
> BM_ReverseModeProductExecute 0.151 ns 0.150 ns 1000000000
> 
> But the benchmark returns this result
> 
> BM_ReverseModeProductExecute 114 ns 114 ns 5956940
> 
> What must be the reason for this ? Also I am on M2 Mac (ARM) so like what tool should I use to actually understand what's the issue ?

Do you mean that the bots give different benchmark results? If so, that's expected because they are different machines with different properties.

rohanjulka19:
> > Additionally on my local for BM_ReverseModeProductExecute I am getting this time
> > BM_ReverseModeProductExecute 0.151 ns 0.150 ns 1000000000
> > But the benchmark returns this result
> > BM_ReverseModeProductExecute 114 ns 114 ns 5956940
> > What must be the reason for this ? Also I am on M2 Mac (ARM) so like what tool should I use to actually understand what's the issue ?
> 
> Do you mean that the bots give different benchmark results? If so, that's expected because they are different machines with different properties.

Yeah but in my laptop the new tape is perfoming better than old tape and benchmark run be the pipeline the old tape is doing better than the new tape

include/clad/Differentiator/NewTape.h
rohanjulka19: Here we are deleting the current block I just added the destroy in the tape destructor to ensure that incase tape terminates without popping all the values, it should then call destructor for all the pending blocks.

rohanjulka19:
```
  ~new_tape_impl() {
    if (m_cur_block) {
      //       destroy();
```

In this line I don't want the compiler optimization and instead keep the code of destroy function separate because if it is added to the destructor it will slow down the destructor even though the function destroy won't be likely be called.  If like we are sure that all the values will always be popped before the tape terminates then I can just remove this function from the tape destructor

vgvassilev:
> > > Additionally on my local for BM_ReverseModeProductExecute I am getting this time
> > > BM_ReverseModeProductExecute 0.151 ns 0.150 ns 1000000000
> > > But the benchmark returns this result
> > > BM_ReverseModeProductExecute 114 ns 114 ns 5956940
> > > What must be the reason for this ? Also I am on M2 Mac (ARM) so like what tool should I use to actually understand what's the issue ?
> > 
> > 
> > Do you mean that the bots give different benchmark results? If so, that's expected because they are different machines with different properties.
> 
> Yeah but in my laptop the new tape is perfoming better than old tape and according to the benchmarks exectued by the pipeline the old tape is doing better than the new tape

Do you run your benchmarks on release builds?

Could you log in the bots and run some profiling there? You can do that by following: https://clad.readthedocs.io/en/latest/user/DevelopersDocumentation.html#debugging-github-runners

Issue #789, opened 2024-02-29
Add support for lambda expressions


vaithak:
Minimal working example:
```cpp
#include "clad/Differentiator/Differentiator.h"
#include <iostream>
double func(double i, double j)
{
  //auto myLambda = [=](double t)
  auto myLambda = [](double t)
  {
    return t*t + 1.0;
  };
  return i + myLambda(j);
}
int main()
{
  auto func_grad = clad::gradient(func);
  double di = 0;
  double dj = 0;
  std::cout << "func(3,4) = " << func(3, 4) << '\n';
  func_grad.execute(3., 4., &di, &dj);
  std::cout << "gradient at (3,4): (" << di << ", " << dj << ")\n";
  return 0;
}
```

PR #955, opened 2024-06-21
Add support for simple lambda expressions in reverse mode

gojakuch:
This commit provides support for primitive lambda expressions with no captures in reverse mode in the same way they are currently supported in the forward mode (#937). That is, the lambda expressions are not visited yet. Instead, the lambda functions are treated as a special case of functors.

Fixes: #789

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -17,6 +17,7 @@
 #include "clad/Differentiator/StmtClone.h"
 
 #include "clang/AST/ASTContext.h"
+#include "clang/AST/ASTLambda.h"
 #include "clang/AST/Expr.h"
 #include "clang/AST/Stmt.h"
 #include "clang/AST/TemplateBase.h"
@@ -1596,13 +1597,20 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     StmtDiff baseDiff;
     // If it has more args or f_darg0 was not found, we look for its pullback
     // function.
+    const auto* MD = dyn_cast<CXXMethodDecl>(FD);
     if (!OverloadedDerivedFn) {
       size_t idx = 0;
 
       /// Add base derivative expression in the derived call output args list if
       /// `CE` is a call to an instance member function.
-      if (const auto* MD = dyn_cast<CXXMethodDecl>(FD)) {
-        if (MD->isInstance()) {
+      if (MD) {
+        if (isLambdaCallOperator(MD)) {
+          QualType ptrType = m_Context.getPointerType(m_Context.getRecordType(
+              FD->getDeclContext()->getOuterLexicalRecordContext()));
+          baseDiff =
+              StmtDiff(Clone(dyn_cast<CXXOperatorCallExpr>(CE)->getArg(0)),
+                       new (m_Context) CXXNullPtrLiteralExpr(ptrType, Loc));
+        } else if (MD->isInstance()) {
           const Expr* baseOriginalE = nullptr;
           if (const auto* MCE = dyn_cast<CXXMemberCallExpr>(CE))
             baseOriginalE = MCE->getImplicitObjectArgument();
@@ -1700,7 +1708,10 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         pullbackRequest.EnableTBRAnalysis = m_DiffReq.EnableTBRAnalysis;
         bool isaMethod = isa<CXXMethodDecl>(FD);
         for (size_t i = 0, e = FD->getNumParams(); i < e; ++i)
-          if (DerivedCallOutputArgs[i + isaMethod])
+          if (MD && isLambdaCallOperator(MD)) {
+            if (const auto* paramDecl = FD->getParamDecl(i))
+              pullbackRequest.DVI.push_back(paramDecl);
+          } else if (DerivedCallOutputArgs[i + isaMethod])
             pullbackRequest.DVI.push_back(FD->getParamDecl(i));
 
         FunctionDecl* pullbackFD = nullptr;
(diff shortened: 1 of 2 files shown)


test/Gradient/Lambdas.C
```diff
@@ -0,0 +1,34 @@
...
+}
+
+double f2(double i, double j) {
+  auto _f = [] (double t, double k) {
+    return t + k;
+  };
+  double x = _f(i + j + _f(j, i), i);
+  return x;
+}
+
```
vgvassilev, line 23: We should check the expected generated code.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1594,7 +1597,13 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       /// Add base derivative expression in the derived call output args list if
       /// `CE` is a call to an instance member function.
       if (const auto* MD = dyn_cast<CXXMethodDecl>(FD)) {
-        if (MD->isInstance()) {
+        if ((isLambda = isLambdaCallOperator(MD))) {
```
vgvassilev, line 1600:
```suggestion
        if (isLambdaCallOperator(MD)) {
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1584,6 +1585,8 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     // requires (slightly) different arguments.
     llvm::SmallVector<Expr*, 16> pullbackCallArgs{};
 
+    bool isLambda = false;
+
```
vgvassilev, line 1589:
```suggestion
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2707,14 +2719,44 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    const auto* declsBegin = DS->decls().begin();
+    if (declsBegin != DS->decls().end() && isa<VarDecl>(*declsBegin)) {
+      auto* VD = dyn_cast<VarDecl>(*declsBegin);
+      QualType QT = VD->getType();
+      if (!QT->isPointerType()) {
+        auto* typeDecl = QT->getAsCXXRecordDecl();
+        // We should also simply copy the original lambda. The differentiation
+        // of lambdas is happening in the `VisitCallExpr`. For now, only the
+        // declarations with lambda expressions without captures are supported.
+        isNotLambda = !(typeDecl && typeDecl->isLambda());
```
vgvassilev, line 2736:
```suggestion
        if (typeDecl && !typeDecl->isLambda())
          continue;
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2707,14 +2719,44 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     bool promoteToFnScope =
         !getCurrentScope()->isFunctionScope() &&
         m_DiffReq.Mode != DiffMode::reverse_mode_forward_pass;
+
+    // If the DeclStmt is not empty, check the first declaration in case it is a
+    // lambda function. This case it treated separately for now and we don't
+    // create a variable for its derivative.
+    bool isNotLambda = true;
```
vgvassilev, line 2726: We should not rely on such state when possible.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2707,14 +2719,44 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+        // We should also simply copy the original lambda. The differentiation
+        // of lambdas is happening in the `VisitCallExpr`. For now, only the
+        // declarations with lambda expressions without captures are supported.
+        isNotLambda = !(typeDecl && typeDecl->isLambda());
+        if (!isNotLambda) {
+          for (auto* D : DS->decls())
+            if (auto* VD = dyn_cast<VarDecl>(D))
+              decls.push_back(VD);
+            else
+              diag(DiagnosticsEngine::Warning, D->getEndLoc(),
```
vgvassilev, line 2742: Can we create a test case which triggers that diagnostic?

lib/Differentiator/ReverseModeVisitor.cpp
gojakuch: this is not the only place I need to perform this check. sorry, I just don't understand why I can't simply save the result of this into a bool variable because this would make more sense than moving the typeDecl variable into a more general scope and checking `typeDecl && !typeDecl->isLambda()` each time I need to check if its a lambda.

test/Gradient/Lambdas.C
```diff
@@ -0,0 +1,83 @@
...
+#include "clad/Differentiator/Differentiator.h"
+
+double f1(double i, double j) {
+  auto _f = [] (double t) {
+    return t*t + 1.0;
+  };
+  return i + _f(j);
+}
+
+// CHECK:     inline void operator_call_pullback(double t, double _d_y, double *_d_t) const;
```
vgvassilev, line 16: How does the body of `operator_call_pullback` look like? We should check that, too.

test/Gradient/Lambdas.C
```diff
@@ -0,0 +1,83 @@
...
+// CHECK-NEXT:     void f1_grad(double i, double j, double *_d_i, double *_d_j) {
+// CHECK-NEXT:         auto _f = [](double t) {
+// CHECK-NEXT:             return t * t + 1.;
```
vgvassilev, line 19: Shouldn't we differentiate `_f` into `2*t`?

test/Gradient/Lambdas.C
gojakuch: yes, I've been working on that, I saw your initial comment on that. I only pushed these changes now, as I tried to fix the failing tests in the same time.

test/Gradient/Lambdas.C
gojakuch: the pullback code is checked in the end of the file, as clad outputs it in this order. it is basically `2*t`, yes.

vgvassilev (approved):
LGTM!

PR #960, opened 2024-06-24
Don't synthesize 0 of the ArraySubscriptExpr type in forward mode

PetroZarytskyi:
Currently, Clad can't differentiate ArraySubscriptExpr of non-numerical types. This happens because it attempts to synthesize a 0 of the same type, which is not always numerical. To avoid this error, we can use ``getZeroInit`` instead.
Here's a reproducer (it's added to the tests in this PR):
```
struct A {
  double mem;
  A(double p_mem = 0) : mem(p_mem) {}
};

double fn18(double i, double j) {
  A v[2] = {2, 3};
  v[0] = 9 * i;
  return v[0].mem;
}
```
Differentiating this in forward mode produces an error.

This bug arises when differentiating HepEmShow.


vgvassilev (approved):
LGTM!

PR #964, opened 2024-07-02
Simplify the reverse mode by having just one DiffMode for it

PetroZarytskyi:
Currently, on master, we have two reverse diff modes: ``DiffMode::reverse`` for gradients and ``DiffMode::experimental_pullback`` for pullbacks. In this PR, they are essentially merged into ``DiffMode::reverse``. This has been achieved by placing a pullback in the gradient overload instead of a gradient function.
Let's consider an example:
```cpp
// Original code:
double f(double a, double b) {
    ...
}

int main() {
    auto df = clad::gradient(f);
    ...
}
```
->
On master:
```cpp
// Generated derivative code:
double f_grad(double a, double b, double *_d_a, double *_d_b) {
    ...
}

// This overload is placed in the clad::gradient call
double f_grad(double a, double b, void *_temp_d_a, void *_temp_d_b) {
    double *_d_a = (double *)_temp_d_a;
    double *_d_b = (double *)_temp_d_b;
    f_grad(a, b, _d_a, _d_b);
}
```
In this PR:
```cpp
// Generated derivative code:
double f_pullback(double a, double b, double _d_y, double *_d_a, double *_d_b) {
    ...
}

// This is placed in the clad::gradient call
double f_grad(double a, double b, void *_temp_d_a, void *_temp_d_b) {
    double *_d_a = (double *)_temp_d_a;
    double *_d_b = (double *)_temp_d_b;
    f_pullback(a, b, 1, _d_a, _d_b);
}
```

Note: To make this system work with error estimation, I had to enable overloads there. To do that, I had to change the type of ``_final_error`` parameters from ``double&`` to ``double*``.

Advantages:
1) On master, we have 11 DiffModes, many of which use the same visitors. Having a unified reverse DiffMode makes the system easier to understand.
2) In RMV, ``Derive`` and ``DerivePullback`` do almost the same job. This PR removes ``DerivePullback`` completely.
3) With this PR, clad does not use overloads for the reverse mode anymore: just one gradient function and one pullback function. This is a great step towards supporting C, which does not have overloads.
4) Differentiating recursive functions used to generate both the gradient and the pullback. Now only the pullback is generated.

Disadvantages:
1) Now gradient forward declaration is only supported with ``void*`` adjoint parameter types. e.g. for a function ``double f(double a, double b)``, it doesn't make sense anymore to forward declare ``void f_grad(double a, double b, double *_d_a, double *_d_b)``. The options ``void f_grad(double a, double b, void *_d_a, void *_d_b)`` and ``void f_pullback(double a, double b, double _d_y, void *_d_a, void *_d_b)`` still work.
However, forward declarations don't seem to be that widely used. For example, when we changed all array_ref adjoint types to pointers in the gradient signature, this didn't break a single ROOT test. The main way to execute derivatives (with ``CladFunction``) works as before.
2) Now all differentiated functions have the pullback ``_d_y`` parameter. This may make it harder to understand the derivative code. Moreover, every time the function has a parameter named ``y``, the pullback parameter will be renamed to ``_d_y0`` to avoid name collisions. This could make the code even more confusing. However, we can fix the last problem by giving the pullback parameter a different name.

### DIFF
--- a/include/clad/Differentiator/FunctionTraits.h
+++ b/include/clad/Differentiator/FunctionTraits.h
@@ -482,8 +482,7 @@ namespace clad {
   // GradientDerivedEstFnTraits specializations for pure function pointer types
   template <class ReturnType, class... Args>
   struct GradientDerivedEstFnTraits<ReturnType (*)(Args...)> {
-    using type = void (*)(Args..., OutputParamType_t<Args, Args>...,
-                          double&);
+    using type = void (*)(Args..., OutputParamType_t<Args, void>..., void*);
   };
 
   /// These macro expansions are used to cover all possible cases of
@@ -495,12 +494,12 @@ namespace clad {
   /// qualifier and reference respectively. The AddNOEX adds cases for noexcept
   /// qualifier only if it is supported and finally AddSPECS declares the
   /// function with all the cases
-#define GradientDerivedEstFnTraits_AddSPECS(var, cv, vol, ref, noex)           \
-  template <typename R, typename C, typename... Args>                          \
-  struct GradientDerivedEstFnTraits<R (C::*)(Args...) cv vol ref noex> {       \
-    using type = void (C::*)(Args..., OutputParamType_t<Args, Args>...,           \
-                             double&) cv vol ref noex;                         \
-  };
+#define GradientDerivedEstFnTraits_AddSPECS(var, cv, vol, ref, noex)         \
+    template <typename R, typename C, typename... Args>                        \
+    struct GradientDerivedEstFnTraits<R (C::*)(Args...) cv vol ref noex> {     \
+      using type = void (C::*)(Args..., OutputParamType_t<Args, void>...,      \
+                               void*) cv vol ref noex;                         \
+    };
 
 #if __cpp_noexcept_function_type > 0
 #define GradientDerivedEstFnTraits_AddNOEX(var, con, vol, ref)                 \
--- a/test/Hessian/TemplateFunctors.C
+++ b/test/Hessian/TemplateFunctors.C
@@ -15,9 +15,9 @@ template <typename T> struct Experiment {
 
 // CHECK: void operator_call_hessian(double i, double j, double *hessianMatrix) {
 // CHECK-NEXT:     Experiment<double> _d_this;
-// CHECK-NEXT:     this->operator_call_darg0_grad(i, j, &_d_this, hessianMatrix + {{0U|0UL}}, hessianMatrix + {{1U|1UL}});
+// CHECK-NEXT:     this->operator_call_darg0_pullback(i, j, 1, &_d_this, hessianMatrix + {{0U|0UL}}, hessianMatrix + {{1U|1UL}});
 // CHECK-NEXT:     Experiment<double> _d_this0;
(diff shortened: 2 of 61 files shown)


PetroZarytskyi:
This is a big change so having different opinions on the PR would be great. We discussed the idea with @vgvassilev.
@vaithak I'd love to know your thoughts.

vaithak:
This looks really good 👍🏼 
Thanks, @PetroZarytskyi, for improving this.

vgvassilev:
This PR also Partially addresses https://github.com/vgvassilev/clad/issues/721.

test/Hessian/TemplateFunctors.C
```diff
@@ -15,9 +15,9 @@ template <typename T> struct Experiment {
 
 // CHECK: void operator_call_hessian(double i, double j, double *hessianMatrix) {
 // CHECK-NEXT:     Experiment<double> _d_this;
-// CHECK-NEXT:     this->operator_call_darg0_grad(i, j, &_d_this, hessianMatrix + {{0U|0UL}}, hessianMatrix + {{1U|1UL}});
+// CHECK-NEXT:     this->operator_call_darg0_pullback(i, j, 1, &_d_this, hessianMatrix + {{0U|0UL}}, hessianMatrix + {{1U|1UL}});
```
vgvassilev, line 18: Passing `1` as hardcoded number for `d_x` will probably be some sort of bugs/confusion. Can we make sure somehow we require literals either `0` or `1`? Maybe some enum constant...

test/Misc/RunDemos.C
```diff
@@ -119,20 +119,20 @@
 //CHECK_FLOAT_SUM:         clad::push(_t1, sum);
 //CHECK_FLOAT_SUM:         sum = sum + x;
 //CHECK_FLOAT_SUM:     }
-//CHECK_FLOAT_SUM:     _d_sum += 1;
+//CHECK_FLOAT_SUM:     _d_sum += _d_y;
 //CHECK_FLOAT_SUM:     for (; _t0; _t0--) {
 //CHECK_FLOAT_SUM:         i--;
 //CHECK_FLOAT_SUM:         {
-//CHECK_FLOAT_SUM:             _final_error += std::abs(_d_sum * sum * 1.1920928955078125E-7);
+//CHECK_FLOAT_SUM:             *_final_error += std::abs(_d_sum * sum * 1.1920928955078125E-7);
```
vgvassilev, line 126: Why do we pass this by pointer now?

include/clad/Differentiator/FunctionTraits.h
```diff
@@ -482,8 +482,7 @@ namespace clad {
   // GradientDerivedEstFnTraits specializations for pure function pointer types
   template <class ReturnType, class... Args>
   struct GradientDerivedEstFnTraits<ReturnType (*)(Args...)> {
-    using type = void (*)(Args..., OutputParamType_t<Args, Args>...,
-                          double&);
+    using type = void (*)(Args..., OutputParamType_t<Args, void>..., void*);
```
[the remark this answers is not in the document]
PetroZarytskyi: This is necessary to enable overloads in error estimation. ``double&`` cannot be converted to ``void*`` so I replaced it with ``void*`` in the overload and ``double*`` in other places. This is the reason ``_final_error`` became pointer-type.

test/Misc/RunDemos.C
PetroZarytskyi: To enable overloads in error estimation (I explained this in more detail above).

PR #967, opened 2024-07-03
Fix the derivative of string literals in forward mode

gojakuch:
This commit makes Clad set the derivative of string literals to an empty string in the forward mode. Differentiating string literals used to produce integer zero literals previously. This is important for enabling custom pushforwards for functions with string arguments and for Kokkos support in particular.

### DIFF
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -1662,10 +1662,9 @@ BaseForwardModeVisitor::VisitCharacterLiteral(const CharacterLiteral* CL) {
 }
 
 StmtDiff BaseForwardModeVisitor::VisitStringLiteral(const StringLiteral* SL) {
-  llvm::APInt zero(m_Context.getIntWidth(m_Context.IntTy), /*value*/ 0);
-  auto* constant0 =
-      IntegerLiteral::Create(m_Context, zero, m_Context.IntTy, noLoc);
-  return StmtDiff(Clone(SL), constant0);
+  return StmtDiff(Clone(SL), StringLiteral::Create(
+                                 m_Context, "", SL->getKind(), SL->isPascal(),
+                                 SL->getType(), utils::GetValidSLoc(m_Sema)));
 }
 
 StmtDiff BaseForwardModeVisitor::VisitWhileStmt(const WhileStmt* WS) {
--- a/test/FirstDerivative/Variables.C
+++ b/test/FirstDerivative/Variables.C
@@ -4,6 +4,7 @@
 
 #include "clad/Differentiator/Differentiator.h"
 #include <cmath>
+#include <string>
 
 double f_x(double x) {
   double t0 = x;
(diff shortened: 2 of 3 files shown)


vgvassilev:
Can we add a test for this?

gojakuch:
I've added this second commit because the name of a test's executable was clashing with another test which caused possible failures on my machine.

vgvassilev:
If we merge the PR with two commits the first one will fail pass the tests, right?

gojakuch:
@vgvassilev yes, it's very likely

vaithak:
This is breaking the semantics of the contract and can require even further hard-coding or handling of such cases. Can you provide an example of why this is required for Kokkos?

gojakuch:
@vaithak well, in Kokkos everything is named with a string, be that a `View` or a `parallel_for` loop. my current issue is that if I try to provide custom pushforwards for things like `View` constructors, they need to accept both `name` and `_d_name`, which are strings. currently, the derivative of a string literal is set to a zero integer, which absolutely breaks this code. For Kokkos, it doesn't really matter what exactly the derivative of a string would be, it just should be a valid string literal of some sort. I can probably make it work with empty strings in the pushforwards. 

I actually didn't understand what this PR breaks, but if the problem is with copying the initial string, I just need it to have a valid type and that's it. like, in reverse mode, I believe, there's an empty string literal generated, that'd already be better.

gojakuch:
and even not considering Kokkos, I think creating an integer zero as a derivative of a string literal is quite weird.

gojakuch:
I've updated the commit to produce empty strings instead and changed the commit/PR messages accordingly

vgvassilev:
I believe @vaithak was asking for the original example that motivated that change. Maybe we can infer more semantics from it and make a better review.

gojakuch:
> I believe @vaithak was asking for the original example that motivated that change. Maybe we can infer more semantics from it and make a better review.

yes, I understand, I didn't want to make it seem rude, sorry @vaithak if that seemed so. I was just been trying to explain that it's not only Kokkos that needs this to be implemented. I was initially motivated by this small example though:

```c++
double fn(double x) {
    const int N = 2;
    Kokkos::View<double**, Kokkos::HostSpace> a("a", N, N);

    int i = 1;
    a(i, i) = x;
    return a(i, i);
} // = x
```

here, to be able to properly create a derivative for the View object, we need to provide a pushforward for the constructor, which I did. But generating `0` as a derivative of `"a"` breaks it at runtime:
```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
Aborted (core dumped)
```
because you can't build a string from a `0`.

vaithak:
No worries @gojakuch, thanks for the example 👍
I agree with your point that return an integer literal is wrong. My reason to ask for an example was to infer if we need the original string, or would empty string also work. 
Empty string would be helpful in future cases, where the parent visitors want to have a check like "do something if the child visitor's derivative is null".
In such cases, we can handle empty string literals similar to the null derivative case.

vaithak:
One suggestion I would have is to add a test case with similar complexity as the intended use case. 
In this case, passing string literal as an argument to a function call, to check/ensure the behaviours of pushforwards.

gojakuch:
> One suggestion I would have is to add a test case with similar complexity as the intended use case. In this case, passing string literal as an argument to a function call, to check/ensure the behaviours of pushforwards.

I actually thought about this. I'm going to add it now. thanks for the suggestion.

vaithak (approved):

PR #971, opened 2024-07-07
Move the TBR analysis in DiffRequest and provide an interface to query it.

vgvassilev:
The idea of this change is to separate the analysis steps from the building step. That would help us detect unsupported cases and provide useful diagnostics without crashing. The centralization will help with enabling the future activity analysis in other modes, too.

Partially addresses #721.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -1,5 +1,7 @@
 #include "clad/Differentiator/DiffPlanner.h"
 
+#include "TBRAnalyzer.h"
+
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/RecursiveASTVisitor.h"
 #include "clang/Basic/SourceManager.h"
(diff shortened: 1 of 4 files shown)


lib/Differentiator/TBRAnalyzer.h
```diff
@@ -224,7 +224,7 @@ class TBRAnalyzer : public clang::RecursiveASTVisitor<TBRAnalyzer> {
   enum Mode { kMarkingMode = 1, kNonLinearMode = 2 };
   /// Tells if the variable at a given location is required to store. Basically,
   /// is the result of analysis.
-  std::set<clang::SourceLocation> m_TBRLocs;
+  std::set<clang::SourceLocation>& m_TBRLocs;
```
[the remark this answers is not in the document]
vgvassilev: This is bogus. cppcoreguidelines-avoid-const-or-ref-data-members says "his check warns when structs or classes that are copyable or movable, and have const-qualified or reference (lvalue or rvalue) data members.". The `TBRAnalyzer` has deleted copy and move operations but the complain is still on. I suspect there is a bug in clang-tidy...

vgvassilev:
@PetroZarytskyi, this change triggers

```
Unexpectedly Passed Tests (1):
  clad :: Gradient/PointersWithTBR.C
```

Do you have a clue why that happens?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -559,6 +561,29 @@ namespace clad {
     return;
   }
 
+  bool DiffRequest::shouldBeRecorded(Expr* E) const {
+    assert(EnableTBRAnalysis && "TBR not enabled!");
+
+    if (!isa<DeclRefExpr>(E) || !isa<ArraySubscriptExpr>(E) ||
+        !isa<MemberExpr>(E))
```
PetroZarytskyi, line 568: I think the condition should be ``!isa<DeclRefExpr>(E) && !isa<ArraySubscriptExpr>(E) &&
        !isa<MemberExpr>(E)``.
         ``DeclRefExpr``, ``ArraySubscriptExpr``, and ``MemberExpr`` are different expression types so at least one of the three conditions ``!isa<...`` will always be ``true``. Because of this, the function always returns true here. This is why we get the ``Unexpectedly Passed`` check (TBR analysis is essentially turned off).

PetroZarytskyi:
Other than my last comment, the PR looks good to me.

lib/Differentiator/DiffPlanner.cpp
vgvassilev: Ah, yeah, good catch!

PetroZarytskyi (approved):
LGTM!

vgvassilev:
The failure on the ubuntu llvm-18 bot is due to some apt-get problem. I do not understand why codecov is not happy with the test coverage and at the same time it complains about reduced coverage. Maybe if we removed code the coverage computation goes wrong...

PR #973, opened 2024-07-08
Add shell.nix with an environment for building and using Clad on NixOS

guitargeek:
I use this nix environment to build Clad standalone and test the reproducers.

Until recently, I could just use my native Arch Linux environment to do
so, but after Arch dropped LLVM 17 this was not possible anymore (Clad
doesn't work the LLVM 18 yet). Therefore, I had to find another
solution, and nix packages works quite well.

It would be nice to have this in the repo so also other people can work
with Clad in NixOS, without having to waste time figuring out the
dependencies and necessary CMake flags.

Also, add a flag `CLAD_DISABLE_TESTS` to the build system, so that Clad can be built without the hack of resetting the compiler flags temporarily by hand.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -324,28 +324,34 @@ if (NOT CLAD_BUILD_STATIC_ONLY)
   add_subdirectory(demos/ErrorEstimation/CustomModel)
   add_subdirectory(demos/ErrorEstimation/PrintModel)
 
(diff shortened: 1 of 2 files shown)


vgvassilev (approved):
I know nothing about Nix but that LGTM ;) Thank you, @guitargeek!

PR #975, opened 2024-07-09
Fix pointer dereference in fwd mode

### LINKED ISSUE #972: Dereferencing of input parameters doesn't work in forward mode
This is related to #961, only that now the dereferenced array is a parameter to the function.

Reproducer:
```c++
// Compile with clang++ -std=c++11 -fplugin=install/clad.so -I./install/include cladIssue.cxx -o cladIssue

#include <iostream>
#include <vector>

double wrapper(double* params, double const* constants) {
   double c0 = *constants;
   return params[0] * c0;
}

int main()
{

   double x = 2.0;
   double y = 2.0;
   double result = 0.0;
   grad.execute(&x, &y, &result);

   std::cout << diff.execute(&x, &y) << std::endl;
}
```
Compiler output:
```txt
error: indirection requires pointer operand ('int' invalid)
1 error generated.
```

I'm using LLVM and Clang 16 to build clad standalone (see #973).

vaithak:
fixes #972


lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1005,8 +1005,11 @@ StmtDiff BaseForwardModeVisitor::VisitDeclRefExpr(const DeclRefExpr* DRE) {
   }
   // Is not a variable or is a reference to something unrelated to independent
   // variable. Derivative is 0.
-  auto zero = ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
-  return StmtDiff(clonedDRE, zero);
+  // If DRE is of type pointer, then the derivative is a null pointer.
+  if (clonedDRE->getType()->isPointerType())
+    return StmtDiff(clonedDRE, nullptr);
+  return StmtDiff(clonedDRE, ConstantFolder::synthesizeLiteral(
+                                 m_Context.IntTy, m_Context, /*val=*/0));
```
vgvassilev, line 1012:
```suggestion
  QualType clonedDreTy = clonedDRE->getType();
  if (clonedDreTy->isPointerType())
    return StmtDiff(clonedDRE, nullptr);
  return StmtDiff(clonedDRE, ConstantFolder::synthesizeLiteral(
                                 clonedDreTy, m_Context, /*val=*/0));
```

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1315,7 +1318,11 @@ StmtDiff BaseForwardModeVisitor::VisitUnaryOperator(const UnaryOperator* UnOp) {
            opKind == UnaryOperatorKind::UO_Imag) {
     return StmtDiff(op, BuildOp(opKind, diff.getExpr_dx()));
   } else if (opKind == UnaryOperatorKind::UO_Deref) {
-    return StmtDiff(op, BuildOp(opKind, diff.getExpr_dx()));
+    Expr* dx = diff.getExpr_dx();
+    if (dx == nullptr)
+      return StmtDiff(op, ConstantFolder::synthesizeLiteral(
+                              m_Context.IntTy, m_Context, /*val=*/0));
+    return StmtDiff(op, BuildOp(opKind, dx));
```
vgvassilev, line 1325:
```suggestion

    if (Expr* dx = diff.getExpr_dx())
      return StmtDiff(op, BuildOp(opKind, dx));
    return StmtDiff(op, ConstantFolder::synthesizeLiteral(
                              m_Context.IntTy, m_Context, /*val=*/0));
```
I know that we are not using this consistently but `IntTy` here is probably not what we want semantically. Shall we give it a try with a `VoidPtrTy` or a `NullPtrTy`?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1005,8 +1005,14 @@ StmtDiff BaseForwardModeVisitor::VisitDeclRefExpr(const DeclRefExpr* DRE) {
...
   // Is not a variable or is a reference to something unrelated to independent
   // variable. Derivative is 0.
-  auto zero = ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
-  return StmtDiff(clonedDRE, zero);
+  // If DRE is of type pointer, then the derivative is a null pointer.
+  if (clonedDRE->getType()->isPointerType())
+    return StmtDiff(clonedDRE, nullptr);
+  QualType clonedDreTy = clad::utils::GetValueType(clonedDRE->getType());
+  if (!clonedDreTy->isRealType())
+    clonedDreTy = m_Context.IntTy; // force int if not real, ex. custom type
```
vgvassilev, line 1013: Should we fix the implementation of `ConstantFolder::synthesizeLiteral` to generate maybe `{}` initializer list which is supposed to call the default constructor of the custom type?

PS: Sorry for making this hard as that's not quite related to this PR...

vaithak: It wouldn't be a good idea. 
It can be done, but we will be assuming that a default constructor exists for this custom type, which properly zero initializes the data members. This assumption may constrain what we can differentiate.

vgvassilev: Sure, but how this is different from what we have now scattered all over the place? I guess I was proposing to centralize the logic there and if the type does not have a default constructor with proper initialization we can complain...

vgvassilev:
Maybe we should revert the PR to the old state where everything passed, merge it and continue with this change separately. That would unblock the release.

vaithak:
> Maybe we should revert the PR to the old state where everything passed, merge it and continue with this change separately. That would unblock the release.

Updated 👍🏼

vgvassilev (approved):
LGTM! Can you open a new one with the changes that we discussed, too?

Issue #974, opened 2024-07-08
Building calls to pushforwards of functions that accept string arguments fails


gojakuch:
Clad fails to build calls to functions with string arguments (at least in the forward mode), specifically like in the example below that I noticed initially:

```c++
#include "clad/Differentiator/Differentiator.h"
#include <iostream>

namespace clad { namespace custom_derivatives {
    ValueAndPushforward<double, double> fs_pushforward(double x, const ::std::string& s, double _d_x, const ::std::string& _d_s) {
        return {x, 1};
    }
}}

double fs(double x, const std::string& s) {
  return x;
}

double f(double x) {
  return fs(x, "text");
}

int main() {
  auto df = clad::differentiate(f, 0);
  std::cout << "f(3)=" << f(3.0) << "; df(3)=" << df.execute(3.0) << '\n';
  return 0;
}
```
From what I understood during my debug-investigation, the following code triggers an equivalent issue (that appears different at first glance):
```c++
#include "clad/Differentiator/Differentiator.h"
#include <iostream>

double f(double x) {
    std::string s = "text";
    return x;
}

int main() {
  auto df = clad::differentiate(f, 0);
  std::cout << "f(3)=" << f(3.0) << "; df(3)=" << df.execute(3.0) << '\n';
  return 0;
}
```

PR #976, opened 2024-07-09
Add support for `std::string` variables and functions with implicit constructor calls in their arguments

gojakuch:
Previously, Clad failed to differentiate functions with `std::string` variables in them or function calls that call a constructor implicitly. For instance, a call
```
f(x, "text");
```
to a function `double f(double, const std::string&)` would have caused a segmentation fault before this commit. Same applies to other user defined arguments, not only those of the `std::string` type.

Fixes: #974

### DIFF
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -2012,17 +2012,17 @@ BaseForwardModeVisitor::VisitCXXConstructExpr(const CXXConstructExpr* CE) {
   // forward mode derived constructor that would require same arguments as of
   // a pushforward function, that is, `{a, 1, b, _d_a, 0., _d_b}`.
   if (CE->getNumArgs() != 1) {
-    if (CE->isListInitialization()) {
-      clonedArgsE = m_Sema.ActOnInitList(noLoc, clonedArgs, noLoc).get();
-      derivedArgsE = m_Sema.ActOnInitList(noLoc, derivedArgs, noLoc).get();
-    } else if (CE->getNumArgs() == 0) {
+    if (CE->getNumArgs() == 0 && !CE->isListInitialization()) {
       // ParenList is empty -- default initialisation.
       // Passing empty parenList here will silently cause 'most vexing
       // parse' issue.
       return StmtDiff();
     } else {
-      clonedArgsE = m_Sema.ActOnParenListExpr(noLoc, noLoc, clonedArgs).get();
-      derivedArgsE = m_Sema.ActOnParenListExpr(noLoc, noLoc, derivedArgs).get();
+      // Rely on the initializer list expressions as they seem to be more
+      // flexible in terms of conversions and other similar scenarios where a
+      // constructor is called implicitly.
+      clonedArgsE = m_Sema.ActOnInitList(noLoc, clonedArgs, noLoc).get();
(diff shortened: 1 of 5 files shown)


test/FirstDerivative/Variables.C
```diff
@@ -102,7 +103,7 @@ double f_string(double x) {
 namespace clad {
 namespace custom_derivatives {
 clad::ValueAndPushforward<double, double> string_test_pushforward(double x, const char s[], double _d_x, const char *_d_s) {
-    return {0, 0};
+    return {1, 0};
```
vgvassilev, line 106: Was the previous value incorrect mathematically?

gojakuch: yes, I noticed this when copying the code from there to make a new test

test/FirstDerivative/Variables.C
```diff
@@ -118,13 +119,48 @@ double f_string_call(double x) {
...
 
+double f_stdstring(double x) {
+  std::string s = "string literal";
+  return x;
+}
+
+// CHECK: double f_stdstring_darg0(double x) {
+// CHECK-NEXT:         double _d_x = 1;
+// CHECK-NEXT:         std::string _d_s = {{[{]"", .*[}]|""}};
+// CHECK-NEXT:         std::string s = {{[{]"string literal", .*[}]|"string literal"}};
```
vaithak, line 130: Why are we not trying to match "std::allocator<char>" precisely as the second argument? Could it be different on different platforms?

test/FirstDerivative/Variables.C
vaithak: Any reason why the outputs of their `execute` method are not tested explicitly? This would have caught otherwise.

test/FirstDerivative/Variables.C
gojakuch: for some reason, I couldn't make this work. I'll try again now

gojakuch: updated this. I was just being dumb last time so it didn't match the parentheses in the allocator expression. should be more precise now

test/FirstDerivative/Variables.C
gojakuch: there were no `check-exec`s in the file previously, I added some for the new tests in this PR

vaithak (approved):
Lgtm

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -2012,17 +2012,19 @@ BaseForwardModeVisitor::VisitCXXConstructExpr(const CXXConstructExpr* CE) {
...
       // parse' issue.
       return StmtDiff();
     } else {
-      clonedArgsE = m_Sema.ActOnParenListExpr(noLoc, noLoc, clonedArgs).get();
-      derivedArgsE = m_Sema.ActOnParenListExpr(noLoc, noLoc, derivedArgs).get();
+      // We use list expressions here, as they seem to be safer to work with for
+      // Clad. In particular, this is used to avoid segmentation fault when
+      // synthesising function calls with string literal values passed to
+      // std::string arguments (issue #974) and other similar scenarios where a
+      // constructor is called implicitly.
```
vgvassilev, line 2025:
```suggestion
      // Rely on the initializer list expressions as they seem to be more
      // flexible in terms of conversions and other similar scenarios where a
      // constructor is called implicitly.
```

PR #977, opened 2024-07-10
Add basic Kokkos support for the original tests of #783

gojakuch:
This commit adds a test from the original Kokkos PR #783 and provides the implementation of the simplest Kokkos features in the form of custom pushforwards. This commit only addresses the forward mode.

### DIFF
--- a/include/clad/Differentiator/KokkosBuiltins.h
+++ b/include/clad/Differentiator/KokkosBuiltins.h
@@ -1,0 +1,54 @@
+// This header file contains the implementation of the Kokkos framework
+// differentiation support in Clad in the form of custom pushforwards and
+// pullbacks. Please include it manually to enable Clad for Kokkos code.
+
+#ifndef CLAD_DIFFERENTIATOR_KOKKOSBUILTINS_H
+#define CLAD_DIFFERENTIATOR_KOKKOSBUILTINS_H
+
+#include <Kokkos_Core.hpp>
+#include "clad/Differentiator/Differentiator.h"
+
+namespace clad::custom_derivatives {
+namespace class_functions {
+/// Kokkos arrays
+template <class DataType, class... ViewParams>
+clad::ValueAndPushforward<Kokkos::View<DataType, ViewParams...>,
+                          Kokkos::View<DataType, ViewParams...>>
+constructor_pushforward(
+    clad::ConstructorPushforwardTag<Kokkos::View<DataType, ViewParams...>>,
+    const ::std::string& name, const size_t& idx0, const size_t& idx1,
+    const size_t& idx2, const size_t& idx3, const size_t& idx4,
+    const size_t& idx5, const size_t& idx6, const size_t& idx7,
+    const ::std::string& d_name, const size_t& d_idx0, const size_t& d_idx1,
+    const size_t& d_idx2, const size_t& d_idx3, const size_t& d_idx4,
+    const size_t& d_idx5, const size_t& d_idx6, const size_t& d_idx7) {
+  return {Kokkos::View<DataType, ViewParams...>(name, idx0, idx1, idx2, idx3,
+                                                idx4, idx5, idx6, idx7),
+          Kokkos::View<DataType, ViewParams...>(
+              "_diff_" + name, idx0, idx1, idx2, idx3, idx4, idx5, idx6, idx7)};
+}
+} // namespace class_functions
+
+/// Kokkos functions
+namespace Kokkos {
+template <typename View1, typename View2, typename T>
+inline void deep_copy_pushforward(const View1& dst, const View2& src, T param,
+                                  const View1& d_dst, const View2& d_src,
+                                  T d_param) {
+  deep_copy(dst, src);
+  deep_copy(d_dst, d_src);
+}
+
+template <class ExecPolicy, class FunctorType>
+inline void
+parallel_for_pushforward(const ::std::string& str, const ExecPolicy& policy,
+                         const FunctorType& functor, const ::std::string& d_str,
+                         const ExecPolicy& d_policy,
+                         const FunctorType& d_functor) {
+  // TODO: implement parallel_for_pushforward
+  return;
+}
+} // namespace Kokkos
+} // namespace clad::custom_derivatives
+
+#endif // CLAD_DIFFERENTIATOR_KOKKOSBUILTINS_H
--- a/unittests/Kokkos/CMakeLists.txt
+++ b/unittests/Kokkos/CMakeLists.txt
@@ -1,5 +1,6 @@
 add_clad_unittest(KokkosTests
   main.cpp
+  ViewAccess.cpp
   ViewBasics.cpp
   ParallelReduce.cpp
   ParallelFor.cpp
--- a/unittests/Kokkos/ViewAccess.cpp
+++ b/unittests/Kokkos/ViewAccess.cpp
@@ -1,0 +1,73 @@
+// Tests originally created by Kim Liegeois
+
+#include "TestUtils.h"
+#include <Kokkos_Core.hpp>
+#include "clad/Differentiator/Differentiator.h"
+#include "clad/Differentiator/KokkosBuiltins.h"
+#include "gtest/gtest.h"
+
+double f(double x, double y) {
+
+  const int N1 = 4;
+
+  Kokkos::View<double* [4], Kokkos::LayoutLeft, Kokkos::HostSpace> a("a", N1);
+  Kokkos::View<double* [4], Kokkos::LayoutLeft, Kokkos::HostSpace> b("b", N1);
+
+  a(0, 0) = x;
+  b(0, 0) = y;
+
+  b(0, 0) += a(0, 0) * b(0, 0);
+
+  return a(0, 0) * a(0, 0) * b(0, 0) + b(0, 0);
+}
+
+double f_2(double x, double y) {
+
+  const int N1 = 4;
+
+  Kokkos::View<double* [4], Kokkos::LayoutLeft, Kokkos::HostSpace> a("a", N1);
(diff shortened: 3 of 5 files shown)


gojakuch:
The exact way this has been implemented in this PR is a subject for a discussion. I'm specifically talking about the way I put all the pushforwards in this file in the unittest directory. I also removed the initial tests I'd written in favour of the original ones from the PR #783 that basically cover the same things.

gojakuch:
@vgvassilev as far as I can see, clang-format and some other checks are failing because there's no valid OpenPGP data being found, so it's probably not the PR's fault.

MihailMihov:
@gojakuch It couldn't resolve the url (`curl: (6) Could not resolve host: apt.llvm.org`), but the url is working now, so maybe just re-run the checks?

gojakuch:
@MihailMihov thanks, I'll do that a bit later.

unittests/Kokkos/KokkosImplementation.h
```diff
@@ -0,0 +1,74 @@
+// This header file contains the implementation of the Kokkos framework
```
vgvassilev, line 1: This file should become `KokkosBuiltins.h`, similar to `[STLBuiltins.h](https://github.com/vgvassilev/clad/blob/master/include/clad/Differentiator/STLBuiltins.h)`. In it we should add the propagators which work with kokkos. This way users should be able to include this file and get them ready to go.

gojakuch: yes, this is exactly the part of the design that I was unsure about. I'll move it as I fix the errors in the checks

unittests/Kokkos/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 add_clad_unittest(KokkosTests
   main.cpp
-  ViewBasics.cpp
```
vgvassilev, line 3: I'd keep the ViewBasics as it has a lot of commented out code that should soon work.

gojakuch:
@vgvassilev the failing check I have right now in this PR is the most important Kokkos one and yet I cannot replicate it on my machine. it works out fine for me. what shall I do in this situation?

vgvassilev:
> @vgvassilev the failing check I have right now in this PR is the most important Kokkos one and yet I cannot replicate it on my machine. it works out fine for me. what shall I do in this situation?

Debug on the bots?

gojakuch:
> Debug on the bots?

how do I do that?

gojakuch:
@vgvassilev so basically I've managed to debug and fix this on the GitHub runner, but only by modifying a line of code in Clad's main code base. the change is quite small and honestly seems reasonable to me, but that's still something I'd like to discuss during today's meeting. I'm pushing it here for now to have all the checks passing, but maybe it'd be better to have it as a separate commit.

include/clad/Differentiator/KokkosBuiltins.h
```diff
@@ -0,0 +1,74 @@
+// This header file contains the implementation of the Kokkos framework
+// differentiation support in Clad in the form of custom pushforwards and
+// pullbacks. Please include it manually to use Clad on Kokkos code.
+
+#ifndef CLAD_KOKKOS_BUILTINS_H
+#define CLAD_KOKKOS_BUILTINS_H
```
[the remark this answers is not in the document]
vgvassilev: That's a good suggestion.

include/clad/Differentiator/KokkosBuiltins.h
```diff
@@ -0,0 +1,74 @@
...
+
+#include <Kokkos_Core.hpp>
```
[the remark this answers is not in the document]
vgvassilev: Maybe wrap this in `_has_include`?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1198,7 +1198,7 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
   // If all arguments are constant literals, then this does not contribute to
   // the gradient.
   // FIXME: revert this when this is integrated in the activity analysis pass.
-  if (!callDiff) {
+  if (!callDiff && !CE->getType().getTypePtr()->isVoidType()) {
```
vgvassilev, line 1201: We should move this as a separate PR with an appropriate test.

vgvassilev (commented):
I think this is heading in a good direction.

@kliegeois, @brian-kelley can you take a look?

include/clad/Differentiator/KokkosBuiltins.h
```diff
@@ -0,0 +1,74 @@
...
+#include "clad/Differentiator/Differentiator.h"
+
+namespace clad {
+namespace custom_derivatives {
+namespace class_functions {
+/// Kokkos arrays
+template <typename ArrayT, typename Space>
+clad::ValueAndPushforward<Kokkos::View<ArrayT, Space>,
+                          Kokkos::View<ArrayT, Space>>
+constructor_pushforward(
```
brian-kelley, line 18:
```suggestion
template <class DataType, class... ViewParams>
clad::ValueAndPushforward<Kokkos::View<DataType, ViewParams...>,
                          Kokkos::View<Kokkos::View<DataType, ViewParams...>>>
constructor_pushforward(
```

The full set of possible ``Kokkos::View`` template parameters is more than ``<datatype, space>`` or ``<datatype, layout, space>``. There could also be a memory traits type, and everything besides the datatype is optional. What about doing something like this to match all the possible cases? This would also let you merge this specialization with the one on line 32.

kliegeois (commented):
Thanks @gojakuch it looks good to me!

brian-kelley:
@gojakuch It also looks good to me, I just see that using variadic template arguments could simplify the ``constructor_pushforward`` specializations and support more general ``Kokkos::View`` types.

include/clad/Differentiator/KokkosBuiltins.h
kliegeois: Good catch @brian-kelley !

unittests/Kokkos/ViewAccess.cpp
```diff
@@ -0,0 +1,76 @@
+// Source:
+// https://github.com/vgvassilev/clad/pull/783/files#diff-c919937640674e1911d1f17d3df2b65426d152400589873ff9ba2f0391048d78
+// Source: https://github.com/kliegeois/clad/tree/kokkos-PR
```
vgvassilev, line 3:
```suggestion
```

unittests/Kokkos/ViewBasics.cpp
```diff
@@ -4,12 +4,12 @@
 // it has been modified to match gtest guidelines and improve readability
 
 #include "ParallelAdd.h"
-#include "TestUtils.h"
+// #include "TestUtils.h"
```
vgvassilev, line 7: Why this is commented out? Maybe we should remove it?

unittests/Kokkos/ViewBasics.cpp
```diff
@@ -4,12 +4,12 @@
...
 #include <Kokkos_Core.hpp>
 #include "clad/Differentiator/Differentiator.h"
 #include "gtest/gtest.h"
 
-double f(double x, double y) {
+double ff(double x, double y) {
```
vgvassilev, line 12: I'd love to use more meaningful names.

gojakuch: alright, I'll rename these. it's just that I couldn't use the same name twice, because this google test thing complains about having functions with the same name, as it compiles all the tests at once, not like clad does with the other tests.

include/clad/Differentiator/KokkosBuiltins.h
gojakuch: thanks for the suggestion, updated this.

gojakuch:
@vgvassilev the `ubu22-clang16-runtime18` check started failing both here and on another PR of mine for some reason related with package setup there, so that's not on my side

include/clad/Differentiator/KokkosBuiltins.h
```diff
@@ -0,0 +1,55 @@
+// This header file contains the implementation of the Kokkos framework
+// differentiation support in Clad in the form of custom pushforwards and
+// pullbacks. Please include it manually to use Clad on Kokkos code.
```
vgvassilev, line 3:
```suggestion
// pullbacks. Please include it manually to enable Clad for Kokkos code.
```

include/clad/Differentiator/KokkosBuiltins.h
```diff
@@ -0,0 +1,74 @@
...
+
+#ifndef CLAD_KOKKOS_BUILTINS_H
+#define CLAD_KOKKOS_BUILTINS_H
+
+#include <Kokkos_Core.hpp>
+#include "clad/Differentiator/Differentiator.h"
```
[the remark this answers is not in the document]
vgvassilev: We can ignore that suggestion.

include/clad/Differentiator/KokkosBuiltins.h
```diff
@@ -0,0 +1,74 @@
...
+          Kokkos::View<ArrayT, Layout, Space>("_diff_" + name, idx0, idx1, idx2,
+                                              idx3, idx4, idx5, idx6, idx7)};
+}
+} // namespace class_functions
+
+/// Kokkos functions
+namespace Kokkos {
+template <typename View1, typename View2, typename T>
+inline void deep_copy_pushforward(const View1& dst, const View2& src, T param,
+                                  const View1& _d_dst, const View2& _d_src,
```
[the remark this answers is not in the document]
vgvassilev: These suggestions are relevant. The `_` identifiers are normally reserved for special builtins and we are not doing anything special here.

vgvassilev (approved):
LGTM!

kliegeois:
Thanks a lot for having merged the PR @vgvassilev and @gojakuch !

Issue #723, opened 2024-01-21
Add support for range-based for loops.


vgvassilev:

7070Shreyash:
Hey @vgvassilev  , Could you provide guidance on resolving this problem? I'm new to the CLAD codebase and would appreciate assistance

krathul:
I would like to work on this issue @vgvassilev. I guess its only the body of the range-based for loop that can be differentiated right?

krathul:
@vgvassilev from what I can understand I need to first differentiate initialization statement, and then the body. And finally construct the ranged for loop block which uses derivates. 
Or will it be possible to transform the range-based for loop to a normal for loop?

PR #980, opened 2024-07-11
Adding support for range-based for loops in the reverse mode

ovdiiuv:
Adding basic support for the range-based for loops, does not yet support break and continue statements inside.

Fixes: #723

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2108,7 +2108,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
           addToCurrentBlock(add_assign, direction::reverse);
         }
       }
-      return {cloneE, derivedE};
+      return {cloneE, derivedE, derivedE};
     } else {
       if (opCode != UO_LNot)
         // We should only output warnings on visiting boolean conditions
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -976,6 +976,131 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     return StmtDiff(condExpr, ResultRef);
   }
 
+  StmtDiff
+  ReverseModeVisitor::VisitCXXForRangeStmt(const CXXForRangeStmt* FRS) {
+    beginScope(Scope::DeclScope | Scope::ControlScope | Scope::BreakScope |
+               Scope::ContinueScope);
+    beginBlock(direction::reverse);
+
+    LoopCounter loopCounter(*this);
+    const VarDecl* LoopVD = FRS->getLoopVariable();
+
+    const Stmt* RangeDecl = FRS->getRangeStmt();
+    const Stmt* BeginDecl = FRS->getBeginStmt();
+    StmtDiff VisitRange = Visit(RangeDecl);
+    StmtDiff VisitBegin = Visit(BeginDecl);
+    Expr* BeginExpr = cast<BinaryOperator>(VisitBegin.getStmt())->getLHS();
+
+    beginBlock(direction::reverse);
+    // Create all declarations needed.
+    auto* BeginDeclRef = cast<DeclRefExpr>(BeginExpr);
+    Expr* d_BeginDeclRef = m_Variables[BeginDeclRef->getDecl()];
+
+    auto* RangeExpr =
+        cast<DeclRefExpr>(cast<BinaryOperator>(VisitRange.getStmt())->getLHS());
+
+    Expr* RangeInit = Clone(FRS->getRangeInit());
+    Expr* AssignRange =
+        BuildOp(BO_Assign, RangeExpr, BuildOp(UO_AddrOf, RangeInit));
+    Expr* AssignBegin =
+        BuildOp(BO_Assign, BeginDeclRef, BuildOp(UO_Deref, RangeExpr));
+    addToCurrentBlock(AssignRange);
+    addToCurrentBlock(AssignBegin);
+    const auto* EndDecl = cast<VarDecl>(FRS->getEndStmt()->getSingleDecl());
+
+    Expr* EndInit = cast<BinaryOperator>(EndDecl->getInit())->getRHS();
+    QualType EndType = CloneType(EndDecl->getType());
+    std::string EndName = EndDecl->getNameAsString();
+    Expr* EndAssign = BuildOp(BO_Add, BuildOp(UO_Deref, RangeExpr), EndInit);
+    VarDecl* EndVarDecl =
+        BuildGlobalVarDecl(EndType, EndName, EndAssign, /*DirectInit=*/false);
+    DeclStmt* AssignEnd = BuildDeclStmt(EndVarDecl);
+
+    addToCurrentBlock(AssignEnd);
+    auto* AssignEndVarDecl =
+        cast<VarDecl>(cast<DeclStmt>(AssignEnd)->getSingleDecl());
+    DeclRefExpr* EndExpr = BuildDeclRef(AssignEndVarDecl);
+    Expr* IncBegin = BuildOp(UO_PreInc, BeginDeclRef);
+
+    beginBlock(direction::forward);
+    DeclDiff<VarDecl> LoopVDDiff = DifferentiateVarDecl(LoopVD);
+    Stmt* AdjLoopVDAddAssign =
+        utils::unwrapIfSingleStmt(endBlock(direction::forward));
+
+    if ((LoopVDDiff.getDecl()->getDeclName() != LoopVD->getDeclName() ||
+         LoopVD->getType() != LoopVDDiff.getDecl()->getType()))
+      m_DeclReplacements[LoopVD] = LoopVDDiff.getDecl();
+    llvm::SaveAndRestore<bool> SaveIsInsideLoop(isInsideLoop,
+                                                /*NewValue=*/true);
+
+    Expr* d_IncBegin = BuildOp(UO_PreInc, d_BeginDeclRef);
+    Expr* d_DecBegin = BuildOp(UO_PostDec, d_BeginDeclRef);
+    Expr* ForwardCond = BuildOp(BO_NE, BeginDeclRef, EndExpr);
+    // Add item assignment statement to the body.
+    const Stmt* body = FRS->getBody();
+    StmtDiff bodyDiff = Visit(body);
+
+    StmtDiff storeLoop = StoreAndRestore(BuildDeclRef(LoopVDDiff.getDecl()));
+    StmtDiff storeAdjLoop =
+        StoreAndRestore(BuildDeclRef(LoopVDDiff.getDecl_dx()));
+
+    addToCurrentBlock(BuildDeclStmt(LoopVDDiff.getDecl_dx()));
+    Expr* CounterIncrement = loopCounter.getCounterIncrement();
+
+    Expr* LoopInit = LoopVDDiff.getDecl()->getInit();
+    LoopVDDiff.getDecl()->setInit(getZeroInit(LoopVDDiff.getDecl()->getType()));
+    addToCurrentBlock(BuildDeclStmt(LoopVDDiff.getDecl()));
+    Expr* AssignLoop =
+        BuildOp(BO_Assign, BuildDeclRef(LoopVDDiff.getDecl()), LoopInit);
+
+    if (!LoopVD->getType()->isReferenceType()) {
+      Expr* d_LoopVD = BuildDeclRef(LoopVDDiff.getDecl_dx());
+      AdjLoopVDAddAssign =
+          BuildOp(BO_Assign, d_LoopVD, BuildOp(UO_Deref, d_BeginDeclRef));
+    }
+
+    beginBlock(direction::forward);
+    addToCurrentBlock(CounterIncrement);
+    addToCurrentBlock(AdjLoopVDAddAssign);
+    addToCurrentBlock(AssignLoop);
+    addToCurrentBlock(storeLoop.getStmt());
+    addToCurrentBlock(storeAdjLoop.getStmt());
+    CompoundStmt* LoopVDForwardDiff = endBlock(direction::forward);
+    CompoundStmt* bodyForward = utils::PrependAndCreateCompoundStmt(
+        m_Sema.getASTContext(), bodyDiff.getStmt(), LoopVDForwardDiff);
+
+    beginBlock(direction::forward);
+    addToCurrentBlock(d_DecBegin);
+    addToCurrentBlock(storeLoop.getStmt_dx());
+    addToCurrentBlock(storeAdjLoop.getStmt_dx());
+    CompoundStmt* LoopVDReverseDiff = endBlock(direction::forward);
+    CompoundStmt* bodyReverse = utils::PrependAndCreateCompoundStmt(
+        m_Sema.getASTContext(), bodyDiff.getStmt_dx(), LoopVDReverseDiff);
+
+    Expr* Inc = BuildOp(BO_Comma, IncBegin, d_IncBegin);
+    Stmt* Forward = new (m_Context) ForStmt(
+        m_Context, /*Init=*/nullptr, ForwardCond, /*CondVar=*/nullptr, Inc,
+        bodyForward, FRS->getForLoc(), FRS->getBeginLoc(), FRS->getEndLoc());
+    Expr* CounterCondition =
+        loopCounter.getCounterConditionResult().get().second;
+    Expr* CounterDecrement = loopCounter.getCounterDecrement();
+
+    Stmt* Reverse = bodyReverse;
+    addToCurrentBlock(Reverse, direction::reverse);
+    Reverse = endBlock(direction::reverse);
+
+    Reverse = new (m_Context)
+        ForStmt(m_Context, /*Init=*/nullptr, CounterCondition,
+                /*CondVar=*/nullptr, CounterDecrement, Reverse,
+                FRS->getForLoc(), FRS->getBeginLoc(), FRS->getEndLoc());
+    addToCurrentBlock(Reverse, direction::reverse);
+    Reverse = endBlock(direction::reverse);
+    endScope();
+
+    return {utils::unwrapIfSingleStmt(Forward),
+            utils::unwrapIfSingleStmt(Reverse)};
+  }
+
   StmtDiff ReverseModeVisitor::VisitForStmt(const ForStmt* FS) {
     beginBlock(direction::reverse);
     LoopCounter loopCounter(*this);
--- a/test/Gradient/Loops.C
+++ b/test/Gradient/Loops.C
@@ -2685,6 +2685,138 @@ double fn33(double i, double j) {
 //CHECK-NEXT:    }
 //CHECK-NEXT:}
 
+double fn34(double x, double y){
+  double r = 0;
+  double a[] = {y, x*y, x*x + y};
+  for(auto& i: a){
+    r+=i;
+  }
+  return r;
+}
+
+//CHECK: void fn34_grad(double x, double y, double *_d_x, double *_d_y) {
+//CHECK-NEXT:     unsigned {{int|long}} _t0;
+//CHECK-NEXT:     double (*_d___range1)[3] = 0;
+//CHECK-NEXT:     double (*__range10)[3] = {};
+//CHECK-NEXT:     double *_d___begin1 = 0;
+//CHECK-NEXT:     double *__begin10 = 0;
+//CHECK-NEXT:     clad::tape<double> _t1 = {};
+//CHECK-NEXT:     clad::tape<double *> _t2 = {};
+//CHECK-NEXT:     clad::tape<double *> _t3 = {};
+//CHECK-NEXT:     double _d_r = 0;
+//CHECK-NEXT:     double r = 0;
+//CHECK-NEXT:     double _d_a[3] = {0};
+//CHECK-NEXT:     double a[3] = {y, x * y, x * x + y};
+//CHECK-NEXT:     _t0 = {{0U|0UL}};
+//CHECK-NEXT:     _d___range1 = &_d_a;
+//CHECK-NEXT:     _d___begin1 = *_d___range1;
+//CHECK-NEXT:     __range10 = &a;
+//CHECK-NEXT:     __begin10 = *__range10;
+//CHECK-NEXT:     double *__end10 = *__range10 + {{3|3L}};
+//CHECK-NEXT:     double *_d_i = 0;
+//CHECK-NEXT:     double *i = 0;
+//CHECK-NEXT:     for (; __begin10 != __end10; ++__begin10 , ++_d___begin1) {
+//CHECK-NEXT:         {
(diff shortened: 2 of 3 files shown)


test/Gradient/Loops.C
```diff
@@ -1717,6 +1717,68 @@ double fn22(double param) {
...
+// CHECK-NEXT:     double *__begin10 = 0;
+// CHECK-NEXT:     clad::tape<double *> _t1 = {};
+// CHECK-NEXT:     clad::tape<double *> _t2 = {};
+// CHECK-NEXT:     double *_d_i = 0;
+// CHECK-NEXT:     double *i = {};
+// CHECK-NEXT:     clad::tape<double> _t3 = {};
+// CHECK-NEXT:     double r = 0;
+// CHECK-NEXT:     double a[3] = {y, x * y, x * x + y};
+// CHECK-NEXT:     _t0 = {{0U|0UL}};
+// CHECK-NEXT:     _d___range1 = &_d_a;
```
parth-07, line 1745: Can we use `std::begin` and `std::end` on C-style arrays? If so, then we should probably use these functions instead of explicitly taking the address of the array and using the array size. It will make the functionality reusable and consistent for more complicated types such as `std::vector`, `std::list`, and so on...

ovdiiuv: We could, but the way it is done is more natural, since clad creates all those statements automatically.

parth-07: How does clad creates statements such as `&_d_a` naturally? You are synthesizing the statements by making Clang API calls whether you are creating `&_d_a` or `std::begin(_d_a)`, right?

ovdiiuv: Yeah, that makes sense. Please correct me if I'm wrong, but that would require the support of the iterators in the reverse mode and clad doesn't support that yet. That's an issue I'm currently working on, but clad doesn't yet support custom pullbacks. If so, I suggest we return on that issue once custom pullbacks are supported.

vgvassilev: @parth-07, @ovdiiuv has opened an issue to track that comment so that we can move forward: #1004

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -973,6 +973,82 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    isInsideLoop = true;
+    const auto* Item = FRS->getLoopVariable();
+    Expr* IncBegin = BuildOp(UO_PreInc, BeginExpr);
+    Expr* d_IncBegin = BuildOp(UO_PreInc, d_BeginDeclRef);
+
+    Expr* d_DecBegin = BuildOp(UO_PostDec, d_BeginDeclRef);
+    Expr* ForwardCond = BuildOp(BO_NE, BeginExpr, EndExpr);
+    // Add item assignment statement to the body.
+    auto* body = utils::PrependAndCreateCompoundStmt(
+        m_Sema.getASTContext(), const_cast<Stmt*>(FRS->getBody()),
```
vgvassilev, line 1027: Shouldn’t we clone the body here?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -973,6 +973,82 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+        BuildOp(BO_Add, BuildOp(UO_Deref, RangeExpr),
+                cast<BinaryOperator>(EndDecl->getInit())->getRHS()),
+        false));
+    addToCurrentBlock(AssignEnd);
+    DeclRefExpr* EndExpr =
+        BuildDeclRef(cast<VarDecl>(cast<DeclStmt>(AssignEnd)->getSingleDecl()));
+
+    llvm::SaveAndRestore<bool> SaveIsInsideLoop(isInsideLoop);
+    isInsideLoop = true;
+    const auto* Item = FRS->getLoopVariable();
```
vgvassilev, line 1019:
```suggestion
    VarDecl* Item = FRS->getLoopVariable();
```

ovdiiuv: I don't think that would compile, will try later

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,83 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+
+    LoopCounter loopCounter(*this);
+    const auto* RangeDecl = FRS->getRangeStmt();
+    const auto* BeginDecl = FRS->getBeginStmt();
+    StmtDiff VisitRange = Visit(RangeDecl);
+    StmtDiff VisitBegin = Visit(BeginDecl);
+
+    beginBlock(direction::reverse);
+    // Create all declarations needed.
+    auto* d_BeginDeclRef =
```
vgvassilev, line 991: We should spell out the type here. I think it is VarDecl.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,83 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+        m_Variables[cast<DeclRefExpr>(
+                        cast<BinaryOperator>(VisitBegin.getStmt())->getLHS())
```
vgvassilev, line 993: We should move `VisitBegin.getStmt())->getLHS()` into a temporary variable since it is used more than once.

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: I bet it will, we should also rename `Item` to `LoopVD` or something like that. `VD` is a common acronym for things like that.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,83 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+                        ->getDecl()];
+
+    auto* RangeExpr =
+        cast<DeclRefExpr>(cast<BinaryOperator>(VisitRange.getStmt())->getLHS());
+    auto* BeginExpr =
+        cast<DeclRefExpr>(cast<BinaryOperator>(VisitBegin.getStmt())->getLHS());
+
+    auto* RangeInit = Clone(FRS->getRangeInit());
```
vgvassilev, line 1001: We should describe in a comment, with code example, what are we trying to produce here with the next several blocks of code.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,83 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    Expr* AssignRange =
+        BuildOp(BO_Assign, RangeExpr, BuildOp(UO_AddrOf, RangeInit));
+    Expr* AssignBegin =
+        BuildOp(BO_Assign, BeginExpr, BuildOp(UO_Deref, RangeExpr));
+    addToCurrentBlock(AssignRange);
+    addToCurrentBlock(AssignBegin);
+    const auto* EndDecl = cast<VarDecl>(FRS->getEndStmt()->getSingleDecl());
```
vgvassilev, line 1008: `getEndStmt` returns a `DeclStmt` which might contain a `[DeclGroupRef](https://clang.llvm.org/doxygen/classclang_1_1DeclGroupRef.html)`. That'd be the case for `int i, j`. Are we sure that is not supported in the range-based for loops. If so, we should add an assertion.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,83 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    auto* AssignEnd = BuildDeclStmt(BuildGlobalVarDecl(
+        CloneType(EndDecl->getType()), EndDecl->getNameAsString(),
+        BuildOp(BO_Add, BuildOp(UO_Deref, RangeExpr),
+                cast<BinaryOperator>(EndDecl->getInit())->getRHS()),
```
vgvassilev, line 1012: These things are better as temporary variables.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,83 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+
+    auto* RangeInit = Clone(FRS->getRangeInit());
+    Expr* AssignRange =
+        BuildOp(BO_Assign, RangeExpr, BuildOp(UO_AddrOf, RangeInit));
+    Expr* AssignBegin =
+        BuildOp(BO_Assign, BeginExpr, BuildOp(UO_Deref, RangeExpr));
+    addToCurrentBlock(AssignRange);
+    addToCurrentBlock(AssignBegin);
+    const auto* EndDecl = cast<VarDecl>(FRS->getEndStmt()->getSingleDecl());
+    auto* AssignEnd = BuildDeclStmt(BuildGlobalVarDecl(
```
vgvassilev, line 1009: Please spell out the type here.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,83 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    isInsideLoop = true;
+    const auto* Item = FRS->getLoopVariable();
+    Expr* IncBegin = BuildOp(UO_PreInc, BeginExpr);
+    Expr* d_IncBegin = BuildOp(UO_PreInc, d_BeginDeclRef);
+
+    Expr* d_DecBegin = BuildOp(UO_PostDec, d_BeginDeclRef);
+    Expr* ForwardCond = BuildOp(BO_NE, BeginExpr, EndExpr);
+    // Add item assignment statement to the body.
+    auto* body = utils::PrependAndCreateCompoundStmt(
+        m_Sema.getASTContext(), Clone(FRS->getBody()),
```
vgvassilev, line 1028: The clone operations are better of outside the calls. We need to spell out the type here, too.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,83 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+        BuildDeclStmt(const_cast<VarDecl*>(Item)));
+
+    StmtDiff BodyDiff =
+        DifferentiateLoopBody(body, loopCounter, nullptr, d_DecBegin,
+                              /*isForLoop=*/true);
+    Stmt* Forward =
+        new (m_Context) ForStmt(m_Context, nullptr, ForwardCond, nullptr,
+                                BuildOp(BO_Comma, IncBegin, d_IncBegin),
+                                BodyDiff.getStmt(), noLoc, noLoc, noLoc);
```
vgvassilev, line 1037: We have a new policy: we try to pass the source locations of the original function instead of `noLoc`.

vgvassilev (commented):
Thanks for working on this. Seems like it is heading in the right direction.

lib/Differentiator/ReverseModeVisitor.cpp
ovdiiuv: It's always a single declaration.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,85 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+
+    Expr* d_DecBegin = BuildOp(UO_PostDec, d_BeginDeclRef);
+    Expr* ForwardCond = BuildOp(BO_NE, BeginDeclRef, EndExpr);
+    // Add item assignment statement to the body.
+    Stmt* bodyClone = Clone(FRS->getBody());
+    CompoundStmt* body = utils::PrependAndCreateCompoundStmt(
+        m_Sema.getASTContext(), bodyClone,
+        BuildDeclStmt(const_cast<VarDecl*>(LoopVD)));
+    StmtDiff BodyDiff =
+        DifferentiateLoopBody(body, loopCounter, nullptr, d_DecBegin,
```
vgvassilev, line 1033:
```suggestion
        DifferentiateLoopBody(body, loopCounter, /*condVarDifff=*/nullptr, d_DecBegin,
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,85 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+                              /*isForLoop=*/true);
+
+    Expr* Inc = BuildOp(BO_Comma, IncBegin, d_IncBegin);
+    Stmt* Forward = new (m_Context)
+        ForStmt(m_Context, nullptr, ForwardCond, nullptr, Inc,
+                BodyDiff.getStmt(), FRS->getForLoc(), noLoc, noLoc);
```
vgvassilev, line 1039:
```suggestion
                BodyDiff.getStmt(), FRS->getForLoc(), FRS->getBeginLoc(), FRS->getEndLoc());
```

We also need to add comments for the `nullptr`s on the line above.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,85 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    Expr* CounterCondition =
+        loopCounter.getCounterConditionResult().get().second;
+    Expr* CounterDecrement = loopCounter.getCounterDecrement();
+    Stmt* Reverse = BodyDiff.getStmt_dx();
+    addToCurrentBlock(Reverse, direction::reverse);
+    Reverse = endBlock(direction::reverse);
+    Reverse = new (m_Context)
+        ForStmt(m_Context, nullptr, CounterCondition, nullptr, CounterDecrement,
+                Reverse, FRS->getForLoc(), noLoc, noLoc);
```
vgvassilev, line 1048: Likewise.

test/Gradient/Loops.C
```diff
@@ -2732,6 +2732,68 @@ double fn33(double i, double j) {
 //CHECK-NEXT:    }
 //CHECK-NEXT:}
 
+double fn34(double x, double y){
+  double r = 0;
+  double a[] = {y, x*y, x*x + y};
+  for(auto& i: a){
```
vgvassilev, line 2738: Maybe we can add a separate test case like
```cpp
for (double i = 1, j = 2; auto i : a) {
        sum_plus_offset += v + i*j;
    }
```

ovdiiuv: I didn't add the support for the initializer statement, I could open an issue on that.

vgvassilev: Ok, please do.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,87 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    CompoundStmt* body = utils::PrependAndCreateCompoundStmt(
+        m_Sema.getASTContext(), bodyClone,
+        BuildDeclStmt(const_cast<VarDecl*>(LoopVD)));
+    StmtDiff BodyDiff = DifferentiateLoopBody(
+        body, loopCounter, /*condVarDiff*/ nullptr, d_DecBegin,
+        /*isForLoop=*/true);
+
+    Expr* Inc = BuildOp(BO_Comma, IncBegin, d_IncBegin);
+    Stmt* Forward = new (m_Context)
+        ForStmt(m_Context, /*Init*/ nullptr, ForwardCond, /*CondVar*/ nullptr,
```
vgvassilev, line 1038:
```suggestion
        ForStmt(m_Context, /*Init=*/nullptr, ForwardCond, /*CondVar=*/nullptr,
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,87 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+                FRS->getEndLoc());
+    Expr* CounterCondition =
+        loopCounter.getCounterConditionResult().get().second;
+    Expr* CounterDecrement = loopCounter.getCounterDecrement();
+    Stmt* Reverse = BodyDiff.getStmt_dx();
+    addToCurrentBlock(Reverse, direction::reverse);
+    Reverse = endBlock(direction::reverse);
+    Reverse = new (m_Context)
+        ForStmt(m_Context, /*Init*/ nullptr, CounterCondition,
+                /*CondVar*/ nullptr, CounterDecrement, Reverse,
```
vgvassilev, line 1049:
```suggestion
        ForStmt(m_Context, /*Init=*/nullptr, CounterCondition,
                /*CondVar=*/nullptr, CounterDecrement, Reverse,
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,87 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    StmtDiff VisitBegin = Visit(BeginDecl);
+    Expr* BeginExpr = cast<BinaryOperator>(VisitBegin.getStmt())->getLHS();
+
+    beginBlock(direction::reverse);
+    // Create all declarations needed.
+    Expr* d_BeginDeclRef = m_Variables[cast<DeclRefExpr>(BeginExpr)->getDecl()];
+
+    auto* RangeExpr =
+        cast<DeclRefExpr>(cast<BinaryOperator>(VisitRange.getStmt())->getLHS());
+    auto* BeginDeclRef = cast<DeclRefExpr>(BeginExpr);
```
vgvassilev, line 996: Let's move that a few lines above to avoid the `cast<DeclRefExpr>`.

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: ping.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,87 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    QualType EndType = CloneType(EndDecl->getType());
+    std::string EndName = EndDecl->getNameAsString();
+    Expr* EndAssign = BuildOp(BO_Add, BuildOp(UO_Deref, RangeExpr), EndInit);
+    VarDecl* EndVarDecl =
+        BuildGlobalVarDecl(EndType, EndName, EndAssign, false);
+    DeclStmt* AssignEnd = BuildDeclStmt(EndVarDecl);
+
+    addToCurrentBlock(AssignEnd);
+    DeclRefExpr* EndExpr =
+        BuildDeclRef(cast<VarDecl>(cast<DeclStmt>(AssignEnd)->getSingleDecl()));
```
vgvassilev, line 1017: Here it's worth moving the long expression to some temporary variables.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,87 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+
+    llvm::SaveAndRestore<bool> SaveIsInsideLoop(isInsideLoop);
+    isInsideLoop = true;
```
vgvassilev, line 1020:
```suggestion
    llvm::SaveAndRestore<bool> SaveIsInsideLoop(isInsideLoop, /*NewValue=*/true);
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,85 @@
...
+    Expr* IncBegin = BuildOp(UO_PreInc, BeginDeclRef);
+    Expr* d_IncBegin = BuildOp(UO_PreInc, d_BeginDeclRef);
+
+    Expr* d_DecBegin = BuildOp(UO_PostDec, d_BeginDeclRef);
+    Expr* ForwardCond = BuildOp(BO_NE, BeginDeclRef, EndExpr);
+    // Add item assignment statement to the body.
+    Stmt* bodyClone = Clone(FRS->getBody());
+    CompoundStmt* body = utils::PrependAndCreateCompoundStmt(
+        m_Sema.getASTContext(), bodyClone,
+        BuildDeclStmt(const_cast<VarDecl*>(LoopVD)));
```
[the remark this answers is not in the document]
vgvassilev: We should clone that variable declaration, too.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,87 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    StmtDiff BodyDiff = DifferentiateLoopBody(
+        body, loopCounter, /*condVarDiff*/ nullptr, d_DecBegin,
```
vgvassilev, line 1033:
```suggestion
        body, loopCounter, /*condVarDiff=*/nullptr, d_DecBegin,
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -974,6 +974,88 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    addToCurrentBlock(AssignRange);
+    addToCurrentBlock(AssignBegin);
+    const auto* EndDecl = cast<VarDecl>(FRS->getEndStmt()->getSingleDecl());
+
+    Expr* EndInit = cast<BinaryOperator>(EndDecl->getInit())->getRHS();
+    QualType EndType = CloneType(EndDecl->getType());
+    std::string EndName = EndDecl->getNameAsString();
+    Expr* EndAssign = BuildOp(BO_Add, BuildOp(UO_Deref, RangeExpr), EndInit);
+    VarDecl* EndVarDecl =
+        BuildGlobalVarDecl(EndType, EndName, EndAssign, false);
```
vgvassilev, line 1012:
```suggestion
        BuildGlobalVarDecl(EndType, EndName, EndAssign, /*DirectInit=*/false);
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -976,6 +976,131 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    Expr* LoopInit = LoopVDDiff.getDecl()->getInit();
+    LoopVDDiff.getDecl()->setInit(getZeroInit(LoopVDDiff.getDecl()->getType()));
+    addToCurrentBlock(BuildDeclStmt(LoopVDDiff.getDecl()));
+    Expr* AssignLoop =
+        BuildOp(BO_Assign, BuildDeclRef(LoopVDDiff.getDecl()), LoopInit);
+
+    if (!LoopVD->getType()->isReferenceType()) {
+      Expr* d_LoopVD = BuildDeclRef(LoopVDDiff.getDecl_dx());
+      AdjLoopVDAddAssign =
+          BuildOp(BO_Assign, d_LoopVD, BuildOp(UO_Deref, d_BeginDeclRef));
```
vgvassilev, line 1059: Can we add a test case to make codecov happy?

vgvassilev (approved):
LGTM!

PR #981, opened 2024-07-12
Use a reference to store the LHS of assignments

### LINKED ISSUE #759: Remove GetInnermostReturnExpr
To know all possible L-values of the LHS of an assignment operation in the reverse mode, we call ``utils::GetInnermostReturnExpr``. We use it to know the variables that we should store and restore in assignments with non-trivial LHS (e.g. ``(a = b) = c;``, ``(cond ? a : b) = c;``). We should instead store and restore the reference to the LHS (which we currently have anyway). For example, we should differentiate ``(a = b) = c;`` as
```
double _t0 = a;
double &_t1 = (a = b);
double _t2 = _t1;
_t1 = c;
...
_t1 = _t2;
...
a = _t0;
```
instead of
```
double _t0 = a;
double &_t2 = (a = b);
double _t1 = a;
_t2 = c;
...
a = _t1;
...
a = _t0;
```
This way we will:
1) Remove GetInnermostReturnExpr.
2) Store fewer variables in cases like ``(cond ? a : b) = c;``, where we currently store both ``a`` and ``b``.
3) Allow function calls on the LHS.

PetroZarytskyi:
This PR is created to fix #759.

### DIFF
--- a/test/Gradient/Assignments.C
+++ b/test/Gradient/Assignments.C
@@ -414,20 +414,19 @@ double f9(double x, double y) {
 //CHECK:   void f9_grad(double x, double y, double *_d_x, double *_d_y) {
 //CHECK-NEXT:       double _d_t = 0;
 //CHECK-NEXT:       double _t0;
-//CHECK-NEXT:       double *_t1;
-//CHECK-NEXT:       double _t2;
+//CHECK-NEXT:       double _t1;
 //CHECK-NEXT:       double t = x;
 //CHECK-NEXT:       _t0 = t;
-//CHECK-NEXT:       _t1 = &(t *= x);
-//CHECK-NEXT:       _t2 = *_t1;
-//CHECK-NEXT:       *_t1 *= y;
+//CHECK-NEXT:       (t *= x);
+//CHECK-NEXT:       _t1 = t;
+//CHECK-NEXT:       t *= y;
 //CHECK-NEXT:       _d_t += 1;
 //CHECK-NEXT:       {
-//CHECK-NEXT:           *_t1 = _t2;
+//CHECK-NEXT:           t = _t1;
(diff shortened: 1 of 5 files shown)


test/Gradient/Assignments.C
```diff
@@ -414,19 +414,20 @@ double f9(double x, double y) {
...
 //CHECK-NEXT:       double _d_t = 0;
 //CHECK-NEXT:       double _t0;
+//CHECK-NEXT:       double *_t1;
 //CHECK-NEXT:       double _t2;
 //CHECK-NEXT:       double t = x;
 //CHECK-NEXT:       _t0 = t;
-//CHECK-NEXT:       double &_t1 = (t *= x);
-//CHECK-NEXT:       _t2 = t;
-//CHECK-NEXT:       _t1 *= y;
+//CHECK-NEXT:       _t1 = &(t *= x);
```
vgvassilev, line 421: Is that right? Aren't we taking the address of a temporary that gets destroyed at the end of the expression?

PetroZarytskyi: The value of ``t *= x`` is a reference to ``t``. So ``&(t *= x)`` should give the address of ``t``. Anyway, we used to declare a reference to the same expression and references are pointers under the hood. So this change shouldn't affect the performance or behavior.

vgvassilev: Ah, okay, I see. Thanks!

vgvassilev:
@PetroZarytskyi, https://github.com/vgvassilev/clad/pull/971 is merged. Can you rebase?

vgvassilev (approved):
LGTM!

vgvassilev:
I believe the codecov complaint is because we removed more code and it thinks the coverage is reduced...

PR #982, opened 2024-07-16
Fix pointer arithmetic in fwd mode

### LINKED ISSUE #978: Pointer arithmetics don't work in forward mode
This is similar to #972, only now it's about pointer arithmetic that backward mode supports but not the forward mode.

It would be nice if backward mode and forward mode have the same level of C++ language support to avoid more surprises like this :slightly_smiling_face:

Reproducer:
```c++
// Compile with clang++ -std=c++11 -fplugin=install/lib/clad.so -I./install/include cladIssue.cxx -o cladIssue

#include <iostream>
#include <vector>

double wrapper(double* params, double const* constants) {
   double c0 = *(constants + 0);
   return params[0] * c0;
}

int main()
{

   double x = 2.0;
   double y = 2.0;
   double result = 0.0;
   grad.execute(&x, &y, &result);

   std::cout << diff.execute(&x, &y) << std::endl;
}
```
Compiler output:
```txt
cladIssue.cxx:9:30: warning: attempt to differentiate unsupported operator,  derivative                          set to 0
   double c0 = *(constants + 0);
                             ^
error: indirection requires pointer operand ('int' invalid)
1 warning and 1 error generated.
```

vaithak:
fixes #978

### DIFF
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -1460,13 +1460,13 @@ BaseForwardModeVisitor::VisitBinaryOperator(const BinaryOperator* BinOp) {
     // including it as a Stmt_dx. Moreover, the fact that Stmt_dx is left
     // nullptr is used for treating expressions like ((A && B) && C) correctly.
     return StmtDiff(opDiff, nullptr);
-  }
-  if (!opDiff) {
+  } else {
     // FIXME: add support for other binary operators
     unsupportedOpWarn(BinOp->getEndLoc());
     opDiff = ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
   }
-  opDiff = folder.fold(opDiff);
+  if (opDiff)
+    opDiff = folder.fold(opDiff);
   // Recover the original operation from the Ldiff and Rdiff instead of
   // cloning the tree.
   Expr* op;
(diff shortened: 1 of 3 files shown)


lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1501,9 +1501,17 @@ BaseForwardModeVisitor::DifferentiateVarDecl(const VarDecl* VD,
...
+  if (VD->getType()->isPointerType() && !initDx) {
+    // initialize with nullptr.
+    // NOLINTBEGIN(cppcoreguidelines-owned-memory)
+    initDx =
+        new (m_Context) CXXNullPtrLiteralExpr(VD->getType(), VD->getBeginLoc());
+    // NOLINTEND(cppcoreguidelines-owned-memory)
+  }
+  VarDecl* VDDerived =
+      BuildVarDecl(VD->getType(), "_d_" + VD->getNameAsString(), initDx,
+                   VD->isDirectInit(), nullptr, VD->getInitStyle());
```
vgvassilev, line 1514:
```suggestion
                   VD->isDirectInit(), /*TSI=*/nullptr, VD->getInitStyle());
```

vaithak:
I am working on one more fix related to handling function calls with pointers, will send a separate PR for that as it takes a bit more debugging. The issue is essentially related to #762, but in forward mode.

PR #986, opened 2024-07-18
Fix type casting of ValueAndPushForward types

### LINKED ISSUE #983: Functions that return integers can't be used in forward mode
Functions that return integers can't be used in forward mode (in backwards mode it works). This pattern is used a lot in binned likelihoods.

Note that when trying to do the same in the ROOT interpreter, there is a segfault instead of the error message.

Reproducer:
```c++
// Compile with clang++ -std=c++11 -fplugin=install/lib/clad.so -I./install/include cladIssue.cxx -o cladIssue

inline unsigned int getBin(double low, double high, double val, unsigned int numBins)
{
   double binWidth = (high - low) / numBins;
   return val >= high ? numBins - 1 : std::abs((val - low) / binWidth);
}

double wrapper(double *params, double const *obs, double const *xlArr)
{
   const double t116 = *(xlArr + getBin(0., 1., obs[0], 1));
   return t116 * params[0];
}

int main()
{
}
```
Output of the compiler:
```txt
cladIssue.cxx:5:21: error: type 'double' cannot be narrowed to 'unsigned int' in initializer list [-Wc++11-narrowing]
inline unsigned int getBin(double low, double high, double val, unsigned int numBins)
                    ^~~~~~
cladIssue.cxx:5:21: note: insert an explicit cast to silence this issue
inline unsigned int getBin(double low, double high, double val, unsigned int numBins)
                    ^~~~~~
                    static_cast<unsigned int>( )
cladIssue.cxx:5:21: error: type 'double' cannot be narrowed to 'unsigned int' in initializer list [-Wc++11-narrowing]
inline unsigned int getBin(double low, double high, double val, unsigned int numBins)
                    ^~~~~~
cladIssue.cxx:5:21: note: insert an explicit cast to silence this issue
inline unsigned int getBin(double low, double high, double val, unsigned int numBins)
                    ^~~~~~
                    static_cast<unsigned int>( )
2 errors generated.
```

vaithak:
Fixes #983

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -20,6 +20,14 @@ namespace clad {
 template <typename T, typename U> struct ValueAndPushforward {
   T value;
   U pushforward;
+
+  // Define the cast operator from ValueAndPushforward<T, U> to
+  // ValueAndPushforward<V, w> where V is convertible to T and W is
+  // convertible to U.
+  template <typename V = T, typename W = U>
+  operator ValueAndPushforward<V, W>() const {
+    return {static_cast<V>(value), static_cast<W>(pushforward)};
+  }
 };
 
 /// It is used to identify constructor custom pushforwards. For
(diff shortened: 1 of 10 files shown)


vgvassilev (approved):
LGTM!

Issue #988, opened 2024-07-18
Void function calls with literal arguments crash Clad in forward mode


gojakuch:
```c++
#include "clad/Differentiator/Differentiator.h"
#include <iostream>

void some_important_void_func(double y) {
}

double f(double x) {
  some_important_void_func(0);
  return x;
}

int main() {
  auto df = clad::differentiate(f, 0);
  df.dump();
  std::cout << "f(3)=" << f(3.0) << "; df(3)=" << df.execute(3.0) << std::endl;
  return 0;
}
```

PR #989, opened 2024-07-18
Prevent Clad from trying to create a void zero literal

gojakuch:
Previously, clad used to try to create a void zero literal when differentiating a call to a void function with literal arguments in the forward mode. This caused it to crash.

Fixes: #988

### DIFF
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -1216,8 +1216,7 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
                                validLoc, llvm::MutableArrayRef<Expr*>(CallArgs),
                                validLoc)
                 .get();
-        auto* zero = ConstantFolder::synthesizeLiteral(CE->getType(), m_Context,
-                                                       /*val=*/0);
+        auto* zero = getZeroInit(CE->getType());
         return StmtDiff(call, zero);
       }
     }
--- a/test/FirstDerivative/CallArguments.C
+++ b/test/FirstDerivative/CallArguments.C
@@ -148,7 +148,7 @@ float f_literal_args_func(float x, float y, float *z) {
 // CHECK-NEXT: float _d_y = 0;
 // CHECK-NEXT: printf("hello world ");
 // CHECK-NEXT: float _t0 = f_literal_helper(0.5, 'a', z, nullptr);
-// CHECK-NEXT: return _d_x * _t0 + x * 0.F;
+// CHECK-NEXT: return _d_x * _t0 + x * 0;
 // CHECK-NEXT: }
 
 inline unsigned int getBin(double low, double high, double val, unsigned int numBins) {
--- a/test/FirstDerivative/FunctionCalls.C
+++ b/test/FirstDerivative/FunctionCalls.C
@@ -184,7 +184,7 @@ double test_9(double x) {
 // CHECK-NEXT: }
 
 void some_important_void_func(double y) {
-    assert(y < 1);
+    assert(y >= 1);
 }
 
 double test_10(double x) {
--- a/test/FirstDerivative/FunctionCalls.C
+++ b/test/FirstDerivative/FunctionCalls.C
@@ -183,6 +183,21 @@ double test_9(double x) {
 // CHECK-NEXT: return _t0.pushforward;
 // CHECK-NEXT: }
 
+void some_important_void_func(double y) {
+    assert(y < 1);
+}
+
+double test_10(double x) {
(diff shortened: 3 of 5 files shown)


gojakuch:
@vgvassilev this should prevent #977 from failing the kokkos check

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1219,9 +1219,15 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
                                validLoc, llvm::MutableArrayRef<Expr*>(CallArgs),
                                validLoc)
                 .get();
-        auto* zero = ConstantFolder::synthesizeLiteral(CE->getType(), m_Context,
-                                                       /*val=*/0);
-        return StmtDiff(call, zero);
+        if (!CE->getType().getTypePtr()->isVoidType()) {
```
vgvassilev, line 1222:
```suggestion
        if (!CE->getType()->isVoidType()) {
```

vgvassilev: On a second thought can we use `VisitorBase::getZeroInit` and teach it to return `nullptr` if we ask to get the zero init of a void type?

gojakuch:
@vgvassilev one check still fails to set up the Ubuntu apt sources

test/FirstDerivative/CallArguments.C
```diff
@@ -148,7 +148,7 @@ float f_literal_args_func(float x, float y, float *z) {
 // CHECK-NEXT: float _d_y = 0;
 // CHECK-NEXT: printf("hello world ");
 // CHECK-NEXT: float _t0 = f_literal_helper(0.5, 'a', z, nullptr);
-// CHECK-NEXT: return _d_x * _t0 + x * 0.F;
```
vgvassilev, line 151: We will need to check if it is a double of float type to retain that type in the literal too. We have to update VisitorBase::getZeroInit with the relevant code that takes into account if the type is a float or a double.

gojakuch: `VisitorBase::getZeroInit` currently only synthesises integer 0 if it's a scalar numerical type and it actually converts it to the correct type implicitly. if I edit this, I'll need to change almost every 0 entry in 37 different tests manually and I don't think it's worth the time, as I don't see what it brings. like, it only generates integer zeroes for the types for which it's safe to do this.

vgvassilev: It’s worth changing because sometimes the produced code cannot compile.

test/FirstDerivative/FunctionCalls.C
```diff
@@ -183,6 +183,21 @@ double test_9(double x) {
 // CHECK-NEXT: return _t0.pushforward;
 // CHECK-NEXT: }
 
+void some_important_void_func(double y) {
+    assert(y < 1);
```
parth-07, line 187: The value of `y` is `1` in the `test_10` function call. The `assert` call can fail.

parth-07 (changes requested):

test/FirstDerivative/CallArguments.C
gojakuch: I cannot think of a situation where a float or double value couldn't be created from `0`. but even if that's my mistake, I'd suggest opening this as another issue if this is worth fixing, because this is not something I've changed, it's been like that before. changing all the tests for this to work out would require some automation, because it takes an unreasonable amount of time if done manually, I think this PR needs to be merged faster than that. I've already spent half an hour on a single file file out of 37 and I'm not even half way through this file. some of the test files are short, but I'd rather have this as a separate PR and then I would automate these changes. again, this PR did not make this change, it's how Clad works on master in most cases, this PR only affects the special cases of when a zero comes from a 'constant function call'.

vgvassilev: One can claim that `0.f` was correct and that PR regressed it. We can argue about that at length but I guess that something you will need to get used to when developing functionality in the core of clad.

If you say that we should merge that with a regression that's fine but then immediately after that you will have to work on the regression which does not save you a lot, right?

gojakuch: well, in that case, I can stick to the less elegant issue of having this being called for void function calls only, not changing the behaviour for other cases for now, but I can add a comment there.

vgvassilev: I guess we can move forward with your approach as long as you open an issue and describe what needs to be done there.

gojakuch: alright, I'll open an issue then

parth-07 (approved):
Looks good.

Issue #829, opened 2024-03-19
Using std::array inside functions to be differentiated?


a-jp:
Hi,

Can I use std::array to hold data that is accessed by functions that are differentiated?  

```c++
double f(double x, double y)
{
  double a[2];
  a[0] = 5;
  a[1] = 4;
  //std::array<double, 2> a{{5, 4}}; // fails
  return a[0] * x * y;
}
```

Thanks

vaithak:
Hi @a-jp, std::array is not yet supported in clad. We are aware of this, along with many other stl features and plan to add support for them in the near future.

a-jp:
Ok thanks

vgvassilev:
I'd like to keep this issue open since supporting `std::array` is an immediate goal for the project.

PR #992, opened 2024-07-19
Add support for `std::array` in the fwd mode

gojakuch:
Fixes: #829

### DIFF
--- a/include/clad/Differentiator/STLBuiltins.h
+++ b/include/clad/Differentiator/STLBuiltins.h
@@ -128,6 +128,80 @@ operator_subscript_pushforward(const ::std::vector<T>* v, unsigned idx,
   return {(*v)[idx], (*d_v)[idx]};
 }
 
+template <typename T, ::std::size_t N>
+constexpr clad::ValueAndPushforward<T&, T&>
+operator_subscript_pushforward(::std::array<T, N>* a, ::std::size_t i,
+                               ::std::array<T, N>* d_a,
+                               ::std::size_t d_i) noexcept {
+  return {(*a)[i], (*d_a)[i]};
+}
+
+template <typename T, ::std::size_t N>
+constexpr clad::ValueAndPushforward<T&, T&>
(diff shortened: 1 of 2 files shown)


vgvassilev:
Can we add support in reverse mode before closing the issue?

gojakuch:
> Can we add support in reverse mode before closing the issue?

I'll see if it's possible. should be.

parth-07:
> I'll see if it's possible. should be.

It would be difficult because we don't support custom derivatives of member functions and constructors in the reverse mode...

parth-07:
@vgvassilev A long time ago, you suggested creating a separate repository to maintain custom derivatives. I liked that idea because custom derivatives can grow a lot. What do you think about that idea now?

vgvassilev:
> @vgvassilev A long time ago, you suggested creating a separate repository to maintain custom derivatives. I liked that idea because custom derivatives can grow a lot. What do you think about that idea now?

Yes, however, the challenge would be keeping it in sync with the changes of clad and making it more difficult for clad to introduce backward incompatible changes. That being said we should do this but probably for libraries like Stl and a few others that represent major usecases for us we should keep those in the repo…

parth-07:
> Yes, however, the challenge would be keeping it in sync with the changes of clad and making it more difficult for clad to introduce backward incompatible changes.

Yes, agreed...

> That being said we should do this but probably for libraries like Stl and a few others that represent major usecases for us we should keep those in the repo…

Okay, this sounds good.

gojakuch:
> > I'll see if it's possible. should be.
> 
> It would be difficult because we don't support custom derivatives of member functions and constructors in the reverse mode...

oh, for some reason I thought this works. not so possible to hit this in this PR then

vgvassilev:
> > > I'll see if it's possible. should be.
> > 
> > 
> > It would be difficult because we don't support custom derivatives of member functions and constructors in the reverse mode...
> 
> oh, for some reason I thought this works. not so possible to hit this in this PR then

Then open another issue for std::array support in reverse mode then.

vgvassilev (approved):
Lgtm!

PR #997, opened 2024-07-21
Directly set the adjoint of the LHS to 0 in assignments

PetroZarytskyi:
Currently, when we set the LHS of an assignment to 0 in the reverse mode, we do this by
```
float _r_d1 = *_d_y;
*_d_y -= _r_d1;
```
This PR simplifies this process to
```
float _r_d1 = *_d_y;
*_d_y = 0;
```

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2357,9 +2357,12 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         oldValue = StoreAndRef(AssignedDiff, direction::reverse, "_r_d",
                                /*forceDeclCreation=*/true);
       if (opCode == BO_Assign) {
-        // Add the statement `dl -= oldValue;`
-        addToCurrentBlock(BuildOp(BO_SubAssign, AssignedDiff, oldValue),
-                          direction::reverse);
+        if (!isPointerOp) {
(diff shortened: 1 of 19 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2357,9 +2357,12 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         oldValue = StoreAndRef(AssignedDiff, direction::reverse, "_r_d",
                                /*forceDeclCreation=*/true);
       if (opCode == BO_Assign) {
-        // Add the statement `dl -= oldValue;`
-        addToCurrentBlock(BuildOp(BO_SubAssign, AssignedDiff, oldValue),
-                          direction::reverse);
+        if (!isPointerOp) {
```
vgvassilev, line 2360: Out of curiosity, which are the pointer operations in which we should the old strategy?

PetroZarytskyi: I didn't understand why pointers didn't work at first. The thing here is that for pointers, ``oldValue`` is ``nullptr`` and ``BuildOp(...)`` returns ``nullptr`` as well. Because of that, ``addToCurrentBlock(...)`` doesn't add anything to the block. So this line didn't do anything for pointers. But the new one does so I had to add a condition.

vgvassilev (approved):
LGTM!

PR #998, opened 2024-07-22
Try to improve literal types

vaithak:
This is related to the review comments of PR #975.


lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1008,8 +1008,11 @@ StmtDiff BaseForwardModeVisitor::VisitDeclRefExpr(const DeclRefExpr* DRE) {
   // If DRE is of type pointer, then the derivative is a null pointer.
   if (clonedDRE->getType()->isPointerType())
     return StmtDiff(clonedDRE, nullptr);
+  QualType literalTy = utils::GetValueType(clonedDRE->getType());
```
vgvassilev, line 1011: We should extend visitorbase::getZeroInit as discussed here https://github.com/vgvassilev/clad/pull/989#discussion_r1684487454

cc: @gojakuch

vgvassilev:
This pull request, if the review comment is addressed, should fix #1001.

kchristin22:
The cuda bug was a result of not handling the enum types correctly. `clad::utils::GetValueType()` doesn't return the underlying type of the enum but the enum itself (as a type). When checking `isRealType()` for that node, the cudaError_t enum passes the test so it's type is not set as an Int. This lead to a compilation error down the line. 

I fixed that part in `clad::utils::GetValueType()` in a branch based on Vaibhav's `cuda-bug` branch. Should I open a new PR and reference this one and continue addressing the comments of this and its previous PR there?

vgvassilev:
> The cuda bug was a result of not handling the enum types correctly. `clad::utils::GetValueType()` doesn't return the underlying type of the enum but the enum itself (as a type). When checking `isRealType()` for that node, the cudaError_t enum passes the test so it's type is not set as an Int. This lead to a compilation error down the line.
> 
> I fixed that part in `clad::utils::GetValueType()` in a branch based on Vaibhav's `cuda-bug` branch. Should I open a new PR and reference this one and continue addressing the comments of this and its previous PR there?

Can you push to this PR?

kchristin22:
> > The cuda bug was a result of not handling the enum types correctly. `clad::utils::GetValueType()` doesn't return the underlying type of the enum but the enum itself (as a type). When checking `isRealType()` for that node, the cudaError_t enum passes the test so it's type is not set as an Int. This lead to a compilation error down the line.
> > I fixed that part in `clad::utils::GetValueType()` in a branch based on Vaibhav's `cuda-bug` branch. Should I open a new PR and reference this one and continue addressing the comments of this and its previous PR there?
> 
> Can you push to this PR?

Sure, but I have no permissions in Vaibhav's repo, @vaithak could you change it temporarily(till this PR is merged)?

vgvassilev:
If you accept my invite you should be able to push to this PR.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1070,8 +1070,11 @@ StmtDiff BaseForwardModeVisitor::VisitDeclRefExpr(const DeclRefExpr* DRE) {
...
+  if (!literalTy->isRealType())
```
vgvassilev, line 1074: Why do we need this special check here?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1374,8 +1377,12 @@ StmtDiff BaseForwardModeVisitor::VisitUnaryOperator(const UnaryOperator* UnOp) {
   } else if (opKind == UnaryOperatorKind::UO_Deref) {
     if (Expr* dx = diff.getExpr_dx())
       return StmtDiff(op, BuildOp(opKind, dx));
-    return StmtDiff(op, ConstantFolder::synthesizeLiteral(
-                            m_Context.IntTy, m_Context, /*val=*/0));
+    QualType literalTy =
+        utils::GetValueType(UnOp->getSubExpr()->getType()->getPointeeType());
+    if (!literalTy->isRealType())
+      literalTy = m_Context.IntTy;
```
vgvassilev, line 1383: I think we should not need this. This gets handled in `ConstantFolder::synthesizeLiteral` anyways.

lib/Differentiator/VisitorBase.cpp
```diff
@@ -379,11 +379,10 @@ namespace clad {
     // FIXME: Consolidate other uses of synthesizeLiteral for creation 0 or 1.
     if (T->isVoidType())
       return nullptr;
-    if (T->isScalarType()) {
+    if (T->isScalarType() && !T->isPointerType()) {
```
vgvassilev, line 382: Why do we need this here? Shouldn't `ConstantFolder::synthesizeLiteral` check if T is a pointer type and synthesize `nullptr`?

lib/Differentiator/VisitorBase.cpp
```diff
@@ -379,9 +379,8 @@
...
     if (T->isVoidType())
       return nullptr;
-    if (T->isScalarType()) {
-      ExprResult Zero =
-          ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
+    else if (T->isScalarType() && !T->isPointerType()) {
+      ExprResult Zero = ConstantFolder::synthesizeLiteral(T, m_Context, 0);
       CastKind CK = m_Sema.PrepareScalarCast(Zero, T);
       return m_Sema.ImpCastExprToType(Zero.get(), T, CK).get();
     }
```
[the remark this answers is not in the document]
vgvassilev: That's a good suggestion.

lib/Differentiator/BaseForwardModeVisitor.cpp
kchristin22: True, it was a leftover from Vaibhav's PR. I will remove these lines.

lib/Differentiator/VisitorBase.cpp
kchristin22: Right now, the pointers are initialized as an empty list, which is equivalent to `nullptr`, the same way clad tapes are initialized. I can alter them to be equal to `nullptr` if that's safer.

vgvassilev: Yes, that'd be more readable, too.

lib/Differentiator/VisitorBase.cpp
```diff
@@ -379,11 +379,10 @@ namespace clad {
...
-    if (T->isScalarType()) {
+    if ((T->isScalarType() || T->isPointerType()) && !T->isReferenceType()) {
       ExprResult Zero =
-          ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
-      CastKind CK = m_Sema.PrepareScalarCast(Zero, T);
-      return m_Sema.ImpCastExprToType(Zero.get(), T, CK).get();
+          ConstantFolder::synthesizeLiteral(T, m_Context, /*val=*/0);
+      return Zero.get();
     }
     return m_Sema.ActOnInitList(noLoc, {}, noLoc).get();
```
vgvassilev, line 387: We should probably sink `ActOnInitList` into `synthesizeLiteral` but let's do it outside of this pr.

test/Gradient/FunctionCalls.C
```diff
@@ -323,8 +323,8 @@ double fn8(double x, double y) {
 // CHECK-NEXT:     double _t1 = std::tanh(1.);
 // CHECK-NEXT:     double _t0 = std::max(1., 2.);
 // CHECK-NEXT:     {
-// CHECK-NEXT:         double _r0 = 0;
-// CHECK-NEXT:         char _r1 = 0;
+// CHECK-NEXT:         double _r0 = 0.;
+// CHECK-NEXT:         char _r1 = 0i8;
```
vgvassilev, line 327: That will fail to compile I think.

test/Gradient/STLCustomDerivatives.C
```diff
@@ -272,94 +272,94 @@ int main() {
 // CHECK-NEXT:     {
 // CHECK-NEXT:         res = _t25;
 // CHECK-NEXT:         double _r_d6 = _d_res;
-// CHECK-NEXT:         {{.*}} _r10 = 0;
+// CHECK-NEXT:         size_type _r10 = {{0U|0UL}};
 // CHECK-NEXT:         {{.*}}class_functions::operator_subscript_pullback(&_t26, 0, _r_d6, &_d_vec, &_r10);
-// CHECK-NEXT:         {{.*}} _r11 = 0;
+// CHECK-NEXT:         size_type _r11 = {{0U|0UL}};
```
vgvassilev, line 277: That will not compiler. We probably mean `std::vector<...>::size_type`. I do not think that's a problem for this PR.

vgvassilev (commented):
We are almost there. See this one comment.

lib/Differentiator/VisitorBase.cpp
kchristin22: I think it makes sense to be in `getZeroInit` as we initialize with an empty list, while synthesizeLiteral is used in other places as well

test/Gradient/STLCustomDerivatives.C
kchristin22: I'll open an issue

test/Gradient/FunctionCalls.C
kchristin22: Yes indeed, fixed

vgvassilev (approved):
LGTM!

PR #999, opened 2024-07-23
Add support for `Kokkos::resize` in the forward mode

gojakuch:
This PR extends the Kokkos support in Clad by providing custom pushforwards for different overloads of the `Kokkos::resize` function.


vgvassilev:
@kliegeois, @brian-kelley can you take a look? Looks good to me.

unittests/Kokkos/ViewBasics.cpp
```diff
@@ -160,4 +160,48 @@ TEST(ViewBasics, TestDeepCopy2Reverse) {
...
+double f_basics_resize_1(double x, double y) {
+  Kokkos::View<double** [3], Kokkos::LayoutLeft, Kokkos::HostSpace> a("a", 3,
+                                                                      2);
+
+  Kokkos::deep_copy(a, 3 * x + y);
+  a(2, 1, 0) = x * y;
+
+  Kokkos::resize(Kokkos::WithoutInitializing, a, 5, 5);
+
+  return a(2, 1, 0);
```
kliegeois, line 174: I would suggest here (and on the following test as well) to change a little bit the test to return a value of the view that has been allocated during the resize.

For example doing something like this:
```
  Kokkos::deep_copy(a, 3 * x + y);

  Kokkos::resize(Kokkos::WithoutInitializing, a, 5, 5);

  a(4, 4, 0) = x * y;

  return a(4, 4, 0);
```

include/clad/Differentiator/KokkosBuiltins.h
```diff
@@ -39,6 +39,29 @@ inline void deep_copy_pushforward(const View1& dst, const View2& src, T param,
...
 }
 
+template <class View>
+inline void
+resize_pushforward(View& v, const size_t n0, const size_t n1, const size_t n2,
+                   const size_t n3, const size_t n4, const size_t n5,
+                   const size_t n6, const size_t n7, View& d_v,
+                   const size_t d_n0, const size_t d_n1, const size_t d_n2,
+                   const size_t d_n3, const size_t d_n4, const size_t d_n5,
+                   const size_t d_n6, const size_t d_n7) {
```
kliegeois, line 49: All these variables are not used in the function body.

I am almost sure that Clad requires to provide pushforward functions that takes as input the primal variables and each one of the derivatives but some code that use that function and that build with warning as errors will not compile due to unused variables.

If I recall correctly, a way to avoid that would be to have either:


```
(void) d_n0; 
(void) d_n1; 
(void) d_n2;
(void) d_n3; 
(void) d_n4; 
(void) d_n5;
(void) d_n6; 
(void) d_n7;
```
in the body of the function

or

replace 
```
                   const size_t d_n0, const size_t d_n1, const size_t d_n2,
                   const size_t d_n3, const size_t d_n4, const size_t d_n5,
                   const size_t d_n6, const size_t d_n7) {
```
with
```
                   const size_t /*d_n0*/, const size_t /*d_n1*/, const size_t /*d_n2*/,
                   const size_t /*d_n3*/, const size_t /*d_n4*/, const size_t /*d_n5*/,
                   const size_t /*d_n6*/, const size_t /*d_n7*/) {
```

include/clad/Differentiator/KokkosBuiltins.h
```diff
@@ -39,6 +39,29 @@ inline void deep_copy_pushforward(const View1& dst, const View2& src, T param,
...
+  ::Kokkos::resize(d_v, n0, n1, n2, n3, n4, n5, n6, n7);
+}
+
+template <class I, class dI, class View>
+inline void resize_pushforward(
+    const I& arg, View& v, const size_t n0, const size_t n1, const size_t n2,
+    const size_t n3, const size_t n4, const size_t n5, const size_t n6,
+    const size_t n7, const dI& d_arg, View& d_v, const size_t d_n0,
+    const size_t d_n1, const size_t d_n2, const size_t d_n3, const size_t d_n4,
+    const size_t d_n5, const size_t d_n6, const size_t d_n7) {
```
kliegeois, line 60: Same as above.

unittests/Kokkos/ViewBasics.cpp
```diff
@@ -160,4 +160,48 @@ TEST(ViewBasics, TestDeepCopy2Reverse) {
...
+double f_basics_resize_2(double x, double y) {
+  Kokkos::View<double** [3], Kokkos::LayoutLeft, Kokkos::HostSpace> a("a", 3,
+                                                                      2);
+
+  Kokkos::deep_copy(a, 3 * x + y);
+  a(2, 1, 0) = x * y;
+
+  Kokkos::resize(a, 5, 5);
+
+  return a(2, 1, 0);
```
kliegeois, line 196: Same as above.

kliegeois (commented):
Overall it looks good to me.
I just have couple of quick comments.

unittests/Kokkos/ViewBasics.cpp
gojakuch: oh sure, I can add these tests, that seems useful, but my initial goal with the tests I wrote was to check if the existing values are preserved after the resize, so I'd keep them

include/clad/Differentiator/KokkosBuiltins.h
gojakuch: sure, I'll address that if it can be a problem potentially.

include/clad/Differentiator/KokkosBuiltins.h
```diff
@@ -40,24 +40,28 @@ inline void deep_copy_pushforward(const View1& dst, const View2& src, T param,
...
-    const I& arg, View& v, const size_t n0, const size_t n1, const size_t n2,
-    const size_t n3, const size_t n4, const size_t n5, const size_t n6,
-    const size_t n7, const dI& d_arg, View& d_v, const size_t d_n0,
-    const size_t d_n1, const size_t d_n2, const size_t d_n3, const size_t d_n4,
-    const size_t d_n5, const size_t d_n6, const size_t d_n7) {
+inline void resize_pushforward(const I& arg, View& v, const size_t n0,
+                               const size_t n1, const size_t n2,
+                               const size_t n3, const size_t n4,
+                               const size_t n5, const size_t n6,
+                               const size_t n7, const dI& d_arg, View& d_v,
```
kliegeois, line 60: Same for `d_arg` 😃

kliegeois (commented):
Thanks for the modifications!

Just one extra comment.

gojakuch:
@vgvassilev @kliegeois I think I've addressed all the suggestions and we can merge it. I put everything in separate commits, we can squash that obviously.

vgvassilev:
@kliegeois, if that looks good to you we can merge it.

kliegeois:
It looks good to me thanks!

PR #1002, opened 2024-07-23
Add demo for comparison with libtorch (and pytorch)

vaithak:


demos/PytorchComparison/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
```
vgvassilev, line 1: Can we add a bot to run this as part of the CI?

demos/PytorchComparison/clad-basic-nn.cpp
```diff
@@ -0,0 +1,153 @@
...
+  std::cout << "Time taken for 100 matrix multiplications: "
+            << elapsed_bench.count() << " seconds" << std::endl;
+}
+
+int main() {
+  // Initialize random number generator.
+  std::mt19937 gen(42);
+
+  // Generate random input data and labels.
+  double x[M * N], y[M];
```
MihailMihov, line 118: Both of the demos seem to benchmark the multiplication XX, which I believe only makes sense if M=N. Maybe transpose one of them so that it's always valid?

demos/PytorchComparison/clad-basic-nn.cpp
```diff
@@ -0,0 +1,153 @@
...
+// Perform gradient descent to optimize the weights.
+void optimize(double x[M * N], double y[M], double w1[N], double w2[N],
+              double w3[N], double w4[3], double lr, int num_iters) {
+  auto grad = clad::gradient(loss_fn, "w1,w2,w3,w4");
+  for (int i = 0; i < num_iters; ++i)
+    performStep(x, y, w1, w2, w3, w4, grad, lr);
+}
+
+// Benchmark the time for 100 matrix multiplications.
+void benchmark_matrix_multiply(double x[M + N], double y[M]) {
```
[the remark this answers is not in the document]
MihailMihov: Also `y` seems to be unused here.

PR #1003, opened 2024-07-24
Don't assume output elements are used in the right order in jacobians

### LINKED ISSUE #479: Incorrect Jacobian generated when Order of assignment to Outputs is changed
Minimum reproducible example:
```cpp
void h(float a, float output[]){
  output[1]=a;
  output[0]=a*a;  
}
int main(){
  auto dh = clad::jacobian(h);
  float a=3;
  float output[2]={0};
  float jc[2]={0};
  dh.execute(a,output,jc);
  printf("Result = {%.2f, %.2f}",jc[0],jc[1]);
}
```

The above example prints: `Result = {0.00, 6.00}`, which is wrong

However if the order of assignment to outputs is correctly given:
```cpp
void h(float a, float output[]){
  output[0]=a*a;  
  output[1]=a;
}
int main(){
  auto dh = clad::jacobian(h);
  float a=3;
  float output[2]={0};
  float jc[2]={0};
  dh.execute(a,output,jc);
  printf("Result = {%.2f, %.2f}",jc[0],jc[1]);
}
```
The result is correct: `Result = {6.00, 1.00}`

PetroZarytskyi:
Currently, when creating jacobians, we assume that ``output`` elements are used in the ascending order. i.e.
```
output[0] = ...;
output[1] = ...;
```
is supported but
```
output[1] = ...;
output[0] = ...;
```
is not supported. We should not make such assumptions.
Fixes #479.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2308,7 +2308,9 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
                                         .get();
                 temp_m_Variables[m_IndependentVars[i]] = result_at_i;
               }
-              m_VectorOutput.push_back(temp_m_Variables);
+              if (m_VectorOutput.size() <= outputArrayCursor)
+                m_VectorOutput.resize(outputArrayCursor + 1);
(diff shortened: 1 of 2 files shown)


vgvassilev:
Can you move the pull request description in the commit message, too?

vgvassilev (approved):
LGTM!

PR #1006, opened 2024-07-26
Add propagators for __builtin_pow.

### LINKED ISSUE #1005: Clad does not have propagators for __builtin_pow
```cpp
double pow2(double x) { return __builtin_pow(x, 2); }

int main() {
  auto df = clad::differentiate(pow2, "x");
  printf("pow2dx(1)=%f\n", df.execute(1));
}
```

Adding
```cpp 


namespace clad { namespace custom_derivatives {
  CUDA_HOST_DEVICE ValueAndPushforward<decltype(::std::pow(T1(), T2())),
                                       decltype(::std::pow(T1(), T2()))>
  __builtin_pow_pushforward(T1 x, T2 exponent, T1 d_x, T2 d_exponent) {
    auto val = ::std::pow(x, exponent);
    auto derivative = (exponent * ::std::pow(x, exponent - 1)) * d_x;
    // Only add directional derivative of base^exp w.r.t exp if the directional
    // seed d_exponent is non-zero. This is required because if base is less than or
    // equal to 0, then log(base) is undefined, and therefore if user only requested
    // directional derivative of base^exp w.r.t base -- which is valid --, the result would
    // be undefined because as per C++ valid number + NaN * 0 = NaN.
    if (d_exponent)
      derivative += (::std::pow(x, exponent) * ::std::log(x)) * d_exponent;
    return {val, derivative};
  }

  }}
```

makes the example work.

This issue reflects root-project/root#15912

vgvassilev:
Fixes #1005.

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -72,6 +72,61 @@ ValueAndPushforward<int, int> cudaDeviceSynchronize_pushforward()
 }
 #endif
 
+CUDA_HOST_DEVICE inline ValueAndPushforward<double, double>
+__builtin_pow_pushforward(double x, double exponent, double d_x,
+                          double d_exponent) {
+  auto val = __builtin_pow(x, exponent);
+  double derivative = (exponent * __builtin_pow(x, exponent - 1)) * d_x;
+  // Only add directional derivative of base^exp w.r.t exp if the directional
+  // seed d_exponent is non-zero. This is required because if base is less than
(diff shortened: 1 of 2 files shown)


vgvassilev:
@vaithak, can you help? It looks like here https://github.com/vgvassilev/clad/blob/69d15a31d11fa38b193d274e2392a086f54006bc/lib/Differentiator/ReverseModeVisitor.cpp#L1797 we fail to find the already generated derivative for `__builtin_pow`'s pullback...

There are two questions that are unclear to me. First why do we need `__builtin_pow`'s pullback and secondly why we fail to find its custom derivative?

PR #1015, opened 2024-07-29
Fix cloning `MemberCallExpr` for older Clang versions

gojakuch:
This small change prevents a segfault from occuring on older versions of Clang when trying to clone some member call expressions. This should fix the failing checks for #992.


gojakuch:
@vgvassilev there are no tests in this PR, but it's a feature that is already used in tests, and the specific test for this fix is basically included in the tests of #992, as the easiest ways to reproduce the original issue seem to involve arrays or other STL structures anyway.

vgvassilev (approved):
Lgtm!

PR #1017, opened 2024-07-31
Move to using xeus-cpp instead of xeus-cling

### LINKED ISSUE #985: Move the binder service from xeus-cling to xeus-cpp

vgvassilev:
Fixes #985.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,5 +1,5 @@
 channels:
   - conda-forge
 dependencies:
-  - xeus-cling
+  - xeus-cpp
   - clad


davidlange6:
fwiw, I tried this and see that it doesn't preserve current behavior of printing results  - eg, the first intro notebook example results the image below. (also a compilation error means restarting the kernel - though I do not believe that to be a regression.

image

vgvassilev:
Yes, if you replace it with `printf` it works but there are still a few issues. I've asked @tharun571 to take a look.

anutosh491:
I was looking into some functionality for xeus-clad (maybe using xeus-cpp) only. Is this good to go ?

tharun571:
> I was looking into some functionality for xeus-clad (maybe using xeus-cpp). Is this good to go ?

I don't think so. There are some functional discrepancies present. Hessian and Jacobian functions do not work as intended with xeus-cpp.

PR #1020, opened 2024-08-01
Support bitwise, shift, comparison, and remainder operators in fwd mode

### LINKED ISSUE #381: Add support for differentiation of bitwise operations
We allow differentiation of integral types and thus we should also allow differentiation of bitwise operators. 

For example, we should support differentiation of statements such as these: 

```cpp
int i, j;
i <<=3;
i = i | j;
i ^= j;
```

PetroZarytskyi:
This PR adds support for bitwise, shift, comparison, remainder, and bitwise not operators. Shift operators are considered differentiable since they essentially represent multiplication by ``2^n`` or ``2^-n``, where ``n`` is the RHS of the shift operators ``<<`` and ``>>``. Not operators are considered differentiable as well because they represent ``2^n - 1 - x`` or ``- 1 - x`` (depending on whether the type is signed) so the derivative is ``-_d_x``. Other operators have unclear differentiable effects and so they are considered non-differentiable.
Fixes #381.

### DIFF
--- a/include/clad/Differentiator/BaseForwardModeVisitor.h
+++ b/include/clad/Differentiator/BaseForwardModeVisitor.h
@@ -102,6 +102,8 @@ public:
   StmtDiff VisitCXXThisExpr(const clang::CXXThisExpr* CTE);
   StmtDiff VisitCXXNewExpr(const clang::CXXNewExpr* CNE);
   StmtDiff VisitCXXDeleteExpr(const clang::CXXDeleteExpr* CDE);
+  StmtDiff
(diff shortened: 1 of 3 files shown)


gojakuch:
probably fixes the #381

vgvassilev:
> probably fixes the #381

@PetroZarytskyi, can you update the commit message including this issue?

PetroZarytskyi:
@vgvassilev I updated the commit message.

vgvassilev (approved):
We can't test this easily and we can ignore codecov for now. LGTM!

PR #1022, opened 2024-08-02
Add support for `Kokkos::parallel_for` in the fwd mode

gojakuch:


gojakuch:
@vgvassilev I don't really get why the kokkos check fails, there's just a bunch of warnings, no errors though. or is there a limit on how many warnings it can get?

gojakuch:
@vgvassilev ping

gojakuch:
@vgvassilev I've finally managed to fix the failing check. can we merge this?

vgvassilev:
@kliegeois, @brian-kelley, can you take a look?

gojakuch:
if we merge #1038 before this, I think it'd make sense to update the tests in this PR accordingly

unittests/Kokkos/ParallelFor.cpp
```diff
@@ -89,4 +90,118 @@ TEST(ParallelFor, ParallelPolynomialReverse) {
...
+  KOKKOS_INLINE_FUNCTION
+  void operator()(const int i) const { res(i) = x * i; }
+};
+
+double parallel_for_functor_simplest_case_intpol(double x) {
+  Kokkos::View<double[5], Kokkos::HostSpace> res("res");
+
+  Foo<Kokkos::View<double[5], Kokkos::HostSpace>> f(res, x);
+
+  f(0);
```
vgvassilev, line 110: Can we add a fixme note to track this being a workaround for a bug in clad?

kliegeois:
@vgvassilev I will try to review this PR by the end of the week.

unittests/Kokkos/ParallelFor.cpp
gojakuch: done

kliegeois:
Does it work with the `generate-source-file` option?
I tried briefly locally and got a bunch of warnings and/or errors.

gojakuch:
> Does it work with the `generate-source-file` option? I tried briefly locally and got a bunch of warnings and/or errors.

I haven't tried that. could you give some clues on what exactly you ran?

kliegeois:
Thanks @gojakuch ! I was going to reply today but that's great that you have been able to reproduce it as well.
I am wondering if we should have some sort of unit tests that verify that `generate-source-file` works with Kokkos based code and that the generated files are as expected.
What do you think @vgvassilev  ?

vgvassilev:
> Thanks @gojakuch ! I was going to reply today but that's great that you have been able to reproduce it as well. I am wondering if we should have some sort of unit tests that verify that `generate-source-file` works with Kokkos based code and that the generated files are as expected. What do you think @vgvassilev ?

Yes, we should. It's on my todo list but still have not got time to work on it. Basically the idea is to enhance the unittest infrastructure to use the infrastructure to generate a file and then use it as part of the compilation. That should be possible with cmake but will need some time to implement.

gojakuch:
> Thanks @gojakuch ! I was going to reply today but that's great that you have been able to reproduce it as well. I am wondering if we should have some sort of unit tests that verify that `generate-source-file` works with Kokkos based code and that the generated files are as expected. What do you think @vgvassilev ?

well, that's the thing. I understand what you ran but it works out fine on my machine (at least while generating). it produces code without warnings, however the produced code is not compilable on its own, but that's a problem of Clang, as far as I'm aware. it's just that some types are not output to text correctly when resolving the template definitions. so I'm not sure it's fixable on Clad's side. at least that's the only thing happening on my machine from what I can see. correct me if you meant smth else.

gojakuch:
the last forced push was just me rebasing the PR

gojakuch:
> > Thanks @gojakuch ! I was going to reply today but that's great that you have been able to reproduce it as well. I am wondering if we should have some sort of unit tests that verify that `generate-source-file` works with Kokkos based code and that the generated files are as expected. What do you think @vgvassilev ?
> 
> @kliegeois well, that's the thing. I understand what you ran but it works out fine on my machine (at least while generating). it produces code without warnings, however the produced code is not compilable on its own, but that's a problem of Clang...

@kliegeois an update: after a quick chat with Vassil earlier today, I'm now convinced that it's fixable on our side. but that falls out of the scope of this PR, so I'll open a separate issue for this and fix it in a separate PR. do you have any suggestions apart from that? if not, I think we can merge this and move on, while I'm working on a fix for that

kliegeois:
@gojakuch sure, that sounds good to me, thanks!

PR #1025, opened 2024-08-04
Add llvm 18 osx ci

vgvassilev:
This is the same as #889 but I cannot login to the bots. Let's try if this will let me use tmate...


mcbarton:
@vgvassilev If we can work out how to get this PR through then I'll look at adding llvm 19 compatibility.

vgvassilev:
> @vgvassilev If we can work out how to get this PR passing then I'll look at adding llvm 19 compatibility.

On my todo list... I have everything set up, don't have the time to look at it just yet.

PR #1029, opened 2024-08-05
Support multiple indices in clad::gradient calls

### LINKED ISSUE #46: Add an option to specify independent variables for gradients
Now we can only compute gradients w.r.t. to every argument of a function.

For efficiency reasons, it seems to be a good idea to be able to specify indices required of independent variables, e.g.:

```
clad::gradient(f, 0, 2, 4)
```
or
```
clad::gradient(f, {0, 2, 4})
```
or even
```
clad::gradient(f, std::index_sequence<0, 2, 4>{})
```

PetroZarytskyi:
Currently, the user can provide a string as the second argument of ``clad::gradient`` to specify independent parameters as a list of comma-separated names. This PR allows users to specify indices alongside with names. e.g.
```
clad::gradient(fn, "0");
clad::gradient(fn, "1, z");
...
```
Previously, it was possible to provide a single index as an integer literal. e.g.
```
clad::gradient(fn, 0);
```
Fixes #46.


vgvassilev (approved):
LGTM!

PR #1037, opened 2024-08-10
Add support of custom `_forw` functions

infinite-void-16:
This commit adds support for custom (user-provided) `_forw` functions.
A `_forw` function, if available, is called in place of the actual
function.

For example, if the primal code contains:

```cpp
someFn(u, v, w);
```

and user has defined a custom `_forw` function for `someFn` as follows:

```cpp
namespace clad {
    void someFn_forw(double u, double v, double w, double *d_u,
      double *d_v, double *dw) {
      // ...
      // ...
    }
  }
}
```

Then clad will generate the derivative function as follows:

```cpp
// forward-pass
clad::custom_derivatives::someFn_forw(u, v, w, d_u, d_v, d_w);
// ...

// reverse-pass; no change in reverse-pass
someFn_pullback(u, v, w, d_u, d_v, d_w);
// ...
```

But more importantly, why do we need such a functionality? Two reasons:

- Supporting reference/pointer return types in the reverse-mode. This
  has been discussed at great length here:
https://github.com/vgvassilev/clad/pull/425 (https://github.com/vgvassilev/clad/pull/425)

- Supporting types whose elements grows dynamically, such as
  `std::vector` and `std::map`. The issue is that we correctly
  need to update the size/property of the adjoint variable when a
  function call updates the size/property of the corresponding primal
  variable. For example: a call to `vec.push_back(...)` should update
  the size of `_d_vec` as well. However, the actual function call does
  not modify the adjoint variable in any way. Here comes `_forw` functions
  to the rescue. `_forw` functions makes it possible to adjust the adjoint
  variable size/properties along with executing the actual function call.

Please note that `_forw` function signature takes adjoint variables as
arguments and return `clad::ValueAndAdjoint<U, V>` to support the
reference/pointer return type.

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -30,6 +30,11 @@ template <typename T, typename U> struct ValueAndPushforward {
   }
 };
 
+template <typename T, typename U> struct ValueAndAdjoint {
+  T value;
+  U adjoint;
+};
+
 /// It is used to identify constructor custom pushforwards. For
 /// constructor custom pushforward functions, we cannot use the same
 /// strategy which we use for custom pushforward for member
--- a/include/clad/Differentiator/STLBuiltins.h
+++ b/include/clad/Differentiator/STLBuiltins.h
@@ -202,6 +202,35 @@ void fill_pushforward(::std::array<T, N>* a, const T& u,
   d_a->fill(d_u);
 }
 
+template <typename T, typename U>
+void push_back_reverse_forw(::std::vector<T>* v, U val, ::std::vector<T>* d_v,
+                            U* d_val) {
+  v->push_back(val);
+  d_v->push_back(0);
+}
+
+template <typename T, typename U>
+void push_back_pullback(::std::vector<T>* v, U val, ::std::vector<T>* d_v,
+                        U* d_val) {
+  *d_val += d_v->back();
+  d_v->pop_back();
+}
+
+template <typename T>
+clad::ValueAndAdjoint<T&, T&> operator_subscript_reverse_forw(
+    ::std::vector<T>* vec, typename ::std::vector<T>::size_type idx,
+    ::std::vector<T>* d_vec, typename ::std::vector<T>::size_type* d_idx) {
+  return {(*vec)[idx], (*d_vec)[idx]};
+}
+
+template <typename T, typename P>
+void operator_subscript_pullback(::std::vector<T>* vec,
+                                 typename ::std::vector<T>::size_type idx,
+                                 P d_y, ::std::vector<T>* d_vec,
+                                 typename ::std::vector<T>::size_type* d_idx) {
+  (*d_vec)[idx] += d_y;
+}
+
 } // namespace class_functions
 } // namespace custom_derivatives
 } // namespace clad
--- a/include/clad/Differentiator/CladUtils.h
+++ b/include/clad/Differentiator/CladUtils.h
@@ -321,6 +321,9 @@ namespace clad {
 
     bool IsMemoryFunction(const clang::FunctionDecl* FD);
     bool IsMemoryDeallocationFunction(const clang::FunctionDecl* FD);
+
+    /// Returns true if QT is a non-const reference type.
+    bool isNonConstReferenceType(clang::QualType QT);
     } // namespace utils
     } // namespace clad
 
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -22,12 +22,8 @@
 #include <cstring>
 
 namespace clad {
-template <typename T, typename U> struct ValueAndAdjoint {
-  T value;
-  U adjoint;
-};
 
-  /// \returns the size of a c-style string
+/// \returns the size of a c-style string
 inline CUDA_HOST_DEVICE unsigned int GetLength(const char* code) {
   const char* code_copy = code;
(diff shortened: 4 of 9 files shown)


include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -30,6 +30,11 @@ template <typename T, typename U> struct ValueAndPushforward {
   }
 };
 
+template <typename T, typename U> struct ValueAndAdjoint {
```
gojakuch, line 33: is it really necessary to move this?

vgvassilev (approved):
Lgtm! @PetroZarytskyi, can you take a look at the change in the TBR part?

include/clad/Differentiator/STLBuiltins.h
```diff
@@ -202,6 +202,35 @@ void fill_pushforward(::std::array<T, N>* a, const T& u,
   d_a->fill(d_u);
 }
 
+template <typename T, typename U>
+void push_back_forw(::std::vector<T>* v, U val, ::std::vector<T>* d_v,
```
gojakuch, line 206: @vgvassilev do you think `_forw` is a good way to name these? it's not a big deal but I feel like it's a bit counter-intuitive and might be easily confused with pushforwards, epsecially for new users. I mean, most people using clad in their projects wouldn't be experts on AD and our naming, so maybe something like `_reverse_forw`, `_rmforw`, or `_rvs_forw` or anything that indicates the usage of this in reverse mode only would be a better option? I think the naming should be at least partially self-explanatory for Clad not to become overcomplicated to new users (since its compile-time nature is already confusing enough for an unprepared mind :D).

test/Gradient/UserDefinedTypes.C
```diff
@@ -326,6 +329,89 @@ double fn9(Tangent t, dcomplex c) {
...
+    return vec[0] + vec[1];
+}
+
+namespace clad {
+    namespace custom_derivatives {
+        namespace class_functions {
+            ::std::vector<size_t> size_update_stack;
+
+            template <typename T>
+            void resize_forw(::std::vector<T> *v,
```
gojakuch, line 354: why is this defined in tests and not in `STLBuiltins.h` if these are `std::vector` operations? also, since this is an STL test, it should probably be put into another test file, not in this one for **user**-defined types :) (but that's optional, of course, it would be nice though). take a look at `test/ForwardMode/STLCustomDerivatives.C` (if I remember correctly) for example. I think these tests should basically mirror that.

gojakuch (commented):
nice! this should help a lot. @parth-07 is this consistent with what you'd thought about custom constructor support in the reverse mode? if so, I think we can do smth similar for them, I suppose

include/clad/Differentiator/STLBuiltins.h
vgvassilev: I agree. @infinite-void-16 what do you think?

include/clad/Differentiator/BuiltinDerivatives.h
infinite-void-16: Yes. It was previously in `Differentiator.h` and we should not include `Differentiator.h` in `BuiltinDerivatives.h` or `STLBuiltins.h`

include/clad/Differentiator/STLBuiltins.h
infinite-void-16: I agree as well. We should also change the suffix of `_forw` functions that Clad itself generates.

test/Gradient/UserDefinedTypes.C
infinite-void-16: This should not go in STLBuiltins.h because these functions using a custom global stack and are thus not thread-safe. I will move them into a separate file.

include/clad/Differentiator/STLBuiltins.h
infinite-void-16: I have changed the suffix to `_reverse_forw`

test/Gradient/UserDefinedTypes.C
infinite-void-16: These custom derivatives should not go in `STLBuiltins.h` because these functions are using a custom global stack and are thus not thread-safe. They only serve as an example of how users can define custom derivatives for the cases where a stack cannot be avoided. 

I liked the idea to move the tests to a different file. I have moved the new tests to `STLCustomDerivatives.C` as suggested.

gojakuch (approved):
LGTM

lib/Differentiator/TBRAnalyzer.cpp
```diff
@@ -203,9 +203,13 @@ TBRAnalyzer::VarData::VarData(QualType QT, bool forceNonRefType) {
...
-    ProfileID nonConstIdxID;
-    auto& idxData = (*m_Val.m_ArrData)[nonConstIdxID];
-    idxData = VarData(QualType::getFromOpaquePtr(elemType));
+    // FIXME: In some cases for Mac, 'elemType' is nullptr for std::vector<T>
+    // internal members.
+    if (elemType) {
+      ProfileID nonConstIdxID;
+      auto& idxData = (*m_Val.m_ArrData)[nonConstIdxID];
+      idxData = VarData(QualType::getFromOpaquePtr(elemType));
+    }
```
vgvassilev, line 212: After #1046 we should not need this part. Can you remove it and see if things still work?

infinite-void-16: I have removed this change, and yes, things seems to work fine now.

PR #1039, opened 2024-08-14
Introduce placeholders in the reverse mode to correctly handle multiplication

PetroZarytskyi:
Currently, when differentiating multiplication in the reverse mode, we need to pass the differentiated LHS when visiting the RHS and vice versa (i.e. ``Visit(R, dR)`` and ``Visit(L, dL)`` where ``dR = LDiff * dfdx`` and ``dL = dfdx * RDiff``). This creates a loop that we break in one of two ways:
1) Create a variable to represent the result of the RHS visitation, use it to visit the LHS, visit the RHS, and then set the value of the variable introduced before to the result of the RHS visitation.
2) Clone the RHS, use it to visit the LHS, then visit the RHS.

The 1st approach is bad because it introduces a new variable that is usually unnecessary.

The 2nd approach is used more frequently but its downside is that cloning is not the same as visiting, some expressions cannot be cloned. e.g.
a) ``x > 0 ? a : b`` is cloned as ``x > 0 ? a : b`` but differentiated as ``_cond ? a : b``, where ``_cond`` is a variable to save the condition ``x > 0``.
b) References are often turned into pointers in the reverse mode. Because of that, ``x`` should be differentiated as ``*x``  and not as just ``x``, which ``Clone`` does. We already have an exception to handle ref-type decl refs as RHS of multiplication, which is removed in this PR.

This PR solves this problem in a third way: by passing a literal expression when visiting LHS and replacing it when the differentiated RHS is known. All of the new logic is sunk into existing functions ``DelayedGlobalStoreAndRef`` and ``Finalize``.

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -400,6 +400,8 @@ namespace clad {
     virtual StmtDiff VisitReturnStmt(const clang::ReturnStmt* RS);
     StmtDiff VisitStmt(const clang::Stmt* S);
     virtual StmtDiff VisitUnaryOperator(const clang::UnaryOperator* UnOp);
+    StmtDiff
+    VisitUnaryExprOrTypeTraitExpr(const clang::UnaryExprOrTypeTraitExpr* UE);
     StmtDiff VisitExprWithCleanups(const clang::ExprWithCleanups* EWC);
     /// Decl is not Stmt, so it cannot be visited directly.
     StmtDiff VisitWhileStmt(const clang::WhileStmt* WS);
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -278,13 +278,14 @@ namespace clad {
       bool isInsideLoop;
       bool isFnScope;
       bool needsUpdate;
+      clang::Expr* Placeholder;
       DelayedStoreResult(ReverseModeVisitor& pV, StmtDiff pResult,
-                         clang::VarDecl* pDeclaration, bool pIsConstant,
-                         bool pIsInsideLoop, bool pIsFnScope,
-                         bool pNeedsUpdate = false)
+                         clang::VarDecl* pDeclaration, bool pIsInsideLoop,
+                         bool pIsFnScope, bool pNeedsUpdate = false,
+                         clang::Expr* pPlaceholder = nullptr)
           : V(pV), Result(pResult), Declaration(pDeclaration),
-            isConstant(pIsConstant), isInsideLoop(pIsInsideLoop),
-            isFnScope(pIsFnScope), needsUpdate(pNeedsUpdate) {}
+            isInsideLoop(pIsInsideLoop), isFnScope(pIsFnScope),
+            needsUpdate(pNeedsUpdate), Placeholder(pPlaceholder) {}
       void Finalize(clang::Expr* New);
     };
 
@@ -297,7 +298,8 @@ namespace clad {
     /// This is what DelayedGlobalStoreAndRef does. E is expected to be the
     /// original (uncloned) expression.
     DelayedStoreResult DelayedGlobalStoreAndRef(clang::Expr* E,
-                                                llvm::StringRef prefix = "_t");
+                                                llvm::StringRef prefix = "_t",
+                                                bool forceStore = false);
 
     struct CladTapeResult {
       ReverseModeVisitor& V;
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -4107,6 +4107,11 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     return Visit(NTTP->getReplacement());
   }
 
+  StmtDiff ReverseModeVisitor::VisitUnaryExprOrTypeTraitExpr(
+      const clang::UnaryExprOrTypeTraitExpr* UE) {
+    return {Clone(UE), Clone(UE)};
+  }
+
   DeclDiff<StaticAssertDecl> ReverseModeVisitor::DifferentiateStaticAssertDecl(
       const clang::StaticAssertDecl* SAD) {
     return DeclDiff<StaticAssertDecl>(nullptr, nullptr);
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2319,25 +2319,8 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       // to reduce cloning complexity and only clones once. Storing it in a
       // global variable allows to save current result and make it accessible
       // in the reverse pass.
-      std::unique_ptr<DelayedStoreResult> RDelayed;
-      StmtDiff RResult;
-      // If R has no side effects, it can be just cloned
-      // (no need to store it).
-
-      // Check if the local variable declaration is reference type, since it is
-      // moved to the global scope and the right side should be recomputed
-      bool promoteToFnScope = false;
-      if (auto* RDeclRef = dyn_cast<DeclRefExpr>(R->IgnoreImplicit()))
-        promoteToFnScope = RDeclRef->getDecl()->getType()->isReferenceType() &&
-                           !getCurrentScope()->isFunctionScope();
-
-      if (!ShouldRecompute(R) || promoteToFnScope) {
-        RDelayed = std::unique_ptr<DelayedStoreResult>(
-            new DelayedStoreResult(DelayedGlobalStoreAndRef(R)));
-        RResult = RDelayed->Result;
-      } else {
-        RResult = StmtDiff(Clone(R));
-      }
+      DelayedStoreResult RDelayed = DelayedGlobalStoreAndRef(R);
+      StmtDiff& RResult = RDelayed.Result;
 
       Expr* dl = nullptr;
       if (dfdx())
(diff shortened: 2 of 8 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3367,9 +3352,51 @@
...
   }
 
-  void ReverseModeVisitor::DelayedStoreResult::Finalize(Expr* New) {
-    if (isConstant || !needsUpdate)
+  void ReverseModeVisitor::DelayedStoreResult::Finalize(Expr* New, Sema& S) {
+    class PlaceholderReplacer
+        : public RecursiveASTVisitor<PlaceholderReplacer> {
+    public:
+      const Expr* placeholder;
+      Sema& m_Sema;
```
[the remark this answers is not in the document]
PetroZarytskyi: @vgvassilev Is this really a problem?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2588,14 +2570,17 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+        StmtDiff& RResult = RDelayed.Result;
         Expr* RStored =
             StoreAndRef(RResult.getRevSweepAsExpr(), direction::reverse);
         addToCurrentBlock(BuildOp(BO_AddAssign, ResultRef,
                                   BuildOp(BO_Div, oldValue, RStored)),
                           direction::reverse);
-        if (!RDelayed.isConstant) {
+        Expr::EvalResult dummy;
+        if (!clad_compat::Expr_EvaluateAsConstantExpr(R, dummy, m_Context) ||
+            RDelayed.needsUpdate) {
```
vgvassilev, line 2583: Can we write a test case here to make codecov happy?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3368,8 +3353,50 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+          if (*iter == placeholder) {
+            Expr* castExpr = newExpr;
+            // Since we are manually replacing the statement, implicit casts are
+            // not generated automatically.
+            QualType exprTy = newExpr->getType();
+            QualType targetTy = cast<Expr>(*iter)->getType();
+            if (exprTy->isIntegralType(m_Context) && targetTy->isFloatingType())
+              castExpr = m_Sema
+                             .ImpCastExprToType(newExpr, targetTy,
+                                                CK_IntegralToFloating)
```
vgvassilev, line 3378: We cannot rely on the specific cast type. We can probably use `m_Sema.PrepareScalarCast(from, to)`.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3368,8 +3353,50 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+      const Expr* placeholder;
+      Sema& m_Sema;
+      ASTContext& m_Context;
+      Expr* newExpr{nullptr};
+      PlaceholderReplacer(const Expr* Placeholder, Sema& S)
+          : placeholder(Placeholder), m_Sema(S), m_Context(S.getASTContext()) {}
+
+      bool VisitExpr(Expr* E) const {
+        for (auto iter = E->child_begin(), e = E->child_end(); iter != e;
+             ++iter)
```
vgvassilev, line 3368:
```suggestion
        for (Stmt *& S : E->children())
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3368,8 +3353,50 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+        : public RecursiveASTVisitor<PlaceholderReplacer> {
+    public:
+      const Expr* placeholder;
+      Sema& m_Sema;
+      ASTContext& m_Context;
+      Expr* newExpr{nullptr};
+      PlaceholderReplacer(const Expr* Placeholder, Sema& S)
+          : placeholder(Placeholder), m_Sema(S), m_Context(S.getASTContext()) {}
+
+      bool VisitExpr(Expr* E) const {
```
vgvassilev, line 3366: Can we explain why iterating over the child is safe and what we want to do in general?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3368,8 +3353,50 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    if (!needsUpdate)
+      return;
+
+    if (Placeholder) {
+      PlaceholderReplacer repl(Placeholder, V.m_Sema);
+      repl.newExpr = New;
+      for (Stmt* S : V.getCurrentBlock(direction::forward))
+        repl.TraverseStmt(S);
+      for (Stmt* S : V.getCurrentBlock(direction::reverse))
+        repl.TraverseStmt(S);
```
vgvassilev, line 3395: We should fold this into the class itself and just call the constructor and eventually a function `Replace`.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -292,7 +293,8 @@ namespace clad {
     /// This is what DelayedGlobalStoreAndRef does. E is expected to be the
     /// original (uncloned) expression.
     DelayedStoreResult DelayedGlobalStoreAndRef(clang::Expr* E,
-                                                llvm::StringRef prefix = "_t");
+                                                llvm::StringRef prefix = "_t",
+                                                bool forceNoRecompute = false);
```
vgvassilev, line 297: Too many negations. Can you clarify?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3385,22 +3412,32 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
-          clad_compat::Expr_EvaluateAsConstantExpr(E, evalRes, m_Context);
       return DelayedStoreResult{*this,
                                 Ediff,
                                 /*Declaration=*/nullptr,
-                                /*isConstant=*/isConst,
                                 /*isInsideLoop=*/false,
                                 /*isFnScope=*/false,
                                 /*pNeedsUpdate=*/false};
     }
+    if (!forceNoRecompute && ShouldRecompute(E)) {
```
vgvassilev, line 3428: Is `ShouldRecompute` not a superset of `forceNoRecompute`?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3385,22 +3412,32 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+      // The value of the literal has no. It's given a very particular value for
+      // easier debugging.
+      Expr* PH = ConstantFolder::synthesizeLiteral(E->getType(), m_Context,
+                                                   /*val=*/31);
```
vgvassilev, line 3432:
```suggestion
                                                   /*val=*/~0U);
```
Is the standard form.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3385,22 +3412,32 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+      return DelayedStoreResult{*this,
+                                StmtDiff{PH, nullptr, nullptr, PH},
+                                /*Declaration=*/nullptr,
+                                /*isInsideLoop*/ false,
```
vgvassilev, line 3436:
```suggestion
                                /*isInsideLoop=*/false,
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3385,22 +3412,32 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
                                 /*isFnScope=*/false,
                                 /*pNeedsUpdate=*/false};
     }
+    if (!forceNoRecompute && ShouldRecompute(E)) {
+      // The value of the literal has no. It's given a very particular value for
+      // easier debugging.
+      Expr* PH = ConstantFolder::synthesizeLiteral(E->getType(), m_Context,
+                                                   /*val=*/31);
+      return DelayedStoreResult{*this,
+                                StmtDiff{PH, nullptr, nullptr, PH},
```
vgvassilev, line 3434:
```suggestion
                                StmtDiff{PH, /*diff=*/nullptr, /*forwSweepDiff=*/nullptr, PH},
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3380,8 +3353,56 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    };
+
+    if (!needsUpdate)
+      return;
+
+    if (Placeholder) {
+      PlaceholderReplacer repl(Placeholder, V.m_Sema);
+      repl.newExpr = New;
+      repl.Replace(V);
+      Result = New;
```
vgvassilev, line 3402:
```suggestion
      PlaceholderReplacer repl(Placeholder, V.m_Sema);
      repl.Replace(V, New, Result);
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3380,8 +3353,56 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+
+      void Replace(ReverseModeVisitor& RMV) {
+        for (Stmt* S : RMV.getCurrentBlock(direction::forward))
+          TraverseStmt(S);
+        for (Stmt* S : RMV.getCurrentBlock(direction::reverse))
+          TraverseStmt(S);
+      }
+
+      // We need to iterate through exprs and then through their children to
+      // replace the placeholders for their parents.
```
vgvassilev, line 3378:
```suggestion
      // We chose iteration rather than visiting because we only do this for simple Expression subtrees and it is not worth it to implement an entire visitor infrastructure for simple replacements.
```

lib/Differentiator/ReverseModeVisitor.cpp

lib/Differentiator/ReverseModeVisitor.cpp

include/clad/Differentiator/ReverseModeVisitor.h
PetroZarytskyi: I changed the name to ``forceStore``.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: I'm not sure I understand. Is this still relevant?

vgvassilev (approved):
Lgtm!

Issue #1036, opened 2024-08-08
Call to `clad::gradient` for CUDA kernels cannot be compiled for GPU


kchristin22:
Since global kernels cannot be called like normal device functions from other device functions, the following command can be compiled only for host like so:
```
#ifndef __CUDA_ARCH__
    auto kernel_g = clad::gradient(kernel);
#endif
```
When executing the kernel though, the device needs to recognize(="see") the kernel. Hence, when trying to execute it, the device is not able to read the kernel (it says: "device symbol not found").

A simple workaround is generating a header file with a fixed name in the build folder(or in inc) like the compiler option `-fgenerate-source-file` does, placing the derived kernel there and including the file in `Differentiator.h`. The user would call the gradient with the macros like above.

There are some other ways that could also work, like loading the kernel to GPU manually using CUDA Modules. However, this may require assemblied code in string and not the c++ code. 

Also, transferring (allocating) a function pointer from host to device using `cudaMemcyToSymbol` does not work for kernels.

Maybe there could be a `gradient_kernel` function that is also a kernel and the returned object is an argument of that function (as kernels are void functions), but it is speculation that it could work and there may be problem with execute down the line.

vgvassilev:
I think I understand the problem. Can you write a simple standalone example with a gradient written by hand which compiles. Then we can think of how to implement that in clad. I'd be interested to see the different scenarios which a gradient on device is useful and its usage pattern in the GPU programming model.

kchristin22:
I made some progress with this and made it work in [cuda-compilation-support](https://github.com/kchristin22/clad/tree/cuda-compilation-support) branch. At compile time, whenever a global kernel is differentiated, the code is stored in a file, compiled using clang cuda and its assembly for the GPU (ptx code) is extracted in another file. The latter file can be loaded at runtime on the GPU, its function handle acquired and launched with a specified configuration.

Note that this code was used only to see what works and some things were taken for granted (the visible name of the kernel in the ptx file for instance) and of course the `CUDA Module` instructions should not be visible to the user. 

I want to work on:
* extracting the ptx code and storing it as string in the CladFunction object, as well as the visible name of the kernel (so all the produced files are deleted after compilation)
* moving the cuda module calls inside an execute function

kchristin22:
Update:
[...]

PR #1041, opened 2024-08-16
Add CUDA derived global kernels' compilation support in reverse mode

kchristin22:
This PR solves #1036. CUDA kernels can now be an arg to a `clad::gradient` function and the derived kernels can be executed successfully. 

Details on issues faced and the final approach can be found [here](https://github.com/vgvassilev/clad/issues/1036#issuecomment-2295242020).


parth-07:
@kchristin22 Can you please explain the solution/design as part of the pull-request description?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -177,9 +183,11 @@ namespace clad {
...
                                Sema& SemaRef) {
     CallExpr* call = this->CallContext;
-    // Index of "code" parameter:
-    auto codeArgIdx = static_cast<int>(call->getNumArgs()) - 1;
-    auto derivedFnArgIdx = codeArgIdx - 1;
+    // Index of "kernelName" parameter:
+    auto kernelNameArgIdx = static_cast<int>(call->getNumArgs()) - 1;
+    auto kernelArgIdx = kernelNameArgIdx - 2;
+    auto codeArgIdx = kernelNameArgIdx - 3;
+    auto derivedFnArgIdx = kernelNameArgIdx - 4;
```
vgvassilev, line 190: Should we not conditionally check if we are in cuda mode or that's host C++?

vgvassilev:
@kchristin22, what happened here? Something to do with rebase/git?

kchristin22:
@vgvassilev I opened a new one to have a clean slate. The old branch is renamed and kept in my repository in case it's ever needed.

vgvassilev:
Ok, but please do not generally do that since we lose review history. Next time just ping me and I can help you.

PR #1042, opened 2024-08-16
Update FileCheck test lines for number literals and types generated on windows

fsfod:
Update FileCheck test lines for number literals and types generated on windows that were checking for `int` or `long` before to `long long` since `long` on 64 bit windows is only 32 bit so size types have to be long long.
This fixes about 20 tests failing on windows with clang-cl.


vgvassilev:
@fsfod, this looks good. Can you give us a recipe how to build llvm in plugins mode to test this patch and try to not regress again?

fsfod:
on Linux?

vgvassilev:
> on Linux?

On windows. In theory we can get a version of llvm, apply a huge patch, build it, cache it and use it in our ci until everything lands.

vgvassilev (approved):
Thank you @fsfod. This lgtm!

PR #1045, opened 2024-08-18
 Add primitive support of reverse-mode constructor custom derivatives

infinite-void-16:
This commit adds primitive support of reverse-mode custom derivatives
for constructors.

Constructors can thought of like a special function. A function which also
constructs an object. Therefore, differentiation of a constructor is
very similar to the differentiation of a function call. Please note that
this includes innocent looking initialization such as:

```
SomeClass C1 = C2;
```

This would involve differentiation of copy-constructor of `SomeClass`.

Let's see a concrete example of how differentiation of a constructor call
looks like:

Original code:

```cpp
SomeClass c(u, v, w);
```

Derivative code:

```cpp
// forward-pass
clad::ValueAndAdjoint<SomeClass, SomeClass> _t0 =
  constructor_reverse_forw(clad::ConstructorReverseForwTag<SomeClass>,
    u, v, w, _d_u, _d_v, _d_w);
SomeClass _d_c(_t0.adjoint);
SomeClass c(_t0.value);

// reverse-pass
{
  double _r0 = 0;
  double _r1 = 0;
  double _r2 = 0;
  constructor_pullback(&c, u, v, w, &_d_c, &_r0, &_r1, &_r2);
  _d_u += _r0;
  _d_v += _r1;
  _d_w += _r2;
}
```

Please note the use of `clad::ConstructorReverseForwTag<T>` for
`constructor_reverse_forw` function. The reasoning and motivation for
this is the same as we have for `clad::ConstructorPushforwardTag<T>`

Reverse-mode custom derivatives for constructor can be
specified as:

```cpp
namespace clad {
namespace class_functions {
clad::ValueAndAdjoint<SomeClass, SomeClass>
constructor_reverse_forw(clad::ConstructorReverseForwTag<SomeClass>,
  ...) {
  // ...
  // ...
}

void constructor_pullback(SomeClass *c, ..., SomeClass *d_c, ...) {
  // ...
}
```

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -16,6 +16,7 @@
 #include "clang/Sema/Sema.h"
 
 #include <array>
+#include <limits>
 #include <memory>
 #include <stack>
 #include <unordered_map>
@@ -101,6 +102,11 @@ namespace clad {
     clang::Expr* BuildCallToCustomForwPassFn(
         const clang::FunctionDecl* FD, llvm::ArrayRef<clang::Expr*> primalArgs,
         llvm::ArrayRef<clang::Expr*> derivedArgs, clang::Expr* baseExpr);
+
+    // clang::Expr* BuildCallToCustomForwPassConstructor(
+    //     const clang::CXXConstructExpr* CE,
+    //     llvm::ArrayRef<clang::Expr*> clonedArgs,
+    //     llvm::ArrayRef<clang::Expr*> derivedArgs);
 
   public:
     using direction = rmv::direction;
@@ -689,6 +695,35 @@ namespace clad {
     }
 
     void PopSwitchStmtInfo() { m_SwitchStmtsData.pop_back(); }
+
+    struct ConstructorPullbackCallInfo {
+      clang::CallExpr* pullbackCE = nullptr;
+      size_t thisAdjointArgIdx = std::numeric_limits<size_t>::max();
+      void updateThisParmArgs(clang::Expr* thisE, clang::Expr* dThisE) const;
+      ConstructorPullbackCallInfo() = default;
+      ConstructorPullbackCallInfo(clang::CallExpr* pPullbackCE,
+                                  size_t pThisAdjointArgIdx)
+          : pullbackCE(pPullbackCE), thisAdjointArgIdx(pThisAdjointArgIdx) {}
+
+      bool empty() const { return !pullbackCE; }
+    };
+
+    void setConstructorPullbackCallInfo(clang::CallExpr* pullbackCE,
+                                        size_t thisAdjointArgIdx) {
+      m_ConstructorPullbackCallInfo = {pullbackCE, thisAdjointArgIdx};
+    }
+
+    ConstructorPullbackCallInfo getConstructorPullbackCallInfo() {
+      return m_ConstructorPullbackCallInfo;
+    }
+
+    void resetConstructorPullbackCallInfo() {
+      m_ConstructorPullbackCallInfo = ConstructorPullbackCallInfo{};
+    }
+
+  private:
+    ConstructorPullbackCallInfo m_ConstructorPullbackCallInfo;
+    bool m_TrackConstructorPullbackInfo = false;
   };
 } // end namespace clad
 
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -103,11 +103,6 @@ namespace clad {
         const clang::FunctionDecl* FD, llvm::ArrayRef<clang::Expr*> primalArgs,
         llvm::ArrayRef<clang::Expr*> derivedArgs, clang::Expr* baseExpr);
 
-    // clang::Expr* BuildCallToCustomForwPassConstructor(
-    //     const clang::CXXConstructExpr* CE,
-    //     llvm::ArrayRef<clang::Expr*> clonedArgs,
-    //     llvm::ArrayRef<clang::Expr*> derivedArgs);
-
   public:
     using direction = rmv::direction;
     clang::Expr* dfdx() {
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -29,14 +29,14 @@
 #include "clang/Sema/Sema.h"
 #include "clang/Sema/SemaInternal.h"
 #include "clang/Sema/Template.h"
-
-#include "llvm/Support/SaveAndRestore.h"
-
-#include <algorithm>
 #include <clang/AST/DeclCXX.h>
 #include <clang/AST/ExprCXX.h>
 #include <clang/AST/OperationKinds.h>
 #include <clang/Sema/Ownership.h>
+
+#include "llvm/Support/SaveAndRestore.h"
+
+#include <algorithm>
 #include <numeric>
 
 #include "clad/Differentiator/CladUtils.h"
@@ -4142,7 +4142,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     for (const auto* arg : CE->arguments()) {
       StmtDiff argDiff{};
       Expr* adjointArg = nullptr;
-      if (utils::IsReferenceOrPointerArg(arg)) {
+      if (utils::IsReferenceOrPointerArg(arg->IgnoreParenImpCasts())) {
         argDiff = Visit(arg);
         adjointArg = argDiff.getExpr_dx();
       } else {
@@ -4172,7 +4172,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         reverseForwAdjointArgs.push_back(adjointArg);
         adjointArgs.push_back(adjointArg);
       } else {
-        if (arg->getType()->isReferenceType())
+        if (utils::IsReferenceOrPointerArg(arg->IgnoreParenImpCasts()))
           reverseForwAdjointArgs.push_back(adjointArg);
         else
           reverseForwAdjointArgs.push_back(getZeroInit(arg->getType()));
@@ -4181,8 +4181,8 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       }
       primalArgs.push_back(argDiff.getExpr());
     }
-    // Try to create a pullback constructor call
 
+    // Try to create a pullback constructor call
     llvm::SmallVector<Expr*, 4> pullbackArgs;
     QualType recordType =
         m_Context.getRecordType(CE->getConstructor()->getParent());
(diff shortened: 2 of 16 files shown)


vgvassilev:
@infinite-void-16, can you rebase this PR?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -33,6 +33,10 @@
 #include "llvm/Support/SaveAndRestore.h"
 
 #include <algorithm>
+#include <clang/AST/DeclCXX.h>
+#include <clang/AST/ExprCXX.h>
+#include <clang/AST/OperationKinds.h>
+#include <clang/Sema/Ownership.h>
```
vgvassilev, line 39: We should probably move this block with the block of clang includes.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2007,8 +2013,18 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
-        !returnType.getNonReferenceType().isConstQualified()) {
+    if (Expr* customForwardPassCE = BuildCallToCustomForwPassFn(
+            FD, CallArgs, DerivedCallOutputArgs, baseExpr)) {
+      if (!utils::isNonConstReferenceType(returnType))
+        return StmtDiff{customForwardPassCE};
+      auto* callRes = StoreAndRef(customForwardPassCE);
+      auto* resValue =
+          utils::BuildMemberExpr(m_Sema, getCurrentScope(), callRes, "value");
+      auto* resAdjoint =
+          utils::BuildMemberExpr(m_Sema, getCurrentScope(), callRes, "adjoint");
```
vgvassilev, line 2024: We should probably spell out the types here.

vgvassilev:
@infinite-void-16, can you rebase this PR?

infinite-void-16:
@vgvassilev I have rebased the pull-request.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4100,31 +4124,150 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+        QualType dArgTy = getNonConstType(arg->getType(), m_Context, m_Sema);
+        VarDecl* dArgDecl = BuildVarDecl(dArgTy, "_r", getZeroInit(dArgTy));
+        prePullbackCallStmts.push_back(BuildDeclStmt(dArgDecl));
+        adjointArg = BuildDeclRef(dArgDecl);
+        argDiff = Visit(arg, BuildDeclRef(dArgDecl));
+      }
+
+      if (utils::isArrayOrPointerType(arg->getType())) {
+        reverseForwAdjointArgs.push_back(adjointArg);
+        adjointArgs.push_back(adjointArg);
```
gojakuch, line 4173: is there a quick way to test this so that Codecov doesn't complain?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4100,31 +4124,150 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+      } else {
+        if (arg->getType()->isReferenceType())
+          reverseForwAdjointArgs.push_back(adjointArg);
```
gojakuch, line 4176: same here

gojakuch (commented):
@infinite-void-16 that seems great, can you run git-clang-format on it so that it passes all the checks? LGTM though

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: Ping.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4381,4 +4539,27 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+  //       GetCladConstructorReverseForwTagOfType(
+  //           CE->getType().withoutLocalFastQualifiers());
+  //   Expr* constructorReverseForwTagArg = m_Sema.BuildCXXTypeConstructExpr(
+  //       m_Context.getTrivialTypeSourceInfo(constructorReverseForwTagT),
+  //       noLoc, MultiExprArg{}, noLoc, /*ListInitialization=*/false).get();
+  //   reverseForwCallArgs.push_back(constructorReverseForwTagArg);
+  //   reverseForwCallArgs.append(clonedArgs.begin(), clonedArgs.end());
+  //   reverseForwCallArgs.append(derivedArgs.begin(), derivedArgs.end());
+  //   std::string reverseForw
+  // }
```
vgvassilev, line 4558: Should we keep this?

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -102,6 +103,11 @@ namespace clad {
         const clang::FunctionDecl* FD, llvm::ArrayRef<clang::Expr*> primalArgs,
         llvm::ArrayRef<clang::Expr*> derivedArgs, clang::Expr* baseExpr);
 
+    // clang::Expr* BuildCallToCustomForwPassConstructor(
+    //     const clang::CXXConstructExpr* CE,
+    //     llvm::ArrayRef<clang::Expr*> clonedArgs,
+    //     llvm::ArrayRef<clang::Expr*> derivedArgs);
```
vgvassilev, line 109: Do we need this commented code?

vgvassilev (commented):
Looks like we are heading at the right direction.

gojakuch:
@vgvassilev I've addressed all the comments and formatted the code. I've also noticed that the check for reference arguments wouldn't have worked the way it was implemented here, so I changed it a bit. let's see if this passes all the checks. you can squash the commits

lib/Differentiator/CladUtils.cpp
```diff
@@ -304,6 +304,23 @@ namespace clad {
       return false;
     }
 
+    bool isReference(const Expr* e) {
+      if (!e)
+        return false;
+      const Expr* expr = e->IgnoreParenImpCasts();
```
vgvassilev, line 310:
```suggestion
      const Expr* expr = e->IgnoreParenImpCasts();
      struct RefChecker: public RecursiveASTVisitor<RefChecker> {
        QualType DRETy;
        bool VisitDeclRefExpr(DeclRefExpr *DRE) {
           DRETy = DRE->getType();
           return false; //abort visitation on first found.
         } DREChecker;
         DRECheker.Visit(E);
         DREChecker.DRETy...
      }
```

lib/Differentiator/CladUtils.cpp
```diff
@@ -305,20 +305,18 @@ namespace clad {
...
-              return true;
-      return false;
+      assert(e && "clad::utils::isReference check failed because the "
+                  "expression was null");
+      struct RefChecker : public RecursiveASTVisitor<RefChecker> {
+        bool isRef = false;
+        bool VisitDeclRefExpr(const DeclRefExpr* DRE) {
+          if (const ValueDecl* decl = DRE->getDecl())
+            isRef = decl->getType()->isReferenceType();
+          return false; // abort visitation on first found.
```
vgvassilev, line 315:
```suggestion
          if (const ValueDecl* decl = DRE->getDecl()) {
            isRef = decl->getType()->isReferenceType();
            return false; // abort visitation on first found.
          }
          return true;
```

vgvassilev: I wonder if we can't use `!E->isLValue()`?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4100,31 +4124,150 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+        prePullbackCallStmts.push_back(BuildDeclStmt(dArgDecl));
+        adjointArg = BuildDeclRef(dArgDecl);
+        argDiff = Visit(arg, BuildDeclRef(dArgDecl));
+      }
+
+      if (utils::isArrayOrPointerType(arg->getType())) {
+        reverseForwAdjointArgs.push_back(adjointArg);
+        adjointArgs.push_back(adjointArg);
+      } else {
+        if (utils::isReference(arg))
```
vgvassilev, line 4175:
```suggestion
        if (arg->IgnoreParenImpCasts()->getType()->isReferenceType())
```
I wonder if that won't just work?

gojakuch: we still need to get the decl, because the expression itself can't tell if it's a reference or not for some reason, so I guess defining a function is more general. plus, this function can then be reused, as this seems to be a common task

gojakuch:
I've actually figured out a better solution now and it also covers another issue I've spotted here.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2808,6 +2814,16 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
           VDDerivedInit = initDiff.getForwSweepExpr_dx();
       }
 
+      if (VD->getType()->isStructureOrClassType()) {
```
vgvassilev, line 2817:
```suggestion
      if (VDType->isStructureOrClassType()) {
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2869,9 +2885,10 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       }
 
       if (VD->getInit()) {
-        if (isa<CXXConstructExpr>(VD->getInit()))
-          initDiff = Visit(VD->getInit());
-        else
+        if (VD->getType()->isStructureOrClassType()) {
```
vgvassilev, line 2888:
```suggestion
        if (VDType->isStructureOrClassType()) {
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4100,31 +4124,150 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+        // SomeClass_pullback(c, u, ..., &_d_c, &_r0, ...);
+        // _d_u += _r0;
+        QualType dArgTy = getNonConstType(arg->getType(), m_Context, m_Sema);
+        VarDecl* dArgDecl = BuildVarDecl(dArgTy, "_r", getZeroInit(dArgTy));
+        prePullbackCallStmts.push_back(BuildDeclStmt(dArgDecl));
+        adjointArg = BuildDeclRef(dArgDecl);
+        argDiff = Visit(arg, BuildDeclRef(dArgDecl));
+      }
+
+      if (utils::isArrayOrPointerType(arg->getType())) {
```
vgvassilev, line 4171: `arg->getType()` seems repeating, can we move it into a temp?

gojakuch: done

vgvassilev (approved):
LGTM!

PR #1018, opened 2024-08-01
Add support for std::initializer_list in the reverse mode

PetroZarytskyi:
This PR adds support for ``std::initializer_list`` in the reverse mode.
Also, this PR adds support for non-array type ranges and removes redundant assignments by simplifying the logic for range-based for loops. Apart from that, it adds support for ``std::begin``/``std::end``, and introduces ``begin``/``end`` functions to ``clad::array``.
Fixes #830.

PR #1046, opened 2024-08-19
Desugar the type before analyzing it in TBR

PetroZarytskyi:
This is necessary to analyze sugared types like ``std::vector<double>::pointer``. 
Unblocks #1037 and #1018.


vgvassilev (approved):
LGTM!

vgvassilev:
I can merge it with codecov complaining. @PetroZarytskyi will open an issue to track the missing test coverage because it also misses a feature.

PR #1047, opened 2024-08-19
Support pointer-valued functions in the reverse mode

PetroZarytskyi:
This PR enables support of pointer-valued functions in the reverse mode by creating ``_forw`` functions for them (as we already do with reference-type functions).

Note: the first commit simplifies how we pass adjoints to ``_forw`` functions. The types used to be converted from ref-types to the corresponding pointer types. There's no need to change the type because the function represents the forward pass. Moreover, non-ref types (like pointer types in this PR) were not handed properly.


vgvassilev (approved):
Lgtm!

Issue #855, opened 2024-04-08
TBR analysis doesn't include the for loop's step in the forward pass when there's a statement inside with a body of its own


kchristin22:
```cpp
double f(double val) {
  for (int i=1; i<5; ++i) {
    val += val;
  }
  return val;
}
```
Gradient:

* Without TBR:
```cpp

void f_grad(double val, double *_d_val) {
    unsigned long _t0;
    int _d_i = 0;
    int i = 0;
    clad::tape<double> _t1 = {};
    _t0 = 0;
    for (i = 1; i < 5; ++i) {
        _t0++;
        clad::push(_t1, val);
        val += val;
    }
    goto _label0;
  _label0:
    *_d_val += 1;
    for (; _t0; _t0--) {
        --i;
        {
            val = clad::pop(_t1);
            double _r_d0 = *_d_val;
            *_d_val += _r_d0;
        }
    }
}
```
* With TBR:
```cpp
void f_grad(double val, double *_d_val) {
    unsigned long _t0;
    int _d_i = 0;
    int i = 0;
    _t0 = 0;
    for (i = 1; i < 5; ++i) {
        _t0++;
        val += val;
    }
    goto _label0;
  _label0:
    *_d_val += 1;
    for (; _t0; _t0--) {
        {
            double _r_d0 = *_d_val;
            *_d_val += _r_d0;
        }
    }
}
```

kchristin22:
Yes, here the result is not compromised, but if it depended on the value of i (e.g. val += i) then it would be wrong. Hence, the increment of i should be included similarly to the case where TBR is not enabled.

PetroZarytskyi:
Do you mean ``--i`` in the reverse pass? It's not included because we don't need the value of ``i``. It is the purpose of TBR analysis to detect what values we don't need to restore to compute the derivatives. Adding ``val += i`` would not change anything because the transformation is linear. However, if we differentiate a function where the value of ``i`` is necessary, we will still have ``--i``. e.g.
```
double f(double val) {
    for (int i=1; i<5; ++i) {
        val += val * i;
    }
    return val;
}
```
gives us
```
void f_grad(double val, double *_d_val) {
    unsigned long _t0;
    int _d_i = 0;
    int i = 0;
    clad::tape<double> _t1 = {};
    _t0 = 0;
    for (i = 1; i < 5; ++i) {
        _t0++;
        clad::push(_t1, val);
        val += val * i;
    }
    goto _label0;
  _label0:
[...]

PR #1051, opened 2024-08-21
Fix collectDataFromPredecessors to collect data from the passed branch

ovdiiuv:
Before the fix, we didn't collect data from varsData at all, and it wasn't merged afterward either. Now, we start collecting data from varsData, ensuring the final data of the CFG block is correct.

For example it used to fail in cases like below, when the LCA is a predecessor of the mergeData in the merge function `TBRAnalyzer::merge(VarsData* targetData, VarsData* mergeData)`.

```cpp
double fn(double x){
  double res = 0;
  double g = 0;
  if(g){
    res+=g*x;
  }
  return res;
}
```
Before this PR the `collectedMergeData` was empty and after it contains data about {res, g, x}, as expected.

Fixes:#855


vgvassilev:
@ovdiiuv, looks good but can you improve the commit message and the pull request title?

PR #1056, opened 2024-08-23
Add support for `Kokkos::parallel_reduce` in the fwd mode

gojakuch:


gojakuch:
I'm marking this as ready for review, although I'll try to resolve a couple more things here before merging. this covers most possible cases (everything not covered by this PR is tracked in the comments in corresponding parts of the code for now). there are multiple commits in this PR, because there were several things that had to be done here, but since it's all Kokkos-related and reviews normally take some time, I put them all into this single PR.

gojakuch:
the reason I haven't covered all the possible cases of `parallel_reduce` here yet is that I have no idea how to make it work with arbitrary reduce parameters in the `parallel_reduce`, since we need to define a pushforward that would take twice as many parameters of correct types for this case. not sure how to go about this for now

vgvassilev:
@kliegeois, can you take a look?

kliegeois:
I will look into this PR very soon.

kliegeois:
I tried the PR locally and it generated I think what is expected. I run into a segfault though but I think this is due to a version mismatch.
Otherwise the PR looks good to me, thanks!

Have you started looking at the lambda as well?

gojakuch:
@kliegeois thanks for the approval

> Have you started looking at the lambda as well?

yes, but lambdas need some general rework in Clad. that's on my list though. I've opened some issues to track progress on that and I'll get to it soon.

Issue #1049, opened 2024-08-20
Differentiating increment inside the if body fails with TBR on


ovdiiuv:
Differentiating
 
```cpp
double f(double val) {
  double res = 0;
  if(val)
    res++;
  return res;
}
```

with TBR on results in an error.

PR #1058, opened 2024-08-24
Fix braceless if differentiation in reverse mode

ovdiiuv:
As in #1049, the statement inside the braceless if isn't included, causing an error. To avoid that now we check if `thenDiff.getStmt_dx()` is empty, if so we create if statement with `!(orig_cond)` condition to ensure correct handling of `else if` conditions.

Fixes: #1049

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -878,10 +878,21 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         elseDiff.getStmt());
     addToCurrentBlock(Forward, direction::forward);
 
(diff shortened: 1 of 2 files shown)


vgvassilev:
Can you add a test to make codecov happy?

ovdiiuv:
> Can you add a test to make codecov happy?

Yes, I wanted to ask you about that. Should I include the generated code in the tests? Now it only checks if the values are right.

vgvassilev:
> > Can you add a test to make codecov happy?
> 
> Yes, I wanted to ask you about that. Should I include the generated code in the tests? Now it only checks if the values are right.

For this particular change you can add the test in `test/Analyses/TBR.cpp`.

vgvassilev (approved):
LGTM!

Issue #812, opened 2024-03-09
Examine clang-15 failure on CUDA Gradient script


kchristin22:
The test of `GradientCuda.cu` was indicated as expected to fail with clang-15. The changes applied to this script in #806 will probably not have any effect on that, so the problem remains. During development though it was shown that errors where produced when CUDA Toolkit and Clang versions were incompatible. 

TODO:

- [ ] Install a CUDA version that is old enough for clang-15 and is fully supported by your system
- [ ] Compile `GradientCuda.cu` with `-fdump-derived-fn` option --> expected result: the gradient function is printed on the terminal
- [ ] Validate the run-time results

vgvassilev:
Hi @kchristin22, is this issue still relevant? Was it not already addressed elsewhere?

vgvassilev:
clang-15 is less relevant, we have clang-20 and we do not seem to have such issues. Feel free to reopen it if needed.

PR #1059, opened 2024-08-25
Enable computation of CUDA global kernels derivative in reverse mode

### LINKED ISSUE #1036: Call to `clad::gradient` for CUDA kernels cannot be compiled for GPU

kchristin22:
This PR closes #1036. The CUDA kernels can now be an arg to a clad::gradient function and the derived kernels can be executed successfully. 

Details on issues faced and the final approach can be found here: https://github.com/vgvassilev/clad/issues/1036#issuecomment-2308891416.

[Old PR's](#1041)  approach involved the use of CUDA API libs to compile and load the computed kernel on the GPU.

### DIFF
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -132,6 +132,30 @@ inline CUDA_HOST_DEVICE unsigned int GetLength(const char* code) {
     return f(static_cast<Args>(args)...);
   }
 
+#ifdef __CUDACC__
+  template <bool EnablePadding, class... Rest, class F, class... Args,
+            class... fArgTypes,
+            typename std::enable_if<EnablePadding, bool>::type = true>
+  CUDA_HOST_DEVICE void
+  execute_with_default_args(list<Rest...>, F f, list<fArgTypes...>, dim3 grid,
+                            dim3 block, size_t shared_mem, cudaStream_t stream,
+                            Args&&... args) {
+    void* argPtrs[] = {(void*)&args..., (void*)static_cast<Rest>(nullptr)...};
+    cudaLaunchKernel((void*)f, grid, block, argPtrs, shared_mem, stream);
+  }
+
+  template <bool EnablePadding, class... Rest, class F, class... Args,
+            class... fArgTypes,
+            typename std::enable_if<!EnablePadding, bool>::type = true>
+  CUDA_HOST_DEVICE void
+  execute_with_default_args(list<Rest...>, F f, list<fArgTypes...>, dim3 grid,
+                            dim3 block, size_t shared_mem, cudaStream_t stream,
+                            Args&&... args) {
+    void* argPtrs[] = {(void*)&args...};
+    cudaLaunchKernel((void*)f, grid, block, argPtrs, shared_mem, stream);
+  }
+#endif
+
   // for executing member-functions
   template <bool EnablePadding, class... Rest, class ReturnType, class C,
             class Obj, class... Args, class... fArgTypes,
@@ -215,9 +239,33 @@ inline CUDA_HOST_DEVICE unsigned int GetLength(const char* code) {
         printf("CladFunction is invalid\n");
         return static_cast<return_type_t<F>>(return_type_t<F>());
       }
+      if (m_CUDAkernel) {
+        printf("Use execute_kernel() for global CUDA kernels\n");
+        return static_cast<return_type_t<F>>(return_type_t<F>());
+      }
       // here static_cast is used to achieve perfect forwarding
       return execute_helper(m_Function, static_cast<Args>(args)...);
     }
+
+#ifdef __CUDACC__
+    template <typename... Args, class FnType = CladFunctionType>
+    typename std::enable_if<!std::is_same<FnType, NoFunction*>::value,
+                            return_type_t<F>>::type
+    execute_kernel(dim3 grid, dim3 block, size_t shared_mem,
+                   cudaStream_t stream, Args&&... args) CUDA_HOST_DEVICE {
+      if (!m_Function) {
+        printf("CladFunction is invalid\n");
+        return static_cast<return_type_t<F>>(return_type_t<F>());
+      }
+      if (!m_CUDAkernel) {
+        printf("Use execute() for non-global CUDA kernels\n");
+        return static_cast<return_type_t<F>>(return_type_t<F>());
+      }
+
+      return execute_helper(m_Function, grid, block, shared_mem, stream,
+                            static_cast<Args>(args)...);
+    }
+#endif
 
     /// `Execute` overload to be used when derived function type cannot be
     /// deduced. One reason for this can be when user tries to differentiate
@@ -270,6 +318,20 @@ inline CUDA_HOST_DEVICE unsigned int GetLength(const char* code) {
           TakeNFirstArgs_t<sizeof...(Args), decltype(f)>{},
           static_cast<Args>(args)...);
       }
+
+#ifdef __CUDACC__
+      /// Helper function for executing non-member global derived functions.
+      template <class Fn, class... Args>
+      CUDA_HOST_DEVICE return_type_t<CladFunctionType>
+      execute_helper(Fn f, dim3 grid, dim3 block, size_t shared_mem,
+                     cudaStream_t stream, Args&&... args) {
+        // `static_cast` is required here for perfect forwarding.
+        return execute_with_default_args<EnablePadding>(
+            DropArgs_t<sizeof...(Args), F>{}, f,
+            TakeNFirstArgs_t<sizeof...(Args), decltype(f)>{}, grid, block,
+            shared_mem, stream, static_cast<Args>(args)...);
+      }
+#endif
 
       /// Helper functions for executing member derived functions.
       /// If user have passed object explicitly, then this specialization will
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -171,12 +171,13 @@ inline CUDA_HOST_DEVICE unsigned int GetLength(const char* code) {
     CladFunctionType m_Function;
     char* m_Code;
     FunctorType *m_Functor = nullptr;
+    bool m_CUDAkernel = false;
 
   public:
-    CUDA_HOST_DEVICE CladFunction(CladFunctionType f,
-                                  const char* code,
-                                  FunctorType* functor = nullptr)
-        : m_Functor(functor) {
+    CUDA_HOST_DEVICE CladFunction(CladFunctionType f, const char* code,
+                                  FunctorType* functor = nullptr,
+                                  bool CUDAkernel = false)
+        : m_Functor(functor), m_CUDAkernel(CUDAkernel) {
       assert(f && "Must pass a non-0 argument.");
       if (size_t length = GetLength(code)) {
         m_Function = f;
@@ -397,10 +398,10 @@ inline CUDA_HOST_DEVICE unsigned int GetLength(const char* code) {
       annotate("G"))) CUDA_HOST_DEVICE
   gradient(F f, ArgSpec args = "",
            DerivedFnType derivedFn = static_cast<DerivedFnType>(nullptr),
-           const char* code = "") {
-      assert(f && "Must pass in a non-0 argument");
-      return CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>, true>(
-          derivedFn /* will be replaced by gradient*/, code);
+           const char* code = "", bool CUDAkernel = false) {
+    assert(f && "Must pass in a non-0 argument");
+    return CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>, true>(
+        derivedFn /* will be replaced by gradient*/, code, nullptr, CUDAkernel);
   }
 
   /// Specialization for differentiating functors.
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -177,9 +177,6 @@ namespace clad {
   void DiffRequest::updateCall(FunctionDecl* FD, FunctionDecl* OverloadedFD,
                                Sema& SemaRef) {
     CallExpr* call = this->CallContext;
-    // Index of "code" parameter:
-    auto codeArgIdx = static_cast<int>(call->getNumArgs()) - 1;
-    auto derivedFnArgIdx = codeArgIdx - 1;
 
     assert(call && "Must be set");
     assert(FD && "Trying to update with null FunctionDecl");
@@ -191,6 +188,19 @@ namespace clad {
     ASTContext& C = SemaRef.getASTContext();
 
     FunctionDecl* replacementFD = OverloadedFD ? OverloadedFD : FD;
+
+    // Index of "CUDAkernel" parameter:
+    int numArgs = static_cast<int>(call->getNumArgs());
+    if (numArgs > 4) {
+      auto kernelArgIdx = numArgs - 1;
+      auto cudaKernelFlag = new (C) CXXBoolLiteralExpr(
+          replacementFD->hasAttr<CUDAGlobalAttr>(), C.BoolTy, noLoc);
+      call->setArg(kernelArgIdx, cudaKernelFlag);
+      numArgs--;
+    }
+    auto codeArgIdx = numArgs - 1;
+    auto derivedFnArgIdx = numArgs - 2;
+
     // Create ref to generated FD.
     DeclRefExpr* DRE =
         DeclRefExpr::Create(C, oldDRE->getQualifierLoc(), noLoc, replacementFD,
--- a/lib/Differentiator/VisitorBase.cpp
+++ b/lib/Differentiator/VisitorBase.cpp
@@ -616,6 +616,13 @@ namespace clad {
     if (auto derMethod = dyn_cast<CXXMethodDecl>(FD)) {
       call = BuildCallExprToMemFn(derMethod, argExprs, useRefQualifiedThisObj);
     } else {
+      // If the function is a global kernel, we need to transform it
+      // into a device function when calling it inside the overload function
+      // which is the final global kernel returned.
+      if (FD->hasAttr<CUDAGlobalAttr>()) {
+        FD->dropAttr<CUDAGlobalAttr>();
+        FD->addAttr(CUDADeviceAttr::CreateImplicit(m_Context));
+      }
       Expr* exprFunc = BuildDeclRef(FD, SS);
       call = m_Sema
                  .ActOnCallExpr(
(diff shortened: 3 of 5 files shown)


unittests/Misc/CallDeclOnly.cpp
```diff
@@ -59,10 +59,10 @@ double sq(double x) { return x * x; }
 double wrapper2(double* params) { return sq(params[0]); }
 
 TEST(CallDeclOnly, CheckCustomDiff) {
-  auto grad = clad::hessian(wrapper2, "params[0]");
+  auto hess = clad::hessian(wrapper2, "params[0]");
   double x = 4.0;
   double dx = 0.0;
-  grad.execute(&x, &dx);
+  hess.execute(&x, &dx);
```
vgvassilev, line 65: This seems unrelated to the current PR. Can this change go in a separate PR?

kchristin22: I changed it here because my changes caused it to fail without this fix. The name had to change because there was already another object above named `grad` that used `clad::gradient` which now has a different signature than `clad::hessian` and returns an extra arg. Should I open a new PR for this?

kchristin22:
> ## [Codecov](https://app.codecov.io/gh/vgvassilev/clad/pull/1059?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Vassil+Vassilev) Report
> 
> Attention: Patch coverage is `89.47368%` with `2 lines` in your changes missing coverage. Please review.
> 
> > Project coverage is 94.32%. Comparing base [(`6f4b081`)](https://app.codecov.io/gh/vgvassilev/clad/commit/6f4b0811d320e2a910b67ebee9eab24f3fbda3e9?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Vassil+Vassilev) to head [(`610850f`)](https://app.codecov.io/gh/vgvassilev/clad/commit/610850ff502bc7daaaea7f85c11e21a9b47f4332?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Vassil+Vassilev).
> > Report is 4 commits behind head on master.
> 
> [Files](https://app.codecov.io/gh/vgvassilev/clad/pull/1059?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Vassil+Vassilev) 	Patch % 	Lines
> [include/clad/Differentiator/Differentiator.h](https://app.codecov.io/gh/vgvassilev/clad/pull/1059?src=pr&el=tree&filepath=include%2Fclad%2FDifferentiator%2FDifferentiator.h&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Vassil+Vassilev#diff-aW5jbHVkZS9jbGFkL0RpZmZlcmVudGlhdG9yL0RpZmZlcmVudGlhdG9yLmg=) 	75.00% 	[2 Missing ⚠️ ](https://app.codecov.io/gh/vgvassilev/clad/pull/1059?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Vassil+Vassilev)
> Additional details and impacted files
> 
> [Impacted file tree graph](https://app.codecov.io/gh/vgvassilev/clad/pull/1059?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Vassil+Vassilev)
> 
> ```diff
> @@            Coverage Diff             @@
> ##           master    #1059      +/-   ##
> ==========================================
> - Coverage   94.34%   94.32%   -0.02%     
> ==========================================
>   Files          55       55              
>   Lines        8311     8323      +12     
> ==========================================
> + Hits         7841     7851      +10     
> - Misses        470      472       +2     
> ```
> 
> [Files](https://app.codecov.io/gh/vgvassilev/clad/pull/1059?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Vassil+Vassilev) 	Coverage Δ 	
> [lib/Differentiator/DiffPlanner.cpp](https://app.codecov.io/gh/vgvassilev/clad/pull/1059?src=pr&el=tree&filepath=lib%2FDifferentiator%2FDiffPlanner.cpp&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Vassil+Vassilev#diff-bGliL0RpZmZlcmVudGlhdG9yL0RpZmZQbGFubmVyLmNwcA==) 	`98.76% <100.00%> (+0.01%)` 	⬆️
> [lib/Differentiator/VisitorBase.cpp](https://app.codecov.io/gh/vgvassilev/clad/pull/1059?src=pr&el=tree&filepath=lib%2FDifferentiator%2FVisitorBase.cpp&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Vassil+Vassilev#diff-bGliL0RpZmZlcmVudGlhdG9yL1Zpc2l0b3JCYXNlLmNwcA==) 	`96.88% <100.00%> (+0.01%)` 	⬆️
> [include/clad/Differentiator/Differentiator.h](https://app.codecov.io/gh/vgvassilev/clad/pull/1059?src=pr&el=tree&filepath=include%2Fclad%2FDifferentiator%2FDifferentiator.h&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Vassil+Vassilev#diff-aW5jbHVkZS9jbGFkL0RpZmZlcmVudGlhdG9yL0RpZmZlcmVudGlhdG9yLmg=) 	`80.00% <75.00%> (-3.34%)` 	⬇️
> [Files](https://app.codecov.io/gh/vgvassilev/clad/pull/1059?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Vassil+Vassilev) 	Coverage Δ 	
> [lib/Differentiator/DiffPlanner.cpp](https://app.codecov.io/gh/vgvassilev/clad/pull/1059?src=pr&el=tree&filepath=lib%2FDifferentiator%2FDiffPlanner.cpp&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Vassil+Vassilev#diff-bGliL0RpZmZlcmVudGlhdG9yL0RpZmZQbGFubmVyLmNwcA==) 	`98.76% <100.00%> (+0.01%)` 	⬆️
> [lib/Differentiator/VisitorBase.cpp](https://app.codecov.io/gh/vgvassilev/clad/pull/1059?src=pr&el=tree&filepath=lib%2FDifferentiator%2FVisitorBase.cpp&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Vassil+Vassilev#diff-bGliL0RpZmZlcmVudGlhdG9yL1Zpc2l0b3JCYXNlLmNwcA==) 	`96.88% <100.00%> (+0.01%)` 	⬆️
> [include/clad/Differentiator/Differentiator.h](https://app.codecov.io/gh/vgvassilev/clad/pull/1059?src=pr&el=tree&filepath=include%2Fclad%2FDifferentiator%2FDifferentiator.h&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Vassil+Vassilev#diff-aW5jbHVkZS9jbGFkL0RpZmZlcmVudGlhdG9yL0RpZmZlcmVudGlhdG9yLmg=) 	`80.00% <75.00%> (-3.34%)` 	⬇️

I have added such a test in `GradientKernels.cu` to cover these two lines indicated:
`test.execute(d_a, d_square);` 
But it doesn't trace it. Does it need to be included in an already present test file?

unittests/Misc/CallDeclOnly.cpp
vgvassilev: Ha, interesting. I am not sure how that worked... Yes, probably worth being a separate PR.

vgvassilev:
> > I need to figure out how to run the cuda tests, too.
> 
> You mean locally?

No, adding a bot that has a cuda device to run the code we develop. That's an action item on me :)

include/clad/Differentiator/Differentiator.h
```diff
@@ -128,6 +128,30 @@ inline CUDA_HOST_DEVICE unsigned int GetLength(const char* code) {
     return f(static_cast<Args>(args)...);
   }
 
+#ifdef __CUDACC__
+  template <bool EnablePadding, class... Rest, class F, class... Args,
+            class... fArgTypes,
+            typename std::enable_if<EnablePadding, bool>::type = true>
+  CUDA_HOST_DEVICE void
+  execute_with_default_args(list<Rest...>, F f, list<fArgTypes...>, dim3 grid,
+                            dim3 block, size_t shared_mem, cudaStream_t stream,
```
vgvassilev, line 137:
```suggestion
                            dim3 block, size_t shared_mem, cudaStream_t stream,
```
Can we wrap that in a CUDA-specific macro, say CUDA_ARGS, which gets expanded only in cuda mode.

This will allow you to wrap the cudaLaunchKernel and friends with `#ifdef __CUDACC__` and allow duplicating the overly complicated templates already...

vgvassilev: Alternatively, we can go even deeper and assume that in cuda mode `Args` has `size_t shared_mem, cudaStream_t stream` and extract it before forwarding the rest.

test/CUDA/GradientKernels.cu
```diff
@@ -0,0 +1,67 @@
...
+// RUN: %cladclang_cuda -I%S/../../include %s -xc++ %cudasmlevel \
+// RUN: --cuda-path=%cudapath -L/usr/local/cuda/lib64 -lcudart_static \
+// RUN: -L%cudapath/lib64/stubs \
+// RUN: -ldl -lrt -pthread -lm -lstdc++ -lcuda -lnvrtc
+
+// REQUIRES: cuda-runtime
+
+// expected-no-diagnostics
+
+// XFAIL: clang-15
```
parth-07, line 13: Why is it XFAIL for clang-15?

include/clad/Differentiator/Differentiator.h
```diff
@@ -210,10 +235,34 @@ inline CUDA_HOST_DEVICE unsigned int GetLength(const char* code) {
         printf("CladFunction is invalid\n");
         return static_cast<return_type_t<F>>(return_type_t<F>());
       }
+      if (m_CUDAkernel) {
+        printf("Use execute_kernel() for global CUDA kernels\n");
```
parth-07, line 239: We should probably assert-out if users use `execute` instead of `execute_kernel` for CUDA kernels. 'printf' seems to be too subtle for reporting such a big error. @vgvassilev What do you think?

include/clad/Differentiator/Differentiator.h
```diff
@@ -210,10 +235,34 @@ inline CUDA_HOST_DEVICE unsigned int GetLength(const char* code) {
...
+      }
       // here static_cast is used to achieve perfect forwarding
       return execute_helper(m_Function, static_cast<Args>(args)...);
     }
 
+#ifdef __CUDACC__
+    template <typename... Args, class FnType = CladFunctionType>
+    typename std::enable_if<!std::is_same<FnType, NoFunction*>::value,
+                            return_type_t<F>>::type
+    execute_kernel(dim3 grid, dim3 block, size_t shared_mem,
```
parth-07, line 250: Do you think it would be a good idea to provide overloads for `execute_kernel` that takes default values for `shared_mem` and `stream` args?

lib/Differentiator/VisitorBase.cpp
```diff
@@ -643,6 +643,13 @@ namespace clad {
     if (auto derMethod = dyn_cast<CXXMethodDecl>(FD)) {
       call = BuildCallExprToMemFn(derMethod, argExprs, useRefQualifiedThisObj);
     } else {
+      // If the function is a global kernel, we need to transform it
+      // into a device function when calling it inside the overload function
+      // which is the final global kernel returned.
+      if (FD->hasAttr<CUDAGlobalAttr>()) {
```
parth-07, line 649: This is great! 

But is `BuildCallExprToFunction` a right place for making this modification? We call `BuildCallExprToFunction` for each call to a function decl `FD`. If the function decl `FD` has to be called multiple times, then `BuildCallExprToFunction` is called multiple times as well. Hence, we check for `CUDAGlobalAttr` multiple times as well.

What do you think about making this change in `ReverseModeVisitor::Derive` just after where we create the derivative `FunctionDecl` and add the attributes?

```
DeclWithContext result = m_Builder.cloneFunction(
        m_DiffReq.Function, *this, DC, noLoc, name, gradientFunctionType);
FunctionDecl* gradientFD = result.first;
m_Derivative = gradientFD;
```

(attributes are added in `cloneFunction` method.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -191,6 +188,19 @@ namespace clad {
     ASTContext& C = SemaRef.getASTContext();
 
     FunctionDecl* replacementFD = OverloadedFD ? OverloadedFD : FD;
+
+    // Index of "CUDAkernel" parameter:
+    int numArgs = static_cast<int>(call->getNumArgs());
+    if (numArgs > 4) {
+      auto kernelArgIdx = numArgs - 1;
+      auto* cudaKernelFlag = new (C) CXXBoolLiteralExpr(
```
parth-07, line 196: Can you please try to use higher-level `Sema::ActOnCXXBoolLiteral` API instead of the lower-level `CXXBoolLiteralExpr` class here? We prefer to use `Sema` APIs where possible because they tend to be more robust.

parth-07 (changes requested):
Thank you for this pull-request. Overall, this looks great!

include/clad/Differentiator/Differentiator.h
vgvassilev: Is that a programmatic error or user error. We can use assert/abort if the error is programmer error.

unittests/Misc/CallDeclOnly.cpp
kchristin22: Ok, will open a new one

test/CUDA/GradientKernels.cu
kchristin22: It was specified like so in the CUDA gradient test and copied it just in case. I can't install an older CUDA version compatible with clang-15 on my laptop to investigate further. There's already [this relevant issue](https://github.com/vgvassilev/clad/issues/812) for this.

include/clad/Differentiator/Differentiator.h
kchristin22: It is a user error basically. They should use the appropriate `execute` function depending on whether their function is a CUDA kernel or not.

lib/Differentiator/DiffPlanner.cpp
kchristin22: Done, thanks for the explanation!

include/clad/Differentiator/Differentiator.h
kchristin22: Sure, I can use the default values of 0 and nullptr

test/CUDA/GradientKernels.cu
parth-07: Oh, okay. Sounds good then.

include/clad/Differentiator/Differentiator.h
kchristin22: Done, but I needed to use constexpr for this, which is c++17. So cuda tests cannot run with an older c++ version. @vgvassilev is that too big of a problem?

include/clad/Differentiator/Differentiator.h
kchristin22: Done

lib/Differentiator/VisitorBase.cpp
parth-07: Please move it to `ReverseModeVisitor::Derive`.

The kernels which are differentiated using the forward-mode would not go through the function `VisitorBase::BuildCallExprToFunction`. This function is called for creating a call expression. It is not necessarily needed for differentiating a function. It is called for the reverse-mode because the gradient overload makes a function call to the actual gradient function.

kchristin22: Oh ok, sure then.

include/clad/Differentiator/Differentiator.h
parth-07: Why do we need `constexpr` for having default values for `shared_mem` and `stream`?

lib/Differentiator/VisitorBase.cpp
kchristin22: I moved it inside CreateGradientOverload, cause otherwise m_Derivative should be a global kernel (when an overload should not be created)

include/clad/Differentiator/Differentiator.h
kchristin22: Since the signature for execute_kernel with default args is: `dim3, dim3, Args&&...`
the function chosen when the args should not be default was the one **above** instead of the correct: `dim3, dim3, size_t, cudaStream_t, Args&&... `

Hence, I kept only the signature mentioned first and check if a cudaStream_t arg is included in Args, otherwise I use default values to call execute_with_default_args

lib/Differentiator/VisitorBase.cpp
parth-07: Okay, sounds good.

parth-07 (approved):
Looks good to me.

parth-07:
@kchristin22 Can you please rebase the PR on top of master?

include/clad/Differentiator/Differentiator.h
kchristin22: @vgvassilev I think it's better to not require the grid parameters to be evaluated at compile time. CUDA doesn't require that in general. So, we can't do this:
`execute_kernel<grid, block, shared_mem, stream>(...)`

PR #1061, opened 2024-08-26
Provide pushforward methods for `Kokkos::View` indexing

gojakuch:
Previously, we relied on automatically generated pushforwards for these operator calls, but this solution is way safer and should work for more machines and Kokkos versions.

### DIFF
--- a/include/clad/Differentiator/KokkosBuiltins.h
+++ b/include/clad/Differentiator/KokkosBuiltins.h
@@ -28,6 +28,75 @@ constructor_pushforward(
           Kokkos::View<DataType, ViewParams...>(
               "_diff_" + name, idx0, idx1, idx2, idx3, idx4, idx5, idx6, idx7)};
 }
+
+/// View indexing
+template <typename View, typename Idx>
+inline clad::ValueAndPushforward<typename View::reference_type,
+                                 typename View::reference_type>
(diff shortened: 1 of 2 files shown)


kliegeois:
This looks great but I would add a last version with 8 arguments as views can have up to 8 dimensions. Thanks!

kliegeois:
@gojakuch now that I am thinking about it, the code of the PR needs to be changed.

In Kokkos, the index type is templated with a different template for each indices. Example:
```
  // rank 4
  template <typename I0, typename I1, typename I2, typename I3>
  KOKKOS_INLINE_FUNCTION constexpr size_type operator()(I0 const& i0,
                                                        I1 const& i1,
                                                        I2 const& i2,
                                                        I3 const& i3) const {
    return i0 + m_dim.N0 * (i1 + m_dim.N1 * (i2 + m_dim.N2 * i3));
  }
```
The merged code works only if `I0`, ... `I3` are the same.
Could you fix that and add a test that uses a view indexing of at least 2 dimensions where one is a `int` and one is a `size_t` index? Thanks!

gojakuch:
@kliegeois I see. I'll open a PR today then

kliegeois:
Thanks a lot @gojakuch !

PR #1062, opened 2024-08-28
Differentiate array parameters in forward mode

### LINKED ISSUE #1035: Unnecessary warning when differentiating a function that mutates its array arg
```
double example(double x, double *y)
{
    y[0] = 3;
    return x * y[0];
}

int main()
{
    double x = 10.0;
    double y[3] = {7.0, 4.6, 6.3};
    auto ex = clad::differentiate(example, "x");
    ex.dump();
    std::cout << ex.execute(x, y) << std::endl;
}
``` 
on Clad version 1.8~dev this gives a compilation warning: 
```
warning: derivative of an assignment attempts to assign to unassignable expr, assignment ignored
    y[0] = 3;
           ^
1 warning generated.
``` 
This warning is a little confusing and could be interpreted as the code not working as intended, when in reality it produces expected results.

PetroZarytskyi:
Currently, we don't create an adjoint of an array parameter in the forward mode because we don't know the size of the array. This means considering it non-differentiable, which is not always the case. This PR does two things:
1. If the parameter is a const size array (e.g. ``double[5]``), use the same type for the adjoint.
2. Otherwise the size is unknown so we have to extend the array dynamically before accessing members.

There are downsides to the approach in 2. : extending the array involves reallocating memory, which costs performance and will cause a segfault if the user stored a pointer to some of the array's elements. However, alternatives are banning array parameters or considering them non-differentiable as before, which will fail way more frequently and produce warnings at max.

Also, please note the function ``extend`` that I introduced is not the same as ``resize`` since the latter can shrink the array if its argument is less than the size of the array. ``extend`` only makes the array bigger.

Fixes #1035.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2616,20 +2616,6 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     if (!VDDerivedInit)
       VDDerivedInit = getZeroInit(VDType);
 
-    // `specialThisDiffCase` is only required for correctly differentiating
-    // the following code:
-    // ```
-    // Class _d_this_obj;
-    // Class* _d_this = &_d_this_obj;
-    // ```
-    // Computation of hessian requires this code to be correctly
-    // differentiated.
-    bool specialThisDiffCase = false;
-    if (const auto* MD = dyn_cast<CXXMethodDecl>(m_DiffReq.Function)) {
-      if (VDDerivedType->isPointerType() && MD->isInstance())
-        specialThisDiffCase = true;
-    }
-
     if (isRefType) {
       initDiff = Visit(VD->getInit());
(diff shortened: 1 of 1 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1605,6 +1605,29 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       return {call, callDiff};
     }
 
+    // FIXME: There is nothing special about the `clad::array::extend` function
+    // but we need to be able to differentiate it to prevent segfaults in
+    // hessians. Once we add support for methods that change their objects, this
+    // section should be removed.
+    if (FDName == "extend")
```
vgvassilev, line 1612: Can we not provide a pullback for this interface?

PetroZarytskyi: I thought this was not possible but I forgot we now support custom ``_forw`` functions. I fixed this. Thank you for pointing this out.

Issue #1070, opened 2024-09-02
Incorrect `std::vector` differentiation in the reverse mode


gojakuch:
```c++
#include "clad/Differentiator/Differentiator.h"
#include "clad/Differentiator/STLBuiltins.h"
#include <vector>
#include <iostream>

double fn(double x, double y) {
  std::vector<double> a;
  a.push_back(x);
  a.push_back(x);
  a[1] = x*x;
  return a[1];
}

int main(int argc, char* argv[]) {
    double dx=0, dy=0;
    auto df = clad::gradient(fn, "x, y");
    std::cout << fn(3, 4) << '\n';
    df.execute(3, 4, &dx, &dy);
    std::cout << dx << ' ' << dy << '\n';
}
```
**expected output:** 
9
6 0
**output:**
9
1 0

my assumption as to why this happens is that we don't match the constructor forw and pullback properly. but I'm not sure, it could be that other things are not matched properly as well or that it's another kind of problem.

gojakuch:
this might be related to #1067

Issue #1066, opened 2024-08-29
Clad generates noncompilable code for STL containers (and constructors in general) in the reverse mode


gojakuch:
The examples can be found in actual Clad tests, but here's a reproducer:
```c++
#include "clad/Differentiator/Differentiator.h"
#include "clad/Differentiator/STLBuiltins.h"
#include <iostream>
#include <vector>

#define show(x) std::cout << #x << ": " << x << "\n";

double fn(double u, double v) {
    std::vector<double> vec;
    vec.push_back(u);
    double &ref = vec[0];
    return ref;
}

int main() {
    auto d_fn = clad::gradient(fn);
    double u = 3, v = 5;
    double du, dv;
    du = dv = 0;
    show(fn(u, v));
    d_fn.execute(u, v, &du, &dv);
    show(du);
    show(dv);
}
```

the currently generated code looks like this:
```c++
void fn_grad(double u, double v, double *_d_u, double *_d_v) {
    std::vector<double> _d_vec({});
    std::vector<double> vec;
    double _t0 = u;
    std::vector<double> _t1 = vec;
    clad::custom_derivatives::class_functions::push_back_reverse_forw(&vec, u, &_d_vec, *_d_u);
    std::vector<double> _t2 = vec;
    clad::ValueAndAdjoint<double &, double &> _t3 = clad::custom_derivatives::class_functions::operator_subscript_reverse_forw(&vec, 0, &_d_vec, _r0);
    double &_d_ref = _t3.adjoint;
    double &ref = _t3.value;
    _d_ref += 1;
    {
        std::vector::size_type _r0 = 0;
        clad::custom_derivatives::class_functions::operator_subscript_pullback(&_t2, 0, 0, &_d_vec, &_r0);
    }
    {
        u = _t0;
        clad::custom_derivatives::class_functions::push_back_pullback(&_t1, _t0, &_d_vec, &*_d_u);
    }
}
```
of course, the generated code might change in Clad, but the problem is at the line:
```c++
clad::ValueAndAdjoint<double &, double &> _t3 = clad::custom_derivatives::class_functions::operator_subscript_reverse_forw(&vec, 0, &_d_vec, _r0);
```
where `_r0` is used without being created before that. the code compiles well with Clad used as a plugin, but it's not a valid piece of pastable C++ code.

PetroZarytskyi:
Fixed by #1207 and #1250.

PR #1076, opened 2024-09-03
Fix custom `reverse_forw`s for operators

gojakuch:
Previously, nullptr used to be set as the derivative of a call in the reverse mode, if there was a custom reverse_forw function available. This issue was overlooked at first, since it doesn't cause any trouble, unless someone decides to use nested operators (such as expressions of the form `a[i] = x*x`).

Fixes: #1070

### DIFF
--- a/test/Gradient/STLCustomDerivatives.C
+++ b/test/Gradient/STLCustomDerivatives.C
@@ -103,17 +103,27 @@ double fn13(double u, double v) {
   return vec[0] + vec[1] + vec[2];
 }
 
+double fn14(double x, double y) {
+  std::vector<double> a;
+  a.push_back(x);
+  a.push_back(x);
+  a[1] = x*x;
+  return a[1];
+}
+
 int main() {
     double d_i, d_j;
     INIT_GRADIENT(fn10);
     INIT_GRADIENT(fn11);
     INIT_GRADIENT(fn12);
     INIT_GRADIENT(fn13);
+    INIT_GRADIENT(fn14);
 
     TEST_GRADIENT(fn10, /*numOfDerivativeArgs=*/2, 3, 5, &d_i, &d_j);  // CHECK-EXEC: {1.00, 1.00}
     TEST_GRADIENT(fn11, /*numOfDerivativeArgs=*/2, 3, 5, &d_i, &d_j);  // CHECK-EXEC: {2.00, 1.00}
     TEST_GRADIENT(fn12, /*numOfDerivativeArgs=*/2, 3, 5, &d_i, &d_j);  // CHECK-EXEC: {4.00, 2.00}
     TEST_GRADIENT(fn13, /*numOfDerivativeArgs=*/2, 3, 5, &d_i, &d_j);  // CHECK-EXEC: {3.00, 0.00}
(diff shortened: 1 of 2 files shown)


vgvassilev (approved):
Great catch! LGTM!

test/Gradient/STLCustomDerivatives.C
```diff
@@ -381,3 +391,41 @@ int main() {
...
+// CHECK-NEXT:          {{.*}}push_back_reverse_forw(&a, x, &_d_a, *_d_x);
+// CHECK-NEXT:          double _t2 = x;
+// CHECK-NEXT:          std::vector<double> _t3 = a;
+// CHECK-NEXT:          {{.*}}push_back_reverse_forw(&a, x, &_d_a, *_d_x);
+// CHECK-NEXT:          std::vector<double> _t4 = a;
+// CHECK-NEXT:          clad::ValueAndAdjoint<double &, double &> _t5 = {{.*}}operator_subscript_reverse_forw(&a, 1, &_d_a, _r0);
+// CHECK-NEXT:          double _t6 = _t5.value;
+// CHECK-NEXT:          _t5.value = x * x;
+// CHECK-NEXT:          std::vector<double> _t7 = a;
+// CHECK-NEXT:          clad::ValueAndAdjoint<double &, double &> _t8 = {{.*}}operator_subscript_reverse_forw(&a, 1, &_d_a, _r1);
```
parth-07, line 409: `nullptr` was being passed to `reverse_forw`s intentionally. We cannot pass adjoints uniformally for non-reference/non-pointer cases. Now, the generated code cannot be compiled because `_r1` is being passed to the `reverse_forw` function call even though it is defined later in the reverse-pass.

gojakuch: I know, I've opened an issue about this (#1066) this week and I think it was partially introduced in the original PR about custom `reverse_forw`s for constructors, so I didn't really register this as an issue. it's still better this way because it at least has the correct logic now. but such misuses of yet-to-be-declared variables are happening all over in the reverse mode (at least that's what I thought), so I didn't pay that much attention to this. I'll add this as a note to #1066 or open a separate issue about this. thanks for your comment, cos I'd have forgotten to track this otherwise.

parth-07: > it's still better this way because it at least has the correct logic now.

I don't think it is the correct logic now. The `reverse_forw` call now refers to a variable that is defined much later, that does not seem like a correct logic to me.

> but such misuses of yet-to-be-declared variables are happening all over in the reverse mode (at least that's what I thought)

Please correct me if I am wrong, I think @PetroZarytskyi made significant efforts to reduce such cases. We certainly should not knowingly add more such cases.

> I'll add this as a note to https://github.com/vgvassilev/clad/issues/1066 or open a separate issue about this. thanks for your comment, cos I'd have forgotten to track this otherwise.

I think passing `nullptr` where adjoint is unavailable in the forward-pass is the right way to go. @vgvassilev What do you recommend here?

vgvassilev: I think I see both points. @gojakuch is unblocked by this change on the one hand and on the other it is problematic. Do we have an alternative way to fix his example cases where the `nullptr` approach breaks?

gojakuch: > Do we have an alternative way to fix his example cases where the nullptr approach breaks?
I opened #1079 yesterday to track this. I think there should be a way, I'll get back to this.

PR #1083, opened 2024-09-05
[cuda] Enable filecheck on tests

vgvassilev:
@kchristin22, could you help me do the same for the other two tests. I was not sure what's the expected result after execution.

### DIFF
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -145,9 +145,9 @@ CUDA_HOST_DEVICE T push(tape<T>& to, ArgsT... val) {
     if (CUDAkernel) {
       void* argPtrs[] = {(void*)&args...};
       cudaLaunchKernel((void*)f, grid, block, argPtrs, shared_mem, stream);
-    } else {
-      return f(static_cast<Args>(args)...);
+      return return_type_t<F>();
     }
+    return f(static_cast<Args>(args)...);
 #else
     return f(static_cast<Args>(args)...);
 #endif
--- a/test/CUDA/ForwardMode.cu
+++ b/test/CUDA/ForwardMode.cu
@@ -7,8 +7,6 @@
 // REQUIRES: cuda-runtime
 //
 // expected-no-diagnostics
-//
-// CHECK-NOT: {{.*error|warning|note:.*}}
 
 #include "clad/Differentiator/Differentiator.h"
 #include "../TestUtils.h"
--- a/test/CUDA/ForwardMode.cu
+++ b/test/CUDA/ForwardMode.cu
@@ -1,12 +1,14 @@
-// RUN: %cladclang_cuda -I%S/../../include %s -xc++ %cudasmlevel \
-// RUN: --cuda-path=%cudapath -L/usr/local/cuda/lib64 -lcudart_static \
-// RUN: -ldl -lrt -pthread -lm -lstdc++ -oForwardMode.out 2>&1 | %filecheck %s
-
-// RUN: ./ForwardMode.out
-
+// RUN: %cladclang_cuda -I%S/../../include --cuda-path=%cudapath \
+// RUN:     --cuda-gpu-arch=%cudaarch %cudaldflags -oForwardMode.out \
+// RUN:     -Xclang -verify %s 2>&1 | %filecheck %s
+//
+// RUN: ./ForwardMode.out | %filecheck_exec %s
+//
 // REQUIRES: cuda-runtime
-
+//
 // expected-no-diagnostics
+//
+// CHECK-NOT: {{.*error|warning|note:.*}}
 
 #include "clad/Differentiator/Differentiator.h"
 #include "../TestUtils.h"
@@ -17,7 +19,7 @@ __global__ void add(double *a, double *b, double *c, int n) {
     c[idx] = a[idx] + b[idx];
 }
 
-// CHECK: void add_pushforward(double *a, double *b, double *c, int n, double *_d_a, double *_d_b, double *_d_c, int _d_n) __attribute__((global)) {
+// CHECK:  __attribute__((global)) void add_pushforward(double *a, double *b, double *c, int n, double *_d_a, double *_d_b, double *_d_c, int _d_n) {
 // CHECK-NEXT:     int _d_idx = 0;
 // CHECK-NEXT:     int idx = threadIdx.x;
 // CHECK-NEXT:     if (idx < n) {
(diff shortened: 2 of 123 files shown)


kchristin22:
Yes, will have a look tomorrow

vgvassilev:
> Yes, will have a look tomorrow

Thanks! Looks like things are even broken for the test I just enabled. Could you take a look at that, too?

kchristin22:
Yes I noticed, will fix that : )

kchristin22:
The `ForwardMode.cu` test runs smoothly locally and outputs the correct result, so it seems that a device is not indeed connected to the self-host. The rest two runners are not connected to an actual device either, so this is why they fail, cause I believe that there aren't any other checks with CUDA set up.

test/CUDA/ForwardMode.cu
kchristin22, line 1: A way to verify that a device is connected and check for potential errors is by checking the return value of a  CUDA function f.i. `cudaMalloc`:
```c++
  cudaError_t error = cudaMalloc(&device_a, n * sizeof(double));
  if (error != cudaSuccess) {
    printf("CUDA error: %s\n", cudaGetErrorString(error));
  }
```

kchristin22:
The `GradientCuda.cu` test should check for (line 119):
```c++
  printf("%f,%f,%f\n", result[0], result[1], result[2]); // CHECK-EXEC: 0.007714,0.007714,0.007714
```
The `GradientKernels.cu` test already includes the CHECK-EXEC statements

vgvassilev:
@kchristin22, now the bot fails properly. Can you take a look at the tests and help us fix them?

vgvassilev:
> @kchristin22, now the bot fails properly. Can you take a look at the tests and help us fix them?

The only remaining failure seems to have discovered a real issue:

https://github.com/vgvassilev/clad/blob/50a0b5f797a2ad3c4f35be523c6a166fdc21fd2a/include/clad/Differentiator/Differentiator.h#L145-L150

Here in we take the `true` branch the function does not return anything. Is the intent to still fall back to the `else` part? The function must return a value.

kchristin22:
> Here in we take the true branch the function does not return anything. Is the intent to still fall back to the else part? The function must return a value.

If the function is a global kernel it is of type void, so it doesn't return anything, but it's being executed by `cudaLaunchKernel`

vgvassilev:
> > Here in we take the true branch the function does not return anything. Is the intent to still fall back to the else part? The function must return a value.
> 
> If the function is a global kernel it is of type void, so it doesn't return anything, but it's being executed by `cudaLaunchKernel`

We will probably need to add some sort of template specialization for void because it seems like the compiler diagnostics cannot

> > Here in we take the true branch the function does not return anything. Is the intent to still fall back to the else part? The function must return a value.
> 
> If the function is a global kernel it is of type void, so it doesn't return anything, but it's being executed by `cudaLaunchKernel`

Here is the original diagnostics:

```
In file included from /home/vvassilev/workspace/sources/clad/test/CUDA/ForwardMode.cu:11:
/home/vvassilev/workspace/sources/clad/test/CUDA/../../include/clad/Differentiator/Differentiator.h:154:3: warning: non-void function does not return a value in all control paths [-Wreturn-type]
  154 |   }
      |   ^
/home/vvassilev/workspace/sources/clad/test/CUDA/../../include/clad/Differentiator/Differentiator.h:331:20: note: in instantiation of function template specialization 'clad::execute_with_default_args<false, double (*)(double, double), int, int, double, double, true>' requested here
  331 |             return execute_with_default_args<EnablePadding>(
      |                    ^
/home/vvassilev/workspace/sources/clad/test/CUDA/../../include/clad/Differentiator/Differentiator.h:245:14: note: in instantiation of function template specialization 'clad::CladFunction<double (*)(double, double), clad::NoObject>::execute_helper<double (*)(double, double), int, int>' requested here
  245 |       return execute_helper(m_Function, m_CUDAkernel, dim3(0), dim3(0),
      |              ^
/home/vvassilev/workspace/sources/clad/test/CUDA/../TestUtils.h:170:14: note: in instantiation of function template specialization 'clad::CladFunction<double (*)(double, double), clad::NoObject>::execute<int, int, double (*)(double, double)>' requested here
  170 |   display(cf.execute(std::forward<Args>(args)...));
      |              ^
/home/vvassilev/workspace/sources/clad/test/CUDA/ForwardMode.cu:123:3: note: in instantiation of function template specialization 'test_utils::run_differentiate<clad::CladFunction<double (*)(double, double), clad::NoObject>, int, int>' requested here
  123 |   TEST_DIFFERENTIATE(fn1, 3, 5);  // CHECK-EXEC: 8000.00
      |   ^
/home/vvassilev/workspace/sources/clad/test/CUDA/../TestUtils.h:226:15: note: expanded from macro 'TEST_DIFFERENTIATE'
  226 |   test_utils::run_differentiate(fn##_diff, __VA_ARGS__);
      |               ^
```

Probably there is something to be tweaked for forward mode?

kchristin22:
> We will probably need to add some sort of template specialization for void because it seems like the compiler diagnostics cannot

I see, I think adding this would work:
`return return_type_t<F>();`

include/clad/Differentiator/Differentiator.h
```diff
@@ -145,9 +145,9 @@ CUDA_HOST_DEVICE T push(tape<T>& to, ArgsT... val) {
     if (CUDAkernel) {
       void* argPtrs[] = {(void*)&args...};
       cudaLaunchKernel((void*)f, grid, block, argPtrs, shared_mem, stream);
-    } else {
-      return f(static_cast<Args>(args)...);
+      return return_type_t<F>();
```
kchristin22, line 148: What I meant was:
```
 if (CUDAkernel) {
      void* argPtrs[] = {(void*)&args...};
      cudaLaunchKernel((void*)f, grid, block, argPtrs, shared_mem, stream);
      return return_type_t<F>();
    } else {
    }
   ```
   Cause the compiler doesn't like that the path of `if (CUDAkernel)` is not returning anything.

If the function is not a kernel, it should be executed as it normally would:
```
else {
}
```

Comment (kchristin22) [reply include/clad/Differentiator/Differentiator.h:148]: @vgvassilev  I think this will solve the warnings:
```
#if defined(__CUDACC__) && !defined(__CUDA_ARCH__)
 if (CUDAkernel) {
      void* argPtrs[] = {(void*)&args...};
      cudaLaunchKernel((void*)f, grid, block, argPtrs, shared_mem, stream);
      return return_type_t<F>();
    } else {
    }
 #else ...
 ```

PR #1084, opened 2024-09-06
Add Varied analysis to the reverse mode

ovdiiuv:
This PR adds Varied analysis support in the reverse mode, fixes wrong jacobian results and updates tests.

### DIFF
--- a/lib/Differentiator/ActivityAnalyzer.cpp
+++ b/lib/Differentiator/ActivityAnalyzer.cpp
@@ -29,6 +29,13 @@ void VariedAnalyzer::Analyze(const FunctionDecl* FD) {
   }
 }
 
+void mergeVarsData(VarsData* targetData, VarsData* mergeData) {
+  for (const clang::VarDecl* i : *mergeData)
+    targetData->insert(i);
+  for (const clang::VarDecl* i : *targetData)
+    mergeData->insert(i);
+}
+
 CFGBlock* VariedAnalyzer::getCFGBlockByID(unsigned ID) {
   return *(m_CFG->begin() + ID);
 }
@@ -86,16 +93,18 @@ void VariedAnalyzer::copyVarToCurBlock(const clang::VarDecl* VD) {
 bool VariedAnalyzer::VisitBinaryOperator(BinaryOperator* BinOp) {
   Expr* L = BinOp->getLHS();
   Expr* R = BinOp->getRHS();
-
+  const auto opCode = BinOp->getOpcode();
   if (BinOp->isAssignmentOp()) {
     m_Varied = false;
     TraverseStmt(R);
     m_Marking = m_Varied;
     TraverseStmt(L);
     m_Marking = false;
-  } else {
-    TraverseStmt(L);
-    TraverseStmt(R);
+  } else if (opCode == BO_Add || opCode == BO_Sub || opCode == BO_Mul ||
+             opCode == BO_Div) {
+    for (auto* subexpr : BinOp->children())
+      if (!isa<BinaryOperator>(subexpr))
+        TraverseStmt(subexpr);
   }
   return true;
 }
@@ -111,18 +120,15 @@ bool VariedAnalyzer::VisitConditionalOperator(ConditionalOperator* CO) {
 bool VariedAnalyzer::VisitCallExpr(CallExpr* CE) {
   FunctionDecl* FD = CE->getDirectCallee();
   bool noHiddenParam = (CE->getNumArgs() == FD->getNumParams());
-  std::set<const clang::VarDecl*> variedParam;
   if (noHiddenParam) {
     MutableArrayRef<ParmVarDecl*> FDparam = FD->parameters();
     for (std::size_t i = 0, e = CE->getNumArgs(); i != e; ++i) {
       clang::Expr* par = CE->getArg(i);
       TraverseStmt(par);
-      if (m_Varied || 1) {
-        m_VariedDecls.insert(FDparam[i]);
-        m_Varied = false;
-      }
+      m_VariedDecls.insert(FDparam[i]);
     }
   }
+  m_Varied = true;
   return true;
 }
 
@@ -150,8 +156,6 @@ bool VariedAnalyzer::VisitUnaryOperator(UnaryOperator* UnOp) {
     m_Marking = true;
   }
   TraverseStmt(E);
-  m_Varied = false;
-  m_Marking = false;
   return true;
 }
 
@@ -161,10 +165,15 @@ bool VariedAnalyzer::VisitDeclRefExpr(DeclRefExpr* DRE) {
   if (isVaried(dyn_cast<VarDecl>(DRE->getDecl())))
     m_Varied = true;
 
-  if (const auto* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
-    if (m_Varied && m_Marking)
-      copyVarToCurBlock(VD);
-  }
+  auto* VD = dyn_cast<VarDecl>(DRE->getDecl());
+  if (!VD)
+    return true;
+
+  if (isVaried(VD))
+    m_Varied = true;
+
+  if (m_Varied && m_Marking)
+    copyVarToCurBlock(VD);
   return true;
 }
 } // namespace clad
--- a/lib/Differentiator/ActivityAnalyzer.cpp
+++ b/lib/Differentiator/ActivityAnalyzer.cpp
@@ -29,11 +29,11 @@ void VariedAnalyzer::Analyze(const FunctionDecl* FD) {
   }
 }
 
-void mergeVarsData(VarsData* targetData, VarsData* mergeData) {
+void mergeVarsData(std::set<const clang::VarDecl*>* targetData,
+                   std::set<const clang::VarDecl*>* mergeData) {
   for (const clang::VarDecl* i : *mergeData)
     targetData->insert(i);
-  for (const clang::VarDecl* i : *targetData)
-    mergeData->insert(i);
+  *mergeData = *targetData;
 }
 
 CFGBlock* VariedAnalyzer::getCFGBlockByID(unsigned ID) {
@@ -156,15 +156,11 @@ bool VariedAnalyzer::VisitUnaryOperator(UnaryOperator* UnOp) {
     m_Marking = true;
   }
   TraverseStmt(E);
+  m_Marking = false;
   return true;
 }
 
-bool VariedAnalyzer::VisitInitListExpr(InitListExpr* ILE) { return true; }
-
 bool VariedAnalyzer::VisitDeclRefExpr(DeclRefExpr* DRE) {
-  if (isVaried(dyn_cast<VarDecl>(DRE->getDecl())))
-    m_Varied = true;
-
   auto* VD = dyn_cast<VarDecl>(DRE->getDecl());
   if (!VD)
     return true;
--- a/lib/Differentiator/ActivityAnalyzer.cpp
+++ b/lib/Differentiator/ActivityAnalyzer.cpp
@@ -45,34 +45,31 @@ void VariedAnalyzer::AnalyzeCFGBlock(const CFGBlock& block) {
     }
   }
 
-  for (const auto succ : block.succs()) {
+  for (const clang::CFGBlock::AdjacentBlock succ : block.succs()) {
     if (!succ)
       continue;
     auto& succData = m_BlockData[succ->getBlockID()];
 
-    if (!succData) {
-      succData = new VarsData(*m_BlockData[block.getBlockID()]);
-      succData->m_Prev = m_BlockData[block.getBlockID()];
-    }
+    if (!succData)
+      succData = createNewVarsData(*m_BlockData[block.getBlockID()]);
 
+    bool shouldPushSucc = true;
     if (succ->getBlockID() > block.getBlockID()) {
       if (m_LoopMem == *m_BlockData[block.getBlockID()])
-        m_shouldPushSucc = false;
+        shouldPushSucc = false;
 
-      // has to be changed
-      for (const auto& i : *m_BlockData[block.getBlockID()])
+      for (const VarDecl* i : *m_BlockData[block.getBlockID()])
         m_LoopMem.insert(i);
     }
 
-    if (m_shouldPushSucc)
+    if (shouldPushSucc)
       m_CFGQueue.insert(succ->getBlockID());
-    m_shouldPushSucc = true;
 
     mergeVarsData(succData.get(), m_BlockData[block.getBlockID()].get());
   }
-
-  // has to be changed
-  for (const auto& i : *m_BlockData[block.getBlockID()])
+  // FIXME: Information about the varied variables is stored in the last block,
+  // so we should be able to get it form there
+  for (const VarDecl* i : *m_BlockData[block.getBlockID()])
     m_VariedDecls.insert(i);
 }
 
@@ -82,7 +79,7 @@ bool VariedAnalyzer::isVaried(const VarDecl* VD) const {
 }
 
 void VariedAnalyzer::copyVarToCurBlock(const clang::VarDecl* VD) {
-  auto& curBranch = getCurBlockVarsData();
+  VarsData& curBranch = getCurBlockVarsData();
   curBranch.insert(VD);
 }
 
@@ -103,7 +100,29 @@ bool VariedAnalyzer::VisitBinaryOperator(BinaryOperator* BinOp) {
   return true;
 }
 
-bool VariedAnalyzer::VisitConditionalOperator(clang::ConditionalOperator* CO) {
+// add branching merging
+bool VariedAnalyzer::VisitConditionalOperator(ConditionalOperator* CO) {
+  TraverseStmt(CO->getCond());
+  TraverseStmt(CO->getTrueExpr());
+  TraverseStmt(CO->getFalseExpr());
+  return true;
+}
+
+bool VariedAnalyzer::VisitCallExpr(CallExpr* CE) {
+  FunctionDecl* FD = CE->getDirectCallee();
+  bool noHiddenParam = (CE->getNumArgs() == FD->getNumParams());
+  std::set<const clang::VarDecl*> variedParam;
+  if (noHiddenParam) {
+    MutableArrayRef<ParmVarDecl*> FDparam = FD->parameters();
+    for (std::size_t i = 0, e = CE->getNumArgs(); i != e; ++i) {
+      clang::Expr* par = CE->getArg(i);
+      TraverseStmt(par);
+      if (m_Varied || 1) {
+        m_VariedDecls.insert(FDparam[i]);
+        m_Varied = false;
+      }
+    }
+  }
   return true;
 }
 
@@ -124,10 +143,19 @@ bool VariedAnalyzer::VisitDeclStmt(DeclStmt* DS) {
 }
 
 bool VariedAnalyzer::VisitUnaryOperator(UnaryOperator* UnOp) {
+  const auto opCode = UnOp->getOpcode();
   Expr* E = UnOp->getSubExpr();
+  if (opCode == UO_AddrOf || opCode == UO_Deref) {
+    m_Varied = true;
+    m_Marking = true;
+  }
   TraverseStmt(E);
+  m_Varied = false;
+  m_Marking = false;
   return true;
 }
+
+bool VariedAnalyzer::VisitInitListExpr(InitListExpr* ILE) { return true; }
 
 bool VariedAnalyzer::VisitDeclRefExpr(DeclRefExpr* DRE) {
   if (isVaried(dyn_cast<VarDecl>(DRE->getDecl())))
--- a/lib/Differentiator/ActivityAnalyzer.cpp
+++ b/lib/Differentiator/ActivityAnalyzer.cpp
@@ -2,113 +2,101 @@
 
 using namespace clang;
 
-namespace clad{
+namespace clad {
 
-void VariedAnalyzer::Analyze(const FunctionDecl* FD) {  
-    // Build the CFG (control-flow graph) of FD.
-    clang::CFG::BuildOptions Options;
-    m_CFG = clang::CFG::buildCFG(FD, FD->getBody(), &m_Context, Options);
+void VariedAnalyzer::Analyze(const FunctionDecl* FD) {
+  // Build the CFG (control-flow graph) of FD.
+  clang::CFG::BuildOptions Options;
+  m_CFG = clang::CFG::buildCFG(FD, FD->getBody(), &m_Context, Options);
 
+  m_BlockData.resize(m_CFG->size());
+  // Set current block ID to the ID of entry the block.
+  CFGBlock* entry = &m_CFG->getEntry();
+  m_CurBlockID = entry->getBlockID();
+  m_BlockData[m_CurBlockID] = createNewVarsData({});
+  for (const VarDecl* i : m_VariedDecls)
+    m_BlockData[m_CurBlockID]->insert(i);
+  // Add the entry block to the queue.
+  m_CFGQueue.insert(m_CurBlockID);
 
-    m_BlockData.resize(m_CFG->size());
-    m_BlockPassCounter.resize(m_CFG->size(), 0);
-
-    // Set current block ID to the ID of entry the block.
-    auto* entry = &m_CFG->getEntry();
-    m_CurBlockID = entry->getBlockID();
-    m_BlockData[m_CurBlockID] = new VarsData();
-    for(const auto& i: m_VariedDecls){
-      m_BlockData[m_CurBlockID]->insert(i);
-    }
-    // Add the entry block to the queue.
-    m_CFGQueue.insert(m_CurBlockID);
-
-    // Visit CFG blocks in the queue until it's empty.
-    while (!m_CFGQueue.empty()) {
-        auto IDIter = std::prev(m_CFGQueue.end());
-        m_CurBlockID = *IDIter;
-        m_CFGQueue.erase(IDIter);
-        CFGBlock& nextBlock = *getCFGBlockByID(m_CurBlockID);
-        VisitCFGBlock(nextBlock);
-    }
+  // Visit CFG blocks in the queue until it's empty.
+  while (!m_CFGQueue.empty()) {
+    auto IDIter = std::prev(m_CFGQueue.end());
+    m_CurBlockID = *IDIter;
+    m_CFGQueue.erase(IDIter);
+    CFGBlock& nextBlock = *getCFGBlockByID(m_CurBlockID);
+    AnalyzeCFGBlock(nextBlock);
+  }
 }
 
 CFGBlock* VariedAnalyzer::getCFGBlockByID(unsigned ID) {
   return *(m_CFG->begin() + ID);
 }
 
-void VariedAnalyzer::VisitCFGBlock(const CFGBlock& block) {
+void VariedAnalyzer::AnalyzeCFGBlock(const CFGBlock& block) {
   // Visit all the statements inside the block.
   for (const clang::CFGElement& Element : block) {
-      if (Element.getKind() == clang::CFGElement::Statement) {
+    if (Element.getKind() == clang::CFGElement::Statement) {
       const clang::Stmt* S = Element.castAs<clang::CFGStmt>().getStmt();
+      // The const_cast is inevitable, since there is no
+      // ConstRecusiveASTVisitor.
+      // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
       TraverseStmt(const_cast<clang::Stmt*>(S));
-      }
+    }
   }
 
-  for(const auto succ: block.succs()){
+  for (const auto succ : block.succs()) {
     if (!succ)
       continue;
     auto& succData = m_BlockData[succ->getBlockID()];
 
-    if(!succData){
+    if (!succData) {
       succData = new VarsData(*m_BlockData[block.getBlockID()]);
       succData->m_Prev = m_BlockData[block.getBlockID()];
     }
 
-
-    if(succ->getBlockID() > block.getBlockID()){
-      if(m_LoopMem == *m_BlockData[block.getBlockID()])
+    if (succ->getBlockID() > block.getBlockID()) {
+      if (m_LoopMem == *m_BlockData[block.getBlockID()])
         m_shouldPushSucc = false;
-      
+
       // has to be changed
-      for(const auto& i : *m_BlockData[block.getBlockID()])
+      for (const auto& i : *m_BlockData[block.getBlockID()])
         m_LoopMem.insert(i);
     }
 
-    if(m_shouldPushSucc){
+    if (m_shouldPushSucc)
       m_CFGQueue.insert(succ->getBlockID());
-    }
     m_shouldPushSucc = true;
 
-    merge(succData, m_BlockData[block.getBlockID()]);
+    mergeVarsData(succData.get(), m_BlockData[block.getBlockID()].get());
   }
 
   // has to be changed
-  for(const auto& i: *m_BlockData[block.getBlockID()])
+  for (const auto& i : *m_BlockData[block.getBlockID()])
     m_VariedDecls.insert(i);
 }
 
-bool VariedAnalyzer::isVaried(const VarDecl* VD){
-    auto& curBranch = getCurBlockVarsData();
-    return curBranch.find(VD) != curBranch.end();
-}
-
-void VariedAnalyzer::merge(VarsData* targetData, VarsData* mergeData) {
-  for(const auto& i: *mergeData){
-    targetData->insert(i);
-  }
-  for(const auto& i: *targetData){
-    mergeData->insert(i);
-  }
+bool VariedAnalyzer::isVaried(const VarDecl* VD) const {
+  const VarsData& curBranch = getCurBlockVarsData();
+  return curBranch.find(VD) != curBranch.end();
 }
 
 void VariedAnalyzer::copyVarToCurBlock(const clang::VarDecl* VD) {
-    auto& curBranch = getCurBlockVarsData();
-    curBranch.insert(VD);
+  auto& curBranch = getCurBlockVarsData();
+  curBranch.insert(VD);
 }
 
 bool VariedAnalyzer::VisitBinaryOperator(BinaryOperator* BinOp) {
   Expr* L = BinOp->getLHS();
   Expr* R = BinOp->getRHS();
 
-  if(BinOp->isAssignmentOp()){
+  if (BinOp->isAssignmentOp()) {
     m_Varied = false;
     TraverseStmt(R);
     m_Marking = m_Varied;
     TraverseStmt(L);
     m_Marking = false;
-  }else{
+  } else {
     TraverseStmt(L);
     TraverseStmt(R);
   }
@@ -120,40 +108,35 @@ bool VariedAnalyzer::VisitConditionalOperator(clang::ConditionalOperator* CO) {
 }
 
 bool VariedAnalyzer::VisitDeclStmt(DeclStmt* DS) {
-  for (auto* D : DS->decls()){
-    if (auto* VD = dyn_cast<VarDecl>(D)) {
-      if(Expr* init = VD->getInit()){
-        m_Varied = false;
-        TraverseStmt(init);
-        m_Marking = true;
-        auto& curBranch = getCurBlockVarsData();
-        if(curBranch.find(VD) == curBranch.end() && m_Varied){
-          copyVarToCurBlock(VD);
-        }
-        m_Marking = false;
-      }
+  for (Decl* D : DS->decls()) {
+    if (!isa<VarDecl>(D))
+      continue;
+    if (Expr* init = cast<VarDecl>(D)->getInit()) {
+      m_Varied = false;
+      TraverseStmt(init);
+      m_Marking = true;
+      if (m_Varied)
+        copyVarToCurBlock(cast<VarDecl>(D));
+      m_Marking = false;
     }
   }
   return true;
 }
 
-
-bool VariedAnalyzer::VisitUnaryOperator(UnaryOperator* UnOp){
+bool VariedAnalyzer::VisitUnaryOperator(UnaryOperator* UnOp) {
   Expr* E = UnOp->getSubExpr();
   TraverseStmt(E);
   return true;
 }
 
 bool VariedAnalyzer::VisitDeclRefExpr(DeclRefExpr* DRE) {
-    if(isVaried(dyn_cast<VarDecl>(DRE->getDecl()))){
-        m_Varied = true;
-    }
+  if (isVaried(dyn_cast<VarDecl>(DRE->getDecl())))
+    m_Varied = true;
 
-    if (const auto* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
-        auto& curBranch = getCurBlockVarsData();
-        if (curBranch.find(VD) == curBranch.end() && m_Varied && m_Marking)
-            copyVarToCurBlock(VD);
-    }
-    return true;
+  if (const auto* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
+    if (m_Varied && m_Marking)
+      copyVarToCurBlock(VD);
+  }
+  return true;
 }
-}
+} // namespace clad
--- a/lib/Differentiator/ActivityAnalyzer.cpp
+++ b/lib/Differentiator/ActivityAnalyzer.cpp
@@ -1,0 +1,159 @@
+#include "ActivityAnalyzer.h"
+
+using namespace clang;
+
+namespace clad{
+
+void VariedAnalyzer::Analyze(const FunctionDecl* FD) {  
+    // Build the CFG (control-flow graph) of FD.
+    clang::CFG::BuildOptions Options;
+    m_CFG = clang::CFG::buildCFG(FD, FD->getBody(), &m_Context, Options);
+
+
+    m_BlockData.resize(m_CFG->size());
+    m_BlockPassCounter.resize(m_CFG->size(), 0);
+
+    // Set current block ID to the ID of entry the block.
+    auto* entry = &m_CFG->getEntry();
+    m_CurBlockID = entry->getBlockID();
+    m_BlockData[m_CurBlockID] = new VarsData();
+    for(const auto& i: m_VariedDecls){
+      m_BlockData[m_CurBlockID]->insert(i);
+    }
+    // Add the entry block to the queue.
+    m_CFGQueue.insert(m_CurBlockID);
+
+    // Visit CFG blocks in the queue until it's empty.
+    while (!m_CFGQueue.empty()) {
+        auto IDIter = std::prev(m_CFGQueue.end());
+        m_CurBlockID = *IDIter;
+        m_CFGQueue.erase(IDIter);
+        CFGBlock& nextBlock = *getCFGBlockByID(m_CurBlockID);
+        VisitCFGBlock(nextBlock);
+    }
+}
+
+CFGBlock* VariedAnalyzer::getCFGBlockByID(unsigned ID) {
+  return *(m_CFG->begin() + ID);
+}
+
+void VariedAnalyzer::VisitCFGBlock(const CFGBlock& block) {
+  // Visit all the statements inside the block.
+  for (const clang::CFGElement& Element : block) {
+      if (Element.getKind() == clang::CFGElement::Statement) {
+      const clang::Stmt* S = Element.castAs<clang::CFGStmt>().getStmt();
+      TraverseStmt(const_cast<clang::Stmt*>(S));
+      }
+  }
+
+  for(const auto succ: block.succs()){
+    if (!succ)
+      continue;
+    auto& succData = m_BlockData[succ->getBlockID()];
+
+    if(!succData){
+      succData = new VarsData(*m_BlockData[block.getBlockID()]);
+      succData->m_Prev = m_BlockData[block.getBlockID()];
+    }
+
+
+    if(succ->getBlockID() > block.getBlockID()){
+      if(m_LoopMem == *m_BlockData[block.getBlockID()])
+        m_shouldPushSucc = false;
+      
+      // has to be changed
+      for(const auto& i : *m_BlockData[block.getBlockID()])
+        m_LoopMem.insert(i);
+    }
+
+    if(m_shouldPushSucc){
+      m_CFGQueue.insert(succ->getBlockID());
+    }
+    m_shouldPushSucc = true;
+
+    merge(succData, m_BlockData[block.getBlockID()]);
+  }
+
+  // has to be changed
+  for(const auto& i: *m_BlockData[block.getBlockID()])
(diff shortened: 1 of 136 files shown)


lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,152 @@
...
+void VariedAnalyzer::Analyze(const FunctionDecl* FD) {
+  // Build the CFG (control-flow graph) of FD.
+  clang::CFG::BuildOptions Options;
+  m_CFG = clang::CFG::buildCFG(FD, FD->getBody(), &m_Context, Options);
+
+  m_BlockData.resize(m_CFG->size());
+  m_BlockPassCounter.resize(m_CFG->size(), 0);
+
+  // Set current block ID to the ID of entry the block.
+  auto* entry = &m_CFG->getEntry();
```
vgvassilev, line 16: We should expand the type here.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,152 @@
...
+  m_CurBlockID = entry->getBlockID();
+  m_BlockData[m_CurBlockID] = new VarsData();
+  for (const auto& i : m_VariedDecls)
```
vgvassilev, line 19: We should expand the type here.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,152 @@
...
+void VariedAnalyzer::VisitCFGBlock(const CFGBlock& block) {
+  // Visit all the statements inside the block.
+  for (const clang::CFGElement& Element : block) {
+    if (Element.getKind() == clang::CFGElement::Statement) {
+      const clang::Stmt* S = Element.castAs<clang::CFGStmt>().getStmt();
+      TraverseStmt(const_cast<clang::Stmt*>(S));
+    }
+  }
+
+  for (const auto succ : block.succs()) {
```
vgvassilev, line 47: We should expand the type here.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,152 @@
...
+    if (!succ)
```
vgvassilev, line 48: When `succ` can be null? Would that even enter the loop?

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,165 @@
...
+      }
+  }
+
+  for(const auto succ: block.succs()){
+    if (!succ)
+      continue;
+    auto& succData = m_BlockData[succ->getBlockID()];
+
+    if(!succData){
+      succData = new VarsData(*m_BlockData[block.getBlockID()]);
```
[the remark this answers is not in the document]
vgvassilev: I am a little uncomfortable in allocating `VarsData` in random places in the code. Could this be done in a single place, or maybe behind some helper function?

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,152 @@
...
+      succData->m_Prev = m_BlockData[block.getBlockID()];
+    }
+
+    if (succ->getBlockID() > block.getBlockID()) {
+      if (m_LoopMem == *m_BlockData[block.getBlockID()])
+        m_shouldPushSucc = false;
+
+      // has to be changed
+      for (const auto& i : *m_BlockData[block.getBlockID()])
```
vgvassilev, line 62: We should expand the type here.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+    if (m_shouldPushSucc)
+      m_CFGQueue.insert(succ->getBlockID());
+    m_shouldPushSucc = true;
+
+    merge(succData, m_BlockData[block.getBlockID()]);
+  }
+
+  // has to be changed
+  for (const auto& i : *m_BlockData[block.getBlockID()])
```
vgvassilev, line 74: We should expand the type here.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,152 @@
...
+bool VariedAnalyzer::isVaried(const VarDecl* VD) {
+  auto& curBranch = getCurBlockVarsData();
+  return curBranch.find(VD) != curBranch.end();
+}
+
+void VariedAnalyzer::merge(VarsData* targetData, VarsData* mergeData) {
+  for (const auto& i : *mergeData)
+    targetData->insert(i);
+  for (const auto& i : *targetData)
+    mergeData->insert(i);
```
vgvassilev, line 87: We should expand the type here.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,152 @@
...
+    TraverseStmt(R);
+  }
+  return true;
+}
+
+bool VariedAnalyzer::VisitConditionalOperator(ConditionalOperator* CO) {
+  return true;
+}
+
+bool VariedAnalyzer::VisitCallExpr(CallExpr* CE) { return true; }
```
vgvassilev, line 116: Can't we omit these, as there is no implementation in there and that's the default behavior.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+bool VariedAnalyzer::VisitDeclStmt(DeclStmt* DS) {
+  for (auto* D : DS->decls()) {
```
vgvassilev, line 119: We should expand the type here.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,152 @@
...
+    if (auto* VD = dyn_cast<VarDecl>(D)) {
+      if (Expr* init = VD->getInit()) {
+        m_Varied = false;
+        TraverseStmt(init);
+        m_Marking = true;
+        auto& curBranch = getCurBlockVarsData();
```
vgvassilev, line 125: We should expand the type here.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,152 @@
...
+        if (curBranch.find(VD) == curBranch.end() && m_Varied)
```
vgvassilev, line 126:
```suggestion
        if (m_Varied && curBranch.find(VD) == curBranch.end())
```

Should be faster.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,152 @@
...
+  TraverseStmt(E);
+  return true;
+}
+
+bool VariedAnalyzer::VisitDeclRefExpr(DeclRefExpr* DRE) {
+  if (isVaried(dyn_cast<VarDecl>(DRE->getDecl())))
+    m_Varied = true;
+
+  if (const auto* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
+    auto& curBranch = getCurBlockVarsData();
```
vgvassilev, line 146: We should expand the type here.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,152 @@
...
+    if (curBranch.find(VD) == curBranch.end() && m_Varied && m_Marking)
```
vgvassilev, line 147:
```suggestion
    if (m_Varied && m_Marking && curBranch.find(VD) == curBranch.end())
```

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    if (const auto* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
+        auto& curBranch = getCurBlockVarsData();
+        if (curBranch.find(VD) == curBranch.end() && m_Varied && m_Marking)
+            copyVarToCurBlock(VD);
+    }
+    return true;
+}
+}
```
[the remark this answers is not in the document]
vgvassilev: That's a good suggestion.

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,114 @@
+#include "clang/AST/RecursiveASTVisitor.h"
```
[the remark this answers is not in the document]
vgvassilev: That's a good suggestion too. We just need to put as an include protector `LIB_DIFFERENTIATOR_ACTIVITYANALYZER_H`.

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,120 @@
...
+
+#include "clad/Differentiator/CladUtils.h"
+#include "clad/Differentiator/Compatibility.h"
+
+#include <map>
+#include <unordered_map>
+#include <algorithm>
+#include <set>
+#include <iterator>
+#include <iostream>
```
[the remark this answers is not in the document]
vgvassilev: Good suggestion, too.

vgvassilev (commented):
Here is a partial review. Please address the clang-tidy complaints as most of them are good. I'd also rely on @PetroZarytskyi here for a review.

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,112 @@
...
+        iterator end() { return m_Data.end(); }
+        iterator find(const clang::VarDecl* VD) { return m_Data.find(VD); }
+        void insert(const clang::VarDecl* VD){m_Data.insert(VD);}
+        void clear() { m_Data.clear(); }
+        std::set<const clang::VarDecl*> updateLoopMem(){return m_Data;}
+    };
+
+    VarsData m_LoopMem;
+    clang::CFGBlock* getCFGBlockByID(unsigned ID);
+    // VarData* getExprVarData(const clang::Expr* E, bool addNonConstIdx = false);
```
PetroZarytskyi: Perhaps this should be removed.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+    if(!succData){
+      succData = new VarsData(*m_BlockData[block.getBlockID()]);
+      succData->m_Prev = m_BlockData[block.getBlockID()];
+    }
+
+
+    if(succ->getBlockID() > block.getBlockID()){
+      if(m_LoopMem == *m_BlockData[block.getBlockID()])
+        m_shouldPushSucc = false;
```
PetroZarytskyi, line 59: Does ``m_shouldPushSucc`` have to be a field in the activity analyzer? I believe the variable is only used in this function and it's always ``true`` before this line. Can we declare it, let's say, on line 59?

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,159 @@
...
+      m_CFGQueue.insert(succ->getBlockID());
+    }
+    m_shouldPushSucc = true;
+
+    merge(succData, m_BlockData[block.getBlockID()]);
+  }
+
+  // has to be changed
+  for(const auto& i: *m_BlockData[block.getBlockID()])
+    m_VariedDecls.insert(i);
```
PetroZarytskyi, line 75: I remember we discussed using ``m_BlockData[0]`` (the final CFG block) to get the results and get rid of ``m_VariedDecls``. Is the idea still relevant?

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,112 @@
...
+class VariedAnalyzer : public clang::RecursiveASTVisitor<VariedAnalyzer>{
+
+
+    bool m_Varied = false;
+    bool m_Marking = false;
+    bool m_shouldPushSucc = true;
+
+    std::set<const clang::VarDecl*>& m_VariedDecls;
+
+    struct VarsData {
```
PetroZarytskyi: I believe you took this struct from TBR. I mostly added it as a wrapper of ``std::set<const clang::VarDecl*>`` able to keep track of the previous block, i.e. ``m_Prev``. Your analysis doesn't seem to utilize ``m_Prev``. If so, you could replace VarsData with plain ``std::set<const clang::VarDecl*>``. 
For simplicity, you could do
 ```using VarsData = std::set<const clang::VarDecl*>;```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2939,7 +2940,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
 
     VarDecl* VDClone = nullptr;
     Expr* derivedVDE = nullptr;
-    if (VDDerived)
+    if (VDDerived && m_DiffReq.shouldHaveAdjoint(const_cast<VarDecl*>(VD)))
```
PetroZarytskyi: We might have fixed this during one of hour calls but I'll leave a reminder just in case. You can make the ``shouldHaveAdjoint`` take a const VarDecl*. This way, we can get rid of the const casts.

test/Analyses/ActivityReverse.cpp
```diff
@@ -0,0 +1,102 @@
...
+  auto F##grad = clad::gradient<clad::opts::enable_aa>(F);\
+  F##grad.execute(x, result);\
+  printf("{%.2f}\n", result[0]); \
+}
+
+int main(){
+    double result[3] = {};
+    TEST(f1, 3);// CHECK-EXEC: {6.00}
+    TEST(f2, 3);// CHECK-EXEC: {6.00}
+}
```
PetroZarytskyi, line 175: The tests look great. Perhaps we should add tests for function calls / pullbacks once they are supported.

PetroZarytskyi (changes requested):
Thank you for your work. I didn't mention pullbacks and call exprs because we already made huge progress there yesterday. Overall, the PR looks awesome. Its simplicity makes me feel that TBR was overcomplicated. I think we made the right decision to start with a simplified implementation.

lib/Differentiator/ActivityAnalyzer.cpp
ovdiiuv: I believe that `m_BlockData[0]` isn't always the last one. There has to be a workaround but I'd keep it the way it is for now

lib/Differentiator/ActivityAnalyzer.cpp
ovdiiuv: If some block isn't executed, it isn't included in the CFG but might be among the successors.
For example something like 
```
double f(double x){
  double g = 0;
  if(0){
    g = x;
  }
  return g;
}
```
requires that.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,169 @@
...
+        m_shouldPushSucc = false;
+
+      // has to be changed
+      for (const VarDecl* i : *m_BlockData[block.getBlockID()])
+        m_LoopMem.insert(i);
+    }
+
+    if (m_shouldPushSucc)
+      m_CFGQueue.insert(succ->getBlockID());
+    m_shouldPushSucc = true;
```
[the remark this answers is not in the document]
PetroZarytskyi: This is valid, the statement is useless now.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2890,14 +2891,17 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
             nullptr, VD->getInitStyle());
     }
 
+    if (!m_DiffReq.shouldHaveAdjoint(const_cast<VarDecl*>(VD)))
```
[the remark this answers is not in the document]
PetroZarytskyi: I believe the const cast is unnecessary now.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3034,6 +3044,25 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    //   auto* decl = dyn_cast<VarDecl>(declStmt->getSingleDecl());
+    //   if(decl){
+    //     if(m_DiffReq.shouldHaveAdjoint(const_cast<VarDecl*>(decl))){
+    //       adjInsert = true;
+    //     }
+    //   }
+    // }else{
+    //   adjInsert = true;
+    // }
+
```
PetroZarytskyi, line 3065: This can be removed.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3060,6 +3089,10 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     CompoundStmt* RCS = endBlock(direction::reverse);
     std::reverse(RCS->body_begin(), RCS->body_end());
     Stmt* ReverseResult = utils::unwrapIfSingleStmt(RCS);
+
+    // if(!adjInsert)
+    //   ReverseResult = nullptr;
+
```
PetroZarytskyi, line 3095: And this.

lib/Differentiator/ActivityAnalyzer.cpp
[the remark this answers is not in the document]
ovdiiuv: Can you please elaborate? I don't think I understand what you mean and why should we do so.

lib/Differentiator/ActivityAnalyzer.cpp
vgvassilev: Then we should be able to write a test case for this.

lib/Differentiator/ActivityAnalyzer.cpp
vgvassilev: We should capture this in a fixme note in the codebase.

vgvassilev (commented):
Please check the resolved conversations as some are not really resolved by adding the requested change.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,165 @@
...
+  for(auto& i: *m_BlockData[block.getBlockID()])
+    m_VariedDecls.insert(i);
+}
+
+bool VariedAnalyzer::isVaried(const VarDecl* VD){
+    auto& curBranch = getCurBlockVarsData();
+    return curBranch.find(VD) != curBranch.end();
+}
+
+void VariedAnalyzer::merge(VarsData* targetData, VarsData* mergeData) {
```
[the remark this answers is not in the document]
vgvassilev: This is a good suggestion.

test/Jacobian/FunctionCalls.C
```diff
@@ -16,24 +16,34 @@ void fn1(double i, double j, double* output) {
 }
```
vgvassilev, line 16: Why did we need to modify these tests when AA is off by default?

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+namespace clad {
+
+void VariedAnalyzer::Analyze(const FunctionDecl* FD) {
+  // Build the CFG (control-flow graph) of FD.
+  clang::CFG::BuildOptions Options;
+  m_CFG = clang::CFG::buildCFG(FD, FD->getBody(), &m_Context, Options);
+
+  m_BlockData.resize(m_CFG->size());
+  m_BlockPassCounter.resize(m_CFG->size(), 0);
```
vgvassilev, line 13: I do not see where this member is being used. Maybe drop it?

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,72 @@
...
+#include <algorithm>
+#include <iterator>
+#include <set>
+#include <unordered_map>
+#include <utility>
+
+using namespace clang;
+
+namespace clad {
+class VariedAnalyzer : public clang::RecursiveASTVisitor<VariedAnalyzer> {
```
vgvassilev, line 19: We should write doxygen-style documentation to this class and it's api.

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,72 @@
...
+#include "clad/Differentiator/CladUtils.h"
+#include "clad/Differentiator/Compatibility.h"
+
+#include <algorithm>
+#include <iterator>
+#include <set>
+#include <unordered_map>
+#include <utility>
+
+using namespace clang;
```
vgvassilev, line 16: We must not include `using namespace` in header files. I wonder why clang-tidy does not complain about it.

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,72 @@
...
+  std::vector<std::unique_ptr<VarsData>> m_BlockData;
+  std::vector<short> m_BlockPassCounter;
+  unsigned m_CurBlockID{};
+  std::set<unsigned> m_CFGQueue;
+
+  void addToVaried(const clang::VarDecl* VD);
+  bool isVaried(const clang::VarDecl* VD);
+
+  void copyVarToCurBlock(const clang::VarDecl* VD);
+  VarsData& getCurBlockVarsData() { return *m_BlockData[m_CurBlockID]; }
```
vgvassilev, line 45:
```suggestion
  VarsData& getCurBlockVarsData() { return *m_BlockData[m_CurBlockID]; }
  const VarsData& getCurBlockVarsData() const { return const_cast<VariedAnalyzer*>(this)->getCurBlockVarsData()}
```

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,72 @@
...
+
+  /// Delete copy/move operators and constructors.
+  VariedAnalyzer(const VariedAnalyzer&) = delete;
+  VariedAnalyzer& operator=(const VariedAnalyzer&) = delete;
+  VariedAnalyzer(const VariedAnalyzer&&) = delete;
+  VariedAnalyzer& operator=(const VariedAnalyzer&&) = delete;
+
+  /// Visitors
+  void Analyze(const clang::FunctionDecl* FD);
+  void VisitCFGBlock(const clang::CFGBlock& block);
```
vgvassilev, line 63: We should probably rename this as the other Visit methods are "derived" from the RecursiveASTVisitor.

lib/Differentiator/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ llvm_add_library(cladDifferentiator
   ReverseModeForwPassVisitor.cpp
   ReverseModeVisitor.cpp
   TBRAnalyzer.cpp
+  ActivityAnalyzer.cpp
```
vgvassilev, line 39: We use alphabetical order.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -615,6 +616,32 @@ namespace clad {
...
 
+  bool DiffRequest::shouldHaveAdjoint(const VarDecl* VD) const {
+    if (!EnableActivityAnalysis)
+      return true;
+
+    if (VD->getType()->isPointerType())
+      return true;
+
+    if (!m_ActivityRunInfo.HasAnalysisRun) {
+      if (!DVI.empty()) {
```
vgvassilev, line 627: Inverting the condition would help readability.

test/Jacobian/FunctionCalls.C
ovdiiuv: Those Jacobin tests weren't correct and the changes I've made somehow fixed the problem with jacobians

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,169 @@
...
+}
+
+bool VariedAnalyzer::VisitDeclStmt(DeclStmt* DS) {
+  for (Decl* D : DS->decls()) {
+    if (auto* VD = dyn_cast<VarDecl>(D)) {
+      if (Expr* init = VD->getInit()) {
+        m_Varied = false;
+        TraverseStmt(init);
+        m_Marking = true;
+        VarsData& curBranch = getCurBlockVarsData();
```
vgvassilev, line 142:
```suggestion
       const VarsData& curBranch = getCurBlockVarsData();
```

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,169 @@
...
+  TraverseStmt(E);
+  return true;
+}
+
+bool VariedAnalyzer::VisitDeclRefExpr(DeclRefExpr* DRE) {
+  if (isVaried(dyn_cast<VarDecl>(DRE->getDecl())))
+    m_Varied = true;
+
+  if (const auto* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
+    VarsData& curBranch = getCurBlockVarsData();
```
vgvassilev, line 163:
```suggestion
   const VarsData& curBranch = getCurBlockVarsData();
```

We should extract this into a utility function which does `getCurBlockVarsData` and `find`.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    merge(succData.get(), m_BlockData[block.getBlockID()].get());
+  }
+  // FIXME: Information about the varied variables is stored in the last block,
+  // so we should be able to get it form there
+  for (const VarDecl* i : *m_BlockData[block.getBlockID()])
+    m_VariedDecls.insert(i);
+}
+
+bool VariedAnalyzer::isVaried(const VarDecl* VD) {
```
vgvassilev, line 75:
```suggestion
bool VariedAnalyzer::isVaried(const VarDecl* VD) const {
```

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,77 @@
...
+  bool m_Varied = false;
+  bool m_Marking = false;
+
+  std::set<const clang::VarDecl*>& m_VariedDecls;
+  using VarsData = std::set<const clang::VarDecl*>;
+  /// @brief A helper method to allocate VarsData
+  /// @param toAssign Parameter to initialize new VarsData with.
+  /// @return Unique pointer to a new object of type Varsdata.
+  static std::unique_ptr<VarsData> createNewVarsData(VarsData toAssign) {
+    return std::unique_ptr<VarsData>(new VarsData(std::move(toAssign)));
```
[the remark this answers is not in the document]
ovdiiuv: make_unique doesn't work on C++11, so I think we could keep it as it is

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,165 @@
...
+  m_CurBlockID = entry->getBlockID();
+  m_BlockData[m_CurBlockID] = createNewVarsData({});
+  for (const VarDecl* i : m_VariedDecls)
+    m_BlockData[m_CurBlockID]->insert(i);
+  // Add the entry block to the queue.
+  m_CFGQueue.insert(m_CurBlockID);
+
+  // Visit CFG blocks in the queue until it's empty.
+  while (!m_CFGQueue.empty()) {
+    auto IDIter = std::prev(m_CFGQueue.end());
```
vgvassilev, line 24: We need to specify the type here.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,165 @@
...
+  return *(m_CFG->begin() + ID);
+}
+
+
+void VariedAnalyzer::VisitCFGBlock(const CFGBlock& block) {
+  // Visit all the statements inside the block.
+  for (const clang::CFGElement& Element : block) {
+      if (Element.getKind() == clang::CFGElement::Statement) {
+      const clang::Stmt* S = Element.castAs<clang::CFGStmt>().getStmt();
+      TraverseStmt(const_cast<clang::Stmt*>(S));
```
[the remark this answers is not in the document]
vgvassilev: We should drop the `const` qualifier from the `block` parameter.

lib/Differentiator/ActivityAnalyzer.cpp
vgvassilev: Still no test going here.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,165 @@
...
+    if (auto* VD = dyn_cast<VarDecl>(D)) {
+      if (Expr* init = VD->getInit()) {
+        m_Varied = false;
+        TraverseStmt(init);
+        m_Marking = true;
+        if (m_Varied)
+          copyVarToCurBlock(VD);
+        m_Marking = false;
+      }
+    }
```
vgvassilev, line 144:
```suggestion
    if (!isa<VarDecl>(D))
       return continue;
    if (Expr* init = cast<VarDecl>(D)->getInit()) {
        m_Varied = false;
        TraverseStmt(init);
        m_Marking = true;
        if (m_Varied)
          copyVarToCurBlock(VD);
        m_Marking = false;
    }
```

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,165 @@
...
+  TraverseStmt(E);
+  return true;
+}
+
+bool VariedAnalyzer::VisitDeclRefExpr(DeclRefExpr* DRE) {
+  if (isVaried(dyn_cast<VarDecl>(DRE->getDecl())))
+    m_Varied = true;
+
+  if (const auto* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
+    if (m_Varied && m_Marking)
```
vgvassilev, line 160: I am not generally happy with the state that we collect up and down the visitation process. Is there a way to avoid this somehow?

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,77 @@
...
+
+#include "clad/Differentiator/CladUtils.h"
+#include "clad/Differentiator/Compatibility.h"
+
+#include <algorithm>
+#include <iterator>
+#include <memory>
+#include <set>
+#include <utility>
+/// @brief Class that implemets Varied part of the Activity analysis.
```
vgvassilev, line 15:
```suggestion
/// Class that implemets Varied part of the Activity analysis.
```

No need of brief - it is deprecated and default now.

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,77 @@
...
+class VariedAnalyzer : public clang::RecursiveASTVisitor<VariedAnalyzer> {
+
+  bool m_Varied = false;
+  bool m_Marking = false;
+
+  std::set<const clang::VarDecl*>& m_VariedDecls;
+  using VarsData = std::set<const clang::VarDecl*>;
+  /// @brief A helper method to allocate VarsData
+  /// @param toAssign Parameter to initialize new VarsData with.
+  /// @return Unique pointer to a new object of type Varsdata.
```
vgvassilev, line 30:
```suggestion
  /// A helper method to allocate VarsData
  /// \param toAssign - Parameter to initialize new VarsData with.
  /// \return Unique pointer to a new object of type Varsdata.
```

This is the style we use. Can you update the rest?

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,77 @@
...
+  static std::unique_ptr<VarsData> createNewVarsData(VarsData toAssign) {
+    return std::unique_ptr<VarsData>(new VarsData(std::move(toAssign)));
+  }
+  VarsData m_LoopMem;
+
+  clang::CFGBlock* getCFGBlockByID(unsigned ID);
+  static void merge(VarsData* targetData, VarsData* mergeData);
```
vgvassilev, line 37:
```suggestion
```

The diagnostic was to make something file static, not a class static.

lib/Differentiator/DiffPlanner.cpp
vgvassilev: I meant inverting the condition of `!DVI.empty()`. We should bring back the `HasNoAnalysisRun` change.

test/Jacobian/FunctionCalls.C
vgvassilev: We should move that change in a separate PR.

vgvassilev (commented):
We should add tests for the missing codecov reports.

lib/Differentiator/ActivityAnalyzer.cpp
ovdiiuv: okay, I was frightened by its type tho.

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,84 @@
...
+#include <set>
+#include <utility>
+/// Class that implemets Varied part of the Activity analysis.
+/// By performing static data-flow analysis, so called Varied variables
+/// are determined, meaning variables that depend on input parameters
+/// in a differentiable way. That result enables us to remove redundant
+/// statements in the reverse mode, improving generated codes efficiency.
+namespace clad {
+using VarsData = std::set<const clang::VarDecl*>;
+static inline void mergeVarsData(VarsData* targetData, VarsData* mergeData) {
```
vgvassilev, line 22: This function should go in the cpp file.

vgvassilev:
Can you rebase this PR?

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,142 @@
...
+      copyVarToCurBlock(VD);
+  }
```
vgvassilev, line 139:
```suggestion
  VarDecl *VD = dyn_cast<VarDecl>(DRE->getDecl());
  if (!VD)
    return true;
  if (isVaried(VD))
    m_Varied = true;

  if (m_Varied && m_Marking)
    copyVarToCurBlock(VD);
```

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,84 @@
...
+  clang::CFGBlock* getCFGBlockByID(unsigned ID);
+
+  clang::ASTContext& m_Context;
+  std::unique_ptr<clang::CFG> m_CFG;
+  std::vector<std::unique_ptr<VarsData>> m_BlockData;
+  unsigned m_CurBlockID{};
+  std::set<unsigned> m_CFGQueue;
+  bool isVaried(const clang::VarDecl* VD) const;
+  void copyVarToCurBlock(const clang::VarDecl* VD);
+  VarsData& getCurBlockVarsData() { return *m_BlockData[m_CurBlockID]; }
```
vgvassilev, line 52: We need documentation here, too.

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,84 @@
...
+  ~VariedAnalyzer() = default;
+
+  /// Delete copy/move operators and constructors.
+  VariedAnalyzer(const VariedAnalyzer&) = delete;
+  VariedAnalyzer& operator=(const VariedAnalyzer&) = delete;
+  VariedAnalyzer(const VariedAnalyzer&&) = delete;
+  VariedAnalyzer& operator=(const VariedAnalyzer&&) = delete;
+
+  /// Runs Varied analysis.
+  /// \param FD Function to run the analysis on.
```
vgvassilev, line 74:
```suggestion
  /// \param[in] FD Function to run the analysis on.
```

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,84 @@
...
+class VariedAnalyzer : public clang::RecursiveASTVisitor<VariedAnalyzer> {
+
+  bool m_Varied = false;
+  bool m_Marking = false;
+  bool m_shouldPushSucc = true;
+
+  std::set<const clang::VarDecl*>& m_VariedDecls;
+  // using VarsData = std::set<const clang::VarDecl*>;
+  /// A helper method to allocate VarsData
+  /// \param toAssign - Parameter to initialize new VarsData with.
```
vgvassilev, line 36:
```suggestion
  /// \param toAssign[in] - Parameter to initialize new VarsData with.
```

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,170 @@
...
+        m_Varied = false;
+      }
+    }
+  }
+  return true;
+}
+
+bool VariedAnalyzer::VisitDeclStmt(DeclStmt* DS) {
+  for (Decl* D : DS->decls()) {
+    if (!isa<VarDecl>(D))
```
vgvassilev, line 131: A test here could be `class C{} c;` that produces a DeclStmt whose Stmt is not a VarDecl.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,170 @@
...
+bool VariedAnalyzer::VisitCallExpr(CallExpr* CE) {
+  FunctionDecl* FD = CE->getDirectCallee();
+  bool noHiddenParam = (CE->getNumArgs() == FD->getNumParams());
+  std::set<const clang::VarDecl*> variedParam;
+  if (noHiddenParam) {
+    MutableArrayRef<ParmVarDecl*> FDparam = FD->parameters();
+    for (std::size_t i = 0, e = CE->getNumArgs(); i != e; ++i) {
+      clang::Expr* par = CE->getArg(i);
+      TraverseStmt(par);
+      if (m_Varied || 1) {
```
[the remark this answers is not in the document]
vgvassilev: This probably should change to what it was before.

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,83 @@
...
+#include <memory>
+#include <set>
+#include <utility>
+/// Class that implemets Varied part of the Activity analysis.
+/// By performing static data-flow analysis, so called Varied variables
+/// are determined, meaning variables that depend on input parameters
+/// in a differentiable way. That result enables us to remove redundant
+/// statements in the reverse mode, improving generated codes efficiency.
+namespace clad {
+using VarsData = std::set<const clang::VarDecl*>;
```
vgvassilev, line 21: This can go in the class definition.

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,83 @@
...
+#include <algorithm>
+#include <iterator>
+#include <memory>
+#include <set>
+#include <utility>
+/// Class that implemets Varied part of the Activity analysis.
+/// By performing static data-flow analysis, so called Varied variables
+/// are determined, meaning variables that depend on input parameters
+/// in a differentiable way. That result enables us to remove redundant
+/// statements in the reverse mode, improving generated codes efficiency.
```
vgvassilev, line 19: The documentation should go to the class definition.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,182 @@
...
+  Expr* E = UnOp->getSubExpr();
+  if (opCode == UO_AddrOf || opCode == UO_Deref) {
+    m_Varied = true;
+    m_Marking = true;
+  }
+  TraverseStmt(E);
+  return true;
+}
+
+bool VariedAnalyzer::VisitInitListExpr(InitListExpr* ILE) { return true; }
```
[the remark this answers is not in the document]
vgvassilev: ping.

test/Jacobian/FunctionCalls.C
```diff
@@ -40,6 +40,8 @@ void fn1(double i, double j, double* output) {
 // CHECK-NEXT:     }
 // CHECK-NEXT: }
 
+
+
```
vgvassilev, line 44: Redundant new lines.

test/Gradient/FunctionCalls.C
```diff
@@ -998,7 +998,7 @@ double sq_defined_later(double x) {
 // CHECK-NEXT:         *_d_b += _d_y;
 // CHECK-NEXT: }
 
-// CHECK: void do_nothing_pullback(double *u, double *v, double *w, double _d_y, double *_d_u, double *_d_v, double *_d_w) {
```
vgvassilev, line 1001: Why did we need this change?

test/Gradient/FunctionCalls.C
```diff
@@ -441,14 +441,14 @@ double do_nothing(double* u, double* v, double* w) {
   return u[0];
 }
 
-// CHECK: void do_nothing_pullback(double *u, double *v, double *w, double _d_y, double *_d_u, double *_d_v, double *_d_w);
```
vgvassilev, line 444: Why this change was needed?

ovdiiuv: We used to remove non varied parameters from the adjoint, but it'd crash cms example, so I removed it for now.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,179 @@
...
+  return true;
+}
+
+bool VariedAnalyzer::VisitUnaryOperator(UnaryOperator* UnOp) {
+  const auto opCode = UnOp->getOpcode();
+  Expr* E = UnOp->getSubExpr();
+  if (opCode == UO_AddrOf || opCode == UO_Deref) {
+    m_Varied = true;
+    m_Marking = true;
+  }
```
PetroZarytskyi, line 157: Why do we need to set these modes to ``true`` here? And even if so, shouldn't we also set ``m_Marking`` to ``false`` afterward?

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,179 @@
...
+  TraverseStmt(E);
+  return true;
+}
+
+bool VariedAnalyzer::VisitInitListExpr(InitListExpr* ILE) { return true; }
+
+bool VariedAnalyzer::VisitDeclRefExpr(DeclRefExpr* DRE) {
+  if (isVaried(dyn_cast<VarDecl>(DRE->getDecl())))
+    m_Varied = true;
```
PetroZarytskyi, line 166: Doesn't this portion of the code repeat what comes next? Perhaps we should keep one of these.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,179 @@
...
+    AnalyzeCFGBlock(nextBlock);
+  }
+}
+
+void mergeVarsData(VarsData* targetData, VarsData* mergeData) {
+  for (const clang::VarDecl* i : *mergeData)
+    targetData->insert(i);
+  for (const clang::VarDecl* i : *targetData)
+    mergeData->insert(i);
+}
```
PetroZarytskyi, line 37: It looks like this function will make ``targetData`` and ``mergeData`` identical. Do we need them both to be modified? Even if we do, we can probably do ``*mergedata = *targetdata`` instead of the last two lines.

PetroZarytskyi (changes requested):

lib/Differentiator/ActivityAnalyzer.cpp
ovdiiuv: yes

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,179 @@
...
+  if (BinOp->isAssignmentOp()) {
+    m_Varied = false;
+    TraverseStmt(R);
+    m_Marking = m_Varied;
+    TraverseStmt(L);
+    m_Marking = false;
+  } else if (opCode == BO_Add || opCode == BO_Sub || opCode == BO_Mul ||
+             opCode == BO_Div) {
+    for (auto* subexpr : BinOp->children())
+      if (!isa<BinaryOperator>(subexpr))
```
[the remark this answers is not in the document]
ovdiiuv: See #1106

lib/Differentiator/ActivityAnalyzer.cpp
ovdiiuv: 1. To support something like `double* ptr = &var`. Pointers are set varied by default, but if `var` isn't varied clad would crash when generating an adjoint for `ptr`.
2. Probably yes.

lib/Differentiator/ActivityAnalyzer.h
ovdiiuv: It'd throw an error because merge is file-static

lib/Differentiator/ActivityAnalyzer.cpp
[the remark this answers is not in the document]
vgvassilev: If this does not do anything you can even drop the VisitInitListExpr altogether.

lib/Differentiator/ActivityAnalyzer.h
vgvassilev: In that case you will use it as `VariedAnalyzer::VarsData` or just make it take `std::set<const clang::VarDecl*>`

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,84 @@
...
+#include <utility>
+
+namespace clad {
+using VarsData = std::set<const clang::VarDecl*>;
+class VariedAnalyzer : public clang::RecursiveASTVisitor<VariedAnalyzer> {
+  /// Class that implemets Varied part of the Activity analysis.
+  /// By performing static data-flow analysis, so called Varied variables
+  /// are determined, meaning variables that depend on input parameters
+  /// in a differentiable way. That result enables us to remove redundant
+  /// statements in the reverse mode, improving generated codes efficiency.
```
vgvassilev, line 23: This does not seem to be at the right place either. It needs to go before the class definition.

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -0,0 +1,84 @@
...
+
+  bool m_Varied = false;
+  bool m_Marking = false;
+
+  std::set<const clang::VarDecl*>& m_VariedDecls;
```
vgvassilev, line 28:
```suggestion
  VarsData& m_VariedDecls;
```

PetroZarytskyi (approved):

PetroZarytskyi:
LGTM

test/Gradient/FunctionCalls.C
vgvassilev: I am not sure I follow, Does not that change the pullback signatures randomly? @PetroZarytskyi.

vgvassilev (commented):
We are almost there. Please add tests on the missing branches.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -0,0 +1,172 @@
...
+      m_VariedDecls.insert(FDparam[i]);
+    }
+  }
+  m_Varied = true;
+  return true;
+}
+
+bool VariedAnalyzer::VisitDeclStmt(DeclStmt* DS) {
+  for (Decl* D : DS->decls()) {
+    if (Expr* init = cast<VarDecl>(D)->getInit()) {
```
vgvassilev, line 137: Here we will crash if `D` is not a `VarDecl`.

vgvassilev (approved):
LGTM!

PR #1085, opened 2024-09-06
Cover the specifics of debugging the architecture check runner in the docs

gojakuch:


docs/userDocs/source/user/DevelopersDocumentation.rst
```diff
@@ -300,6 +300,37 @@ Now, to ssh into the GitHub runner do, simply do::
...
+.. code-block:: bash
+
+   bash
+   vim /home/runner/work/_temp/%UNIQUE_FILENAME%
+
+So you need to switch to bash to have access to some text editors and then inside Vim (or any other editor you find there, you can even use ``cat`` for this), you need to edit the contents of this file to only contain the following:
+
+.. code-block:: bash
+
+   export CC=/usr/bin/clang-17
```
vgvassilev, line 322: Isn’t that specific to the current compiler setup and could go out of date easily?

vgvassilev: Sorry. Take it back. I understand it now.

vgvassilev (approved):
Lgtm!

docs/userDocs/source/user/DevelopersDocumentation.rst
gojakuch: oh, I'm not sure about that really. depends on how the architecture check is supposed to be performed (like which compiler should be used and such)

Issue #1087, opened 2024-09-10
Invalid code generated in the gradient when using operators


gojakuch:
reproducer:
```c++
#include "clad/Differentiator/Differentiator.h"
#include <iostream>

namespace TN {
    template <typename T>
    struct Test2 {
        T operator()(T x) {
            return 2*x;
        }
        T operator[](T x) {
            return 4*x;
        }
    };
}

namespace clad {
namespace custom_derivatives {
namespace class_functions {
    template <typename T>
    void operator_subscript_pullback(::TN::Test2<T>* obj, T x, T d_u, ::TN::Test2<T>* d_obj, T* d_x) {
        (*d_x) += 4*d_u;
    }
}}}

double fn2(double x, double y) {
    TN::Test2<double> t;
    auto q = t[x];
    return q;
}

int main() {
  double dx, dy;
  auto df2 = clad::gradient(fn2);
  std::cout << "f2(3, 4)=" << fn2(3.0, 4.0) << '\n';
  dx = 0; dy = 0;
  df2.execute(3.0, 4.0, &dx, &dy);
  std::cout << "df/dx(3, 4)=" << dx << '\n';
  std::cout << "df/dy(3, 4)=" << dy << '\n';
}
```
generates code with the following line:
```c++
double q = operator[](x);
```
which is invalid and should be replaced with `double q = t[x];`

vgvassilev:
`double q = operator[](x);` is missing `double q = t.operator[](x);`. However, I agree it should appear the way you suggested.

Issue #1050, opened 2024-08-21
Clad is not building the nested name qualifiers correctly


gojakuch:
There's a plenty of examples in Kokkos unittests, for instance. Using Clad with the `generate-source-file` option thus produces invalid code with some nested name qualifiers missing.

Here's a reproducer:

```c++
#include "clad/Differentiator/Differentiator.h"
#include <iostream>

namespace TN {
    struct Test {
        static int multiplier;
    };
    int Test::multiplier = 3;
}

double fn(double x) {
    return TN::Test::multiplier*x;
}

int main() {
  auto df = clad::differentiate(fn, "x");
  std::cout << "f(3)=" << fn(3.0);
  std::cout << "; df(3)=" << df.execute(3.0) << '\n';
  return 0;
}
```

In the generated code, the `TN::Test::multiplier` is just used as `multiplier` which isn't correct.

gojakuch:
this was partially solved in #1088. we now support namespace names in the nested name chain, but nothing else, so the reproducer still produces invalid code. an example of what now works can be found in the ValidCodeGen tests in clad.

vgvassilev:
Any clue how to support this?

gojakuch:
well, I'd been trying to solve this issue fully in #1088 but as we saw to little avail, since the debug build kept crushing. I tried a bunch of different approaches and none worked for the debug build. but there should be a way. my guess is that we need to somehow use locations properly, but there's a chance that I overlooked something because of .

vgvassilev:
Can you give me a small standalone reproducer to take a look?

gojakuch:
the reproducer in the issue text is still relevant

vgvassilev:
Fixed in v1.8: https://godbolt.org/z/c3EKGYbfr

PR #1088, opened 2024-09-10
Fix the generation of invalid code in some common cases

gojakuch:
This commit fixes the way Clad generates code. Specifically, it addresses the way operators appear in the generated code in the reverse mode and the way nested name qualifiers are built in both modes. This includes a partial fix to #1050.

Fixes: #1087

### DIFF
--- a/lib/Differentiator/VisitorBase.cpp
+++ b/lib/Differentiator/VisitorBase.cpp
@@ -236,11 +236,38 @@ namespace clad {
   }
 
   DeclRefExpr* VisitorBase::BuildDeclRef(DeclaratorDecl* D,
-                                         const CXXScopeSpec* SS /*=nullptr*/) {
+                                         const CXXScopeSpec* SS /*=nullptr*/,
+                                         ExprValueKind VK /*=VK_LValue*/) {
     QualType T = D->getType();
     T = T.getNonReferenceType();
     return cast<DeclRefExpr>(clad_compat::GetResult<Expr*>(
-        m_Sema.BuildDeclRefExpr(D, T, VK_LValue, D->getBeginLoc(), SS)));
+        m_Sema.BuildDeclRefExpr(D, T, VK, D->getBeginLoc(), SS)));
+  }
+
+  DeclRefExpr* VisitorBase::BuildDeclRef(DeclaratorDecl* D,
+                                         NestedNameSpecifier* NNS,
+                                         ExprValueKind VK /*=VK_LValue*/) {
+    std::vector<NestedNameSpecifier*> NNChain;
+    CXXScopeSpec CSS;
+    while (NNS) {
+      NNChain.push_back(NNS);
+      NNS = NNS->getPrefix();
+    }
+
+    std::reverse(NNChain.begin(), NNChain.end());
+
+    for (size_t i = 0; i < NNChain.size(); ++i) {
+      NNS = NNChain[i];
+      // FIXME: this needs to be extended to support more NNS kinds. An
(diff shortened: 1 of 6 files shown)


lib/Differentiator/VisitorBase.cpp
```diff
@@ -243,6 +243,34 @@ namespace clad {
...
+    std::vector<NestedNameSpecifier*> NNChain;
+    CXXScopeSpec CSS;
+    while (NNS) {
+      NNChain.push_back(NNS);
+      NNS = NNS->getPrefix();
+    }
+
+    std::reverse(NNChain.begin(), NNChain.end());
+
+    for (auto& n : NNChain) {
```
vgvassilev, line 257: I am wondering if we can't use (or draw inspiration from) https://github.com/llvm-mirror/clang/blob/aa231e4be75ac4759c236b755c57876f76e3cf05/lib/AST/QualTypeNames.cpp#L211

gojakuch: I saw it but I didn't dive deep into reading it, since I'd thought it's unrelated, cos it's about type names only.

vgvassilev: The routine is practically the same and more complete to what we are trying to do here. It’s not possible to call it though. Maybe we should at least add a fixme and refer to it as a potential place to steal code from. It has a lot of tests and presumably should work more reliably.

gojakuch: alright, a fixme sounds fair. I'll also try to maybe incorporate some more things from it

gojakuch:
@vgvassilev thanks for the rebase! I'm still working on fixing the last check, seems that it hasn't been affected that much

gojakuch:
@vgvassilev it finally passed all the checks, I'm just going to squash everything now

test/ValidCodeGen/ValidCodeGen.C
```diff
@@ -0,0 +1,80 @@
+// XFAIL: asserts
```
vgvassilev, line 1: Why do we fail in debug builds here?

vgvassilev (commented):
Can you address the clang-tidy comments?

test/ValidCodeGen/ValidCodeGen.C
gojakuch: oh, seems like it was actually doing something, now I remember. unfortunately, the debug build is very picky as it runs some additional internal checks and so it doesn't allow building nested names with class names the way I do. I have a couple of FIXME messages in that place and once they are addressed the XFAIL line can be removed. the problem is that inside Clang they're using locations to build these nested names, while I'm using other things here, since we don't really have valid locations to provide to those Clang methods (providing valid locations from the original source code doesn't seem to work here). my approach works for nested names with namespace and basic class names, but the debug build runs into some asserts if you try to extend a CXXScopeSpec with a class name this way.

I can remove the XFAIL and update the tests so that they don't contain a class name in the nested name specifier chain and then open another separate issue for this case. does that sound good? or we can just not flag the issue #1050 as solved in this PR and update it to say that it's only partially solved.

vgvassilev: Maybe update the tests and remove the xfail flag. Then we can adjust them back when fixing the issue #1050.

gojakuch: @vgvassilev all done. the checks have passed

vgvassilev (approved):
LGTM!

PR #1090, opened 2024-09-11
Move checks in CladFunction to compile-time

MihailMihov:


test/ForwardMode/ConstevalTest.C
```diff
@@ -0,0 +1,58 @@
+// RUN: %cladclang %s -I%S/../../include -std=c++23 -oConstevalTest.out | %filecheck %s
+// RUN: ./ConstevalTest.out | %filecheck_exec %s
+// XFAIL: *
```
vgvassilev, line 3: Why the tests are marked with XFAIL? Can we make a test that makes sure this does not regress again into some runtime evaluation?

MihailMihov: The changes that would allow constexpr/consteval are not in this pr so I disabled them and we can enable them in that pr, or should I remove them from here?

vgvassilev: In that case we should not add the tests as part of this pr.

tools/ClangPlugin.cpp
```diff
@@ -90,6 +90,12 @@ namespace clad {
         CGOpts.PassPlugins.push_back(CladSoPath.str());
       }
 #endif // CLANG_VERSION_MAJOR > 8
+
+      // Add define for CLAD_SO_LOADED, so that CladFunction::CladFunction()
```
vgvassilev, line 94:
```suggestion
      // Add define for __CLAD_SO_LOADED, so that CladFunction::CladFunction()
```

vgvassilev (approved):
LGTM!

PR #1092, opened 2024-09-12
Add support of CUDA builtins

kchristin22:


test/CUDA/GradientKernels.cu
```diff
@@ -73,7 +123,61 @@ int main(void) {
...
+  int *d_in;
+  cudaMalloc(&d_in, 10 * sizeof(int));
+
+  auto add = clad::gradient(add_kernel, "in, out");
+  add.execute_kernel(dim3(1), dim3(10, 1, 1), dummy_out, dummy_in, d_out, d_in);
+  cudaDeviceSynchronize();
+
+  int *res = (int*)malloc(10 * sizeof(int));
+  cudaMemcpy(res, d_in, 10 * sizeof(int), cudaMemcpyDeviceToHost);
+  for(int i = 0; i < 10; i++) {
```
parth-07, line 149: Why are we not using lit/FileCheck for verifying derivatives?

parth-07 (commented):
Looks great!

test/CUDA/GradientKernels.cu
kchristin22: I didn't want to print every value when I ran the test locally. I will change this to a smaller array and print the result at once

test/CUDA/GradientKernels.cu
```diff
@@ -73,7 +123,48 @@ int main(void) {
...
+  }
+  int *d_out;
+  cudaMalloc(&d_out, 5 * sizeof(int));
+  cudaMemcpy(d_out, out, 5 * sizeof(int), cudaMemcpyHostToDevice);
+
+  int *d_in;
+  cudaMalloc(&d_in, 5 * sizeof(int));
+
+  auto add = clad::gradient(add_kernel, "in, out");
+  add.execute_kernel(dim3(1), dim3(5, 1, 1), dummy_out, dummy_in, d_out, d_in);
```
parth-07, line 144: Shouldn't `dummy_out` and `dummy_in` pointers be of size 5 ints? Currently, they are of size 1 int.

test/CUDA/GradientKernels.cu
```diff
@@ -73,7 +123,48 @@ int main(void) {
...
+  int *out = (int*)malloc(5 * sizeof(int));
+  for(int i = 0; i < 5; i++) {
+    out[i] = 5;
+  }
+  int *d_out;
+  cudaMalloc(&d_out, 5 * sizeof(int));
+  cudaMemcpy(d_out, out, 5 * sizeof(int), cudaMemcpyHostToDevice);
+
+  int *d_in;
+  cudaMalloc(&d_in, 5 * sizeof(int));
```
parth-07, line 141: Please initialize `d_in` values to `0` to avoid any undefined behavior.

parth-07 (changes requested):

parth-07 (approved):
Looks good to me.

parth-07 (commented):
Can you please squash all the commits into one?

kchristin22:
> Can you please squash all the commits into one?

This will be done automatically when merging this PR

parth-07:
> This will be done automatically when merging this PR

The primary goal of squashing commits manually in this case is to properly write and structure the commit message.

test/CUDA/GradientKernels.cu
```diff
@@ -34,46 +34,158 @@ void fake_kernel(int *a) {
...
+//CHECK-NEXT:         int _r_d0 = _d_out[index0];
+//CHECK-NEXT:         _d_in[index0] += _r_d0;
+//CHECK-NEXT:     }
+//CHECK-NEXT: }
+
+__global__ void add_kernel_2(int *out, int *in) {
+  out[threadIdx.x] += in[threadIdx.x];
+}
+
+// CHECK:    void add_kernel_2_grad(int *out, int *in, int *_d_out, int *_d_in) {
```
parth-07, line 58: Can you investigate why we don't have `__attribute__((device))` here?

Clang is printing `__attribute__((device))` if we replace `clang::CUDADeviceAttr::CreateImplicit` call in `m_Derivative->addAttr(clang::CUDADeviceAttr::CreateImplicit(m_Context));` with `m_Derivative->addAttr(clang::CUDADeviceAttr::Create(m_Context));`

kchristin22: Yes, I've seen that. The attribute is "hidden" only for the compiler to see when using implicit. I can change that to the explicit creation (`Create` instead of `CreateImplicit`) for clarity purposes if necessary in another PR.

kchristin22: My concern is that it may confuse the user into thinking that it is not a kernel actually being executed (as we don't print the overload)

parth-07: Please create an issue for using `Create` instead of `CreateImplicit`. 

> My concern is that it may confuse the user into thinking that it is not a kernel actually being executed (as we don't print the overload)

I think it is more misleading to not show any attribute at all. Having the attribute present is also necessary for customers wanting to independently use Clad generated-derivatives.

kchristin22: >  Having the attribute present is also necessary for customers wanting to independently use Clad generated-derivatives.

But the attribute is not the correct one if the users want to execute the function themselves, as they would expect to have the multi-threaded execution offered in a global kernel. The attribute should be `__global__` instead of `__device__`. A work-around would be to prepend the string `"__global__ "`  before the printing or dumping to source file if we want this to be intuitive.

parth-07:
The commit message says:

> Added support of CUDA grid configuration builtin variables. Builtins tested: threadIdx, blockIdx, blockDim, gridDim, warpSize

The pull-request does not seem to have tests for `gridDim` and `warpSize`. Can you either remove them from the commit message or add tests for them?

PR #1093, opened 2024-09-14
Store and restore outputaArray elements for jacobians

ovdiiuv:
Before this PR, outputArray's elements weren't stored if they were changed. It would also fix an issue with enabling activity analysis.


PetroZarytskyi (approved):
Currently, this is not going to change the results of jacobians since we don't support the ``output`` variable being used on the RHS of assignments (meaning it's never TBR). However, I see why this change in RMV is necessary for activity analysis. LGTM!

PR #1094, opened 2024-09-15
Add basic support for `std::tie` and tuples in the fwd mode

gojakuch:


gojakuch:
@vgvassilev we should deprecate c++11 as you've suggested

vgvassilev:
> @vgvassilev we should deprecate c++11 as you've suggested

Because of using `index_sequence`?

gojakuch:
> > @vgvassilev we should deprecate c++11 as you've suggested
> 
> Because of using `index_sequence`?

yes and also the `constexpr` stuff. I'm guessing you might suggest using some reimplementation of the `index_sequence`?

vgvassilev:
> > > @vgvassilev we should deprecate c++11 as you've suggested
> > 
> > 
> > Because of using `index_sequence`?
> 
> yes and also the `constexpr` stuff. I'm guessing you might suggest using some reimplementation of the `index_sequence`?

Yeah, maybe in a separate compatibility header file...

test/ForwardMode/STLCustomDerivatives.C
```diff
@@ -1,4 +1,4 @@
-// RUN: %cladclang %s -std=c++14 -I%S/../../include -oSTLCustomDerivatives.out | %filecheck %s
+// RUN: %cladclang -std=c++17 %s -I%S/../../include -oSTLCustomDerivatives.out | %filecheck %s
```
vgvassilev, line 1: Why do we need c++17 here?

vgvassilev (commented):
Can you squash all commits into a commit with an appropriate commit message?

test/ForwardMode/STLCustomDerivatives.C
gojakuch: we don't. I turned it on and forgot to switch it back to c++14.

vgvassilev (approved):
LGTM!

PR #1102, opened 2024-09-21
Fix appendage of nullptrs to args of a CUDA kernel

kchristin22:
Previously, we were appending `nullptr` to the rest of the args, whose addresses are accessed when `cudaLaunchKernel` was used, but `nullptr` cannot be referenced. This PR fixes that by pushing back the address of a param initialized to `nullptr` as many times needed.

This PR will include changes to the tests of #1092 once merged, to show that the problem is solved before it's ready for merging.


parth-07 (approved):
Looks good.

Issue #1071, opened 2024-09-02
Crash when differentiating some `std::vector` code in the reverse mode


gojakuch:
reproducer:

```c++
#include "clad/Differentiator/Differentiator.h"
#include "clad/Differentiator/STLBuiltins.h"
#include <vector>
#include <iostream>

double fn(double x, double y) {
  std::vector<double> a;
  a.push_back(0); // the problem is with pushing literal values for some reason
  a[1] = x*x;
  return a[1];
}

int main(int argc, char* argv[]) {
    double dx=0, dy=0;
    auto df = clad::gradient(fn, "x, y");
    std::cout << fn(3, 4) << '\n';
    df.execute(3, 4, &dx, &dy);
    std::cout << dx << ' ' << dy << '\n';
}
```

PR #1103, opened 2024-09-21
Fix some cases of `std::vector::push_back` in the rvs mode

gojakuch:
Fixes: #1071


gojakuch:
@vgvassilev the clang-tidy-review check fails because of some docker issue

vgvassilev (approved):
Lgtm!

PR #1109, opened 2024-10-03
Add support for differentiation of immediate functions

MihailMihov:


tools/ClangPlugin.cpp
```diff
@@ -137,6 +137,20 @@ namespace clad {
...
+#if CLANG_VERSION_MAJOR > 16
+      for (DiffRequest& request : m_DiffRequestGraph.getNodes()) {
+        if (!request.ImmediateMode ||
+            (!request.Function->isConstexpr() &&
+             !request.Function->isImmediateFunction()))
+          continue;
+
+        m_DiffRequestGraph.setCurrentProcessingNode(request);
+        ProcessDiffRequest(request);
+        m_DiffRequestGraph.markCurrentNodeProcessed();
```
vgvassilev, line 150:
```suggestion
      for (DiffRequest& request : m_DiffRequestGraph.getNodes()) {
        if (request.ImmediateMode && 
            (request.Function->isConstexpr() ||
             request.Function->isImmediateFunction())) {
        m_DiffRequestGraph.setCurrentProcessingNode(request);
        ProcessDiffRequest(request);
        m_DiffRequestGraph.markCurrentNodeProcessed();
        }
```

I am not sure why we need to iterate over all of the requests, though...

include/clad/Differentiator/Differentiator.h
```diff
@@ -192,54 +192,87 @@ CUDA_HOST_DEVICE T push(tape<T>& to, ArgsT... val) {
...
+    }
 
+    constexpr CUDA_HOST_DEVICE CladFunction(CladFunctionType f,
+                                            FunctorType* functor = nullptr,
+                                            bool CUDAkernel = false)
+      requires(ImmediateMode)
+        : m_Function(f), m_Code("<constexpr functions don't have support for "
+                                "printing the derivative yet>"),
+          m_Functor(functor), m_CUDAkernel(CUDAkernel) {
+#ifndef __CLAD_SO_LOADED
```
vgvassilev, line 232: How about renaming this into simply `__CLAD__`?

tools/ClangPlugin.cpp
MihailMihov: In this method `HandleTopLevelDeclForClad` we are given the new `DeclGroupRef` and then the `DiffCollector` adds the new nodes to the differentiation graph, but we don't really have a way to see what is new. I don't think it would be correct to move the processing part inside the collector as that's not it's purpose. Maybe there should be some way for the collector to give references to just the nodes that it added on it's last traversal? Should I do that or do you have something else in mind?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -220,31 +232,35 @@ namespace clad {
...
-
-      StringLiteral* SL = utils::CreateStringLiteral(C, Out.str());
-      Expr* newArg =
-        SemaRef.ImpCastExprToType(SL,
-                                  Arg->getType(),
-                                  CK_ArrayToPointerDecay).get();
-      call->setArg(codeArgIdx, newArg);
+    if (derivedFnArgIdx != -1) {
+      // Add the "&" operator
+      auto newUnOp =
```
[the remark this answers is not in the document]
vgvassilev: That seems a good suggestion.

tools/ClangPlugin.cpp
vgvassilev: I am wondering if we shouldn't check if there is a `FunctionDecl` in `DGR` and ask it if it were an immediate function?

MihailMihov: `DGR` is the top level decl, so to find the just the immediate `FunctionDecl`s we'd need to traverse all of it again, which doesn't sound like a great idea, when that's what `DiffCollector` already does, or am I misunderstanding?

vgvassilev: Ah, yes you are right. Please then just invert the condition as suggested and I think we might be good to go.

MihailMihov: Okay, I'll do that, but also I think that I'll have some documentation ready by tomorrow, so maybe I can include that in this PR.

vgvassilev: Agreed. We need a demo but that can be in a separate PR.

docs/userDocs/source/user/UsingImmediateMode.rst
```diff
@@ -0,0 +1,70 @@
+Using Clad-generated derivatives in an immediate context
+**********************************************************
+
+The derivatives that Clad generates are valid C++ code, which
+could in theory be executed at compile-time (or in an immediate
+context as the C++ standard calls it). When a function is differentiated
+all attributes, such as `constexpr` and `consteval` are kept, but
```
vgvassilev, line 7:
```suggestion
all specifiers, such as `constexpr` and `consteval` are kept, but
```

docs/userDocs/source/user/UsingImmediateMode.rst
```diff
@@ -0,0 +1,70 @@
...
+
+Currently to get a `CladFunction` that is usable in immediate
+mode the user has to pass `clad::immediate_mode` to the differentiation
+function and that removes the ability to dump the generated derivative,
+but it may be possible to add support for that in the future.
+
+Usage of Clad's immediate mode
+================================================
+
+The following code snippet shows how one can request Clad to use the immediate mode for
```
vgvassilev, line 27: We tend to break at 80 columns.

docs/userDocs/source/user/UsingImmediateMode.rst
```diff
@@ -0,0 +1,70 @@
...
+
+Use cases supported by Clad's immediate mode
+================================================
+
+Currently Clad's immediate mode is primarily meant to be used in the forward mode
+(`clad::differentiate`) as internal data structures that Clad needs for differentiating
+loops, etc. are not yet usable in an immediate context.
+
+Both `constexpr` and `consteval` are supported as Clad doesn't actually rely on these
+specific keywords for it's support, but instead uses clang's API to determine if the functions
```
vgvassilev, line 69:
```suggestion
specific keywords for its support, but instead uses clang's API to determine if the functions
```

tools/ClangPlugin.cpp
```diff
@@ -137,6 +137,19 @@ namespace clad {
       SetRequestOptions(opts);
       DiffCollector collector(DGR, CladEnabledRange, m_DiffRequestGraph, S,
                               opts);
+
+#if CLANG_VERSION_MAJOR > 16
+      for (DiffRequest& request : m_DiffRequestGraph.getNodes()) {
+        if (request.ImmediateMode &&
+            (request.Function->isConstexpr() ||
+             request.Function->isImmediateFunction())) {
```
vgvassilev, line 145: Any chance to write a test that hits here?

MihailMihov: It seems like `isImmediateFunction` isn't being hit, because `isConstexpr` is checking the `ConstexprSpecKind`, which contains information for constexpr, consteval and constinit and even if I remove the `isImmediateFunction` check the tests still pass, but I believe that is only the case, because with the current implementation we are guaranteed that clang would've set the `ConstexprSpecKind` before we process the `FunctionDecl`. So would it be better to keep that assumption and remove the second check or keep it, even though it's not tested currently?

vgvassilev (approved):
Lgtm!

PR #1110, opened 2024-10-03
Remove excessive parameters from Derive functions

PetroZarytskyi:
``Derive`` functions have access to the original function and the diff request through ``m_DiffRef``, there's no need to pass those as parameters. The same applies to ``m_Functor`` as it's also stored in ``m_DiffRef``.

Partially addresses #721.


vgvassilev:
Can you fix the clang-tidy and code-cov complaints? Also please mention in the commit message that this partially addresses #721.

vgvassilev (approved):
LGTM!

PR #1111, opened 2024-10-07
Add support of CUDA device pullbacks

kchristin22:
Add support for device pullback functions.

1. If the function is a global kernel, we know that all its args reside in global memory. These are stored in a container (`m_GlobalArgs`)
2. If the function is a device function, it was called inside a kernel or it's a nested device call (the first one was called by a kernel).
 2.1. If the user has called `clad::gradient` on the kernel, the `m_GlobalArgs` set is not empty. When a call is encountered, we iterate on the args of the call and identify which reside in global memory. If such args are found, we store their index in another container (`m_GlobalArgsIndexes`) and include it in the differentiation request of the device function. When the latter is derived later on using `DerivePullback()`, we check whether `m_GlobalArgsIndexes` is empty, or in another words if any of its call args are global, and if so, match them with the parameters of this function's signature (add them in the local `m_GlobalArgs` container). The same procedure is followed for nested calls. 
 2.2. If the user didn't call `clad::gradient` on the kernel, but only to the device function, then we can't tell at compile time which runtime args are global. As a result the user should copy the data to local memory first and back again
   ```cpp
   __global__ kernel(double *in, double *out, double *_d_in, double *_d_out){
   auto test = clad::gradient(device); 
   test.execute(in, out, _d_in, _d_out); // use local vars here and then assign the result back to the global args
   } 
   ```

**NOTE**: Currently `Clad` doesn't support variables in shared memory, which would be considered global as well.

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -38,6 +38,11 @@ namespace clad {
     // several private/protected members of the visitor classes.
     friend class ErrorEstimationHandler;
     llvm::SmallVector<const clang::ValueDecl*, 16> m_IndependentVars;
+    /// Set used to keep track of parameter variables w.r.t which the
+    /// the derivative (gradient) is being computed. This is separate from the
+    /// m_Variables map because all other intermediate variables will
+    /// not be stored here.
+    std::unordered_set<const clang::ValueDecl*> m_ParamVarsWithDiff;
     /// In addition to a sequence of forward-accumulated Stmts (m_Blocks), in
     /// the reverse mode we also accumulate Stmts for the reverse pass which
     /// will be executed on return.
@@ -51,6 +56,8 @@ namespace clad {
     /// that will be put immediately in the beginning of derivative function
     /// block.
     Stmts m_Globals;
+    /// Global GPU args of the function.
+    std::unordered_set<const clang::ParmVarDecl*> m_GlobalArgs;
     //// A reference to the output parameter of the gradient function.
     clang::Expr* m_Result;
     /// A flag indicating if the Stmt we are currently visiting is inside loop.
@@ -432,7 +439,7 @@ namespace clad {
 
     /// Helper function that checks whether the function to be derived
     /// is meant to be executed only by the GPU
-    bool shouldUseCudaAtomicOps();
+    bool shouldUseCudaAtomicOps(const clang::Expr* E);
 
     /// Add call to cuda::atomicAdd for the given LHS and RHS expressions.
     ///
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -57,7 +57,7 @@ namespace clad {
     /// block.
     Stmts m_Globals;
     /// Global GPU args of the function.
-    std::unordered_set<const clang::ParmVarDecl*> m_GlobalArgs;
+    std::unordered_set<const clang::ParmVarDecl*> m_CUDAGlobalArgs;
     //// A reference to the output parameter of the gradient function.
     clang::Expr* m_Result;
     /// A flag indicating if the Stmt we are currently visiting is inside loop.
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -106,12 +106,12 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
 
   bool ReverseModeVisitor::shouldUseCudaAtomicOps(const Expr* E) {
     // Same as checking whether this is a function executed by the GPU
-    if (!m_GlobalArgs.empty())
+    if (!m_CUDAGlobalArgs.empty())
       if (const auto* DRE = dyn_cast<DeclRefExpr>(E))
         if (const auto* PVD = dyn_cast<ParmVarDecl>(DRE->getDecl()))
           // we need to check whether this param is in the global memory of the
           // GPU
-          return m_GlobalArgs.find(PVD) != m_GlobalArgs.end();
+          return m_CUDAGlobalArgs.find(PVD) != m_CUDAGlobalArgs.end();
 
     return false;
   }
@@ -455,7 +455,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     // global memory of the GPU
     if (m_DiffReq->hasAttr<clang::CUDAGlobalAttr>())
       for (auto param : params)
-        m_GlobalArgs.emplace(param);
+        m_CUDAGlobalArgs.emplace(param);
 
     llvm::ArrayRef<ParmVarDecl*> paramsRef =
         clad_compat::makeArrayRef(params.data(), params.size());
@@ -563,7 +563,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
 
     auto derivativeName =
         utils::ComputeEffectiveFnName(m_DiffReq.Function) + "_pullback";
-    for (auto index : m_DiffReq.GlobalArgsIndexes)
+    for (auto index : m_DiffReq.CUDAGlobalArgsIndexes)
       derivativeName += "_" + std::to_string(index);
     auto DNI = utils::BuildDeclarationNameInfo(m_Sema, derivativeName);
 
@@ -608,14 +608,14 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     m_Derivative->setParams(params);
     // Match the global arguments of the call to the device function to the
     // pullback function's parameters.
-    if (!m_DiffReq.GlobalArgsIndexes.empty())
-      for (auto index : m_DiffReq.GlobalArgsIndexes)
-        m_GlobalArgs.emplace(m_Derivative->getParamDecl(index));
+    if (!m_DiffReq.CUDAGlobalArgsIndexes.empty())
+      for (auto index : m_DiffReq.CUDAGlobalArgsIndexes)
+        m_CUDAGlobalArgs.emplace(m_Derivative->getParamDecl(index));
     // If the function is a global kernel, all its parameters reside in the
     // global memory of the GPU
     else if (m_DiffReq->hasAttr<clang::CUDAGlobalAttr>())
       for (auto param : params)
-        m_GlobalArgs.emplace(param);
+        m_CUDAGlobalArgs.emplace(param);
     m_Derivative->setBody(nullptr);
 
     if (!m_DiffReq.DeclarationOnly) {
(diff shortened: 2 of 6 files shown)


vgvassilev:
In a separate PR we should probably move the cuda specific builtins into a separate header.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1579,8 +1579,20 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
           // FIXME: not sure if this is generic.
           // Don't update derivatives of record types.
           if (!VD->getType()->isRecordType()) {
-            auto* add_assign = BuildOp(BO_AddAssign, it->second, dfdx());
-            // Add it to the body statements.
+            auto indepedentVar = m_ParamVariables.find(VD);
+            Expr* add_assign = nullptr;
+            if (indepedentVar != std::end(m_ParamVariables) &&
+                shouldUseCudaAtomicOps()) {
+              // if the derived param is not an array subscript and thus stored
```
parth-07, line 1584: Why don't we need to get the pointer/array for array subscript expressions? Both `*_d_param` and `_d_param[idx]` have non-pointer types, right?

parth-07 (changes requested):
We should have separate PRs for enabling kernel calls in a host function and device function call inside a kernel.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -38,6 +38,11 @@ namespace clad {
     // several private/protected members of the visitor classes.
     friend class ErrorEstimationHandler;
     llvm::SmallVector<const clang::ValueDecl*, 16> m_IndependentVars;
+    /// Map used to keep track of parameter variables w.r.t which the
+    /// the derivative (gradient) is being computed. This is separate from the
+    /// m_Variables map because all other intermediate variables will
+    /// not be stored here.
+    std::unordered_map<const clang::ValueDecl*, clang::Expr*> m_ParamVariables;
```
parth-07, line 45: It is only used to determine if a `ValueDecl*` is a function parameter, can we make it a `std::unordered_set` instead of map?

parth-07:
Can you please rebase the pull-request on top of `origin/master`?

vgvassilev:
> Can you please rebase the pull-request on top of `origin/master`?

Done. @kchristin22 we need a better commit message.

test/CUDA/GradientKernels.cu
```diff
@@ -288,6 +288,186 @@ __global__ void add_kernel_7(double *a, double *b) {
...
+//CHECK-NEXT:        _d_out[index0] = 0.;
+//CHECK-NEXT:        double _r0 = 0.;
+//CHECK-NEXT:        device_with_device_call_pullback_0_1_3(in, val, _r_d0, _d_in, &_r0);
+//CHECK-NEXT:        _d_val += _r0;
+//CHECK-NEXT:    }
+//CHECK-NEXT:}
+
+// CHECK: __attribute__((device)) void device_fn_pullback_1(double in, double val, double _d_y, double *_d_in, double *_d_val) {
+//CHECK-NEXT:    {
+//CHECK-NEXT:                *_d_in += _d_y;
```
parth-07, line 417: Why are we not using `atomicAdd` here for `_d_in` and `_d_val`?

kchristin22: Because local vars are passed to this function for `_d_in` and `_d_val` when it's being called in `kernel_with_device_call_grad_0_2`:
```
double _r0 = 0.;
double _r1 = 0.;
device_fn_pullback_1(in[index0], val, _r_d0, &_r0, &_r1);
atomicAdd(_d_val, _r1);
```

kchristin22:
> Done. @kchristin22 we need a better commit message.

On it!

test/CUDA/GradientKernels.cu
parth-07: Oh, okay, You are right.

test/CUDA/GradientKernels.cu
```diff
@@ -288,6 +288,186 @@ __global__ void add_kernel_7(double *a, double *b) {
...
+//CHECK-NEXT:                *_d_val += _d_y;
+//CHECK-NEXT:    }
+//CHECK-NEXT:}
+
+// CHECK: __attribute__((device)) void device_fn_2_pullback_0_1(double *in, double val, double _d_y, double *_d_val) {
```
parth-07, line 422: Why don't we have `_d_in` parameter in this pullback function?

kchristin22: Because `in` was not part of the `clad::gradient` arg list

parth-07: That should not matter. We still need to compute derivatives with respect to `in` to correctly compute derivatives w.r.t `out` and `val`.

kchristin22: How could it affect the result though since we treat `in` as a value? Why should we be interested in its adjoint?

parth-07: In the current example, this would not matter whether we compute adjoints with respect to `in` or not. But in a general case, it will matter.

Consider a simple example:

```cpp
  v = u;
  double res = v;
  return res;
}

auto fn_grad = clad::gradient(fn, "u");
```
Here, `v` is not a part of `clad::gradient` arg list. However, we still need to compute its derivative to correctly compute the derivative of `u`. The `clad::gradient` arg list contains the arguments in whose derivative the user is interested in. It does not mean that the remaining arguments are constants/non-differentiable.

kchristin22: Yes, but `clad` would already define `_d_v` internally as a local var, wouldn't it?

parth-07: It ideally should, but it seems that Clad does not declare local var `_d_v` if `v` is a pointer. That's understandable because we do not know the size of the array/pointer, however, this does make Clad behavior inconsistent between a pointer/array arg and a non-pointer arg. Let's leave it for now.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -51,6 +56,8 @@ namespace clad {
     /// that will be put immediately in the beginning of derivative function
     /// block.
     Stmts m_Globals;
+    /// Global GPU args of the function.
+    std::unordered_set<const clang::ParmVarDecl*> m_GlobalArgs;
```
parth-07, line 60: Can we please rename `m_GlobalArgs` to `m_CUDAGlobalArgs` for more clarity and intent?

parth-07 (approved):
Looks good, except for one nitpick comment.

kchristin22:
Done!

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -104,10 +104,16 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
-    return m_DiffReq->hasAttr<clang::CUDAGlobalAttr>() ||
-           (m_DiffReq->hasAttr<clang::CUDADeviceAttr>() &&
-            !m_DiffReq->hasAttr<clang::CUDAHostAttr>());
+  bool ReverseModeVisitor::shouldUseCudaAtomicOps(const Expr* E) {
+    // Same as checking whether this is a function executed by the GPU
+    if (!m_CUDAGlobalArgs.empty())
+      if (const auto* DRE = dyn_cast<DeclRefExpr>(E))
+        if (const auto* PVD = dyn_cast<ParmVarDecl>(DRE->getDecl()))
+          // we need to check whether this param is in the global memory of the
+          // GPU
```
vgvassilev, line 113:
```suggestion
          // Check whether this param is in the global memory of the GPU
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -587,6 +601,12 @@
       m_ExternalSource->ActAfterCreatingDerivedFnParams(params);
 
     m_Derivative->setParams(params);
+    if (!m_DiffReq.GlobalArgsIndexes.empty())
+      for (auto index : m_DiffReq.GlobalArgsIndexes)
+        m_GlobalArgs.emplace(m_Derivative->getParamDecl(index));
+    else if (m_DiffReq->hasAttr<clang::CUDAGlobalAttr>())
+      for (auto param : params)
```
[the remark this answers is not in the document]
vgvassilev: That's a good suggestion.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -587,6 +606,16 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    // Match the global arguments of the call to the device function to the
+    // pullback function's parameters.
+    if (!m_DiffReq.CUDAGlobalArgsIndexes.empty())
+      for (auto index : m_DiffReq.CUDAGlobalArgsIndexes)
+        m_CUDAGlobalArgs.emplace(m_Derivative->getParamDecl(index));
+    // If the function is a global kernel, all its parameters reside in the
+    // global memory of the GPU
+    else if (m_DiffReq->hasAttr<clang::CUDAGlobalAttr>())
+      for (auto param : params)
+        m_CUDAGlobalArgs.emplace(param);
```
vgvassilev, line 618: Codecov complains about missing coverage here: can we add a test that goes here?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2237,12 +2302,17 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
           ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, i);
       Expr* gradElem = BuildArraySubscript(gradRef, {idx});
       Expr* gradExpr = BuildOp(BO_Mul, dfdx, gradElem);
-      PostCallStmts.push_back(BuildOp(BO_AddAssign, outputArgs[i], gradExpr));
+      if (shouldUseCudaAtomicOps(outputArgs[i]))
+        PostCallStmts.push_back(
+            BuildCallToCudaAtomicAdd(outputArgs[i], gradExpr));
```
vgvassilev, line 2307: Can we add a test case here?

lib/Differentiator/ReverseModeVisitor.cpp
kchristin22: Yes, good observation. This should actually be included in the other PR regarding kernel pullbacks. There's a test there. I'm removing it for now.

lib/Differentiator/ReverseModeVisitor.cpp
kchristin22: After taking a closer look into this, this condition would always be false, since this function is only called when the callback function doesn't include any pointers as args. When this is the case, local vars (f.i. `_r0`) are used as outputs call args (`&_r0` -> `double *_d_a` (output callback function parameter))

vgvassilev (approved):
LGTM!

PR #1113, opened 2024-10-08
Fix synth literal function for enums

kchristin22:
Tested in PR  #1111

### DIFF
--- a/lib/Differentiator/ConstantFolder.cpp
+++ b/lib/Differentiator/ConstantFolder.cpp
@@ -7,6 +7,7 @@
 //----------------------------------------------------------------------------//
 
 #include "ConstantFolder.h"
+#include "clad/Differentiator/Compatibility.h"
 
 #include "clang/AST/ASTContext.h"
 
@@ -141,7 +142,17 @@ namespace clad {
     // SourceLocation noLoc;
     Expr* Result = 0;
     QT = QT.getCanonicalType();
-    if (QT->isPointerType()) {
+    if (QT->isEnumeralType()) {
+      llvm::APInt APVal(C.getIntWidth(QT), val,
+                        QT->isSignedIntegerOrEnumerationType());
+      Result = clad::synthesizeLiteral(
+          dyn_cast<EnumType>(QT)->getDecl()->getIntegerType(), C, APVal);
+      Expr* cast = ImplicitCastExpr::Create(
+          C, QT, clang::CastKind::CK_IntegralCast, Result, nullptr,
+          CLAD_COMPAT_ExprValueKind_R_or_PR_Value
+              CLAD_COMPAT_CLANG12_CastExpr_DefaultFPO);
+      Result = cast;
+    } else if (QT->isPointerType()) {
       Result = clad::synthesizeLiteral(QT, C);
     } else if (QT->isBooleanType()) {
       Result = clad::synthesizeLiteral(QT, C, (bool)val);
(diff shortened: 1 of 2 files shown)


vgvassilev (commented):
Can we add a test for it?

lib/Differentiator/ConstantFolder.cpp
```diff
@@ -141,7 +142,17 @@ namespace clad {
     // SourceLocation noLoc;
     Expr* Result = 0;
     QT = QT.getCanonicalType();
-    if (QT->isPointerType()) {
+    if (QT->isEnumeralType()) {
+      llvm::APInt APVal(C.getIntWidth(QT), val,
+                        QT->isSignedIntegerOrEnumerationType());
+      Result = clad::synthesizeLiteral(
+          dyn_cast<EnumType>(QT)->getDecl()->getIntegerType(), C, APVal);
+      Expr* cast = ImplicitCastExpr::Create(
```
parth-07, line 150: Do we need to explicitly add the implicit cast? If so, can you please check if the generated code is valid -- can be compiled independently of Clad?

kchristin22: You're right, I didn't think about that. I changed it to `static_cast`.

lib/Differentiator/ConstantFolder.cpp
```diff
@@ -141,7 +142,20 @@ namespace clad {
...
-    if (QT->isPointerType()) {
+    if (QT->isEnumeralType()) {
+      llvm::APInt APVal(C.getIntWidth(QT), val,
+                        QT->isSignedIntegerOrEnumerationType());
+      Result = clad::synthesizeLiteral(
+          dyn_cast<EnumType>(QT)->getDecl()->getIntegerType(), C, APVal);
+      SourceLocation noLoc;
+      Expr* cast = CXXStaticCastExpr::Create(
+          C, QT, CLAD_COMPAT_ExprValueKind_R_or_PR_Value,
+          clang::CastKind::CK_IntegralCast, Result, nullptr,
```
vgvassilev, line 153:
```suggestion
          clang::CastKind::CK_IntegralCast, Result, /*CXXCastPath=*/nullptr,
```

vgvassilev (approved):
LGTM!

lib/Differentiator/ConstantFolder.cpp
```diff
@@ -141,7 +142,20 @@ namespace clad {
...
     Expr* Result = 0;
     QT = QT.getCanonicalType();
-    if (QT->isPointerType()) {
+    if (QT->isEnumeralType()) {
+      llvm::APInt APVal(C.getIntWidth(QT), val,
+                        QT->isSignedIntegerOrEnumerationType());
+      Result = clad::synthesizeLiteral(
+          dyn_cast<EnumType>(QT)->getDecl()->getIntegerType(), C, APVal);
+      SourceLocation noLoc;
+      Expr* cast = CXXStaticCastExpr::Create(
```
parth-07, line 151: Can we please use `clang::Sema::BuildCXXNamedCast` here? `BuildCXXNamedCast` will automatically determine the correct values for `ExprValueKind`, `CXXCastPath`, and `FPOptionsOverride`.

kchristin22: This regards the whole function and how it's implemented. For each case we create the literal manually, but we can alter that to include Sema as an arg instead of the AST context and use Sema's Build functions for all type cases. @vgvassilev what's your opinion on this?

vgvassilev: Yeah, that'd be annoying to update all call sites. Probably worth doing but maybe after the other PR is complete and merged?

kchristin22: Sounds good

PR #1114, opened 2024-10-12
Add support of cuda kernels as pullback functions

kchristin22:


include/clad/Differentiator/ExternalRMVSource.h
```diff
@@ -134,6 +134,16 @@ class ExternalRMVSource {
       llvm::SmallVectorImpl<clang::Expr*>& derivedCallArgs,
       llvm::SmallVectorImpl<clang::Expr*>& ArgResult, bool asGrad) {}
 
+  /// This is called just before finalising `VisitCUDAKernelCallExpr`.
+  ///
+  /// \param KCE kernel call expression that is being visited.
+  /// \param CallArgs
```
parth-07, line 140: `CallArgs` and `ArgResultDecls` mentioned in comments are not present in the function.

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -118,7 +118,8 @@ namespace clad {
     clang::Expr* BuildCallToCustomDerivativeOrNumericalDiff(
         const std::string& Name, llvm::SmallVectorImpl<clang::Expr*>& CallArgs,
         clang::Scope* S, clang::DeclContext* originalFnDC,
-        bool forCustomDerv = true, bool namespaceShouldExist = true);
+        bool forCustomDerv = true, bool namespaceShouldExist = true,
+        clang::Expr* config = nullptr);
```
parth-07, line 122: Can we rename `config` to `execConfig` or `CUDAExecConfig` for more clarity?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -610,7 +610,11 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     if (!m_DiffReq.CUDAGlobalArgsIndexes.empty())
       for (auto index : m_DiffReq.CUDAGlobalArgsIndexes)
         m_CUDAGlobalArgs.emplace(m_Derivative->getParamDecl(index));
-
+    // If the function is a global kernel, all its parameters reside in the
```
parth-07, line 613: This comment is not strictly true. For example, for the kernel `__global__ void SomeKernel(float *a, int val)`, `val` is passed by value and will not reside in the global memory of the GPU. However, its derivative will reside in the global memory of the GPU so the behavior is correct. Can you please update this comment?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2268,12 +2272,403 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+    if (!FD) {
+      diag(DiagnosticsEngine::Warning, KCE->getEndLoc(),
+           "Differentiation of only direct calls is supported. Ignored");
+      return StmtDiff(Clone(KCE));
+    }
+
+    Expr* config = Clone(KCE->getConfig());
+
+    // If the function is non_differentiable, return zero derivative.
+    if (clad::utils::hasNonDifferentiableAttribute(KCE)) {
```
parth-07, line 2287: Can you please add a test for non-differentiable CUDA kernel?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2268,12 +2272,403 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+          m_Sema
+              .ActOnCallExpr(getCurrentScope(), Clone(KCE->getCallee()),
+                             validLoc, ClonedArgs, validLoc, config)
+              .get();
+      // Creating a zero derivative
+      auto* zero = ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context,
+                                                     /*val=*/0);
+
+      // Returning the function call and zero derivative
+      return StmtDiff(Call, zero);
```
parth-07, line 2304: Why are we returning `zero` instead of a `nullptr` here?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2268,12 +2272,403 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     return {};
   }
 
+  StmtDiff
+  ReverseModeVisitor::VisitCUDAKernelCallExpr(const CUDAKernelCallExpr* KCE) {
```
parth-07, line 2276: It seems that there is a lot of code-duplication in `VisitCallExpr` and `VisitCUDAKernelCallExpr`. Can you please refactor the codes to reduce code-duplication? Can we accomodate `CUDAKernelCallExpr` functionality in `VisitCallExpr` instead of having a new visitor function?

parth-07 (changes requested):

test/CUDA/GradientKernels.cu
```diff
@@ -412,6 +412,81 @@ __global__ void kernel_with_nested_device_call(double *out, double *in, double v
...
+//CHECK-NEXT:        {
+//CHECK-NEXT:            res = clad::pop(_t1);
+//CHECK-NEXT:            double _r_d0 = _d_res;
+//CHECK-NEXT:            _d_out_host[i] += _r_d0;
+//CHECK-NEXT:        }
+//CHECK-NEXT:    }
+//CHECK-NEXT:    {
+//CHECK-NEXT:        unsigned long _r0 = 0UL;
+//CHECK-NEXT:        cudaMemcpyKind _r1 = static_cast<cudaMemcpyKind>(0U);
+//CHECK-NEXT:        clad::custom_derivatives::cudaMemcpy_pullback(out_host, out, 10 * sizeof(double), cudaMemcpyDeviceToHost, _d_out_host, _d_out, &_r0, &_r1);
```
parth-07, line 481: This will copy the contents of `_d_out_host` to `_d_out`, right? We should not override the seeds (`_d_out`) provided by a user. Users set the seed values to get the derivatives they want.

test/CUDA/GradientKernels.cu
```diff
@@ -412,6 +412,81 @@ __global__ void kernel_with_nested_device_call(double *out, double *in, double v
...
+  return res;
+}
+
+// CHECK: void fn_memory_grad(double *out, double *in, double *_d_out, double *_d_in) {
+//CHECK-NEXT:    int _d_i = 0;
+//CHECK-NEXT:    int i = 0;
+//CHECK-NEXT:    clad::tape<double> _t1 = {};
+//CHECK-NEXT:    kernel_call<<<1, 10>>>(out, in);
+//CHECK-NEXT:    cudaDeviceSynchronize();
+//CHECK-NEXT:    double *_d_out_host = (double *)malloc(10 * sizeof(double));
```
parth-07, line 450: Should we zero-initialize `_d_out_host`? Currently, its value is undefined and its value is being used. Accessing an undefined value can result in undefined-behavior.

test/CUDA/GradientKernels.cu
```diff
@@ -412,6 +412,81 @@ __global__ void kernel_with_nested_device_call(double *out, double *in, double v
...
+//CHECK-NEXT:            double _r_d0 = _d_res;
+//CHECK-NEXT:            _d_out_host[i] += _r_d0;
+//CHECK-NEXT:        }
+//CHECK-NEXT:    }
+//CHECK-NEXT:    {
+//CHECK-NEXT:        unsigned long _r0 = 0UL;
+//CHECK-NEXT:        cudaMemcpyKind _r1 = static_cast<cudaMemcpyKind>(0U);
+//CHECK-NEXT:        clad::custom_derivatives::cudaMemcpy_pullback(out_host, out, 10 * sizeof(double), cudaMemcpyDeviceToHost, _d_out_host, _d_out, &_r0, &_r1);
+//CHECK-NEXT:    }
+//CHECK-NEXT:    kernel_call_pullback<<<1, 10>>>(out, in, _d_out, _d_in);
```
parth-07, line 483: Should we call `cudaDeviceSynchronize` before `cudaFree` calls?

parth-07 (changes requested):

test/CUDA/GradientKernels.cu
kchristin22: These should be executed in order so no need

test/CUDA/GradientKernels.cu
kchristin22: Yes, but this is out of the scope of this PR

test/CUDA/GradientKernels.cu
kchristin22: I updated that part to use a local array for cudaMemcpy and then add the data to _d_out (via atomicAdd if its a device pointer)

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -82,6 +82,55 @@ ValueAndPushforward<int, int> cudaDeviceSynchronize_pushforward()
     __attribute__((host)) {
   return {cudaDeviceSynchronize(), 0};
 }
+
+template <typename T>
+__global__ void atomicAdd_kernel(T* destPtr, T* srcPtr, size_t N) {
+  for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < N;
```
parth-07, line 88:
```suggestion
  for (size_t i = blockIdx.x * blockDim.x + threadIdx.x; i < N;
```

`int i` and `size_t N` have different types, thus different ranges. You can use `auto` as well.

lib/Differentiator/ReverseModeVisitor.cpp
parth-07: This review has not been addressed. The same comment was present at two places. It seems that you have fixed the similar comment at other place. Can you please fix this as well?

test/CUDA/GradientKernels.cu
```diff
@@ -412,6 +412,81 @@ __global__ void kernel_with_nested_device_call(double *out, double *in, double v
...
+  kernel_call<<<1, 10>>>(out, in);
+}
+
+// CHECK: void fn_grad(double *out, double *in, double *_d_out, double *_d_in) {
+//CHECK-NEXT:     kernel_call<<<1, 10>>>(out, in);
+//CHECK-NEXT:     kernel_call_pullback<<<1, 10>>>(out, in, _d_out, _d_in);
+//CHECK-NEXT: }
+
+double fn_memory(double *out, double *in) {
+  kernel_call<<<1, 10>>>(out, in);
```
parth-07, line 430: Can you please add another test in which the function first copies content from the host memory to device memory before calling the kernel?

test/CUDA/GradientKernels.cu
parth-07: That sounds great!

test/CUDA/GradientKernels.cu
parth-07: It is directly affecting this PR. We should not rely on undefined behavior nor should we merge test containing undefined behavior. Can you please fix the zero-initialize issue in a separate PR?

parth-07 (changes requested):

test/CUDA/GradientKernels.cu
parth-07: > meaning the previous GPU ops will finish before free is executed as they are all executed in the default stream.

What if there are no `cudaFree` calls? Should we call `cudaDeviceSynchronize` in that case?

> Maybe it would be worth it to add them in case the user uses another stream for the kernel, as they should also include cudaDeviceSynchronize() in the original kernel. What do you think?

I believe we should not add additional `cudaDeviceSynchronize` calls because of multiple streams currently. We can add support for multiple stream by adding extra synchronization calls if the need arises.

kchristin22: They are supported for derivative kernel execution though. I can remove the `cudaDeviceSynchronize` calls from `cudaMemcpy_pullback`, but I think it doesn't hurt to keep them, as they will be added in the future either way.

test/CUDA/GradientKernels.cu
kchristin22: Yes, I altered this example to make it complete

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1940,7 +1952,8 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       OverloadedDerivedFn =
           m_Builder.BuildCallToCustomDerivativeOrNumericalDiff(
               customPushforward, pushforwardCallArgs, getCurrentScope(),
-              const_cast<DeclContext*>(FD->getDeclContext()));
+              const_cast<DeclContext*>(FD->getDeclContext()), true, true,
```
vgvassilev, line 1955:
```suggestion
              const_cast<DeclContext*>(FD->getDeclContext()), /*forCustomDerv=*/true, /*namespaceShouldExist=*/true,
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2041,7 +2054,8 @@
       OverloadedDerivedFn =
           m_Builder.BuildCallToCustomDerivativeOrNumericalDiff(
               customPullback, pullbackCallArgs, getCurrentScope(),
-              const_cast<DeclContext*>(FD->getDeclContext()));
+              const_cast<DeclContext*>(FD->getDeclContext()), true, true,
```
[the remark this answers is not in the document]
vgvassilev: Likewise.

parth-07 (approved):
Looks good.

PR #1116, opened 2024-10-14
Kokkos reverse mode improvements

gojakuch:


vgvassilev:
@kliegeois, @brian-kelley can you take a look?

brian-kelley (commented):
This looks fine to me! My only suggestion is that you could probably get away with hardcoding ``size_t`` as the type for all the resize dimensions and their derivatives, since those arguments to ``Kokkos::resize`` are always size_t.

vgvassilev:
> This looks fine to me! My only suggestion is that you could probably get away with hardcoding `size_t` as the type for all the resize dimensions and their derivatives, since those arguments to `Kokkos::resize` are always size_t.

Thanks for the comment. I am confused, are you suggesting reduce the template specializations to one which takes a size_t argument?

gojakuch:
> This looks fine to me! My only suggestion is that you could probably get away with hardcoding `size_t` as the type for all the resize dimensions and their derivatives, since those arguments to `Kokkos::resize` are always size_t.

I think it's better to keep this as templates, since sometimes Clad generates numbers of different types depending on the argument and thus may fail to find this custom derivative in some edge cases (such as when using both literal and variable indices of different types as arguments).

vgvassilev:
> > This looks fine to me! My only suggestion is that you could probably get away with hardcoding `size_t` as the type for all the resize dimensions and their derivatives, since those arguments to `Kokkos::resize` are always size_t.
> 
> I think it's better to keep these as templates, since sometimes Clad generates numerical values of different types depending on the argument and thus may fail to find this custom derivative in some edge cases (such as when using both literal and variable indices of different types as arguments).

Can we check if that is really limitation? If we can drop all of these template arguments it would be become significantly more readable.

gojakuch:
I've rebased this PR and managed to reduce the number of lines by having `std::size_t` arguments in some cases. seems like it's no longer an issue there.

vgvassilev:
@brian-kelley, were these changes what you were aiming for?

brian-kelley:
@vgvassilev @gojakuch Yes, that's what I had in mind. Thanks!

PR #1117, opened 2024-10-15
Store/restore reference args only if they are lvalue and non-const

PetroZarytskyi:
Currently, in the reverse mode, the argument is stored when the function parameter is of a reference type and the argument is not a temporary expression (e.g. ``0`` when bound to a ``const int&`` parameter). A better approach to this is checking if the parameter is an lvalue non-const reference. This lets us avoid unnecessary stores and errors:
```
double g(const double& t) {...}

double f (...) {
  double x = ...;
  const double y = ...;
  g(x);   // No need to store x since the function isn't able to modify it
  g(y);   // An attempt to restore y will lead to an error because it's a const
}
```
This PR fixes the issue in the ReverseModeVisitor and TBRAnalyzer.


vgvassilev (approved):
LGTM!

PR #1120, opened 2024-10-18
Add Useful Analysis to the forward mode

ovdiiuv:


lib/Differentiator/UsefulAnalyzer.cpp
```diff
@@ -0,0 +1,151 @@
...
+}
+
+void UsefulAnalyzer::AnalyzeCFGBlock(const CFGBlock& block) {
+
+  for (auto ib = block.end(); ib != block.begin() - 1; ib--) {
+    if (ib->getKind() == clang::CFGElement::Statement) {
+
+      const clang::Stmt* S = ib->castAs<clang::CFGStmt>().getStmt();
+      llvm::errs() << "\nblockid: " << m_CurBlockID << "\n";
+      S->dump();
```
PetroZarytskyi: Should this be removed?

lib/Differentiator/UsefulAnalyzer.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+static void mergeVarsData(std::set<const clang::VarDecl*>* targetData,
+                          std::set<const clang::VarDecl*>* mergeData) {
+  for (const clang::VarDecl* i : *mergeData)
+    targetData->insert(i);
+  *mergeData = *targetData;
+}
+
+void UsefulAnalyzer::AnalyzeCFGBlock(const CFGBlock& block) {
+
```
PetroZarytskyi, line 52: I guess no need to have a new line here.

PetroZarytskyi (approved):
Very glad to see useful analysis finally being supported!
I think we can use both useful and varied analyses in both reverse and forward modes, but this is an improvement that can be made outside of this PR. I don't see anything major to hold the PR.
LGTM!

PR #1122, opened 2024-10-23
Add support for parameter specification in Varied Analysis

ovdiiuv:
This PR fixes Varied Analysis to work with parameter specification and improves function calls support; it's no longer assumed that every function is varied by default.


PetroZarytskyi:
LGTM

Issue #682, opened 2023-12-13
Missing num-diff fallback warnings in nested function calls


guitargeek:
Reproducer:
```c++
// Compile with clang++ -std=c++20 -fplugin=/usr/lib/clad.so missing_warning.cxx -o missing_warning

#include "clad/Differentiator/Differentiator.h"

#include <cmath>

inline double innerWrapper(double x)
{
   return std::lgamma(x); // call any function that needs num. diff.
}

double wrapper(double *params)
{
   // When calling the num-diffed function directly, you get the correct
   // warning:
   // return std::lgamma(params[0]);
   return innerWrapper(params[0]);
}

int main()
{
   auto grad = clad::gradient(wrapper, "params");
}
```

When calling `std::lgamma` directly, you get this correct warning when compiling:
```txt
warning: Falling back to numerical differentiation for 'lgamma' since no suitable overload was found and clad could not derive it. To disable this feature, compile your programs with -DCLAD_NO_NUM_DIFF.
1 warning generated.
```
However, the warning is absent in the reproducer with the nested function call.

It would be nice if you also get the warning in that situation, because these warnings are very useful to spot residual num. diffs in the gradients.

PR #1124, opened 2024-10-23
Initialize derivative pointers that are allocated through malloc or realloc

kchristin22:
This PR identifies calls to `malloc` or `realloc` and for the derivative pointers it:
* replaces `malloc` with `calloc`
* adds a `calloc` call after `realloc`

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -153,6 +153,37 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     return atomicAddCall;
   }
 
+  Expr* ReverseModeVisitor::CheckAndBuildCallToCalloc(Expr* RHS) {
+    Expr* size = nullptr;
+    if (auto* callExpr = dyn_cast<CallExpr>(RHS))
+      if (auto* implCast = dyn_cast<ImplicitCastExpr>(callExpr->getCallee()))
+        if (auto* declRef = dyn_cast<DeclRefExpr>(implCast->getSubExpr()))
+          if (auto* FD = dyn_cast<FunctionDecl>(declRef->getDecl())) {
+            if (FD->getNameAsString() == "malloc")
+              size = callExpr->getArg(0);
+            else if (FD->getNameAsString() == "realloc")
+              size = callExpr->getArg(1);
+          }
+
+    if (size) {
+      llvm::SmallVector<Expr*, 2> args;
+      if (auto* BinOp = dyn_cast<BinaryOperator>(size)) {
+        if (BinOp->getOpcode() == BO_Mul) {
+          Expr* lhs = BinOp->getLHS();
+          Expr* rhs = BinOp->getRHS();
+          if (auto* sizeofCall = dyn_cast<UnaryExprOrTypeTraitExpr>(rhs))
+            args = {lhs, sizeofCall};
+          else if (auto* sizeofCall = dyn_cast<UnaryExprOrTypeTraitExpr>(lhs))
+            args = {rhs, sizeofCall};
+          if (!args.empty())
+            return GetFunctionCall("calloc", "", args);
+        }
+      }
+    }
+
+    return {};
+  }
+
   ReverseModeVisitor::ReverseModeVisitor(DerivativeBuilder& builder,
                                          const DiffRequest& request)
       : VisitorBase(builder, request), m_Result(nullptr) {}
@@ -1697,25 +1728,6 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         !isa<CXXOperatorCallExpr>(CE))
       return StmtDiff(Clone(CE));
 
-    // If all arguments are constant literals, then this does not contribute to
-    // the gradient.
-    // FIXME: revert this when this is integrated in the activity analysis pass.
-    if (!isa<CXXMemberCallExpr>(CE) && !isa<CXXOperatorCallExpr>(CE)) {
-      bool allArgsAreConstantLiterals = true;
-      for (const Expr* arg : CE->arguments()) {
-        // if it's of type MaterializeTemporaryExpr, then check its
-        // subexpression.
-        if (const auto* MTE = dyn_cast<MaterializeTemporaryExpr>(arg))
-          arg = clad_compat::GetSubExpr(MTE)->IgnoreImpCasts();
-        if (!arg->isEvaluatable(m_Context)) {
-          allArgsAreConstantLiterals = false;
-          break;
-        }
-      }
-      if (allArgsAreConstantLiterals)
-        return StmtDiff(Clone(CE), Clone(CE));
-    }
-
     SourceLocation Loc = CE->getExprLoc();
 
     // Stores the call arguments for the function to be derived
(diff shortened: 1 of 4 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -153,6 +153,37 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     return atomicAddCall;
   }
 
+  Expr* ReverseModeVisitor::CheckAndBuildCallToCalloc(Expr* RHS) {
```
parth-07, line 156: Why is this function so complicated? Instead of manually parsing the `size` expression to determine `calloc` arguments, can we simply create the call as `calloc(1, sizeExpr)` where `sizeExpr` is from the original malloc call `malloc(sizeExpr)`?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2842,6 +2877,19 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         ComputeEffectiveDOperands(Ldiff, Rdiff, derivedL, derivedR);
         addToCurrentBlock(BuildOp(opCode, derivedL, derivedR),
                           direction::forward);
+        if (opCode == BO_Assign && derivedR)
```
parth-07, line 2880: Why is this required? I thought this PR is only modifying the differentiation of `malloc` calls.

lib/Differentiator/VisitorBase.cpp
```diff
@@ -531,11 +531,14 @@ namespace clad {
...
                                      llvm::SmallVectorImpl<Expr*>& callArgs) {
-    NamespaceDecl* NSD =
-        utils::LookupNSD(m_Sema, nmspace, /*shouldExist=*/true);
-    DeclContext* DC = NSD;
+    NamespaceDecl* NSD = nullptr;
     CXXScopeSpec SS;
-    SS.Extend(m_Context, NSD, noLoc, noLoc);
+
+    if (!nmspace.empty()) {
+      NSD = utils::LookupNSD(m_Sema, nmspace, /*shouldExist=*/true);
```
parth-07, line 538: Instead of adding so many conditions in this function, can you please modify `utils::LookupNSD` to return top-level namespace declaration if `nmspace` is an empty string?

parth-07 (changes requested):

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -153,6 +153,37 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     return atomicAddCall;
   }
 
+  Expr* ReverseModeVisitor::CheckAndBuildCallToCalloc(Expr* RHS) {
+    Expr* size = nullptr;
+    if (auto* callExpr = dyn_cast<CallExpr>(RHS))
+      if (auto* implCast = dyn_cast<ImplicitCastExpr>(callExpr->getCallee()))
+        if (auto* declRef = dyn_cast<DeclRefExpr>(implCast->getSubExpr()))
```
vgvassilev, line 160:
```suggestion
      if (auto* declRef = dyn_cast<DeclRefExpr>(callExpr->IgnoreImpCasts()))
```

test/Gradient/Pointers.C
```diff
@@ -407,8 +407,8 @@ double cStyleMemoryAlloc(double x, size_t n) {
 
 // CHECK: void cStyleMemoryAlloc_grad_0(double x, size_t n, double *_d_x) {
 // CHECK-NEXT:     size_t _d_n = 0UL;
-// CHECK-NEXT:     T *_d_t = (T *)malloc(n * sizeof(T));
-// CHECK-NEXT:     T *t = (T *)malloc(n * sizeof(T));
+// CHECK-NEXT:     T *_d_t = (T *)calloc(n, sizeof(T));
```
vgvassilev, line 410: Instead of rewiring malloc to calloc, can we do malloc and them memset? This would theoretically work better if we have a user-provided `malloc` operation.

kchristin22: I tried that in the beginning, but if the user uses `memset` afterwards on the same pointer, the `memset` call would be repeated again:
```cpp
T *_d_t = (T *)malloc(n * sizeof(T));
memset(_d_t, 0, n * sizeof(T)); // emplace memset call after malloc
T *t = (T *)malloc(n * sizeof(T));
memset(_d_t, 0, n * sizeof(T));
memset(t, 0, n * sizeof(T));
```
And I don't think we should skip creating the second `memset`- meaning when we encounter a `memset` call in the original function, we should assign its derivative on the derivative pointer as well, because it's not necessary for a `memset` call to follow a `malloc` call.

I'm not sure I understand why using `memset` instead should make a difference when the `malloc` function is user provided, as either way we wouldn't be able to tell it's a memory allocation function, so we wouldn't add the `memset` call afterwards f.i. Currently, we only check for C memory functions.

lib/Differentiator/ReverseModeVisitor.cpp
kchristin22: `calloc` function is used like so:
```cpp
void* calloc( size_t num, size_t size );

Allocates memory for an array of num objects of size and initializes all bytes in the allocated storage to zero.
```
While the same `malloc` call would look like: (T*)malloc(num * size)

So we have to split that `sizeExpr` into num and size.

lib/Differentiator/ReverseModeVisitor.cpp
kchristin22: I took it one step further and covered realloc as well where there's not a decl stmt, but this may be required for malloc functions also. For example:
```cpp
double *ptr = nullptr;
ptr = (double *)malloc(10 * sizeof(double));
```

test/Gradient/Pointers.C
vgvassilev: Here is the example I have in mind: https://godbolt.org/z/rfabzjTa8

If we rewire the memory calls that would ignore the use-defined logic in `malloc`...

lib/Differentiator/ReverseModeVisitor.cpp
kchristin22: I see what you mean now, I was worried that it wouldn't initialize the whole block correctly if we treated it as one, but I think it should work after all. I will change that.

kchristin22: I will revert back to `memset` after Vassil's comment so I mark this as resolved.

lib/Differentiator/VisitorBase.cpp
kchristin22: If I understand correctly, the function you're mentioning already does that, so I can revert my changes altogether. Can you take a look?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1788,6 +1790,25 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       return StmtDiff();
     }
 
+    // If all arguments are constant literals, then this does not contribute to
+    // the gradient.
+    // FIXME: revert this when this is integrated in the activity analysis pass.
```
vgvassilev, line 1795: @ovdiiuv, can you comment here?

ovdiiuv: I thought about that recently. I don't think it has to be reverted, rather modified to track arguments that are elements of constant sets for example. This would resolve #682.

vgvassilev: But that should be part of the analysis not here, right? That is, it needs to be implemented on your end.

lib/Differentiator/VisitorBase.cpp
parth-07: `utils::LookupNSD` does not seem to return top-level namespace (translation unit decl) if `nmspec` is an empty string. Why do you think it does?

lib/Differentiator/ReverseModeVisitor.cpp
ovdiiuv: No, I think it's the right place.

parth-07 (approved):

PR #1126, opened 2024-10-29
Lambda support in the reverse mode

### LINKED ISSUE #1054: Change lambda function differentiation to produce valid source code and work with lambda functions with captures (reverse mode)
currently, if a function contains a lambda expression, Clad treats it as a special case of a functor. there's no general support for lambdas with captures yet. however, I believe at this point we should stop treating lambda functions as functors but instead generate separate lambda functions representing the pullback of the initial lambda. this way Clad wouldn't need to modify the lambda class that is impossible for the user to manipulate.

pullbacks of the functions that accept lambdas as their arguments would then need to accept both the initial lambda `f` and its pullback `d_f`.

this ensures valid source code generation.

gojakuch:
Potentially fixes: #1054

### DIFF
--- a/include/clad/Differentiator/CladUtils.h
+++ b/include/clad/Differentiator/CladUtils.h
@@ -158,6 +158,8 @@ namespace clad {
     clang::StringLiteral* CreateStringLiteral(clang::ASTContext& C,
                                               llvm::StringRef str);
 
+    bool isLambdaQType(clang::QualType QT);
+
     /// Returns true if `QT` is Array or Pointer Type, otherwise returns false.
     bool isArrayOrPointerType(clang::QualType QT);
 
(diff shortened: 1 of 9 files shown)


gojakuch:
I rebased this PR (up to some most recent changes) and tried to continue pushing this through, but to no avail. I don't think this is quickly fixable in the matter weeks anymore, so I'm leaving all of the code for my attempts committed here, in case anyone decides to take this off at some point. it should be possible to implement this in general, but I've been beating around the bush here for too long and I've got no idea how exactly can this be implemented correctly. I could generate some correct code at some point, but Clad would crash after that because the AST was ill-formed. now it's all completely gone. maybe I'll get back to this later and remake this from scratch, once I've got more time on my hands

vgvassilev:
> I rebased this PR (up to some most recent changes) and tried to continue pushing this through, but to no avail. I don't think this is quickly fixable in the matter weeks anymore, so I'm leaving all of the code for my attempts committed here, in case anyone decides to take this off at some point. it should be possible to implement this in general, but I've been beating around the bush here for too long and I've got no idea how exactly can this be implemented correctly. I could generate some correct code at some point, but Clad would crash after that because the AST was ill-formed. now it's all completely gone. maybe I'll get back to this later and remake this from scratch, once I've got more time on my hands

Hi @gojakuch. Thanks for the efforts. It is really unfortunate outcome of the project but I hope we can continue from where you are leaving it. Best of luck and we are looking forward to having you back again!

PR #1131, opened 2024-11-03
Avoid creation of local derivative of const parameter

### LINKED ISSUE #1130: Derivative of const value input of function is declared as const
When a function has a const input that doesn't affect the derivative, but is part of a binary operation, its derivative is updated. However, this derivative is currently declared as const, so the following example, for instance, produces an error:
```cpp
double constPointer(double y, const double x) {
  return y * x;
}

auto const_test = clad::gradient(constPointer, "y");

void constPointer_grad_0(double y, const double x, double *_d_y) {
    const double _d_x = 0.;
    {
        *_d_y += 1 * x;
        _d_x += y * 1; // this line produces the error
    }
}

```

kchristin22:
Closes #1130

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -4566,6 +4566,12 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     return subExprDiff;
   }
 
+  StmtDiff ReverseModeVisitor::VisitCXXConstCastExpr(
+      const clang::CXXConstCastExpr* CCE) {
+    StmtDiff subExprDiff = Visit(CCE->getSubExpr(), dfdx());
+    return {Clone(CCE), subExprDiff.getExpr_dx()};
+  }
+
   clang::QualType ReverseModeVisitor::ComputeAdjointType(clang::QualType T) {
     if (T->isReferenceType()) {
       QualType TValueType = utils::GetValueType(T);
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -706,6 +706,8 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
           i == m_DiffReq->getNumParams() - 1)
         continue;
       auto VDDerivedType = param->getType();
+      if (VDDerivedType.isConstQualified())
+        continue;
       // We cannot initialize derived variable for pointer types because
       // we do not know the correct size.
       if (utils::isArrayOrPointerType(VDDerivedType))
--- a/test/Gradient/Assignments.C
+++ b/test/Gradient/Assignments.C
@@ -819,6 +819,34 @@ double f23(double x, double y) {
 //CHECK-NEXT:     }
 //CHECK-NEXT: }
 
+double constVal(double y, const double x) {
+  const double z = y;
+  y *= z;
+  return y * x;
+}
+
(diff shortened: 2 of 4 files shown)


test/Gradient/Assignments.C
```diff
@@ -819,6 +819,37 @@ double f23(double x, double y) {
 //CHECK-NEXT:     }
 //CHECK-NEXT: }
 
+double constVal(double y, const double x) {
+  const double z = y;
+  y *= z;
+  return y * x;
+}
```
vgvassilev, line 826: Can you add a test doing something like:
```cpp
double constValConstCast(double y, const double x) {
  double z = const_cast<double>(y);
  y *= z;
  return y * x;
}
```

kchristin22: You mean to try and make z const through const_cast?

vgvassilev: No, the opposite. We strip away the constness here.

vgvassilev: Sorry, I saw my mistake. Corrected the example.

kchristin22: I added it and verified the results, let me know if it's ok

vgvassilev (approved):
LGTM!

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -706,6 +706,8 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
           i == m_DiffReq->getNumParams() - 1)
         continue;
       auto VDDerivedType = param->getType();
+      if (VDDerivedType.isConstQualified())
```
parth-07, line 709: Sorry for the late comment. We should not skip creating the derivative variable for `const` `class` type variables. For example, in the below function, `a` is `const` but (I think) we still need to create `_d_a`.

```cpp
struct A {
double val;
double identity(double i) const {
  return i;
}
};

double fn(double u, const A a) {
  double res = a.identity(u);
  return res;
}
```

kchristin22: I see...
```cpp
void fn_grad_0(double u, const A a, double *_d_u) {
    A _d_a = {};
    A _t0 = a;
    double _d_res = 0.;
    double res = a.identity(u);
    _d_res += 1;
    {
        double _r0 = 0.;
        _t0.identity_pullback(u, _d_res, &_d_a, &_r0);
        *_d_u += _r0;
    }
}
void identity_pullback(double i, double _d_y, A *_d_this, double *_d_i) const {
    *_d_i += _d_y;
}
```
Typically, we don't need _d_a, right? The issues will arise by the way pullback signatures are formed or numerical diff handles these args as far as I understand. I guess for the first case we can skip creating the param if the original call arg is a const function param.

parth-07: > The issues will arise by the way pullback signatures are formed or numerical diff handles these args as far as I understand.

Yes, you are right.

> I guess for the first case we can skip creating the param if the original call arg is a const function param.

Yes, I think so. However, I think it might be simpler to create non-`const` derivatives of `const` parameters and leave it up to activity/useful analysis to determine which parts of the code can be dropped.

kchristin22: Agreed, I'll open a PR with the appropriate change.

vgvassilev: We should check if the underlying type is of aggregate type and copyable/assignable, otherwise the PR seems fine, right?

PR #1132, opened 2024-11-03
Add tensor contraction demo

kchristin22:


test/CUDA/tensor_demo.cu
```diff
@@ -0,0 +1,245 @@
+// RUN: %cladclang_cuda -I%S/../../include --cuda-path=%cudapath \
+// RUN:     --cuda-gpu-arch=%cudaarch %cudaldflags -otensor_demo.out %s
+// RUN: ./tensor_demo.out | %filecheck_exec %s
+// REQUIRES: cuda-runtime
+
```
vgvassilev, line 5: We will probably need some documentation of what's the intent of this demo.

kchristin22: Yeap, will add a section in the documentation for cuda support in general and include that as well

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1893,7 +1893,65 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         QualType dArgTy = getNonConstType(arg->getType(), m_Context, m_Sema);
         VarDecl* dArgDecl = BuildVarDecl(dArgTy, "_r", getZeroInit(dArgTy));
         PreCallStmts.push_back(BuildDeclStmt(dArgDecl));
-        CallArgDx.push_back(BuildDeclRef(dArgDecl));
+        DeclRefExpr* dArgRef = BuildDeclRef(dArgDecl);
+        if (isa<CUDAKernelCallExpr>(CE)) {
```
vgvassilev, line 1897: I feel a little uncomfortable that this change is only tested by the demo. Can we add these changes in a separate PR along with appropriate tests and have a PR only adding the demo?

kchristin22: Sure, no problem!

PR #1133, opened 2024-11-03
Fix _r local vars being passed to non-ref cuda kernel pullbacks

kchristin22:


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1893,7 +1893,65 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+          LookupResult result(m_Sema, idInfo, SourceLocation(),
+                              Sema::LookupOrdinaryName);
+          m_Sema.LookupName(result, m_Sema.getCurScope());
+          EnumDecl* cudaMemcpyKindDecl = nullptr;
+          for (NamedDecl* decl : result)
+            if (auto* enumDecl = dyn_cast<EnumDecl>(decl)) {
+              cudaMemcpyKindDecl = enumDecl;
+              break;
+            }
+          assert(cudaMemcpyKindDecl && "cudaMemcpyKind not found");
```
vgvassilev, line 1918: Can we change that to a clang diagnostic?

vgvassilev:
Any reasonable way to test the diagnostics?

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -91,7 +91,7 @@ __global__ void atomicAdd_kernel(T* destPtr, T* srcPtr, size_t N) {
 }
 
 template <typename T>
-void cudaMemcpy_pullback(T* destPtr, T* srcPtr, size_t count,
+void cudaMemcpy_pullback(T* destPtr, const T* srcPtr, const size_t count,
```
parth-07, line 94: Is `const` required in `const size_t count`?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1893,7 +1893,72 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+          m_Sema.LookupName(result, m_Sema.getCurScope());
+          EnumDecl* cudaMemcpyKindDecl = nullptr;
+          for (NamedDecl* decl : result)
+            if (auto* enumDecl = dyn_cast<EnumDecl>(decl)) {
+              cudaMemcpyKindDecl = enumDecl;
+              break;
+            }
+          if (!cudaMemcpyKindDecl) {
+            diag(DiagnosticsEngine::Warning, CE->getEndLoc(),
+                 "Failed to create cudaMemcpy call; cudaMemcpyKind not found");
```
parth-07, line 1920: We should error-out or at least stop creating this function call if `cudaMemcpyKindDecl` is not found. Otherwise, we have an undefined behavior when calling `cudaMemcpyKindDecl->enumerators()`.

include/clad/Differentiator/BuiltinDerivatives.h
kchristin22: No, you're right, it's a not a ref or a pointer

lib/Differentiator/ReverseModeVisitor.cpp
kchristin22: I agree. I like the idea of the diagnostics, being more readable than the previous `assert` stmt I had, so I changed it to an `Error` instead to avoid creating an output for the user to execute. In case this is hit, there will be some NULL stmts there maybe or hit an internal clang assert, but in the first case it will be more likely they will be able to tell why `cudaMemcpyKind` was needed in the first place.

kchristin22:
> Any reasonable way to test the diagnostics?

I guess if we only omit the include paths for cuda, but these are included in the CMakeLists right?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1893,7 +1893,72 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         QualType dArgTy = getNonConstType(arg->getType(), m_Context, m_Sema);
         VarDecl* dArgDecl = BuildVarDecl(dArgTy, "_r", getZeroInit(dArgTy));
         PreCallStmts.push_back(BuildDeclStmt(dArgDecl));
-        CallArgDx.push_back(BuildDeclRef(dArgDecl));
+        DeclRefExpr* dArgRef = BuildDeclRef(dArgDecl);
+        if (isa<CUDAKernelCallExpr>(CE)) {
+          // Create variables to be allocated on the device and passed to
+          // kernel. These need to be pointers because cudaMlloc expects a
```
vgvassilev, line 1899:
```suggestion
          // kernel. These need to be pointers because cudaMalloc expects a
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1893,7 +1893,72 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+          IdentifierInfo* idInfo = &m_Context.Idents.get("cudaMemcpyKind");
+          LookupResult result(m_Sema, idInfo, SourceLocation(),
+                              Sema::LookupOrdinaryName);
+          m_Sema.LookupName(result, m_Sema.getCurScope());
+          EnumDecl* cudaMemcpyKindDecl = nullptr;
+          for (NamedDecl* decl : result)
+            if (auto* enumDecl = dyn_cast<EnumDecl>(decl)) {
+              cudaMemcpyKindDecl = enumDecl;
+              break;
+            }
```
vgvassilev, line 1917: Why we iterate here?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1893,7 +1893,72 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+          if (!cudaMemcpyKindDecl) {
+            diag(DiagnosticsEngine::Error, CE->getEndLoc(),
+                 "Failed to create cudaMemcpy call; cudaMemcpyKind not found");
+          }
+          QualType cudaMemcpyKindType =
+              m_Context.getTypeDeclType(cudaMemcpyKindDecl);
+          EnumConstantDecl* deviceToHostEnumDecl = nullptr;
+          for (EnumConstantDecl* enumConst :
+               cudaMemcpyKindDecl->enumerators()) {
+            if (enumConst->getName() == "cudaMemcpyDeviceToHost") {
```
vgvassilev, line 1927: We should use lookups. Then we can diagnose easier if we did not find anything.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1893,7 +1893,72 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
+              sizeLiteral, deviceToHostDeclRef};
+          PostCallStmts.push_back(
+              GetFunctionCall("cudaMemcpy", "", cudaMemcpyArgs));
+          llvm::SmallVector<Expr*, 3> freeArgs = {BuildDeclRef(dArgDeclCUDA)};
+          PostCallStmts.push_back(GetFunctionCall("cudaFree", "", freeArgs));
+
+          // Update arg to be passed to pullback call
+          dArgRef = BuildDeclRef(dArgDeclCUDA);
+        }
+        CallArgDx.push_back(dArgRef);
```
vgvassilev, line 1961: We need some documentation for this entire block explaining what we are trying to do.

include/clad/Differentiator/CladUtils.h
```diff
@@ -128,6 +128,16 @@ namespace clad {
                                         clang::DeclContext* DC1,
                                         clang::DeclContext* DC2);
 
+    /// Finds the qualified name `name` in the declaration context `DC`.
+    ///
+    /// \param[in] name
+    /// \param[in] S
+    /// \param[in] DC
+    /// \returns lookup result.
+    clang::LookupResult LookupQualifiedName(llvm::StringRef name,
```
parth-07, line 137: This utility function is going to be very helpful!

include/clad/Differentiator/Differentiator.h
```diff
@@ -126,7 +126,7 @@ CUDA_HOST_DEVICE T push(tape<T>& to, ArgsT... val) {
 #if defined(__CUDACC__) && !defined(__CUDA_ARCH__)
     if (CUDAkernel) {
       constexpr size_t totalArgs = sizeof...(args) + sizeof...(Rest);
-      std::array<void*, totalArgs> argPtrs = {static_cast<void*>(&args)...,
+      std::array<void*, totalArgs> argPtrs = {(void*)(&args)...,
```
parth-07, line 129: Did you change the cast here because `static_cast` does not change `const`ness?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1891,7 +1891,81 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         QualType dArgTy = getNonConstType(arg->getType(), m_Context, m_Sema);
         VarDecl* dArgDecl = BuildVarDecl(dArgTy, "_r", getZeroInit(dArgTy));
         PreCallStmts.push_back(BuildDeclStmt(dArgDecl));
-        CallArgDx.push_back(BuildDeclRef(dArgDecl));
+        DeclRefExpr* dArgRef = BuildDeclRef(dArgDecl);
+        if (isa<CUDAKernelCallExpr>(CE)) {
+          // Create variables to be allocated and initialized on the device, and
+          // then be passed to the kernel pullback.
+          //
+          // These need to be pointers because cudaMalloc expects a double
```
parth-07, line 1899: nitpick: double pointer => pointer-to-pointer.

include/clad/Differentiator/Differentiator.h
kchristin22: I think it was due to the original call args being `const` and so they couldn't be cast to `void *`. I thought about using `const_cast` after our discussions on the other PR, but I would have to determine which args are `const` and use `const_cast` inside the `static_cast` only for them.

parth-07: We only cast adjoint args to `void *`, right? Can you please tell in which case will an adjoint arg be of `const` type?

kchristin22: These are all the args that will be passed to the kernel, both for the forward pass args and their adjoints. The adjoints are not const, but the original parameters may be.

parth-07: Oh, I see.

parth-07 (approved):
Looks good.

Issue #428, opened 2022-04-20
Multiple test assertion failures on master


grimmmyshini:
There are multiple assertion failures on debug master.  Of all, some include: 

```
void clang::CodeGen::CodeGenFunction::EmitCallArgs(clang::CodeGen::CallArgList&, const T*, llvm::iterator_range<clang::Stmt::CastIterator<clang::Expr, const clang::Expr* const, const clang::Stmt* const> >, clang::CodeGen::CodeGenFunction::AbstractCallee, unsigned int, clang::CodeGen::CodeGenFunction::EvaluationOrder) [with T = clang::FunctionProtoType]: 
Assertion `(isGenericMethod || ((*I)->isVariablyModifiedType() || (*I).getNonReferenceType()->isObjCRetainableType() || getContext() .getCanonicalType((*I).getNonReferenceType()) .getTypePtr() == getContext() .getCanonicalType((*Arg)->getType()) .getTypePtr())) && "type mismatch in call argument!"' failed.
```

and 

```
void clang::ClassTemplateSpecializationDecl::setPointOfInstantiation(clang::SourceLocation): 
Assertion `Loc.isValid() && "point of instantiation must be valid!"' failed.
```

vaithak:
We can close this one to,o now that we have a CI test for debug mode and it is passing.
cc: @vgvassilev

vgvassilev:
@vaithak, not really because they are marked as `XFAIL: asserts` we still have work to do here.

vgvassilev:
This depends on the approach solving #248 where we delay all of the differentiation until the end of the translation unit.

PR #1137, opened 2024-11-13
Always use valid location when generating operators

PetroZarytskyi:
A valid location is required by the debug clang.
The commit is taken from #1121.

Partially addresses #428


vgvassilev (approved):
Lgtm!

PR #1138, opened 2024-11-14
Improve array_expression/array/array_ref operators

PetroZarytskyi:
This PR introduces assignment operators between ``array_ref`` and ``array``, and replaces all ``+``, ``-``, ``*``, ``/``operators between ``array_expression``/``array``/``array_ref``/scalar with a single template operator per operation.


vgvassilev (approved):
LGTM!

PR #1139, opened 2024-11-14
Add CUDA demos

kchristin22:
This PR adds the following demos:
* custom TensorContraction demo
* NVIDIA's Black-Scholes sample

### DIFF
--- a/demos/CUDA/BlackScholes/BlackScholes.cu
+++ b/demos/CUDA/BlackScholes/BlackScholes.cu
@@ -35,10 +35,12 @@
  * DISCLAIMER: The following file has been slightly modified to ensure
  * compatibility with Clad and to serve as a Clad demo. Specifically, parts of
  * the original `main` function have been moved to a separate function to use
- * `clad::gradient` on. Furthermore, original print statements have been removed
- * and new helper functions are now included in the file to verify the
- * gradient's results. The original file is available in NVIDIA's cuda-samples
- * repository on GitHub.
+ * `clad::gradient` on. Furthermore, Clad cannot clone printf statements so some
+ * original print statements have been ommitted. The same applies to the
+ * checkCudaErrors function.
+ * New helper functions are included in another file and invoked here to verify
+ * the gradient's results. The original file is available in NVIDIA's
+ * cuda-samples repository on GitHub.
  *
  * Relevant documentation regarding the problem at hand can be found in NVIDIA's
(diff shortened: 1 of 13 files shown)


demos/CUDA/BlackScholes/BlackScholes.cu
```diff
@@ -0,0 +1,405 @@
...
+
+/*
+ * This sample evaluates fair call and put prices for a
+ * given set of European options by Black-Scholes formula.
+ * See supplied whitepaper for more explanations.
+ */
+
+/*
+ * DISCLAIMER: The following file has been modified slightly to make it
+ * compatible with Clad. The original file can be found at NVIDIA's cuda-samples
```
vgvassilev, line 36: Can we describe what the modifications are on a high-level?

demos/CUDA/TensorContraction.cu
```diff
@@ -0,0 +1,266 @@
+//--------------------------------------------------------------------*- C++ -*-
```
vgvassilev, line 1: Can we get this demo in a separate PR?

demos/CUDA/BlackScholes/BlackScholes.cu
kchristin22: Hmm I can mention that these changes refer to the original `main` function whose parts have been moved to a separate function to use `clad::gradient` on. There's also the loss of the original print statements and the addition of helper functions to verify the results. Would this do as a whole?

demos/CUDA/TensorContraction.cu
kchristin22: Certainly

demos/CUDA/BlackScholes/BlackScholes.cu
kchristin22: How about now?

vgvassilev: Yes, or just make a diff and put it in comment if it’s not big.

kchristin22: I think that would be confusing, because I kind of restructured the file

kchristin22:
I added as many parts as possible from the original demo to minimize the diff. You're free to have a look in case you have any suggestions!

kchristin22:
@vgvassilev Let's postpone this merge till the r value PR is fixed and merged, as this functionality is needed to run the demo

vgvassilev (approved):
LGTM!

PR #1140, opened 2024-11-14
Use clad::zero_vector instead of 0 in the vectorized forward mode

### LINKED ISSUE #527: Incorect values for Jacobian when one of the outputs is built with intermediate steps
Code to reproduce: 
```
int add(int a, int b) { return a + b ;}
void tes1(int a, int b, int* res){ 
    res[0] = a*10 + b*b*9;
    res[0] = add(res[0], 10);
    res[1] = a*a*9 + b*b*10;
}
void tes2(int a, int b, int* res){ 
    res[0] = a*10 + b*b*9;
    res[0] = res[0] + 10;
    res[1] = a*a*9 + b*b*10;
}
auto jacobian = clad::jacobian(test1);
int matr[4];
int res[2] = {0, 0};
jacobian.execute(3, 5, res, matr);
```
Both versions will output: `[64, 190, 0, 0]` as the Jacobian result gets summed column wise. 

However, if the intermediate step is not present, then the corect result is computed: `[10, 90, 54, 100]`:

```
void tes3t(int a, int b, int* res){ 
    res[0] = a*10 + b*b*9 + 10;
    res[1] = a*a*9 + b*b*10;
}
```

In case of `test1` and `test2` the `jacobianMatrix` indexing is at fault:

```
--------------
{
        int _r6 = 1 * 9;
        int _r7 = _r6 * _t5;
        jacobianMatrix[0UL] += _r7;
        int _r8 = _t6 * _r6;
        jacobianMatrix[0UL] += _r8;
        int _r9 = 1 * 10;
        int _r10 = _r9 * _t7;
        jacobianMatrix[1UL] += _r10;
        int _r11 = _t8 * _r9;
        jacobianMatrix[1UL] += _r11;
    }
    {
        int _jac0 = 0;
        int _jac1 = 0;
        add_pullback(_t4, 10, 1, &_jac0, &_jac1);
        int _r4 = _jac0;
        int _r5 = _jac1;
    }
    {
        int _r0 = 1 * 10;
        jacobianMatrix[0UL] += _r0;
        int _r1 = 1 * 9;
        int _r2 = _r1 * _t2;
        jacobianMatrix[1UL] += _r2;
        int _r3 = _t3 * _r1;
        jacobianMatrix[1UL] += _r3;
    }
--------------

```

In the case of `test3` the Jacobian is generated as: 

```
--------------
{
        int _r4 = 1 * 9;
        int _r5 = _r4 * _t4;
        jacobianMatrix[2UL] += _r5;
        int _r6 = _t5 * _r4;
        jacobianMatrix[2UL] += _r6;
        int _r7 = 1 * 10;
        int _r8 = _r7 * _t6;
        jacobianMatrix[3UL] += _r8;
        int _r9 = _t7 * _r7;
        jacobianMatrix[3UL] += _r9;
    }
    {
        int _r0 = 1 * 10;
        jacobianMatrix[0UL] += _r0;
        int _r1 = 1 * 9;
        int _r2 = _r1 * _t2;
        jacobianMatrix[1UL] += _r2;
        int _r3 = _t3 * _r1;
        jacobianMatrix[1UL] += _r3;
    }
--------------

```

PetroZarytskyi:
This is necessary to ensure type safety. ``0`` represents a zero vector but is only properly treated as a zero vector when it gets added/subtracted with a ``clad::array``. Initializing a ``clad::array`` with ``0`` will create not a zero vector but an empty array (with zero elements). To fix this, we needed exceptions for all declarations. We would need to make new exceptions when passing parameters to pushforwards.

After merging this PR and #1121, #527 will be fixed automatically.

### DIFF
--- a/include/clad/Differentiator/BaseForwardModeVisitor.h
+++ b/include/clad/Differentiator/BaseForwardModeVisitor.h
@@ -58,12 +58,12 @@ public:
   StmtDiff VisitCXXDefaultArgExpr(const clang::CXXDefaultArgExpr* DE);
   StmtDiff VisitDeclRefExpr(const clang::DeclRefExpr* DRE);
   StmtDiff VisitDeclStmt(const clang::DeclStmt* DS);
-  StmtDiff VisitFloatingLiteral(const clang::FloatingLiteral* FL);
+  virtual StmtDiff VisitFloatingLiteral(const clang::FloatingLiteral* FL);
   StmtDiff VisitForStmt(const clang::ForStmt* FS);
   StmtDiff VisitIfStmt(const clang::IfStmt* If);
   StmtDiff VisitImplicitCastExpr(const clang::ImplicitCastExpr* ICE);
   StmtDiff VisitInitListExpr(const clang::InitListExpr* ILE);
-  StmtDiff VisitIntegerLiteral(const clang::IntegerLiteral* IL);
+  virtual StmtDiff VisitIntegerLiteral(const clang::IntegerLiteral* IL);
   StmtDiff VisitMemberExpr(const clang::MemberExpr* ME);
   StmtDiff VisitParenExpr(const clang::ParenExpr* PE);
   virtual StmtDiff VisitReturnStmt(const clang::ReturnStmt* RS);
--- a/include/clad/Differentiator/VectorForwardModeVisitor.h
+++ b/include/clad/Differentiator/VectorForwardModeVisitor.h
@@ -69,6 +69,8 @@ public:
   /// For example: for size = 4, the returned expression is: {0, 0, 0, 0}
   clang::Expr* getZeroInitListExpr(size_t size, clang::QualType type);
 
+  StmtDiff VisitFloatingLiteral(const clang::FloatingLiteral* FL) override;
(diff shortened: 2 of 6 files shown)


include/clad/Differentiator/Array.h
```diff
@@ -51,6 +51,12 @@ template <typename T> class array {
 
   CUDA_HOST_DEVICE array(const array<T>& arr) : array(arr.m_arr, arr.m_size) {}
 
+  template <typename U>
+  CUDA_HOST_DEVICE array(clad::array<U> arr)
+      : m_arr(new T[arr.size()]{static_cast<T>(T())}), m_size(arr.size()) {
```
vgvassilev, line 56: This seems to initialize only the first element: https://godbolt.org/z/MWs96WxfP

PetroZarytskyi: Agreed. The elements are initialized in the body anyway. Also, the parameter should have been a const reference. I probably copied this from a different constructor and forgot to adjust it properly. Thank you for drawing my attention to this bit.

vgvassilev:
Now the array is uninitialized.

vgvassilev (approved):
LGTM!

PR #1141, opened 2024-11-15
Add CUDA 3D tensor contraction demo

kchristin22:


vgvassilev:
Did you make sure that the output result is as expected?

kchristin22:
Yes, I ran a python script using `torch` instead of Clad for the same set of inputs

vgvassilev:
> Yes, I ran a python script using `torch` instead of Clad for the same set of inputs

Awesome! Great work!

vgvassilev (approved):
LGTM!

PR #1142, opened 2024-11-17
Fix struct initialization and return stmts

kchristin22:


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1894,11 +1894,8 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       const auto* PVD = FD->getParamDecl(
           i - static_cast<unsigned long>(isMethodOperatorCall));
       if (PVD->getType()->isRValueReferenceType()) {
-        IdentifierInfo* PVDII = PVD->getIdentifier();
         IdentifierInfo* RValueName = CreateUniqueIdentifier("_r");
-        auto* param = CloneParmVarDecl(PVD, PVDII);
-        param->setDeclName(RValueName);
-        PVD = param;
+        const_cast<ParmVarDecl*>(PVD)->setDeclName(RValueName);
```
vgvassilev, line 1898: That does not seem right. Why are we changing the name of the target function. Generally we clone and update the cloned values.

kchristin22: This PR tries to fix args not having a name. For example:
```cpp
inline constexpr void operator_equal_pullback(float2 &&, float2 _d_y, float2 *_d_this, float2 *_d_) noexcept {
    float _t0 = this->x;
    this->x = .x; // no float2 var specified
    float _t1 = this->y;
    this->y = .y; // same here
    {
        this->y = _t1;
        float _r_d1 = (*_d_this).y;
        (*_d_this).y = 0.F;
        (*_d_).y += _r_d1;
    }
    {
        this->x = _t0;
        float _r_d0 = (*_d_this).x;
        (*_d_this).x = 0.F;
        (*_d_).x += _r_d0;
    }
}
```

The PR converts this to:
```cpp
inline constexpr void operator_equal_pullback(float2 &&_r14, float2 _d_y, float2 *_d_this, float2 *_d__r14) noexcept {
    float _t0 = this->x;
    this->x = _r14.x;
    float _t1 = this->y;
    this->y = _r14.y;
    {
        this->y = _t1;
        float _r_d1 = (*_d_this).y;
        (*_d_this).y = 0.F;
        (*_d__r14).y += _r_d1;
    }
    {
        this->x = _t0;
        float _r_d0 = (*_d_this).x;
        (*_d_this).x = 0.F;
        (*_d__r14).x += _r_d0;
    }
}
```

vgvassilev: Looks like we need to fix the signature generation somewhere else? When we are creating the function declaration itself?

kchristin22: The thing is that even if we change the pullback signature, the original function's body has a reference to that parameter (f.i. . `this->x = .x;`). We could trace those references but I think the solution right now is more neat.

vgvassilev: It seems like a hack to me to change the AST after they were created. That means that the produced AST is inconsistent and refers to some nodes which were not correctly cloned.

kchristin22: They vars are cloned as supposed to but it is the original function that has a body that shouldn't compile:
```cpp
__host__ __device__ constexpr float2& operator=(float2&&) noexcept {
    this->x = .x;
    this->y = .y;
    return *this;
}
```

vgvassilev: I think I understand now. When we create a pullback for a given function and that function has no parameter names but only types we get into that state. I believe the fix should be at the place where we produce the pullback itself, we can detect if a parameter has no name and give it a name.

kchristin22: You mean change the parameter itself right? Which means that we would use `const_cast`

vgvassilev: I mean at the place where we do `FunctionDecl::Create(<pullback>)`.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4781,6 +4779,10 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     }
 
     for (auto* PVD : m_DiffReq->parameters()) {
+      if (PVD->getNameAsString().empty()) {
+        IdentifierInfo* newName = CreateUniqueIdentifier("_r");
```
vgvassilev, line 4783: The changes should be in newPVD a few lines below. Maybe we can have more useful autogenerated name such as <type>_arg0

kchristin22: As mentioned above, the original function should change as well, not just the pullback, as the original statements are included in the pullback and they are derived accordingly. So if the param is not found in the arg, then they will be derived in the wrong way.

vgvassilev (approved):
Pretty neat! LGTM!

PR #1144, opened 2024-11-19
Add doc page for usage of Clad with CUDA

kchristin22:


docs/userDocs/source/user/UsingCladOnCUDACode.rst
```diff
@@ -0,0 +1,67 @@
...
+The following CUDA features are not yet supported:
+* The use of shared memory within the original function
+* Synchronization primitives like `__syncthreads()` and `cudaDeviceSynchronize()`
+* Other CUDA host functions apart from those listed in the previous section
+
+
+Implementation details
+================================================
+
+For a brief overview of the implementation, please refer to the team's related `blog post <https://compiler-research.org/blogs/gsoc24_christina_koutsou_project_final_blog/>`_.
```
vgvassilev, line 62: Shouldn't it be the inverse? The blog post referring to the documentation?

kchristin22: This page refers to the usage only, as I noticed that this is the pattern followed in the current documentation. I can make a folder where we provide more details regarding the implementation too.

vgvassilev: Ok, another item to open an issue to for in terms of documentation improvement...

kchristin22: Should I open such a separate issue or add an item on the list included in a relevant issue?

vgvassilev: Either way should be fine. Maybe opening a new one is slightly better I think..

vgvassilev (approved):
LGTM!

PR #1147, opened 2024-11-20
Clone printf and fprintf calls

### LINKED ISSUE #1146: Support the use of printf statements in the original function
These statements should be identified and cloned successfully in Clad

kchristin22:
Closes #1146

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1456,6 +1456,7 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       return StmtDiff(Clone(CE));
     }
 
+    // FIXME: Revisit this when variadic functions are supported.
     if (FD->getNameAsString() == "printf" || FD->getNameAsString() == "fprintf")
       return StmtDiff(Clone(CE));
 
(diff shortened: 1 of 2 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1456,6 +1456,9 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
       return StmtDiff(Clone(CE));
     }
 
+    if (FD->getNameAsString() == "printf" || FD->getNameAsString() == "fprintf")
```
vgvassilev, line 1459: Can we add a fixme to revisit after the variadic pullbacks issue is closed?

test/CUDA/GradientKernels.cu
```diff
@@ -520,6 +521,7 @@ void launch_add_kernel_4(int *out, int *in, const int N) {
 
 // CHECK: void launch_add_kernel_4_grad_0_1(int *out, int *in, const int N, int *_d_out, int *_d_in) {
 //CHECK-NEXT:    int _d_N = 0;
+//CHECK-NEXT:    printf("Launching add_kernel_4 for size: %d\n", N);
```
vgvassilev, line 524: Can we add a test where printf is in a loop printing the loop variable value?

lib/Differentiator/ReverseModeVisitor.cpp
kchristin22: Added

test/CUDA/GradientKernels.cu
kchristin22: Done as well

kchristin22:
This PR will be revisited when #1150 is fixed.

PR #1148, opened 2024-11-21
Don't mark non-varied constant params

ovdiiuv:


lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -124,21 +124,54 @@ bool VariedAnalyzer::VisitCallExpr(CallExpr* CE) {
...
 }
 
 bool VariedAnalyzer::VisitDeclStmt(DeclStmt* DS) {
   for (Decl* D : DS->decls()) {
+
+    QualType VDTy = cast<VarDecl>(D)->getType();
+    QualType innermost = VDTy;
+    while (innermost->isPointerType())
+      innermost = innermost->getPointeeType();
+    if (VDTy->isPointerType() && !innermost.isConstQualified()) {
```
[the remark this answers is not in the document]
vgvassilev: This is a good suggestion. You should first check for `isArrayType` I think...

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -153,6 +153,19 @@ struct DiffRequest {
   bool isVaried(const clang::Expr* E) const;
   std::string ComputeDerivativeName() const;
   bool HasIndependentParameter(const clang::ParmVarDecl* PVD) const;
+
+  void addVariedDecl(const clang::VarDecl* init) {
+    this->m_ActivityRunInfo.VariedDecls.insert(init);
```
vgvassilev, line 158:
```suggestion
    m_ActivityRunInfo.VariedDecls.insert(init);
```

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -153,6 +153,19 @@ struct DiffRequest {
...
+  }
+
+  std::set<const clang::VarDecl*>& getVariedDecls() const {
+    return this->m_ActivityRunInfo.VariedDecls;
```
vgvassilev, line 162:
```suggestion
    return m_ActivityRunInfo.VariedDecls;
```

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -153,6 +153,19 @@ struct DiffRequest {
...
+  }
+
+  void setVariedDecls(std::set<const clang::VarDecl*> init) {
```
vgvassilev, line 165: Is this function used?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -625,11 +625,12 @@ namespace clad {
...
-      DVI[i].print(Out);
-      if (i != e - 1)
-        Out << ',';
-    }
+    else
+      for (unsigned i = 0, e = DVI.size(); i < e; i++) {
+        DVI[i].print(Out);
+        if (i != e - 1)
+          Out << ',';
+      }
```
vgvassilev, line 633: Is this change part of this PR? Why is it needed?

ovdiiuv: When differentiating something like:

```cpp
double test1(double x, double y) {
  return 2*y*nested1(y) * 3 * x * nested1(x);
}

int main(){
  auto f_grad = clad::gradient(test1);
return 0;
}
```

we would have non-empty Args and DVI which are same, after calling updateDiffParamsInfo in DiffCollector::VisitCallExpr. Before this PR, DVI wasn't updated properly.

vgvassilev: Can we move that in a separate PR?

ovdiiuv: This is a really temporary workaround and I'm not sure it worth it. I could open an issue on moving updateDiffParamsInfo from ProcessDiffRequest to DiffCollector::VisitCallExpr, which would solve the issue.

vgvassilev: Can you do it instead of opening an issue?

vgvassilev: Let's move that into a separate PR and update test/Misc/TimingsReport.C with a proper test.

test/Analyses/ActivityReverse.cpp
```diff
@@ -317,6 +317,77 @@ double fn9(double x, double const *obs)
 // CHECK-NEXT: }
 
 
+void f10_1(double x, double* t){
```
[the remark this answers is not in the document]
ovdiiuv: f9 does that, basically test that came in Jonas's issue

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -34,11 +35,13 @@
   } m_TbrRunInfo;
 
   mutable struct ActivityRunInfo {
-    std::set<const clang::VarDecl*> ToBeRecorded;
+    std::set<const clang::VarDecl*> VariedDecls;
```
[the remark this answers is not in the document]
vgvassilev: These are not bad suggestions.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1091,7 +1079,21 @@ namespace clad {
     request.CallContext = E;
     request.BaseFunctionName = utils::ComputeEffectiveFnName(request.Function);
 
+    if (m_ParentReq)
+      for (auto decl : m_ParentReq->getVariedDecls())
+        request.addVariedDecl(decl);
```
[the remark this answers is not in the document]
ovdiiuv: added

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1091,7 +1079,24 @@ namespace clad {
...
+    // FIXME: Here we copy all varied declarations down to the pullback, has to
+    // be removed once AA and TBR are completely reworked, with better
+    // branch-merging.
+    if (m_ParentReq)
+      for (auto decl : m_ParentReq->getVariedDecls())
+        request.addVariedDecl(decl);
+
     llvm::SaveAndRestore<const DiffRequest*> Saved(m_ParentReq, &request);
+
+    // Move once #1325 is merged.
```
PetroZarytskyi, line 1091: Is this supposed to be a FIXME?

test/Analyses/ActivityReverse.cpp
```diff
@@ -168,20 +168,15 @@ double f4(double x){
...
-// CHECK-NEXT:     *_d_v += 2 * _d_n;
-// CHECK-NEXT:     *_d_u += 2 * _d_k;
-// CHECK-NEXT: }
+// CHECK: void f4_1_pullback(double v, double u, double _d_y, double *_d_v, double *_d_u) {  
+// CHECK-NEXT:     double k = 2 * u;  
+// CHECK-NEXT:     double _d_n = 0.;  
+// CHECK-NEXT:     double n = 2 * v;  
+// CHECK-NEXT:     _d_n += _d_y * k;  
+// CHECK-NEXT:     *_d_v += 2 * _d_n;  
+// CHECK-NEXT: }
```
PetroZarytskyi, line 177: This looks awesome! As far as I understand, things get simplified because we don't differentiate w.r.t. `u` anymore. Should we account for that in the name and signature of `f4_1`? i.e. `f4_1_pullback_0(double v, double u, double _d_y, double *_d_v)`. Although I don't think it has be in this PR.

PetroZarytskyi (commented):
The PR looks good now. I only left a few comments.

Also, I just thought it would be amazing if we could lift our limitation on non-const pointer parameters if they are not varied.

test/Analyses/ActivityReverse.cpp
vgvassilev: Ping.

ovdiiuv: done

vgvassilev (approved):
LGTM!

PR #1160, opened 2024-12-03
No need to handle recursive calls separately in the reverse mode

PetroZarytskyi:
Currently, we declare pullbacks before generating the body. This means that when ``f`` is differentiated, ``f_pullback`` has already been declared. Therefore, ``HandleNestedDiffRequest`` is capable of finding it.


PetroZarytskyi:
The diff looks like the code was rewritten, although, it just removed the block after ``if (FD == m_DiffReq.Function &&...`` and left the else-block as it was. The diff is confusing because of the indentation.

vgvassilev:
clang-tidy seem to be broken again...

PR #1161, opened 2024-12-03
Use a single point to process non-differentiable functions in the reverse mode

PetroZarytskyi:
Right now, we have a bunch of reasons to consider a function non-differentiable and not produce a pullback. In all cases, we clone the call and return early in different ways, almost all of which have flaws. For example:

Return on line ``1488``: it's not necessary to pass 0 as the adjoint of non-differentiable entities in the reverse mode. Also, cloning arguments is dangerous for multiple reasons (e.g. sometimes we introduce operators, brackets, etc. which are not generated when cloning) and doesn't account for their possible differentiable side effects.

Return on line ``1520``: Again, cloning is dangerous.

Return on line ``1642``: Cloning again. Also, providing the cloned call as the adjoint doesn't make much sense.

Return on line ``1664``: ``dfdx()`` of the call shouldn't be passed to its arguments.

This PR replaces all of these cases to the same place and fixes the mentioned issues.


vgvassilev (approved):
LGTM!

PR #1163, opened 2024-12-05
Initialize adjoints of aggregate types with init lists

PetroZarytskyi:
With aggregate types, we don't have to rely on the default initialization or custom constructor forward passes to zero-initialize its fields:
```
struct aggrStruct {
  double a = 1;
  float b;
  ...
};
```
This option is incorrect since ``_d_x.a`` will be initialized to ``1``:
```
aggrStruct _d_x{};
```
This option could be correct but is just unnecessary:
```
... _t1 = custom_reverse_forward(...);
aggrStruct _d_x = _t1.adjoint; // unnecessary
```
This PR instead implements a simpler solution:
```
aggrStruct _d_x{0., 0.F};
```

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -34,7 +34,9 @@
 #include <clang/AST/OperationKinds.h>
 #include <clang/Sema/Ownership.h>
 
+#include "llvm/ADT/SmallString.h"
 #include "llvm/Support/SaveAndRestore.h"
+#include <llvm/Support/raw_ostream.h>
 
 #include <algorithm>
 #include <numeric>
@@ -4281,12 +4283,22 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
             CE->getConstructor(), primalArgs, reverseForwAdjointArgs,
             /*baseExpr=*/nullptr)) {
       if (RD->isAggregate()) {
-        diag(DiagnosticsEngine::Note, CE->getConstructor()->getBeginLoc(),
-             "No need to provide a custom constructor forward sweep for an "
-             "aggregate type.");
+        SmallString<128> Name_class;
+        llvm::raw_svector_ostream OS_class(Name_class);
+        RD->getNameForDiagnostic(OS_class, m_Context.getPrintingPolicy(),
+                                 /*qualified=*/true);
         diag(DiagnosticsEngine::Warning, CE->getBeginLoc(),
-             "No need to provide a custom constructor forward sweep for an "
-             "aggregate type.");
+             "'%0' is an aggregate type and its constructor does not require a "
+             "user-defined forward sweep function",
+             {OS_class.str()});
+        const FunctionDecl* constr_forw =
+            cast<CallExpr>(customReverseForwFnCall)->getDirectCallee();
+        SmallString<128> Name_forw;
+        llvm::raw_svector_ostream OS_forw(Name_forw);
+        constr_forw->getNameForDiagnostic(
+            OS_forw, m_Context.getPrintingPolicy(), /*qualified=*/true);
+        diag(DiagnosticsEngine::Note, constr_forw->getBeginLoc(),
+             "'%0' is defined here", {OS_forw.str()});
       }
       Expr* callRes = StoreAndRef(customReverseForwFnCall);
       Expr* val =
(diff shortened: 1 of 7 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1309,7 +1309,16 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
         clonedExprs[i] = Visit(ILE->getInit(i), member_acess).getExpr();
       }
       Expr* clonedILE = m_Sema.ActOnInitList(noLoc, clonedExprs, noLoc).get();
-      return StmtDiff(clonedILE);
+
+      const CXXRecordDecl* RD = ILEType->getAsCXXRecordDecl();
+      Expr* adjointInit = nullptr;
+      if (RD && RD->isAggregate()) {
+        llvm::SmallVector<Expr*, 4> adjParams;
+        for (const FieldDecl* FD : RD->fields())
```
vgvassilev, line 1317: Can we write a test case with a private data member?

PetroZarytskyi: I believe aggregates cannot have private data members

lib/Differentiator/VisitorBase.cpp
```diff
@@ -412,11 +412,20 @@ namespace clad {
...
-          ConstantFolder::synthesizeLiteral(T, m_Context, /*val=*/0);
-      return Zero.get();
+    if ((T->isScalarType() || T->isPointerType()) && !T->isReferenceType())
+      return ConstantFolder::synthesizeLiteral(T, m_Context, /*val=*/0);
+
+    if (isa<ConstantArrayType>(T)) {
+      Expr* zero = ConstantFolder::synthesizeLiteral(T, m_Context, /*val=*/0);
+      return m_Sema.ActOnInitList(noLoc, {zero}, noLoc).get();
     }
+    if (const auto* RD = T->getAsCXXRecordDecl())
```
vgvassilev, line 422: Can you make codecov happy?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4286,6 +4280,10 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
     if (Expr* customReverseForwFnCall = BuildCallToCustomForwPassFn(
             CE->getConstructor(), primalArgs, reverseForwAdjointArgs,
             /*baseExpr=*/nullptr)) {
+      if (RD->isAggregate())
+        diag(DiagnosticsEngine::Warning, CE->getBeginLoc(),
```
vgvassilev, line 4284: Is here the place where we found the custom forward reverse function and we diagnose we do not need it? If not we should move the check there, and point to the declaration itself with a `DiagnosticsEngine::Note`(CE->getDecl()->getBeginLoc()). And of course we should add a test for it...

PetroZarytskyi: I kept the diagnostics in the same place but now the location comes from the declaration. Is it better now?

vgvassilev: We need both a warning on the call site and a note on our he definition. We also need a test.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4281,12 +4281,25 @@ Expr* getArraySizeExpr(const ArrayType* AT, ASTContext& context,
...
-             "No need to provide a custom constructor forward sweep for an "
-             "aggregate type.");
+        // set up printing policy
+        clang::LangOptions LangOpts;
+        LangOpts.CPlusPlus = true;
+        clang::PrintingPolicy Policy(LangOpts);
+        Policy.Bool = true;
+        SmallString<128> Name_class;
+        llvm::raw_svector_ostream OS_class(Name_class);
+        RD->getNameForDiagnostic(OS_class, Policy, /*qualified=*/true);
```
vgvassilev, line 4291: You can get the policy from the ASTContext.

test/Gradient/UserDefinedTypes.C
```diff
@@ -467,6 +467,53 @@ void fn13(double *x, double *y, int size)
...
+::clad::ValueAndAdjoint<::std::array<double, N>, ::std::array<double, N>> // expected-note {{'clad::custom_derivatives::class_functions::constructor_reverse_forw<2UL>' is defined here}}
+constructor_reverse_forw(::clad::ConstructorReverseForwTag<::std::array<double, N>>) {
+  ::std::array<double, N> a;
+  ::std::array<double, N> d_a;
+  return {a, d_a};
+}
+}}}
+
+double fn15(double x, double y) {
+  std::array<double, 2> arr; // expected-warning {{'std::array<double, 2>' is an aggregate type and its constructor does not require a user-defined forward sweep function}}
```
vgvassilev, line 501: This does not need to be a `std::array` because we will remove its reverse function. It can be something as simple as an empty class...

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: Fixed

vgvassilev (approved):
Lgtm!

PR #1164, opened 2024-12-06
Don't consider arrays as a special case in DifferentiateVarDecl

PetroZarytskyi:
This PR improves the support for array types in ``getZeroInit`` and simplifies ``DifferentiateVarDecl`` using that change.

Note: most of the diff comes from the change in indentation


vgvassilev (approved):
LGTM!

PR #1171, opened 2024-12-14
Respect shadow declarations when writing propagators.

vgvassilev:
In cases where the public declaration is introduced with using declaration pointing to an internal namespace with the implementation details, we should put the propagator function in the namespace of the public function and not the implementation. That would allow users to position their pullbacks in the same namespace structure as the used functions.

Depends on #1172


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -247,9 +248,29 @@ static void registerDerivative(FunctionDecl* derivedFD, Sema& semaRef) {
...
+    const DeclContext* originalFnDC = nullptr;
+
+    // FIXME: callSite must not be null but it comes when we try to build
+    // a numerical diff call. We should merge both paths and remove the
+    // special branches being taken for propagators and numerical diff.
+    if (callSite) {
+      // Check if the callSite is not associated with a shadow declaration.
+      if (const auto* ME = dyn_cast<CXXMemberCallExpr>(callSite)) {
+        originalFnDC = ME->getMethodDecl()->getParent();
+      } else if (const auto* CE = dyn_cast<CallExpr>(callSite)) {
```
PetroZarytskyi, line 263: ``CXXMemberCallExpr`` is a ``CallExpr ``. If we remove this if-block, will the next one handle member calls correctly?

vgvassilev: We call getMethodDecl there which is not part of CallExpr. I had to add this explicitly to cover some cases iirc.

PetroZarytskyi (approved):

PetroZarytskyi:
LGTM!

Issue #1098, opened 2024-09-17
Incorrect gradient differentiating parenthesis


ovdiiuv:
Differentiating 
```cpp
double f5(double x){
  return -1 * (-x);
}
```
gives us

```cpp
void f5_grad(double x, double *_d_x) {
    *_d_x += --1 * 1;
}
```

PR #1180, opened 2024-12-20
Fix for Parenthesis issue in gradients

toshit3q34:
This resolves the issue with two or more consecutive negative signs while dumping the code.
While copying the code to m_Code, changes include iterating the loop and checking when there are two or more consecutive negative signs and resolves them to be either '-' or ' ' depending on the parity.

Eg.
-(-(-1))*(x) 
Before :  _d_x += - - -1 * 1;
After : _d_x += -1 * 1;

The differentiate remains intact.
Fixes : #1098

### DIFF
--- a/lib/Differentiator/VisitorBase.cpp
+++ b/lib/Differentiator/VisitorBase.cpp
@@ -27,6 +27,7 @@
 #include "clang/Sema/Template.h"
 
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/Casting.h"
 
 #include <algorithm>
 #include <numeric>
@@ -400,7 +401,24 @@ namespace clad {
     // Debug clang requires the location to be valid
     if (!OpLoc.isValid())
       OpLoc = utils::GetValidSLoc(m_Sema);
+    // Call function for UnaryMinus
+    if (OpCode == UO_Minus)
+      return ResolveUnaryMinus(E->IgnoreCasts(), OpLoc);
     return m_Sema.BuildUnaryOp(nullptr, OpLoc, OpCode, E).get();
+  }
+  Expr* VisitorBase::ResolveUnaryMinus(Expr* E, SourceLocation OpLoc) {
+    if (auto* UO = llvm::dyn_cast<clang::UnaryOperator>(E)) {
+      if (UO->getOpcode() == clang::UO_Minus)
+        return (UO->getSubExpr())->IgnoreParens();
+    }
+    Expr* E_LHS = E;
+    while (auto* BO = llvm::dyn_cast<BinaryOperator>(E_LHS))
+      E_LHS = BO->getLHS();
+    if (auto* UO = llvm::dyn_cast<clang::UnaryOperator>(E_LHS->IgnoreCasts())) {
+      if (UO->getOpcode() == clang::UO_Minus)
+        E = m_Sema.ActOnParenExpr(E->getBeginLoc(), E->getEndLoc(), E).get();
+    }
+    return m_Sema.BuildUnaryOp(nullptr, OpLoc, clang::UO_Minus, E).get();
   }
 
   Expr* VisitorBase::BuildOp(clang::BinaryOperatorKind OpCode, Expr* L, Expr* R,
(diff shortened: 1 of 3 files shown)


include/clad/Differentiator/Differentiator.h
```diff
@@ -41,6 +41,37 @@ inline CUDA_HOST_DEVICE unsigned int GetLength(const char* code) {
   return count;
 }
 
+// Fixes consecutive negative signs
+// (-1)*(-x) gives *_d_x += 1 * 1; instead of
+// *_d_x += --1 * 1;
+CUDA_HOST_DEVICE void ResolveParenthesis(char* temp, const char* code) {
```
vgvassilev, line 47: We do not do textual parsing and transformations. The solution must be on AST level in one of our Visitor classes. You should check our documentation for more details.

include/clad/Differentiator/Differentiator.h
```diff
@@ -59,615 +90,598 @@ CUDA_HOST_DEVICE T push(tape<T>& to, ArgsT... val) {
...
 
-  /// Add value to the end of the tape, return the same value.
-  /// A specialization for clad::array_ref types to use in reverse mode.
-  template <typename T, typename U>
-  CUDA_HOST_DEVICE clad::array_ref<T> push(tape<clad::array_ref<T>>& to,
-                                           U val) {
-    to.emplace_back(val);
-    return val;
-  }
+/// Add value to the end of the tape, return the same value.
```
vgvassilev, line 93: Running clang-format on the entire file is not a good idea. We should run it only on the changes with `git-clang-format HEAD~`.

vgvassilev (changes requested):

toshit3q34:
@vgvassilev Thanks! I went through the documentation and have fixed the issue in the `BuildParens` function of `VisitorBase` class. The tests compile without a problem but CHECK-NEXT shows error in 5 tests due to mistmaching body statements.
Example : 
Expected : `-_d_x * x + -x * _d_x`
After the changes : `(-_d_x) * x + (-x) * _d_x`

What should be done to deal with such instances or those 5 tests need to be changed ?

toshit3q34:
@vgvassilev Codecov test is failing because there is no test to check those lines.
Example : `f(x) = (-1) * (-(-x))`
So should I add a file for such tests ?

lib/Differentiator/VisitorBase.cpp
```diff
@@ -400,8 +400,31 @@ namespace clad {
     // Debug clang requires the location to be valid
     if (!OpLoc.isValid())
       OpLoc = utils::GetValidSLoc(m_Sema);
+    // Call function for UnaryMinus
+    if (OpCode == UO_Minus)
+      return RemoveFirstUnaryMinus(E->IgnoreCasts(), OpLoc);
     return m_Sema.BuildUnaryOp(nullptr, OpLoc, OpCode, E).get();
   }
+  Expr* VisitorBase::RemoveFirstUnaryMinus(Expr* E, SourceLocation OpLoc) {
```
vgvassilev, line 408: Rather than removing the minus we should figure out where we synthesize the minus and we need parens.

toshit3q34: I tried to do that at first and was able to do it as well. However for some cases like :
`f(x) = (-(-1)) * (-(-(x)))` if we try to parenthesize, the output becomes `(-(-(-(-1))*1))` which looks quite cluttered and redundant. Similar thing also happens when we try for higher powers like 
`f(x) = (-1) * (-x) * (-x) * (-x)`
Would you recommend continuing to parenthesize in such cases, or would it be acceptable to simplify the representation for clarity?

vgvassilev: It's better to add parenthesis only at the places where it is necessary.

toshit3q34: Okay sure. Can you please specify the expected output of `f_grad()` for `f(x) = (-(-1)) * (-(-(x)))` so that I have a basis to work with and can modify accordingly. Thanks.

toshit3q34:
@vgvassilev I hope you have reviewed the code. I have done the following two things :
1. Reduce unary minus when done on the same operand to avoid clutter. example : `---2 becomes -2`
2. Add parenthesis when unary minus is added in front of compound statement as asked before. 
example : `--1*x becomes -(-1*x)`

The codecov tests fail because there are no tests which check either of the above conditions. 
Is there anything else that I need to change ?

vgvassilev:
Can you add a few tests demonstrating this pull request works?

toshit3q34:
@vgvassilev I have added a test file containing two tests which covers the above stated scenarios. You may review it now.

parth-07 (approved):
Looks good.

vgvassilev (approved):
LGTM!

Issue #1199, opened 2024-12-30
clad.dylib gives a dlopen error in MacOS Sonoma


RS2007:
### Machine specifications:
OS: macOS 14.5 23F79 arm64 (Sonoma) 
Running on a Mac M2 Chip

Installation was done following the MacOS installation instructions in the README

### Commands used to run the differentiator example from README:

```bash
Hedwig :: /tmp/clad » clang++ -std=c++11 -I./inst/include -fplugin=./inst/lib/clad.dylib test.cpp
error: unable to load plugin './inst/lib/clad.dylib': 'dlopen(./inst/lib/clad.dylib, 0x0009): symbol not found in flat namespace '__ZN4llvm9DebugFlagE''
Hedwig :: /tmp/clad » cat test.cpp
#include <iostream>
#include "clad/Differentiator/Differentiator.h"

double f(double x, double y) { return x * y; }

int main() {
  // Call clad to generate the derivative of f wrt x.
  auto f_dx = clad::differentiate(f, "x");
  // Execute the generated derivative function.
  std::cout << f_dx.execute(/*x=*/3, /*y=*/4) << std::endl;
  // Dump the generated derivative code to standard output.
  f_dx.dump();
}
```

RS2007:
Closing this, since the issue was that clang-15 was being used instead of clang-12 (bundled with llvm-12).

RS2007:
Should this probably be mentioned in the readme? Since most users would have their default clang installation elsewhere and this issue might be mitigated if they are warned beforehand

vgvassilev:
Yes, probably worth mentioning it more explicitly. Do you think you can open a PR?

vgvassilev:
The fix was a clarification in the documentation. Thank you @RS2007 for providing the bug report and its fix too :)

PR #1200, opened 2024-12-30
Update README with clang version note for macos

RS2007:
Adding instructions in the README for mac users, to not use their systemwide clang installation while using clad and instead use the version of clang inside homebrew llvm-12.

Using the systemwide clang(15.0.0) installation leads to the following issue:  #1199

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,7 +273,7 @@ make && make install
 make check-clad
 ```
 
-> **NOTE**: If you are using clad as a clang plugin after building it from source, use the version of clang from your llvm@12 installation directory(/opt/homebrew/opt/llvm@12/bin/clang). Using the systemwide clang binary may not work.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -272,7 +272,11 @@ cmake ../clad -DLLVM_DIR=/opt/homebrew/opt/llvm@12/lib/cmake/llvm -DClang_DIR=/o
 make && make install
 make check-clad
 ```
-###  Developer Environment - Build LLVM, Clang and Clad from source:
+
+> **NOTE**: If you are using clad as a clang plugin after building it from source, use the version of clang from your llvm@12 installation directory(/opt/homebrew/opt/llvm@12/bin/clang). Using the systemwide clang binary may not work.
```
vgvassilev, line 276:
```suggestion
> **NOTE**: If you are using clad as a clang plugin after building it from source, please make sure that you uses the same compiler version you built clad against. Apple distributed clang does not work because Apple has disabled clang plugins.
```

PR #1207, opened 2025-01-03
Don't pass derived args to forw pass unless they are references

PetroZarytskyi:
``_reverse_forward`` functions only contain the forward pass, which only affects the derivatives on references/pointers. For other types, there is no point in passing the adjoint. Moreover, doing so is often incorrect because derived arguments are generated for the reverse pass, e.g.
```
// forward pass
operator_subscript_reverse_forw(&vec, 0, &_d_vec, _r0); // `_r0` is declared later
...
// reverse pass
size_type _r0 = 0UL;
operator_subscript_pullback(&_t2, 0, 1, &_d_vec, &_r0);
```
In this example, replacing the first occurrence of _r0 with 0 will still be correct.

### DIFF
--- a/test/Gradient/STLCustomDerivatives.C
+++ b/test/Gradient/STLCustomDerivatives.C
@@ -221,9 +221,9 @@ int main() {
 // CHECK-NEXT:     std::vector<double> _t1 = vec;
 // CHECK-NEXT:     {{.*}}class_functions::push_back_reverse_forw(&vec, v, &_d_vec, *_d_v);
 // CHECK-NEXT:     std::vector<double> _t2 = vec;
-// CHECK-NEXT:     {{.*}}ValueAndAdjoint<double &, double &> _t3 = {{.*}}class_functions::operator_subscript_reverse_forw(&vec, 0, &_d_vec, _r0);
+// CHECK-NEXT:     {{.*}}ValueAndAdjoint<double &, double &> _t3 = {{.*}}class_functions::operator_subscript_reverse_forw(&vec, 0, &_d_vec, {{0U|0UL|0}});
 // CHECK-NEXT:     std::vector<double> _t4 = vec;
-// CHECK-NEXT:     {{.*}}ValueAndAdjoint<double &, double &> _t5 = {{.*}}class_functions::operator_subscript_reverse_forw(&vec, 1, &_d_vec, _r1);
+// CHECK-NEXT:     {{.*}}ValueAndAdjoint<double &, double &> _t5 = {{.*}}class_functions::operator_subscript_reverse_forw(&vec, 1, &_d_vec, {{0U|0UL|0}});
 // CHECK-NEXT:     {
 // CHECK-NEXT:         size_type _r0 = {{0U|0UL}};
 // CHECK-NEXT:         {{.*}}class_functions::operator_subscript_pullback(&_t2, 0, 1, &_d_vec, &_r0);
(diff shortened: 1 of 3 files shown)


vgvassilev (commented):
Any chance for a tests? Also can you make the commit message match the PR description?

test/Gradient/STLCustomDerivatives.C
```diff
@@ -319,23 +319,23 @@ int main() {
...
 // CHECK-NEXT:     double _t10 = res;
 // CHECK-NEXT:     std::vector<double> _t11 = vec;
-// CHECK-NEXT:     {{.*}}ValueAndAdjoint<double &, double &> _t12 = {{.*}}class_functions::operator_subscript_reverse_forw(&vec, 0, &_d_vec, _r4);
+// CHECK-NEXT:     {{.*}}ValueAndAdjoint<double &, double &> _t12 = {{.*}}class_functions::operator_subscript_reverse_forw(&vec, 0, &_d_vec, {{0U|0UL|0}});
 // CHECK-NEXT:     std::vector<double> _t13 = vec;
-// CHECK-NEXT:     {{.*}}ValueAndAdjoint<double &, double &> _t14 = {{.*}}class_functions::operator_subscript_reverse_forw(&vec, 1, &_d_vec, _r5);
+// CHECK-NEXT:     {{.*}}ValueAndAdjoint<double &, double &> _t14 = {{.*}}class_functions::operator_subscript_reverse_forw(&vec, 1, &_d_vec, {{0U|0UL|0}});
 // CHECK-NEXT:     std::vector<double> _t15 = vec;
-// CHECK-NEXT:     {{.*}}ValueAndAdjoint<double &, double &> _t16 = {{.*}}class_functions::operator_subscript_reverse_forw(&vec, 2, &_d_vec, _r6);
+// CHECK-NEXT:     {{.*}}ValueAndAdjoint<double &, double &> _t16 = {{.*}}class_functions::operator_subscript_reverse_forw(&vec, 2, &_d_vec, {{0U|0UL|0}});
```
parth-07, line 326: Writing FileCheck `CHECK`s that satisfy all compiler/runtime combinations seems can get very time-consuming. @vgvassilev @PetroZarytskyi What do you think about testing the FileCheck `CHECK`s for just one compiler/runtime?

parth-07 (approved):

PetroZarytskyi:
> Any chance for a tests?

@vgvassilev The coverage decreased because we no longer visit ``InitListExpr`` without ``dfdx`` (which is, ideally, never supposed to happen). And because of that, ``InitListExpr`` is never cloned either. This is not a bad thing itself since cloning is dangerous and should only be applied to simple exprs like literals or decl refs.

vgvassilev:
> > Any chance for a tests?
> 
> @vgvassilev The coverage decreased because we no longer visit `InitListExpr` without `dfdx` (which is, ideally, never supposed to happen). And because of that, `InitListExpr` is never cloned either. This is not a bad thing itself since cloning is dangerous and should only be applied to simple exprs like literals or decl refs.

Understood I can merge.

test/Gradient/STLCustomDerivatives.C
vgvassilev: Yes, alternatively we can just puth `auto` for the `ValueAndAdjoint`. I do not think we will need a lot of regex magic after we landed #1171.

PR #1211, opened 2025-01-06
Improve derivative registration.

vgvassilev:
Once the derivative's FunctionDecl is created it needs to be registered properly in the AST. Before this patch `registerDerivative` ignored many checks which make sure the produced code is valid C++. This patch enforces more rigorous checks for some declaration kinds such as standalone functions and some cases of class functions. Lambdas will be handled in subsequent patches.

The challenge is to produce valid C++ code when differentiating classes. Consider:
```cpp
class A1 {
  double f(double x) { return x * x; }
  double f_grad(double x, double *d_x); // forward declaration of the gradient.
};
```

In such cases clad will now produce a correct out-of-line definition of `f_grad`: `double A1::f_grad(double x, double *d_x) { ... }`.

In cases where the gradient function is not declared as part of the class signature (external libraries such as STL), we continue to handle them as before. That will be dealt with in a subsequent patches.

This also fixes assertions in debug mode when there are virtual functions.

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -8,8 +8,12 @@
 
 #include "JacobianModeVisitor.h"
 #include "clang/AST/ASTContext.h"
+#include "clang/AST/Attr.h"
+#include "clang/AST/Decl.h"
 #include "clang/AST/ExprCXX.h"
 #include "clang/AST/TemplateBase.h"
+#include "clang/Basic/LLVM.h" // isa, dyn_cast
+#include "clang/Basic/Specifiers.h"
 #include "clang/Sema/Lookup.h"
 #include "clang/Sema/Overload.h"
 #include "clang/Sema/Scope.h"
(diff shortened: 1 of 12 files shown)


lib/Differentiator/VisitorBase.cpp
```diff
@@ -666,8 +666,7 @@ namespace clad {
                      .get();
       isArrow = false;
     }
-    NestedNameSpecifierLoc NNS(FD->getQualifier(),
-                               /*Data=*/nullptr);
+    NestedNameSpecifierLoc NNS /* = FD->getQualifierLoc()*/;
```
PetroZarytskyi, line 669: Does this have to be commented out?

PetroZarytskyi:
Looks good to me!

lib/Differentiator/VisitorBase.cpp
vgvassilev: This would print out the class name and we will have `this->SomeClass::someFn`. We should probably re-enable it for static functions. I plan to do that outside of this PR.

test/Misc/RunDemos.C
```diff
@@ -4,8 +4,6 @@
 // RUN: %cladclang %S/../../demos/RosenbrockFunction.cpp -I%S/../../include 2>&1
 // RUN: %cladclang %S/../../demos/ComputerGraphics/smallpt/SmallPT.cpp -I%S/../../include 2>&1
 
-// XFAIL: asserts
```
parth-07, line 7: This is nice!

parth-07 (approved):
Looks good!

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -120,20 +110,29 @@ static void registerDerivative(FunctionDecl* derivedFD, Sema& semaRef) {
...
+          continue;
         if (!hasAttribute(returnedFD, Attr->getKind()))
           returnedFD->addAttr(Attr->clone(m_Context));
+      }
+    }
+
+    // Generated member function should be called outside of class definitions
+    // even if their original function had different access specifier.
+    if (isa<CXXMethodDecl>(returnedFD))
+      returnedFD->setAccess(AS_public);
```
parth-07, line 135: Will this generate non-private derivative functions for private member functions?

vgvassilev: Correct, we have a that now breaks because we try to call a pushforward that was declared private...

lib/Differentiator/VisitorBase.cpp
vgvassilev: Ha. This is valid C++:
```cpp
class C {
    int f() {return 1;};
    int g() { return this->C::f(); }
};

int main() {}
```

Every day I learn something new...

parth-07: C++ is an _interesting_ language to say the least...

`this->C::f` notation was surprising. I think it is perhaps useful when the function `f` is overridden by a child class and we want to refer to the `foo` from the parent class.

PR #1212, opened 2025-01-07
Support std::initializer_list parameters in the reverse mode

### LINKED ISSUE #1082: Clad crashes when trying to initialise a class type declaration in the reverse mode
reproducer:
```c++
#include "clad/Differentiator/STLBuiltins.h"
#include <iostream>

double fn(double x, double y) {
    std::vector<double> v{1, 2, 3}; // this doesn't work
    return v[0];
}

    double dx, dy;
    auto df = clad::gradient(fn, "x, y");
    std::cout << fn(3, 4) << '\n';
    dx = 0; dy = 0;
    df.execute(3, 4, &dx, &dy);
}
```
same thing is happening with other types

PetroZarytskyi:
Previously, we replaced ``std::initializer_list`` variables with ``clad::array`` in the reverse mode so that they become modifiable. This PR moves the logic for type replacement from ``RMV::DifferentiateVarDecl`` to a``RMV::CloneType`` to use it to handle constructors with ``std::initializer_list`` parameters.

Fixes #1082.

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -200,6 +200,14 @@ namespace clad {
     /// \returns True if the statement was added to the block, false otherwise.
     bool AddToGlobalBlock(clang::Stmt* S) { return addToBlock(S, m_Globals); }
 
+    /// Updates size references in VariableArrayType and replaces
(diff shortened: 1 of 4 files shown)


vgvassilev (approved):
Lgtm!

PR #1214, opened 2025-01-12
Improve formatting of DiffRequest::dump and make it available in gdb.

vgvassilev:


vgvassilev:
This is mostly a gdb convenience function, not to be directly tested. I'd ignore codecov...

PR #1221, opened 2025-01-18
Add Ubuntu arm runners to ci

mcbarton:
This PR will add both the Ubuntu 22.04 and Ubuntu 24.04 arm runners to Clads ci.


.github/workflows/ci.yml
```diff
@@ -468,6 +468,196 @@ jobs:
             compiler: clang-16
             clang-runtime: '18'
 
+          - name: ubu22-arm-gcc10-runtime14
```
vgvassilev, line 471: Probably 2-3 bots for the newest versions of llvm should be sufficient.

vgvassilev (approved):
LGTM!

PR #1222, opened 2025-01-19
Add primitive support for `std::unique_ptr<T>`

parth-07:
This commit adds primitive support for std::unique_ptr<T>. Important changes introduced:

- Custom derivatives are added for std::unique_ptr<T> constructor, and operator* member function.
- `constructor_reverse_forw` has been modified to use move constructor in-case the object is not copyable.
- Variable-restore is disabled for non-copyable types.

### DIFF
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -729,5 +729,18 @@ namespace clad {
       return QT->isReferenceType() &&
              !QT.getNonReferenceType().isConstQualified();
     }
+
+    bool isCopyable(const clang::CXXRecordDecl* RD) {
+      if (RD->defaultedCopyConstructorIsDeleted())
+        return false;
+      if (RD->hasUserDeclaredCopyConstructor()) {
+        std::string qualifiedName = RD->getQualifiedNameAsString();
+        // FIXME: I don't know why Clang things that unique_ptr has
+        // user-declared copy constructor.
+        if (qualifiedName == "std::unique_ptr")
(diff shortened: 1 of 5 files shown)


lib/Differentiator/CladUtils.cpp
```diff
@@ -729,5 +729,18 @@ namespace clad {
...
     }
+
+    bool isCopyable(const clang::CXXRecordDecl* RD) {
+      if (RD->defaultedCopyConstructorIsDeleted())
+        return false;
+      if (RD->hasUserDeclaredCopyConstructor()) {
+        std::string qualifiedName = RD->getQualifiedNameAsString();
+        // FIXME: I don't know why Clang things that unique_ptr has
+        // user-declared copy constructor.
+        if (qualifiedName == "std::unique_ptr")
```
vgvassilev, line 740: Can we use this generic check in finding out if something is a smart pointer or not: https://github.com/compiler-research/CppInterOp/blob/c781d0e7fb27199c53557b3d490e530c72d0930e/lib/Interpreter/CppInterOp.cpp#L310

parth-07: I think we can use this. However, I was hoping that in the (near) future, we can fix this function such that we do not require special condition at all.

Using isSmartPointer check may not be valid for all cases because `shared_ptr` is copyable but `unique_ptr` is not. This function should return false for any type that is not-copyable.

vgvassilev (approved):
Lgtm!

PR #1223, opened 2025-01-21
Support differentiating functions that return CXXConstructExpr

### LINKED ISSUE #1216: Correctly generate pullbacks with object return types in the reverse mode
```cpp
struct A {
  double x;
  A() = default;
  A(double px) : x(px) {}
  A(const A& b) : x(b.x) {}
};

namespace clad {
namespace class_functions {
  void constructor_pullback(A* a, double x, A* d_a, double* d_x) {
      *d_x += d_a->x;
  }
  void constructor_pullback(A* a, const A& b, A* d_a, A* d_b) {
      d_b->x += d_a->x;
  }
}}}

A operator+(const A& a, const A& b) {
  A res;
  res.x = a.x + b.x;
  return res;
}

  A t1{1};
  A t2{x};
  A sum = t1 + t2;
  return sum.x;
}


    auto df = clad::gradient(f);
    double dx, dy;
    dx = 0; dy = 0;
    df.execute(3, 4, &dx, &dy);
}
```
Expected output:
```1 0```
Output:
```segmentation fault```

Note that the bug happens because of the way ``operator+`` is differentiated:
```
void operator_plus_pullback(const A &a, const A &b, A _d_y, A *_d_a, A *_d_b) {
    A res;
    A _d_res({});
    clad::zero_init(_d_res);
    double _t0 = res.x;
    res.x = a.x + b.x;
    ...::constructor_pullback(nullptr, res, nullptr, &_d_res); // <--- This line corresponds to the return-stmt
    {                            // Instead of the last nullptr we are supposed to have _d_y
        res.x = _t0;
        double _r_d0 = _d_res.x;
        _d_res.x = 0.;
        (*_d_a).x += _r_d0;
        (*_d_b).x += _r_d0;
    }
}
```

### PR DESCRIPTION
**Pull request by PetroZarytskyi**
The goal of the changes is to support ``ReturnStmt`` with ``CXXConstructExpr``, e.g.
```
SimpleFunctions1 operator*(const SimpleFunctions1& rhs) {
  return SimpleFunctions1(rhs.x, this->y);
}
```
Currently, we have 2 problems in the way:
1) When forming the ``constructor_pullback`` call, we need to initialize a parameter of type ``SimpleFunctions1`` with a constructor. This is possible either with ``SimpleFunctions1(x, y)`` or with an ``InitListExpr`` ``{x, y}``. However, instead of doing the former, we generate a ``ParenListExpr`` ``(x, y)``. Clang can build the constructor call itself when dealing with decl inits but not here.
2) The pullback argument is not propagated to the constructor pullback; therefore, we get a zero derivative.

This PR solves 2) completely and fixes 1) for usage in ``ReturnStmt`` and as a parameter.
Originally, the PR was opened to support all tests from #984. Therefore, all of them are added.
Fixes #1216.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -4151,6 +4151,10 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     // responsible for updating these args.
     Expr* thisE = getZeroInit(recordPointerType);
     Expr* dThisE = getZeroInit(recordPointerType);
+    if (!m_TrackConstructorPullbackInfo && dfdx() &&
+        m_DiffReq.Mode == DiffMode::experimental_pullback)
+      dThisE = BuildOp(UnaryOperatorKind::UO_AddrOf, dfdx(),
+                       m_DiffReq->getLocation());
 
     pullbackArgs.push_back(thisE);
     pullbackArgs.append(primalArgs.begin(), primalArgs.end());
@@ -4170,7 +4174,6 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       if (m_TrackConstructorPullbackInfo) {
         setConstructorPullbackCallInfo(llvm::cast<CallExpr>(customPullbackCall),
                                        primalArgs.size() + 1);
-        m_TrackConstructorPullbackInfo = false;
       }
     }
     // FIXME: If no compatible custom constructor pullback is found then try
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/Differentiator/ReverseModeVisitor.cpp [hunk 1]
```diff
@@ -4151,6 +4151,10 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     // responsible for updating these args.
     Expr* thisE = getZeroInit(recordPointerType);
     Expr* dThisE = getZeroInit(recordPointerType);
+    if (!m_TrackConstructorPullbackInfo && dfdx() &&
+        m_DiffReq.Mode == DiffMode::experimental_pullback)
```
PetroZarytskyi, line 4155: This check is only necessary to prevent a crash in the case described in #1224. If that PR is merged, the line can be removed.

vgvassilev (approved):
LGTM!

PR #1224, opened 2025-01-21
Don't accept functions of object return types in clad::gradient

PetroZarytskyi:
``clad::gradient`` cannot differentiate functions with multiple outputs. This is a limitation of the reverse mode AD which can only be overcome by vectorization of adjoints (essentially, having a separate adjoint element for every output). Similarly, ``clad::gradient`` cannot differentiate functions of object types since, most of the time, classes have more than just one real field. 
However, we have one test for this case:
```
MyStruct fn12(MyStruct s) {
  s = {2 * s.a, 2 * s.b + 2};
  return s;
}
```
The derivative is called as follows:
```
MyStruct s = {1.0, 2.0}, d_s = {1.0, 1.0};
auto fn12_test = clad::gradient(fn12);
fn12_test.execute(s, &d_s);
print(d_s); // CHECK-EXEC: {2.00, 2.00}
```
This example is misleading because the derivative is actually given to the gradient as an adjoint parameter, this has nothing to do with the return value of ``fn12``.
The generated code of ``fn12_grad`` will be identical if we remove the return value altogether, i.e.
```
void fn12(MyStruct s) {
  s = {2 * s.a, 2 * s.b + 2};
}
```
This PR adds ``warning`` and ``note`` in such cases and adds a test with the void version of ``fn12`` to prove the results are identical.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -215,14 +215,25 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
       const_cast<DiffRequest&>(m_DiffReq).Mode = DiffMode::reverse;
 
+    QualType returnTy = m_DiffReq->getReturnType();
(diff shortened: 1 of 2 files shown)


vgvassilev (approved):
LGTM!

PR #1225, opened 2025-01-22
Support VisitCXXFunctionalCastExpr in the reverse mode

### LINKED ISSUE #1215: Support constructor calls passed as arguments in the reverse mode
Currently, we only differentiate constructor calls if they are attached to a variable declaration. For example, we don't support constructor calls being used as call arguments to a function/operator. e.g.

```cpp
struct A {
  double x;
  A() = default;
  A(double px) : x(px) {}
};

namespace clad {
namespace class_functions {
  void constructor_pullback(A* a, double x, A* d_a, double* d_x) {
      *d_x += d_a->x;
  }
}}}

double operator+(const double& val, const A& a) {
  return a.x + val;
}

  return 2 + A(x);     // A(x) is ignored
  
  // However, if instead we use the next lines, everything works as intented
  // A b(x);
  // return 2 + b;
}


    auto df = clad::gradient(f);
    double dx, dy;
    dx = 0; dy = 0;
    df.execute(3, 4, &dx, &dy);
}
```
Expected output:
```1 0```
Output:
```0 0```

PetroZarytskyi:
Fixes #1215


vgvassilev (approved):
LGTM!

PR #1226, opened 2025-01-22
Remove VisitWithExplicitNoDfDx and replace it with plain Visit.

PetroZarytskyi:
In the reverse mode, ``VisitWithExplicitNoDfDx(const Stmt* S)`` visits ``S`` while pushing ``nullptr`` to the ``dfdx()`` stack. This is exactly the behaviour we get when calling ``Visit(const Stmt* S, Expr* dfdS = nullptr)`` with a single parameter. In other words, ``VisitWithExplicitNoDfDx(S)`` is the same as ``Visit(S)``. This PR removes this function.

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -128,28 +128,6 @@ namespace clad {
       return result;
     }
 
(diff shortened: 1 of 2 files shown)


vgvassilev (approved):
Lgtm!

PR #1228, opened 2025-01-22
Don't return zero initialization as the derivative of InitListExpr

PetroZarytskyi:
A zero-initializing ``InitListExpr`` is not supposed to be returned as the derivative in the reverse mode, but rather by ``getZeroInit()``.


PetroZarytskyi:
I didn't mention this in the PR/commit messages but I left this bit by mistake in #1179.

vgvassilev (approved):
Lgtm!

PR #1229, opened 2025-01-22
Don't use original stmts in the generated code

PetroZarytskyi:
It's not correct to use the original expressions ``L`` and ``R`` in the generated code. Also, it doesn't make sense why we exclusively differentiate ``DeclRefExpr``.


vgvassilev (approved):
Lgtm!

PR #1230, opened 2025-01-22
Remove m_DerivativeForForwSweep from StmtDiff

PetroZarytskyi:
Depends on #1228.
Currently, ``m_DerivativeForForwSweep`` is used to store adjoint to the reference to a given expression. e.g.
1) for ``x``,  it stores ``_d_x``
2) for ``x[i]``, ``_d_x[i]``

etc.

We only use ``m_DerivativeForForwSweep`` to initialize reference-type variables. However, to initialize pointers, we use ``getExpr_dx()``. Pointers and references are equivalent and should be initialized the same way. When differentiating expressions, we initialize ``m_DerivativeForForwSweep`` either with the same value as ``Expr_dx`` or with ``nullptr``.

### DIFF
--- a/include/clad/Differentiator/VisitorBase.h
+++ b/include/clad/Differentiator/VisitorBase.h
@@ -31,15 +31,12 @@ namespace clad {
   class StmtDiff {
   private:
     std::array<clang::Stmt*, 2> data;
-    clang::Stmt* m_DerivativeForForwSweep;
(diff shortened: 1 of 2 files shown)


vgvassilev:
Can you rebase?

vgvassilev (approved):
LGTM!

PR #1232, opened 2025-02-03
Don't set init and init style of a declaration manually

PetroZarytskyi:
This PR addresses 2 issues:
1) Initialization styles should not be set manually, instead, they should be deduced by clang based on the types of the declaration and init, as well as on whether the declaration is direct.
 e.g. direct initialization of `std::vector vec` with `other` should look like
```
std::vector vec(other);
```
and not
```
std::vector vec = other;
```

2) Sometimes we initialize declarations using ``Decl::setInit``, which is a low-level function used by clang internally. This PR introduces a wrapper of ``Sema::AddInitializerToDecl`` called ``SetDeclInit``.

### DIFF
--- a/include/clad/Differentiator/VisitorBase.h
+++ b/include/clad/Differentiator/VisitorBase.h
@@ -273,6 +273,15 @@ namespace clad {
     /// \returns Expression with correct Unary Operator placement.
     clang::Expr* ResolveUnaryMinus(clang::Expr* E, clang::SourceLocation OpLoc);
     clang::Expr* BuildParens(clang::Expr* E);
+    /// Sets Init as the initializer of the declaration VD and compute its
+    /// initialization kind.
+    ///\param[in] VD - variable declaration
+    ///\param[in] Init - can be nullptr, then only initialization kind is
+    /// computed.
+    ///\param[in] DirectInit - tells whether the initialization is
+    /// direct.
+    void SetDeclInit(clang::VarDecl* VD, clang::Expr* Init = nullptr,
+                     bool DirectInit = false);
     /// Builds variable declaration to be used inside the derivative
     /// body.
     /// \param[in] Type The type of variable declaration to build.
@@ -478,7 +487,7 @@ namespace clad {
                                      clang::Sema::AA_Casting);
       assert(!ICAR.isInvalid() && "Invalid implicit conversion!");
       // Assign the resulting expression to the variable declaration
-      VD->setInit(ICAR.get());
+      SetDeclInit(VD, ICAR.get());
     }
 
     /// Build a call to member function through Base expr and using the function
--- a/include/clad/Differentiator/VisitorBase.h
+++ b/include/clad/Differentiator/VisitorBase.h
@@ -15,6 +15,9 @@
 #include "clang/AST/StmtVisitor.h"
 #include "clang/Sema/ParsedAttr.h"
 #include "clang/Sema/Sema.h"
+#include <clang/AST/Type.h>
+#include <llvm/ADT/StringRef.h>
+
 #include <array>
 #include <stack>
 #include <unordered_map>
@@ -284,9 +287,7 @@ namespace clad {
     clang::VarDecl*
     BuildVarDecl(clang::QualType Type, clang::IdentifierInfo* Identifier,
                  clang::Scope* scope, clang::Expr* Init = nullptr,
-                 bool DirectInit = false, clang::TypeSourceInfo* TSI = nullptr,
-                 clang::VarDecl::InitializationStyle IS =
-                     clang::VarDecl::InitializationStyle::CInit);
+                 bool DirectInit = false, clang::TypeSourceInfo* TSI = nullptr);
     /// Builds variable declaration to be used inside the derivative
     /// body.
     /// \param[in] Type The type of variable declaration to build.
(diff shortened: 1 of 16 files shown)


include/clad/Differentiator/VisitorBase.h
```diff
@@ -270,6 +272,10 @@ namespace clad {
     /// \returns Expression with correct Unary Operator placement.
     clang::Expr* ResolveUnaryMinus(clang::Expr* E, clang::SourceLocation OpLoc);
     clang::Expr* BuildParens(clang::Expr* E);
+    /// Function to set Init as the initializer of the declaration VD. Init can
+    /// be ``nullptr``. DirectInit tells whether the initialization is direct.
```
vgvassilev, line 276:
```suggestion
    /// Sets Init as the initializer of the declaration VD and compute its
    /// initialization kind. 
    ///\param[in] VD - variable declaration 
    ///\param[in] Init - can a nullptr, then only initialization kind is computed.
    ///\param[in] DirectInit - tells whether the initialization is direct.
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3005,15 +2997,11 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
-              m_Sema.AddInitializerToDecl(
-                  decl, Clone(getArraySizeExpr(AT, m_Context, *this)), true);
-              decl->setInitStyle(VarDecl::InitializationStyle::CallInit);
-            } else {
-              m_Sema.AddInitializerToDecl(decl, getZeroInit(VD->getType()),
-                                          /*DirectInit=*/true);
-              decl->setInitStyle(VarDecl::InitializationStyle::CInit);
-            }
+            if (const auto* AT = dyn_cast<ArrayType>(VD->getType()))
+              SetDeclInit(decl, Clone(getArraySizeExpr(AT, m_Context, *this)),
```
vgvassilev, line 3001: Shouldn't we sink this in `getZeroInit`?

lib/Differentiator/VectorForwardModeVisitor.cpp
```diff
@@ -502,9 +502,8 @@ StmtDiff VectorForwardModeVisitor::VisitReturnStmt(const ReturnStmt* RS) {
   // If we are in vector mode, we need to wrap the return value in a
   // vector.
   QualType cladArrayType = GetCladArrayOfType(utils::GetValueType(retType));
-  auto dVectorParamDecl =
-      BuildVarDecl(cladArrayType, "_d_vector_return", derivedRetValE, false,
-                   nullptr, VarDecl::InitializationStyle::CallInit);
+  VarDecl* dVectorParamDecl = BuildVarDecl(cladArrayType, "_d_vector_return",
```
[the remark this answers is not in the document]
vgvassilev: Not a bad suggestion I think...

lib/Differentiator/VisitorBase.cpp
```diff
@@ -107,29 +107,44 @@ namespace clad {
...
     delete oldScope;
   }
+
+  void VisitorBase::SetDeclInit(VarDecl* VD, Expr* Init, bool DirectInit) {
+    if (!Init) {
+      // Clang sets inits only once. Therefore, ActOnUninitializedDecl does
+      // not reset the init and we have to do it manually.
+      VD->setInit(nullptr);
+      m_Sema.ActOnUninitializedDecl(VD);
+      return;
```
vgvassilev, line 117: Does `ActOnUninitializedDecl` set the InitStyle, too? If so we should update the comment.

vgvassilev (commented):
This looks pretty awesome. See comments inline.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: Here, arrays are converted to ``clad::array`` to promote them to the global scope.  And zero-initialization of clad::arrays is done with the size. The problem is that we cannot pass the original type because it's not ``clad::array`` and we cannot pass the derived type ``clad::array`` because it doesn't contain the array size.

lib/Differentiator/VisitorBase.cpp
PetroZarytskyi: Does init style matter when it comes to uninitialized decls?

vgvassilev: Depends on what clang does... Probably does not but then we need to update the documentation comment of `SetDeclInit`.

lib/Differentiator/VectorForwardModeVisitor.cpp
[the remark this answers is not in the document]
PetroZarytskyi: Done.

include/clad/Differentiator/VisitorBase.h
PetroZarytskyi: Done.

include/clad/Differentiator/VisitorBase.h
```diff
@@ -16,6 +16,8 @@
 #include "clang/Sema/ParsedAttr.h"
 #include "clang/Sema/Sema.h"
 #include <array>
+#include <clang/AST/Type.h>
+#include <llvm/ADT/StringRef.h>
```
vgvassilev, line 20: We should put these before array and sort alphabetically.

include/clad/Differentiator/VisitorBase.h
```diff
@@ -270,6 +272,13 @@ namespace clad {
     /// \returns Expression with correct Unary Operator placement.
     clang::Expr* ResolveUnaryMinus(clang::Expr* E, clang::SourceLocation OpLoc);
     clang::Expr* BuildParens(clang::Expr* E);
+    /// Sets Init as the initializer of the declaration VD and compute its
+    /// initialization kind. 
+    ///\param[in] VD - variable declaration 
+    ///\param[in] Init - can a nullptr, then only initialization kind is computed.
```
vgvassilev, line 278:
```suggestion
    ///\param[in] Init - can be nullptr, then only initialization kind is
    ///                           computed.
```

This needs to be updated according to the comment on the implementation of the function.

lib/Differentiator/VectorForwardModeVisitor.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include "clang/AST/TemplateName.h"
 #include "clang/Sema/Lookup.h"
+#include <clang/AST/Decl.h>
```
vgvassilev, line 8: That needs to be sorted..

include/clad/Differentiator/VisitorBase.h
PetroZarytskyi: Done.

include/clad/Differentiator/VisitorBase.h
PetroZarytskyi: I tried ordering everything alphabetically but clang-format reverted all changes. I separated the includes into different blocks, clang-format seems to be okay with this. I hope this is better.

vgvassilev (approved):
LGTM!

PR #1234, opened 2025-02-04
Fix clang-format workflow

mcbarton:
The clang format workflow is broken because it tries to get a git-clang-format by wgetting it. Git-clang-format is not used as far as I can tell from the workflow, so should be safe to just remove the 'Download git-clang-format' section of the workflow.


vgvassilev (approved):
Let’s see if that works out.

PR #1240, opened 2025-02-14
Force differentiation w.r.t. all non-const pointer/array type parameters

PetroZarytskyi:
We have an old problem that we need the user to differentiate w.r.t. pointer/array parameters. In those cases, the user does not provide an adjoint for the parameters and we have to initialize them ourselves. And while it's clear how to deal with numerical types, e.g.
```
f_grad_0(double x, double y, double* _d_x) { // the user only wants the derivative w.r.t. x
    double _d_y = 0; // so we initialize _d_y ourselves
...
```
we cannot do the same with array/pointer types:
```
f_grad_0(double x, double* y, double* _d_x) { // the user only wants the derivative w.r.t. x
    double* _d_y = ???; // we don't know the size of y and, therefore, cannot initialize _d_y
...
```

It's worth noting that while the user is not directly interested in ``_d_y`` from the last example, ``_d_y`` can be used in the derivative internally. However, we can guarantee that ``_d_y`` will not be useful if ``y`` is const. Here's why this is the case:
In the reverse mode, "``y`` depends on ``x``" results in "``_d_x`` depends on ``_d_y``", e.g.
```
y = x;  -->  _d_x += _d_y;
```
Therefore, if ``y`` cannot be modified (i.e. is const), ``_d_y`` will not be used in the reverse pass.

This PR makes a compromise and forces differentiation w.r.t. non-const array/pointer parameters while allowing not differentiating w.r.t. const ones.

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -16,6 +16,7 @@
 #include "clang/Sema/Sema.h"
 
 #include <llvm/ADT/ArrayRef.h>
+#include <llvm/ADT/SmallVector.h>
 
 #include <array>
 #include <limits>
@@ -43,6 +44,7 @@ namespace clad {
     // a separate namespace, as well as add getters/setters function of
     // several private/protected members of the visitor classes.
     friend class ErrorEstimationHandler;
+    llvm::SmallVector<const clang::ParmVarDecl*, 16> m_NonIndepParams;
     /// In addition to a sequence of forward-accumulated Stmts (m_Blocks), in
     /// the reverse mode we also accumulate Stmts for the reverse pass which
     /// will be executed on return.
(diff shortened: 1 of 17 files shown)


include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -43,7 +43,7 @@ namespace clad {
     // a separate namespace, as well as add getters/setters function of
     // several private/protected members of the visitor classes.
     friend class ErrorEstimationHandler;
-    llvm::SmallVector<const clang::ValueDecl*, 16> m_IndependentVars;
+    llvm::SmallVector<const clang::ParmVarDecl*, 16> m_NonIndepParams;
```
vgvassilev, line 46: Can we move this in the `DiffRequest` class?

PetroZarytskyi: I think it would be quite useful to do this on a higher level because we generate these parameter adjoints ``_d_x`` in all modes. However, ``m_NonIndepParams `` contains cloned variables that are not available on the ``DiffRequest`` stage. We could map the original variables to the cloned ones but we don't have such a map right now. I think we could make this work with some rearrangements in cloning.

parth-07 (changes requested):
Looks good to me, apart from one nitpick comment.

parth-07 (approved):
Looks good! It was a much needed functionality.

PR #1243, opened 2025-02-17
No need to check if InitListExpr of a RecordType is a struct

### LINKED ISSUE #685: Clad is not happy with empty initializer lists
In RooFit, we were recently dealing with code like this:
```c++
#include <iostream>

class Identity {
    public:
    double operator()(double u) { return u + 1; }
};

double algo(double x) {
    //Identity di; // works!
    Identity di{}; // doesn't work!
    double val = di(x);
    return val * val;
}

int main() {
    auto grad = clad::gradient(algo);
    double x=3, dx=0;
    grad.execute(x, &dx);
    show(dx); // This prints: x: 6
}
```
It gave the following error:
```txt
clang-16: /home/rembserj/PKGBUILDs/clad/src/clad/lib/Differentiator/ReverseModeVisitor.cpp:1227: clad::StmtDiff clad::ReverseModeVisitor::VisitInitListExpr(const clang::InitListExpr*): Assertion `isValueAndPushforward && "Only InitListExpr that represents arrays or ValueAndPushforward " "Object initialization is supported"' failed.
```
Maybe it should also be considered acceptable if the initializer list is empty? Or is there a reason why it should not work? It would be nice if it works for compatibility reasons.

PetroZarytskyi:
If a ``RecordType`` can be initialized with an ``InitListExpr``, it's an aggregate and its fields are accessible. Checking if it is a struct is redundant.
Also, we handle ``InitListExpr`` of type ``ValueAndPushforward`` as a special case. This is unnecessary since the general logic for ``RecordType`` works fine with ``ValueAndPushforward``.
Fixes #685.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1176,47 +1176,20 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       return StmtDiff(clonedILE);
     }
     // Check if type is a CXXRecordDecl
-    if (!utils::IsCladValueAndPushforwardType(ILEType) &&
-        ILEType->isRecordType()) {
-      for (unsigned i = 0, e = ILE->getNumInits(); i < e; i++) {
-        // fetch ith field of the struct.
-        auto field_iterator = ILEType->getAsCXXRecordDecl()->field_begin();
-        std::advance(field_iterator, i);
-        Expr* member_acess = utils::BuildMemberExpr(
-            m_Sema, getCurrentScope(), dfdx(), (*field_iterator)->getName());
-        clonedExprs[i] = Visit(ILE->getInit(i), member_acess).getExpr();
-      }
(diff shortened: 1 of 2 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1175,48 +1175,21 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
-      }
-      Expr* clonedILE = m_Sema.ActOnInitList(noLoc, clonedExprs, noLoc).get();
-      return StmtDiff(clonedILE);
-    }
-
-    // FIXME: This is a makeshift arrangement to differentiate an InitListExpr
-    // that represents a ValueAndPushforward type. Ideally this must be
-    // differentiated at VisitCXXConstructExpr
+    // Check if type is a CXXRecordDecl
 #ifndef NDEBUG
```
vgvassilev, line 1179: We can drop this `ifndef` now.

vgvassilev (approved):
LGTM! Codecov confuses itself and we can ignore the ubuntu arm bot...

Issue #760, opened 2024-02-14
Make tbr analysis default


vgvassilev:

vgvassilev:
Depends on #855.

vgvassilev:
Fixed by https://github.com/vgvassilev/clad/pull/1245

PR #1245, opened 2025-02-18
Make TBR default

ovdiiuv:
Fixes:#760


vgvassilev:
@guitargeek, can we check if this PR works with RooFit?

ovdiiuv:
> @guitargeek, can we check if this PR works with RooFit?

I have already checked that on Cms and Atlas, so as long as those remain the same there is no need.

guitargeek:
Yes, the CMS and ATLAS likelihoods that I use for validations are still the same that I sent you guys. But building ROOT against Clad with this PR and see if the tests pass also does no harm. Doing that now.

guitargeek:
Works fine for RooFit after this patch, which is fixing compatibility with ROOT master independently of this PR:
  * https://github.com/root-project/root/pull/17786

PetroZarytskyi:
LGTM! I'm happy to see TBR working by default. Let's now hope we don't get many new bugs.

PetroZarytskyi (approved):

PR #1246, opened 2025-02-19
Add support of variadic functions as call exprs in Forward mode

kchristin22:
Changes applied:
* Use Call arg instead of Function param to store differentiation call's args
* Create Unique Identifier for each param expanded from the pack of a variadic function
* Add test

### DIFF
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -282,11 +282,27 @@ QualType BaseForwardModeVisitor::ComputeDerivativeFunctionType() {
 void BaseForwardModeVisitor::SetupDerivativeParameters(
     llvm::SmallVectorImpl<ParmVarDecl*>& params) {
   const FunctionDecl* FD = m_DiffReq.Function;
-  for (ParmVarDecl* PVD : FD->parameters()) {
+  const FunctionDecl* FDPattern = nullptr;
+  unsigned FDPatternNumParams = 0;
+  if (FD->isTemplateInstantiation()) {
+    FDPattern = FD->getTemplateInstantiationPattern();
+    FDPatternNumParams = FDPattern->getNumParams();
+  }
+  for (unsigned i = 0, n = FD->getNumParams(); i < n; ++i) {
+    const ParmVarDecl* PVD = FD->getParamDecl(i);
     IdentifierInfo* PVDII = PVD->getIdentifier();
     // Implicitly created special member functions have no parameter names.
     if (!PVD->getDeclName())
       PVDII = CreateUniqueIdentifier("param");
+
+    if (FDPattern) {
+      const ParmVarDecl* OrigPVD =
+          i >= FDPatternNumParams
+              ? FDPattern->getParamDecl(FDPatternNumParams - 1)
+              : FDPattern->getParamDecl(i);
+      if (OrigPVD->isParameterPack())
+        PVDII = CreateUniqueIdentifier(PVDII->getName());
+    }
 
     auto* newPVD = CloneParmVarDecl(PVD, PVDII,
                                     /*pushOnScopeChains=*/true,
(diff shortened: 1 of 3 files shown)


test/ForwardMode/VariadicCall.C
```diff
@@ -3,20 +3,23 @@
 
 #include "clad/Differentiator/Differentiator.h"
 
-#include "../TestUtils.h"
+// CHECK: warning: function 'printf' was not differentiated because clad failed to differentiate it and no suitable overload was found in namespace 'custom_derivatives'
```
vgvassilev, line 6: We should add `-Xclang -verify` and use the syntax of `// expected-warning ...`

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -295,7 +311,8 @@ void BaseForwardModeVisitor::SetupDerivativeParameters(
     if (PVD == m_IndependentVar)
       m_IndependentVar = newPVD;
 
-    if (!PVD->getDeclName()) // We can't use lookup-based replacements
+    if (PVD->getDeclName() !=
+        newPVD->getDeclName()) // We can't use lookup-based replacements
```
vgvassilev, line 315:
```suggestion
    // We can't use lookup-based replacements
    if (PVD->getDeclName() != newPVD->getDeclName()) 
```

PetroZarytskyi (commented):
The PR looks amazing. I left only a small question.

test/ForwardMode/VariadicCall.C
```diff
@@ -0,0 +1,52 @@
...
+
+#include "clad/Differentiator/Differentiator.h"
+
+template <typename... T>
+double fn1(double x, T... y) {
+    return (x * ... * y);
+}
+
+double fn2(double x, double y) {
+    printf("x is %f, y is %f\n", x, y); // expected-warning {{function 'printf' was not differentiated because clad failed to differentiate it and no suitable overload was found in namespace 'custom_derivatives'}}
```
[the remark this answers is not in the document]
kchristin22: It was an easy example of a variadic function that came to mind : ) Does it look too odd here?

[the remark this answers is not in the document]
PetroZarytskyi: I just thought it didn't test variadic properties. But perhaps there's nothing bad about this test.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1118,7 +1136,11 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
     // If original argument is an RValue and function expects an RValue
     // parameter, then convert the cloned argument and the corresponding
     // derivative to RValue if they are not RValue.
-    QualType paramType = FD->getParamDecl(i - skipFirstArg)->getType();
+    QualType paramType;
+    if (FD->isVariadic() && (i - skipFirstArg) >= numParams)
+      paramType = CE->getArg(i - skipFirstArg)->getType();
```
kchristin22, line 1141: @PetroZarytskyi @vgvassilev the `printf` example was to cover this line, cause the template functions are not considered variadic.

To write one without the template, I would have to use builtin functions like `va_start`, `va_end` and `va_arg`, for which the pushforwards are not so obvious. It would need further discussion, but I don't think this is a priority right now as we don't have such a use case.

Should I bring back the `printf` example or leave it like that?

kchristin22: Thoughts? @vgvassilev @PetroZarytskyi

vgvassilev: Makes sense to me assuming you capture that in a fixme in the test and open an issue.

kchristin22: So add `printf` stmt and open an issue on `va_start` etc.?

PR #1247, opened 2025-02-24
Support default arguments of object types in the reverse mode

PetroZarytskyi:
The major use case of this feature is supporting ``std::initializer_list``-based constructors of STL containers. Most of them have a second default argument for the allocator.


vgvassilev (approved):
LGTM!

PR #1248, opened 2025-02-24
Compute the differentiation plan before differentiation starts.

### LINKED ISSUE #1009: Improve the printing of the DynamicGraph
https://github.com/vgvassilev/clad/blob/69d15a31d11fa38b193d274e2392a086f54006bc/include/clad/Differentiator/DynamicGraph.h#L112 outputs something like:
```
f1__order_1__mode_hessian: #0 (source), (done)
f2__order_1__mode_hessian: #1 (source), (done)
f3__order_1__mode_hessian: #2 (source), (done)
f4__order_1__mode_hessian: #3 (source), (done)
f5__order_1__mode_hessian: #4 (source), (done)
f6__order_1__mode_hessian: #5 (source), (done)
f7__order_1__mode_hessian: #6 (source), (done)
f8__order_1__mode_hessian: #7 (source), (done)
f1__order_1__mode_forward: #8, (done)
f1_darg0__order_1__mode_reverse: #9, (done)
f2__order_1__mode_forward: #10, (done)
f2_darg0__order_1__mode_reverse: #11, (done)
f3__order_1__mode_forward: #12, (done)
f3_darg0__order_1__mode_reverse: #13, (done)
f4__order_1__mode_forward: #14, (done)
f4_darg0__order_1__mode_reverse: #15, (done)
f5__order_1__mode_forward: #16, (done)
f5_darg0__order_1__mode_reverse: #17, (done)
f6__order_1__mode_forward: #18, (done)
f6_darg0__order_1__mode_reverse: #19, (done)
f6__order_1__mode_forward: #20, (done)
f6_darg1__order_1__mode_reverse: #21, (done)
f7__order_1__mode_forward: #22, (done)
f7_darg0__order_1__mode_reverse: #23, (done)
f7__order_1__mode_forward: #24, (done)
f7_darg1__order_1__mode_reverse: #25, (done)
sin_pushforward__order_1__mode_pullback: #26, (done)
cos_pushforward__order_1__mode_pullback: #27, (done)
exp_pushforward__order_1__mode_pullback: #28, (done)
log_pushforward__order_1__mode_pullback: #29, (done)
pow_pushforward__order_1__mode_pullback: #30, (done)
__builtin_pow_pushforward__order_1__mode_pullback: #31, (unprocessed)
__builtin_pow__order_1__mode_pullback: #32, (unprocessed)
0 -> 8
0 -> 9
1 -> 10
1 -> 11
2 -> 12
2 -> 13
3 -> 14
3 -> 15
4 -> 16
4 -> 17
5 -> 18
5 -> 19
5 -> 20
5 -> 21
6 -> 22
6 -> 23
6 -> 24
6 -> 25
9 -> 26
9 -> 27
11 -> 28
13 -> 29
15 -> 30
30 -> 31
31 -> 32
```

It would be more understandable if we visualize it they way we visualize the AST with indentations such as:
```
`f1__order_1__mode_hessian: #0 (source), (done)
|   `f1__order_1__mode_forward: #8, (done)
|   `f1_darg0__order_1__mode_reverse: #9, (done)
|  ...
```

vgvassilev:
This patch creates a DiffRequest graph statically and we can plan accordingly the sequence of differentiation steps. The key advantage is that we can remove the recursive calls to ProcessDiffRequest. Recursive diff requests are not good because Clang is not designed to build functions while building other functions.

This patch does not solve all nested requests just yet but lays the foundation of infrastructure that will help removing the remaining parts in an upcoming pull request.

Fixes #1009.

### DIFF
--- a/include/clad/Differentiator/DiffPlanner.h
+++ b/include/clad/Differentiator/DiffPlanner.h
@@ -7,7 +7,9 @@
 
 #include "clang/AST/RecursiveASTVisitor.h"
 
+#include "llvm/ADT/DenseSet.h"
 #include "llvm/Support/Compiler.h"
+#include "llvm/Support/SaveAndRestore.h"
 #include "llvm/Support/raw_ostream.h"
 
 #include <iterator>
@@ -120,15 +122,15 @@ public:
   ///      function will be differentiated w.r.t. to its every parameter.
   void UpdateDiffParamsInfo(clang::Sema& semaRef);
 
-  /// Define the == operator for DiffRequest.
+  /// Allow comparing DiffRequests.
   bool operator==(const DiffRequest& other) const {
-    // either function match or previous declaration match
+    // Note that CallContext is always different and we should ignore it.
     return Function == other.Function &&
            BaseFunctionName == other.BaseFunctionName &&
            CurrentDerivativeOrder == other.CurrentDerivativeOrder &&
            RequestedDerivativeOrder == other.RequestedDerivativeOrder &&
-           CallContext == other.CallContext && Args == other.Args &&
-           Mode == other.Mode && EnableTBRAnalysis == other.EnableTBRAnalysis &&
+           Args == other.Args && Mode == other.Mode &&
+           EnableTBRAnalysis == other.EnableTBRAnalysis &&
            EnableVariedAnalysis == other.EnableVariedAnalysis &&
            DVI == other.DVI && use_enzyme == other.use_enzyme &&
            DeclarationOnly == other.DeclarationOnly;
@@ -172,16 +174,27 @@ public:
     /// If set it means that we need to find the called functions and
     /// add them for implicit diff.
     ///
-    const clang::FunctionDecl* m_TopMostFD = nullptr;
+    const DiffRequest* m_TopMostReq = nullptr;
     clang::Sema& m_Sema;
 
     const RequestOptions& m_Options;
+
+    llvm::DenseSet<const clang::FunctionDecl*> m_Traversed;
+
+    bool m_IsTraversingTopLevelDecl = true;
 
   public:
     DiffCollector(clang::DeclGroupRef DGR, DiffInterval& Interval,
                   clad::DynamicGraph<DiffRequest>& requestGraph, clang::Sema& S,
                   RequestOptions& opts);
     bool VisitCallExpr(clang::CallExpr* E);
+    bool TraverseFunctionDeclOnce(const clang::FunctionDecl* FD) {
+      llvm::SaveAndRestore<bool> Saved(m_IsTraversingTopLevelDecl, false);
+      if (m_Traversed.count(FD))
+        return true;
+      m_Traversed.insert(FD);
+      return TraverseDecl(const_cast<clang::FunctionDecl*>(FD));
+    }
 
   private:
     bool isInInterval(clang::SourceLocation Loc) const;
--- a/include/clad/Differentiator/DiffPlanner.h
+++ b/include/clad/Differentiator/DiffPlanner.h
@@ -151,6 +151,7 @@ public:
   bool shouldBeRecorded(clang::Expr* E) const;
   bool shouldHaveAdjoint(const clang::VarDecl* VD) const;
   std::string ComputeDerivativeName() const;
+  bool HasIndependentParameter(const clang::ParmVarDecl* PVD) const;
 };
 
   using DiffInterval = std::vector<clang::SourceRange>;
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -158,9 +158,9 @@ namespace clad {
       ASTContext& C = semaRef.getASTContext();
 
       if (auto ND = dyn_cast<NamespaceDecl>(DC)) {
-        CSS.Extend(C, ND,
-                   /*NamespaceLoc=*/utils::GetValidSLoc(semaRef),
-                   /*ColonColonLoc=*/utils::GetValidSLoc(semaRef));
+        if (!ND->isInline())
+          CSS.Extend(C, ND, /*NamespaceLoc=*/utils::GetValidSLoc(semaRef),
+                     /*ColonColonLoc=*/utils::GetValidSLoc(semaRef));
       } else if (auto RD = dyn_cast<CXXRecordDecl>(DC)) {
         auto RDQType = RD->getTypeForDecl()->getCanonicalTypeInternal();
         auto RDTypeSourceInfo = C.getTrivialTypeSourceInfo(RDQType);
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -263,6 +263,10 @@ namespace clad {
       if (!DC)
         DC = C.getTranslationUnitDecl();
       S.LookupQualifiedName(Result, DC);
+
+      if (auto* CXXRD = dyn_cast<CXXRecordDecl>(DC))
+        Result.setNamingClass(CXXRD);
+
       return Result;
     }
 
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -11,10 +11,12 @@
 #include "clang/AST/Attr.h"
 #include "clang/AST/Decl.h"
 #include "clang/AST/ExprCXX.h"
+#include "clang/AST/OperationKinds.h"
 #include "clang/AST/TemplateBase.h"
 #include "clang/AST/Type.h"
 #include "clang/Basic/LLVM.h" // isa, dyn_cast
 #include "clang/Basic/Specifiers.h"
+#include "clang/Basic/TokenKinds.h"
 #include "clang/Sema/Lookup.h"
 #include "clang/Sema/Overload.h"
 #include "clang/Sema/Scope.h"
@@ -343,13 +345,14 @@ static void registerDerivative(FunctionDecl* dFD, Sema& S,
           Member.setIdentifier(&m_Context.Idents.get(Name), Loc);
           bool isArrow = Base->getType()->isPointerType();
           // FIXME: update SS here?
-          auto ME = m_Sema
-                        .ActOnMemberAccessExpr(S, Base, Loc,
-                                               isArrow ? tok::TokenKind::arrow
-                                                       : tok::TokenKind::period,
-                                               SS, noLoc, Member,
-                                               /*ObjCImpDecl=*/nullptr)
-                        .get();
+          auto* ME =
+              m_Sema
+                  .ActOnMemberAccessExpr(S, Base, Loc,
+                                         isArrow ? tok::TokenKind::arrow
+                                                 : tok::TokenKind::period,
+                                         SS, noLoc, Member,
+                                         /*ObjCImpDecl=*/nullptr)
+                  .get();
           if (noOverloadExists(ME, MARargs.drop_front()))
             return nullptr;
 
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -12,6 +12,7 @@
 #include "clang/AST/Decl.h"
 #include "clang/AST/ExprCXX.h"
 #include "clang/AST/TemplateBase.h"
+#include "clang/AST/Type.h"
 #include "clang/Basic/LLVM.h" // isa, dyn_cast
 #include "clang/Basic/Specifiers.h"
 #include "clang/Sema/Lookup.h"
@@ -35,6 +36,7 @@
 #include "llvm/Support/SaveAndRestore.h"
 
 #include <algorithm>
+#include <cstddef>
 #include <string>
 
 #include "clad/Differentiator/CladUtils.h"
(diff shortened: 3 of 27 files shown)


include/clad/Differentiator/DiffPlanner.h
```diff
@@ -127,7 +129,7 @@ struct DiffRequest {
            BaseFunctionName == other.BaseFunctionName &&
            CurrentDerivativeOrder == other.CurrentDerivativeOrder &&
            RequestedDerivativeOrder == other.RequestedDerivativeOrder &&
-           CallContext == other.CallContext && Args == other.Args &&
+           /*CallContext == other.CallContext &&*/ Args == other.Args &&
```
PetroZarytskyi, line 132: Should this be left commented out?

vgvassilev: Good point, however, I want to make sure that nobody thinks this is an oversight of the implementation of the comparison operator. OTOH, re-adding it will break pretty much everything so I could delete it...

parth-07 (commented):
The change looks good! It will simplify a lot of things. Initial review comments.

test/Gradient/FunctionCalls.C
```diff
@@ -527,6 +677,27 @@ double recFun (double x, double y) {
 }
 
 //CHECK: void recFun_pullback(double x, double y, double _d_y0, double *_d_x, double *_d_y);
+//CHECK: void recFun_pullback(double x, double y, double _d_y0, double *_d_x, double *_d_y) {
```
PetroZarytskyi, line 680: Why do we still have a forward declaration here?

include/clad/Differentiator/DiffPlanner.h
vgvassilev: Addressed.

lib/Differentiator/CladUtils.cpp
```diff
@@ -158,9 +158,9 @@ namespace clad {
       ASTContext& C = semaRef.getASTContext();
 
       if (auto ND = dyn_cast<NamespaceDecl>(DC)) {
-        CSS.Extend(C, ND,
-                   /*NamespaceLoc=*/utils::GetValidSLoc(semaRef),
-                   /*ColonColonLoc=*/utils::GetValidSLoc(semaRef));
+        if (!ND->isInline())
```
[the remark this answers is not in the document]
vgvassilev: Yes, we ignore the inline namespace as we start producing code which differs from the code as written. Eg: `custom_derivatives::std::__1::custom_pullback`. That's why we strip it here.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -58,7 +59,14 @@ static void registerDerivative(FunctionDecl* dFD, Sema& S,
                                const DiffRequest& R) {
   DeclContext* DC = dFD->getLexicalDeclContext();
   LookupResult Previous(S, dFD->getNameInfo(), Sema::LookupOrdinaryName);
-  S.LookupQualifiedName(Previous, dFD->getParent());
+  // Template instantiations of function templates should not be considered
+  // redeclarations.
+  // FIXME: Currently we produce a FunctionDecl per instantiation, however, we
+  // should follow closer what clang does, namely building a
```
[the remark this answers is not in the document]
vgvassilev: I think it is doable because the primary template should be the same across all implicit instantiations. For the explicit specializations and partial specializations the logic should be different.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -874,62 +903,162 @@ namespace clad {
     return false;
   }
 
+  static bool HasCustomDerivativeForDecl(Sema& S, const DiffRequest& R) {
```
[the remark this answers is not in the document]
vgvassilev: Done.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -874,62 +903,162 @@ namespace clad {
...
+        request.Mode = DiffMode::experimental_pushforward;
+      else if (m_TopMostReq->Mode == DiffMode::reverse)
+        request.Mode = DiffMode::experimental_pullback;
+      else {
+        // propagatorReq.Mode = request.Mode;
+      }
+      request.VerboseDiags = false;
+      request.EnableTBRAnalysis = m_TopMostReq->EnableTBRAnalysis;
+      request.EnableVariedAnalysis = m_TopMostReq->EnableVariedAnalysis;
+      // propagatorReq.CUDAGlobalArgsIndexes = globalCallArgs;
```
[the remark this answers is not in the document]
vgvassilev: Yes, addressed.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -874,62 +898,183 @@ namespace clad {
...
+      if (!isa<CXXNullPtrLiteralExpr>(A) && !isa<CharacterLiteral>(A) &&
+          !isa<CXXBoolLiteralExpr>(A) && !isa<FixedPointLiteral>(A) &&
+          !isa<ImaginaryLiteral>(A) && !isa<IntegerLiteral>(A) &&
+          !isa<ObjCBoolLiteralExpr>(A) && !isa<ObjCStringLiteral>(A) &&
+          !isa<FloatingLiteral>(A) && !isa<ObjCArrayLiteral>(A) &&
+          !isa<StringLiteral>(A) && !isa<CompoundLiteralExpr>(A))
+        return false; // non-constant found.
+    }
+    return true; // all constants.
+  }
```
PetroZarytskyi, line 962: Might be well known but I'll reiterate just in case: unlike on master, with the current state of this PR, activity analysis doesn't influence scheduling derivatives anymore, e.g.
```
double f8(double x) {
  double u = 9;
  double c = f8_1(u, u);
  return c;
}
```
is differentiated on master as
```
void f8_grad(double x, double *_d_x) {
    double u = 9;
    double c = f8_1(u, u);
}
```
and with this PR
```
void f8_1_pullback(double v, double u, double _d_y, double *_d_v, double *_d_u) {
    *_d_v += _d_y;
}
void f8_grad(double x, double *_d_x) {
    double u = 9;
    double c = f8_1(u, u);
}
```
This should be easy to address. We need to attach varied + constness analysis here. I already moved this logic to DiffPlanner with 964e2546e77b72174e7444733ef115fcb6945340 (as part of #1254) so it might come in handy.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -874,62 +898,183 @@ namespace clad {
...
+        for (size_t i = 0, e = FD->getNumParams(); i < e; ++i) {
+          // if (MD && isLambdaCallOperator(MD)) {
+          const auto* paramDecl = FD->getParamDecl(i);
+          request.DVI.push_back(paramDecl);
+
+          //}
+          // FIXME:
+          // else if (DerivedCallOutputArgs[i + (bool)MD]) {
+          //  propagatorReq.DVI.push_back(FD->getParamDecl(i));
+          //}
```
PetroZarytskyi, line 1049: We don't need the lines commented out above, right?

PetroZarytskyi (commented):
I added a few comments but the PR looks great. Looking forward to it being merged.

test/Gradient/FunctionCalls.C
vgvassilev: Good catch! Looks like now we generate both for an unknown reason. I will take a look...

lib/Differentiator/DiffPlanner.cpp
vgvassilev: Can we do that as a subsequent PR. This PR is already getting out of hand :)

lib/Differentiator/DiffPlanner.cpp
vgvassilev: That's intentional. I want to make sure I do not forget that obscure logic that is essentially in the visitors when deciding whether to build a pullback..

lib/Differentiator/DiffPlanner.cpp
PetroZarytskyi: Sure, sounds good.

lib/Differentiator/DiffPlanner.cpp
PetroZarytskyi: Do those bits essentially mean we need to start considering the differentiability of the base? If so, should we replace this with a fixme explaining that?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -874,62 +901,183 @@ namespace clad {
...
+    const Expr* callSite = R.CallContext;
+    const DeclContext* originalFnDC = nullptr;
+    // Check if the callSite is not associated with a shadow declaration.
+    if (const auto* ME = dyn_cast<CXXMemberCallExpr>(callSite)) {
+      originalFnDC = ME->getMethodDecl()->getParent();
+    } else if (const auto* CE = dyn_cast<CallExpr>(callSite)) {
+      const Expr* Callee = CE->getCallee()->IgnoreParenCasts();
+      if (const auto* DRE = dyn_cast<DeclRefExpr>(Callee))
+        originalFnDC = DRE->getFoundDecl()->getDeclContext();
+      else if (const auto* MemberE = dyn_cast<MemberExpr>(Callee))
```
PetroZarytskyi, line 920: Is this case possible to cover? Don't all CallExprs with MemberExpr callees automatically become MemberCallExprs (which is considered above)?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -874,62 +901,183 @@ namespace clad {
...
+        originalFnDC = MemberE->getFoundDecl().getDecl()->getDeclContext();
+    } else if (const auto* CtorExpr = dyn_cast<CXXConstructExpr>(callSite)) {
+      originalFnDC = CtorExpr->getConstructor()->getDeclContext();
```
PetroZarytskyi, line 923: I think this cannot be covered because we don't differentiate constructors automatically yet. After this PR is merged, I'm going to open a PR about automatic constructor pullbacks. Let's then see if this bit gets covered.

vgvassilev: In theory you can submit a pr against that branch.

PetroZarytskyi: I would also have to open it against #1254 which currently seems to conflict with this branch.

lib/Differentiator/CladUtils.cpp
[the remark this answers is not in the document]
parth-07: I might be missing something, but it seems that for the below example, the `CXXScopeSpec CSS` (in the function `BuildNNS`) corresponding to the inline namespace `C` would not contain `A::B::` scope information. It seems that we only extend `CSS` for non-inline namespaces.

```cpp
namespace A {
  namespace B {
    inline namespace C {
      // ...
    }
  }
}
```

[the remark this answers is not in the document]
parth-07: Oh, I see. `A::B::` will be correctly added to `CXXScopeSpec CSS`, thanks to the recursive nature of the function. Please ignore my previous comment.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -274,18 +308,65 @@ static void registerDerivative(FunctionDecl* dFD, Sema& S,
...
+      // FIXME: Figure out how to assert here in cases where we have provided
+      // both a clad-generated derivative and a user-provided one.
+      // LookupResult R1 = utils::LookupQualifiedName(Name, m_Sema,
+      // originalFnDC); assert((R1.empty() || R1.getFoundDecl() ==
+      // R.getFoundDecl()) &&
+      //        "We clad built a derivative for entity which"
+      //        "has a custom derivative!");
+#endif // NDEBUG
+    } else {
+      // Try to find if clad already built a derivative.
```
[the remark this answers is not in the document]
parth-07: For now, I think the current change is good. We should perhaps do the refactoring later. What do you think about separating this function responsibilities into two functions: `BuildCallToExistingDerivative` and `BuildCallToNumericalDiff`? `BuildCallToExistingDerivative` would consider both the custom derivatives and the Clad built derivatives.

lib/Differentiator/DerivativeBuilder.cpp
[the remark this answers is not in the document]
parth-07: Will we need to directly differentiate a `TemplateDecl` to get the primary template of derivatives?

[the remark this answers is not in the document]
vgvassilev: Yes. That's the idea.

lib/Differentiator/DerivativeBuilder.cpp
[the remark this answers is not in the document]
vgvassilev: That's my plan. I really want to split the calls to the numerical diff because logic there is quite entangled because of it. After the complete move to static graphs we will know ahead of time when we are building a call to numerical diff and probably will remove the implementation of it in this function...

lib/Differentiator/DiffPlanner.cpp
vgvassilev: I think this is when there is a FieldDecl access. Currently if I remove it nothing breaks. Should I remove it?

parth-07 (approved):
Looks really good!

lib/Differentiator/DiffPlanner.cpp
PetroZarytskyi: For what it's worth, this can indeed be covered. For example, if the base of the CallExpr is a FieldDecl access of function pointer type. Perhaps we don't support them, but I think it's good we kept this bit.

vgvassilev: Sorry, I clicked merge too fast. Yes, I decided to leave it because on the other hand the codecov does not report uncovered branches for this one.

PR #1250, opened 2025-02-25
Correctly handle memory taken by objects in loops

PetroZarytskyi:
This PR addresses 2 issues:
1) Currently, object adjoints copy the structure of the original variables (e.g. ``std::vector<...> _d_x(x);``). However, if this happens in a loop, the structure of the adjoint ``_d_x`` might change every iteration. In the reverse pass, we will be left with whatever adjoint was left after the last iteration. This could lead to errors if, for example, the corresponding array had more elements on some previous iteration.
2) When calling ``_reverse_forward`` functions, we store the result as a reference. The problem is that the result is stored as a local variable in the forward pass and cannot be used in the reverse mode. This was left unnoticed because the memory location of the local reference was not overwritten. However, when dealing with a local variable inside a loop, it's guaranteed to be overwritten on next iterations. The solution is to store the references on a tape.
This PR also introduces a test for both issues. In theory, these issues could be considered and tested separately but each one breaks other tests when being introduced on its own. This happens because those tests are currently bugged but work under specific conditions.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2728,9 +2728,9 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       if (VDDerived && isInsideLoop) {
         Stmt* assignToZero = nullptr;
         Expr* declRef = BuildDeclRef(VDDerived);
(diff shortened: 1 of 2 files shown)


vgvassilev (approved):
LGTM!

PR #1256, opened 2025-03-05
Fix static graph cuda

kchristin22:
Allows atomicAdd in device functions based on the global args passed to it


vgvassilev:
Thank you @kchristin22!!

PR #1257, opened 2025-03-05
No need to provide the constructed object to a constructor pullback

PetroZarytskyi:
Constructor pullbacks do not require the original constructed object, only its derivative ``_d_this``. This is similar to how regular function pullbacks do not require the original return value, only the pullback ``_d_y`` argument. Currently, the first argument of constructor pullbacks is never used and, in some cases, is even set to ``nullptr``.

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -1633,11 +1633,10 @@ using std::sqrt_pushforward;
 
 namespace class_functions {
 template <typename T, typename U>
-void constructor_pullback(ValueAndPushforward<T, U>* lhs,
-                          ValueAndPushforward<T, U> rhs,
-                          ValueAndPushforward<T, U>* d_lhs,
+void constructor_pullback(ValueAndPushforward<T, U> rhs,
+                          ValueAndPushforward<T, U>* d_this,
                           ValueAndPushforward<T, U>* d_rhs) {
-  d_rhs->pushforward += d_lhs->pushforward;
+  d_rhs->pushforward += d_this->pushforward;
 }
 } // namespace class_functions
 } // namespace custom_derivatives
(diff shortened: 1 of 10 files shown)


parth-07 (commented):
This is an interesting change. Thank you for creating the PR and initiating this discussion.

>  This is similar to how regular function pullbacks do not require the original return value, only the pullback _d_y argument.

Yes, we do not require the original return value of regular functions in their corresponding pullbacks. However, a constructor is not a regular function and does not return any value. A constructor is more similar to a member function than a free (regular) function because the constructor body can access the constructed data members similar to how data members can be accessed by member functions. This indicates that the object data members are created _before_ the constructor body even begins! To differentiate the statements containing the data members, the pullback of both member functions and constructors would need the (constructed object).

> Currently, the first argument of constructor pullbacks is never used and, in some cases, is even set to nullptr.

Yes, I think that is true so far. But I don't think that would always be the case. Especially once we start automatically differentiating constructors. Please let me know if I am overlooking something obvious. Let's consider this contrived example:

```cpp
struct SomeStruct {
  SomeStruct(int u, int v) {
    m_A = u * v;
    m_B = m_A * u;
  }
  int m_A;
  int m_B;
```

reverse-pass adjoint statements of the primal statement `m_B = m_A * u` in the `SomeStruct` constructor pullback would be something like:

```cpp
d_this->m_A += d_this->m_B * u;
d_u += d_this->m_B * (constructed_object)->m_A;
```

We can, in theory, compute the value of `(constructed_object)->m_A` as part of the constructor pullback so that we do not require the constructed object as a parameter to the constructor pullback, but I don't think this additional pain would be worth the efforts. It would also lead to inconsistency in the pullback signatures of member functions and constructors.

**Update**

It seems that I was overlooking something obvious. Constructed object passed to the pullback would _always_ be zero-initialized and thus need not be passed. It seems okay to consider constructors more similar to free (regular) functions.

parth-07 (approved):

PetroZarytskyi:
Hi @parth-07! Thank you for the analysis. It's cool you mentioned automatic constructor pullbacks because I was working on them and this PR is actually motivated by that work. The problem was exactly the case with ``nullptr`` objects, such pullbacks just crashed. I realized I had to generate this zero object and felt it would be meaningful to hide that logic in the pullback.

Issue #1095, opened 2024-09-15
Add support for `std::get` and template arguments in pushforwards/pullbacks


gojakuch:
currently, if there's a function accepting a template argument that cannot be resolved automatically by inferring it from its arguments (like `std::get` for example), it's not possible to use the automatically generated pushforward/pullback for it:

```c++
#include "clad/Differentiator/Differentiator.h"
#include <iostream>

template<int N>
double take(double x, double y) {
    return 0;
}
template<>
double take<0>(double x, double y) {
    return x;
}
template<>
double take<1>(double x, double y) {
    return y;
}

double fn(double u) {
    double a = take<0>(u, 2*u);
    double b = take<1>(u, 2*u);
    return a+b; // = 3*u
}

int main() {
    auto df = clad::differentiate(fn, "u");
    std::cout << "f(3)=" << fn(3.0);
    std::cout << "; df(3)=" << df.execute(3.0) << '\n';
}
```
**expected output:**
f(3)=9; df(3)=3
**output:**
f(3)=9; df(3)=2

while Clad does create two different pushforwards for these specialisations, it only calls one of them, failing to distinguish between those.

`std::get` when used as `std::get<10>(my_tuple)`, for instance, would cause this too, which is why it's important to solve this issue to fully support STL or even tuples. 

this must be solved in both fwd and rvs modes in a way that also allows custom derivatives to be provided for these functions.

PR #1261, opened 2025-03-10
Add name generation for explicitly specialized template functions

pmozil:
Funcitons in clad are identified by `clad::ComputeEffectiveFnName`. Explicit specializations or functions with integral template arguments were all identified with a signle function, resulting in the given bug. The change in this merge request is that specialization functions are now generated like this:
```cpp
#include <iostream>
#include <tuple>

double fn(double u) {
    std::tuple<double, double> tmp{u, 2*u};
    double a = std::get<0>(tmp);
    double b = std::get<1>(tmp);

    return a + b;
}

int main() {
    auto df = clad::differentiate(fn, "u");
    df.dump();
}
```

Two pushforwards will be generated - `get_specialization___i_0_pushforward` and `get_specialization___i_1_pushforward`. instead of a single `get_pushforward` as previously. `std::get` works fine with this merge request, as does the code in the example for #1095

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2616,20 +2616,6 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     if (!VDDerivedInit)
       VDDerivedInit = getZeroInit(VDType);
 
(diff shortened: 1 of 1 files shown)


vgvassilev:
We don't want to format the whole file. Please use `git-clang-format HEAD~` instead.

vgvassilev:
The diff should be minimal. Do you want me to help you out?

PR #1268, opened 2025-03-14
Improve differentiation of object-type decls in the reverse mode

### LINKED ISSUE #800: Compiler error, no matching constructor

```c++

struct T {
  T(int){}
};

double f(double x){
  T t{0};
  return x;
}

int main() {
  double derivative;
  f_dx.execute(3, &derivative);
}
```

```
s.cpp:8:8: error: no matching constructor for initialization of 'T'
double f(double x){
       ^
s.cpp:5:3: note: candidate constructor not viable: requires 1 argument, but 0 were provided
  T(int){}
  ^
s.cpp:4:8: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 0 were provided
struct T {
       ^
s.cpp:4:8: note: candidate constructor (the implicit move constructor) not viable: requires 1 argument, but 0 were provided
1 error generated.

```

**Setup**: Clad built from the GitHub master branch at commit d7e5434c on Ubuntu 22.04 with Clang 11.

PetroZarytskyi:
This PR addresses two issues:
1) Constructor inits in ``RMV::DifferentiateVarDecl`` are differentiated at a point when ``_d_this`` argument is unknown. It is initialized with ``nullptr`` and is updated once the derivative is known. This goes against the regular logic of the reverse mode in Clad since the derivative is usually known before the visitation of the corresponding expr starts. Because of this,  ``ConstructorPullbackCallInfo`` was introduced to keep track of updating the derivative object. In this PR, the code in ``RMV::DifferentiateVarDecl`` is refactored so that the derivative is known before the differentiation of the constructor starts.
2) In ``RMV::DifferentiateVarDecl``, when an object adjoint is initialized with the original object (i.e. ``_d_x(x);``) to preserve the structure, the adjoint is first temporarily initialized with the default constructor. If the default constructor is deleted, an error is produced. Instead of the default constructor, we can use a placeholder ``*nullptr``. Fixes #800.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2561,13 +2561,10 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     if (isPointerType && VD->getInit() && isa<CXXNewExpr>(VD->getInit()))
       isInitializedByNewExpr = true;
 
-    ConstructorPullbackCallInfo constructorPullbackInfo;
-
     bool isConstructInit =
         VD->getInit() && isa<CXXConstructExpr>(VD->getInit()->IgnoreImplicit());
     const CXXRecordDecl* RD = VD->getType()->getAsCXXRecordDecl();
     bool isNonAggrClass = RD && !RD->isAggregate();
-    bool emptyInitListInit = isNonAggrClass;
 
     // VDDerivedInit now serves two purposes -- as the initial derivative value
     // or the size of the derivative array -- depending on the primal type.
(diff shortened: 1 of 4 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2641,24 +2651,24 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
-      if (VDDerived && !clad::utils::hasNonDifferentiableAttribute(VD)) {
+      if (VDDerived) {
         derivedE = BuildDeclRef(VDDerived);
         if (isInitializedByNewExpr)
           derivedE = BuildOp(UnaryOperatorKind::UO_Deref, derivedE);
       }
 
-      if (VD->getInit() && !isConstructInit)
+      if (VD->getInit()) {
+        m_TrackVarDeclConstructor = true;
```
vgvassilev, line 2668: How about sSaveAndRestore?

PetroZarytskyi: Fixed. It will work the same way because this bit is never recursive. But perhaps it makes sense as a proper way to handle a global state variable.

vgvassilev (approved):
Lgtm!

Issue #1125, opened 2024-10-28
Calling templated functions inside function doesn't work in backwards mode


guitargeek:
Reproducer as a ROOT macro:

```c++
#include <Math/CladDerivator.h>

template <class T>
inline double foo(double *params) { return params[0]; }
inline double bar(double *params) { return params[0]; }

double wrapper(double *params)
{
   return foo<int>(params); // doesn't work!
   // return bar(params); // works!
}

#pragma clad ON
void gradient_request() { clad::gradient(wrapper, "params"); }
#pragma clad OFF

// To run as a ROOT macro
void macro()
{
   std::vector<double> parametersVec{.5};
   std::vector<double> gradientVec(parametersVec.size());

   auto func = [&](std::span<double> params) { return wrapper(params.data()); };
   auto grad = [&](std::span<double> params, std::span<double> out) {
      return wrapper_grad(parametersVec.data(), out.data());
   };

   grad(parametersVec, gradientVec);

   auto numDiff = [&](int i) {
      const double eps = 1e-6;
      std::vector<double> p{parametersVec};
      p[i] = parametersVec[i] - eps;
      double funcValDown = func(p);
      p[i] = parametersVec[i] + eps;
      double funcValUp = func(p);
      return (funcValUp - funcValDown) / (2 * eps);
   };

   for (std::size_t i = 0; i < parametersVec.size(); ++i) {
      std::cout << i << ":" << std::endl;
      std::cout << "  numr : " << numDiff(i) << std::endl;
      std::cout << "  clad : " << gradientVec[i] << std::endl;
   }
}
```

Output:

```txt
Processing macro.C...
0:
  numr : 1
  clad : 0
```

I could work around this limitation in RooFit by avoiding template functions in the generated code, but at some point it becomes unavoidable if we want to use functors as parameters.

This follows up on the following issue:
  * https://github.com/vgvassilev/clad/issues/637

If this works, this would finally enable us to use more accurate routines for numeric integration that take function parameters, closing the gap between RooFit AD and regular RooFit.

parth-07:
I am unable to reproduce this issue. I tried to reproduce it outside the ROOT environment.

```cpp
#include "clad/Differentiator/Differentiator.h"
#include <iostream>
#define show(x) std::cout<< #x << ": " << x << "\n";

template <class T>
inline double foo(double *params) { return params[0]; }
inline double bar(double *params) { return params[0]; }

double wrapper(double *params)
{
[...]

PR #1277, opened 2025-03-15
Add information about function template specialization to `DerivativeBuilder::cloneFunction`

pmozil:
Template specialization was not taken into account for different functions.
This was an issue for functions like `std::get`.
Code like this:
```cpp
#include <iostream>
#include <tuple>

double fn(double u) {
    std::tuple<double, double> tmp{u, 2 * u};
    double a = std::get<0>(tmp);
    double b = std::get<1>(tmp);

    return a + b;
}

int main() {
    auto df = clad::differentiate(fn, "u");
}
```
Only used one derivative for function specializations, and thus tit resulted in such returns:
```
f(3)=9; df(3)=2
```

After the merge requests, the expected result is achieved, which is:
```
f(3)=9; df(3)=3
```

The changes to the codebase are restricted only to non-member functions, and the logic is limited to `DerivativeBuilder::cloneFunctions`, while the changes to tests are purely cosmetic, since now function specializations are generated as such:
```cpp
template<> double dn_pushforward<template_args...>(...)
```
as opposed to the previous
```cpp
double fn_pushforward(...)
```


vgvassilev:
@pmozil, that looks like progress. There is some information how to debug the bots: https://clad.readthedocs.io/en/latest/user/DevelopersDocumentation.html#debugging-github-runners

pmozil:
Thanks!

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -141,39 +147,46 @@ static void registerDerivative(FunctionDecl* dFD, Sema& S,
...
-              NewFTD = FTD;
-              break;
+
+          // Check if returnedFD is already associated with a template
+          if (!returnedFD->getDescribedFunctionTemplate() &&
+              !returnedFD->isFunctionTemplateSpecialization()) {
+
+            // Look for existing template in current context
+            FunctionTemplateDecl* ExistingFTD = nullptr;
+            DeclContext::lookup_result Lookup =
```
vgvassilev, line 157: Probably Sema::LookupTemplateName is a routine to look into.

Generally, I think we should construct a small program that has a primary template and a template specialization. Then we should build clang in debug mode and see what it does to find and attach the template specialization. Then we should replay the code sequence.

pmozil: Thanks for the feedback!
I've debugged clang, and it seems that clang does the same thing as I did.
It uses `setFunctionTemplateSpecialization` for instantiations as well.
The major difference was that it also used `clang::Sema::CheckFunctionTemplateSpecialization` to check for valid specialization, but this is validated anyways by clang, so I don't see a point in doing that here.

I did, however, find a better way to check whether template argument lists are the same.

pmozil:
Forgot to account for `isSameTemplateParamList` not existing, will revert to `isSameTemplateParam` rn

pmozil:
Since the prior versions of clang don't have `isSameTemplateParameter` either, need to revert back to original comparion of template parameters

vgvassilev:
> Since the prior versions of clang don't have `isSameTemplateParameter` either, need to revert back to original comparion of template parameters

I think you were on the right track. We should use these interfaces and find out what was used for this in older clang versions. Then we should add it to Compatibility.h

pmozil:
Ah thanks, didn't know about it.
Will do so then

include/clad/Differentiator/Compatibility.h
```diff
@@ -660,5 +660,133 @@ static inline const DeclSpec& Sema_ActOnStartOfLambdaDefinition_ScopeOrDeclSpec(
...
+  for (unsigned i = 0; i < FD1->getNumParams(); ++i) {
+    if (!Context.hasSameType(FD1->getParamDecl(i)->getType(),
+                             FD2->getParamDecl(i)->getType())) {
+      ParamsMatch = false;
+      break;
+    }
+  }
+
+  return ParamsMatch &&
+         Context.isSameTemplateParameterList(FTD1->getTemplateParameters(),
```
vgvassilev, line 786: Wouldn’t it make more sense if you cut and paste this function for versions earlier than clang 14? That way this routine can be implemented outside of this file.

pmozil: Oh, right, that's way better
Will also change that shortly

vgvassilev: Honestly I am not sure why we need the function. Doesn’t clang have functionality to do what you are trying to do? We must be having the same problems as the rest of the normal C++ code being written by users…

pmozil: Sorry, not sure which function you mean, do you mean `isSameTemplateParameterList`?
If so, it's just that it was not present in `ASTContext.h` before clang 12, and for osx, it seems, before clang 13,
So that's why i put a stand-in for it for earlier versions

vgvassilev: I think I can follow. We need to generate a primary template decl when we see a first specialization. That's probably something we can merge in once the bots are green. Then we should open an issue and a new pull request where when we see a primary template we insert a DiffRequest to differentiate the primary template and when we see a implicit instantiation we just take the primary template and instantiate it. This would be very close to what clang would do for regular code.

pmozil: You're right, it seems to me as well that this is the best way to do this.
If i understood correctly, you mean to differentiate a primary template, and instantiate the function derivative is it's an implicit instantiation. Then, in another issue, a problem of creating derivatives for explicit instantiations should be created, right?
If so, it wouldn't require the comparison of template parameter lists and i could revert those changes

pmozil: Great! I'll limit this PR to creating derivatives for implicit instantiations only, and then open another issue for explicit instantiations

vgvassilev:
@pmozil, can check which open issues this PR fixes?

pmozil:
Yeah, ofc, will do so in abt 1-2 hrs

pmozil:
Closes #1095 #1125 (not for template specialization cases)

vgvassilev:
> Closes #1095 #1125 (not for template specialization cases)

Are you sure these are completely fixed? We should open new issues for the parts that are not fixed...

pmozil:
Ah, understood.
Yeah the parts involving specialization still apply, so I'll open an issue that covers those parts rn

pmozil:
Actually, i checked for two cases with specialization and not just instantiation:
one being this:
```cpp
#include <iostream>
#include <tuple>


template<int>
double take(double a, double b) { return 0.0L; }
template<>
double take<0>(double a, double b) { return a; }
template<>
double take<1>(double a, double b) { return b; }

double fn(double u) {
    double a = take<0>(u, 2*u);
    double b = take<1>(u, 2*u);
    double c = take<2>(u, 2*u);

    return a + b;
}

int main() {
    auto df = clad::differentiate(fn, "u");
    df.dump();
}
```

And another this:
```cpp
#include <vector>
#include <span>
#include <iostream>

template <int i>
inline double foo(double *params) { return params[i]; }
template <>
inline double foo<1>(double *params) { return params[0]; }
inline double bar(double *params) { return params[0]; }

{
   return foo<1>(params); // doesn't work!
   // return bar(params); // works!
}

int main()
{
   auto fn = clad::gradient(wrapper, "params");
   std::vector<double> parametersVec{.5};

   auto func = [&](std::span<double> params) { return wrapper(params.data()); };
      return fn.execute(parametersVec.data(), out.data());
   };

   auto numDiff = [&](int i) {
      const double eps = 1e-6;
      double funcValDown = func(p);
      double funcValUp = func(p);
   };

   }
}
```

And both work as expected, meaning that they give proper derivatives. Seems that specialization is covered as well, at least for simple functions  - templates in templater or variadic templates may not still work.

vgvassilev (approved):
Lgtm!

PR #1279, opened 2025-03-16
Use memcpy instead of manually copying data, and add malloc check

### LINKED ISSUE #1278: Possible use of memcpy instead of manually doing it
While going through the code to solve the issue #1274, I found out the constructor https://github.com/vgvassilev/clad/blob/e7fd7c63f556b6fb54c07e9ca2fdad0667546ed7/include/clad/Differentiator/Differentiator.h#L277 does manual copy of bytes from one memory location to other. I believe this could be optimized by using memcpy instead of manually doing this. Also, I guess, we should check if the malloc call was successfull before doing this?
Please let me know, @parth-07 @vgvassilev if this looks legtimate and I will make a PR for this.

jayantpranjal0:
Closes #1278

### DIFF
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -283,10 +283,15 @@ CUDA_HOST_DEVICE T push(tape<T>& to, ArgsT... val) {
                            "you pass clad.so to clang.");
 #endif
       size_t length = GetLength(code);
-      char* temp = (char*)malloc(length + 1);
-      m_Code = temp;
-      while ((*temp++ = *code++))
-        ;
+      m_Code = (char*)malloc(length + 1);
+      if (m_Code)
+        memcpy((void*)m_Code, code, length + 1);
+      else
+#ifdef __CUDACC__
+        printf("stderr: Error: Failed to allocate memory for m_Code\n");
+#else
+        fprintf(stderr, "Error: Failed to allocate memory for m_Code\n");
+#endif
     }
 #endif
 


include/clad/Differentiator/Differentiator.h
```diff
@@ -283,10 +283,13 @@ CUDA_HOST_DEVICE T push(tape<T>& to, ArgsT... val) {
...
-      char* temp = (char*)malloc(length + 1);
-      m_Code = temp;
-      while ((*temp++ = *code++))
-        ;
+      m_Code = (char*)malloc(length+1);
+      if(m_Code) {
+        memcpy((void*)m_Code, code, length+1);
+      }
+      else {
+        printf("Failed to allocate memory for m_Code\n");
```
parth-07, line 291: Perhaps it should be reported with the prefix 'error:' and go in the error stream instead of the output stream.

jayantpranjal0:
@parth-07 I have updated the code with your suggestions.

parth-07 (approved):
Looks good.

include/clad/Differentiator/Differentiator.h
vgvassilev: I think Parth meant `fprintf(stderr...`.

jayantpranjal0: Updated

include/clad/Differentiator/Differentiator.h
```diff
@@ -283,10 +283,11 @@ CUDA_HOST_DEVICE T push(tape<T>& to, ArgsT... val) {
...
       size_t length = GetLength(code);
-      char* temp = (char*)malloc(length + 1);
-      m_Code = temp;
-      while ((*temp++ = *code++))
-        ;
+      m_Code = (char*)malloc(length + 1);
+      if (m_Code)
+        memcpy((void*)m_Code, code, length + 1);
+      else
+        fprintf(stderr, "Error: Failed to allocate memory for m_Code\n");
```
kchristin22, line 290: Hi! `fprintf` and `stderr` can only be compiled for a host (CPU), so you need to differentiate the execution path for device (GPU) and host (CPU). There are examples like that inside `Differentiator.h`. You can use `printf` for device code : )

vgvassilev: @jayantpranjal0, any news on this?

jayantpranjal0: Sorry, I was out for few days. I will made a PR by tomorrow.

jayantpranjal0: I have pushed some changes. Also, I was thinking of maybe creating a logging system, which would handle these by default, or also have some more explicitly defined functions for different use cases. What are your opinions @vgvassilev @kchristin22 @parth-07 ?

vgvassilev (approved):
Lgtm!

jayantpranjal0:
@vgvassilev Should I also create a issue, to create a class or some other utility for uniform logging? 
I can take it up.

vgvassilev:
> @vgvassilev Should I also create a issue, to create a class or some other utility for uniform logging? I can take it up.

I do not think that should be a priority. For example, we want to keep the Differentiator.h simple without including a lot of headers, etc.

PR #1286, opened 2025-03-18
Automatically generate constructor pullbacks

### LINKED ISSUE #1044: Adjoints of class-type arguments are not updated in function call differentiation
Adjoints of class-type arguments are not updated in function call differentiation.

Reproducible example:

```cpp
class A {
public:
  double data = 0;
};

double add(A a, double u) {
    return a.data + u;
}

    A a;
    a.data = v;
    double res = 0;
    res = add(a, u);
    return res;
}

int main() {
}
```

The derivative code for `res = add(a, u)` is:

```
 {
        res = _t1;
        double _r_d1 = _d_res;
        _d_res = 0;
        A _r0 = {};
        double _r1 = 0;
        add_pullback(a, u, _r_d1, &_r0, &_r1);
        *_d_u += _r1;
    }
```

Please note that this code is not updating `_d_a` with the temporary gradient `_r0`.

PetroZarytskyi:
This PR introduces the automatic generation of constructor pullbacks when custom ones are not provided. This is especially useful for differentiating implicit copy/move constructors because it's not obvious to the user that a custom constructor pullback must be provided (see #1044).

Fixes #1044

This is still a draft because there are design choices that need to be discussed. For example, how exactly to differentiate `this` object.


test/Gradient/UserDefinedTypes.C
```diff
@@ -1017,10 +1064,49 @@ int main() {
...
+// CHECK-NEXT:        SimpleFunctions1::constructor_pullback(this->x * rhs.x, this->y * rhs.y, &_d_y, &_r0, &_r1);
 // CHECK-NEXT:        (*_d_this).x += _r0 * rhs.x;
 // CHECK-NEXT:        (*_d_rhs).x += this->x * _r0;
 // CHECK-NEXT:        (*_d_this).y += _r1 * rhs.y;
 // CHECK-NEXT:        (*_d_rhs).y += this->y * _r1;
 // CHECK-NEXT:    }
 // CHECK-NEXT:}
+
+// CHECK:  static void constructor_pullback(double px, SimpleFunctions1 *_d_this, double *_d_px) {
+// CHECK-NEXT:      char _temp0[16] = {0};
```
parth-07, line 1076: It is cool to use a char array to avoid the constructor call here!

However, what happens if the struct/class alignment requirement is different from a char array? For example, a struct has an alignment requirement of 8, and a char array might have an alignment requirement of 1.

parth-07: We will not have this alignment issue if we use `malloc`, as you have used in the previous commits of this PR. But explicitly calling `free` is not much fun either...

vgvassilev: We can use `unsigned char alignas(T) tmp[ sizeof(T) ];` but this is still UB :)

PetroZarytskyi: Going to mark this resolved since we settled on the malloc approach

PetroZarytskyi:
The commits about try stmts and lambdas fix crashes that arise from unsupported features. We attempt to differentiate them because we now differentiate all constructors, including the ones added in #1288 just to make sure they aren't differentiated.

include/clad/Differentiator/KokkosBuiltins.h
```diff
@@ -50,19 +50,15 @@ constructor_reverse_forw(
...
-                          ::Kokkos::View<DataType, ViewParams...>* d_this,
-                          const ::std::string* /*d_name*/,
-                          const size_t& /*d_idx0*/, const size_t* /*d_idx1*/,
-                          const size_t* /*d_idx2*/, const size_t* /*d_idx3*/,
-                          const size_t* /*d_idx4*/, const size_t* /*d_idx5*/,
-                          const size_t* /*d_idx6*/, const size_t* /*d_idx7*/) {}
-
+template <class DataType, size_t N, class... ViewParams>
+void constructor_pullback(
+    const char (&name)[N], const size_t idx0, const size_t idx1,
```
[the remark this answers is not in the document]
PetroZarytskyi: This is the way they defined it in kokkos

PR #1288, opened 2025-03-18
Generate pullbacks of linear constructors

PetroZarytskyi:
This PR enables differentiation of all constructors that are either
1) Generated by the compiler
2) Have an empty body and all inits equal to either constants or ``DeclRefExpr``

This is a simplified version of a bigger work covering all constructors.


test/Gradient/UserDefinedTypes.C
```diff
@@ -793,6 +781,60 @@ double fn23(double x){
 // CHECK-NEXT:      *_d_x += 1;
 // CHECK-NEXT:  }
 
+class B {
+public:
+  double data = 0;
+};
```
vgvassilev, line 787: Can we add:
```cpp
struct S1{
double d;
S() : d([](){return 12.;}()) {}
};
struct S2{
double i;
double d;
S() : i(1.), d([](){i *= 32; return 12.;}()) {}
};
```

vgvassilev: Yes, and we will have a `CHECK-NOT:` clause.

vgvassilev: > And the builtin CXXConstructorDecl::isTrivial() seems to be only applied to compiler-generated constructors.

https://clang.llvm.org/doxygen/classclang_1_1Sema.html#af8eede5ad60ca466d9dc9abd594a58ee

include/clad/Differentiator/STLBuiltins.h
```diff
@@ -627,6 +627,9 @@ constructor_reverse_forw(clad::ConstructorReverseForwTag<::std::unique_ptr<T>>,
   return {::std::unique_ptr<T>(p), ::std::unique_ptr<T>(d_p)};
 }
 
+template <typename T>
+void constructor_pullback(T* p, ::std::unique_ptr<T>* dthis, T* dp) noexcept {};
```
vgvassilev, line 631: Why do we need this?

lib/Differentiator/CladUtils.cpp
```diff
@@ -307,6 +307,29 @@ namespace clad {
...
+                             const clang::ASTContext& C) {
+      // Trivial constructors are linear
+      if (CD->isTrivial())
+        return true;
+      // If the body is not empty, the constructor is not considered linear
+      if (auto* CS = dyn_cast<CompoundStmt>(CD->getBody())) {
+        if (!CS->body_empty())
+          return false;
+      } else
+        return false;
```
vgvassilev, line 320:
```suggestion
      if (!isa_or_null<CompoundStmt>(CD->getBody())
        return false;
      auto* CS = cast<CompoundStmt>(CD->getBody());
      if (!CS->body_empty())
          return false;
```

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -16,6 +16,7 @@
 #include "clang/AST/RecursiveASTVisitor.h"
 #include "clang/AST/StmtVisitor.h"
 #include "clang/Sema/Sema.h"
+#include <clang/AST/DeclCXX.h>
```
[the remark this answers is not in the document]
PetroZarytskyi: Done.

lib/Differentiator/CladUtils.cpp
```diff
@@ -307,6 +307,27 @@ namespace clad {
       return QT->isArrayType() || QT->isPointerType();
     }
 
+    bool isLinearConstructor(const clang::CXXConstructorDecl* CD,
+                             const clang::ASTContext& C) {
+      // Trivial constructors are linear
+      if (CD->isTrivial())
+        return true;
+      // If the body is not empty, the constructor is not considered linear
+      auto* CS = cast<CompoundStmt>(CD->getBody());
```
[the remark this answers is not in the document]
PetroZarytskyi: Thank you for the example. The original version of the PR actually accounted for the body not being a compound stmt. I removed that part when I failed to find an example and had to fix the coverage. Anyway, I changed it back and added the test you provided.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -110,10 +110,12 @@ static void registerDerivative(FunctionDecl* dFD, Sema& S,
     TypeSourceInfo* TSI = m_Context.getTrivialTypeSourceInfo(functionType);
     if (isa<CXXMethodDecl>(FD)) {
       CXXRecordDecl* CXXRD = cast<CXXRecordDecl>(DC);
+      StorageClass SC = isa<CXXConstructorDecl>(FD)
+                            ? SC_Static
```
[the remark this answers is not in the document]
PetroZarytskyi: We need to make it static because otherwise, we would have to provide `this` object. It doesn't have to be provided for the actual constructor because it's a special case but the derivative is not a constructor. I added a comment.

lib/Differentiator/CladUtils.cpp
PetroZarytskyi: Done.

include/clad/Differentiator/STLBuiltins.h
PetroZarytskyi: You're right, we don't. I probably took it from the other PR by accident.

lib/Differentiator/DerivativeBuilder.cpp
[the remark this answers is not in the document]
vgvassilev: Can you add a comment explaining this?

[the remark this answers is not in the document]
PetroZarytskyi: Done.

Issue #1265, opened 2025-03-13
Define a pullback function as having an empty body when it's forward declared but not needed


kchristin22:
Pullback functions are forward declared, but Clad may find out that they're not needed for the gradient and thus not produce a function definition afterwards, leading to a compilation error if you include the derived code in a source file.

Example:
```
void fn2_pullback(...);
void fn1_grad(..){
    ...
    fn2_pullback(...);
    ...
}
```

Solution:
```
void fn2_pullback(...);
void fn1_grad(..){
    ...
    fn2_pullback(...);
    ...
}
void fn2_pullback(...){};
```

Or if possible:
```
void fn2_pullback(...){};
void fn1_grad(..){
    ...
    fn2_pullback(...);
    ...
}
```

PetroZarytskyi:
Hi @Danbaekman! Feel free to contribute to any issue by submitting a pull request.

parth-07:
>Define a pullback function as having an empty body when it's forward declared but not needed

Can we remove the unneeded pullback function calls instead of defining the pullback functions with empty body?

vgvassilev:
Fixed by 0bab89dc ("Do not generate calls to pullbacks with empty bodies") — the empty-body approach discussed here. Regression coverage updated in `test/Gradient/FunctionCalls.C` and `test/Gradient/STLCustomDerivatives.C`. Closing.

PR #1290, opened 2025-03-20
Fix issue #1265: Add empty function bodies for unused pullback functions

salmazekri:
**Problem**
When Clad generates gradient code, it forward declares pullback functions that may not be used in the final derivative computation. This leads to linker errors when the pullback functions are declared but not defined.

**Solution Implemented**
1-Added the ability to detect unused pullback functions and generate empty definitions for them.
2-Added new member functions to [DerivativeBuilder]
_bool IsUsedInGradient(const clang::FunctionDecl* FD);
void HandlePullbackDecl(clang::FunctionDecl* FD);_
3-The code checks if a pullback function is actually used in gradient computation, and if not, generates an empty definition.

**Changes made per file:**
1. DerivativeBuilder.h - Added `IsUsedInGradient()` and `HandlePullbackDecl()` method declarations to handle unused pullbacks
2. DerivativeBuilder.cpp - Added implementations of `IsUsedInGradient()` and `HandlePullbackDecl()` to detect and handle unused pullbacks
3. UnusedPullbacks.cpp - Added new test file to verify empty definitions are generated for unused pullback functions
4. CMakeLists.txt - Added UnusedPullbacks.h to CLAD_INCLUDE_FILES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -315,6 +315,9 @@ if (CLAD_INCLUDE_DOCS)
   add_subdirectory(docs)
 endif()
 
+set(CLAD_INCLUDE_FILES
+  include/clad/Differentiator/UnusedPullbacks.h
+)
 
 if (NOT CLAD_BUILD_STATIC_ONLY)
   include(AddClad)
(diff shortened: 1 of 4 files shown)


vgvassilev:
I’d think we should check if the body of a function is empty or not in the DiffPlanner.

salmazekri:
> I’d think we should check if the body of a function is empty or not in the DiffPlanner. @vgvassilev 

I modified DiffPlanner.cpp and tested with a testfile (positive), I put them both in a gist in the link below. If it is OK, I can proceed to PR

https://gist.github.com/salmazekri/9d434a67b98497ce41c4e383aac35c0e

vgvassilev:
Sure, whatever leads to a successful review of your PR.

vgvassilev:
This pull request changes the files textually and is not going in the right direction.

PR #1291, opened 2025-03-20
Fix unexpected attempted differentiation of non-differentiation function error

### LINKED ISSUE #1269: Unexpected attempted differentiation of non-differentiation function error
Reproducible example:

```cpp
__attribute__((annotate("non_differentiable")))
void do_nothing(double u, double v) {}

  do_nothing(u, v);
  return u + v; 
}

int main() {
  auto d_f1 = clad::gradient(fn);
}
```

Output:
```
 error: attempted differentiation of function 'do_nothing', which is marked as non-differentiable
  do_nothing(u, v);
```

In this example, Clad is requested to differentiate `fn` and this function internally calls a non-differentiable function `do_nothing`. Clad should not try to differentiate `do_nothing` as it is a non-differentiable function.

Errant404:
Fixes #1269.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -1040,6 +1040,12 @@ namespace clad {
       }
       m_TopMostReq = &request;
     } else {
+      // If the function contains annotation of non_differentiable, then Clad
+      // should not produce any derivative expression for that function call,
+      // and the function call in the primal should be used as it is.
+      if (clad::utils::hasNonDifferentiableAttribute(E))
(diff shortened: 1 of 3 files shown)


vgvassilev:
Please add a test case.

vgvassilev:
I think we are waiting on @parth-07 to take a look.

parth-07 (approved):
Looks good to me!

@Errant404 can you please squash the commits and improve the commit message.

vgvassilev:
Your commit message and PR should have Fixes: #issue_number as a last paragraph

Issue #1108, opened 2024-10-02
clad::gradient and clad::differentiate produce different results under certain numerical conditions


ZeptoStarling:
clad::gradient returns NaN values due to (unused) std::acos in the function definition receiving invalid input, while clad::differentiate calculates the derivatives. So this:
```
double f(double C, double A)
{
	double a = std::acos(-C / A); // valid input is -1 <= -C/A <= 1
	return 2 * C * C * A;
}

int main()
{
	auto f_grad = clad::gradient(f);
	double dC = 0, dA = 0;
	f_grad.execute(/*C=*/5, /*A=*/3, &dC, &dA);
	std::cout << "clad::gradient results: " << std::endl;
	std::cout << dC << " " << dA << std::endl;
	std::cout << "clad::differentiate results: " << std::endl;
	auto f_dC = clad::differentiate(f, "C");
	std::cout << f_dC.execute(/*C=*/5, /*A=*/3) << " ";
	auto f_dA = clad::differentiate(f, "A");
	std::cout << f_dA.execute(/*C=*/5, /*A=*/3) << std::endl;
}
```
on Clad version 1.8~dev produces these results:
```
clad::gradient results: 
nan -nan
clad::differentiate results: 
60 50
```

Shubhanjan-GGOD:
Hello, I have worked on this issue. I will raise a PR today.

Vedant2005goyal:
Hi
I found this issue to be an interesting one and decided to work on.
The reverse mode didn't gave the correct result because it used to store intermediate values in **TAPE** during forward pass and later plays the tape in reverse order during backward pass. The problem arises due to NAN value of acos(-5/3) being stored in tape during the forward pass and that messes everything during backward pass even if the return statement has nothing to do with the acos(-5/3). And that is the reason there is no issue with the forward mode since it does not use TAPE.
Code generated by the clad:

<img width="823" height="296" alt="Image" src="https://github.com/user-attachments/assets/bd6a3f8d-b752-49ba-b804-ce2f20f445f0" />

It shows that the derivative of acos is computed and also its primal value was also stored in TAPE. This needs to be fixed. For working on this i am deep diving in the code and just want to know if I should explore something specific in the code for fixing this or any other inputs that might help me.
Thanks :)

PR #1292, opened 2025-03-20
fix of gradient computation for unused acos #1108

Shubhanjan-GGOD:
#1108 Fixed Gradient Computation of Unused acos function.

Description -This PR improves Clad’s handling of acos in differentiation by ensuring it is skipped only when it does not contribute to the return computation of the parent function.

Annotated acos as non-differentiable so that Clad skips differentiation .

Introduced logic to override this behavior when acos is directly or indirectly used in the return computation of the parent function.

Updated VisitCallExpr to check if acos affects the function’s return value before deciding whether to differentiate it.

Key Changes-
1.Added isUsedInReturnComputation method-

Checks if a given CallExpr (function call) is used in the return computation of the parent function.

2.Added findReturnStmt method-

Finds the return statement of a given function to analyze its dependencies.

3.Added containsExpr method-

Determines whether a given expression is present inside a statement, used to track acos usage.

4.Updated VisitCallExpr in ReverseModeVisitor-

5Now, acos is skipped only if it does not contribute to the function’s return value.

Calls isUsedInReturnComputation to make an informed decision.

Files changed -
1.ReverseModeVisitor.h-

Added method declarations:

bool isUsedInReturnComputation(const clang::CallExpr* CE);

const clang::Stmt* findReturnStmt(const clang::FunctionDecl* FD);

bool containsExpr(const clang::Stmt* S, const clang::Expr* E);

2.ReverseModeVisitor.cpp-

->Implemented the newly declared methods:

isUsedInReturnComputation: Determines if acos is part of the return computation.

findReturnStmt: Retrieves the return statement of a function.

containsExpr: Checks if a specific expression is inside a statement.

->Modified VisitCallExpr to use isUsedInReturnComputation before deciding whether to skip differentiation of acos.

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -32,691 +32,681 @@ template <typename T> class SmallVectorImpl;
 }
 
 namespace clad {
-  class ErrorEstimationHandler;
-  class ExternalRMVSource;
-  class MultiplexExternalRMVSource;
+class ErrorEstimationHandler;
+class ExternalRMVSource;
+class MultiplexExternalRMVSource;
 
-  /// A visitor for processing the function code in reverse mode.
-  /// Used to compute derivatives by clad::gradient.
-  class ReverseModeVisitor
-      : public clang::ConstStmtVisitor<ReverseModeVisitor, StmtDiff>,
-        public VisitorBase {
-  protected:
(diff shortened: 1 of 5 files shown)


vgvassilev:
Can you undo the clang-format changes so that we are able to see your real code modifications?

vgvassilev:
Closing due to lack of activity.

PR #1296, opened 2025-03-22
Do not dereference _d_this by default

PetroZarytskyi:
In the reverse mode, we still consider the adjoint of `this` to be `(*_d_this)`. We have had this inconsistency since the time we didn't support pointers to avoid using them. Now, we can support pointers perfectly fine and there is no reason to introduce type inconsistency between the clone and the adjoint.


vgvassilev (approved):
This is awesome!

PR #1303, opened 2025-03-25
Remove const qualifier from pullback parameters

### LINKED ISSUE #1302: Error when differentiating a function of const object return type
When trying to differentiate
```
struct Vector3 {
    double x, y, z;
    Vector3(double px = 0, double py = 0, double pz = 0) : x(px), y(py), z(pz) {}
};

  const Vector3 operator*(double a, const Vector3& v) {
    return {a * v.x, a * v.y, a * v.z};
  }
  
    Vector3 v(x, x, y);
    Vector3 w = 2*v;
    return w.x;
  }
```
we get 
```
error: cannot initialize a parameter of type 'Vector3 *' with an rvalue of type 'const Vector3 *'
  const Vector3 operator*(double a, const Vector3& v) {
                ^
note: passing argument to parameter '_d_this' here
^
void operator_star_pullback(double a, const Vector3 &v, const Vector3 _d_y, double *_d_a, Vector3 *_d_v) {
    {
        double _r0 = 0.;
        double _r1 = 0.;
        double _r2 = 0.;
        *_d_a += _r0 * v.x;
        (*_d_v).x += a * _r0;
        *_d_a += _r1 * v.y;
        (*_d_v).y += a * _r1;
        *_d_a += _r2 * v.z;
        (*_d_v).z += a * _r2;
    }
}
...
```
This happens because the pullback argument ``_d_y`` in ``operator_star_pullback`` is const qualified. We usually remove the const qualifier for all adjoints but we don't do that with pullback parameters.

PetroZarytskyi:
Fixes #1302.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -4372,6 +4372,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
 
     QualType oRetTy = FD->getReturnType();
     QualType dRetTy = oRetTy.getNonReferenceType();
+    dRetTy = getNonConstType(dRetTy, m_Context, m_Sema);
 
     // FIXME: We ignore the pointer return type for pullbacks.
     bool HasRet = false;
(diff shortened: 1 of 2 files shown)


vgvassilev (approved):
Lgtm!

PR #1305, opened 2025-03-26
Consider const methods non-differentiable

PetroZarytskyi:
Currently, we consider a function non-differentiable if it doesn't have any pointer/reference params and `dfdx`, but it also cannot be a method/operator since that's an equivalent of having a reference-type parameter. However, we can lower this requirement for const methods.


vgvassilev (approved):
LGTM!

PR #1308, opened 2025-03-26
Consider calls with const reference params non-differentiable

PetroZarytskyi:
Currently, we consider a function non-differentiable if it doesn't have any pointer/reference params and dfdx. This PR lowers the requirement to non-const pointer/reference params.


vgvassilev (approved):
Lgtm!

PR #1309, opened 2025-03-26
Support pointer bases in member exprs in reverse mode

PetroZarytskyi:
* this PR is a draft because it's done on top of #1308


vgvassilev:
Looks like we have to rebase this.

PR #1312, opened 2025-03-26
Support non-void functions with no dfdx

### LINKED ISSUE #1304: Error when differentiating a non-void function with its return unused in the reverse mode
Currently, if a function call of non-void return type has no dfdx, we have a crash. An easy reproducer:
```
double g(double& x) {
  return x++;
}
  
  g(x);
  return x;
}

int main() {
  clad::gradient(f);
  return 0;
}
```
produces
```
void g_pullback(double &x, double _d_y, double *_d_x) {
    *_d_x += _d_y;
}
clangtest.cpp:14:5: error: too few arguments to function call, expected 3, have 2
    g(x);
    ^
note: 'g_pullback' declared here
void f_grad(double x, double y, double *_d_x, double *_d_y) {
    double _t0 = x;
    g(x);
    *_d_x += 1;
    x = _t0;
}
1 error generated.
```

PetroZarytskyi:
Fixes #1304.


vgvassilev (approved):
Lgtm!

PR #1319, opened 2025-03-27
Use static_cast when casting a call arg of a cuda device call to ParmVarDecl

kchristin22:


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1084,7 +1084,8 @@ namespace clad {
           // Try to match it against the global arguments
           Expr* ArgE = E->getArg(i)->IgnoreParens()->IgnoreParenCasts();
           if (const auto* DRE = dyn_cast<DeclRefExpr>(ArgE)) {
-            const auto* PVD = cast<ParmVarDecl>(DRE->getDecl());
+            const ParmVarDecl* PVD =
+                static_cast<const ParmVarDecl*>(DRE->getDecl());
```
vgvassilev, line 1088:
```suggestion
            const auto* PVD = cast<const ParmVarDecl*>(DRE->getDecl());
```

kchristin22: Why should we avoid `static_cast`? Otherwise I could create a new `ParmVarDecl` when `DRE->getDecl()` is not already one.

vgvassilev: I guess I do not understand what this PR is doing then.

kchristin22: when there's a device call inside a kernel, we use the call args as params for the device function instead of the original ones, so if one of the arguments passed to the device call is a local var of the kernel, its decl is of type VarDecl and clang can't convert it to a ParmVarDecl

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1084,7 +1084,7 @@ namespace clad {
           // Try to match it against the global arguments
           Expr* ArgE = E->getArg(i)->IgnoreParens()->IgnoreParenCasts();
           if (const auto* DRE = dyn_cast<DeclRefExpr>(ArgE)) {
-            const auto* PVD = cast<ParmVarDecl>(DRE->getDecl());
```
kchristin22, line 1087:
```
  auto* VD = DRE->getDecl();
  if (isa<ParmVarDecl>(VD)) { // check if it's a kernel param 
    const auto* PVD = cast<ParmVarDecl>(VD); // clang can cast 
    request.DVI.push_back(PVD);
    request.CUDAGlobalArgsIndexes.push_back(i); // we know we should use atomic ops here
  } else {
    const ParmVarDecl* PVD =
        utils::BuildParmVarDecl(m_Sema, m_Sema.CurContext,
                                VD->getIdentifier(), VD->getType()); // create a param from the VarDecl
    request.DVI.push_back(PVD);
  }
```
@vgvassilev do you prefer this approach?

vgvassilev: Yes, I failed to understand the other one

vgvassilev (approved):
Lgtm!

Issue #1074, opened 2024-09-03
Use `VisitorBase::getZeroInit()` or `ConstantFolder::synthesizeLiteral()`  for assignments and initializations of variables


kchristin22:
Currently not all variables initialized or assigned to a value use `VisitorBase::getZeroInit()` or `ConstantFolder::synthesizeLiteral()` accordingly. Hence, the code is mixed or these functions are not used at all, as the Forward Mode tests indicate.

Example:

```c++

#include "clad/Differentiator/Differentiator.h"

double fn(const double* arr, size_t n) {
  double sum = 0;
  for (size_t i=0; i < n; ++i) {
    sum += arr[i];
  }
  return sum;
}


int main() {
  auto grad = clad::gradient(fn);
  return 0;
}
```

The code produced is:
```c++
void fn_grad(const double *arr, size_t n, double *_d_arr, size_t *_d_n) {
    size_t _d_i = 0UL;
    size_t i = 0UL;
    clad::tape<double> _t1 = {};
    double _d_sum = 0.;
    double sum = 0; // should be 0.
    unsigned long _t0 = 0UL;
    for (i = 0; ; ++i) {
        {
            if (!(i < n))
                break;
        }
        _t0++;
        clad::push(_t1, sum);
        sum += arr[i];
    }
    _d_sum += 1;
    for (;; _t0--) {
        {
            if (!_t0)
                break;
        }
        --i;
        {
            sum = clad::pop(_t1);
            double _r_d0 = _d_sum;
            _d_arr[i] += _r_d0;
        }
    }
}
```

As the comment indicates, the var `sum` should be initialized as so:
```c++
    double sum = 0.; 
```

M-Israr-Ul-Haq:
Hi! I would like to take this issue if no one else is working on it.
Could you please assign it to me?

Also, Should the fix apply to all the variable initializations generated by the forward and reverse mode visitors, or only to specific cases surfaced by the forward-mode tests?

PR #1323, opened 2025-03-28
Remove short int suffix in constant creation

kchristin22:
Previously:
```
short _d_y = 0i16; // doesn't compile
```
Now:
```
short _d_y = 0; 
```

### DIFF
--- a/lib/Differentiator/ConstantFolder.cpp
+++ b/lib/Differentiator/ConstantFolder.cpp
@@ -162,6 +162,9 @@ namespace clad {
     } else if (QT->isIntegralType(C)) {
       if (QT->isAnyCharacterType())
         QT = C.IntTy;
+      if (auto* BT = dyn_cast<BuiltinType>(QT.getTypePtr()))
+        if (BT->getKind() == BuiltinType::Short)
+          QT = C.IntTy;
       llvm::APInt APVal(C.getIntWidth(QT), val,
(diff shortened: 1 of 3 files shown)


lib/Differentiator/ConstantFolder.cpp
```diff
@@ -162,6 +162,9 @@ namespace clad {
     } else if (QT->isIntegralType(C)) {
       if (QT->isAnyCharacterType())
         QT = C.IntTy;
+      if (const auto* BT = dyn_cast<BuiltinType>(QT.getTypePtr()))
+        if (BT->getKind() == BuiltinType::Short)
+          QT = C.IntTy;
```
vgvassilev, line 167: Why codecov is not happy here?

kchristin22: There's #1074

vgvassilev: I guess, it means that the test does not make the code to go via this codepath, which is expected as the test worked before?

kchristin22: `synthesizeLiteral` doesn't seem to be invoked at all in this file

vgvassilev (approved):
Apparently there is a difference between a short and short int.. weird. Lgtm!

Issue #1295, opened 2025-03-21
Diffrequest is created dynamically when the definition is after clad::differentiate


ovdiiuv:
reproducer:
```cpp
#include "clad/Differentiator/Differentiator.h"
extern "C" int printf(const char* fmt, ...);

double sq_defined_later(double);

double g(double x){
    return 1;
}

int main() {
  int x = 4;
  auto df = clad::differentiate(sq_defined_later, "x");
  return 0;
}

double sq_defined_later(double x) {
    return g(x);
}
```

gives us

```cpp
clad::ValueAndPushforward<double, double> g_pushforward(double x, double _d_x);
double sq_defined_later_darg0(double x) {
    double _d_x = 1;
    clad::ValueAndPushforward<double, double> _t0 = g_pushforward(x, _d_x);
    return _t0.pushforward;
}

=========== //output when calling BaseForwardModeVisitor::Derive
clad::ValueAndPushforward<double, double> g_pushforward(double x, double _d_x) {
    return {(double)1, (double)0};
}
```

vgvassilev:
I think we can solve this by moving the analysis part at the end of the TU.

PetroZarytskyi:
@vgvassilev 
> I think we can solve this by moving the analysis part at the end of the TU.

That's what I thought too. I think we can accumulate ``clad::differentiate/gradient`` calls at ``CladPlugin::HandleTopLevelDecl`` but run the ``DiffCollector`` at the end of the TU.

vgvassilev:
Fixed by #1325

PR #1325, opened 2025-03-28
Move Diffcollector to HandleTU

ovdiiuv:
Fixes:#1295

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -83,18 +83,16 @@ namespace clad {
           // Emit error diagnostics
           if (R.empty()) {
             const char diagFmt[] = "'%0' has no defined operator()";
-            auto diagId =
-                m_SemaRef.Diags.getCustomDiagID(DiagnosticsEngine::Level::Error,
-                                                diagFmt);
+            auto diagId = m_SemaRef.Diags.getCustomDiagID(
+                DiagnosticsEngine::Level::Error, diagFmt);
             m_SemaRef.Diag(m_BeginLoc, diagId) << RD->getName();
             return false;
           } else if (!R.isSingleResult()) {
             const char diagFmt[] =
                 "'%0' has multiple definitions of operator(). "
                 "Multiple definitions of call operators are not supported.";
-            auto diagId =
-                m_SemaRef.Diags.getCustomDiagID(DiagnosticsEngine::Level::Error,
-                                                diagFmt);
+            auto diagId = m_SemaRef.Diags.getCustomDiagID(
+                DiagnosticsEngine::Level::Error, diagFmt);
             m_SemaRef.Diag(m_BeginLoc, diagId) << RD->getName();
 
             // Emit diagnostics for candidate functions
@@ -1015,6 +1013,7 @@ namespace clad {
       // The root of the differentiation request graph should update the
       // CladFunction object with the generated call.
       request.CallUpdateRequired = true;
+      request.CallContext = E;
 
       request.Args = E->getArg(1);
       // FIXME: We should call UpdateDiffParamsInfo unconditionally, however,
@@ -1059,6 +1058,7 @@ namespace clad {
       request.VerboseDiags = false;
       request.EnableTBRAnalysis = m_TopMostReq->EnableTBRAnalysis;
       request.EnableVariedAnalysis = m_TopMostReq->EnableVariedAnalysis;
+      request.CallContext = E;
 
       // const auto* MD = dyn_cast<CXXMethodDecl>(FD);
       if (m_TopMostReq->CUDAGlobalArgsIndexes.empty()) {
@@ -1101,7 +1101,6 @@ namespace clad {
 
     if (isCallOperator(m_Sema.getASTContext(), request.Function))
       request.Functor = cast<CXXMethodDecl>(request.Function)->getParent();
-    request.CallContext = E;
     request.BaseFunctionName = utils::ComputeEffectiveFnName(request.Function);
 
     // FIXME: Here we copy all varied declarations down to the pullback, has to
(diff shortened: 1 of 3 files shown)


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -39,14 +39,19 @@ namespace clad {
   ///
   /// \param[in] call A clad differentiation function call expression
   /// \param SemaRef Reference to Sema
-  DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
+  DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef,
+                              bool emitdiags = true) {
```
vgvassilev, line 43:
```suggestion
                              bool Complain = true) {
```

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -191,11 +191,12 @@ struct DiffRequest {
     llvm::DenseSet<const clang::FunctionDecl*> m_Traversed;
 
     bool m_IsTraversingTopLevelDecl = true;
+    bool m_CollectOnlyConstExpr;
 
   public:
     DiffCollector(clang::DeclGroupRef DGR, DiffInterval& Interval,
                   clad::DynamicGraph<DiffRequest>& requestGraph, clang::Sema& S,
-                  RequestOptions& opts);
+                  RequestOptions& opts, bool collectconstexpr);
```
vgvassilev, line 199:
```suggestion
                  RequestOptions& opts, bool constexprOnly = false);
```

tools/ClangPlugin.cpp
```diff
@@ -125,19 +125,16 @@ namespace clad {
...
+#if CLANG_VERSION_MAJOR > 16
       Sema& S = m_CI.getSema();
-
       if (!m_DerivativeBuilder)
         m_DerivativeBuilder.reset(
             new DerivativeBuilder(S, *this, m_DFC, m_DiffRequestGraph));
-
       RequestOptions opts{};
       SetRequestOptions(opts);
       DiffCollector collector(DGR, CladEnabledRange, m_DiffRequestGraph, S,
```
vgvassilev, line 135: I think things will become much simpler if we check whether the `DGR` is a constexpr and collect in place.

tools/ClangPlugin.cpp
```diff
@@ -132,8 +132,15 @@ namespace clad {
             new DerivativeBuilder(S, *this, m_DFC, m_DiffRequestGraph));
       RequestOptions opts{};
       SetRequestOptions(opts);
-      DiffCollector collector(DGR, CladEnabledRange, m_DiffRequestGraph, S,
-                              opts, true);
+      if (DGR.isSingleDecl()) {
```
vgvassilev, line 135: We need something like:
for (Decl *D: DGR)
  if (not FunctionDecl)
    continue;
  ...

tools/ClangPlugin.cpp
```diff
@@ -478,6 +482,17 @@ namespace clad {
...
     void CladPlugin::HandleTranslationUnit(ASTContext& C) {
+      Sema& S = m_CI.getSema();
+      if (!m_DerivativeBuilder)
+        m_DerivativeBuilder.reset(
+            new DerivativeBuilder(S, *this, m_DFC, m_DiffRequestGraph));
+      RequestOptions opts{};
+      SetRequestOptions(opts);
+
+      for (auto DCI : m_DelayedCalls)
+        DiffCollector collector(DCI.m_DGR, CladEnabledRange, m_DiffRequestGraph,
```
vgvassilev, line 493: We should also make sure we skip constexpr things that we have already visited here.

tools/ClangPlugin.cpp
```diff
@@ -132,14 +132,16 @@ namespace clad {
...
+
+      for (Decl* D : DGR) {
+        if (!isa<FunctionDecl>(D))
+          continue;
+        FunctionDecl* FD = cast<FunctionDecl>(D);
+        if (!FD->isConstexpr())
+          continue;
+        DiffCollector collector(DGR, CladEnabledRange, m_DiffRequestGraph, S,
+                                opts);
+        break;
```
vgvassilev, line 144:
```suggestion
        if (FD->isConstexpr())
          DiffCollector collector(DGR, CladEnabledRange, m_DiffRequestGraph, S,
                                opts);
```

tools/ClangPlugin.cpp
```diff
@@ -490,8 +492,16 @@ namespace clad {
...
 
       for (auto DCI : m_DelayedCalls)
-        DiffCollector collector(DCI.m_DGR, CladEnabledRange, m_DiffRequestGraph,
-                                S, opts);
+        for (Decl* D : DCI.m_DGR) {
+          if (isa<FunctionDecl>(D)) {
+            FunctionDecl* FD = cast<FunctionDecl>(D);
+            if (FD->isConstexpr())
+              continue;
+          }
```
vgvassilev, line 500:
```suggestion
          if (const auto* FD = dyn_cast<FunctionDecl>(D))
            if (FD->isConstexpr())
              continue;
```

tools/ClangPlugin.cpp
```diff
@@ -125,19 +125,24 @@ namespace clad {
...
 
-#if CLANG_VERSION_MAJOR > 16
+      for (Decl* D : DGR) {
+        if (!isa<FunctionDecl>(D))
+          continue;
+        FunctionDecl* FD = cast<FunctionDecl>(D);
+        if (FD->isConstexpr())
+          DiffCollector collector(DGR, CladEnabledRange, m_DiffRequestGraph, S,
+                                  opts);
+        break;
```
vgvassilev, line 143: No break here.

tools/ClangPlugin.cpp
```diff
@@ -478,6 +483,23 @@ namespace clad {
...
+      SetRequestOptions(opts);
+
+      for (auto DCI : m_DelayedCalls)
+        for (Decl* D : DCI.m_DGR) {
+          if (const auto* FD = dyn_cast<FunctionDecl>(D))
+            if (FD->isConstexpr())
+              continue;
+          DiffCollector collector(DCI.m_DGR, CladEnabledRange,
+                                  m_DiffRequestGraph, S, opts);
+          break;
```
vgvassilev, line 500: Likewise

tools/ClangPlugin.cpp
ovdiiuv: shouldn't it be inside that if?

tools/ClangPlugin.cpp
ovdiiuv: why? we should traverse the whole group only once.

vgvassilev: The decl group is an array of decls. That is double x = 12, y; we want to visit both, right?

ovdiiuv: yes, we could either visit both or none.

vgvassilev: After a discussion offline we agreed to have a comment explaining why the break stmts are needed.

PR #1328, opened 2025-03-31
Reduce tape usage for objects

### LINKED ISSUE #1276: Incorrect derivative result because class type variable are not restored
Clad emits incorrect derivative result because class type variable are not restored.

Reproducible example:

```cpp

class A {
public:
  void increment() { data++; }
  void setData(double u) { data = u; }
  double data = 0;
};

  double res = 0;
  A a;
  a.setData(u);
  res += a.data * v;
  a.increment();
  return res;
}

int main() {
  double du = 0, dv = 0;
  d_fn.execute(3, 5, &du, &dv);
  show(du);
  show(dv);
}
```

**Output:**
```
du: 5
dv: 4
```

**Expected:**
```
du: 5
dv: 3
```

PetroZarytskyi:
We switched from the `store per use` to `store per change` system in reverse mode a while ago, i.e.
instead of
```
_t0 = x;
f(x);  // x passed by reference
...
f_pullback(_t0); // we use a different _t variable for every use
```
we do
```
_t0 = x;
f(x); // x passed by reference
...
x = _t0; // this way, we can reuse x many times per one change
f_pullback(x);
```
However, we haven't completed it for objects. Other parts of Clad expect original objects to be set to correct values, this inconsistency sometimes results in bugs like #1276. This PR implements the `store per change` system for method call bases.
Another big advantage of this approach is decreased tape usage, as we don't have to store bases of const methods of ones represented with temporary expressions. Also, in the future, we can extend TBR analysis and make it go into called functions so that we can remove even more stores.

Fixes #1276.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1502,7 +1502,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     // Determine the base of the call if any.
     const auto* MD = dyn_cast<CXXMethodDecl>(FD);
     const Expr* baseOriginalE = nullptr;
-    if (MD && MD->isInstance()) {
+    if (MD && MD->isInstance() && !isLambdaCallOperator(MD)) {
       if (const auto* MCE = dyn_cast<CXXMemberCallExpr>(CE))
         baseOriginalE = MCE->getImplicitObjectArgument();
       else if (const auto* OCE = dyn_cast<CXXOperatorCallExpr>(CE))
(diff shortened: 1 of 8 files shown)


include/clad/Differentiator/STLBuiltins.h
```diff
@@ -409,14 +409,14 @@ void push_back_reverse_forw(::std::vector<T>* v, U val, ::std::vector<T>* d_v,
 }
 
 template <typename T, typename U, typename pU>
-void push_back_pullback(::std::vector<T>* v, U val, ::std::vector<T>* d_v,
+void push_back_pullback(const ::std::vector<T>* v, U val, ::std::vector<T>* d_v,
```
vgvassilev, line 412: Can we split this into a separate PR with also diagnostic?

PetroZarytskyi: Fixed.

PR #1332, opened 2025-04-03
Take const base in custom pullbacks of const methods

PetroZarytskyi:
We often provide custom STL pullbacks with non-const base arguments while the original method is const. This may lead overload mismatch and Clad attempting to differentiate these methods.
Here's a simple example:
```
retTy method(...) const {
...
```
should be differentiated as
```
retTy method_pullback(const classTy* a, ...) {
...
```
instead of
```
retTy method_pullback(classTy* a, ...) {
...
```


PetroZarytskyi:
This PR takes some changes from #1328.

vgvassilev (approved):
LGTM!

PR #1333, opened 2025-04-03
Update google test to its latest release

vgvassilev:


mcbarton:
@vgvassilev After 12.1 Google Test required a minimum of C++14 (see https://github.com/google/googletest/releases). Given Clad has some old versions of llvm it builds against then I don't think the upgrade is possible for Clad anytime soon.

PR #1337, opened 2025-04-07
`DerivativeBuilde::cloneFunction` Speficy from which function the template was instantiated

pmozil:
In llvm-20, the previous flow for cloning templated functions failed - ` getTemplateInstantiationPattern()` went into infinite recursion.
This change mends that


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -194,12 +194,16 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
             }
 
             // Now specialize the function correctly
-            TemplateArgumentList* TALCopy =
-                TemplateArgumentList::CreateCopy(m_Context, TAL->asArray());
-
-            returnedFD->setFunctionTemplateSpecialization(
-                ExistingFTD, TALCopy, nullptr,
-                FD->getTemplateSpecializationKindForInstantiation());
+            if (FD->getDescribedFunctionTemplate() != nullptr) {
```
vgvassilev, line 197:
```suggestion
            if (FD->getDescribedFunctionTemplate()) {
```

pmozil:
@vgvassilev seems that creating a primary template and then specializing is really the only way to go through with this.
Currently implemented basic primary template creation, will change if tests don't pass

vgvassilev:
> @vgvassilev seems that creating a primary template and then specializing is really the only way to go through with this. Currently implemented basic primary template creation, will change if tests don't pass

Any progress on this? Seems to be holding off our llvm20 support (#1267).

PR #1351, opened 2025-04-12
Don't generate += statements of non-real types in reverse mode

PetroZarytskyi:
When visiting a decl ref, we increase its adjoint with ``dfdx()``, e.g. ``x = y`` -> ``_d_y += _d_x``. This operation only makes sense to real types.


vgvassilev (approved):
Lgtm! Thank you!

PR #1355, opened 2025-04-14
Schedule member functions statically

PetroZarytskyi:


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -937,27 +937,29 @@ namespace clad {
...
-    LookupResult Found(S, DNInfo, Sema::LookupOrdinaryName);
-    S.LookupQualifiedName(Found, DC);
-
-    return !Found.empty();
+    llvm::SmallVector<std::string, 1> Names;
+    std::string oName = R.BaseFunctionName;
+    if (R.Mode != DiffMode::experimental_pullback || R->getNumParams() == 1)
+      Names.push_back(oName + "_pushforward");
+    if (R.Mode == DiffMode::experimental_pullback)
+      Names.push_back(oName + "_pullback");
```
vgvassilev, line 952:
```suggestion
    std::string Name;
    std::string oName = ;
    if (R.Mode != DiffMode::experimental_pullback || R->getNumParams() == 1)
      Name = R.BaseFunctionName + "_pushforward";
    if (R.Mode == DiffMode::experimental_pullback)
      Name = R.BaseFunctionName  + "_pullback");
```

vgvassilev: Not sure why that changed actually.

PetroZarytskyi: When we differentiate a nested function call in the reverse mode and the function has only one parameter, we first look for its pushforward (because the forward mode is more efficient in that case). If not found, we look for a pullback. Doing a lookup of just one name means we ignore potential custom pullbacks.

PetroZarytskyi: However, I have now realized that we also currently check that the function is not a method decl and has no pointer/reference parameters. This will fix the test that got broken. I can remove it for now if it's better.

vgvassilev: I am confused. If we don't need it probably is better to keep the diff as small as possible.

PetroZarytskyi: I have changed this bit. However, we'll probably have to fix this in some form later. Does it look better?

vgvassilev (approved):
LGTM!

PR #1357, opened 2025-04-16
Add Windows arm job to ci

mcbarton:
According to https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/ Windows arm Github runners are now available for free for public repositories. Therefore I have added them to Clads ci. Hopefully they build first time like Windows x86, but won't know until the workflow runs. Just adding for llvm 20 for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,11 @@ jobs:
             compiler: msvc
             clang-runtime: '20'
 
+          - name: win11-arm-msvc-runtime20
+            os: windows-11-arm
+            compiler: msvc
+            clang-runtime: '20'
+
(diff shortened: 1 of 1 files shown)


mcbarton:
The failed job has nothing to do with my changes. There was a failure by one of the Ubuntu jobs, while installing llvm 12.

mcbarton:
Not entirely sure this line https://github.com/vgvassilev/clad/blob/fb3b8960e6a41aae1e895d406dc32baa4a076ca2/.github/workflows/ci.yml#L731 will work as expected for Windows arm, given Conda doesn't support Windows on arm at the moment. Will need to investigate further.

vgvassilev:
Is that good to go?

PR #1358, opened 2025-04-16
Make visitor destructors virtual

PetroZarytskyi:


vgvassilev (approved):
LGTM!

include/clad/Differentiator/VisitorBase.h
```diff
@@ -694,10 +694,14 @@ namespace clad {
...
                                    clang::Expr*& derivedR);
 
+    virtual ~VisitorBase() = 0;
+
   private:
     clang::TemplateDecl* m_CladConstructorPushforwardTag = nullptr;
     clang::TemplateDecl* m_CladConstructorReverseForwTag = nullptr;
   };
+
+  VisitorBase::~VisitorBase() = default;
```
vgvassilev, line 704: This needs to be pinned in a source file. That is probably in DerivativeBuilder.cpp. We should probably remote the `=0`, too...

vgvassilev (approved):
Lgtm!

PR #1361, opened 2025-04-19
Don't account for output parameter in indepVarCount

ovdiiuv:
Before this PR, we would assume that ouput parameters are inputs as well and hence create larger `one_hot_vector` arrays and `identity_matrix` matricies than needed. In this PR we assume all parameters with `_clad_out_` prefix are the output parameters. For example, the function

```cpp
void f_10(float a, double _clad_out_output[]){
  _clad_out_output[0]= a * a;  
  _clad_out_output[1] = _clad_out_output[0];
}
```
produced
```cpp
void f_10_jac(float a, double _clad_out_output[], clad::matrix<double> *_d_vector__clad_out_output) {
    unsigned long indepVarCount = _d_vector__clad_out_output->rows() + 1UL;
                                                ...
```
and after

```cpp
void f_10_jac(float a, double _clad_out_output[], clad::matrix<double> *_d_vector__clad_out_output) {
    unsigned long indepVarCount = 1UL;
                                                ...
```

### DIFF
--- a/lib/Differentiator/JacobianModeVisitor.cpp
+++ b/lib/Differentiator/JacobianModeVisitor.cpp
@@ -67,7 +67,11 @@ DerivativeAndOverload JacobianModeVisitor::Derive() {
   llvm::SmallVector<ParmVarDecl*, 16> params;
   llvm::SmallVector<ParmVarDecl*, 16> derivedParams;
   llvm::SmallVector<DeclStmt*, 8> adjointDecls;
-  for (const auto* PVD : FD->parameters()) {
+
+  auto origParams = FD->parameters();
+  for (size_t i = 0, e = origParams.size(); i < e; ++i) {
+    const ParmVarDecl* PVD = origParams[i];
+
     IdentifierInfo* PVDII = PVD->getIdentifier();
     auto* newPVD = CloneParmVarDecl(PVD, PVDII,
                                     /*pushOnScopeChains=*/true,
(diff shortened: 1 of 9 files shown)


lib/Differentiator/JacobianModeVisitor.cpp
```diff
@@ -67,7 +67,12 @@ DerivativeAndOverload JacobianModeVisitor::Derive() {
   llvm::SmallVector<ParmVarDecl*, 16> params;
   llvm::SmallVector<ParmVarDecl*, 16> derivedParams;
   llvm::SmallVector<DeclStmt*, 8> adjointDecls;
-  for (const auto* PVD : FD->parameters()) {
+
+  auto origParams = FD->parameters();
+  for (size_t i = 0; i < origParams.size(); ++i) {
```
vgvassilev, line 72:
```suggestion
  for (size_t i = 0; e = origParams.size(); i < e; ++i) {
```

lib/Differentiator/JacobianModeVisitor.cpp
```diff
@@ -89,11 +94,13 @@ DerivativeAndOverload JacobianModeVisitor::Derive() {
       derivedExpr = utils::BuildParenExpr(m_Sema, derivedExpr);
       Expr* getSize = BuildCallExprToMemFn(BuildDeclRef(derivedPVD),
                                            /*MemberFunctionName=*/"rows", {});
-      if (!m_IndVarCountExpr)
-        m_IndVarCountExpr = getSize;
-      else
-        m_IndVarCountExpr =
-            BuildOp(BinaryOperatorKind::BO_Add, m_IndVarCountExpr, getSize);
+      if (!isLast) {
```
vgvassilev, line 97:
```suggestion
      if (i != e - 1) {
```

lib/Differentiator/JacobianModeVisitor.cpp
```diff
@@ -89,11 +94,13 @@ DerivativeAndOverload JacobianModeVisitor::Derive() {
...
+        if (!m_IndVarCountExpr)
```
vgvassilev, line 98: Under what conditions we may not have `m_IndVarCountExpr`?

ovdiiuv: On the first iteration, if the parameter is an array.

vgvassilev:
@PetroZarytskyi, can you take a look at this PR?

test/Jacobian/Pointers.C
```diff
@@ -1,28 +1,28 @@
-// RUN: %cladclang -Xclang -plugin-arg-clad -Xclang -disable-tbr %s -I%S/../../include -oPointers.out 2>&1 | %filecheck %s
-// RUN: ./Pointers.out | %filecheck_exec %s
-// RUN: %cladclang %s -I%S/../../include -oPointers.out
-// RUN: ./Pointers.out | %filecheck_exec %s
+// RUN: %cladclang -Xclang -plugin-arg-clad -Xclang -disable-tbr %s -I%S/../../include -oPointers._clad_out_out 2>&1 | %filecheck %s
+// RUN: ./Pointers._clad_out_out | %filecheck_exec %s
+// RUN: %cladclang %s -I%S/../../include -oPointers._clad_out_out
+// RUN: ./Pointers._clad_out_out | %filecheck_exec %s
```
PetroZarytskyi, line 4: I suppose this was accidental.

PetroZarytskyi (approved):
I was thinking about other solutions to this problem, like using additional parameters of ``clad::jacobian``, e.g.
```
double f(double x, double y, double z, double* output);

// give all non-output parameters
clad::jacobian(f, "x, y, z");

// or add functionality for output parameters
clad::jacobian(f, "", "output");
```
However, option 1 would be a nightmare with many parameters, and option 2 is simply ugly. So really, I agree the proposed solution involving parameter naming is the best here.
LGTM!

PR #1362, opened 2025-04-21
Enable custom derivative lookups on the DiffPlanner stage

PetroZarytskyi:
Currently, in DiffPlanner, we check if a given function has a custom derivative based on the name, i.e. we consider that ``f`` has a custom derivative if a function with the name ``f_pullback`` exists. Since names are not unique, this is insufficient and results in a fallback to dynamic derivative scheduling when names overlap. This is noticeable when we deal with operators and constructors, as they don't have unique names in Clad. This PR implements a full lookup that compares the signatures of derivatives (including templates).
Something we need to take into account:
Before this PR, we called custom derivatives every time it worked with a given set of arguments. Now, however, we only call custom derivatives if there is an exact match of parameter types. For example:
```
double g(double x) {
    return x;
}
// g_pushforward will NOT be considered a derivative for g, even if a given set of arguments can be converted to float
clad::ValueAndPushforward<float, float> g_pushforward(float x, float dx) {
    return {x, dx};
}
```
This could be both an advantage and a disadvantage. On the one hand, the lookup is more picky, making writing custom derivatives harder. On the other hand, this brings clarity to handling overloads since we require direct correspondence between functions and derivatives.

Another huge advantage is that currently, we handle custom derivative lookups separately for every generated function type (in all ``Derive``, ``VisitCallExpr``, and ``VisitCXXConstructExpr``). After this PR, lookups will be handled on a higher level, which will allow us to remove all of that logic.

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -23,6 +23,7 @@
 #include "clang/Sema/Sema.h"
 #include "clang/Sema/SemaInternal.h"
 #include "clang/Sema/Template.h"
+#include <clad/Differentiator/DiffMode.h>
 #include "clad/Differentiator/BaseForwardModeVisitor.h"
 #include "clad/Differentiator/CladUtils.h"
 #include "clad/Differentiator/DiffPlanner.h"
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -26,6 +26,7 @@
 #include "clad/Differentiator/CladConfig.h"
 #include "clad/Differentiator/CladUtils.h"
 #include "clad/Differentiator/Compatibility.h"
+#include "clad/Differentiator/DerivativeBuilder.h"
 
 #include <algorithm>
 #include <string>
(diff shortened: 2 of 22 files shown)


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -276,9 +266,10 @@ namespace clad {
 
   DiffCollector::DiffCollector(DeclGroupRef DGR, DiffInterval& Interval,
                                clad::DynamicGraph<DiffRequest>& requestGraph,
-                               clang::Sema& S, RequestOptions& opts)
+                               clang::Sema& S, RequestOptions& opts,
+                               DerivativeBuilder* builder)
```
vgvassilev, line 270: This seems to be breaking layering. Why do we need the DerivativeBuilder here?

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -293,6 +295,94 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
...
+    clang::sema::TemplateDeductionInfo Info(noLoc);
+    clang::TemplateArgumentListInfo ExplicitTemplateArgs;
+    S.DeduceTemplateArguments(Template, &ExplicitTemplateArgs, DerivativeType,
+                              Specialization, Info);
+    return Specialization;
+  }
+
+  bool
+  DerivativeBuilder::LookupCustomDerivativeDecl(const DiffRequest& request) {
+    auto DFI = m_DFC.Find(request);
```
vgvassilev, line 311: We can probably move `m_DFC` in the diff planner and drop the dependency on the derivative builder.

lib/Differentiator/DiffPlanner.cpp
PetroZarytskyi: I mentioned this when first discussing my work at a weekly meeting. ``DerivativeBuilder`` seemed like the best place to keep the lookup in. All of the existing lookup functions are in there. At first, I wanted to move it all to ``DiffPlanner`` but the lookup requires derivative types and names. Even after simplifying the system, it utilizes mode-dependent virtual visitor methods (e.g., to build adjoint parameter types) and many utility functions. For instance, to build the derivative type, we often need ``clad::ValueAndAdjoint``, ``clad::array``, and others, all of which are looked up and built in visitors.
I'm not against moving the lookup to ``DiffPlanner`` but this is likely to involve moving a lot of ``VisitorBase`` functions there too, including the ones that are now virtual. Do we still want to do this? Or do you see a better solution?

lib/Differentiator/DerivativeBuilder.cpp
PetroZarytskyi: ``m_DFC`` is part of the ``CladPlugin``, ``DerivativeBuilder`` only has a reference to it. We could also pass it to DiffPlanner, but I guess this is not the main problem.

lib/Differentiator/DiffPlanner.cpp
vgvassilev: Okay, I see. Can we transform this into a FIXME and an issue? I think we should be able to simplify that logic in the visitors and the virtual functions, too.

PetroZarytskyi:
@vgvassilev
> This seems to be breaking layering. Why do we need the DerivativeBuilder here?

I moved all the necessary functionality from VisitorBase to CladUtils and the lookup to the DiffPlanner. If it looks good now, I'll squash commits in a meaningful way.

vgvassilev:
> @vgvassilev
> 
> > This seems to be breaking layering. Why do we need the DerivativeBuilder here?
> 
> I moved all the necessary functionality from VisitorBase to CladUtils and the lookup to the DiffPlanner. If it looks good now, I'll squash commits in a meaningful way.

Let me know when you are done. This mostly looks good.

PetroZarytskyi:
@vgvassilev 
I squashed all necessary commits and removed the ones that were not relevant to the new implementation.

vgvassilev (approved):
Great job! LGTM!

PR #1370, opened 2025-05-03
Make static-only scheduling default for pullbacks

PetroZarytskyi:
The goal of this PR is to mostly remove dynamic pullback scheduling and dynamic custom derivative lookups. There are 3 temporary exceptions that were too complex to address in this PR:
1) Error estimation. It currently uses singleton objects m_ErrorEstHandler and m_EstModel, which are cleared after each error_estimate request. This requires the pullback to be derived at the same time to access the singleton objects.
2) Hessians. All high-order derivatives are still scheduled dynamically.
3) Calls that have non-differentiable arguments. This only happens when the user doesn't request derivatives with respect to certain const pointer parameters. Such information is only deduced during visitation. The only way to know it in the diffplanner is with activity analysis. Once activity analysis is enabled by default, we can deprecate this feature with activity analysis off.

The second commit of this PR makes pushforward scheduling in RMV more consistent. Currently, when dealing with simple single-argument functions in the reverse mode, we look up custom pushforwards but generate pullbacks. In this PR, we also start generating pushforwards in such cases. The primary goal is to make the generation/lookup of such derivatives fully static.

### DIFF
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -426,6 +426,14 @@ namespace clad {
     clang::QualType GetNonConstValueType(clang::QualType T) {
       QualType valueType = GetValueType(T);
       valueType.removeLocalConst();
+      // If the const-ness of the type is hidden with sugar, e.g.
+      // `class_name<double>::const_value_type`, the approach above
+      // does not work and we have to desugar the type explicitly.
+      QualType canonicalType = valueType.getCanonicalType();
+      if (canonicalType.isConstQualified()) {
+        canonicalType.removeLocalConst();
+        return canonicalType;
+      }
       return valueType;
     }
 
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -938,9 +938,11 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
     clang::FunctionDecl* Specialization = nullptr;
     clang::sema::TemplateDeductionInfo Info(noLoc);
     clang::TemplateArgumentListInfo ExplicitTemplateArgs;
-    S.DeduceTemplateArguments(Template, &ExplicitTemplateArgs, DerivativeType,
-                              Specialization, Info);
-    return Specialization;
+    auto R = S.DeduceTemplateArguments(Template, &ExplicitTemplateArgs,
+                                       DerivativeType, Specialization, Info);
+    if (R == clad_compat::CLAD_COMPAT_TemplateSuccess)
+      return Specialization;
+    return nullptr;
   }
 
   bool DiffCollector::LookupCustomDerivativeDecl(const DiffRequest& request) {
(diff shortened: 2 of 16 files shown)


lib/Differentiator/CladUtils.cpp
```diff
@@ -426,6 +426,14 @@ namespace clad {
...
       QualType valueType = GetValueType(T);
       valueType.removeLocalConst();
+      // If the const-ness of the type is hidden with sugar, e.g.
+      // `class_name<double>::const_value_type`, the approach above
+      // does not work and we have to desugar the type explicitly.
+      QualType canonicalType = valueType.getCanonicalType();
+      if (canonicalType.isConstQualified()) {
+        canonicalType.removeLocalConst();
+        return canonicalType;
+      }
```
PetroZarytskyi, line 436: This is supposed to be the last "uncovered" bit in this PR. It's only used in Kokkos, which, I believe, is not included in the coverage test. This part was designed to work only under certain circumstances because desugaring types might sometimes cause problems.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1158,24 +1148,19 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
...
-              request.CUDAGlobalArgsIndexes.push_back(i);
-            } else {
-              // create a param from the VarDecl
-              const ParmVarDecl* PVD =
-                  utils::BuildParmVarDecl(m_Sema, m_Sema.CurContext,
-                                          VD->getIdentifier(), VD->getType());
-              request.DVI.push_back(PVD);
+        if (!m_TopMostReq->CUDAGlobalArgsIndexes
+                 .empty()) { // CUDA device function call in global kernel
+                             // gradient
```
vgvassilev, line 1153:
```suggestion
        // CUDA device function call in global kernel gradient
        if (!m_TopMostReq->CUDAGlobalArgsIndexes.empty()) { 
```

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1143,8 +1133,8 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
       request.EnableUsefulAnalysis = m_TopMostReq->EnableUsefulAnalysis;
       request.CallContext = E;
 
-      // const auto* MD = dyn_cast<CXXMethodDecl>(FD);
-      if (m_TopMostReq->CUDAGlobalArgsIndexes.empty()) {
+      if (request.Mode != DiffMode::experimental_pushforward) {
+        // const auto* MD = dyn_cast<CXXMethodDecl>(FD);
```
vgvassilev, line 1137:
```suggestion
```

test/Misc/TimingsReport.C
```diff
@@ -5,7 +5,7 @@
 #include "clad/Differentiator/Differentiator.h"
 // CHECK: Timers for Clad Funcs
 // CHECK_STATS: *** INFORMATION ABOUT THE DIFF REQUESTS
-// CHECK_STATS-NEXT: <double nested1(double c)>[name=nested1, order=1, mode=pushforward, args='c']: #0 (source), (done)
```
vgvassilev, line 8: Why did we need to change this?

PetroZarytskyi: In pushforwards, DVI is usually not specified, and the differentiation happens wrt every parameter. Once we introduced static scheduling, DiffPlanner started creating DVI for pushforwards (simply including all parameters). This inconsistency became a problem in the commit ``Differentiate single-argument...`` because ``FindDerivedFunction`` matches DVI. You can take a look at the commit ``Differentiate single-argument...``.

vgvassilev (approved):
LGTM. Please squash into relevant set of commits.

PR #1373, opened 2025-05-04
Rework the time reporting.

vgvassilev:
This patch implements two separate timers. One per Analysis and one per DiffRequest. It also allows this logic to be triggered by setting the env variable CLAD_ENABLE_TIMING.

### DIFF
--- a/lib/Differentiator/Timers.cpp
+++ b/lib/Differentiator/Timers.cpp
@@ -1,0 +1,131 @@
+#include "clad/Differentiator/Timers.h"
+
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/Statistic.h"
+#include "llvm/ADT/StringMap.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/ManagedStatic.h"
+#include "llvm/Support/Timer.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include <cassert>
+#include <functional>
+#include <memory>
+#include <string>
+
+namespace clad {
+
+// Implementation heavily inspired by llvm::TimePassesHandler.
+class CladTimeInfo {
+  /// Group of timers that are part of the analysis stage.
+  llvm::TimerGroup m_AnalysisTG;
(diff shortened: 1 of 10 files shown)


tools/ClangPlugin.cpp
```diff
@@ -190,8 +208,8 @@ namespace clad {
     FunctionDecl* CladPlugin::ProcessDiffRequest(DiffRequest& request) {
       Sema& S = m_CI.getSema();
       if (!m_DerivativeBuilder)
-        m_DerivativeBuilder.reset(
-            new DerivativeBuilder(S, *this, m_DFC, m_DiffRequestGraph));
+        m_DerivativeBuilder = std::make_unique<DerivativeBuilder>(
+            S, *this, m_DFC, m_DiffRequestGraph);
```
PetroZarytskyi, line 212: I am surprised this works because I remember having problems with ``std::make_unique`` on older C++ versions (it appeared in C++14). Have we dropped the support for C++11?

PetroZarytskyi (approved):
The PR looks amazing! I only left a couple of small questions, but nothing too important.

tools/ClangPlugin.cpp
vgvassilev: Yes, we switched to require c++14 as minimum C++ version after we dropped clang9.

lib/Differentiator/Timers.cpp
```diff
@@ -0,0 +1,131 @@
...
+namespace clad {
+
+// Implementation heavily inspired by llvm::TimePassesHandler.
+class CladTimeInfo {
+  /// Group of timers that are part of the analysis stage.
+  llvm::TimerGroup m_AnalysisTG;
+  /// Group of timers that are part of the transformation stage.
+  llvm::TimerGroup m_DiffTG;
+
+  using TimerVector = llvm::SmallVector<std::unique_ptr<llvm::Timer>, 4>;
```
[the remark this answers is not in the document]
vgvassilev: Good point. Made it `llvm::SmallVector<llvm::Timer, 4>`.

PR #1381, opened 2025-05-11
Add template specialization handling and primary template derivation for specializations

pmozil:
This merge request is a repeat of #1277 . Clang-20 has improved handling of template definitions, and it recursed infinitely on the previous implementation, this pull request is to handle  template specializations properly and mend the issue

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -23,6 +23,7 @@
 #include "clang/Sema/Sema.h"
 #include "clang/Sema/SemaInternal.h"
 #include "clang/Sema/Template.h"
+#include <cinttypes>
 #include <clad/Differentiator/DiffMode.h>
 #include "clad/Differentiator/BaseForwardModeVisitor.h"
 #include "clad/Differentiator/CladUtils.h"
@@ -40,6 +41,7 @@
 
 #include <algorithm>
 #include <cstddef>
+#include <memory_resource>
 #include <string>
 
 #include "clad/Differentiator/CladUtils.h"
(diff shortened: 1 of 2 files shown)


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -146,6 +146,48 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
...
+
+        const TemplateArgumentList* TAL = FD->getTemplateSpecializationArgs();
+        TemplateArgumentList* TALCopy =
+            TemplateArgumentList::CreateCopy(m_Context, TAL->asArray());
+
+        returnedFD->setFunctionTemplateSpecialization(
+            returnedFTD, TALCopy, nullptr,
+            FD->getTemplateSpecializationKindForInstantiation());
+
+        std::cout << "\n-------------------------------------------------------"
```
parth-07, line 181: Remove the debug statements.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -524,6 +566,45 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
...
+        primaryFDRequest.Function =
+            FD->getPrimaryTemplate()->getTemplatedDecl();
+
+        auto* primaryFD = HandleNestedDiffRequest(primaryFDRequest);
+
+        FunctionTemplateDecl* NewFTD = nullptr;
+        auto Results = m_Context.getTranslationUnitDecl()->lookup(
+            primaryFD->getNameInfo().getName());
+
+        for (NamedDecl* ND : Results) {
```
parth-07, line 582: It seems to be repeated. Can you please move into a separate function to reuse code?

parth-07 (changes requested):
Please add tests for the feature.

pmozil:
Thanks for the feedback! This is currently a draft pull request, but there's no feature on github to mark one as such.
Will remoev debug statements and add tests after i get the feature to work!
Thanks again for your remarks on this, will use them as reminders!

PR #1382, opened 2025-05-11
Improve support for rvalue references in the reverse mode

PetroZarytskyi:
This PR improves support for static casts to rvalue references, which used to be ignored by Clad in the reverse mode. They are often generated by the compiler as part of move operators. This PR also improves support for rvalue reference arguments in ``_pullback`` and ``_reverse_forw`` functions. A test for the newly supported functionality is also added.


vgvassilev (approved):
Lgtm!

PR #1384, opened 2025-05-13
Only generate pushforwards of functions of real params in the reverse mode

PetroZarytskyi:
Currently, in the reverse mode, if a nested function has only one non-reference/pointer parameter, we generate a pushforward instead of a pullback for efficiency and simplicity. However, we also need to check that the parameter is of a real type. Since class types usually represent multiple real numbers, such functions cannot be differentiated in forward mode.

### DIFF
--- a/include/clad/Differentiator/CladUtils.h
+++ b/include/clad/Differentiator/CladUtils.h
@@ -393,6 +393,10 @@ namespace clad {
 
     bool IsDifferentiableType(clang::QualType T);
 
+    /// Returns true if FD can be differentiated as a pushforward
+    /// And be used in the reverse mode.
+    bool canUsePushforwardInRevMode(const clang::FunctionDecl* FD);
+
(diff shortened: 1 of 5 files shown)


include/clad/Differentiator/CladUtils.h
```diff
@@ -393,6 +393,10 @@ namespace clad {
 
     bool IsDifferentiableType(clang::QualType T);
 
+    /// Returns true if FD can be differentiated as a pushforward
+    /// And be used in the reverse mode.
+    bool usePushforwardInRevMode(const clang::FunctionDecl* FD);
```
vgvassilev, line 398:
```suggestion
    bool canUsePushforwardInRevMode(const clang::FunctionDecl* FD);
```

PetroZarytskyi: Addressed.

Issue #1383, opened 2025-05-13
Edge case `std::pow(0., 0.)` doesn't work correctly


guitargeek:
Reproducer as a ROOT macro:
```c++
#include <Math/CladDerivator.h>

double roo_codegen_0(double *params)
{
   return std::pow(params[0], params[1]);
}

#pragma clad ON
void gradient_request()
{
   clad::gradient(roo_codegen_0);
}
#pragma clad OFF
// clang-format off

std::vector<double> parametersVec = {0., 0.};

// clang-format on

// To run as a ROOT macro
void my_nll_full()
{
   std::vector<double> gradientVec(parametersVec.size());

   auto func = [&](std::span<double> params) { return roo_codegen_0(params.data()); };
   auto grad = [&](std::span<double> params, std::span<double> out) {
      return roo_codegen_0_grad(parametersVec.data(), out.data());
   };

   grad(parametersVec, gradientVec);

   auto numDiff = [&](int i) {
      const double eps = 1e-2;
      std::vector<double> p{parametersVec};
      p[i] = parametersVec[i] - eps;
      double funcValDown = func(p);
      p[i] = parametersVec[i] + eps;
      double funcValUp = func(p);
      return (funcValUp - funcValDown) / (2 * eps);
   };

   for (std::size_t i = 0; i < parametersVec.size(); ++i) {
      std::cout << i << ":" << std::endl;
      std::cout << "  numr : " << numDiff(i) << std::endl;
      std::cout << "  clad : " << gradientVec[i] << std::endl;
   }
}
```
Output:
```
0:
  numr : 0
  clad : -nan
1:
  numr : -inf
  clad : -nan
```

It would be good if `d/dx pow(x, 0.)` is correctly recognized as `zero` by Clad, since this can actually happen in production code and is mathematically valid:

https://www.wolframalpha.com/input?i=d%2Fdx+x%5E0+at+x+%3D+0

vgvassilev:
@a-elrawy, can you take a look?

PR #1385, opened 2025-05-14
Fix edge case for std::pow(0., 0.)

a-elrawy:
This pr address #1383 by fixing the edge case for std::pow(0., 0.) in the pow_pushforward and pow_pullback functions.

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -144,6 +144,8 @@ CUDA_HOST_DEVICE inline ValueAndPushforward<double, double>
 __builtin_pow_pushforward(double x, double exponent, double d_x,
                           double d_exponent) {
   auto val = __builtin_pow(x, exponent);
+  if (exponent == 0 && d_exponent == 0)
+    return {val, 0};
   double derivative = (exponent * __builtin_pow(x, exponent - 1)) * d_x;
   // Only add directional derivative of base^exp w.r.t exp if the directional
   // seed d_exponent is non-zero. This is required because if base is less than
@@ -160,6 +162,8 @@ CUDA_HOST_DEVICE inline ValueAndPushforward<float, float>
 __builtin_powf_pushforward(float x, float exponent, float d_x,
                            float d_exponent) {
   auto val = __builtin_powf(x, exponent);
+  if (exponent == 0 && d_exponent == 0)
+    return {val, 0};
   float derivative = (exponent * __builtin_powf(x, exponent - 1)) * d_x;
   // Only add directional derivative of base^exp w.r.t exp if the directional
   // seed d_exponent is non-zero. This is required because if base is less than
@@ -1296,6 +1300,8 @@ template <typename T1, typename T2, typename dT1, typename dT2,
 CUDA_HOST_DEVICE ValueAndPushforward<T_out, dT_out>
 pow_pushforward(T1 x, T2 exponent, dT1 d_x, dT2 d_exponent) {
   T_out val = ::std::pow(x, exponent);
+  if (exponent == static_cast<T2>(0) && d_exponent == static_cast<dT2>(0))
+    return {val, static_cast<dT_out>(0)};
   dT_out derivative = (exponent * ::std::pow(x, exponent - 1)) * d_x;
   // Only add directional derivative of base^exp w.r.t exp if the directional
   // seed d_exponent is non-zero. This is required because if base is less than
--- a/test/FirstDerivative/BuiltinDerivatives.C
+++ b/test/FirstDerivative/BuiltinDerivatives.C
@@ -490,6 +490,14 @@ double f_erfc(double x) { return std::erfc(x); }
 // CHECK-NEXT:     return _t0.pushforward;
 // CHECK-NEXT: }
 
+double f_pow_zero(double x, double y) { return std::pow(x, y); }
+// CHECK: double f_pow_zero_darg0(double x, double y) {
+// CHECK-NEXT:     double _d_x = 1;
+// CHECK-NEXT:     double _d_y = 0;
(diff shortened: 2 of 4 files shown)


vgvassilev (commented):
What about the `__builtin_pow_`?

a-elrawy:
> What about the `__builtin_pow_`?

I'm not sure if it has the same issue. I'll test and implement it if it does

a-elrawy:
> What about the `__builtin_pow_`?

I've added it for now, though I'm not sure how to test it

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -153,6 +153,8 @@ __builtin_pow_pushforward(double x, double exponent, double d_x,
   // = NaN.
   if (d_exponent)
     derivative += (__builtin_pow(x, exponent) * __builtin_log(x)) * d_exponent;
+  else if (exponent == 0)
+    derivative = 0;
```
guitargeek, line 157: I think the way how this is fixed is a bit confusing, because you first compute the `nan` a few lines before...
```c++
double derivative = (exponent * __builtin_pow(x, exponent - 1)) * d_x;
```
... and then you're resetting the derivative if `exponent == 0`. Can this not be done directly when you compute `derivative` to begin with? Like:
```c++
double derivative = exponent == 0. ? 0. : (exponent * __builtin_pow(x, exponent - 1)) * d_x;
```

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -1305,14 +1315,20 @@ pow_pushforward(T1 x, T2 exponent, dT1 d_x, dT2 d_exponent) {
   // = NaN.
   if (d_exponent)
     derivative += (::std::pow(x, exponent) * ::std::log(x)) * d_exponent;
+  else if (exponent == static_cast<T2>(0))
+    derivative = static_cast<dT_out>(0);
```
guitargeek, line 1319: Also here, I think we can check if `exponent == 0` when computing `derivative` initially.

guitargeek (commented):
Thank you very much for fixing this! Functionally, the PR looks correct to me, but I have some suggestions to avoid redundant checks and computations.

include/clad/Differentiator/BuiltinDerivatives.h
a-elrawy: Yes, I did it like that because I wasn't sure about d_exponent in pow(0,0).
The gradient for it now is `std::pow(0.0, 0.0) = {0.000000, -nan}`. 
Does this work? or should it be `std::pow(0.0, 0.0) = {0.000000, 0.000000}`
If it's not zero, I can just add it like this in the beginning of the function 
 `if (exponent == 0 && d_x == 1)
      return {val, 0};`

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -178,7 +182,10 @@ CUDA_HOST_DEVICE inline void __builtin_pow_pullback(double x, double exponent,
                                                     double* d_exponent) {
   auto t =
       __builtin_pow_pushforward(x, exponent, /*d_x=*/1., /*d_exponent=*/0.);
-  *d_x += t.pushforward * d_y;
+  if (exponent == 0)
+    *d_x = 0;
+  else
+    *d_x += t.pushforward * d_y;
```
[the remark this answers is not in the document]
a-elrawy: yes, you're absolutely right! Thanks for noticing. I'll remove it.

include/clad/Differentiator/BuiltinDerivatives.h
vgvassilev: Maybe we can also have an early return.

guitargeek: Indeed it should be `{0.000000, -nan}`. The derivative with respect to exponent is undefined, because there is a discontinuity. See:

```txt
root [0] std::pow(0., 0.)
(double) 1.0000000
root [1] std::pow(0., 0.00001)
(double) 0.0000000
root [2] std::pow(0., -0.00001)
(double) inf
root [3]
```

For x, there is no discontinuity and the result is just a constant:
```txt
root [0] std::pow(0., 0.)
(double) 1.0000000
root [1] std::pow(0.00001, 0.)
(double) 1.0000000
root [2] std::pow(-0.00001, 0.)
(double) 1.0000000
root [3]
```

a-elrawy:
> Thank you very much for fixing this! Functionally, the PR looks correct to me, but I have some suggestions to avoid redundant checks and computations.

Thanks a lot for the review and the suggestions! I've modified it now

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -144,6 +144,8 @@ CUDA_HOST_DEVICE inline ValueAndPushforward<double, double>
 __builtin_pow_pushforward(double x, double exponent, double d_x,
                           double d_exponent) {
   auto val = __builtin_pow(x, exponent);
+  if (exponent == 0 && d_x == 1)
```
guitargeek, line 147:
```suggestion
  if (exponent == 0 && d_exponent == 0)
```
You assume that it will be either `d_x == 1 && d_exponent == 0`, or `d_x == 0 && d_exponent == 1`. But I'm afraid this might not generalize, if `d_x` might also be a non-zero value that is not `1`. @vgvassilev, can this happen?

Maybe it's better to do the check `d_exponent == 0` instead? Also in the other pushforwards.

vgvassilev: > @vgvassilev, can this happen?

No idea, and I'd suggest if we are not sure to be conservative and assume it might happen..

vgvassilev:
@guitargeek, can we do another round of review?

guitargeek (commented):
Thanks a lot for addressing the comments! Looks good now to me!

vgvassilev (approved):
LGTM!

PR #1386, opened 2025-05-14
Do not ignore Imp casts when calling IsReferenceOrPointerArg in RMV::VisitCXXConstructExpr

PetroZarytskyi:
Ignoring implicit casts results in some arguments being considered as by-reference even though they are passed by value. Doing this is incorrect because passing a parameter by value uses implicit copying, which affects the derived code.


vgvassilev (approved):
LGTM!

PR #1387, opened 2025-05-19
Don't differentiate w.r.t. parameters of non-differentiable types in the reverse mode

PetroZarytskyi:
Currently, the diff planner does not check if parameters are of non-differentiable types. The argument corresponding to such a parameter will not have an adjoint, and RMV will request a different derivative without it. Because of this, the scheduled derivative will not be used and will possibly cause errors.
This PR also addresses a different problem: VarDecls of non-differentiable types used to be reused instead of being cloned, i.e., a declaration from `f` was directly used in `f_grad`. This leads to incorrect behavior and causes errors.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -1148,6 +1148,11 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
         for (size_t i = 0, e = FD->getNumParams(); i < e; ++i) {
           // if (MD && isLambdaCallOperator(MD)) {
           const auto* paramDecl = FD->getParamDecl(i);
+          QualType paramElemTy = utils::GetValueType(paramDecl->getType());
+          const CXXRecordDecl* RD = paramElemTy->getAsCXXRecordDecl();
+          if (RD && clad::utils::hasNonDifferentiableAttribute(RD))
+            continue;
+
           request.DVI.push_back(paramDecl);
 
           //}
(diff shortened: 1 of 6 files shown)


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1148,6 +1148,11 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
         for (size_t i = 0, e = FD->getNumParams(); i < e; ++i) {
           // if (MD && isLambdaCallOperator(MD)) {
           const auto* paramDecl = FD->getParamDecl(i);
+          QualType paramElemTy = utils::GetValueType(paramDecl->getType());
+          const CXXRecordDecl* RD = paramElemTy->getAsCXXRecordDecl();
+          if (RD && clad::utils::hasNonDifferentiableAttribute(RD))
+            continue;
```
vgvassilev, line 1154: Nice!

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2946,15 +2945,28 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
-          (typeDecl && clad::utils::hasNonDifferentiableAttribute(typeDecl))) {
+      if (typeDecl && typeDecl->isLambda()) {
         for (auto* D : DS->decls())
           if (auto* VD = dyn_cast<VarDecl>(D))
             decls.push_back(VD);
         Stmt* DSClone = BuildDeclStmt(decls);
         return StmtDiff(DSClone, nullptr);
       }
+
+      if (typeDecl && clad::utils::hasNonDifferentiableAttribute(typeDecl)) {
```
vgvassilev, line 2956: I suspect we need this because not all propagators are generated in the diff planner, right? Probably we need a fixme and we should open an issue to track that.

vgvassilev (approved):
Looks good, please address the comments, and feel free to merge it with appropriate set of commits.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: I moved the type non-diff check to ``utils::hasNonDifferentiableAttribute`` in the last commit. Would you like to do something else there? I think ``utils::hasNonDifferentiableAttribute`` is already the central point for handling the attribute.

PR #1388, opened 2025-05-20
Add support for sparsity pattern generation for jacobians

ovdiiuv:
This PR introduces very basic support for sparsity patterns for jacobians. Sparsity pattern that is used here is called **Compressed Sparse Row** a collection of three sets: `vals, column_index, row_index`. See more [here](https://en.wikipedia.org/wiki/Sparse_matrix#Compressed_sparse_row_(CSR,_CRS_or_Yale_format)). 

Consider following example:
```cpp
void f1(double a, double b, double c, double _clad_out_output[]) {
  _clad_out_output[0] = a;
  _clad_out_output[1] = a + b;
  _clad_out_output[2] = a + b + c;
}
```
Essentially we need to determine whether `_clad_out_output[i]` depends on a certain input variable. For that Dependency Analysis is implemented, which traverses the body of the function and does the following:

1. Identifies `_clad_out_output[i] = LHS` expressions.
2. Traverses LHS and adds `{i, independent_variable_index}` to the `OutputDependencySet` that idicates that `i-th` output depends on certain `independent_variable`(we numerate independent variables for consistency, hence `independent_variable_index`).

Having `OutputDependencySet` enables us to compute `column_index` and`row_index`. Iterating over it and assigning respective jacobian entry to `vals` give us the desired pattern.

TODO(not in this PR tho):

1. Enable handling of array parameters. Since we numerate independent variables we also have to account for fixed-size array parameters.
2. Add loop support. For that we would need some kind of analysis to determine if a loop is "standard" and keep info about it.
3. Enable branches. Right now we do not have branch-merging like on TBR or AA.
4. Don't account for the input parameters that are not in the DVI.

### DIFF
--- a/include/clad/Differentiator/CladConfig.h
+++ b/include/clad/Differentiator/CladConfig.h
@@ -31,6 +31,7 @@ enum opts : unsigned {
   disable_tbr = 1 << (ORDER_BITS + 3),
   enable_va = 1 << (ORDER_BITS + 5),
   disable_va = 1 << (ORDER_BITS + 6),
+  enable_sp = 1 << (ORDER_BITS + 8),
   enable_ua = 1 << (ORDER_BITS + 9),
   disable_ua = 1 << (ORDER_BITS + 10),
 
--- a/include/clad/Differentiator/CladUtils.h
+++ b/include/clad/Differentiator/CladUtils.h
@@ -15,6 +15,7 @@
 #include "llvm/ADT/StringRef.h"
 
 #include <string>
+#include <utility>
 
 namespace clang {
   class ASTContext;
@@ -24,6 +25,15 @@ namespace clang {
 
 namespace clad {
   namespace utils {
+  // move
+  struct compare {
+    bool operator()(const std::pair<int, int> a,
+                    const std::pair<int, int> b) const {
+      if (a.second == b.second)
+        return (a.first < b.first);
+      return (a.second <= b.second);
+    }
+  };
     /// If `FD` is an overloaded operator, returns a name, unique for
     /// each operator, that can be used to create valid C++ identifiers.
     /// Otherwise if `FD` is an ordinary function, returns the name of the
--- a/include/clad/Differentiator/DiffPlanner.h
+++ b/include/clad/Differentiator/DiffPlanner.h
@@ -1,6 +1,7 @@
 #ifndef CLAD_DIFF_PLANNER_H
 #define CLAD_DIFF_PLANNER_H
 
+#include "clad/Differentiator/CladUtils.h"
 #include "clad/Differentiator/DerivedFnCollector.h"
 #include "clad/Differentiator/DiffMode.h"
 #include "clad/Differentiator/DynamicGraph.h"
@@ -18,6 +19,7 @@
 
 #include <iterator>
 #include <set>
+#include <utility>
 
 namespace clang {
 class CallExpr;
@@ -52,6 +54,11 @@ private:
     std::set<const clang::VarDecl*> UsefulDecls;
     bool HasAnalysisRun = false;
   } m_UsefulRunInfo;
+
+  mutable struct DependencySparsityInfo {
+    std::set<std::pair<int, int>, utils::compare> OutputDependencySet;
+    bool HasAnalysisRun = false;
+  } m_DependencySparsityInfo;
 
 public:
   /// Function to be differentiated.
@@ -191,6 +198,9 @@ public:
   }
   std::set<const clang::VarDecl*>& getUsefulDecls() const {
     return m_UsefulRunInfo.UsefulDecls;
+  }
+  std::set<std::pair<int, int>, utils::compare>& getDependencySet() const {
+    return m_DependencySparsityInfo.OutputDependencySet;
   }
 };
 
--- a/include/clad/Differentiator/DiffPlanner.h
+++ b/include/clad/Differentiator/DiffPlanner.h
@@ -81,6 +81,7 @@ public:
   bool EnableTBRAnalysis = false;
   bool EnableVariedAnalysis = false;
   bool EnableUsefulAnalysis = false;
+  bool EnableSparsity = false;
   /// A flag specifying whether this differentiation is to be used
   /// in immediate contexts.
   bool ImmediateMode = false;
@@ -153,6 +154,7 @@ public:
            Args == other.Args && Mode == other.Mode &&
            EnableTBRAnalysis == other.EnableTBRAnalysis &&
            EnableVariedAnalysis == other.EnableVariedAnalysis &&
+           EnableSparsity == other.EnableSparsity &&
            EnableUsefulAnalysis == other.EnableUsefulAnalysis &&
            DVI == other.DVI && use_enzyme == other.use_enzyme &&
            DeclarationOnly == other.DeclarationOnly && Global == other.Global &&
@@ -200,6 +202,7 @@ public:
     bool EnableTBRAnalysis = false;
     bool EnableVariedAnalysis = false;
     bool EnableUsefulAnalysis = false;
+    bool EnableSparsity = false;
   };
 
   class DiffCollector: public clang::RecursiveASTVisitor<DiffCollector> {
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -17,6 +17,7 @@
 #include "FunctionTraits.h"
 #include "Matrix.h"
 #include "NumericalDiff.h"
+#include "SparsityPattern.h"
 #include "Tape.h"
 
 #include <array>
@@ -395,7 +396,7 @@ CUDA_HOST_DEVICE T push(tape<T>& to, ArgsT... val) {
       template <class Fn, class... Args>
       constexpr CUDA_HOST_DEVICE return_type_t<CladFunctionType>
       execute_helper(Fn f, CUDA_ARGS Args&&... args) const {
-        // `static_cast` is required here for perfect forwarding.
+        //  `static_cast` is required here for perfect forwarding.
 #if defined(__CUDACC__)
         if constexpr (sizeof...(Args) >= 2) {
           auto secondArg =
--- a/include/clad/Differentiator/SparsityPattern.h
+++ b/include/clad/Differentiator/SparsityPattern.h
@@ -1,0 +1,64 @@
+#ifndef CLAD_DIFFERENTIATOR_SPARSITYPATTERN_H
+#define CLAD_DIFFERENTIATOR_SPARSITYPATTERN_H
+
+#include <assert.h>
+
+#include <cstddef>
+#include <initializer_list>
+#include <type_traits>
+
+namespace clad {
+template <typename T> class sparsity_pattern {
+  int* m_col = nullptr;
+  int* m_row = nullptr;
+  T* m_vals = nullptr;
+
+  std::size_t m_col_size;
+  std::size_t m_row_size;
+
+public:
+  sparsity_pattern() = default;
+
+  void set_col_idx(std::initializer_list<int> col) {
+    m_col_size = col.size();
+    // NOLINTNEXTLINE(cppcoreguidelines-owning-memory)
+    m_col = new int[m_col_size];
+    // NOLINTNEXTLINE(cppcoreguidelines-owning-memory)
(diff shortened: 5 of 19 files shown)


include/clad/Differentiator/CladUtils.h
```diff
@@ -24,6 +25,15 @@ namespace clang {
 
 namespace clad {
   namespace utils {
+  // move
```
vgvassilev, line 28: Can you clarify that comment?

include/clad/Differentiator/DependencySparsityAnalyzer.h
```diff
@@ -0,0 +1,71 @@
+#ifndef CLAD_DIFFERENTIATOR_DEPENDENCYSPARSITYANALYZER_H
```
vgvassilev, line 1: This header file can go in `lib`.

include/clad/Differentiator/DependencySparsityAnalyzer.h
```diff
@@ -0,0 +1,71 @@
...
+namespace clad {
+
+// struct compare{
+//     bool operator()(const std::pair<int, int> a, const std::pair<int, int> b)
+//     const{
+//         if(a.first != b.first)
+//             return (a.second > b.second);
+//         return (a.first > b.first);
+//     }
+// };
```
vgvassilev: Seems like a stray comment.

include/clad/Differentiator/Differentiator.h
```diff
@@ -163,6 +164,7 @@ CUDA_HOST_DEVICE T push(tape<T>& to, ArgsT... val) {
   constexpr CUDA_HOST_DEVICE return_type_t<F>
   execute_with_default_args(list<Rest...>, F f, list<fArgTypes...>,
                             CUDA_ARGS CUDA_REST_ARGS Args&&... args) {
+    // assert(0 && "default_args");
```
vgvassilev: Why did we need to disable this assert?

lib/Differentiator/DependencySparsityAnalyzer.cpp
```diff
@@ -0,0 +1,105 @@
...
+      // The const_cast is inevitable, since there is no
+      // ConstRecusiveASTVisitor.
+      // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
+      TraverseStmt(const_cast<clang::Stmt*>(S));
+    }
+  }
+
+  for (const clang::CFGBlock::AdjacentBlock succ : block.succs()) {
+    if (!succ)
+      continue;
```
vgvassilev, line 52: Can we cover that with a test?

lib/Differentiator/DependencySparsityAnalyzer.cpp
```diff
@@ -0,0 +1,105 @@
...
+    m_CFGQueue.insert(succ->getBlockID());
+  }
+}
+
+CFGBlock* DependencySparsityAnalyzer::getCFGBlockByID(unsigned ID) {
+  return *(m_CFG->begin() + ID);
+}
+
+bool DependencySparsityAnalyzer::VisitBinaryOperator(BinaryOperator* BinOp) {
```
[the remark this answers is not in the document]
vgvassilev: Seems like a good comment.

lib/Differentiator/DependencySparsityAnalyzer.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+bool DependencySparsityAnalyzer::VisitArraySubscriptExpr(
+    ArraySubscriptExpr* ASE) {
+  auto* dASE = dyn_cast<DeclRefExpr>(ASE->getBase()->IgnoreImpCasts());
+  llvm::StringRef Name = dASE->getDecl()->getName();
+
+  if (Name.contains("_clad_out_")) {
+    const Expr* idx = ASE->getIdx();
+    if (isa<IntegerLiteral>(idx)) {
```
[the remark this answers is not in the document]
vgvassilev: Likewise.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -851,6 +855,8 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
                       "TBR analysis is not meant for forward mode AD.");
       return true;
     }
+    if (enable_sp_in_req)
+      request.EnableSparsity = enable_sp_in_req;
```
vgvassilev, line 860: Worth having a test case.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1207,8 +1213,21 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
         analyzer.Analyze(request.Function);
       }
 
+      // if(request.EnableSparsity)
+      //   llvm::errs() << "\n============\n";
```
vgvassilev: Stray comment.

lib/Differentiator/JacobianModeVisitor.cpp
```diff
@@ -260,6 +262,243 @@
...
+    Expr* colIdxE = IntegerLiteral::Create(m_Context, llvm::APInt(32, p.first),
+                                           m_Context.IntTy, noLoc);
+    Elements.emplace_back(colIdxE);
+  }
+  return m_Sema.ActOnInitList(noLoc, Elements, noLoc).get();
+}
+
+clang::Expr*
+ComputeRowIdx(clang::ASTContext& m_Context, clang::Sema& m_Sema,
+              std::set<std::pair<int, int>, utils::compare> depset) {
```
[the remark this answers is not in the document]
vgvassilev: Seem like a good suggestion.

lib/Differentiator/JacobianModeVisitor.cpp
```diff
@@ -260,6 +262,241 @@ DerivativeAndOverload JacobianModeVisitor::Derive() {
...
+      outputType = dParam.param->getType()->getPointeeType();
+    args.push_back(dParam.param);
+  }
+
+  std::string derivedFnName = m_DiffReq.BaseFunctionName + "_sparse_jac";
+  if (args.size() != FD->getNumParams()) {
+    for (const ValueDecl* arg : args) {
+      const auto* it = std::find(FD->param_begin(), FD->param_end(), arg);
+      auto idx = std::distance(FD->param_begin(), it);
+      derivedFnName += ('_' + std::to_string(idx));
```
vgvassilev, line 324: Can you add a test here?

vgvassilev (commented):
Seems like we are heading in a good direction. Please take a look at my partial review; clang-tidy comments and we also should add a demo and documentation of this new feature.

include/clad/Differentiator/CladUtils.h
ovdiiuv: I wonder what is the best place for the compare struct.

include/clad/Differentiator/Differentiator.h
ovdiiuv: there are bunch of debug comments, will remove all of them

include/clad/Differentiator/CladUtils.h
vgvassilev: Probably somewhere close to where is being used? But why do you need to define a call operator for this type?

include/clad/Differentiator/CladConfig.h
```diff
@@ -31,6 +31,7 @@ enum opts : unsigned {
   disable_tbr = 1 << (ORDER_BITS + 3),
   enable_va = 1 << (ORDER_BITS + 5),
```
parth-07, line 32: @vgvassilev A bit unrelated. Is there any particular reason for keeping separate bits for enabling / disabling a feature? We have separate bits for enable_va / disable_va, enable_tbr / disable_tbr and enable_ua/disable_ua.

include/clad/Differentiator/CladConfig.h
```diff
@@ -31,6 +31,7 @@ enum opts : unsigned {
...
   disable_va = 1 << (ORDER_BITS + 6),
+  enable_sp = 1 << (ORDER_BITS + 8),
```
parth-07, line 34: Nitpick: Why didn't we use the 7th bit?

include/clad/Differentiator/CladUtils.h
```diff
@@ -24,6 +25,15 @@ namespace clang {
...
+  struct compare {
+    bool operator()(const std::pair<int, int> a,
```
parth-07, line 30: We should pass `a` and `b` as const reference here.

parth-07: Also, can you please add a description explaining what exactly is being compared here? Perhaps the struct name can be more informative than `compare`?

include/clad/Differentiator/Differentiator.h
```diff
@@ -17,6 +17,7 @@
 #include "FunctionTraits.h"
 #include "Matrix.h"
 #include "NumericalDiff.h"
+#include "SparsityPattern.h"
```
parth-07, line 20: Is it being used in this file?

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -53,6 +55,11 @@ struct DiffRequest {
     bool HasAnalysisRun = false;
   } m_UsefulRunInfo;
 
+  mutable struct DependencySparsityInfo {
+    std::set<std::pair<int, int>, utils::compare> OutputDependencySet;
```
parth-07, line 59: `utils::compare` should be defined in `DiffPlanner` instead of `utils::` if it is only being used for DependencySparsityInfo.

parth-07: I see that this `std::set<...>` type is getting used at a lot of places. Can you please expose this type as a member of `DependencySparsityInfo` so that the type does not have to be repeated everywhere?

include/clad/Differentiator/SparsityPattern.h
```diff
@@ -0,0 +1,68 @@
...
+  int* m_row = nullptr;
+  T* m_vals = nullptr;
+
+  std::size_t m_col_size;
+  std::size_t m_row_size;
+
+public:
+  sparsity_pattern() = default;
+  ~sparsity_pattern() {
+      delete[] m_col;
```
parth-07, line 22: Is there any particular reason for using manual memory allocation instead of, let's say, `std::vector` here?

lib/Differentiator/DependencySparsityAnalyzer.cpp
```diff
@@ -0,0 +1,105 @@
...
+  m_CFG = clang::CFG::buildCFG(FD, FD->getBody(), &m_Context, Options);
+
+  // Set current block ID to the ID of entry the block.
+  CFGBlock* entry = &m_CFG->getEntry();
+  m_CurBlockID = entry->getBlockID();
+
+  int i = 0;
+  for (const auto& par : FD->parameters()) {
+    if (!utils::isArrayOrPointerType(par->getType())) {
+      m_ParameterNum[dyn_cast<VarDecl>(par)] = i;
```
parth-07, line 20: `dyn_cast` will return `nullptr` if `par` is not a `VarDecl` and then the expression may have an undefined behavior. Can we use `llvm::cast` instead?

lib/Differentiator/DependencySparsityAnalyzer.cpp
```diff
@@ -0,0 +1,105 @@
...
+  return *(m_CFG->begin() + ID);
+}
+
+bool DependencySparsityAnalyzer::VisitBinaryOperator(BinaryOperator* BinOp) {
+  Expr* L = BinOp->getLHS();
+  Expr* R = BinOp->getRHS();
+  const auto opCode = BinOp->getOpcode();
+  if (BinOp->isAssignmentOp()) {
+    TraverseStmt(L);
+    // m_MarkingMode = true;
```
parth-07, line 67: Should this be un-commented?

include/clad/Differentiator/SparsityPattern.h
ovdiiuv: No, apart from consistency with other classes in clad.

include/clad/Differentiator/DiffPlanner.h
ovdiiuv: Yes, thanks for pointing that out.

PR #1391, opened 2025-05-27
Schedule constructor pullbacks statically

PetroZarytskyi:
This PR makes the constructor pullback scheduling the default for both custom and clad-generated derivatives. The only exception is Hessians, which still only support dynamical scheduling.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -179,7 +179,9 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
 
   void DiffRequest::updateCall(FunctionDecl* FD, FunctionDecl* OverloadedFD,
                                Sema& SemaRef) {
-    CallExpr* call = this->CallContext;
+    assert(isa<CallExpr>(this->CallContext) &&
+           "Trying to update an unsupported expression");
+    auto* call = cast<CallExpr>(this->CallContext);
 
     assert(call && "Must be set");
     assert(FD && "Trying to update with null FunctionDecl");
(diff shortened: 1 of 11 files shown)


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1252,4 +1254,59 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
     }
     return true;
   }
+
+  bool DiffCollector::VisitCXXConstructExpr(CXXConstructExpr* E) {
+    DiffRequest request;
+    CXXConstructorDecl* CD = E->getConstructor();
```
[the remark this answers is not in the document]
PetroZarytskyi: Makes sense, done.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1252,4 +1253,57 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
     }
     return true;
   }
+
+  bool DiffCollector::VisitCXXConstructExpr(CXXConstructExpr* E) {
+    DiffRequest request;
```
vgvassilev, line 1258: That’s performance sensitive code and probably that allocation should go down, after the early exits too.

PetroZarytskyi: Makes sense, I refactored the code to avoid declaring vars before the early exit.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1252,4 +1253,56 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
...
+      return true;
+
+    // Don't build propagators for calls that do not contribute in
+    // differentiable way to the result.
+    if (allArgumentsAreLiterals(E->arguments(), m_ParentReq))
+      return true;
+
+    // FIXME: add support for the forward mode
+    if (m_TopMostReq->Mode != DiffMode::reverse)
+      return true;
```
vgvassilev, line 1269: That's faster, should go before `allArgumentsAreLiterals`.

vgvassilev (approved):
Lgtm!

Issue #665, opened 2023-12-04
Create a benchmark script


vgvassilev:
Our CI has already benchmark comparison action which compares the benchmarks against the submitted pull request:

https://github.com/vgvassilev/clad/blob/ec76b704ced2bb7471b25b51a7c24273374700d9/.github/workflows/ci.yml#L755-L781

We should turn that into a script `benchmark_compare.py` that can take two revisions and produce a comparison output in the same way. That'd allow us to run on local machines, not only in the CI.

jayantpranjal0:
@vgvassilev instead of writing a python script wouldn't a bash script work?

jayantpranjal0:
Got it. Thanks!

vgvassilev:
@aditimjoshi, can you take a look?

PR #1394, opened 2025-06-01
Fix issue #665: Created a benchmark script

aditimjoshi:
Fixes issue #665. Added a benchmark script which takes two revisions (baseline and current) and finds and compares the benchmarks of both on the local machine.

### DIFF
--- a/benchmark/benchmark_compare.py
+++ b/benchmark/benchmark_compare.py
@@ -1,0 +1,47 @@
+import subprocess
+import argparse
+import os
+from pathlib import Path
+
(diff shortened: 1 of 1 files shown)


vgvassilev:
> > Fixes issue #665. Added a benchmark script which takes two revisions (baseline and current) and finds and compares the benchmarks of both on the local machine.
> 
> Can you change the description and the commit message to end with:
> 
> `Fixes #665` which will make github link these two and close the issue upon merging?

I will fix it during the squash but next time please follow that rule.

PR #1395, opened 2025-06-02
Add template specialization handling and primary template derivation for specializations 

pmozil:
This merge request is a repeat of https://github.com/vgvassilev/clad/pull/1381 . Clang-20 has improved handling of template definitions, and it recursed infinitely on the previous implementation, this pull request is to handle template specializations properly and mend the issue.

Closed last one on accident when trying to make pipelines work


lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -136,11 +136,31 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
 
       const TemplateArgumentList* TAL =
           R.Function->getTemplateSpecializationArgs();
-      TemplateArgumentList* TALCopy =
-          TemplateArgumentList::CreateCopy(S.getASTContext(), TAL->asArray());
-      dFD->setFunctionTemplateSpecialization(
-          SpecFTD, TALCopy, nullptr,
-          R.Function->getTemplateSpecializationKind());
+
+      void* location = nullptr;
```
```suggestion
      void *InsertPos = nullptr;
```

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -66,6 +67,92 @@ DerivativeBuilder::~DerivativeBuilder() {}
 static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
   DeclContext* DC = D->getLexicalDeclContext();
   if (auto* dFD = dyn_cast<FunctionDecl>(D)) {
+    // Don't specialize cxx methods -
+    // private methods break clad, refer to the
+    // todo regarding proper handling of private fields below
+    bool shouldSkipSpecialization = R.Function->isOverloadedOperator();
+    if (dyn_cast<CXXMethodDecl>(R.Function) != nullptr)
```
```suggestion
    if (isa<CXXMethodDecl>(R.Function))
```

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -66,6 +67,92 @@ DerivativeBuilder::~DerivativeBuilder() {}
...
+      shouldSkipSpecialization = true;
```
vgvassilev: We can probably early return and simplify stuff below.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -66,6 +67,92 @@ DerivativeBuilder::~DerivativeBuilder() {}
...
+      auto Results = DC->lookup(dFD->getNameInfo().getName());
+
+      for (NamedDecl* ND : Results) {
+        if (auto* FTD = dyn_cast<FunctionTemplateDecl>(ND)) {
+          SpecFTD = FTD;
+          break;
+        }
+
+        if (auto* FD = dyn_cast<FunctionDecl>(ND)) {
+          if (auto* FTD = FD->getDescribedFunctionTemplate()) {
```
vgvassilev: Please expand the type here -- we only use auto if the reader can easily deduce the type by looking at the right hand side.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -66,6 +67,92 @@ DerivativeBuilder::~DerivativeBuilder() {}
...
+          }
+
+          if (auto* FTD = FD->getPrimaryTemplate()) {
+            SpecFTD = FTD;
+            break;
+          }
+        }
+      }
+
+      // If no template found, create a dummy one
```
```suggestion
      // If no template found, create a forward declaration.
```

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -66,6 +67,92 @@ DerivativeBuilder::~DerivativeBuilder() {}
...
+          SpecFTD->findSpecialization(TAL->asArray(), location);
+      (void)(location);
+
+      if (SpecFD != nullptr) {
+        dFD->setFunctionTemplateSpecialization(
+            SpecFTD, TALCopy, nullptr, SpecFD->getTemplateSpecializationKind());
+      } else {
+        dFD->setFunctionTemplateSpecialization(SpecFTD, TALCopy, nullptr,
+                                               TSK_ExplicitSpecialization);
+      }
```
vgvassilev: Don't we need to call `AddSpecialization`?

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -145,7 +232,7 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
               ? VB.Clone(FD->getTrailingRequiresClause())
               : nullptr);
 
-      returnedFD->setAccess(FD->getAccess());
+      returnedFD->setAccess(AS_public);
```
vgvassilev: We need a comment why this changes the visibility.

vgvassilev (commented):
Here is a partial review. Looks like we are going in a good direction. Thank you!

lib/Differentiator/DerivativeBuilder.cpp
pmozil: `AddSpecialization` is a private method, and it's called in `setFunctionTemplateSpecialization`

PR #1398, opened 2025-06-04
Update a test of a break-stmt to avoid regression in #1123

### LINKED ISSUE #1123: If stmt of break cond is not added before compound for-loop inc diff when using TBR analysis in reverse mode
**This issue is opened in advance of PR #833 getting merged so not to forget this afterwards**

When we have a break stmt in a for loop, we shouldn't perform the for loop increment differentiation in the first iteration in the reverse pass (as the original stmt is also not performed in the forward pass after the break stmt). This is why we add an if stmt to check whether this is the path that the break stmt corresponds to in the reverse pass, to not perform the inc diff stmt. Examples and code found in PR #833.
When the inc diff is a compound stmt it seems that this if-check is not added when using TBR analysis, even though it is added without it:  

* Without TBR analysis:
```c++
if (clad::size(_t3) != 0 && clad::back(_t3) != 1) {
            res = clad::pop(_t2);
            double _r_d1 = _d_res;
            _d_res = 0.;
            *_d_i += 7 * _r_d1 * j;
            *_d_j += 7 * i * _r_d1;
            _d_c += 0;
            --c;
}
 ```
* With TBR analysis:
 ```
 {
            res = clad::pop(_t2);
            double _r_d1 = _d_res;
            _d_res = 0.;
            *_d_i += 7 * _r_d1 * j;
            *_d_j += 7 * i * _r_d1;
            _d_c += 0;
            --c;
 }
 ```

PetroZarytskyi:
This test wasn't updated by mistake. Since ``CHECK-NEXT:`` doesn't care about the way the line starts, a whole if-stmt could be dropped without the test failing.
Fixes #1123.

### DIFF
--- a/test/Gradient/Loops.C
+++ b/test/Gradient/Loops.C
@@ -2086,7 +2086,7 @@ double fn26(double i, double j) {
 // CHECK-NEXT:             if (!_t0)
 // CHECK-NEXT:                 break;
 // CHECK-NEXT:         }
-// CHECK-NEXT:         {
+// CHECK-NEXT:         if (_t0 != _numRevIterations0 || (clad::back(_t3) != 1)) {
 // CHECK-NEXT:             res = clad::pop(_t2);
 // CHECK-NEXT:             double _r_d1 = _d_res;
 // CHECK-NEXT:             _d_res = 0.;


vgvassilev (approved):
LGTM!

PR #1400, opened 2025-06-08
Enable static derivative scheduling in the forward mode by default

PetroZarytskyi:
This has been achieved with the exception of Hessians and high-order derivatives.

The PR also solves a problem that arose along the way: it turned out to be dangerous to instantiate custom derivative templates on the diffplanner stage, which led to linker errors. The new custom derivative lookup system postpones the actual overload resolution until the ``ProcessDiffRequest`` stage and utilizes clang infrastructure (``Sema::ActOnCallExpr``) for that.

Note: Just like in the reverse mode, the new static custom derivative lookup system demands exact parameter type match between the original function and the custom derivative, e.g., for a function
```
double f(double x);
```
Only a function of this signature will be considered a custom derivative:
```
clad::ValueAndPushforward<double, double> f(double x, double _d_x);
```
For example, these will NOT be accepted:
```
clad::ValueAndPushforward<float, float> f(float x, float _d_x);
clad::ValueAndPushforward<double, double> f(const double x, double _d_x);
clad::ValueAndPushforward<double, double> f(double& x, double _d_x);
```
A template is supposed to be able to be instanstiated to the exact required signature.


test/FirstDerivative/CallArguments.C
```diff
@@ -138,16 +138,22 @@ float f_literal_helper(float x, char ch, float* p, float* q) {
...
+// CHECK-NEXT:         return {x * x, _d_x * x + x * _d_x};
+// CHECK-NEXT:     return {-x * x, -_d_x * x + -x * _d_x};
+// CHECK-NEXT: }
+
 float f_literal_args_func(float x, float y, float *z) {
   printf("hello world ");
-  return x * f_literal_helper(0.5, 'a', z, nullptr);
-}
+  return x * f_literal_helper(x, 'a', z, nullptr);
+} // x ^ 3
```
PetroZarytskyi, line 150: While moving some responsibilities to the diff planner, I accidentally improved non-diff call handling. Here, the original call ``f_literal_helper(0.5, 'a', z, nullptr)`` takes no differentiable parameters and so the pullback wasn't generated at all. It pretty much defeated the purpose of the test so I replaced ``0.5`` with ``x`` to make the call differentiable again. This also changed the numerical results, which can be seen below (``0.25`` became ``3``). Not sure if it's better this way or the way it was.

test/FirstDerivative/FunctionCallsWithResults.C
```diff
@@ -81,8 +81,7 @@ float test_4(float x) {
 
 // CHECK: float test_4_darg0(float x) {
 // CHECK-NEXT: float _d_x = 1;
-// CHECK-NEXT: clad::ValueAndPushforward<float, float> _t0 = clad::custom_derivatives::overloaded_pushforward();
-// CHECK-NEXT: return _t0.pushforward;
+// CHECK-NEXT: return 0.F;
```
PetroZarytskyi, line 84: Same thing as what happened above. This call is non-differentiable, and it doesn't make sense that it had a derivative here. The other thing is that the provided custom derivative returns ``4``, which is obviously not the correct derivative. Do we want to use custom derivatives even if we know the function is non-differentiable?

To be clear: we never considered custom derivatives for non-differentiable calls in the diff planner, it just worked that way in the ``BaseForwardModeVisitor``.

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -560,7 +560,8 @@ CUDA_HOST_DEVICE void expl_pullback(T x, U d_y, T* d_x) {
 // 2.2 exp2, exp2f, exp2l
 template <typename T, typename dT>
 CUDA_HOST_DEVICE ValueAndPushforward<T, dT> exp2_pushforward(T x, dT d_x) {
-  return {::std::exp2(x), ::std::exp2(x) * ::std::log(2) * d_x};
+  return {::std::exp2(x),
+          static_cast<dT>(::std::exp2(x) * ::std::log(2) * d_x)};
```
vgvassilev, line 560: Why we started needing a static cast here?

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -108,6 +108,10 @@ struct DiffRequest {
   // A flag to enable the use of enzyme for backend instead of clad
   bool use_enzyme = false;
 
+  /// UnresolvedLookupExpr or DeclRefExpr representing the custom derivative
+  /// overload
+  clang::Expr* CustomDerivative = nullptr;
```
vgvassilev, line 113: Should we update `operator==` with this? That is, should two `DiffRequest` compare equal if they point to two different custom derivatives?

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -219,12 +223,10 @@ struct DiffRequest {
 
     bool m_IsTraversingTopLevelDecl = true;
 
-    DerivedFnCollector& m_DFC;
-
   public:
     DiffCollector(clang::DeclGroupRef DGR, DiffInterval& Interval,
                   clad::DynamicGraph<DiffRequest>& requestGraph, clang::Sema& S,
-                  RequestOptions& opts, DerivedFnCollector& DFC);
```
vgvassilev, line 227: Nice!

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -924,27 +923,51 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
...
+      CXXScopeSpec SS;
+      return S.BuildDeclarationNameExpr(SS, Found, /*ADL*/ false).get();
+    }
     return nullptr;
   }
 
-  bool DiffCollector::LookupCustomDerivativeDecl(const DiffRequest& request) {
-    auto DFI = m_DFC.Find(request);
-    if (DFI.IsValid())
-      return true;
```
vgvassilev, line 947: I wonder if that would have a performance impact...

test/FirstDerivative/BuiltinDerivatives.C
```diff
@@ -391,7 +391,7 @@ double f_fmodf(double x, double y) { return ::fmodf(x, y); }
 // CHECK: double f_fmodf_darg0(double x, double y) {
 // CHECK-NEXT:     double _d_x = 1;
 // CHECK-NEXT:     double _d_y = 0;
-// CHECK-NEXT:     {{.*}}ValueAndPushforward<double, double> _t0 = {{.*}}fmodf_pushforward(x, y, _d_x, _d_y);
+// CHECK-NEXT:     {{.*}}ValueAndPushforward<float, float> _t0 = {{.*}}fmodf_pushforward(x, y, _d_x, _d_y);
```
vgvassilev, line 394: Can we mention in the commit messages that now we are much more precise in the template type deduction? That's now looks more accurate.

tools/ClangPlugin.cpp
```diff
@@ -279,17 +279,24 @@ void InitTimers();
           OverloadedDerivativeDecl = DFI.OverloadedDerivedFn();
           alreadyDerived = true;
         } else {
-          auto deriveResult = m_DerivativeBuilder->Derive(request);
```
vgvassilev, line 282: Can't we sink `ProcessCustomDerivative` under `Derive`?

lib/Differentiator/DiffPlanner.cpp
PetroZarytskyi: Oh, good catch! I wanted to look into this and completely forgot.

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -1201,7 +1202,8 @@ CUDA_HOST_DEVICE void atanhl_pullback(T x, U d_z, T* d_x) {
 // 6.1 erf, erff, erfl
 template <typename T, typename dT>
 CUDA_HOST_DEVICE ValueAndPushforward<T, dT> erf_pushforward(T x, dT d_x) {
-  return {::std::erf(x), (2 / ::std::sqrt(M_PI)) * ::std::exp(-x * x) * d_x};
+  return {::std::erf(x),
+          static_cast<dT>(2 / ::std::sqrt(M_PI)) * ::std::exp(-x * x) * d_x};
```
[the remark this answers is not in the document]
PetroZarytskyi: @vgvassilev, cmath is already included. Is this a bug?

include/clad/Differentiator/DiffPlanner.h
PetroZarytskyi: This is an interesting question. In theory, other fields of the requests are supposed to fully define the custom derivatives since they contain the original function and the diff args, as well as all of the other requirements. Maybe we could even add an assert. The other question is whether the exact ``clang::Expr*`` will match. I'm guessing it's possible to have equivalent exprs but not equal.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -33,6 +33,7 @@
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/IR/Constants.h"
+#include "llvm/Support/Casting.h"
 #include "llvm/Support/SaveAndRestore.h"
```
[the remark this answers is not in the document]
PetroZarytskyi: This is the header with cast_or_null, which, as it complained, wasn't directly included. Do I not understand something?

tools/ClangPlugin.cpp
PetroZarytskyi: I did it as a separate commit for now. Is this what you had in mind?

test/FirstDerivative/BuiltinDerivatives.C
PetroZarytskyi: I added it both to the commit message and the PR message.

PetroZarytskyi: Also, I think we need to add diagnostics that show the user the required signature (assuming a custom derivative of the same name was written). I basically already implemented this for debug purposes but right now, it's going to make a mess because of ``constructor_pullback``. I remember we thought of renaming them to ``ClassNam_ctor_pullback``, which would help here.

lib/Differentiator/BaseForwardModeVisitor.cpp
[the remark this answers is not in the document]
vgvassilev: Is claims that the header is not used.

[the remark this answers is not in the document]
PetroZarytskyi: I know. I added it after clang-tidy complained that cast_or_null was not directly included. It didn't suggest a specific header, so I looked it up in the clang documentation. I added the one that was mentioned there, but now clang-tidy claims it's unused.

include/clad/Differentiator/DiffPlanner.h
vgvassilev: Ok, if we decide to not add it as part of the `operator==` then we should add a comment there so that people don't add it by accident.

include/clad/Differentiator/STLBuiltins.h
```diff
@@ -731,7 +731,7 @@ template <typename Tuple> auto second_half_tuple_tie(const Tuple& tpl) {
   return select_tuple_elements_tie(tpl, second_half);
 }
 
-template <typename... Args> auto tie_pushforward(Args&&... args) noexcept {
```
vgvassilev: Why did we have to make this change?

lib/Differentiator/BaseForwardModeVisitor.cpp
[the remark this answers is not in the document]
vgvassilev: I believe you need to include `#include "clang/Basic/LLVM.h"`

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -924,27 +923,51 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
   static bool allArgumentsAreLiterals(const CallExpr::arg_range& args,
                                       const DiffRequest* request) {
     return std::none_of(args.begin(), args.end(), [&request](const Expr* A) {
+      // A->dump();
```
vgvassilev: A stray comment?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -924,27 +923,51 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
...
-      return Specialization;
+  static Expr* getOverloadExpr(clang::Sema& S, const std::string& Name,
+                               DeclContext* DC, QualType DerivativeType) {
+    IdentifierInfo* II = &S.getASTContext().Idents.get(Name);
+    DeclarationNameInfo DNInfo(II, utils::GetValidSLoc(S));
+    LookupResult Found(S, DNInfo, Sema::LookupOrdinaryName);
+    S.LookupQualifiedName(Found, DC);
+    bool overloadFound = false;
+    for (NamedDecl* candidate : Found) {
+      if (auto* usingShadow = dyn_cast<UsingShadowDecl>(candidate))
```
vgvassilev, line 940: I guess this routine should have sole documentation. Why we strip the `UsingShadowDecl`s, etc.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -924,27 +923,51 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
...
+        candidate = usingShadow->getTargetDecl();
+      if (auto* FD = dyn_cast<FunctionDecl>(candidate)) {
+        if (utils::SameCanonicalType(FD->getType(), DerivativeType)) {
```
vgvassilev: Not part of this PR but that should be `IsSameCanonicalType`

test/FirstDerivative/BuiltinDerivatives.C
vgvassilev: Makes sense. We generally want to extend that to show us the full signature of the expected pullback..

tools/ClangPlugin.cpp
vgvassilev: Yes, exactly.

include/clad/Differentiator/STLBuiltins.h
PetroZarytskyi: Good catch, that was just a part of the debug process. I reverted this change

lib/Differentiator/DiffPlanner.cpp
PetroZarytskyi: Thank you, removed.

lib/Differentiator/DiffPlanner.cpp
PetroZarytskyi: I have now renamed it as the first commit in the PR.

lib/Differentiator/BaseForwardModeVisitor.cpp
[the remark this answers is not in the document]
PetroZarytskyi: Thank you! Added ``#include "clang/Basic/LLVM.h"`` and removed ``#include "llvm/Support/Casting.h"``.

include/clad/Differentiator/DiffPlanner.h
PetroZarytskyi: I added a comment.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -478,6 +479,43 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
...
+        // that satisfies the given diff request. Now, we perform the
+        // overload resolution using Sema::ActOnCallExpr to make sure we
+        // follow the c++ standard.
+        llvm::SmallVector<const ValueDecl*, 4> diffParams{};
+        for (const DiffInputVarInfo& VarInfo : request.DVI)
+          diffParams.push_back(VarInfo.param);
+        QualType DerivativeType =
+            utils::GetDerivativeType(m_Sema, request.Function, request.Mode,
+                                     diffParams, /*moveBaseToParams=*/true);
+        // Generate dummy inits
```
vgvassilev, line 494: Can you expand here. Like what we generate and what's the idea of the dummy inits.

include/clad/Differentiator/BuiltinDerivatives.h
PetroZarytskyi: After the PR, I started getting this error:
```
error: non-constant-expression cannot be narrowed from type 'double' to 'float' in initializer list
```
in the test
```
double f_exp2f_darg0(double x) {
    double _d_x = 1;
    {{.*}}ValueAndPushforward<float, float> _t0 = {{.*}}exp2f_pushforward(x, _d_x);
    //      ^ used to be ValueAndPushforward<double, double>
    return _t0.pushforward;
}
```
I'm not sure if it's supposed to arise, but the message looks reasonable. Should I look into this deeper?

vgvassilev: Probably worth understanding why? For some reason we pick the float overload of `std::exp2`.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -928,23 +927,46 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
...
+        if (Specialization && !Specialization->isTemplated() &&
+            Specialization->getReturnType()->isUndeducedAutoType()) {
+          overloadFound = true;
+          break;
+        }
+      }
+    }
+    if (overloadFound) {
+      CXXScopeSpec SS;
+      return S.BuildDeclarationNameExpr(SS, Found, /*ADL*/ false).get();
```
vgvassilev:
```suggestion
      return S.BuildDeclarationNameExpr(SS, Found, /*ADL=*/false).get();
```

vgvassilev (approved):
LGTM. Please turn the commits into an atomic set and then feel free to land it.

include/clad/Differentiator/BuiltinDerivatives.h
PetroZarytskyi: The error arises because ``std::exp2f`` is used and ``exp2f_pushforward`` is templated and doesn't work well with float. The same error can be reproduced by hand by ``exp2f_pushforward(1.F, 2.F)``.
The other question is that we picked the overload ``exp2f_pushforward(float, float)`` and not ``exp2f_pushforward(double, double)``, which clang would pick based on args ``x`` and ``_d_x``.

vgvassilev: I think during our voice call we concluded it was fine.

lib/Differentiator/DiffPlanner.cpp
PetroZarytskyi: Fixed.

lib/Differentiator/DerivativeBuilder.cpp
PetroZarytskyi: Good idea. I added more documentation.

lib/Differentiator/DiffPlanner.cpp
PetroZarytskyi: Agree. I added more documentation.

include/clad/Differentiator/BuiltinDerivatives.h
PetroZarytskyi: Yes, I summarized it here just in case.

PR #1402, opened 2025-06-09
Implement more descriptive pretty stack printing for generated derivatives.

vgvassilev:
This prints:
```
2.	Building code for '<double wrapper_function(double x, double (*func)(double))>[name=wrapper_function, order=1, mode=pushforward, args='', tbr]'
 #0 0x0000000104d4c680 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/Users/vvassilev/workspace/builds/llvm-project-clad/bin/clang-18+0x102c14680)
 #1 0x0000000104d4cc64 PrintStackTraceSignalHandler(void*) (/Users/vvassilev/workspace/builds/llvm-project-clad/bin/clang-18+0x102c14c64)
 #2 0x0000000104d4a91c llvm::sys::RunSignalHandlers() (/Users/vvassilev/workspace/builds/llvm-project-clad/bin/clang-18+0x102c1291c)
 #3 0x0000000104d4dd80 SignalHandler(int) (/Users/vvassilev/workspace/builds/llvm-project-clad/bin/clang-18+0x102c15d80)
 #4 0x0000000185424624 (/usr/lib/system/libsystem_platform.dylib+0x1804ac624)
...
```

@PetroZarytskyi, can you help me put what would help us make more sense out of a stack trace? I thought it might be a good idea to print also the last statement of the forward and reverse passes. Would that help?


PetroZarytskyi:
Probably not for this PR, but I think we need to include DVI in the DiffRequest dump. I did it locally while working on static graph PRs as it's crucial for matching requests and caused the most trouble.

vgvassilev:
> This functionality looks very useful. At first, I thought it would be amazing to see the last original statement we started visiting (ideally, with location). This would be enough info for 80% of the cases when I need to run debuggers. But perhaps this would be hard to implement. Dumping the last generated stmts seems like a great idea as it does pretty much the same job.

Did you mean:
```
2.	Building code for '<double wrapper_function(double x, double (*func)(double))>[name=wrapper_function, order=1, mode=pushforward, args='', tbr]'
While visiting <IntegerLiteral> [ '/Users/vvassilev/workspace/sources/clad/test/Misc/Test.cpp:11:22', '/Users/vvassilev/workspace/sources/clad/test/Misc/Test.cpp:11:22']

Begin Stmt Dump ---
2
End Stmt Dump ---
```

include/clad/Differentiator/BaseForwardModeVisitor.h
```diff
@@ -39,6 +39,11 @@ class BaseForwardModeVisitor
   ///
   DerivativeAndOverload Derive() override;
 
+  virtual StmtDiff Visit(const clang::Stmt* S) {
+    m_CurVisitedStmt = S;
+    return clang::ConstStmtVisitor<BaseForwardModeVisitor, StmtDiff>::Visit(S);
+  }
+
```
PetroZarytskyi, line 51: Do we need this to be a virtual function?

vgvassilev: Good point. Removed the virtual keyword.

include/clad/Differentiator/BaseForwardModeVisitor.h
```diff
@@ -39,6 +39,16 @@ class BaseForwardModeVisitor
   ///
   DerivativeAndOverload Derive() override;
 
+  StmtDiff Visit(const clang::Stmt* S) {
+    m_CurVisitedStmt = S;
+#ifndef NDEBUG
+    // Enable testing of the pretty printing of the state when clad crashes.
+    if (const char* Env = std::getenv("CLAD_FORCE_CRASH"))
+      std::terminate();
```
vgvassilev, line 47: This is tested but only run in debug mode. Our codecov infrastructure does not do run in debug mode that's why it complains.

PR #1404, opened 2025-06-11
Fix issue #793: Implement slab based tape

### LINKED ISSUE #793: Do not relocate the tape elements

aditimjoshi:
Fixes #793. Modifies tape implementation to a slab based structure instead of relocating elements each time the tape grows in size.

### DIFF
--- a/include/clad/Differentiator/Tape.h
+++ b/include/clad/Differentiator/Tape.h
@@ -1,168 +1,275 @@
 #ifndef CLAD_TAPE_H
 #define CLAD_TAPE_H
 
+#include "clad/Differentiator/CladConfig.h"
 #include <cassert>
+#include <cstddef>
 #include <cstdio>
+#include <iterator>
 #include <memory>
+#include <new>
 #include <type_traits>
 #include <utility>
-#include "clad/Differentiator/CladConfig.h"
 
 namespace clad {
-  /// Dynamically-sized array (std::vector-like), primarily used for storing
-  /// values in reverse-mode AD inside loops.
-  template <typename T>
-  class tape_impl {
-    T* _data = nullptr;
-    std::size_t _size = 0;
-    std::size_t _capacity = 0;
-  public:
-    using reference = T&;
-    using const_reference = const T&;
-    using pointer = T*;
-    using const_pointer = const T*;
 
-    using size_type = std::size_t;
-    using difference_type = std::ptrdiff_t;
-    using value_type = T;
-    using iterator = pointer;
-    using const_iterator = const_pointer;
+template <typename T, std::size_t SBO_SIZE, std::size_t SLAB_SIZE>
+class tape_impl;
 
-    CUDA_HOST_DEVICE ~tape_impl(){
-      destroy(begin(), end());
-      // delete the old data here to make sure we do not leak anything.
-      ::operator delete(const_cast<void*>(
-            static_cast<const volatile void*>(_data)));
-    }
+/// A forward iterator for traversing elements in `clad::tape_impl`.
+/// This iterator supports standard forward iteration operations, including:
+/// - Dereferencing (`*`, `->`)
+/// - Increment (`++`)
+/// - Equality and inequality comparisons
+template <typename T, std::size_t SBO_SIZE = 64, std::size_t SLAB_SIZE = 1024>
+class tape_iterator {
+  using tape_t = clad::tape_impl<T, SBO_SIZE, SLAB_SIZE>;
+  tape_t* m_tape;
+  std::size_t m_index;
 
-    /// Move values from old to new storage
-    CUDA_HOST_DEVICE T* AllocateRawStorage(std::size_t _capacity) {
-      #ifdef __CUDACC__
-        // Allocate raw storage (without calling constructors of T) of new capacity.
-        T* new_data = static_cast<T*>(::operator new(_capacity * sizeof(T)));
-      #else
-        T *new_data =
-          static_cast<T *>(::operator new(_capacity * sizeof(T), std::nothrow));
-      #endif
-      return new_data;
-    }
+public:
+  using iterator_category = std::forward_iterator_tag;
+  using value_type = T;
+  using difference_type = std::ptrdiff_t;
+  using pointer = T*;
+  using reference = T&;
 
-    /// Add new value of type T constructed from args to the end of the tape.
-    template <typename... ArgsT>
-    CUDA_HOST_DEVICE void emplace_back(ArgsT&&... args) {
-      if (_size >= _capacity)
-        grow();
-      ::new (const_cast<void*>(static_cast<const volatile void*>(end())))
-          T(std::forward<ArgsT>(args)...);
-      _size += 1;
-    }
+  CUDA_HOST_DEVICE tape_iterator() : m_tape(nullptr), m_index(0) {}
+  CUDA_HOST_DEVICE tape_iterator(tape_t* tape, std::size_t index)
+      : m_tape(tape), m_index(index) {}
 
-    CUDA_HOST_DEVICE std::size_t size() const { return _size; }
-    CUDA_HOST_DEVICE iterator begin() {
-      return reinterpret_cast<iterator>(_data);
-    }
-    CUDA_HOST_DEVICE const_iterator begin() const {
-      return reinterpret_cast<const_iterator>(_data);
-    }
-    CUDA_HOST_DEVICE iterator end() {
-      return reinterpret_cast<iterator>(_data) + _size;
-    }
-    CUDA_HOST_DEVICE const_iterator end() const {
-      return reinterpret_cast<const_iterator>(_data) + _size;
-    }
+  CUDA_HOST_DEVICE reference operator*() const { return (*m_tape)[m_index]; }
 
-    /// Access last value (must not be empty).
-    CUDA_HOST_DEVICE reference back() {
-      assert(_size);
-      return begin()[_size - 1];
-    }
-    CUDA_HOST_DEVICE const_reference back() const {
-      assert(_size);
-      return begin()[_size - 1];
-    }
+  CUDA_HOST_DEVICE pointer operator->() const { return &(*m_tape)[m_index]; }
 
-    CUDA_HOST_DEVICE reference operator[](std::size_t i) {
-      assert(i < _size);
-      return begin()[i];
+  CUDA_HOST_DEVICE tape_iterator& operator++() {
+    ++m_index;
+    return *this;
+  }
+
+  CUDA_HOST_DEVICE tape_iterator operator++(int) {
+    tape_iterator tmp = *this;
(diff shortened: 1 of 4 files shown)


include/clad/Differentiator/Tape.h
```diff
@@ -9,159 +9,135 @@
...
+          while (last->next) last = last->next;
+          last->next = new_slab;
+        }
+      }
+
+      Slab* slab = _head;
+      size_t i = _size / slab_size;
+      while (i--) slab = slab->next;
+
+      ::new (const_cast<void*>(static_cast<const volatile void*>(slab->elements() + (_size % slab_size)))) T(std::forward<ArgsT>(args)...);
```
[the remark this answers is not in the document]
vgvassilev: There must be some idiom that's used in STL for such operations instead of these casts...

include/clad/Differentiator/Tape.h
```diff
@@ -9,160 +9,185 @@
 #include "clad/Differentiator/CladConfig.h"
 
 namespace clad {
-  /// Dynamically-sized array (std::vector-like), primarily used for storing
-  /// values in reverse-mode AD inside loops.
   template <typename T>
   class tape_impl {
-    T* _data = nullptr;
+    struct Slab {
+      alignas(T) char raw_data[1024 * sizeof(T)];
```
vgvassilev: Please move the `1024` into a constant explaining why it was picked -- probably move the constexpr variable a bit up.

include/clad/Differentiator/Tape.h
```diff
@@ -9,160 +9,185 @@
 #include "clad/Differentiator/CladConfig.h"
 
 namespace clad {
-  /// Dynamically-sized array (std::vector-like), primarily used for storing
-  /// values in reverse-mode AD inside loops.
```
vgvassilev, line 13: We should also add some documentation for the new implementation.

include/clad/Differentiator/Tape.h
vgvassilev: Can you also test with bigger objects where `sizeof(T)` influences the overall size?

include/clad/Differentiator/Tape.h
```diff
@@ -9,160 +9,185 @@
...
   template <typename T>
```
vgvassilev: Here we should take `SBO_SIZE` and `SLAB_SIZE` as parameters with default values.

include/clad/Differentiator/Tape.h
```diff
@@ -9,160 +9,185 @@
...
   class tape_impl {
-    T* _data = nullptr;
+    struct Slab {
+      alignas(T) char raw_data[1024 * sizeof(T)];
+      Slab* next;
+      Slab() : next(nullptr) {}
```
kchristin22: Hi, great work! I tested it on CUDA, and to compile it should be:
```
CUDA_HOST_DEVICE Slab() : next(nullptr) {}
```

include/clad/Differentiator/Tape.h
```diff
@@ -9,160 +9,185 @@
...
+      T* elements() { return reinterpret_cast<T*>(raw_data); }
```
kchristin22: same here:
```
CUDA_HOST_DEVICE T *elements() { return reinterpret_cast<T*>(raw_data); } 
```

include/clad/Differentiator/Tape.h
```diff
@@ -9,139 +9,155 @@
...
-      _size += 1;
+      if (_size < SBO_SIZE) {
+        // Store in SBO buffer
+        ::new (const_cast<void*>(static_cast<const volatile void*>(sbo_elements() + _size)))
+            T(std::forward<ArgsT>(args)...);
+      } else {
+        // Transition to dynamic storage if needed
+        if (m_using_sbo) {
+          m_using_sbo = false;
+        }
```
vgvassilev:
```suggestion
        if (m_using_sbo)
          m_using_sbo = false;
```

include/clad/Differentiator/Tape.h
```diff
@@ -9,139 +9,155 @@
 #include "clad/Differentiator/CladConfig.h"
 
 namespace clad {
-  /// Dynamically-sized array (std::vector-like), primarily used for storing
-  /// values in reverse-mode AD inside loops.
-  template <typename T>
+  /// A dynamic slab-based vector-like container with Small Buffer Optimization (SBO),
+  /// primarily used for storing values in reverse-mode AD.
+  /// Stores elements in a static buffer first, then falls back to dynamically
+  /// allocated linked slabs if capacity exceeds SBO.
```
vgvassilev, line 15:
```suggestion
  /// allocated linked slabs if capacity exceeds small buffer optimization (SBO).
```

include/clad/Differentiator/Tape.h
```diff
@@ -1,168 +1,252 @@
...
-      // allocation properly.
-      for (; first != last; ++first, (void)++current) {
-        ::new (const_cast<void*>(
-            static_cast<const volatile void*>(clad_addressof(*current))))
-          T(std::move(*first));
-      }
+      // Construct element in-place
+      ::new (const_cast<void*>(static_cast<const volatile void*>(
+          slab->elements() + ((_size - SBO_SIZE) % SLAB_SIZE))))
+          T(std::forward<ArgsT>(args)...);
```
vgvassilev:
```suggestion
      T* location = slab->elements() + ((_size - SBO_SIZE) % SLAB_SIZE);
      ::new (static_cast<void*>(location)) T(std::forward<ArgsT>(args)...);
```
Can we do something like this?

include/clad/Differentiator/Tape.h
```diff
@@ -1,168 +1,252 @@
...
+
+  /// Remove the last value from the tape.
+  CUDA_HOST_DEVICE void pop_back() {
+    assert(_size);
+    _size--;
+    at(_size)->~T();
+  }
+
+private:
+  // Returns pointer to element at specified index, handling SBO or slab lookup
```
vgvassilev:
```suggestion
  /// Returns pointer to element at specified index, handling SBO or slab lookup
```

include/clad/Differentiator/Tape.h
```diff
@@ -1,168 +1,252 @@
...
+/// slabs if capacity exceeds SBO.
+template <typename T, std::size_t SBO_SIZE = 64, std::size_t SLAB_SIZE = 1024>
+class tape_impl {
+  /// A block of contiguous storage allocated dynamically when SBO capacity is
+  /// exceeded.
+  struct Slab {
+    alignas(T) char raw_data[SLAB_SIZE * sizeof(T)]{};
+    Slab* next;
+    CUDA_HOST_DEVICE Slab() : next(nullptr) {}
+    CUDA_HOST_DEVICE T* elements() { return reinterpret_cast<T*>(raw_data); }
```
vgvassilev: From c++17 onward we can/should use:
`return std::launder(reinterpret_cast<T*>(addr));`

Can we have an ifdef?

include/clad/Differentiator/Tape.h
```diff
@@ -1,168 +1,252 @@
...
+  };
 
-    CUDA_HOST_DEVICE const_reference operator[](std::size_t i) const {
-      assert(i < _size);
-      return begin()[i];
-    }
+  alignas(T) char m_static_buffer[SBO_SIZE * sizeof(T)];
```
vgvassilev:
```suggestion
  std::aligned_storage_t<sizeof(T), alignof(T)> data[N];
```
Maybe something like that?

vgvassilev (commented):
We are getting very close to merging this.

include/clad/Differentiator/Tape.h
aditimjoshi: I'll do this for both the sbo_elements() functions as well right?

include/clad/Differentiator/Tape.h
aditimjoshi: I tried this but it won't work unless we add (const_cast<void*>(static_cast<const volatile void*>(location))). For just (static_cast<void*>(location)) it gives the error 
error: static_cast from 'const int *' to 'void *' is not allowed

include/clad/Differentiator/Tape.h
vgvassilev: Yes, please.

include/clad/Differentiator/Tape.h
aditimjoshi: std::aligned_storage_t was introduced in c++-14. I think version 11 had std::aligned_storage but it's going to be deprecated in later versions.

include/clad/Differentiator/Tape.h
```diff
@@ -1,168 +1,252 @@
...
+    return m_index != other.m_index;
+  }
+};
+
+/// A dynamic slab-based vector-like container with Small Buffer Optimization
+/// (SBO), primarily used for storing values in reverse-mode AD. Stores elements
+/// in a static buffer first, then falls back to dynamically allocated linked
+/// slabs if capacity exceeds SBO.
+template <typename T, std::size_t SBO_SIZE = 64, std::size_t SLAB_SIZE = 1024>
+class tape_impl {
```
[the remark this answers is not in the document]
vgvassilev: @aditimjoshi, do we want to try to address this?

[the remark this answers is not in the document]
aditimjoshi: The old tape implementation had also defined a destructor without defining the rest so I was thought it would be fine but if there's a case I missed where it will cause errors or if it seems better to address it then I can add it.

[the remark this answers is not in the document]
vgvassilev: Let's write them.

include/clad/Differentiator/Tape.h
```diff
@@ -1,168 +1,330 @@
...
+  using const_reference = const T&;
+  using pointer = T*;
+  using const_pointer = const T*;
+  using size_type = std::size_t;
+  using difference_type = std::ptrdiff_t;
+  using value_type = T;
+  using iterator = tape_iterator<T, SBO_SIZE, SLAB_SIZE>;
+  using const_iterator = tape_iterator<const T, SBO_SIZE, SLAB_SIZE>;
+
+  CUDA_HOST_DEVICE tape_impl() = default;
```
vgvassilev, line 125: That's enough -- we should not encourage people to copy the tape. Perhaps once we have a valid use case (so far we don't) we can implement the move operations. Let's keep only this line and delete the other operators with a comment explaining that we don't want the tape to be copied (AD does not generally need that).

PR #1406, opened 2025-06-13
Call UpdateDiffParamsInfo in the Diffplanner unconditionally

PetroZarytskyi:
Before this PR, ``UpdateDiffParamsInfo`` was called either in the ``Diffplanner`` or ``ProcessDiffRequest`` depending on certain conditions. This leads to inconsistency in when ``DVI`` (diff params) is built and whether it's accessible in the ``DiffPlanner``. Also, some diagnostics are done in ``UpdateDiffParamsInfo`` and could therefore be accidentally duplicated. This PR makes the call to ``UpdateDiffParamsInfo`` in the ``Diffplanner`` unconditional, and only leaves other calls for dynamic derivative scheduling and high-order derivatives.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -322,6 +322,11 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
     auto& C = semaRef.getASTContext();
     const Expr* diffArgs = Args;
     const FunctionDecl* FD = Function;
+    // On this stage, templated functions are not yet fully instantiated.
+    // Uninstantiated functions lack much information like we need: they don't
+    // have bodies and information about defition. Therefore, we have to force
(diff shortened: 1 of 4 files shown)


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -322,6 +322,7 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
     auto& C = semaRef.getASTContext();
     const Expr* diffArgs = Args;
     const FunctionDecl* FD = Function;
+    semaRef.PerformPendingInstantiations();
```
vgvassilev, line 329: Can you add a comment why we need this?

PetroZarytskyi: I added one. Does it look good?

vgvassilev (approved):
Lgtm!

PR #1407, opened 2025-06-13
Update the SmallPT ray tracer demo

PetroZarytskyi:
The current demo of the SmallPT ray tracer doesn't work for multiple reasons:
1) One of the gradients was implemented with forward mode. There was a mistake that all three derivatives of the function ``distance_func`` were taken w.r.t. the ``x`` parameter and not ``x``, ``y``, and ``z``. After that, the vector was normalized. In other words, the result is always either ``(1, 1, 1)`` or ``(-1, -1, -1)`` (or error if the derivative is exactly 0). So Clad didn't do any useful job in the demo.
2) The demo expects Clad to differentiate virtual functions in a specific way, which it currently doesn't seem to be able to do. i.e.
```
class BaseClass {
    virtual double distance_func(...) {...}
   
    // In theory, we want this method to work in derived classes and
    // use a virtual `distance_func_grad` but this doesn't work
    Vec normal(...) {
        auto df = clad::gradient(distance_func);
        ...
    }
}
```
First, this PR replaces forward mode with reverse, which is more useful for gradients. Second, all the demo now produces correct results.
This PR leaves a couple of FIXMEs: 
1) We don't have normal way to call ``CladFunction`` of member functions as ``CladFunction::execute`` doesn't have a way to provide a base for the method. Once it's supported, we can get rid of helper functions like ``sphere_distance_func `` and directly use the virtual member function ``distance_func``. 
2) Polymorphism is essentially unsupported, so the ``normal`` (the one that calls ``clad::gradient``) method function has to be rewritten in every derived class.


PetroZarytskyi:
Also, we might want to modify the demo itself, as it defines two shapes (``Sphere`` and ``HyperbolicSolid``) and only uses one (``Sphere``).

demos/ComputerGraphics/smallpt/SmallPT.cpp
```diff
@@ -112,25 +107,9 @@ class ImplicitSolid : public Solid {
...
-    //double Nx = distance_func_dx.execute(*this, pt.x, pt.y, pt.z);
-    //double Ny = distance_func_dy.execute(*this, pt.x, pt.y, pt.z);
-    //double Nz = distance_func_dz.execute(*this, pt.x, pt.y, pt.z);
-    double Nx = distance_func_darg0(pt.x, pt.y, pt.z);
-    double Ny = distance_func_darg1(pt.x, pt.y, pt.z);
-    double Nz = distance_func_darg2(pt.x, pt.y, pt.z);
-
-    return Vec(Nx, Ny, Nz).norm();
-  }
+  // FIXME: Once Clad fully supports virtual methods,
```
vgvassilev, line 108: Can we track this in an issue?

demos/ComputerGraphics/smallpt/SmallPT.cpp
```diff
@@ -190,20 +163,14 @@ class Sphere : public ImplicitSolid {
...
-
-    return Vec(Nx, Ny, Nz).norm();
-    //return Vec(Nx, Ny, Nz); // nabla f of signed distance functions is always unit vector
+    Vec result{};
+    Vec dummy{};
+    // FIXME: We should differentiate distance_func directly once clad
+    // provides functionality to call CladFunction of member functions.
+    auto dist_grad = clad::gradient(sphere_distance_func, "x, y, z, p");
+    dist_grad.execute(pt.x, pt.y, pt.z, p, r, &result.x, &result.y, &result.z,
+                      &dummy);
```
alexander-penev: The function _sphere_distance_func_ is used to describe a sphere with a given center and a given radius. We then look for the gradient of this function at a given point (_x_, _y_, _z_). This is necessary to find the implicit geometry (i.e. sphere).
```sphere_distance_func(double x, double y, double z, const Vec &p, double r)```
The parameters _p_ and _r_ are parameters describing the sphere, i.e. center and radius, and at least for what the function is used for now, we don't need to differentiate by them.

When this code was written, **clad** had many limitations, so the function has parameters _x_, _y_, _z_ instead of a vector _V_, for example. If now we can differentiate by a vector, it would be better for the interface of this and similar functions to become:
```sphere_distance_func(Vec p, const Vec &p0, const double r)```

PetroZarytskyi: Agreed. I'll change x, y, z to a single vector.

demos/ComputerGraphics/smallpt/SmallPT.cpp
PetroZarytskyi: I opened issues #1409, #1410, and #1411.

demos/ComputerGraphics/smallpt/SmallPT.cpp
PetroZarytskyi: I changed all x, y, z triplets to single Vec.

alexander-penev (approved):

PR #1408, opened 2025-06-13
Schedule simple partial derivatives statically

### LINKED ISSUE #1371: Performance regression in generating Higgs combination gradients with Clad 1.10
See the following explanation if the issue:
https://github.com/root-project/root/pull/18383#issuecomment-2848773894

Please let me know how I can help, besides building a standalone reproducer, which I am doing now.

PetroZarytskyi:
This is a draft PR because it's implemented on top of #1406.

This PR adds simple propagation of parameter non-differentiability to nested functions. This system only works with array parameters, which currently cause all partial pullbacks. This system only recognizes decl refs of non-differentiable parameters, but nothing more complicated. The general problem can only be solved with activity analysis. For example,
```
double f (const double* x, double y) { // x is non-diff
    inner_func(x, y); // We propagate non-differentiability of x to inner_func
}
```
but
```
double f (const double* x, double y) { // x is non-diff
    const double* z = x + 4; // The current system still doesn't understand that z is also non-diff
    inner_func(z, y); // Non-differentiability is NOT propagated
}
```
This is especially important for efficiency of Clad because currently when partial pullbacks are needed, this leads to derivatives being generated twice, as shown in #1371.
Fixes #1371.


PetroZarytskyi:
The decrease in coverage happened because a case with ``noOverloadExists`` in ``DerivativeBuilder::LookupCustomDerivativeOrNumericalDiff`` is now uncovered. This is part of the dynamical scheduling system, which we wanted to get rid of. I think it's not a good idea to remove the line but I also don't know if we should look for corner examples to cover this.

test/CUDA/GradientKernels.cu
```diff
@@ -336,19 +336,14 @@ __global__ void dup_kernel_with_device_call_2(double *out, const double *in, dou
...
+// CHECK: __attribute__((device)) void device_fn_2_pullback_1(const double *in, double val, double _d_y, double *_d_val) {
 //CHECK-NEXT:    unsigned int _t1 = blockIdx.x;
 //CHECK-NEXT:    unsigned int _t0 = blockDim.x;
 //CHECK-NEXT:    int _d_index = 0;
 //CHECK-NEXT:    int index0 = threadIdx.x + _t1 * _t0;
-//CHECK-NEXT:    {
-//CHECK-NEXT:        _d_in[index0] += _d_y;
-//CHECK-NEXT:        *_d_val += _d_y;
-//CHECK-NEXT:    }
+//CHECK-NEXT:    *_d_val += _d_y;
```
vgvassilev, line 344: Are we sure this change is good?

PetroZarytskyi: I think the way this diff looks is a little misleading. There used to be 2 pullbacks ``device_fn_2_pullback_1``: the one with the full DVI was scheduled statically and never used, the one with partial DVI was scheduled dynamically and used in a function below. After this PR, the pullback with full DVI is not scheduled, and that same pullback with partial DVI is now scheduled statically.

vgvassilev (approved):
Lgtm!

PR #1416, opened 2025-06-19
Update call argument storing in the reverse mode

PetroZarytskyi:
This PR moves the storing system of ref-type call arguments to the same framework as the rest of Clad. This leads to 2 changes in the produced code:
1) Only original variables (not temporaries `_t`) are used as pullback args
```
// forward pass
_t0 = a;
modify(_t0); // a is modified so we store it
...
// reverse pass
a = _t0;
modify_pullback(_t0, ...)
modify_pullback(a, ...);
```
This is safe because pullbacks have the same storing system as the gradient, meaning they are responsible for restoring a to its original state.

2) In loops, the tape is handled a little differently:
```
a = clad::back(_t0);
modify_pullback(a, ...);
clad::pop(_t0);
```
became

```
a = clad::pop(_t0);
modify_pullback(a, ...);
```
This PR also updates the documentation of this part.


vgvassilev (approved):
Lgtm!

PR #1417, opened 2025-06-22
Avoid repetetive CFG construction

ovdiiuv:
Before this PR we used to build CFG after entering each analysis. Instead we now use **AnalysisDeclContext** to pass down to the analyzers. TBR call is moved to the DiffCollector. Since TBR was never enabled for nested functions and would fail, it is explicitly disabled for non topmost requests for now.


include/clad/Differentiator/DiffPlanner.h
```diff
@@ -56,6 +57,7 @@ struct DiffRequest {
 public:
   /// Function to be differentiated.
   const clang::FunctionDecl* Function = nullptr;
+  clang::AnalysisDeclContext* ADContext = nullptr;
```
vgvassilev: In an AD system `ADContext` probably does not mean what it does for clang. Probably we should use `AnalysisDC` or `AnalysisCtx`. And more broadly don't we want a reference to the `AnalysisDeclContextManager`

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1312,9 +1319,22 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
     if (m_Sema.isStdInitializerList(recordTy, /*elemType=*/nullptr))
       return true;
 
-    if (!LookupCustomDerivativeDecl(request))
+    if (!LookupCustomDerivativeDecl(request)) {
+      AnalysisDeclContextManager* ADCM = new AnalysisDeclContextManager(request.Function->getASTContext());
```
vgvassilev: Is there a way to make this cover all of the diff plan for a given request?

include/clad/Differentiator/DiffPlanner.h
vgvassilev: This should probably be a unique_ptr...

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1213,22 +1211,31 @@
       request.Function = request.Function->getDefinition();
 
     if (!LookupCustomDerivativeDecl(request)) {
-      if (m_TopMostReq->EnableVariedAnalysis &&
-          m_TopMostReq->Mode == DiffMode::reverse) {
+      // Create analysis manager once for all analyses.
+      AnalysisDeclContextManager* ADCM = new AnalysisDeclContextManager(request.Function->getASTContext());
```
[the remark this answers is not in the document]
PetroZarytskyi: Can we use a unique pointer?

lib/Differentiator/TBRAnalyzer.cpp
```diff
@@ -280,18 +290,27 @@ void TBRAnalyzer::copyVarToCurBlock(const clang::VarDecl* VD) {
 }
 
 void TBRAnalyzer::addVar(const clang::VarDecl* VD, bool forceNonRefType) {
+  llvm::errs() << "\n==============21\n";
```
PetroZarytskyi: Can you remove the debug stmts here and below?

lib/Differentiator/TBRAnalyzer.cpp
```diff
@@ -586,13 +606,21 @@ bool TBRAnalyzer::VisitDeclRefExpr(DeclRefExpr* DRE) {
...
       addVar(VD);
+      llvm::errs() << "\n===========3\n";
+
       if (clang::Expr* init = VD->getInit()) {
+
         setMode(Mode::kMarkingMode);
         TraverseStmt(init);
         resetMode();
+
         auto& VDExpr = getCurBlockVarsData()[VD];
```
PetroZarytskyi, line 576: Up to here

lib/Differentiator/TBRAnalyzer.cpp
```diff
@@ -219,10 +219,20 @@ TBRAnalyzer::VarData::VarData(QualType QT, const ASTContext& C,
...
+
+    // FIXME: For some reason if a variable is of kokkos-type we enter an
+    // infinite loop.
+    bool isInKokkosNS = false;
+    if (const auto* ns =
+            llvm::dyn_cast<clang::NamespaceDecl>(recordDecl->getDeclContext()))
+      if (ns->getName() == "Kokkos")
+        isInKokkosNS = true;
+
+    if (!isInKokkosNS)
```
PetroZarytskyi: Can you open a brief issue mentioning this?

PetroZarytskyi (changes requested):
Please address clang-tidy complaints about the headers and memory ownership (I mean, make ADContext a unique_ptr). Other than that, the PR seems good.

include/clad/Differentiator/DiffPlanner.h
ovdiiuv: I would assume so, but `AnalysisDeclContextManager` from what I have seen isn't a manager really.

ovdiiuv: I thought of it, but then copying DiffRequests would not be possible.

lib/Differentiator/DiffPlanner.cpp
ovdiiuv: No, it is left for the case we don't schedule a diffrequest statically. Right now we do not have any tests that have dynamic pullbacks as far as I know.

include/clad/Differentiator/DiffPlanner.h
PetroZarytskyi: Oh, I guess you're right.
@vgvassilev should we do shared_ptr or raw?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1213,22 +1211,33 @@
       request.Function = request.Function->getDefinition();
 
     if (!LookupCustomDerivativeDecl(request)) {
-      if (m_TopMostReq->EnableVariedAnalysis &&
-          m_TopMostReq->Mode == DiffMode::reverse) {
+      // Create analysis manager once for all analyses.
+      AnalysisDeclContextManager* ADCM =
```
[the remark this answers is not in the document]
ovdiiuv: Is that something worthy?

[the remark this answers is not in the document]
vgvassilev: Yes, but we probably need to make this a unique_ptr.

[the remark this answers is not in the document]
ovdiiuv: It would be destroyed once it is out of scope then. I could make it a shared_ptr, add it to DiffRequest or keep as is.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -6,13 +6,9 @@ using namespace clang;
...
 void VariedAnalyzer::Analyze(const FunctionDecl* FD) {
-  // Build the CFG (control-flow graph) of FD.
-  clang::CFG::BuildOptions Options;
-  m_CFG = clang::CFG::buildCFG(FD, FD->getBody(), &m_Context, Options);
-
-  m_BlockData.resize(m_CFG->size());
+  m_BlockData.resize(m_AnalysisDC->getCFG()->size());
   // Set current block ID to the ID of entry the block.
-  CFGBlock* entry = &m_CFG->getEntry();
+  CFGBlock* entry = &m_AnalysisDC->getCFG()->getEntry();
```
vgvassilev:
```suggestion
  CFGBlock& entry = m_AnalysisDC->getCFG()->getEntry();
```

lib/Differentiator/TBRAnalyzer.cpp
```diff
@@ -229,6 +229,10 @@ TBRAnalyzer::getIDSequence(const clang::Expr* E,
       if (opCode == UO_Deref)
         IDSequence.push_back(ProfileID());
       E = UO->getSubExpr();
+    } else if (isa<clang::CXXStaticCastExpr>(E)) {
+      // FIXME: We might enter here when moving objects using static_cast, for
+      // example. Check if we need some logic here.
+      break;
```
PetroZarytskyi: I already fixed this locally. Instead of this, you can better replace the line above
```
E = E->IgnoreImplicit();
```
with 
```
E = E->IgnoreCasts();
```
This should help.

lib/Differentiator/TBRAnalyzer.cpp
vgvassilev: Did we address this comment? Could we make the analysis run only on finding reassignments?

PetroZarytskyi: I don't know if this was addressed. @ovdiiuv ?

PetroZarytskyi: This is the only part that concerns me. We enable TBR in some new cases, get an error, and ignore it. This is fine, assuming it's easy to fix. But we should look into this pretty much immediately.

ovdiiuv: I thought we agreed upon keeping TBR as is, but I could do it, yes. @PetroZarytskyi I will open an issue once merged, is that fine with you?

PetroZarytskyi: I think it's fine

PetroZarytskyi (approved):
LGTM!

lib/Differentiator/TBRAnalyzer.cpp
vgvassilev: I agree we should keep the tbr as is means that we should drop the kokkos check and make the analysis on demand as it was.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -6,14 +6,10 @@ using namespace clang;
...
-  // Build the CFG (control-flow graph) of FD.
-  clang::CFG::BuildOptions Options;
-  m_CFG = clang::CFG::buildCFG(FD, FD->getBody(), &m_Context, Options);
-
-  m_BlockData.resize(m_CFG->size());
+  m_BlockData.resize(m_AnalysisDC->getCFG()->size());
   // Set current block ID to the ID of entry the block.
-  CFGBlock* entry = &m_CFG->getEntry();
-  m_CurBlockID = entry->getBlockID();
+  CFGBlock entry = m_AnalysisDC->getCFG()->getEntry();
```
vgvassilev:
```suggestion
  CFGBlock &entry = m_AnalysisDC->getCFG()->getEntry();
```

ovdiiuv: Didn't you ask to remove it?

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -12,6 +12,8 @@
 #include "clad/Differentiator/DerivedFnCollector.h"
 #include "clad/Differentiator/DiffPlanner.h"
 
+#include "../lib/Differentiator/TBRAnalyzer.h"
+
```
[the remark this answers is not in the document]
vgvassilev: Why is this header included here?

include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -87,6 +89,7 @@ namespace clad {
     clad::DynamicGraph<DiffRequest>& m_DiffRequestGraph;
     std::unique_ptr<utils::StmtClone> m_NodeCloner;
     clang::NamespaceDecl* m_BuiltinDerivativesNSD;
+    ContextMap& m_AllAnalysisDC;
```
vgvassilev: Now that you have the manager with the lifetime of the DiffPlanner you can still have the Context* in the diff request, right?

lib/Differentiator/ActivityAnalyzer.cpp
vgvassilev: I asked to remove the pointer and replace it with a reference.

include/clad/Differentiator/DerivativeBuilder.h
ovdiiuv: No need to. We now access it through DerivativeBuilder.

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -123,6 +128,8 @@ struct DiffRequest {
   /// This will be particularly useful for pushforward and pullback functions.
   bool DeclarationOnly = false;
 
+  clang::AnalysisDeclContext* m_AnalysisDC;
```
vgvassilev, line 131: Do we need to edit `bool operator==` taking into account the new member?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1313,9 +1315,17 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
...
 
-    if (!LookupCustomDerivativeDecl(request))
+    if (!LookupCustomDerivativeDecl(request)) {
+      clang::CFG::BuildOptions Options;
+      std::unique_ptr<AnalysisDeclContext> AnalysisDC =
+          std::make_unique<AnalysisDeclContext>(nullptr, request.Function,
+                                                Options);
+
+      request.m_AnalysisDC = AnalysisDC.get();
+      m_AllAnalysisDC.push_back(std::move(AnalysisDC));
```
vgvassilev:
```suggestion
      m_AllAnalysisDC.push_back(std::move(AnalysisDC));
      request.m_AnalysisDC = m_AllAnalysisDC.back().get();    
```

Should be clearer.

lib/Differentiator/TBRAnalyzer.h
```diff
@@ -221,10 +222,7 @@ class TBRAnalyzer : public clang::RecursiveASTVisitor<TBRAnalyzer> {
   /// a new one).
   std::vector<int> m_ModeStack;
 
-  ASTContext& m_Context;
-
-  /// clang::CFG of the function being analysed.
-  std::unique_ptr<clang::CFG> m_CFG;
+  clang::AnalysisDeclContext* m_AnalysisDC;
```
vgvassilev: What about making `TBRAnalyzer::Analyze` take a `DiffRequest` and get rid of this data member?

include/clad/Differentiator/DiffPlanner.h
ovdiiuv: I don't think so, we sometimes construct requests to find the existing ones -- that would make it much more complex.

vgvassilev: Ok, in that case we should add a comment there explaining this so that it does not get added on a later stage.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1214,22 +1214,26 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
       request.Function = request.Function->getDefinition();
 
     if (!LookupCustomDerivativeDecl(request)) {
-      if (m_TopMostReq->EnableVariedAnalysis &&
-          m_TopMostReq->Mode == DiffMode::reverse) {
+      clang::CFG::BuildOptions Options;
+      std::unique_ptr<AnalysisDeclContext> AnalysisDC =
+          std::make_unique<AnalysisDeclContext>(nullptr, request.Function,
```
vgvassilev:
```suggestion
          std::make_unique<AnalysisDeclContext>(/*AnalysisDeclContextManager=*/nullptr, request.Function,
```

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1313,9 +1317,18 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
     if (m_Sema.isStdInitializerList(recordTy, /*elemType=*/nullptr))
       return true;
 
-    if (!LookupCustomDerivativeDecl(request))
+    if (!LookupCustomDerivativeDecl(request)) {
+      clang::CFG::BuildOptions Options;
+      std::unique_ptr<AnalysisDeclContext> AnalysisDC =
+          std::make_unique<AnalysisDeclContext>(nullptr, request.Function,
```
vgvassilev:
```suggestion
          std::make_unique<AnalysisDeclContext>(/*AnalysisDeclContextManager=*/nullptr, request.Function,
```

lib/Differentiator/TBRAnalyzer.h
```diff
@@ -33,7 +35,7 @@ class TBRAnalyzer : public clang::RecursiveASTVisitor<TBRAnalyzer> {
 
   ProfileID getProfileID(const Expr* E) const {
     ProfileID profID;
-    E->Profile(profID, m_Context, /* Canonical */ true);
+    E->Profile(profID, m_ASTContext, /* Canonical */ true);
```
vgvassilev:
```suggestion
    E->Profile(profID, m_ASTContext, /* Canonical=*/true);
```

PR #1418, opened 2025-06-24
Simplify `getExprVarData` in TBR and add a test

PetroZarytskyi:
This PR solves 2 problems related to TBR:
1) ``getExprVarData`` has a parameter ``addNonConstIdx``, which was probably meant for efficiency when building ``VarData``. The difference, however, is likely negligible, as the function will return earlier only under quite specific circumstances. There's no need to overcomplicate the analysis with this.
2) The system handling array elements in TBR is quite complicated, but we don't have a single test that directly checks the produced code. This PR adds a test involving non-constant indices.


vgvassilev (approved):
Lgtm!

PR #1421, opened 2025-06-25
Switch from Visit to Traverse functions in TBR

PetroZarytskyi:
Currently, `TBRAnalyzer` is implemented as a `RecursiveASTVisitor`, meaning it automatically visits all sub-stmts recursively. However, in TBR, we often need to visit sub-stmts in a specific way, e.g.
When analyzing
```
x = y;
```
we should visit ``x`` with settings according to the fact that ``x`` is being overwritten, and then reset the settings. The same goes for ``y``.
Therefore, we cannot perform the analysis in the order, in which `RecursiveASTVisitor` visits this stmt: `x = y`, then `x`, then `y`. Instead, we call ``TraverseStmt`` manually inside ``VisitBinaryOperator``. Because of this we traverse stmts more times than we need to:
```
1) x = y;   ->  also x and y manually
2) x          -> automatically, unnecessary
3) y          -> automatically, unnecessary
```
We can show that for nested stmts, this will bump complexity from linear to quadratic. The same problem also occurs for other stmts like ``CallExpr``, where we visit the ``DeclRefExpr`` of the function.
This can be solved by switching from ``Visit`` functions to ``Traverse`` and returning ``false`` to prevent the visitor from entering the child nodes (returning ``false`` in ``Visit`` functions stops the whole visitation). The only downside is that ``Traverse`` only works with exact node type matches, e.g. ``TraverseCallExpr`` is not called for a ``CXXMemberCallExpr`` even though it's a subclass of ``CallExpr``. 
We can solve this by introducing dummy functions like
```
TBRAnalyzer::TraverseCXXMemberCallExpr(const CXXMemberCallExpr* MCE) {
  TBRAnalyzer::VisitCallExpr(MCE);
  return false;
}
```


lib/Differentiator/TBRAnalyzer.cpp
```diff
@@ -769,10 +770,20 @@ bool TBRAnalyzer::VisitCallExpr(clang::CallExpr* CE) {
     }
   }
   resetMode();
-  return true;
+  return false;
 }
 
-bool TBRAnalyzer::VisitCXXConstructExpr(clang::CXXConstructExpr* CE) {
+bool TBRAnalyzer::TraverseCXXMemberCallExpr(clang::CXXMemberCallExpr* CE) {
+  TBRAnalyzer::TraverseCallExpr(CE);
```
vgvassilev, line 777: Wouldn't that go naturally without explicitly specifying it?

PetroZarytskyi: Sadly, no. As I explained in the end of the PR message, Traverse only works with exact type matches, sub-classes are not visited. For what it's worth, I introduced this after breaking some tests.

vgvassilev: Thanks!

PR #1422, opened 2025-06-25
Move ProfileID generation to a single point in TBR

PetroZarytskyi:
TBR has a complicated logic that parses `Expr`. There is a function `getExprVarData` that converts it to the type `VarData` that represents information about variables, e.g.
```getExprVarData:  arr[4].k  -->  VarData{arr, 4, k}```
However, VarData does not store information about every element in the sequence, rather points to the end of it. While this is enough as an identifier of the underlying value, it's not enough for more complex analysis. The function `overlay` needs the whole sequence and so it performs its own parsing. This has lead to bugs like #1369, where certain cases were accounted for in `getExprVarData` but not in `overlay`, which is used less frequently. This PR moves the parsing of `Expr` to a single point in `getIDSequence`. `getExprVarData` then uses the sequence to get the VarData. 

I think we should also consider dropping `VarData` altogether and using the sequence of IDs as the identifier.

### DIFF
--- a/lib/Differentiator/TBRAnalyzer.cpp
+++ b/lib/Differentiator/TBRAnalyzer.cpp
@@ -27,7 +27,7 @@ void TBRAnalyzer::setIsRequired(VarData& varData, bool isReq) {
     for (auto& pair : *varData.m_Val.m_ArrData)
       setIsRequired(pair.second, isReq);
   else if (varData.m_Type == VarData::REF_TYPE && varData.m_Val.m_RefData)
-    if (auto* data = getExprVarData(varData.m_Val.m_RefData))
+    if (auto* data = getVarDataFromExpr(varData.m_Val.m_RefData))
       setIsRequired(*data, isReq);
 }
 
@@ -81,7 +81,7 @@ bool TBRAnalyzer::findReq(const VarData& varData) {
       if (findReq(pair.second))
         return true;
   } else if (varData.m_Type == VarData::REF_TYPE && varData.m_Val.m_RefData) {
-    if (auto* data = getExprVarData(varData.m_Val.m_RefData)) {
+    if (auto* data = getVarDataFromExpr(varData.m_Val.m_RefData)) {
       if (findReq(*data))
         return true;
     }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lib/Differentiator/TBRAnalyzer.cpp [hunk 1]
```diff
@@ -225,20 +196,41 @@ void TBRAnalyzer::overlay(const clang::Expr* E) {
...
+          continue;
+        }
       }
-      innermostDRE = DRE;
-      cond = false;
-    } else
-      return;
+      innerVD = VD;
+      break;
+      return VD;
```
vgvassilev:
```suggestion
      return VD;
```

lib/Differentiator/TBRAnalyzer.cpp
```diff
@@ -579,9 +571,15 @@ bool TBRAnalyzer::TraverseDeclStmt(DeclStmt* DS) {
         // VarData of the RHS variable.
         llvm::SmallVector<Expr*, 4> ExprsToStore;
         utils::GetInnermostReturnExpr(init, ExprsToStore);
-        if (VDExpr.m_Type == VarData::REF_TYPE && !ExprsToStore.empty())
-          // NOLINTNEXTLINE(cppcoreguidelines-pro-type-union-access)
-          VDExpr.m_Val.m_RefData = ExprsToStore[0];
+        if (VDExpr.m_Type == VarData::REF_TYPE) {
+          if (ExprsToStore.size() == 1)
```
vgvassilev, line 576: It's worth having a comment why size of 1 is treated specially.

PetroZarytskyi: I added a comment.

lib/Differentiator/TBRAnalyzer.cpp
PetroZarytskyi: The original code indeed doesn't make sense because the return is inaccessible. I removed it now. 
Do you want to have a return there? This was the way I wrote that loop originally, but then it either had no return after the loop (which caused a warning) or an uncovered return. I also thought of rewriting this as a recursive function, but I guess it would be a little harder to debug.

vgvassilev: I don’t want to have a return but either or. One of both is fine.

PR #1424, opened 2025-06-28
Rework SetIsReq and overlay and combine them into a single function

### LINKED ISSUE #1423: If a constant element is used, mark the default index as used as well in TBR
TBR considers elements of arrays as separate entities. In order to handle elements that are not defined at compile-time (`arr[i]` as opposed to `arr[3]`), TBR has default indices. The default index is set to `used` if at least one element of the array has been used. Note: if we're unsure if something is `used`, we always assume it is for safety.
Therefore, if we set
1) a known index (`arr[3]`) to `used`, we also set the default index (`arr[i]`) to `used`.
2) a known index (`arr[3]`) to `unused`, we should not make further assumptions.
3) an unknown index (`arr[i]`) to `used`, we set ALL indices to `used`.
4) We can never set the unknown index (`arr[i]`) to `unused` because it corresponds to all indices, and there's no operation that guarantees all elements of an array to be overwritten.

Currently, there's `TBR::SetIsRequired` that sets the exact element we handle to the required `used` state. This fully covers 2) and 4). There is also `TBR::overlay` that is designed to cover case 3) and set all matching indices to `used` as well. However, case 1) is not handled properly, which lets us reproduce the bug like this:
```
double f6(double x, double y) {
  double arr[4];
  arr[0] = y;
  x *= arr[0]; // arr[0] gets used, but arr[default] isn't labeled as used
  int i = 0;
  arr[i] = 0; // tbr is supposed to make an asumption that arr[i] might be arr[0]
  return x;
}
```
I believe it was wrong to have one main function `TBR::SetIsRequired` and additional functions like `TBR::overlay` to cover up gaps in the logic. This leads to code repetition and makes it harder to understand the logic behind this. We should handle all of these cases at a single point.

PetroZarytskyi:
The problem this PR solves is explained in detail in #1423. For short, when we set vars to `used`/`unused` in TBR, we handle constant array indices (like `arr[3]`) in `TBR::SetIsRequired` and partly cover the gaps in the logic in `TBR::overlay`. This PR moves handling of all cases to `TBR::SetIsRequired` and fixes the bug in #1423. Also, this PR removes the state variable `TBR::m_NonConstIndexFound` that was used to detect the cases when `TBR::overlay` should be called.
Apart from that, `ProfileID` sequences are now reversed to the right order for easier handling: the sequence of `arr[2].k` used to be `{k, 2}` and is now `{2, k}`. Iterating backwards through the sequence in recursion resulted in code that was quite hard to read.

Fixes #1423

### DIFF
--- a/lib/Differentiator/TBRAnalyzer.cpp
+++ b/lib/Differentiator/TBRAnalyzer.cpp
@@ -1,5 +1,6 @@
 #include "TBRAnalyzer.h"
 
+#include <algorithm>
 #include <cassert>
 
 #include "clang/AST/Decl.h"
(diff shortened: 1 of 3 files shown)


vgvassilev (approved):
LGTM!

PR #1435, opened 2025-07-04
Print the differentiation plan early when -print-stats is passed.

vgvassilev:
This can give a hint to users debug clad crashes.


vgvassilev:
@MaxSagebaum, perhaps that would help narrowing down the source of issues.

PR #1440, opened 2025-07-08
Support std::pair constructors natively

PetroZarytskyi:
This PR attempts to add support for std::pair constructors without specifying their custom derivatives. There were 2 challenges: 
1) Support base initializers in derived class constructors when building constructor pullbacks. Also, add support for them in TBR.
2) Add support for std::forward.


test/Gradient/Constructors.C
```diff
@@ -285,6 +285,167 @@ double fn4(double i, double j) {
...
+// CHECK-NEXT:  }
+
+double fn6(double x, double y) {
+    argByValWrapper g(x, y);
+    return g.z;
+} // x^2 * y
+
+// CHECK:  static void constructor_pullback(double v, double u, argByValWrapper *_d_this, double *_d_v, double *_d_u) {
+// CHECK-NEXT:      argByValWrapper *_this = (argByValWrapper *)malloc(sizeof(argByValWrapper));
+// CHECK-NEXT:      new (_this) argByValWrapper(v);
```
vgvassilev:
```suggestion
// CHECK-NEXT:      argByValWrapper *_this = argByValWrapper(v);
```

test/Gradient/Constructors.C
```diff
@@ -285,6 +285,167 @@ double fn4(double i, double j) {
...
+// CHECK-NEXT:  }
+
+double fn7(double x, double y) {
+    argByValWrapper g(x, false);
+    return g.z;
+} // x^3
+
+// CHECK:  static void constructor_pullback(double v, bool arg, argByValWrapper *_d_this, double *_d_v, bool *_d_arg) {
+// CHECK-NEXT:      argByValWrapper *_this = (argByValWrapper *)malloc(sizeof(argByValWrapper));
+// CHECK-NEXT:      new (static_cast<argByVal *>(_this)) argByVal(v);
```
vgvassilev, line 381: What do we lose if we call `argByValWrapper *_this = new argByValWrapper(...)`? The memory will be initialized with some values but we will overwrite them. That would not lead to significant performance regression I'd think...

PetroZarytskyi: Well, some constructor calls have side effects and shouldn't be called twice. A real-world example would be move-constructors. If that's better, we can check if the constructor doesn't have any side effects by examining its parameter types or other factors. Also, I think the question is more about the logic in constructor pullbacks in general than about this change in particular.

test/Gradient/Constructors.C
vgvassilev: I think we are blocked by resolving this. Probably it is safe to do it the way you have but it is really a undefined behavior in standards terms...

PetroZarytskyi: Even though this approach has limitations for constructors with side-effects, the PR doesn't solve this problem either. I agree this is better at the moment. I implemented this.

include/clad/Differentiator/Differentiator.h
```diff
@@ -122,7 +122,7 @@ CUDA_HOST_DEVICE T push(tape<T, SBO_SIZE, SLAB_SIZE>& to, ArgsT... val) {
     // copying the sequence of characters to the destination [27.5.1(3)].
     // (C++ has deprecated the volatile qualifiers. However, we drop them here
     // to make sure things still work with codebases which still have them)
-    std::memcpy(const_cast<T*>(&t), tmp, sizeof(T));
+    std::memcpy((void*)const_cast<T*>(&t), tmp, sizeof(T));
```
vgvassilev: Why did we need to make this change?

PetroZarytskyi: I already don't remember the exact reason, but there was a warning because of an implicit cast to `void*`.

PetroZarytskyi: @vgvassilev It seems I was confused. Removing this explicit conversion doesn't break anything.

PR #1441, opened 2025-07-09
Don't create CUDA atomics for basic indices

ovdiiuv:
Before this PR, we unconditionally produced atomic operations for writes to global gpu memory. For example, the gradient of the kernel below:

```cpp

__global__ void indices_perm(int *out, int *in) {
   int index = threadIdx.x + blockIdx.x * blockDim.x;
   out[index] += in[index];
 }
``` 
 
used to have an `atomicAdd` in the reverse pass corresponding to the `out[index] += in[index]` statement. The way `index` is defined makes it injective, meaning no two threads in any two blocks would share the same `index`. In cases like this no data races occur, so we can avoid using atomic operations.

### DIFF
--- a/include/clad/Differentiator/CladUtils.h
+++ b/include/clad/Differentiator/CladUtils.h
@@ -408,6 +408,10 @@ namespace clad {
     /// create modifiable adjoints.
     clang::QualType replaceStdInitListWithCladArray(clang::Sema& S,
                                                     clang::QualType origTy);
+    /// Currently is only used for CUDA in the reverse mode. Determines whether
+    /// an expression, most likely an index, is injective, meaning no two
+    /// threads have the same value.
+    bool isInjective(const clang::Expr* E, clang::ASTContext& ctx);
     } // namespace utils
     } // namespace clad
 
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -52,12 +52,14 @@
 #include <algorithm>
 #include <cstddef>
 #include <iterator>
+#include <memory>
 #include <numeric>
+#include <string>
+#include <utility>
+#include <vector>
 
 #include "clad/Differentiator/CladUtils.h"
 #include "clad/Differentiator/Compatibility.h"
-
-using namespace clang;
 
 namespace clad {
 
@@ -126,27 +128,42 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
   bool ReverseModeVisitor::shouldUseCudaAtomicOps(const Expr* E) {
     if (!m_Context.getLangOpts().CUDA)
       return false;
-
-    if (!isa<DeclRefExpr>(E))
-      return false;
-
-    const auto* DRE = cast<DeclRefExpr>(E);
-
-    if (const auto* PVD = dyn_cast<ParmVarDecl>(DRE->getDecl())) {
-      if (m_DiffReq->hasAttr<clang::CUDAGlobalAttr>())
-        // Check whether this param is in the global memory of the GPU
-        return m_DiffReq.HasIndependentParameter(PVD);
-      if (m_DiffReq->hasAttr<clang::CUDADeviceAttr>()) {
-        for (auto index : m_DiffReq.CUDAGlobalArgsIndexes) {
-          const auto* PVDOrig = m_DiffReq->getParamDecl(index);
-          if ("_d_" + PVDOrig->getNameAsString() == PVD->getNameAsString() &&
-              (utils::isArrayOrPointerType(PVDOrig->getType()) ||
-               PVDOrig->getType()->isReferenceType()))
-            return true;
+    if (const auto* DRE = dyn_cast<DeclRefExpr>(E)) {
+      if (const auto* PVD = dyn_cast<ParmVarDecl>(DRE->getDecl())) {
+        if (m_DiffReq->hasAttr<clang::CUDAGlobalAttr>())
+          // Check whether this param is in the global memory of the GPU
+          return m_DiffReq.HasIndependentParameter(PVD);
+        if (m_DiffReq->hasAttr<clang::CUDADeviceAttr>()) {
+          for (auto index : m_DiffReq.CUDAGlobalArgsIndexes) {
+            const auto* PVDOrig = m_DiffReq->getParamDecl(index);
+            if ("_d_" + PVDOrig->getNameAsString() == PVD->getNameAsString() &&
+                (utils::isArrayOrPointerType(PVDOrig->getType()) ||
+                 PVDOrig->getType()->isReferenceType()))
+              return true;
+          }
+        }
+      }
+    } else if (const auto* ASE = dyn_cast<ArraySubscriptExpr>(E)) {
+      const auto* base =
+          dyn_cast<DeclRefExpr>(ASE->getBase()->IgnoreImpCasts());
+      if (const auto* PVD = dyn_cast<ParmVarDecl>(base->getDecl())) {
+        const auto* idx = ASE->getIdx();
+        if (m_DiffReq->hasAttr<clang::CUDAGlobalAttr>())
+          // Check whether this param is in the global memory of the GPU and
+          // if index is injective.
+          return m_DiffReq.HasIndependentParameter(PVD) &&
+                 !clad::utils::isInjective(idx, m_Context);
+        if (m_DiffReq->hasAttr<clang::CUDADeviceAttr>()) {
+          for (auto index : m_DiffReq.CUDAGlobalArgsIndexes) {
+            const auto* PVDOrig = m_DiffReq->getParamDecl(index);
+            if (PVDOrig->getNameAsString() == PVD->getNameAsString() &&
+                (utils::isArrayOrPointerType(PVDOrig->getType()) ||
+                 PVDOrig->getType()->isReferenceType()))
+              return !clad::utils::isInjective(idx, m_Context);
+          }
         }
       }
     }
-
     return false;
   }
 
@@ -1376,7 +1393,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     // Create the (target += dfdx) statement.
     if (dfdx()) {
       Expr* add_assign = nullptr;
-      if (shouldUseCudaAtomicOps(target))
+      if (shouldUseCudaAtomicOps(ASE))
         add_assign = BuildCallToCudaAtomicAdd(result, dfdx());
       else
         add_assign = BuildOp(BO_AddAssign, result, dfdx());
--- a/test/CUDA/GradientKernels.cu
+++ b/test/CUDA/GradientKernels.cu
@@ -77,7 +77,7 @@ __global__ void add_kernel_3(int *out, int *in) {
 //CHECK-NEXT:    {
 //CHECK-NEXT:        out[index0] = _t0;
 //CHECK-NEXT:        int _r_d0 = _d_out[index0];
-//CHECK-NEXT:        atomicAdd(&_d_in[index0], _r_d0);
+//CHECK-NEXT:        _d_in[index0] += _r_d0;
 //CHECK-NEXT:    }
 //CHECK-NEXT:}
 
@@ -347,7 +347,7 @@ __global__ void dup_kernel_with_device_call_2(double *out, const double *in, dou
 //CHECK-NEXT:    int _d_index = 0;
 //CHECK-NEXT:    int index0 = threadIdx.x + blockIdx.x * blockDim.x;
 //CHECK-NEXT:    {
-//CHECK-NEXT:        atomicAdd(&_d_in[index0], _d_y);
+//CHECK-NEXT:        _d_in[index0] += _d_y;
 //CHECK-NEXT:        *_d_val += _d_y;
 //CHECK-NEXT:    }
 //CHECK-NEXT:}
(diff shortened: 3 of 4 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -56,7 +56,10 @@
 #include "clad/Differentiator/CladUtils.h"
 #include "clad/Differentiator/Compatibility.h"
 
-using namespace clang;
+#include "clang/ASTMatchers/ASTMatchFinder.h"
```
vgvassilev: What is the advantage of having an ast matcher support? Clad has been deliberately not relying on this so that we have clearer code and better control.

ovdiiuv: The other way to do it is to have loop that unfolds binary operations. That would get pretty ugly, since we would need a lot of checks for the LHS/RHS and really hard make changes. Matchers are flexible and it’s easy to add new ones.

vgvassilev: You can have a nested visitor.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -133,30 +136,238 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+              return false;
+          }
+        } else if (current->Node.isExpr()) {
+          if (!current->Node.m_E)
+            return false;
+          std::string actualName =
+              current->Node.m_E->getNameInfo().getAsString();
+          if (actualName != expected) {
+            if (i % 2 == 0 && actualName == patternIdx[i + 1])
+              *shouldInvert = true;
```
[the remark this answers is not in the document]
ovdiiuv: not relevant, I believe

test/CUDA/GradientKernels.cu
```diff
@@ -689,6 +689,116 @@ void launch_add_kernel_4(int *out, int *in, const int N) {
 //CHECK-NEXT:    cudaFree(_d_out_dev);
 //CHECK-NEXT:}
 
+__global__ void indices_perm(int *out, int *in) {
+  int index1 = threadIdx.x + blockIdx.x * blockDim.x;
```
kchristin22, line 693: I think for now we should cover only the cases of const decl refs, and change that when we add functionality for the rest.
```suggestion
  const int index1 = threadIdx.x + blockIdx.x * blockDim.x;
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -133,30 +135,220 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+          }
+        }
+      }
+    } else if (const auto* ASE = dyn_cast<ArraySubscriptExpr>(E)) {
+      if (m_DiffReq->hasAttr<clang::CUDAGlobalAttr>()) {
+        const auto* idx = ASE->getIdx();
+        const auto* base =
+            dyn_cast<DeclRefExpr>(ASE->getBase()->IgnoreImpCasts());
+        if (const auto* PVD = dyn_cast<ParmVarDecl>(base->getDecl())) {
+          if (m_DiffReq->hasAttr<clang::CUDAGlobalAttr>()) {
```
kchristin22: If we only check the const indexes of arrays, we can change that to check the CUDAGlobalArgsIndexes list so we can cover the device pullbacks as well.
```suggestion
          if (!m_DiffReq.CUDAGlobalArgsIndexes.empty()) {
```

kchristin22 (commented):
I think this approach is great, we can cover the const cases now and extend it to re-assignments later. This could be done by adding the decl refs of indexes that are injective to a list as we traverse them in the Inject checker and then update these decl refs when we ecnounter an assignment in the original visitor (reverse mode visitor). We don't have to store them in this PR, it could be left for the re-assignment one.

lib/Differentiator/ReverseModeVisitor.cpp
ovdiiuv: We have discussed this I think, apparently adding support for the device functions isn't that easy, so this should be a subject of some other PR.

test/CUDA/GradientKernels.cu
ovdiiuv: We don't check for the reassignmnets yet, but I think we shouldn't change the tests. I'm working on that right now so we could open an issue in the meanwhile.

lib/Differentiator/ReverseModeVisitor.cpp
kchristin22: Yes, I'm talking about device pullbacks instead of device gradients. Like in this case:
```
__device__ double fn(const double *in, double val) {
    int index = threadIdx.x + blockIdx.x * blockDim.x;
    return in[index] + val;
}

__global__ void kernel(double *out, const double *in, double val)
{
    int index = threadIdx.x + blockIdx.x * blockDim.x;
    out[index] = fn(in, val);
}

int main(){
    auto grad = clad::gradient(kernel, "out, in");
    return 0;
}

```

```
__attribute__((device)) void fn_pullback_0(const double *in, double val, double _d_y, double *_d_in, double *_d_val) {
    int _d_index = 0;
    int index0 = threadIdx.x + blockIdx.x * blockDim.x;
    {
        atomicAdd(&_d_in[index0], _d_y); // we can tell this is global memory address from the call in the kernel we derive
        *_d_val += _d_y;
    }
}
void kernel_grad_0_1(double *out, const double *in, double val, double *_d_out, double *_d_in) {
    double _d_val = 0.;
    int _d_index = 0;
    int index0 = threadIdx.x + blockIdx.x * blockDim.x;
    out[index0] = fn(in, val);
    {
        double _r_d0 = _d_out[index0];
        _d_out[index0] = 0.;
        double _r0 = 0.;
        fn_pullback_0(in, val, _r_d0, _d_in, &_r0);
        _d_val += _r0;
    }
}
```

`CUDAGlobalArgsIndexes` is not empty when:
* we use `clad::gradient` on a kernel and we are deriving this kernel function
* we use `clad::gradient` on a kernel and we are passing global params to a device call inside that kernel

test/CUDA/GradientKernels.cu
kchristin22: I meant changing the implementation as well, restricting it to const decl refs for now. Only suggested that so we can change the check with the CUDAGloblArgsIndexes (the other comment), because we saw that when the index was the parameter of a for loop it didn't account the reassignment (the step) and didn't use atomic ops. But if we only checked the const, this test wouldn't be affected, we would still use atomic ops. 

It's either-or, we can leave it as is for now and remember to change this line in the other PR for the re-assignments, so we can extend this analysis to device pullbacks. We can do that if you prefer it : )

lib/Differentiator/ReverseModeVisitor.cpp
ovdiiuv: For pullbacks as well. We need to make sure that the memory is global, which requires more then checking whether there are any elements in `m_DiffReq.CUDAGlobalArgsIndexes`.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -133,30 +135,226 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+          return true;
+
+        return comparePatternToTree(current->left.get(), patternIdx,
+                                    2 * i + 1) &&
+               comparePatternToTree(current->right.get(), patternIdx, 2 * i);
+      }
+
+      bool isInjectiveIdx(const clang::Expr* E) {
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
+        TraverseStmt(const_cast<clang::Expr*>(E));
```
vgvassilev: Can we avoid const_casts its UB.

test/CUDA/GradientKernels.cu
ovdiiuv: Why changing an implementation to check for indices to be const just to change it back later? I say we open an issue about that we don't check for liveness and I resolve it next.

kchristin22: I was only suggesting checking for const so we don't remove atomicAdd in for loops in the already included tests. But I see that this is not altered in the tests (e.g. `add_kernel_4`). How come these were not evaluated as injective? You only check if the index is declared as a binary op for now? If yes, I think this is a better work-around than checking for const for now and removing it later : ) Just mention in the issue that in the next PR you go deeper into a declaration:
```
int index = threadIdx.x + blockIdx.x * blockDim.x;
int i = index;
_d_in[i] += ...;
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -133,30 +135,226 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+                (utils::isArrayOrPointerType(PVDOrig->getType()) ||
+                 PVDOrig->getType()->isReferenceType()))
+              return true;
+          }
+        }
+      }
+    } else if (const auto* ASE = dyn_cast<ArraySubscriptExpr>(E)) {
+      const auto* base =
+          dyn_cast<DeclRefExpr>(ASE->getBase()->IgnoreImpCasts());
+      if (const auto* PVD = dyn_cast<ParmVarDecl>(base->getDecl())) {
```
kchristin22, line 149: Checking if it's a parameter is common in both branches (DRE and ASE), so we should have that as the outer if-condition

test/CUDA/GradientKernels.cu
ovdiiuv: Yes. Agreed.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -134,30 +136,226 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+      };
+      std::unique_ptr<IdxNode> m_Root;
+      IdxNode* m_ParentNode = nullptr;
+
+      enum class side { left, right } m_Side;
+
+    public:
+      InjectiveCheckerExpr() = default;
+
+      bool comparePatternToTree(const IdxNode* current,
```
vgvassilev: We need some more documentation about what this routine does.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -134,30 +136,226 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+                                const std::vector<std::string>& patternIdx,
+                                size_t i = 1) {
+
+        if (!current && (i >= patternIdx.size() || patternIdx[i].empty()))
```
vgvassilev: Under what conditions current can be a nullptr?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -134,30 +136,226 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+          TraverseStmt(L);
+          m_ParentNode = currPtr;
+
+          m_Side = side::right;
+          TraverseStmt(R);
+        }
+        return true;
+      }
+
+      bool TraverseDeclRefExpr(clang::DeclRefExpr* DRE) {
```
vgvassilev: Why not `VisitDeclRefExpr` but `TraverseDeclRefExpr`?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -134,30 +136,226 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+          if (isConstL && isConstR)
+            return false;
+
+          if (!isConstL && isConstR)
+            return TraverseStmt(L);
+
+          if (isConstL && !isConstR)
+            return TraverseStmt(R);
+
+          return isInjectiveE(BinOp);
```
```suggestion
          if (!isConstL && isConstR)
            return TraverseStmt(L);
          
          if (!isConstL && !isConstR)
            return sInjectiveE(BinOp);

          if (!isConstR)
            return TraverseStmt(R);
          return TraverseStmt(L);
```

kchristin22 (approved):

lib/Differentiator/ReverseModeVisitor.cpp
ovdiiuv: I think we have converged that traverse methods are better, since they do not visit subexpressions automatically, meaning if x=y is visited and not traversed, we would visit LHS, RHS and then x, y as subexpressions of x=y.

vgvassilev: Can we capture that as a comment in the appropriate place..

lib/Differentiator/ReverseModeVisitor.cpp
ovdiiuv: I don't think it can, that's something that's left from the previous implementation.

ovdiiuv: Each node has a left and right children whatsoever, those might be not initialized.

vgvassilev: In that case, I'd move the 
```        if (!current || i >= patternIdx.size() || patternIdx[i].empty())
          return false;
```
before that statement and remove the `!current` checks from the rest.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -134,30 +136,232 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+            return isInjectiveE(init->IgnoreImpCasts());
+        }
+        return false;
+      }
+
+      bool TraverseIntegerLiteral(clang::IntegerLiteral* IL) { return false; }
+
+    } checker(ctx);
+
+    return checker.isInjectiveIdx(E);
```
vgvassilev: Looks like this logic is useful beyond the reverse mode visitor. Can we move it somewhere in utils for example?

include/clad/Differentiator/CladUtils.h
```diff
@@ -408,6 +408,12 @@ namespace clad {
     /// create modifiable adjoints.
     clang::QualType replaceStdInitListWithCladArray(clang::Sema& S,
                                                     clang::QualType origTy);
+    /// Helper fucntion for isInjective.
+    bool isInjectiveE(const clang::Expr* E);
```
vgvassilev: We don't need that as part of our api. It can be cpp file static.

PR #1443, opened 2025-07-12
Add support for std::shared_ptr and std::weak_ptr and improve differentiation of object params

### LINKED ISSUE #781: Add support for smart pointers
We should be able to differentiate `std::unique_ptr`, `std::shared_ptr` and `std::weak_ptr`.

PetroZarytskyi:
This PR adds support for ``std::shared_ptr`` and ``std::weak_ptr`` by adding custom derivatives and by fixing an issue we had when handling value-type object parameters. 
In particular, the code for the pullback of 
```
... = f(x, ...); // Assume `x` is passed by value, i.e. by copying
```
was built with the following pattern:
```
Class_type _r0 = {};
f_pullback(x, ..., &_r0, ...);
Class_type::constructor_pullback(x, &_r0, &_d_x); // corresponds to the copy constructor
```
While this logic works for purely numerical types like ``std::pair`` or ``std::complex`` (with numerical elements), it breaks for types that use pointer fields like ``std::smart_ptr`` or even ``std::vector`` as ``f_pullback`` expects all pointer fields of ``_r0`` to be properly initialized.
This PR solves this issue by initializing objects with the result of the corresponding copy-constructor in the forward pass:
```
clad::ValueAndAdjoint<...> _t0 = ...::constructor_reverse_forw(..., x, _d_x);
...
Class_type _r0 = _t0.adjoint;
f_pullback(x, ..., &_r0, ...);
Class_type::constructor_pullback(x, &_r0, &_d_x);
```
Please note that the call to ``constructor_reverse_forw`` was also generated before this PR.
It's also worth noting that this strategy reflects what we do to handle copy-initialization in object-type declarations. 

Fixes #781.

### DIFF
--- a/include/clad/Differentiator/STLBuiltins.h
+++ b/include/clad/Differentiator/STLBuiltins.h
@@ -790,6 +790,35 @@ void constructor_pullback(const ::std::shared_ptr<T>& p,
                           ::std::shared_ptr<T>* dthis,
                           ::std::shared_ptr<T>* dp) noexcept {}
 
+template <typename T, typename U>
+void constructor_pullback(const ::std::shared_ptr<T>& p, U*,
+                          ::std::weak_ptr<T>* dthis, ::std::shared_ptr<T>* dp,
+                          U*) noexcept {}
+
+// std::weak_ptr<T> custom derivatives...
+template <typename T, typename U>
+clad::ValueAndAdjoint<::std::weak_ptr<T>, ::std::weak_ptr<T>>
+constructor_reverse_forw(clad::ConstructorReverseForwTag<::std::weak_ptr<T>>,
+                         U p, U d_p) {
+  return {::std::weak_ptr<T>(p), ::std::weak_ptr<T>(d_p)};
+}
+template <typename T, typename U>
+void constructor_pullback(U&& p, ::std::weak_ptr<T>* dthis, U* dp) noexcept {}
+
+template <typename T, typename U>
+void constructor_pullback(const U& p, ::std::weak_ptr<T>* dthis,
+                          U* dp) noexcept {}
+
+template <typename T>
+clad::ValueAndAdjoint<::std::shared_ptr<T>, ::std::shared_ptr<T>>
+lock_reverse_forw(const ::std::weak_ptr<T>* p,
+                  const ::std::weak_ptr<T>* dp) noexcept {
+  return {p->lock(), dp->lock()};
+}
+
+template <typename T>
+void lock_pullback(const ::std::weak_ptr<T>* p, ::std::shared_ptr<T> dthis,
+                   ::std::weak_ptr<T>* dp) noexcept {}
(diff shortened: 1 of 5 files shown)


vgvassilev:
Can you mention what issue this pr intends to close?

include/clad/Differentiator/STLBuiltins.h
```diff
@@ -809,6 +840,17 @@
                                           second_half_tuple(t));
 }
 
+// std::make_shared<T> custom derivatives...
+clad::ValueAndAdjoint<::std::shared_ptr<double>, ::std::shared_ptr<double>>
+make_shared_reverse_forw(double& x, double& dx) {
+  return {::std::make_shared<double>(x), ::std::make_shared<double>(dx)};
+}
+
+void make_shared_pullback(double& x, ::std::shared_ptr<double> dthis,
```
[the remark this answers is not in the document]
vgvassilev: Non-templated definitions should be marked inline.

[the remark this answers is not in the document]
PetroZarytskyi: Oh, that was a mistake. Everything should be templated. I fixed it

vgvassilev (approved):
LGTM!

include/clad/Differentiator/STLBuiltins.h
```diff
@@ -681,26 +681,20 @@ constructor_reverse_forw(clad::ConstructorReverseForwTag<::std::unique_ptr<T>>,
 template <typename T>
 void constructor_pullback(T* p, ::std::unique_ptr<T>* dthis, T* dp) noexcept {};
 
+// operator* custom derivatives
 template <typename T>
-clad::ValueAndAdjoint<T&, T&>
-operator_star_reverse_forw(const ::std::unique_ptr<T>* u,
-                           const ::std::unique_ptr<T>* d_u) {
+clad::ValueAndAdjoint<decltype(*(T{}))&, decltype(*(T{}))&>
```
parth-07, line 716: While this should hold good for STL types, I feel that we should not have such generic custom derivatives, as there is always a risk of accidentally using a custom derivative where it should not be used.

PetroZarytskyi: We discussed this with Vassil and thought it was fine to assume such behavior of ``operator*``. But maybe you're right. I changed it to only work with smart pointers and std iterators. Do you think it's good now?

parth-07: Yes, thank you.

PetroZarytskyi:
@parth-07 Some platforms had a common base class for std::weak_ptr and std::shared_ptr, and simple functions like operator* were written for that. So I did some metaprogramming to allow the types that std::shared_ptr can be implicitly converted to.

vgvassilev:
The windows bot seems stuck but not due to this pr.

PR #1444, opened 2025-07-13
Add primitive docs for custom derivatives in Clad sphinx docs

arima-04:


docs/userDocs/source/index.rst
```diff
@@ -11,26 +11,26 @@ Release v\ |version|.
...
+support of language features. It supports reverse-mode AD (a.k.a backpropagation)
+as well as forward-mode AD. It also facilitates computation of hessian matrix and
 jacobian matrix of any arbitrary function.
 
-Automatic differentiation solves all the usual problems of numerical 
-differentiation (precision loss) and symbolic differentiation 
-(inefficient code produced). If you are just getting started with clad, then please 
+Automatic differentiation solves all the usual problems of numerical
+differentiation (precision loss) and symbolic differentiation
+(inefficient code produced). If you are just getting started with clad, then please
```
vgvassilev:
```suggestion
Automatic differentiation solves all the usual problems of numerical
differentiation such as precision loss and symbolic differentiation such as
limitations to closed form expressions and expression swell. If you are just getting started with clad, then please
```

docs/userDocs/source/user/CustomDerivatives.rst
```diff
@@ -0,0 +1,385 @@
...
+.. contents::
+
+What are custom derivatives?
+============================
+
+Clad allows users to specify custom derivatives for functions. It is useful when a more
+efficient or numerically stable expression for derivatives is known, or when Clad is unable
+to differentiate a function. Clad is unable to differentiate a function if its definition is
+in a library and source code is not available, or when the function code contains a C++ feature
+that Clad does not support yet.
```
vgvassilev, line 24:
```suggestion
that Clad does not support yet.

The custom derivatives mechanism enable hybrid AD approaches where Clad can work synergetically with tools based on operator overloading, for example. Another use of custom derivatives is to connect to third party libraries which need specific rules for differentiation such as linear solvers. The mechanism is useful to connect differentiable code to code which does not have differentiable properties, such as neural networks or other ML models.
```

docs/userDocs/source/user/CustomDerivatives.rst
```diff
@@ -0,0 +1,385 @@
...
+
+Custom derivatives are defined as C++ functions in clad-specific namespaces. Whenever Clad
+needs to differentiate a function, it will first look if a custom derivative for the function
+is available. If so, Clad will use the custom derivative instead of differntiating the function
+using AD.
+
+Clad supports 3 types of custom derivatives: :code:`pushforward`, :code:`pullback` and :code:`reverse_forw`.
+Each type has a distinct use-case. Clad internally differentiates functions using the
+same types as we have for custom derivatives. For example, if a Clad needs to differentiate
+a function :code:`foo` that has a custom derivative :code:`foo_pushforward` defined, then it will use
```
vgvassilev:
```suggestion
a function :code:`foo` that has a custom derivative :code:`foo_pushforward` defined, then it will use
```

Can we use different wording from `foo` and `bar`? `fn1`, `fn2` or `f`, `g` or anything similar seems better.

docs/userDocs/source/user/CustomDerivatives.rst
```diff
@@ -0,0 +1,385 @@
...
+to differentiate a function. Clad is unable to differentiate a function if its definition is
+in a library and source code is not available, or when the function code contains a C++ feature
+that Clad does not support yet.
+
+Custom derivatives are defined as C++ functions in clad-specific namespaces. Whenever Clad
+needs to differentiate a function, it will first look if a custom derivative for the function
+is available. If so, Clad will use the custom derivative instead of differntiating the function
+using AD.
+
+Clad supports 3 types of custom derivatives: :code:`pushforward`, :code:`pullback` and :code:`reverse_forw`.
```
vgvassilev: Can we connect the `pushforward` and `pullback` terminology to the differential geometry by referencing some appropriate resource?

docs/userDocs/source/user/CustomDerivatives.rst
```diff
@@ -0,0 +1,385 @@
...
+The custom derivatives for free functions needs to be defined under
+:code:`clad::custom_derivatives` and for class functions (both static and non-static) under
+:code:`clad::custom_derivatives::class_functions`.
+
+If a free function is defined in a namespace :code:`A::B::C`, then the custom derivative for
+the function must be defined in the same namespace sequence under :code:`clad::custom_derivatives`,
+i.e, :code:`clad::custom_derivatives::A::B::C`. The custom derivatives under
+:code:`clad::custom_derivatives::class_functions` do not follow this rule. The custom derivatives
+for the class functions must all be defined directly in :code:`clad::custom_derivatives::class_functions`
+regardless of the class's namespace.
```
vgvassilev:
```suggestion
regardless of the class's namespace.

Note that non-templated free functions defined in a header file need to be marked `inline` to avoid issues with symbol duplication just like any other C++ entity defined in a header file.
```

docs/userDocs/source/user/CustomDerivatives.rst
```diff
@@ -0,0 +1,385 @@
...
+  } // namespace custom_derivatives
+  } // namespace clad
+
+:code:`clad::ConstructorPushforwardTag<::Coordinates>` is used to identify the
+class for which the constructor pushforward is defined. The member function
+custom derivatives do not require this tag because the custom derivative function
+takes the class object as the first argument, which is sufficient to identify
+the class.
+
+constructor pullback custom derivative
```
vgvassilev:
```suggestion
Constructor pullback custom derivative
```

docs/userDocs/source/user/CustomDerivatives.rst
```diff
@@ -0,0 +1,385 @@
...
+    d_coordinates->z = 0;
+  }
+  } // namespace class_functions
+  } // namespace custom_derivatives
+  } // namespace clad
+
+Note that the constructor pullback does not need anything such as
+:code:`clad::ConstructorPushforwardTag<::Coordinates>`. It is because
+the constructor pullback takes :code:`d_coordinates` as an argument, which can be
+used to identify the class for which the constructor pullback is defined.
```
vgvassilev, line 437: Can we add a section on `clad::zero`?

vgvassilev (commented):
This looks quite promising.
@PetroZarytskyi, can you take a look and possibly review this PR, too.
@davidlange6, can you check English?

docs/userDocs/source/user/CustomDerivatives.rst
PetroZarytskyi: I think you meant ``clad::zero_init``?

vgvassilev: Yes.

docs/userDocs/source/user/CustomDerivatives.rst
```diff
@@ -0,0 +1,385 @@
...
+that Clad does not support yet.
+
+Custom derivatives are defined as C++ functions in clad-specific namespaces. Whenever Clad
+needs to differentiate a function, it will first look if a custom derivative for the function
+is available. If so, Clad will use the custom derivative instead of differntiating the function
+using AD.
+
+Clad supports 3 types of custom derivatives: :code:`pushforward`, :code:`pullback` and :code:`reverse_forw`.
+Each type has a distinct use-case. Clad internally differentiates functions using the
+same types as we have for custom derivatives. For example, if a Clad needs to differentiate
```
PetroZarytskyi:
```suggestion
same types as we have for custom derivatives. For example, if Clad needs to differentiate
```

docs/userDocs/source/user/CustomDerivatives.rst
```diff
@@ -0,0 +1,385 @@
...
+:code:`foo_pushforward` to compute the derivative of :code:`foo`. Otherwise, Clad will try to automatically
+generate :code:`foo_pushforward`. The outside observable behavior of :code:`foo_pushforward` should be the
+same in both the cases. Clad does not make any efforts to ensure that the custom derivatives have the correct
+behavior. It is the user's responsibility to ensure that the custom derivatives are correct.
+
+You need to define :code:`pushforward` custom derivative if you want the derivative to be used
+by the Clad forward mode AD (:code:`clad::differentiate`). The :code:`pullback` custom derivatives is
+used by the Clad reverse mode AD (:code:`clad::gradient`). :code:`reverse_forw` is a weird custom derivative
+type because it is not meant to differentiate anything. It is used by the Clad reverse mode AD
+to determine the ajoint of a function's return value for functions which returns a reference or
```
PetroZarytskyi:
```suggestion
to determine the adjoint of a function's return value for functions which returns a reference or
```

docs/userDocs/source/user/CustomDerivatives.rst
```diff
@@ -0,0 +1,385 @@
...
+
+Pushforward custom derivatives
+===============================
+
+The :ref:`pushforward <PushforwardFunctions>` custom derivative is used by the Clad forward mode AD.
+Pushforward functions *pushes* sensitivities of the inputs to the sensitivities of the outputs.
+Put simply, it computes partial derivative of function's output with respect to some independent
+variable. This independent variable does not necessarily have to be the function's input.
+This functionality can be easily understood with the help of an example, so let's set aside
+the mathematics jargon.
```
PetroZarytskyi, line 90: In the mathematics jargon, pushforwards are directional derivatives (+ the primal function). So basically, the derivative is computed at the point ``{u, v}`` in the direction ``{du, dv}`` (again, plus the primal). I think it would clarify everything for those who know what it is. We could mention this very briefly either here or at the end of the section

docs/userDocs/source/user/CustomDerivatives.rst
```diff
@@ -0,0 +1,385 @@
...
+  } // namespace clad
+
+Here :code:`du` and :code:`dv` are the adjoints of the function arguments.
+
+Some important things to note here:
+- The reverse-forward custom derivative function name must be :code:`<function_name>_reverse_forw`.
+
+- The reverse-forward custom derivative function must take the same number of arguments as the
+  original function, followed by the adjoints of the function's arguments. The adjoint of a
+  function argument has the same type as the function argument after removing the :code:`const`
```
PetroZarytskyi, line 263: So I just checked on master, and the const is not removed. I think adjoint types are identical at the moment

docs/userDocs/source/user/CustomDerivatives.rst
```diff
@@ -0,0 +1,385 @@
...
+    void foo_pullback(A *a, double u, double v, double dr, A *da, double *du, double *dv) {
+      *du += dr * a->val1;
+      da->val1 += dr * u;
+      *dv += dr * a->val2;
+      da->val2 += dr * v;
+    }
+  } // namespace class_functions
+  } // namespace custom_derivatives
+  } // namespace clad
+
```
PetroZarytskyi, line 316: Can you briefly mention that for const methods, it will be ``const A* a, const A* _d_a`` for pushforwards and ``const A* a, A* _d_a`` for pullbacks?

docs/userDocs/source/user/CustomDerivatives.rst
```diff
@@ -0,0 +1,385 @@
...
+custom derivatives do not require this tag because the custom derivative function
+takes the class object as the first argument, which is sufficient to identify
+the class.
+
+constructor pullback custom derivative
+----------------------------------------
+
+Constructor pullback custom derivatives are more similar to the ordinary pullback
+functions. Constructor pullback functions do not have the same problem as of constructor
+pushforward functions of initializing the primal object and the derivative object. After all,
```
PetroZarytskyi, line 395: Sadly, constructor_pullback doesn't have the same problem because the initialization must be done in the forward pass, and so this responsibility lies on constructor_reverse_forw. By the way, we need a section about that. We can often avoid using that by 
```
A a(args);
```
-->
```
A a(args);
A da(a);
clad::zero_init(a);
```
But really, we can only avoid constructor_reverse_forw either for iterable containers, simple numerical structures (no pointer fields) like std::pair, or when the object is initialized with the default constructor. For complex types like std::vector, we rely on constructor_reverse_forw.

arima-04:
Thank you for the detailed reviews @vgvassilev and @PetroZarytskyi. I will implement the reviews soon.

docs/userDocs/source/user/CustomDerivatives.rst
arima-04: Yes, but I think the better place for going into more detail about pushforward and pullback would be `docs/userDocs/source/user/CoreConcepts.rst`. Please let me know your thoughts on this.

docs/userDocs/source/user/CustomDerivatives.rst
arima-04: Oh, thank you for mentioning this. I will update the docs.

docs/userDocs/source/user/CustomDerivatives.rst
vgvassilev: We can refactor this on a later stage. I'd say we reference something and keep track this comment and move on to merge the PR.

docs/userDocs/source/user/CustomDerivatives.rst
arima-04: Yes, sure. I have changed the function naming from `foo`/`bar` to `fn`/`g`.

docs/userDocs/source/user/CustomDerivatives.rst
parth-07: Done. I have mentioned this along with the example, so that the users can refere to the example for the names `u` and `v`.

docs/userDocs/source/user/CustomDerivatives.rst
parth-07: It seems that this issue got fixed recently, as now on master, the const is removed.

docs/userDocs/source/user/CustomDerivatives.rst
arima-04: I have mentioned why the constructor pullback custom derivatives does not have the initialization problem.

Currently, the docs does not cover object construction through custom derivatives. I would like to cover this topic as a follow-up PR.

docs/userDocs/source/user/CustomDerivatives.rst
arima-04: I have made the pullback and pushforward terms here hyperlink to the corresponding Wikipedia-articles. Please let me know if you have a better resource in mind to add as a reference here.

docs/userDocs/source/user/CustomDerivatives.rst
arima-04: Done. I have mentioned this along with the example, so that the users can refer to the example for the names `u ` and `v`.

docs/userDocs/source/user/CustomDerivatives.rst
arima-04: Done.

docs/userDocs/source/user/CustomDerivatives.rst
arima-04: It seems that this issue is fixed on master, as now `const` seems to be removed for adjoint types.

docs/userDocs/source/user/CustomDerivatives.rst
arima-04: I have mentioned why the pullback constructors does not have the same problem of initialization as of pushforward constructors. 


Currently, the docs does not cover object construction through custom derivatives. I would like to cover this topic as a follow-up PR.

arima-04:
> @arima-04, ping.

I will get back to this PR tomorrow. I couldn't get a chance to run the documentation through google docs.

arima-04:
@davidlange6 @vgvassilev 

I'm sorry for the delay. I have updated the PR with the english / grammar fixes from Google docs and grammarly.

PR #1445, opened 2025-07-13
Add support for functions that modify their parameters

PetroZarytskyi:
Currently, if a pointer argument is passed to a function and the pointee value is changed, Clad doesn't support restoring it in the reverse sweep. This PR introduces ``clad::restore_tracker`` that can return the program to the state prior to the function call, e.g.
```
double* arr = ...;
f(arr); // takes arr as a non-const pointer `double*` and modifies it
```
->
```
clad::restore_tracker _tracker0 = {};
f_reverse_forw(arr, _d_arr, _tracker0); // _tracker0 stores all the changes to `arr`
...
_tracker0.restore(); // arr gets restored
f_pullback(arr, _d_arr);
```
``f_reverse_forw`` would then be generated like this:
```
void f(double* arr) {
  for (int i=0; ...) {
    arr[i] = ...;
  }
}
```
->
```
void f_reverse_forw(double* arr, double* _d_arr, clad::restore_tracker& _tracker0) {
  for (int i=0; ...) {
    _tracker0.store(arr[i]); // stores the memory address of arr[i] as well as its value
    arr[i] = ...;
  }
}
```
Note: if ``clad::restore_tracker::store`` is applied to a given location multiple times, it should ignore all calls except the first one because its purpose is only to return everything to the state before the function call. Also, when dealing with nested `reverse_forw` calls, we should just pass the same restore_tracker.
This is turned off for methods for now because it would interfere with our STL custom derivatives, as essentially all `reverse_forw` functions would now require the ``clad::restore_tracker`` parameter.


lib/Differentiator/CladUtils.cpp
```diff
@@ -879,6 +905,43 @@ namespace clad {
...
+      if (T->isReferenceType())
+        return !T.getNonReferenceType().isConstQualified();
+      if (T->isPointerType())
+        return !T->getPointeeType().isConstQualified();
+      if (const auto* RT = T->getAs<RecordType>()) {
+        const RecordDecl* RD = RT->getDecl();
+        if (const auto* CRD = dyn_cast<CXXRecordDecl>(RD))
+          if (!isCopyable(CRD))
+            return true;
+        for (const auto* field : RD->fields())
```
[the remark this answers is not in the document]
PetroZarytskyi: Should we do this? To me, the existing
```
for (const auto* field : RD->fields())
  if (isMemoryType(field->getType()))
    return true;
```
looks way cleaner than
```
if (std::any_of(RD->field_begin(), RD->field_end(),
                [](const FieldDecl* field) {
                  return isMemoryType(field->getType());
                }))
  return true;
```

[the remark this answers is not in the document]
vgvassilev: I agree. Let’s disable that check.

lib/Differentiator/ReverseModeForwPassVisitor.cpp
```diff
@@ -145,10 +150,36 @@ ReverseModeForwPassVisitor::BuildParams(DiffParams& diffParams) {
...
+    return {};
+  if (const auto* DRE = dyn_cast<DeclRefExpr>(E->IgnoreCasts())) {
+    const auto* VD = cast<VarDecl>(DRE->getDecl());
+    if (!VD->getType()->isReferenceType())
+      return {};
+  }
+  m_IsRequired = true;
+  Expr* storeCall =
+      BuildCallExprToMemFn(m_SmartTape, /*MemberFunctionName=*/"store", {E});
+  return {storeCall, nullptr};
```
vgvassilev:
```suggestion
  return {storeCall, /*?=*/nullptr};
```

lib/Differentiator/CladUtils.cpp
```diff
@@ -927,6 +955,33 @@ namespace clad {
       return false;
     }
 
+    bool hasMemoryTypeParams(const FunctionDecl* FD) {
+      // FIXME: enable for methods
+      if (isa<CXXMethodDecl>(FD) || FD->isOverloadedOperator())
+        return false;
+      // if (const auto* MD = dyn_cast<CXXMethodDecl>(FD))
```
vgvassilev: Do we want to keep this commented out code? If so, we should also add some fixme comment.

lib/Differentiator/CladUtils.cpp
```diff
@@ -927,6 +955,33 @@ namespace clad {
...
+      //   if (MD->isInstance() && !MD->isConst())
+      //     return true;
+      // FIXME: clad::restore_tracker is not thread-safe.
+      // We shoudn't disable reverse_forw for CUDA
+      if (FD->hasAttr<clang::CUDAGlobalAttr>() ||
+          FD->hasAttr<clang::CUDADeviceAttr>() ||
+          FD->hasAttr<clang::CUDAHostAttr>())
+        return false;
+      for (const ParmVarDecl* PVD : FD->parameters()) {
+        // Ignore dummy params
```
vgvassilev: Can you expand that comment. Why we have dummy params to ignore?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2127,6 +2137,28 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+      Expr* tapeExpr = nullptr;
+      QualType lastParamType =
+          calleeFnForwPassFD->parameters().back()->getType();
+      QualType tapeType = utils::GetRestoreTrackerType(m_Sema);
+      if (utils::hasMemoryTypeParams(FD) &&
+          utils::GetValueType(lastParamType) == tapeType) {
+        if (m_RestoreTracker) {
+          tapeExpr = m_RestoreTracker;
+        } else {
+          VarDecl* tapeDecl =
```
vgvassilev: We have renamed the smart tape class, however, variables and other identifiers still refer to the tracker as tape.

lib/Differentiator/CladUtils.cpp
PetroZarytskyi: There's a FIXME right above on line 959.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: Good catch. I looked up all "tape" mentions in the diff. I think I changed all of them.

lib/Differentiator/CladUtils.cpp
PetroZarytskyi: I expanded this comment and the other one you mentioned. Let me know if you want any other adjustments.

include/clad/Differentiator/RestoreTracker.h
```diff
@@ -0,0 +1,37 @@
...
+#include <cstdint>
+#include <cstring>
+#include <map>
+#include <utility>
+#include <vector>
+
+namespace clad {
+
+// NOLINTBEGIN(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+class restore_tracker {
```
vgvassilev, line 22: We need documentation here.

include/clad/Differentiator/RestoreTracker.h
```diff
@@ -0,0 +1,37 @@
...
+  std::map<char*, std::vector<uint8_t>> m_data;
```
vgvassilev: We need to explain the mapping. Is it value, raw_memory pair?

vgvassilev:
```suggestion
  using RawMemory = std::vector<uint8_t>;
  using Value = char*;
  std::map<Value, RawMemory> m_data;
```

include/clad/Differentiator/RestoreTracker.h
```diff
@@ -0,0 +1,37 @@
...
+  template <typename T> void store(const T& val) {
+    // Don't overwrite
+    if (m_data.find((char*)&val) != m_data.end())
+      return;
+    std::vector<uint8_t> buffer(sizeof(T));
+    std::memcpy(buffer.data(), &val, sizeof(T));
+    m_data.emplace((char*)&val, std::move(buffer));
+  }
+  void restore() {
+    for (auto& pair : m_data) {
```
vgvassilev:
```suggestion
    for (std::pair<Value, RawMemory>& pair : m_data) {
```

include/clad/Differentiator/ThrustDerivatives.h
```diff
@@ -378,6 +393,16 @@ void transform_pullback(InputIt1 first1, InputIt1 last1, InputIt2 first2,
...
+template <typename InputIt1, typename InputIt2, typename OutputIt,
+          typename BinaryOp>
+clad::ValueAndAdjoint<OutputIt, OutputIt>
+transform_reverse_forw(InputIt1 first1, InputIt1 last1, InputIt2 first2,
+                       OutputIt result, BinaryOp op, InputIt1 d_first1,
+                       InputIt1 /*d_last1*/, InputIt2 /*d_first2*/,
+                       OutputIt /*d_result*/, BinaryOp /*d_op*/) {
+  return {::thrust::transform(first1, last1, first2, result, op), {}};
+}
+
```
vgvassilev, line 405: Why we needed to add here thrust entities?

lib/Differentiator/CladUtils.cpp
```diff
@@ -890,6 +891,33 @@ namespace clad {
           clad_compat::ElaboratedTypeKeyword_None, NS, TT);
     }
 
+    clang::QualType GetRestoreTrackerType(clang::Sema& S) {
+      static QualType T;
+      if (T.isNull()) {
```
```suggestion
      if (!T.isNull())
        return T;
```

lib/Differentiator/CladUtils.cpp
```diff
@@ -890,6 +891,33 @@ namespace clad {
...
+        DeclarationName TrackerName =
+            &S.getASTContext().Idents.get("restore_tracker");
+        LookupResult TrackerR(S, TrackerName, noLoc, Sema::LookupUsingDeclName,
+                              CLAD_COMPAT_Sema_ForVisibleRedeclaration);
+        S.LookupQualifiedName(TrackerR, CladNS, CSS);
+        assert(!TrackerR.empty() && "cannot find clad::restore_tracker");
+
+        // This will instantiate restore_tracker<T> type and return it.
+        auto* RD = cast<RecordDecl>(TrackerR.getFoundDecl());
+        T = S.getASTContext().getRecordType(RD);
```
```suggestion
        ASTContext &C = S.getASTContext();
        T = C.getRecordType(RD);
```

lib/Differentiator/CladUtils.cpp
```diff
@@ -890,6 +891,33 @@ namespace clad {
...
+        // Get clad namespace and its identifier clad::.
+        NestedNameSpecifier* NS = CSS.getScopeRep();
+
+        // Create elaborated type with namespace specifier,
+        // i.e. class<T> -> clad::class<T>
+        T = S.getASTContext().getElaboratedType(
```
```suggestion
        T = C.getElaboratedType(
```

lib/Differentiator/CladUtils.cpp
vgvassilev: It is unclear that these were related to the commented lines...

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1303,7 +1304,7 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
           utils::isNonConstReferenceType(OCE->getArg(0)->getType());
     // Functions with side-effects require TBR.
     bool requestTBR = request.EnableTBRAnalysis &&
-                      (needsForwPass || isNonConstMethod) &&
+                      (scheduleForwPass || isNonConstMethod) &&
```
vgvassilev: Why TBR is dependent on `scheduleForwPass`?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1375,6 +1378,8 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
       }
     }
 
+    if (scheduleForwPass)
+      m_DiffRequestGraph.addNode(forwPassRequest, /*isSource=*/true);
```
vgvassilev: Can we avoid spilling of conditional code on `scheduleForwPass` somehow?

lib/Differentiator/ReverseModeForwPassVisitor.cpp
```diff
@@ -141,10 +144,35 @@ ReverseModeForwPassVisitor::BuildParams(DiffParams& diffParams) {
       m_Variables[*it] = BuildDeclRef(dPVD), m_DiffReq->getLocation();
     }
   }
+  if (utils::hasMemoryTypeParams(m_DiffReq.Function)) {
+    QualType trackerTy = utils::GetRestoreTrackerType(m_Sema);
+    trackerTy = m_Sema.getASTContext().getLValueReferenceType(trackerTy);
+    auto* trackerPVD = utils::BuildParmVarDecl(
```
vgvassilev: Spell out the type.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2127,6 +2137,28 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+          trackerExpr = BuildDeclRef(trackerDecl);
+          Expr* restoreCall = BuildCallExprToMemFn(
+              BuildDeclRef(trackerDecl), /*MemberFunctionName=*/"restore",
+              /*ArgExprs=*/{}, Loc);
+          it = std::begin(block) + insertionPoint;
+          block.insert(it, restoreCall);
+        }
+      }
+      if (trackerExpr)
+        CallArgs.push_back(trackerExpr);
```
vgvassilev: We need some comments for this block.

parth-07:
I was casually scrolling and stumbled upon this PR. This looks really great! I like the solution here for restoring the values. 

> _tracker0.store(arr[i]); // stores the memory address of arr[i] as well as its value

Will this solution work well for the user-defined types? I suppose it should be possible using storing the size and `memcpy`. However, are there any limitations that you are aware of?

include/clad/Differentiator/RestoreTracker.h
parth-07: Do you think it will be beneficial to run this loop parallelly?

lib/Differentiator/CladUtils.cpp
PetroZarytskyi: Would it be better if I put the commented-out part first? Alternatively, I could remove it

include/clad/Differentiator/RestoreTracker.h
PetroZarytskyi: I added more documentation to the class. Does it look better?

include/clad/Differentiator/RestoreTracker.h
PetroZarytskyi: This loop can be parallel, it's not supposed to touch common memory.
@vgvassilev What do you think? Should we use OpenMP or something else? I'm thinking, however, that ``std::map`` will probably not work with most of them. Would it be worth switching to a different type?

lib/Differentiator/DiffPlanner.cpp
PetroZarytskyi: I moved reverse_forw scheduling because we used to schedule them before traversing the function. Meaning if ``f`` calls ``g``, then ``f_reverse_forw`` was scheduled first. I can probably move the whole block handling reverse_forw after the function traversal.

include/clad/Differentiator/RestoreTracker.h
```diff
@@ -0,0 +1,55 @@
...
+/// f_reverse_forw(..., _tracker0);
+/// ...
+/// _tracker0.restore();
+/// f_pullback(...);
+/// We use it when we have to pass information between nested calls and
+/// clad::tape is not viable.
+class restore_tracker {
+  // m_data consists of pairs of memory locations and bitwise values
+  using RawMemory = std::vector<uint8_t>;
+  using Location = char*;
```
vgvassilev:
```suggestion
  using Address = char*;
```

include/clad/Differentiator/RestoreTracker.h
```diff
@@ -0,0 +1,55 @@
...
+#include <cassert>
+#include <cstdint>
+#include <cstring>
+#include <map>
+#include <utility>
+#include <vector>
+
+namespace clad {
+
+// NOLINTBEGIN(cppcoreguidelines-pro-bounds-pointer-arithmetic)
```
vgvassilev: I'd move these around the pointer arithmetic to annotate them better.

lib/Differentiator/CladUtils.cpp
vgvassilev: I guess, let's keep it as it is.

lib/Differentiator/CladUtils.cpp
```diff
@@ -927,6 +955,37 @@ namespace clad {
...
+      //   if (MD->isInstance() && !MD->isConst())
+      //     return true;
+      // FIXME: clad::restore_tracker is not thread-safe.
+      // We shoudn't disable reverse_forw for CUDA
+      if (FD->hasAttr<clang::CUDAGlobalAttr>() ||
+          FD->hasAttr<clang::CUDADeviceAttr>() ||
+          FD->hasAttr<clang::CUDAHostAttr>())
+        return false;
+      for (const ParmVarDecl* PVD : FD->parameters()) {
+        // We ignore dummy params because they are not essential for the
```
vgvassilev: What are `dummy` parameters in `FD`?

parth-07:
@PetroZarytskyi Firstly, thank you for taking the time out to describe this in detail!

> And if a user-defined type is modified with a method of some sort, then that's another call expr, and we can generate a nested reverse_forw and pass the _tracker further.

Oh, I see. Yes, you are right. Assignment operators on user-defined types are also a method call in-disguise.

> We should be completely sure that we store something that exists outside of the function. Obviously, we can construct an example when it's impossible to tell

This is interesting...

> double& ref = <some sophisticated condition> ? x : local;

Depending upon the condition, it can be impossible to determine this at compile-time.

> _tracker0.ignore(local);

This approach may have some shortcomings for local array variables, and class objects (`MyTensor tensor; tensor.default_val = 11;`, we would only be ignoring the address of `tensor`, not its members). I agree with your argument that it is rarely useful and should not be prioritized. However, it seems like an interesting problem to resolve in a generic way.

> However, we can always pick a class, an object of which is impossible to restore (just make it non-assignable, non-movable, and don't provide any obvious setters), so I don't know if this is a major problem.

Yes, this should not be a problem. We should definitely impose some sanity conditions such as class must be movable.

> For example, we can easily store/restore the capacity of a vector with hand-written code but I'm almost sure there's no standard-guaranteed way to get a reference to the capacity.

I think class member access is one of the recurring pain points. Additionally, any code involving memory construction / destruction quickly gets too complicated. In my opinion these cases are best handled by handwritten custom derivatives / specialized logic. I believe `restore_tracker` might be beneficial here as well.

For example:

Let's consider the below primal code:

```cpp
double &returnUorV(double &u, double &v) {
   return u;
}

void g(std::vector<double> &tensor, double val) {
   if (tensor.size() < 2)
     tensor.resize(2, val); 
   return;
}

  std::vector<double> tensor(1, u);
  double& some_val = returnUorV(u, v);
  g(tensor, some_val);
  double res = tensor[0] + tensor[1];
  return res;
}
```

The derivative for `fn` would be something like:

```cpp
void fn_grad(double u, double v, double *d_u, double *d_v) {
   std::vector<double> tensor(1, u);
   std::vector<double> d_tensor(tensor);
   clad::zero_init(d_tensor);
   // ...
   g(tensor, some_val);
   // ...
   // ...
   double _r0 = 0;
   // Needs to restore size of tensor!!!!!
   g_pullback(tensor, some_val, &d_tensor, _r0);
   // ...
}
```

We need to restore the size of `tensor` before `g_pullback` in order to get the correct derivative.

To do this, we can define the `g_reverse_forw` custom derivative as:

```cpp
void g_reverse_forw(std::vector<double> &tensor, double val, clad::restore_tracker tracker0) {
  if (tensor.size() < 2) {
    tracker.store_call(std::bind(&vector<double>::resize, tensor, tensor.size())); 
    tensor.resize(2, val);
  }
  return;
}
```

`tracker.restore()` can restore the `tensor` size by calling all the function calls stored by using `tracker.store_call`. This being said, this solution seems too complex to use. I think one nice benefit is that, thanks to `restore_tracker`, no change would be required in `g_pullback` for supporting this case.

PetroZarytskyi:
@parth-07 
> we would only be ignoring the address of tensor, not its members

Exactly. For that, we need to store memory intervals and not just simple addresses. That actually was part of this PR but I reverted it because I never used ``ignore`` in clad.

> tracker.store_call(std::bind(&vector<double>::resize, tensor, tensor.size()));

Hm, this is a very interesting idea, it could provide more flexibility. I think we can consider that if we want to enable this system for custom derivatives.

lib/Differentiator/CladUtils.cpp
PetroZarytskyi: I added a better explanation.

include/clad/Differentiator/RestoreTracker.h
PetroZarytskyi: I think they were meant for the code that was removed eventually. I removed these lines

lib/Differentiator/ReverseModeVisitor.cpp

lib/Differentiator/DiffPlanner.cpp
PetroZarytskyi: I added a commit doing that, and also making TBR scheduling separate from reverse_forw. Does this look better?

include/clad/Differentiator/ThrustDerivatives.h
PetroZarytskyi: For short, because these functions require reverse_forw. That's because they initialize arrays, that usually corresponds to some adjoint operations in the forward pass too. In this case, we don't care because the adjoint is a parameter and is properly initialized outside of the function. An example with local variables would probably require a reverse_forw. I didn't want to write a proper reverse_forw without seeing an actual use case.

Currently, on master, we don't even attempt to generate reverse_forw for most functions that might require it, and we only rely on custom derivatives. This PR extends the set of functions that we generate reverse_forw for. The work on nested TBR was necessary to get rid of as many new unnecessary reverse_forw as possible, but the system is still not perfect, and it produces a couple of unnecessary ones (although the one above does most likely require a reverse_forw).

vgvassilev: Can we track that into a separate issue?

lib/Differentiator/CladUtils.cpp
```diff
@@ -1164,12 +1235,18 @@ namespace clad {
...
-          !isa<CXXConstructorDecl>(FD)) {
+      if (forCustomDerv && !thisTy.isNull() && !isa<CXXConstructorDecl>(FD)) {
         FnTypes.insert(FnTypes.begin(), thisTy);
         EPI.TypeQuals.removeConst();
       }
 
+      if (mode == DiffMode::reverse_mode_forward_pass &&
+          shouldUseRestoreTracker(FD) && !forCustomDerv) {
+        QualType trackerTy = GetRestoreTrackerType(S);
+        trackerTy = S.getASTContext().getLValueReferenceType(trackerTy);
```
```suggestion
        trackerTy = C.getLValueReferenceType(trackerTy);
```

include/clad/Differentiator/ThrustDerivatives.h
PetroZarytskyi: I opened #1538. Let me know if you have better ideas about the title. The current one doesn't really cover the part about reverse_forw.

include/clad/Differentiator/RestoreTracker.h
vgvassilev: Let’s track this in a separate issue. We can’t really use openmp but we could use standard parallelism available in c++.

vgvassilev: We could use some of the explicit vectorization primitives.

lib/Differentiator/CladUtils.cpp
PetroZarytskyi: Fixed.

include/clad/Differentiator/RestoreTracker.h
PetroZarytskyi: I opened #1539.

PR #1450, opened 2025-07-16
Add primitive diagnostics for type mismatches of custom derivatives

PetroZarytskyi:
Currently, when a custom derivative is not accepted because of a signature mismatch, it's just silently ignored. This PR adds simple diagnostics telling the user explicitly that the provided derivative was not used, which signature was required, and which overloads were provided, as well as their signatures.
This is the first step to make writing custom derivatives easier. Ideally, we should add more diagnostics later, e.g., if the return type is wrong / no pullback parameter is provided for a custom pullback, etc. It would be even better to detect incorrect naming, but this seems way harder to do.


vgvassilev (approved):
LGTM!

PR #1452, opened 2025-07-17
Fix #1451: Support more complex subscript operator use

### LINKED ISSUE #1451: Clad fails to differentiate subscript operators
When differentiating through subscript operators of a container like `std::vector` where the subscript result is subsequently used/would be modified in the reverse pass, clad crashes.

Reproducible example:
```cpp

struct tmp {
  float z = 0;
  tmp(float val) : z(val) {}
  tmp() = default;
  void operator+=(const tmp &other) {
    z += other.z;
  }
  float forward(const float &x) const {
    return x + z;
  }
};

struct layer {
  std::vector<tmp> w;
  float forward(const float &inp) const {
    float x = inp;
    for (int i=0;i<w.size();i++) {
      x = w[i].forward(x);
    }
    return x;
  }
};

float func(const layer &l, float x) {
  return l.forward(x);
}

int main() {
  auto grad = clad::gradient(func, "0");
  grad.dump();
}
```

Clad crashes with:
```
error: too many arguments to function call, expected 1, have 2
   27 |       x = w[i].forward(x); // Forward through each tmp in the layer
      |            ~^~
```

Rohan-T144:
When the result of the subscript operator may affect the derivative, such as with `list[i].modify(x)`, clad currently marks it as nondifferentiable and skips creating the derivative/adjoint. This makes it follow the usual differentiation path.

Closes #1451.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -29,6 +29,7 @@
 #include "clang/AST/TemplateBase.h"
 #include "clang/AST/Type.h"
 #include "clang/Basic/LLVM.h" // for clang::isa
+#include "clang/Basic/OperatorKinds.h"
 #include "clang/Basic/SourceLocation.h"
 #include "clang/Basic/TargetInfo.h"
 #include "clang/Basic/TokenKinds.h"
@@ -1591,8 +1592,16 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     // derived function. In the case of member functions, `implicit`
     // this object is always passed by reference.
     if (!nonDiff && !dfdx() && !utils::HasAnyReferenceOrPointerArgument(FD) &&
-        (!baseOriginalE || MD->isConst()))
-      nonDiff = true;
+        (!baseOriginalE || MD->isConst())) {
+      // The result of the subscript operator may affect the derivative, such as
+      // in a case like `list[i].modify(x)`. This makes clad handle those
+      // normally.
+      if (const auto* OCE = dyn_cast<CXXOperatorCallExpr>(CE)) {
+        if (OCE->getOperator() != clang::OverloadedOperatorKind::OO_Subscript)
+          nonDiff = true;
+      } else
+        nonDiff = true;
+    }
 
     // If all arguments are constant literals, then this does not contribute to
     // the gradient.
@@ -2073,6 +2082,15 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     } // Recreate the original call expression.
 
     if (const auto* OCE = dyn_cast<CXXOperatorCallExpr>(CE)) {
+      if (OCE->getOperator() == clang::OverloadedOperatorKind::OO_Subscript) {
+        // If the operator is subscript, we should return the adjoint expression
+        auto AdjointCallArgs = CallArgs;
+        CallArgs.insert(CallArgs.begin(), baseDiff.getExpr());
+        AdjointCallArgs.insert(AdjointCallArgs.begin(), baseDiff.getExpr_dx());
+        call = BuildOperatorCall(OCE->getOperator(), CallArgs);
+        Expr* call_dx = BuildOperatorCall(OCE->getOperator(), AdjointCallArgs);
+        return StmtDiff(call, call_dx);
+      }
       if (isMethodOperatorCall)
         CallArgs.insert(CallArgs.begin(), baseDiff.getExpr());
       call = BuildOperatorCall(OCE->getOperator(), CallArgs);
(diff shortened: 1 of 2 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1591,8 +1592,19 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
     // this object is always passed by reference.
     if (!nonDiff && !dfdx() && !utils::HasAnyReferenceOrPointerArgument(FD) &&
-        (!baseOriginalE || MD->isConst()))
-      nonDiff = true;
+        (!baseOriginalE || MD->isConst())) {
+      bool isSubscriptOperator = false;
+      // Check if the current CallExpr is an OperatorCallExpr for operator[].
+      if (const auto* OCE = dyn_cast<CXXOperatorCallExpr>(CE)) {
+        if (OCE->getOperator() == clang::OverloadedOperatorKind::OO_Subscript)
+          isSubscriptOperator = true;
```
```suggestion
        if (OCE->getOperator() != clang::OverloadedOperatorKind::OO_Subscript)
          nonDiff = true;
```

That should simplify things and you don't need the rest of the changes I guess.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2076,6 +2088,13 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       if (isMethodOperatorCall)
         CallArgs.insert(CallArgs.begin(), baseDiff.getExpr());
       call = BuildOperatorCall(OCE->getOperator(), CallArgs);
+      if (OCE->getOperator() == clang::OverloadedOperatorKind::OO_Subscript) {
+        // If the operator is subscript, we should return the adjoint expression
+        CallArgs.erase(CallArgs.begin());
+        CallArgs.insert(CallArgs.begin(), baseDiff.getExpr_dx());
+        Expr* call_dx = BuildOperatorCall(OCE->getOperator(), CallArgs);
+        return StmtDiff(call, call_dx);
```
vgvassilev: @PetroZarytskyi, can you think of a way to do that without having to erase already added args?

test/Gradient/Loops.C
```diff
@@ -3260,6 +3261,67 @@ double fn41(double u, double v) {
...
+float fn42(const layer &l, float x) {
+  return l.forward(x);
+}
+//CHECK: void forward_pullback(const float &inp, float _d_y, layer *_d_this, float *_d_inp) const {
+//CHECK-NEXT:    int _d_i = 0;
+//CHECK-NEXT:    int i = 0;
+//CHECK-NEXT:    clad::tape<float> _t1 = {};
+//CHECK-NEXT:    float _d_x = 0.F;
+//CHECK-NEXT:    float x = inp;
+//CHECK-NEXT:    unsigned long _t0 = 0UL;
```
vgvassilev:
```suggestion
//CHECK-NEXT:    unsigned long _t0 = {{0U|0UL|0ULL}};
```

vgvassilev (approved):
LGTM!

test/Gradient/Loops.C
```diff
@@ -3260,6 +3261,67 @@ double fn41(double u, double v) {
...
+float fn42(const layer &l, float x) {
+  return l.forward(x);
+}
+//CHECK: void forward_pullback(const float &inp, float _d_y, layer *_d_this, float *_d_inp) const {
+//CHECK-NEXT:    int _d_i = 0;
+//CHECK-NEXT:    int i = 0;
+//CHECK-NEXT:    clad::tape<float> _t1 = {};
+//CHECK-NEXT:    float _d_x = 0.F;
+//CHECK-NEXT:    float x = inp;
+//CHECK-NEXT:    unsigned long _t0 = {{0U|0UL|0ULL}};
```
vgvassilev:
```suggestion
//CHECK-NEXT:    unsigned {{int|long}} _t0 = {{0U|0UL|0ULL}};
```

test/Gradient/Loops.C
```diff
@@ -3260,6 +3261,67 @@ double fn41(double u, double v) {
...
+//CHECK-NEXT:            if (!_t0)
+//CHECK-NEXT:                break;
+//CHECK-NEXT:        }
+//CHECK-NEXT:        i--;
+//CHECK-NEXT:        {
+//CHECK-NEXT:            x = clad::pop(_t1);
+//CHECK-NEXT:            float _r_d0 = _d_x;
+//CHECK-NEXT:            _d_x = 0.F;
+//CHECK-NEXT:            this->w[i].forward_pullback(x, _r_d0, &_d_this->w[i], &_d_x);
+//CHECK-NEXT:            size_type _r0 = 0UL;
```
vgvassilev:
```suggestion
//CHECK-NEXT:            size_type _r0 = {{0U|0UL}};
```

PR #1456, opened 2025-07-20
Move TBR infrastructure to AnalysisBase

ovdiiuv:


include/clad/Differentiator/AnalysisBase.h
```diff
@@ -0,0 +1,210 @@
...
+#include <map>
+#include <unordered_map>
+using namespace clang;
+
+namespace clad {
+/// ProfileID is the key type for ArrMap used to represent array indices
+/// and object fields.
+using ProfileID = clad_compat::FoldingSetNodeID;
+
+static ProfileID getProfileID(const Expr* E, ASTContext& Context) {
```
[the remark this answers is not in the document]
ovdiiuv: It is used infact.

include/clad/Differentiator/AnalysisBase.h
```diff
@@ -0,0 +1,227 @@
...
+
+#include "llvm/ADT/FoldingSet.h"
+
+#include <cstddef>
+#include <map>
+#include <memory>
+#include <unordered_map>
+#include <utility>
+
+using namespace clang;
```
vgvassilev:
```suggestion
```
We should not do using namespaces in header files. Wondering why clang-tidy did not complain..

lib/Differentiator/TBRAnalyzer.cpp
```diff
@@ -287,20 +65,8 @@ void TBRAnalyzer::setIsRequired(const clang::Expr* E, bool isReq) {
...
-
-TBRAnalyzer::VarData*
-TBRAnalyzer::getVarDataFromDecl(const clang::VarDecl* VD) {
-  auto* branch = &getCurBlockVarsData();
-  while (branch) {
-    auto it = branch->find(VD);
-    if (it != branch->end())
-      return &it->second;
-    branch = branch->m_Prev;
+    AnalysisBase::setIsRequired(data, isReq, IDSequence);
```
PetroZarytskyi, line 67: Do we need ``AnalysisBase::`` here? ``TBRAnalyzer`` is inherited from ``AnalysisBase`` and doesn't seem to override ``setIsRequired``

lib/Differentiator/TBRAnalyzer.cpp
```diff
@@ -318,12 +84,12 @@ void TBRAnalyzer::Analyze(const DiffRequest& request) {
...
   if (MD && !MD->isStatic()) {
     VarData& thisData = getCurBlockVarsData()[nullptr];
-    thisData = VarData(MD->getThisType(), m_Context);
+    thisData = VarData(MD->getThisType(), m_AnalysisDC->getASTContext());
     // We have to set all pointer/reference parameters to tbr
     // since method pullbacks aren't supposed to change objects.
     // constructor pullbacks don't take `this` as a parameter
     if (!isa<CXXConstructorDecl>(FD))
-      setIsRequired(&thisData);
+      AnalysisBase::setIsRequired(&thisData);
```
PetroZarytskyi, line 91: Same here

PetroZarytskyi (changes requested):
Apart from clang-tidy's complaints, the comment I left is the only issue. Everything else looks great. I think this is a good step towards bringing our analyses to the same framework.

lib/Differentiator/TBRAnalyzer.cpp
ovdiiuv: `setIsRequired` in `TBRAnalyzer.cpp` is TBR specific.

PetroZarytskyi: Okay, makes sense

PetroZarytskyi (approved):
Looks good to me! We'll see where specific functionality should be when we use this for other analyses, but that's something we can correct along the way.

Issue #1457, opened 2025-07-22
Regression in RooFit likelihoods


guitargeek:
The current Clad master crashes in cases like these:
```c++
#include <Math/CladDerivator.h>

inline double innerFunc(double const *params, unsigned int n, double const *low, double const *high)
{
   double total = 0.;
   for (std::size_t i = 0; i < n; ++i) {
      total += high[i] - low[i] + params[i] + total; // random compute
   }

   return total;
}

double myfunc(double *params, const double *obs, const double *xlArr)
{
   double t135[] = {params[0], params[1], params[2]};
   return innerFunc(t135, 3, xlArr, xlArr + 3);
}

#pragma clad ON
void gradient_request()
{
   clad::gradient(myfunc, "params");
}
#pragma clad OFF

// To run as a ROOT macro (if the file is called debug.C)
void debug() {}
```
The stack trace is:
```txt
   ------------------------------------------------------------------
  | Welcome to ROOT 6.37.01                        https://root.cern |
  | (c) 1995-2025, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Jan 01 1980, 00:00:00                 |
  | From heads/master@v6-37-01-7477-g6d7f915844e                     |
  | With g++ (GCC) 14.3.0                                            |
  | Try '.help'/'.?', '.demo', '.license', '.credits', '.quit'/'.q'  |
   ------------------------------------------------------------------

root [0]
Processing debug.C...

 *** Break *** segmentation violation



===========================================================
There was a crash.
This is the entire stack trace of all threads:
===========================================================
#0  0x00007f78e8b0a827 in wait4 () from /nix/store/g2jzxk3s7cnkhh8yq55l4fbvf639zy37-glibc-2.40-66/lib/libc.so.6
#1  0x00007f78e8a56b6b in do_system () from /nix/store/g2jzxk3s7cnkhh8yq55l4fbvf639zy37-glibc-2.40-66/lib/libc.so.6
#2  0x00007f78e95268c8 in TUnixSystem::Exec (this=0x1ae2f280, shellcmd=<optimized out>) at /home/rembserj/code/root/root_src/core/unix/src/TUnixSystem.cxx:2160
#3  TUnixSystem::StackTrace (this=0x1ae2f280) at /home/rembserj/code/root/root_src/core/unix/src/TUnixSystem.cxx:2451
[...]

PR #1459, opened 2025-07-23
Check if m_AnalysisDC is valid

ovdiiuv:
Fixes: #1457 

In the example 

```cpp

inline double innerFunc(double const *params, unsigned int n, double const *low, double const *high)
{
   double total = 0.;
      total += high[i] - low[i] + params[i] + total; // random compute
   }

   return total;
}
```

we have non-diff parameters, which is deduced during the visitation process, so the pullback is scheduled both statically and dynamically. TBR has to be run only on the pullbacks with defined `m_AnalysisDC`, hence the check in the `shouldBeRecorded` is added.

Once AA is default we wouldn't need to schedule those pullbacks dynamically, so this fix is somewhat temporary.

### DIFF
--- a/include/clad/Differentiator/DiffPlanner.h
+++ b/include/clad/Differentiator/DiffPlanner.h
@@ -8,15 +8,19 @@
 #include "clad/Differentiator/Timers.h"
 
 #include "clang/AST/Decl.h"
+#include "clang/AST/DeclBase.h"
 #include "clang/AST/ExprCXX.h"
 #include "clang/AST/RecursiveASTVisitor.h"
+#include "clang/Analysis/AnalysisDeclContext.h"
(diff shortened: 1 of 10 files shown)


include/clad/Differentiator/DiffPlanner.h
```diff
@@ -123,6 +128,8 @@ struct DiffRequest {
   /// This will be particularly useful for pushforward and pullback functions.
   bool DeclarationOnly = false;
 
+  clang::AnalysisDeclContext* m_AnalysisDC;
```
vgvassilev:
```suggestion
  clang::AnalysisDeclContext* m_AnalysisDC = nullptr;
```

PR #1462, opened 2025-07-26
Tensor-specific Clad changes

Rohan-T144:
Adds utility to check for any tensor types (including libtorch tensors), and forces differentiated variables to be copy-constructed to retain shape/size information.


lib/Differentiator/CladUtils.cpp
```diff
@@ -1380,5 +1380,21 @@ namespace clad {
       } while (E->IgnoreImplicit() != E);
       return false;
     }
+
+    bool isCladTorchTensor(const clang::QualType T) {
+      if (T.getAsString() == "std::vector<Tensor<float> >")
```
vgvassilev: I am wondering if it would be better if we have this as an attribute. It should not block this PR but maybe we should capture that with a comment and an issue.

r-timmaraju: That makes sense, I'll add a comment and make an issue on it.

test/Gradient/Cladtorch.C
```diff
@@ -0,0 +1,31 @@
...
+  float data;
+};
+}
+
+float fn1(const cladtorch::Tensor& t) {
+  auto b = t;
+  return b.data;
+}
+// CHECK: void fn1_grad(const cladtorch::Tensor &t, cladtorch::Tensor *_d_t) {
+// CHECK-NEXT:     cladtorch::cladtorch::Tensor _d_b(b);
```
vgvassilev: The double `cladtorch::cladtorch` seems to be a bug. Can you open an issue about it?

lib/Differentiator/CladUtils.cpp
```diff
@@ -1380,5 +1380,21 @@ namespace clad {
...
+        return true;
+      if (const auto* CXXRD = T->getAsCXXRecordDecl()) {
+        if (CXXRD->getNameAsString() == "Tensor") {
+          if (const auto* NS = clang::dyn_cast<clang::NamespaceDecl>(
+                  CXXRD->getDeclContext())) {
+            auto ns_name = NS->getNameAsString();
+            return ns_name == "cladtorch" || ns_name == "torch" ||
+                   ns_name == "at";
+          }
```
vgvassilev: I am thinking what if we have a special namespace in the `custom_derivatives` namespace which contains a set of forward declarations for the accepted tensor-like classes. Is the `torch::Tensor` class forward declarable?

r-timmaraju: I don't think `torch::Tensor` is forward declarable - that might not work.

test/Gradient/Cladtorch.C
r-timmaraju: Opened #1485.

lib/Differentiator/CladUtils.cpp
vgvassilev: Hm... Here is how to it seems they do it -- https://github.com/pytorch/pytorch/blob/6e8865fbc161270e2ffc52817e6c667df417a3f7/aten/src/ATen/core/TensorBase.h#L37-L40

lib/Differentiator/CladUtils.cpp
```diff
@@ -1466,5 +1466,21 @@ namespace clad {
       } while (E->IgnoreImplicit() != E);
       return false;
     }
+
+    bool isCladTorchTensor(const clang::QualType T) {
```
```suggestion
    bool isCladTorchTensor(clang::QualType T) {
```

lib/Differentiator/CladUtils.cpp
```diff
@@ -1466,5 +1466,68 @@ namespace clad {
       } while (E->IgnoreImplicit() != E);
       return false;
     }
+
+    bool isTensorLike(clang::Sema& SemaRef, clang::QualType T) {
```
vgvassilev, line 1492: Can we add a quick check for not `TemplateSpecializationType` or `CXXRecordDecl`. Should filter out a lot of the calls.

lib/Differentiator/CladUtils.cpp
```diff
@@ -1466,5 +1466,68 @@ namespace clad {
...
+      // Handle std::vector as a special case:
+      // We check if the type is a vector, and if so, recursively check its
+      // template argument.
+      if (const auto* TST = T->getAs<clang::TemplateSpecializationType>()) {
+        if (TST->getTemplateName()
+                .getAsTemplateDecl()
+                ->getQualifiedNameAsString() == "std::vector") {
```
vgvassilev: We can steal this check here: https://github.com/llvm/llvm-project/blob/9b0b2385b2b5f638af2f94181950b0f0d4819123/clang/lib/StaticAnalyzer/Checkers/LLVMConventionsChecker.cpp#L78

lib/Differentiator/CladUtils.cpp
```diff
@@ -1466,5 +1466,68 @@ namespace clad {
...
+      // Not a class, struct, or union, so it can't be a tensor.
+      if (!CXXRD)
+        return false;
+
+      // Find the special `tensor_like` namespace.
+      // This looks for `clad::tensor_like`.
+      clang::NamespaceDecl* CladNS =
+          utils::LookupNSD(SemaRef, "clad", /*shouldExist*/ true);
+      clang::NamespaceDecl* TensorLikeNS = utils::LookupNSD(
+          SemaRef, "tensor_like", /*shouldExist*/ false, CladNS);
```
```suggestion
      clang::NamespaceDecl* CladNS =
          utils::LookupNSD(SemaRef, "clad", /*shouldExist=*/true);
      clang::NamespaceDecl* TensorLikeNS = utils::LookupNSD(
          SemaRef, "tensor_like", /*shouldExist=*/false, CladNS);
```

lib/Differentiator/CladUtils.cpp
```diff
@@ -1466,5 +1466,68 @@ namespace clad {
...
+      if (!TensorLikeNS)
+        return false;
+
+      // Get the original type's name and its enclosing namespace.
+      clang::IdentifierInfo* TypeName = CXXRD->getIdentifier();
+      if (!TypeName)
```
vgvassilev, line 1529: Can we test that. Should be something like:
```cpp
struct {
    int i = 12;
} a;
```

lib/Differentiator/CladUtils.cpp
```diff
@@ -1466,5 +1466,68 @@ namespace clad {
...
+        return false;
+
+      const auto* OriginalDC = CXXRD->getDeclContext();
+      if (!clang::isa<clang::NamespaceDecl>(OriginalDC)) {
+        // The type is not in a namespace (e.g., it's a nested class or in
+        // global scope). You could extend this logic if needed, but for now,
+        // we'll assume tensor types are in a namespace like `at` or
+        // `cladtorch`.
+        return false;
+      }
```
```suggestion
      const auto* OriginalDC = CXXRD->getDeclContext();
      // The type is not in a namespace (e.g., it's a nested class or in
      // global scope). You could extend this logic if needed, but for now,
      // we'll assume tensor types are in a namespace like `at` or
      // `cladtorch`.
      if (!OriginalDC->isNamespace())
        return false;
```

lib/Differentiator/CladUtils.cpp
```diff
@@ -1466,5 +1466,68 @@ namespace clad {
...
+      const auto* OriginalNS = clang::cast<clang::NamespaceDecl>(OriginalDC);
+      clang::IdentifierInfo* OriginalNSName = OriginalNS->getIdentifier();
+      if (!OriginalNSName)
```
vgvassilev, line 1540: To trigger this we will need something in an anonymous namespace.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -411,10 +411,30 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
           }
           continue;
         }
-        auto VDDerivedType = utils::getNonConstType(paramTy, m_Sema);
+        auto VDDerivedType =
+            utils::GetNonConstValueType(paramTy); // Type(paramTy, m_Sema);
```
vgvassilev: I do not understand the comment. Maybe put it on the previous line and elaborate it.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -411,10 +411,30 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+          ParmVarDecl* newFuncParam = nullptr;
+          for (auto* p : m_Derivative->parameters()) {
+            if (p->getName() == param->getName()) {
+              newFuncParam = p;
+              break;
+            }
+          }
+          assert(
+              newFuncParam &&
+              "Could not find corresponding parameter in derivative function");
```
vgvassilev, line 428:
```suggestion
          assert(newFuncParam &&
              "Could not find corresponding parameter in derivative function");
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -411,10 +411,30 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+            utils::GetNonConstValueType(paramTy); // Type(paramTy, m_Sema);
+        Expr* initExpr = nullptr;
+        bool isDirectInit = false;
+        if (clad::utils::isTensorLike(m_Sema, VDDerivedType)) {
+          ParmVarDecl* newFuncParam = nullptr;
+          for (auto* p : m_Derivative->parameters()) {
+            if (p->getName() == param->getName()) {
+              newFuncParam = p;
+              break;
+            }
```
vgvassilev, line 424: @PetroZarytskyi, is there a better way to trigger direct initialization for the class type?

lib/Differentiator/CladUtils.cpp
```diff
@@ -1466,5 +1466,84 @@ namespace clad {
...
       return false;
     }
+
+    bool isTensorLike(clang::Sema& SemaRef, clang::QualType T) {
+      if (!isa<TemplateSpecializationType>(T) && !T->isRecordType())
+        return false;
+      // Handle std::vector as a special case:
+      // We check if the type is a vector, and if so, recursively check its
+      // template argument. These functions are taken from
+      // https://github.com/llvm/llvm-project/blob/aadc708e78568f1ec5713dd4ba768e77044b651d/clang/lib/StaticAnalyzer/Checkers/LLVMConventionsChecker.cpp#L78
```
```suggestion
      // We check if the type is a std::vector, and if so, recursively check its
      // template argument. 
```

lib/Differentiator/CladUtils.cpp
```diff
@@ -1466,5 +1466,84 @@ namespace clad {
...
+            if (Arg.getKind() == clang::TemplateArgument::Type)
+              return isTensorLike(SemaRef, Arg.getAsType());
+          }
+        }
+      }
+
+      const auto* CXXRD = T->getAsCXXRecordDecl();
+      // Not a class or struct, so it can't be a tensor.
+      if (!CXXRD)
+        return false;
```
```suggestion
```

This is now dead code.

lib/Differentiator/CladUtils.cpp
```diff
@@ -1466,5 +1466,84 @@ namespace clad {
...
+      const auto* OriginalNS = clang::cast<clang::NamespaceDecl>(OriginalDC);
+      clang::IdentifierInfo* OriginalNSName = OriginalNS->getIdentifier();
+      if (!OriginalNSName)
+        return false;
+
+      clang::NamespaceDecl* FoundMirroredNS =
+          utils::LookupNSD(SemaRef, OriginalNSName->getName(),
+                           /*shouldExist=*/false, TensorLikeNS);
+      if (!FoundMirroredNS)
+        return false;
```
vgvassilev, line 1547: Any ideas how to test this branch?

test/Gradient/Switch.C
```diff
@@ -701,7 +701,7 @@ double fn24(double x, double y, Op op) {
 }
 
 // CHECK: void fn24_grad_0_1(double x, double y, Op op, double *_d_x, double *_d_y) {
-// CHECK-NEXT:    Op _d_op = static_cast<Op>(0U);
+// CHECK-NEXT:    {{.*}} _d_op = {{.*}};
```
vgvassilev, line 683: Out of curiosity what is being produced here now? Why we have to glob?

r-timmaraju: On some specific platforms, `unsigned int _d_op = 0U;` is being produced instead, although I don't think this is ever actually used.

https://github.com/vgvassilev/clad/actions/runs/16872164064/job/47788847026#:~:text=Op%20_d_op%20=%20static_cast%3COp%3E(0U)

PR #1465, opened 2025-07-27
Fix loop increment crash 1436

### LINKED ISSUE #1436: Clad crashes if the loop increment expression is empty
Clad crashes if the loop increment expression is empty.

Godbolt reproducer: https://godbolt.org/z/oP78c9nzb

mohsinm-dev:
## Summary
  Fixes the segmentation fault that occurs when Clad attempts to differentiate for loops with empty increment expressions.

  ## Problem
  Issue #1436 reported that Clad would crash with a segfault when processing code like:

 ```
 for (int i = 0; i != 1 ;) {  // Empty increment
      sum += u + v;
      break;
  }
```

  The crash occurred in ReverseModeVisitor::VisitForStmt() when:
  1. dyn_cast<DeclRefExpr>(FS->getInc()) was called on nullptr
  2. cast<CompoundStmt>(incDiff.getStmt()) was called on uninitialized StmtDiff

  Solution

  Added defensive null checks before processing increment expressions:
  - Check FS->getInc() before calling dyn_cast
  - Guard increment processing with proper null checks
  - Ensure compound statement casting is safe

  Testing

  - ✅ Original failing case now compiles without segfault
  - ✅ Empty increment loops produce correct derivatives
  - ✅ Normal increment loops work as before (regression test)
  - ✅ Added comprehensive test in test/Misc/LoopIncrementFix.C

  Changes

  - lib/Differentiator/ReverseModeVisitor.cpp: Added null checks for increment handling
  - test/Misc/LoopIncrementFix.C: Added test case to prevent regression

  Fixes #1436

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 /.vscode
 /inst
 .cache/
+
+# Development activity log (personal tracking)
+activity.md
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1026,7 +1026,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     if (FS->getCond())
       std::tie(condDiff, condExprDiff) = DifferentiateSingleExpr(FS->getCond());
 
-    const auto* IDRE = dyn_cast<DeclRefExpr>(FS->getInc());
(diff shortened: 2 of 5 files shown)


guitargeek:
Master restructured this as `if (const Expr* inc = FS->getInc())`. I ran `for (int i=0; i<3;)` through `clad::gradient`. No crash, `dx=3`. Issue #1436 closed.

PR #1468, opened 2025-07-28
Don't create dynamic arrays for static arrays declarations inside loops

### LINKED ISSUE #1466: Clad creates a dynamic array for static array declarations inside for loops
Example:
```

double fn(int sum) {
  sum = 0; 
  for (int i = 0; i < 1; i++) {
      int a[1] = {1};
      sum += a[0];
  }
  return sum;
}

int main() { 
    return 0;
}
```
Produced code:
```
void fn_grad(int sum, int *_d_sum) {
    int _d_i = 0;
    int i = 0;
    clad::tape<clad::array<int> > _t1 = {};
    int _d_a[1] = {0};
    clad::array<int> a(1UL); // this should be the original static declaration
    sum = 0;
    unsigned long _t0 = 0UL;
    for (i = 0; ; i++) {
        {
            if (!(i < 1))
                break;
        }
        _t0++;
        clad::push(_t1, std::move(a)) , a = {1};
        sum += a[0];
    }
   . . .
}
```

PetroZarytskyi:
This PR fixes #1466. The only reason C arrays were replaced with dynamic ``clad::array`` is that C arrays are non-assignable and hard to handle. This PR creates exceptions in the way our store/restore system works so that such arrays can be handled.

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -252,6 +252,11 @@ namespace clad {
                                      llvm::StringRef prefix = "_t",
                                      bool moveToTape = false);
 
+    /// Build element-wise move between 2 arrays, e.g.
+    /// `std::move(std::begin(from), std::end(from), std::begin(to));`
+    clang::Expr* BuildArrayAssignment(clang::Expr* output, clang::Expr* input,
+                                      direction d);
+
     //// A type returned by DelayedGlobalStoreAndRef
     /// .Result is a reference to the created (yet uninitialized) global
     /// variable. When the expression is finally visited and rebuilt, .Finalize
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -59,19 +59,18 @@ using tape = tape_impl<T>;
 /// Add value to the end of the tape, return the same value.
 template <typename T, std::size_t SBO_SIZE = 64, std::size_t SLAB_SIZE = 1024,
           typename... ArgsT>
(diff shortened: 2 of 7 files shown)


include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -252,6 +252,11 @@ namespace clad {
                                      llvm::StringRef prefix = "_t",
                                      bool moveToTape = false);
 
+    /// Build element-wise move between 2 arrays, e.g.
+    /// `std::move(std::begin(from), std::end(from), std::begin(to));`
+    clang::Expr* BuildArrayAssignment(clang::Expr* output, clang::Expr* input,
```
vgvassilev, line 257: What happens if the elements are not movable?

PetroZarytskyi: Well, currently, the whole store/restore system for local declarations is based on moving. So if we want to support non-movable objects in this case, perhaps we can start with some examples and discuss how we want to approach this. But again, to be clear, we don't need to declare ``Non_Movable x[N]`` inside a loop to get an error, ``Non_Movable x`` would be enough.

vgvassilev: Ok, can we track this discussion in an issue?

kchristin22: I don't think there's an equivalent CUDA function for `std::move`. `std::array` is supported and it will be easy to incorporate that:
```
double arr[3] = {0}; --> std::array<double, 3> arr = {0};
...
for (...) {
   double (&&_t1)[3] = {seed, seed * i, seed + i};    --> clad::push(_t1, arr), arr = {seed, seed * i, seed + i};
   clad::push(_t2, arr) , std::move(std::begin(_t1), std::end(_t1), std::begin(arr));
   ...
}
```

PR #1472, opened 2025-07-31
Add custom derivative for thrust::reduce

a-elrawy:
This PR adds a custom derivative for thrust::reduce to support reverse-mode automatic differentiation of common parallel reduction patterns in Thrust.

Current supported operators: 

- plus
- maximum
- minimum
- multiplies

### DIFF
--- a/.github/workflows/clang-tidy-review.yml
+++ b/.github/workflows/clang-tidy-review.yml
@@ -35,7 +35,7 @@ jobs:
         with:
           build_dir: build
           apt_packages: cmake,libxml2,libxml2-dev,libtinfo-dev,zlib1g-dev,libzstd-dev,libthrust-dev
-          exclude: "test/*,unittests/*,benchmark/*,demos/*"
+          exclude: "test/*,unittests/*"
           split_workflow: true
           config_file: .clang-tidy
           cmake_command: >
--- a/.github/workflows/clang-tidy-review.yml
+++ b/.github/workflows/clang-tidy-review.yml
@@ -34,8 +34,8 @@ jobs:
         id: review
         with:
           build_dir: build
-          apt_packages: cmake,libxml2,libxml2-dev,libtinfo-dev,zlib1g-dev,libzstd-dev
-          exclude: "test/*,unittests/*"
+          apt_packages: cmake,libxml2,libxml2-dev,libtinfo-dev,zlib1g-dev,libzstd-dev,libthrust-dev
+          exclude: "test/*,unittests/*,benchmark/*,demos/*"
           split_workflow: true
           config_file: .clang-tidy
           cmake_command: >
--- a/include/clad/Differentiator/ThrustDerivatives.h
+++ b/include/clad/Differentiator/ThrustDerivatives.h
@@ -1,0 +1,183 @@
+#ifndef CLAD_DIFFERENTIATOR_THRUSTDERIVATIVES_H
+#define CLAD_DIFFERENTIATOR_THRUSTDERIVATIVES_H
+
+#include <cstddef>
+#include <iterator>
+#include <thrust/count.h>
+#include <thrust/device_vector.h>
+#include <thrust/execution_policy.h>
+#include <thrust/for_each.h>
+#include <thrust/functional.h>
+#include <thrust/iterator/constant_iterator.h>
+#include <thrust/iterator/zip_iterator.h>
+#include <thrust/reduce.h>
+#include <thrust/transform.h>
+#include <thrust/tuple.h>
+#include <type_traits>
+
+#include "clad/Differentiator/Differentiator.h"
+
+namespace clad::custom_derivatives::thrust {
+
+template <typename Iterator, typename T, typename BinaryOp>
+void reduce_pullback(Iterator first, Iterator last, T init, BinaryOp op,
+                     T d_output, Iterator* d_first, Iterator* d_last, T* d_init,
+                     BinaryOp* d_op) {
+  size_t n = ::thrust::distance(first, last);
+
+  auto d_first_const_ptr = ::thrust::raw_pointer_cast((*d_first).base());
+  auto d_first_ptr = const_cast<T*>(d_first_const_ptr);
+  ::thrust::device_ptr<T> d_first_dev_ptr(d_first_ptr);
+
+  if constexpr (::std::is_same_v<BinaryOp, ::thrust::plus<T>>) {
+    if (d_init)
+      *d_init += d_output;
+
+    struct add_d_output {
+      T d_output;
+      add_d_output(T d) : d_output(d) {}
+      __host__ __device__ void operator()(T& x) const { x += d_output; }
+    };
+
+    if (n > 0) {
+      ::thrust::for_each(d_first_dev_ptr, d_first_dev_ptr + n,
+                         add_d_output(d_output));
+    }
+
+  } else if constexpr (::std::is_same_v<BinaryOp, ::thrust::maximum<T>>) {
+    auto maxValue = ::thrust::reduce(first, last, init, ::thrust::maximum<T>());
+    auto k = (init == maxValue) ? 1 : 0;
+    k += ::thrust::count(first, last, maxValue);
+
+    if (k > 0) {
+      auto gradient = d_output / k;
+      if (d_init && init == maxValue)
+        *d_init += gradient;
+
+      struct max_grad_functor {
+        T max_val, gradient;
+        max_grad_functor(T mv, T g) : max_val(mv), gradient(g) {}
+        __host__ __device__ void
+        operator()(::thrust::tuple<T&, const T&> t) const {
+          if (::thrust::get<1>(t) == max_val)
+            ::thrust::get<0>(t) += gradient;
+        }
+      };
+
+      if (n > 0) {
+        auto iter = ::thrust::make_zip_iterator(
+            ::thrust::make_tuple(d_first_dev_ptr, first));
+        ::thrust::for_each(iter, iter + n,
+                           max_grad_functor(maxValue, gradient));
+      }
+    }
+
+  } else if constexpr (::std::is_same_v<BinaryOp, ::thrust::minimum<T>>) {
+    auto minValue = ::thrust::reduce(first, last, init, ::thrust::minimum<T>());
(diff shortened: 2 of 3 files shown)


include/clad/Differentiator/ThrustDerivatives.h
```diff
@@ -0,0 +1,183 @@
...
+  ::thrust::device_ptr<T> d_first_dev_ptr(d_first_ptr);
+
+  if constexpr (::std::is_same_v<BinaryOp, ::thrust::plus<T>>) {
+    if (d_init)
+      *d_init += d_output;
+
+    struct add_d_output {
+      T d_output;
+      add_d_output(T d) : d_output(d) {}
+      __host__ __device__ void operator()(T& x) const { x += d_output; }
```
vgvassilev:
```suggestion
      CUDA_HOST_DEVICE void operator()(T& x) const { x += d_output; }
```

include/clad/Differentiator/ThrustDerivatives.h
```diff
@@ -0,0 +1,183 @@
...
+
+    if (k > 0) {
+      auto gradient = d_output / k;
+      if (d_init && init == maxValue)
+        *d_init += gradient;
+
+      struct max_grad_functor {
+        T max_val, gradient;
+        max_grad_functor(T mv, T g) : max_val(mv), gradient(g) {}
+        __host__ __device__ void
```
vgvassilev:
```suggestion
        CUDA_HOST_DEVICE void
```

include/clad/Differentiator/ThrustDerivatives.h
```diff
@@ -0,0 +1,183 @@
...
+
+    if (k > 0) {
+      auto gradient = d_output / k;
+      if (d_init && init == minValue)
+        *d_init += gradient;
+
+      struct min_grad_functor {
+        T min_val, gradient;
+        min_grad_functor(T mv, T g) : min_val(mv), gradient(g) {}
+        __host__ __device__ void
```
vgvassilev:
```suggestion
        CUDA_HOST_DEVICE void
```

include/clad/Differentiator/ThrustDerivatives.h
```diff
@@ -0,0 +1,183 @@
...
+      };
+      T non_zero_product = ::thrust::reduce(
+          ::thrust::make_transform_iterator(first, replace_zero_with_one()),
+          ::thrust::make_transform_iterator(last, replace_zero_with_one()),
+          init, ::thrust::multiplies<T>());
+
+      struct single_zero_grad_functor {
+        T gradient;
+        single_zero_grad_functor(T g) : gradient(g) {}
+        __host__ __device__ void
```
vgvassilev:
```suggestion
        CUDA_HOST_DEVICE void
```

include/clad/Differentiator/ThrustDerivatives.h
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+  } else if constexpr (::std::is_same_v<BinaryOp, ::thrust::multiplies<T>>) {
+    size_t zero_count = ::thrust::count(first, last, 0);
+    bool init_is_zero = (init == 0);
+
+    if (zero_count > 1 || (zero_count == 1 && init_is_zero)) {
+    } else if (zero_count == 1 && !init_is_zero) {
+      struct replace_zero_with_one {
+        __host__ __device__ T operator()(const T& x) const {
```
vgvassilev:
```suggestion
        CUDA_HOST_DEVICE T operator()(const T& x) const {
```

include/clad/Differentiator/ThrustDerivatives.h
```diff
@@ -0,0 +1,183 @@
...
+#include <thrust/for_each.h>
+#include <thrust/functional.h>
+#include <thrust/iterator/constant_iterator.h>
+#include <thrust/iterator/zip_iterator.h>
+#include <thrust/reduce.h>
+#include <thrust/transform.h>
+#include <thrust/tuple.h>
+#include <type_traits>
+
+#include "clad/Differentiator/Differentiator.h"
```
vgvassilev: Why do we need this header for?

a-elrawy:
> @a-elrawy, can you fix the clang-tidy complaints?

Should be fixed now

include/clad/Differentiator/ThrustDerivatives.h
```diff
@@ -0,0 +1,179 @@
...
+      }
+    }
+
+  } else if constexpr (::std::is_same_v<BinaryOp, ::thrust::multiplies<T>>) {
+    size_t zero_count = ::thrust::count(first, last, 0);
+    bool init_is_zero = (init == 0);
+
+    if (zero_count > 1 || (zero_count == 1 && init_is_zero)) {
+    } else if (zero_count == 1 && !init_is_zero) {
+      struct replace_zero_with_one {
```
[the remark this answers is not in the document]
vgvassilev: Doesn’t this interface return a T that has uninitialized members?

include/clad/Differentiator/ThrustDerivatives.h
```diff
@@ -0,0 +1,179 @@
...
+    } else {
+      auto product =
+          ::thrust::reduce(first, last, init, ::thrust::multiplies<T>());
+
+      if (d_init)
+        *d_init += d_output * product / init;
+
+      struct multiplies_grad_functor {
+        T d_output, product;
+        multiplies_grad_functor(T d, T p) : d_output(d), product(p) {}
```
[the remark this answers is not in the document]
vgvassilev: Is product initialized here?

I’d prefer different data members to be on separate lines.

include/clad/Differentiator/ThrustDerivatives.h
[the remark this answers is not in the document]
a-elrawy: I don't think it has data members that would need initialization. I've added `// NOLINTNEXTLINE(cppcoreguidelines-pro-type-member-init)` to remove the warning

include/clad/Differentiator/ThrustDerivatives.h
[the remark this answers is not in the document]
a-elrawy: Yes, it should be initialized here via the constructer. the value is from `thrust::reduce`. 

I have updated them to be in separate lines now.

include/clad/Differentiator/ThrustDerivatives.h
```diff
@@ -0,0 +1,184 @@
...
+      auto product =
+          ::thrust::reduce(first, last, init, ::thrust::multiplies<T>());
+
+      if (d_init)
+        *d_init += d_output * product / init;
+
+      struct multiplies_grad_functor {
+        T d_output;
+        T product;
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-member-init)
```
vgvassilev, line 149: I do not understand why we need this suppression.

a-elrawy: I still get this warning from clang-tidy 
```
warning: constructor does not initialize these fields: T [[cppcoreguidelines-pro-type-member-init]]

include/clad/Differentiator/ThrustDerivatives.h:150:
```

but it should be initialized correctly already. I think the problem happen when there's a return type after `CUDA_HOST_DEVICE`. The warning isn't there when it's `void`

vgvassilev: Interesting... is that some clang-tidy bug that we are seeing?

a-elrawy: Yes, I think so

vgvassilev:
Can you rebase this PR and click resolve all resolved clang-tidy comments.

.github/workflows/clang-tidy-review.yml
```diff
@@ -34,8 +34,8 @@ jobs:
         id: review
         with:
           build_dir: build
-          apt_packages: cmake,libxml2,libxml2-dev,libtinfo-dev,zlib1g-dev,libzstd-dev
-          exclude: "test/*,unittests/*"
+          apt_packages: cmake,libxml2,libxml2-dev,libtinfo-dev,zlib1g-dev,libzstd-dev,libthrust-dev
+          exclude: "test/*,unittests/*,benchmark/*,demos/*"
```
vgvassilev:
```suggestion
          exclude: "test/*,unittests/*"
```

include/clad/Differentiator/ThrustDerivatives.h
```diff
@@ -0,0 +1,184 @@
...
+                         add_d_output(d_output));
+    }
+
+  } else if constexpr (::std::is_same_v<BinaryOp, ::thrust::maximum<T>>) {
+    auto maxValue = ::thrust::reduce(first, last, init, ::thrust::maximum<T>());
+    auto k = (init == maxValue) ? 1 : 0;
+    k += ::thrust::count(first, last, maxValue);
+
+    if (k > 0) {
+      auto gradient = d_output / k;
```
kchristin22, line 49: Hi : ), why should we normalize the gradient here? Is this how AD works in these cases?

a-elrawy: Yes, I think so; because max/min is non-differentiable in this case (when there's non-unique max/min), we use the norm subgradient to split the gradient equally between them.

vgvassilev: We have `min` and `max` in the form of pullbacks here:

https://github.com/vgvassilev/clad/blob/ec1cfbd60fa83234a580100a18a7248c93f7916a/include/clad/Differentiator/BuiltinDerivatives.h#L1340

Why we can't use the same approach for `::thrust::maximum` and friends?

a-elrawy: This PR is mainly for the reduction operation. I'll later work on the other normal operations like max/min. They'll have a similar approach like these ones.

vgvassilev: Can we add them first and build this PR on top of them because it makes some assumptions about them not being available.

a-elrawy: Sure, I'll add them, though I'm not sure if it'll affect this PR. I think both are separate.

PR #1481, opened 2025-08-10
Enable TBR in the tests of the produced code

PetroZarytskyi:
Currently, TBR analysis is turned on by default, but it's still manually disabled in the tests. This makes it harder to see what our real produced code looks like. Turning it on in tests will also allow us to see which improvements need to be made on the analysis side.


vgvassilev (approved):
LGTM!

PR #1484, opened 2025-08-11
Analyze local DeclStmt in TBR

PetroZarytskyi:
At some point, we started promoting local variables (e.g., inside loops/ifs) to the function global scope so that they are accessible in the reverse pass. Every time we do that inside a loop, we generate store/restore statements. For example,
```
while (...) {
    double local_var = ...;
    ...
}
```
is differentiated as
```
// forward pass
double local_var;
clad::tape<double> _t0;
while (...) {
    clad::push(_t0, local_var), local_var = ...;
    ...
}
...
// reverse pass
while (...) {
    ...
    local_var = clad::pop(_t0);
}
```
However, the values of such local variables are sometimes unused in the reverse pass. And this is exactly the purpose of TBR analysis to remove such unnecessary pushes. This PR attaches TBR to this system.

### DIFF
--- a/include/clad/Differentiator/DiffPlanner.h
+++ b/include/clad/Differentiator/DiffPlanner.h
@@ -180,7 +180,7 @@ public:
   void print(llvm::raw_ostream& Out) const;
   LLVM_DUMP_METHOD void dump() const { print(llvm::errs()); }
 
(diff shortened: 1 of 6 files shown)


vgvassilev (approved):
LGTM!

PR #1491, opened 2025-08-18
Add support for OpenMP in forward mode

Errant404:
This PR introduces OpenMP support for Clad's forward mode, covering directives like `#pragma omp parallel for` and `#pragma omp parallel`, along with five commonly used clauses: `shared`, `firstprivate`, `private`, `lastprivate`, and `reduction`.

### DIFF
--- a/lib/Differentiator/BaseForwardModeVisitorOpenMP.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitorOpenMP.cpp
@@ -12,6 +12,8 @@
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/Frontend/OpenMP/OMP.h.inc"
 
+#include <cassert>
+
 using namespace clang;
 using namespace llvm::omp;
 
@@ -90,14 +92,13 @@ StmtDiff BaseForwardModeVisitor::VisitOMPExecutableDirective(
   ArrayRef<OMPClause*> Clauses = D->clauses();
   TClauses.reserve(Clauses.size());
   for (auto* I : Clauses) {
-    if (I) {
-      CLAD_COMPAT_CLANG19_SemaOpenMP(m_Sema).StartOpenMPClause(
-          I->getClauseKind());
-      OMPClause* Clause = Visit(I);
-      CLAD_COMPAT_CLANG19_SemaOpenMP(m_Sema).EndOpenMPClause();
-      if (Clause)
-        TClauses.push_back(Clause);
-    }
+    assert(I);
+    CLAD_COMPAT_CLANG19_SemaOpenMP(m_Sema).StartOpenMPClause(
+        I->getClauseKind());
+    OMPClause* Clause = Visit(I);
+    CLAD_COMPAT_CLANG19_SemaOpenMP(m_Sema).EndOpenMPClause();
+    assert(Clause);
+    TClauses.push_back(Clause);
   }
   StmtDiff AssociatedStmt;
   if (D->hasAssociatedStmt() && D->getAssociatedStmt()) {


lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -2204,4 +2205,110 @@ clang::Expr* BaseForwardModeVisitor::BuildCustomDerivativeConstructorPFCall(
       customPushforwardName, customPushforwardArgs, getCurrentScope(), CE);
   return pushforwardCall;
 }
+
+clang::OMPClause* BaseForwardModeVisitor::VisitOMPReductionClause(
```
vgvassilev: This should probably go into a new file called `BaseForwardModeVisitorOpenMP.cpp`

lib/Differentiator/BaseForwardModeVisitorOpenMP.cpp
```diff
@@ -0,0 +1,176 @@
...
+  DeclarationNameInfo NameInfo = C->getNameInfo();
+  return CLAD_COMPAT_CLANG19_SemaOpenMP(m_Sema).ActOnOpenMPReductionClause(
+      Vars, CLAD_COMPAT_CLANG21_getModifier(C), C->getBeginLoc(),
+      C->getLParenLoc(), C->getModifierLoc(), C->getColonLoc(), C->getEndLoc(),
+      ReductionIdScopeSpec, NameInfo);
+}
+
+OMPClause* BaseForwardModeVisitor::VisitOMPClause(const OMPClause* C) {
+  if (!C)
+    return nullptr;
```
vgvassilev: Is that reachable code?

lib/Differentiator/BaseForwardModeVisitorOpenMP.cpp
```diff
@@ -0,0 +1,176 @@
...
+  case OMPC_private:
+    return VisitOMPPrivateClause(cast<OMPPrivateClause>(C));
+  case OMPC_firstprivate:
+    return VisitOMPFirstprivateClause(cast<OMPFirstprivateClause>(C));
+  case OMPC_lastprivate:
+    return VisitOMPLastprivateClause(cast<OMPLastprivateClause>(C));
+  case OMPC_shared:
+    return VisitOMPSharedClause(cast<OMPSharedClause>(C));
+  case OMPC_reduction:
+    return VisitOMPReductionClause(cast<OMPReductionClause>(C));
```
vgvassilev: Is that dispatch mechanism needed? We should probably just override these methods instead of the `VisitOMPClause`...

lib/Differentiator/BaseForwardModeVisitorOpenMP.cpp
```diff
@@ -0,0 +1,176 @@
...
+  for (auto* I : Clauses) {
+    if (I) {
+      CLAD_COMPAT_CLANG19_SemaOpenMP(m_Sema).StartOpenMPClause(
+          I->getClauseKind());
+      OMPClause* Clause = VisitOMPClause(I);
+      CLAD_COMPAT_CLANG19_SemaOpenMP(m_Sema).EndOpenMPClause();
+      if (Clause)
+        TClauses.push_back(Clause);
+    } else {
+      TClauses.push_back(nullptr);
```
vgvassilev: Can we come up with a test here?

include/clad/Differentiator/BaseForwardModeVisitor.h
```diff
@@ -123,6 +125,20 @@ class BaseForwardModeVisitor
...
+  StmtDiff
+  VisitOMPParallelForDirective(const clang::OMPParallelForDirective* D);
+  clang::OMPClause* VisitOMPClause(const clang::OMPClause* C);
+  clang::OMPClause* VisitOMPPrivateClause(const clang::OMPPrivateClause* C);
+  clang::OMPClause*
+  VisitOMPFirstprivateClause(const clang::OMPFirstprivateClause* C);
+  clang::OMPClause*
+  VisitOMPLastprivateClause(const clang::OMPLastprivateClause* C);
+  clang::OMPClause* VisitOMPSharedClause(const clang::OMPSharedClause* C);
+  clang::OMPClause* VisitOMPReductionClause(const clang::OMPReductionClause* C);
```
PetroZarytskyi, line 147: We generally use ``Visit-`` naming only when we actually override some visitor methods. In other cases, to avoid confusion, we use other prefixes like ``Differentiate-``: ``DifferentiateVarDecl``, ``DifferentiateSingleStmt``, ``DifferentiateCtorInit``. You could also use ``Derive-`` to make it shorter, or whatever prefix you think would make the most sense in this context.

I'm specifically talking about the functions ``VisitOMPClause``, ``VisitOMPPrivateClause``, ``VisitOMPFirstprivateClause``, ``VisitOMPLastprivateClause``, ``VisitOMPSharedClause``, ``VisitOMPReductionClause``.

lib/Differentiator/BaseForwardModeVisitorOpenMP.cpp
Errant404: You mean repurposing the existing Visit() as the dispatch method?
But that won't work since OMPClause isn't a Stmt. 
We have to create a separate dispatch mechanism – which is also how LLVM handles it (see https://github.com/llvm/llvm-project/blob/llvmorg-21.1.4/clang/lib/Sema/SemaOpenMP.cpp#L17488).

lib/Differentiator/BaseForwardModeVisitorOpenMP.cpp
Errant404: I actually think this is mostly unreachable code in practice, because by the time clad processes it, the AST generated by Clang should be robust enough. 
I just added this because I saw similar handling in Clang's codebase. Maybe we should remove it?

lib/Differentiator/BaseForwardModeVisitorOpenMP.cpp
Errant404: Same as https://github.com/vgvassilev/clad/pull/1491#discussion_r2478687471
I actually can't think of a test case that would cover this code.

lib/Differentiator/BaseForwardModeVisitorOpenMP.cpp
vgvassilev: You can take a look at https://github.com/llvm/llvm-project/blob/222fc11f2b8f25f6a0f4976272ef1bb7bf49521d/clang/lib/AST/OpenMPClause.cpp#L2309 

The visitor supports these and I think my suggestion is correct...

Errant404: You're right. I've already implemented the relevant changes.

include/clad/Differentiator/BaseForwardModeVisitor.h
Errant404: After [6979511](https://github.com/vgvassilev/clad/pull/1491/commits/69795117af3bdf8e923beff8011e73b70f63be79), we now override several visitor methods.

lib/Differentiator/BaseForwardModeVisitorOpenMP.cpp
vgvassilev: So maybe replace the else part with an assert?

Errant404: Done.

include/clad/Differentiator/BaseForwardModeVisitor.h
PetroZarytskyi: Amazing! This is better than what I suggested.

PetroZarytskyi (approved):
LGTM. Great work!

PR #1504, opened 2025-08-25
Only use the reverse mode in error estimation

### LINKED ISSUE #1498: "too many arguments to function call" Error for generated `fun_pushforward` with arc length example
When running Clad v2.0 on the following example:

```c++
#ifndef ARCLEN_HPP
#define ARCLEN_HPP

#include <cmath>
#include <iostream>
#define ABS(x) (((x) < 0.0) ? (-(x)) : (x))

#define PI 3.1415926535897932L
#define ANS 5.795776322412856L
#define EPS 1e-10

template <typename T, typename L = T> T fun(T x) {
  L d1 = 1.0;
  L d2 = d1; // also d1 in original
  T t3 = x;  // also t1 in original
  int k;
  for (k = 1; k <= 5; k += 1) {
    d2 = 2.0 * d2;
    t3 = t3 + sin(d2 * x) / d2;
  }
  return t3;
}

template <typename T, typename L = T> T do_fun(int iterations) {
  T t2;
  T h = PI/iterations;
  T t1 = 0;
  T s1 = 0;
  int i;
  for (i = 1; i <= iterations; i += 1) {
    t2 = fun(i * h);
    s1 = s1 + sqrt(h * h + (t2 - t1) * (t2 - t1));
    t1 = +t2;
  }
  return s1;
}

#endif

int main() {
  clad::estimate_error(do_fun<double>);
}
```

The following error is emitted:

```
/benchmarks/arclen.hpp:24:41: error: too many arguments to function call, expected 2, have 3
   12 | template <typename T, typename L = T> T fun(T x) {
      |                                         ~
   13 |   L d1 = 1.0;
   14 |   L d2 = d1; // also d1 in original
   15 |   T t3 = x;  // also t1 in original
   16 |   int k;
   17 |   for (k = 1; k <= 5; k += 1) {
   18 |     d2 = 2.0 * d2;
   19 |     t3 = t3 + sin(d2 * x) / d2;
   20 |   }
   21 |   return t3;
   22 | }
   23 |
   24 | template <typename T, typename L = T> T do_fun(int iterations) {
      |                                         ^~~~
/benchmarks/arclen.hpp:12:41: note: 'fun_pushforward' declared here
   12 | template <typename T, typename L = T> T fun(T x) {
```

Clad v1.9 did not have this issue. I have pinpointed this to commit d749a83251f1739995491707c3713dc7f0ef30c0 before which example compiled without errors. For reference I am using Clad as follows:

```
clang++-19 -std=c++14 -fplugin=$CLAD_PLUGIN -fplugin-arg-clad--fdump-derived-fn -I$CLAD_INCLUDE_DIR example.cpp
```

PetroZarytskyi:
We sometimes use pushforwards in the reverse mode for simplicity/efficiency. To be precise, when a nested non-member function has only one numerical by-value parameter, then we are safe to replace the pullback with a pushforward, even in the reverse mode. However, the approach only works to compute the gradient, in error estimation, we need only pullbacks. Issue #1498 occurs because Clad attempts to pass the additional ``_final_error`` arg to a pushforward.
Fixes #1498.

### DIFF
--- a/test/ErrorEstimation/LoopsAndArrays.C
+++ b/test/ErrorEstimation/LoopsAndArrays.C
@@ -310,7 +310,7 @@ double func6(double x) {
 //CHECK-NEXT:     clad::tape<double> _t1 = {};
 //CHECK-NEXT:     double _d_sum = 0.;
 //CHECK-NEXT:     double sum = 0;
-//CHECK-NEXT:     unsigned long _t0 = 0UL;
+//CHECK-NEXT:     unsigned {{int|long}} _t0 = {{0U|0UL}};
 //CHECK-NEXT:     for (i = 1; i < 10; i++) {
 //CHECK-NEXT:         _t0++;
 //CHECK-NEXT:         clad::push(_t1, sum);
--- a/test/ErrorEstimation/LoopsAndArrays.C
+++ b/test/ErrorEstimation/LoopsAndArrays.C
@@ -282,10 +282,60 @@ double func5(double* x, double* y, double* output) {
 //CHECK-NEXT:     _final_error += std::abs(1. * _ret_value0 * {{.+}});
 //CHECK-NEXT: }
 
+double fun(double x) {
+    return x*x;
+}
+
+//CHECK: void fun_pullback(double x, double _d_y, double *_d_x, double &_final_error) {
+//CHECK-NEXT:     double _ret_value0 = 0.;
+//CHECK-NEXT:     _ret_value0 = x * x;
+//CHECK-NEXT:     {
+//CHECK-NEXT:         *_d_x += _d_y * x;
+//CHECK-NEXT:         *_d_x += x * _d_y;
+//CHECK-NEXT:     }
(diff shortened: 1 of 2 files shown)


test/ErrorEstimation/LoopsAndArrays.C
```diff
@@ -282,10 +282,60 @@ double func5(double* x, double* y, double* output) {
...
+  return sum;
+}
+
+//CHECK: void func6_grad(double x, double *_d_x, double &_final_error) {
+//CHECK-NEXT:     int _d_i = 0;
+//CHECK-NEXT:     int i = 0;
+//CHECK-NEXT:     clad::tape<double> _t1 = {};
+//CHECK-NEXT:     double _d_sum = 0.;
+//CHECK-NEXT:     double sum = 0;
+//CHECK-NEXT:     unsigned long _t0 = 0UL;
```
vgvassilev:
```suggestion
//CHECK-NEXT:     unsigned {{int|long}} _t0 = {{0U|0UL}};
```

vgvassilev (approved):
LGTM!

PR #1506, opened 2025-08-25
Find used parameter values in nested TBR

PetroZarytskyi:
After #1490, TBR can detect which parameters have been modified. This PR extends the analysis of the same class of functions to also track whether the function uses the given parameter value, which allows for more optimization in TBR. In particular, now we can avoid storing some arguments even if they are modified. Previously, we assumed that every parameter value was used, which is not true, for example, for most STL methods, e.g., ``push_back``, ``fill``, etc.

During the work on this PR, there occured a bug, the reason for which was that return stmts were not properly visited in TBR. Fixing that caused regressions in some tests, and the goal of the second commit is to deal with those regressions.
The core idea there is that we don't have to store operands if a whole expr is already stored. In particular, we often store LHS/RHS of multiplication/division.

### DIFF
--- a/lib/Differentiator/TBRAnalyzer.cpp
+++ b/lib/Differentiator/TBRAnalyzer.cpp
@@ -63,8 +63,12 @@ void TBRAnalyzer::setIsRequired(const clang::Expr* E, bool isReq) {
   // Make sure the current branch has a copy of VarDecl for VD
   auto& curBranch = getCurBlockVarsData();
   for (const VarDecl* iterVD : vars) {
-    if (m_ModifiedParams && !isReq && (!iterVD || isa<ParmVarDecl>(iterVD)))
-      (*m_ModifiedParams)[m_Function].insert(iterVD);
+    if (m_ModifiedParams && (!iterVD || isa<ParmVarDecl>(iterVD))) {
+      if (!isReq)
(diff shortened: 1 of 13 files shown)


lib/Differentiator/TBRAnalyzer.cpp
```diff
@@ -63,8 +63,12 @@ void TBRAnalyzer::setIsRequired(const clang::Expr* E, bool isReq) {
   // Make sure the current branch has a copy of VarDecl for VD
   auto& curBranch = getCurBlockVarsData();
   for (const VarDecl* iterVD : vars) {
-    if (m_ModifiedParams && !isReq && (!iterVD || isa<ParmVarDecl>(iterVD)))
-      (*m_ModifiedParams)[m_Function].insert(iterVD);
+    if (m_ModifiedParams && (!iterVD || isa<ParmVarDecl>(iterVD))) {
+      if (!isReq)
+        (*m_ModifiedParams)[m_Function].insert(iterVD);
+      else if (m_ModeStack.back() & Mode::kMarkingMode)
```
vgvassilev, line 69:
```suggestion
      else if (m_ModeStack.back() && Mode::kMarkingMode)
```

PR #1508, opened 2025-08-26
Reimplement VA using AnalysisBase infrastructure

ovdiiuv:
This PR reworks varied analysis using the infrastructure in `AnalysisBase`with support for loops, function calls, and pointer/reference handling. This is the first step toward making varied analysis default.

### DIFF
--- a/lib/Differentiator/ActivityAnalyzer.cpp
+++ b/lib/Differentiator/ActivityAnalyzer.cpp
@@ -1,18 +1,47 @@
 #include "ActivityAnalyzer.h"
+#include "AnalysisBase.h"
+
+#include "clang/AST/Decl.h"
+#include "clang/AST/Expr.h"
+#include "clang/AST/ExprCXX.h"
+#include "clang/AST/Stmt.h"
 #include "clang/AST/Type.h"
+
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/Casting.h"
+
+#include <memory>
+#include <set>
 
 using namespace clang;
 
 namespace clad {
 
-void VariedAnalyzer::Analyze(const FunctionDecl* FD) {
+void VariedAnalyzer::Analyze() {
   m_BlockData.resize(m_AnalysisDC->getCFG()->size());
   // Set current block ID to the ID of entry the block.
+
   CFGBlock& entry = m_AnalysisDC->getCFG()->getEntry();
   m_CurBlockID = entry.getBlockID();
-  m_BlockData[m_CurBlockID] = createNewVarsData({});
-  for (const VarDecl* i : m_VariedDecls)
-    m_BlockData[m_CurBlockID]->insert(i);
+
+  m_BlockData[m_CurBlockID] = std::make_unique<VarsData>();
+
+  for (const auto* i : m_DiffReq.getVariedDecls()) {
+    addVar(i, /*forceInit=*/true);
+    setIsRequired(getVarDataFromDecl(i));
+  }
+
+  auto paramsRef = m_DiffReq.Function->parameters();
+  // If parameter was not marked as varied, add it's VarData and mark
+  // non-varied.
+  for (const auto& par : paramsRef) {
+    if (m_DiffReq.getVariedDecls().find(par) ==
+        m_DiffReq.getVariedDecls().end()) {
+      addVar(par, /*forceInit=*/true);
+      setIsRequired(getVarDataFromDecl(par), /*isReq=*/false);
+    }
+  }
+
   // Add the entry block to the queue.
   m_CFGQueue.insert(m_CurBlockID);
 
@@ -21,20 +50,9 @@ void VariedAnalyzer::Analyze(const FunctionDecl* FD) {
     auto IDIter = std::prev(m_CFGQueue.end());
     m_CurBlockID = *IDIter;
     m_CFGQueue.erase(IDIter);
-    CFGBlock& nextBlock = *getCFGBlockByID(m_CurBlockID);
+    CFGBlock& nextBlock = *getCFGBlockByID(m_AnalysisDC, m_CurBlockID);
     AnalyzeCFGBlock(nextBlock);
   }
-}
-
-void mergeVarsData(std::set<const clang::VarDecl*>* targetData,
-                   std::set<const clang::VarDecl*>* mergeData) {
-  for (const clang::VarDecl* i : *mergeData)
-    targetData->insert(i);
-  *mergeData = *targetData;
-}
-
-CFGBlock* VariedAnalyzer::getCFGBlockByID(unsigned ID) {
-  return *(m_AnalysisDC->getCFG()->begin() + ID);
 }
 
 void VariedAnalyzer::AnalyzeCFGBlock(const CFGBlock& block) {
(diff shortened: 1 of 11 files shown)


test/Analyses/ActivityReverse.cpp
```diff
@@ -458,6 +454,101 @@ double f13(double x, const double* obs){
...
+  b = x2;
+    while(i){
+      x4 = 1;
+      x3 = a;
+      x2 = x1;
+      x1 = x;
+    }
+  }
+  return x1;
+}
```
PetroZarytskyi, line 474: The generated code is correct, but the loops are never executed because ``i`` is zero. I think it makes it easy for somebody in the future to break the test and not realize it because the numerical check passes. Can we adjust the test a little so that the loop is executed, and the analysis affects the numerical results?

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -21,43 +27,57 @@ namespace clad {
...
-  [[nodiscard]] const VarsData& getCurBlockVarsData() const {
-    return const_cast<VariedAnalyzer*>(this)->getCurBlockVarsData();
+  // void copyVarToCurBlock(const clang::VarDecl* VD);
+  void markExpr(const clang::Stmt* S) { m_ResSet.insert(S); }
+  /// Checks if data of blocks blocks with a given order remained same after a
+  /// loop is traversed.
+  bool areBlocksFixed(int order) {
+    if (m_LoopQueueMod.empty())
+      return false;
+    for (auto& i : m_LoopQueueMod)
```
PetroZarytskyi: @vgvassilev Should we write the type explicitly as ``std::pair<unsigned int, LoopInfo>&``?

Once we switch to C++17, we should probably use ``auto& [key, value]``.

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -21,43 +27,57 @@ namespace clad {
...
-  bool isVaried(const clang::VarDecl* VD) const;
+  // bool isVaried(const clang::VarDecl* VD);
   /// Adds varied variable to current branch.
   /// \param[in] VD - Variable declaration.
-  void copyVarToCurBlock(const clang::VarDecl* VD);
-  VarsData& getCurBlockVarsData() { return *m_BlockData[m_CurBlockID]; }
-  [[nodiscard]] const VarsData& getCurBlockVarsData() const {
-    return const_cast<VariedAnalyzer*>(this)->getCurBlockVarsData();
+  // void copyVarToCurBlock(const clang::VarDecl* VD);
+  void markExpr(const clang::Stmt* S) { m_ResSet.insert(S); }
```
PetroZarytskyi, line 38: This is great. We should also move to this system in TBR. I'll open an issue after this PR is merged.

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -21,43 +27,57 @@ namespace clad {
...
 
-  clang::AnalysisDeclContext* m_AnalysisDC;
-  std::vector<std::unique_ptr<VarsData>> m_BlockData;
-  unsigned m_CurBlockID{};
-  std::set<unsigned> m_CFGQueue;
   /// Checks if a variable is on the current branch.
   /// \param[in] VD - Variable declaration.
   /// @return Whether a variable is on the current branch.
-  bool isVaried(const clang::VarDecl* VD) const;
+  // bool isVaried(const clang::VarDecl* VD);
```
PetroZarytskyi: I think we can remove this line (as well as the comment above).

lib/Differentiator/ActivityAnalyzer.h
```diff
@@ -21,43 +27,57 @@ namespace clad {
...
   /// Adds varied variable to current branch.
   /// \param[in] VD - Variable declaration.
-  void copyVarToCurBlock(const clang::VarDecl* VD);
-  VarsData& getCurBlockVarsData() { return *m_BlockData[m_CurBlockID]; }
-  [[nodiscard]] const VarsData& getCurBlockVarsData() const {
-    return const_cast<VariedAnalyzer*>(this)->getCurBlockVarsData();
+  // void copyVarToCurBlock(const clang::VarDecl* VD);
```
PetroZarytskyi: We can also remove this with the comment.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -136,57 +209,100 @@ bool VariedAnalyzer::VisitCallExpr(CallExpr* CE) {
...
+        hasVariedArg = true;
+        m_DiffReq.addVariedDecl(FDparam[i]);
+      }
 
       m_Marking = false;
       m_Varied = false;
     }
   }
+  // is that so?
+  m_Varied = hasVariedArg;
```
PetroZarytskyi, line 188: I think we cannot do better than this with small changes. If you think the comment should stay, let's change it to a fixme with a brief explanation.

lib/Differentiator/ActivityAnalyzer.h
vgvassilev: We could write the type but probably better to typedef the pair into something more meaningful first.

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -136,57 +174,87 @@ bool VariedAnalyzer::VisitCallExpr(CallExpr* CE) {
...
+  return false;
 }
 
-bool VariedAnalyzer::VisitDeclRefExpr(DeclRefExpr* DRE) {
+bool VariedAnalyzer::TraverseDeclRefExpr(DeclRefExpr* DRE) {
   auto* VD = dyn_cast<VarDecl>(DRE->getDecl());
-  if (!VD)
-    return true;
+  // if (!VD)
+  //   return false;
```
PetroZarytskyi:
```suggestion
  assert(VD && "...")
```

lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -136,57 +174,87 @@ bool VariedAnalyzer::VisitCallExpr(CallExpr* CE) {
...
+    setVaried(DRE);
+    m_DiffReq.addVariedDecl(VD);
+    markExpr(DRE);
+  } else if (m_Marking)
+    setVaried(DRE, false);
 
-  if (isVaried(VD))
-    m_Varied = true;
+  if (VarData* data = getVarDataFromDecl(VD))
+    m_Varied = findReq(*data);
```
PetroZarytskyi: Shouldn't it be ``m_Varied |= findReq(*data);`` so that we don't overwrite possibly varied previous traversed stmts?

lib/Differentiator/ActivityAnalyzer.cpp
ovdiiuv: As far as I know not every `DeclRefExpr` correspond to a `VarDecl`.

lib/Differentiator/AnalysisBase.cpp
```diff
@@ -227,10 +243,12 @@ bool AnalysisBase::findReq(const Expr* E) {
...
 }
 
-void AnalysisBase::merge(VarData& targetData, VarData& mergeData) {
+bool AnalysisBase::merge(VarData& targetData, VarData& mergeData) {
   if (targetData.m_Type == VarData::FUND_TYPE) {
+    bool bam = targetData.m_Val.m_FundData;
     targetData.m_Val.m_FundData =
         targetData.m_Val.m_FundData || mergeData.m_Val.m_FundData;
+    return bam != (targetData.m_Val.m_FundData || mergeData.m_Val.m_FundData);
   } else if (targetData.m_Type == VarData::OBJ_TYPE) {
```
PetroZarytskyi, line 252: ```bam != (targetData.m_Val.m_FundData || mergeData.m_Val.m_FundData)```
is the same as (if we plug in the assignment on line 249)
```bam != (bam || mergeData.m_Val.m_FundData || mergeData.m_Val.m_FundData)```
which is the same as
```bam != (bam || mergeData.m_Val.m_FundData)```
the expr above is true if and only if ``bam==false`` and ``mergeData.m_Val.m_FundData==true``, so this is the same as
```!bam && mergeData.m_Val.m_FundData```

All this is to say that we can replace this with a line
```bool targetModified = !targetData.m_Val.m_FundData && mergeData.m_Val.m_FundData``` on line 248 and return it.

PetroZarytskyi (approved):
LGTM!

PR #1510, opened 2025-08-27
Refactor nested TBR

### LINKED ISSUE #1497: Restructure TBR scheduling in DiffPlanner
After PR #1490, TBR analysis of some functions is scheduled in advance in the DiffPlanner. Some solutions, like ``m_TBROnly`` and ``RequestTBR``, worsen readability. We need to refactor this code while maintaining the positive impact of nested TBR.

PetroZarytskyi:
After #1490, there have been concerns about how nested TBR was implemented. This PR moves the information about modified parameters from the ``ClangPlugin`` to ``DiffRequest`` (each one only has the necessary information), removes the state variable ``m_TBROnly`` in the ``DiffPlanner``, and removes the field ``RequestTBR`` from ``DiffRequest``.

Fixes #1497.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -642,7 +642,8 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
     if (!m_TbrRunInfo.HasAnalysisRun && !isLambdaCallOperator(Function) &&
         Function->isDefined() && m_AnalysisDC) {
       TimedAnalysisRegion R("TBR " + BaseFunctionName);
-      TBRAnalyzer analyzer(m_AnalysisDC, getToBeRecorded());
+      TBRAnalyzer analyzer(m_AnalysisDC, getToBeRecorded(),
+                           &getModifiedParams());
       analyzer.Analyze(*this);
     }
     auto found = m_TbrRunInfo.ToBeRecorded.find(S->getBeginLoc());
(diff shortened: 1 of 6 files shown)


PetroZarytskyi:
@vgvassilev I think clang-format is broken here. It did the same for me locally, and the resulting code doesn't even compile.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1295,10 +1297,16 @@ DeclRefExpr* getArgFunction(CallExpr* call, Sema& SemaRef) {
 
       // Recurse into call graph.
       TraverseFunctionDeclOnce(request.Function);
-    }
 
-    if (m_TBROnly)
-      request.Mode = DiffMode::unknown;
+      if (requestTBR) {
+        TimedAnalysisRegion R("TBR " + request.BaseFunctionName);
+        auto& modifiedParams = request.getModifiedParams();
```
vgvassilev: We should probably spell the type here. If it is too long we can do a `using ...` in the DiffRequest..

PetroZarytskyi: Done. Do you have preferences about the naming? You can take a look at the last commit.

vgvassilev (approved):
Lgtm!

lib/Differentiator/DiffPlanner.cpp
vgvassilev: Should be fine. Feel free to land it when the ci passes.

PR #1511, opened 2025-08-28
Add custom derivative for thrust::copy

a-elrawy:
This pull request adds a new custom derivative for the thrust::copy function.

### DIFF
--- a/test/CUDA/ThrustCopy.cu
+++ b/test/CUDA/ThrustCopy.cu
@@ -34,6 +34,25 @@ void copy(const thrust::device_vector<double>& src,
 // CHECK-NEXT:     }
 // CHECK-NEXT: }
 
+void copy_with_return_value_stored(const thrust::device_vector<double>& src,
+                                   thrust::device_vector<double>& dst) {
+    auto it = thrust::copy(src.begin(), src.end(), dst.begin());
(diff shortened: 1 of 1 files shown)


test/CUDA/ThrustCopy.cu
```diff
@@ -0,0 +1,61 @@
...
+#include "clad/Differentiator/ThrustDerivatives.h"
+#include "../TestUtils.h"
+
+#include <thrust/device_vector.h>
+#include <thrust/host_vector.h>
+#include <thrust/copy.h>
+
+void copy(const thrust::device_vector<double>& src,
+                 thrust::device_vector<double>& dst) {
+    thrust::copy(src.begin(), src.end(), dst.begin());
```
vgvassilev, line 25: Could you please add a test that uses the return value of `thrust::copy`.

a-elrawy: The return value is the end iterator of the result vector. I've added a test that have it declared instead of the `{}`.

a-elrawy: Also, I had to cast it to void, because `warning: clad::gradient only supports differentiation functions of real return types. Return stmt ignored.`

PR #1513, opened 2025-08-29
Support not differentiating w.r.t. const references in reverse mode

### LINKED ISSUE #1411: Not differentiating w.r.t. const references produces an error
When we differentiate
```
double f (const double& x, const double& y) {
   return x + y;
}
```
w.r.t. both parameters (``clad::gradient(f);``), everything works as expected. But when we attempt to differentiate only w.r.t. the first param (``clad::gradient(f, "x");``), we get an error:
```
error: non-const lvalue reference to type 'double' cannot bind to an initializer list temporary
   51 | double f (const double& x, const double& y) {
      |        ^
void f_grad_0(const double &x, const double &y, double *_d_x) {
    double &_d_y = <recovery-expr>({});
    {
        *_d_x += 1;
        _d_y += 1;
    }
}
```
It happens because Clad still generates ``_d_y`` of type ``double&`` (`const` is removed) and attempts to zero-initialize it.
const references are often used for efficiency instead of corresponding non-reference types and should in theory be handled similarly.

Note:
In the reverse mode, if `x` affects `y`, then `_d_y` affects `_d_x`, e.g. `x = y;` --> `_d_y += _d_x;`. Therefore, if a parameter `p` is constant (cannot be affected by anything inside the function), it means nothing is affected by `_d_p`. So if we the user is not interested in `_d_p` of a constant `p`, then it's guaranteed to be useless. We need to figure out whether we want to remove adjoints in such cases as it might lead to new partial pullbacks, which already caused problems as described in #1371.

PetroZarytskyi:
This PR fixes issue #1411. If the user doesn't request a derivative w.r.t. a const reference, that derivative will, in theory, not be used. However, in this PR, we still generate it to avoid partial pullbacks.

Fixes #1411.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -412,6 +412,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
           continue;
         }
         auto VDDerivedType = utils::getNonConstType(paramTy, m_Sema);
+        VDDerivedType = VDDerivedType.getNonReferenceType();
         auto* VDDerived =
             BuildGlobalVarDecl(VDDerivedType, "_d_" + param->getNameAsString(),
                                getZeroInit(VDDerivedType));
--- a/test/Gradient/DiffInterface.C
+++ b/test/Gradient/DiffInterface.C
@@ -73,6 +73,18 @@ double f_1(double x, double y, double z) {
 //CHECK-NEXT:       }
 //CHECK-NEXT:   }
 
+double f_2 (const double& x, const double& y) {
+   return x + y;
+}
+
+//CHECK:   void f_2_grad_0(const double &x, const double &y, double *_d_x) {
+//CHECK-NEXT:       double _d_y = 0.;
+//CHECK-NEXT:       {
+//CHECK-NEXT:           *_d_x += 1;
+//CHECK-NEXT:           _d_y += 1;
+//CHECK-NEXT:       }
+//CHECK-NEXT:   }
+
 #define TEST(F, ...)                                                           \
   {                                                                            \
     result[0] = 0;                                                             \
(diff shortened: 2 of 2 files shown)


vgvassilev (approved):
LGTM!

Rohan-T144:
I just tested this PR with my test file, and the full LLM/cladtorch:

https://github.com/vgvassilev/clad/blob/7f69656fac5329eef4923e8058eb95e6e443469e/test/Gradient/Cladtorch.C#L51-L58

It fixes it so that clad doesn't error when trying to compile this anymore.

However the generated code does not copy-initialize from the argument. I.e., it does `cladtorch::Tensor _d_u = {};` instead of `_d_u(u)`. 

It also doesn't fix the main issue I had where the fully implemented `cladtorch::Tensor` type is not copy-initialized for intermediate tensors - i.e. it still does

```
cladtorch::FTensor x = /* some expr */;
cladtorch::FTensor _d_x = {};
// instead of cladtorch::FTensor _d_x(x); clad::zero_init(_d_x);
```
And so it crashes when we try to run the training loop

I'm not completely sure why that happens for the actual Tensor class when clad does copy-initialize for a simpler class, but I think actualy issue is that it hits some edge-case in the code that decides for copy-initialization, which is why I made my PR.

Note that this doesn't block the PR, it just doesn't fix my use-case yet.

vgvassilev:
> I just tested this PR with my test file, and the full LLM/cladtorch:
> 
> https://github.com/vgvassilev/clad/blob/7f69656fac5329eef4923e8058eb95e6e443469e/test/Gradient/Cladtorch.C#L51-L58
> 
> It fixes it so that clad doesn't error when trying to compile this anymore.
> 
> However the generated code does not copy-initialize from the argument. I.e., it does `cladtorch::Tensor _d_u = {};` instead of `_d_u(u)`.
> 
> It also doesn't fix the main issue I had where the fully implemented `cladtorch::Tensor` type is not copy-initialized for intermediate tensors - i.e. it still does
> 
> ```
> cladtorch::FTensor x = /* some expr */;
> cladtorch::FTensor _d_x = {};
> // instead of cladtorch::FTensor _d_x(x); clad::zero_init(_d_x);
> ```
> 
> And so it crashes when we try to run the training loop
> 
> I'm not completely sure why that happens for the actual Tensor class when clad does copy-initialize for a simpler class, but I think actualy issue is that it hits some edge-case in the code that decides for copy-initialization, which is why I made my PR.
> 
> Note that this doesn't block the PR, it just doesn't fix my use-case yet.

@PetroZarytskyi, do we somehow find a default constructor in the class and use it because we cannot find a copy constructor or we prefer the default over the copy?

It's a bit annoying that we will have to copy the object in such a case instead of synthesizing a const reference -- we should know that we don't change the state of the object anyway...

PR #1515, opened 2025-08-30
Find used parameter values in nested TBR 

PetroZarytskyi:
After #1490, TBR can detect which parameters have been modified. This PR extends the analysis of the same class of functions to also track whether the function uses the given parameter value, which allows for more optimization in TBR. In particular, now we can avoid storing some arguments even if they are modified.

This PR is a less powerful version of #1506 that works consistently across all platforms. To get the benefits from STL, we need to improve the system of nested TBR to analyze which fields of objects are modified/used. However, even in its current form, this PR is sufficient to unblock #1445. 

During the work on this PR, there occured a bug, the reason for which was that return stmts were not properly visited in TBR. Fixing that caused regressions in some tests, and the goal of the second commit is to deal with those regressions.
The core idea there is that we don't have to store operands if a whole expr is already stored. In particular, we often store the LHS/RHS of multiplication/division.


vgvassilev (approved):
LGTM!

PR #1516, opened 2025-08-30
Enable TBR in pointer arithmetics

PetroZarytskyi:
Although we have improved support for pointers in TBR and enabled TBR in all pointer-related tests, we still don't use it in pointer operations. This PR enables TBR in pointer arithmetics.


vgvassilev (approved):
Lgtm!

PR #1518, opened 2025-09-02
Rework finding of the target function to differentiate.

vgvassilev:
This patch teaches clad to find the target function that was initially meant including going through the inheritance chain.


test/Functors/Simple.C
```diff
@@ -37,8 +37,6 @@ public:
 // CHECK-NEXT: float _d_y = 0;
 // CHECK-NEXT: SimpleExpression _d_this_obj;
 // CHECK-NEXT: SimpleExpression *_d_this = &_d_this_obj;
-// CHECK-NEXT: float _d_x0 = 0;
```
PetroZarytskyi, line 40: Why did this disappear?

vgvassilev: I have no idea. I believe that we don’t build a fake DeclRefExpr and some other optimization kicks in..

PR #1523, opened 2025-09-06
Schedule first order derivatives for hessians statically

PetroZarytskyi:
Hessians represent second-order multiple-arg derivatives. They utilize two layers of derivatives: forward and then reverse. In this PR, forward mode-derivatives are now scheduled statically. 
Diagonal Hessians utilize double-forward mode differentiation. This PR makes diagonal Hessians fully statically-scheduled. To do this, it was necessary to add support for higher-order derivatives in DerivedFnCollector::Find.


vgvassilev (approved):
Lgtm!

PR #1524, opened 2025-09-07
Enable static custom derivatives lookups for topmost diffrequests

PetroZarytskyi:
Currently, we perform most lookups statically in ``DiffPlanner::LookupCustomDerivativeDecl``. However, we have a separate function ``DerivativeBuilder::LookupCustomDerivativeDecl`` that's used exclusively for topmost requests (i.e. ``_grad``/``_darg0`` and not ``pushforward``/``pullback`` etc.) performed in visitors. This PR fully moves this obligation to ``DiffPlanner::LookupCustomDerivativeDecl``. Also, the old system only supported custom derivatives represented by ``FunctionDecl`` with matching types. Now, the set of supported features is the same as for other modes: templates, using shadow decls, etc.


PetroZarytskyi:
@vgvassilev While working on this PR, I encountered an inconsistency in the way custom derivatives are dumped. Before the PR, nested derivative types (``_pushforward/_pullbacl/...``) were never dumped, and the topmost ones (``_grad/_darg0/...``) were always dumped. In this PR, the custom derivative is dumped if ``request.CallUpdateRequired`` is ``true``. I could also iterate all topmost modes, but that seemed a little ugly. Because of this, some custom derivatives are now not displayed.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1042,35 +1042,38 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
...
+        request.Mode == DiffMode::pullback ||
+        request.Mode == DiffMode::vector_pushforward) {
+      if (const auto* ME = dyn_cast<CXXMemberCallExpr>(callSite)) {
+        fnDecl = ME->getMethodDecl();
+      } else if (const auto* CE = dyn_cast<CallExpr>(callSite)) {
+        const Expr* Callee = CE->getCallee()->IgnoreParenCasts();
+        if (const auto* DRE = dyn_cast<DeclRefExpr>(Callee))
+          fnDecl = DRE->getFoundDecl();
+        else if (const auto* MemberE = dyn_cast<MemberExpr>(Callee))
+          fnDecl = MemberE->getFoundDecl().getDecl();
```
vgvassilev, line 1057: We should figure out how to cover these lines.

PetroZarytskyi: I think the only cases of non-member CallExprs with MemberExpr callees are something crazy like
```
struct S {
    void (*fp)(int);
};

double f(double x) {
    S s;
    s.fp(42);   // This is a CallExpr, callee is a MemberExpr referring to fp
    return 0;
}
```
But we're far from that. All of our infrastructure ignores non-direct calls.
I'm afraid that in this PR, we can either remove / comment out this bit or ignore CodeCov.

vgvassilev: I thought that was covered by tests already but it wasn't.

vgvassilev (approved):
LGTM!

PR #1527, opened 2025-09-07
Use Stmt* as identifiers in TBR instead of SourceLocation

PetroZarytskyi:
Currently, we use ``clang::SourceLocation`` to propagate information from ``TBR`` to ``ReverseModeVisitor``. Source locations are unreliable for a couple of reasons:
1) Some stmts don't have it. For example, if the code is compiler- or clad-generated.
2) Some stmts have the same location. For example,
```
a.getX() = 1;
```
Since ``a`` and ``a.getX()`` have the same location, we cannot differentiate between the information about storing ``a`` in the ``CallExpr`` ``a.getX()`` and about storing ``a.getX()`` in the assignment ``a.getX() = 1``. Storing stmts by pointers allows us to consider these as separate entities.


vgvassilev (approved):
LGTM!

PR #1528, opened 2025-09-09
Remove ID str from range-based loops iterators

ovdiiuv:
Before this PR, when differentiating range-based loops we would, for example, create `__begin10` iterators instead of `__begin1`. This was caused by variables starting with `_` being reserved for Clad's internal logic.


vgvassilev (approved):
Lgtm!

PR #1531, opened 2025-09-11
Add tail pointer and capacity variable to tape

aditimjoshi:
Tail pointer added to point to the last slab and prevent iterating through the whole tape for every push and capacity variable added to keep track of the capacity of the tape and prevent unnecessary allocations of slabs.

### DIFF
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -54,7 +54,7 @@ inline CUDA_HOST_DEVICE unsigned int GetLength(const char* code) {
 
 /// Tape type used for storing values in reverse-mode AD inside loops.
 template <typename T, std::size_t SBO_SIZE = 64, std::size_t SLAB_SIZE = 1024>
-using tape = tape_impl<T>;
+using tape = tape_impl<T, SBO_SIZE, SLAB_SIZE>;
 
 /// Add value to the end of the tape, return the same value.
 template <typename T, std::size_t SBO_SIZE = 64, std::size_t SLAB_SIZE = 1024,


include/clad/Differentiator/Tape.h
```diff
@@ -204,7 +203,22 @@ class tape_impl {
...
+      std::size_t offset = (m_size - SBO_SIZE) % SLAB_SIZE;
+      destroy_element(m_tail->elements() + offset);
+      if (offset == 0) {
+        Slab* slab = m_head;
+        Slab* prev = m_head;
+        std::size_t idx = (m_size - SBO_SIZE) / SLAB_SIZE;
+        while (idx--) {
+          prev = slab;
+          slab = slab->next;
+        }
```
parth-07, line 218: Can you please check this code again?

Considering 0-based indexing, if `idx` is `i`th, then the `tail` should point to the `i`th element of the linked list, right? It seems that with the current code, if the `idx` is `1`, then the tail will point to the `0`th element of the linked list, instead of the 1st element.

parth-07 (commented):
I think we should add more intensive tests for the tape, if we do not have such tests already. Most tests that we have do not require large number of push/pop in the tape, and thus we are only testing tape when there are small number of push/pop. So far, it was fine because tape implementation was relatively simple.

include/clad/Differentiator/Tape.h
aditimjoshi: It's not 0-based indexing in this case. For idx=0 it goes into the sbo and it should point to m_head and then for idx=1 it means its moving from the 2nd to the 1st slab so it would point to m_head and if idx=2 it means its moving from 3rd to 2nd slab so in first iteration its m_head and then in the 2nd iteration its the 2nd slab and so on.

parth-07: Oh, I see. Thanks for the explanation.

PR #1536, opened 2025-09-13
Fix range-based loops and add early traverse for regular loops in VA

ovdiiuv:
This PR fixes two issues in loop analysis that led to incorrect behavior. First one occured because we used to merge all data into the condition block before traversing a successor block, which could cause problems like entering an infinite loop in cases like:
```cpp
while(b = 1)
  b = x
```
or codition not being the last block to be analyzed. Second thing is that the analysis made assumptions based on block numeration, which do not hold for the range-based loops where the CFG is inverted.

Also, we now run numerical check for Gradient/Loops.C with VA on.


vgvassilev (commented):
This looks reasonable to me. Can somebody clean up the clang-tidy complaints?

vgvassilev (approved):
Lgtm!

PR #1541, opened 2025-09-16
Move argument differentiation into a separate function

PetroZarytskyi:
In the reverse mode, we have quite sophisticated rules about how to differentiate call arguments. This logic should be the same for ``CallExpr`` and ``CXXConstructExpr`` but we repeat it in these functions independently. This leads to frequent confusion, as some features are supported in one but not in the other. Apart from making it harder to add support for new features, it also makes it a lot harder to track them properly. This PR moves function/constructor argument differentiation to a separate function ``DifferentiateCallArg``. It now returns a triplet: ``{<original arg>, <arg for pullback>, <reverse_forw arg>}``. In practice, it will look somewhat like this: ``{x, &_r0, _d_x}``.

Apart from that, it allowed us to move some other argument processing to ``DifferentiateCallArg``. In particular, now ``pullback`` and ``reverse_forw`` arguments are generated inside the function. Before, we used to have more lists of arguments for different stages of processing: ``CallArgs``, ``CallArgDx``, ``DerivedCallArgs``, ``DerivedOutputArgs``, ``pullbackArgs`` (which still doesn't include a list of ``reverse_forw`` args).

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1502,6 +1502,191 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     return StmtDiff(Clone(FL));
   }
 
+  static bool isNAT(QualType T) {
+    T = utils::GetValueType(T);
+    if (const auto* RT = T->getAs<RecordType>()) {
+      const RecordDecl* RD = RT->getDecl();
+      if (RD->getNameAsString() == "__nat")
+        return true;
+    }
+    return false;
+  }
+
+  StmtDiff ReverseModeVisitor::DifferentiateCallArg(
+      const Expr* arg, const ParmVarDecl* param,
+      llvm::SmallVectorImpl<clang::Stmt*>& PreCallStmts, bool isNonDiff,
+      bool isCUDAKernel) {
+    StmtDiff result;
+    StmtDiff argDiff{};
+    // FIXME: We handle parameters with default values by setting them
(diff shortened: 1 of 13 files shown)


PetroZarytskyi:
@vgvassilev I made so many commits just in case for debugging (they are technically independent). In practice, I would squash pretty much all of them

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1502,6 +1502,191 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     return StmtDiff(Clone(FL));
   }
 
+  static bool isNAT(QualType T) {
+    T = utils::GetValueType(T);
+    if (const auto* RT = T->getAs<RecordType>()) {
+      const RecordDecl* RD = RT->getDecl();
+      if (RD->getNameAsString() == "__nat")
+        return true;
```
vgvassilev, line 1510: Do we have any chance to test this?

PetroZarytskyi: This is covered on some platforms, but those are ignored by codecov. The whole commit "fix arm nuc" is about making that system work properly.

PetroZarytskyi: This causes 3 out of 5 "uncovered" lines. The other 2 are some kind of diagnostics in CUDA that we don't test.

vgvassilev: Ok.

vgvassilev (approved):
LGTM!

PR #1545, opened 2025-09-17
Do not generate empty pullbacks

PetroZarytskyi:
Currently, a lot of our pullback calls are done to empty functions. Ideally, we should treat them as non-differentiable, but it is currently impossible to get information before visiting call arguments. In this PR, call exprs to empty function pullbacks are no longer built. Empty constructor pullbacks are never registered and so never used. It wasn't possible to do the same with regular functions because this triggers diagnostics about the function not being differentiated.

### DIFF
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -132,17 +132,11 @@ namespace clad {
 
     bool hasEmptyBody(const clang::FunctionDecl* FD) {
       FD = FD->getCanonicalDecl();
-      Stmt* body = FD->getBody();
       if (const FunctionDecl* TIP = FD->getTemplateInstantiationPattern())
-        body = TIP->getBody();
-      while (auto CS = dyn_cast_or_null<CompoundStmt>(body)) {
-        if (CS->size() == 0)
-          return true;
-        if (CS->size() > 1)
-          return false;
-        body = *CS->body_begin();
-      }
-      return false;
+        FD = TIP;
+      if (!FD->hasBody())
+        return false;
+      return utils::unwrapIfSingleStmt(FD->getBody()) == nullptr;
     }
 
     CompoundStmt* PrependAndCreateCompoundStmt(ASTContext& C, Stmt* initial,
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -130,6 +130,21 @@ namespace clad {
       return CS;
     }
 
+    bool hasEmptyBody(const clang::FunctionDecl* FD) {
+      FD = FD->getCanonicalDecl();
+      Stmt* body = FD->getBody();
+      if (const FunctionDecl* TIP = FD->getTemplateInstantiationPattern())
+        body = TIP->getBody();
+      while (auto CS = dyn_cast_or_null<CompoundStmt>(body)) {
+        if (CS->size() == 0)
+          return true;
+        if (CS->size() > 1)
+          return false;
+        body = *CS->body_begin();
+      }
+      return false;
+    }
+
     CompoundStmt* PrependAndCreateCompoundStmt(ASTContext& C, Stmt* initial,
                                                Stmt* S) {
       llvm::SmallVector<Stmt*, 16> block;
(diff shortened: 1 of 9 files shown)


lib/Differentiator/CladUtils.cpp
```diff
@@ -130,6 +130,21 @@ namespace clad {
...
+      FD = FD->getCanonicalDecl();
+      Stmt* body = FD->getBody();
+      if (const FunctionDecl* TIP = FD->getTemplateInstantiationPattern())
+        body = TIP->getBody();
+      while (auto CS = dyn_cast_or_null<CompoundStmt>(body)) {
+        if (CS->size() == 0)
+          return true;
+        if (CS->size() > 1)
+          return false;
+        body = *CS->body_begin();
```
vgvassilev: Is it realistic to expect functions of the sort:
```cpp
void f(double x) {
  {
    // nothing
  }
}
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2097,33 +2099,30 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+        // If the derivative is called through _darg0 instead of _grad.
+        Expr* d = BuildOp(BO_Mul, dfdx(), OverloadedDerivedFn);
+        auto* UnOp = cast<UnaryOperator>(CallArgDx[0]);
+        Expr* addGrad = BuildOp(BO_AddAssign, Clone(UnOp->getSubExpr()), d);
+        it = block.insert(it, addGrad);
+        it++;
+      } else {
+        // Insert the CallExpr to the derived function
+        it = block.insert(it, OverloadedDerivedFn);
+        it++;
```
vgvassilev: Shouldn't that go in the `OverloadedDerivedFn` and `!asGrad` to become an else clause of it?

PetroZarytskyi: If I understood correctly, not really. ``!asGrad`` represents if we use forward mode (_darg or pushforward). Then, we do ``_r0 += dfdx() * f_pushforward(x, 1.).pushforward`` (we could actually do it a little simpler, but that's a separate topic). ``OverloadedDerivedFn`` can, in theory, represent both a call to a pullback and a call to a pushforward. Anyway, I refactored the code a bit.

lib/Differentiator/CladUtils.cpp
PetroZarytskyi: We do have that pattern sometimes. In this PR, it was useful to remove empty constructor pullbacks. However, I think you're right that we should focus on not generating such functions rather than analyzing them. I fixed the thing that caused these nested CompoundStmts and simplified the function.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: I also think that we shouldn't do these insertions, but that's also a different topic.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2075,21 +2077,23 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     // calls.
     llvm::SmallVector<Stmt*, 16> PostCallStmts{};
     if (pullbackFD) {
```
vgvassilev, line 2079:
```suggestion
    if (pullbackFD && !utils::hasEmptyBody(pullbackFD)) {
```

Can this become something like that?

PetroZarytskyi: Then, it will attempt to use numerical diff (because of the else-clause)

PetroZarytskyi: If not the numerical diff, I would move this logic to the diff planner and not register empty derivatives at all. That's what I did with constructors. But with normal functions, it doesn't work because of the diagnostics/num diff. In the future, I think we should move all the decision-making to the diffplanner, but I don't think it's possible until we switch to static-only scheduling.

vgvassilev: Yes, understood. Maybe we should track the switch to static scheduling of numerical diff into a separate issue.

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: Can we open an issue about that, too?

PR #1548, opened 2025-09-21
[ci] Move the valgrind checks to the latest supported llvm.

vgvassilev:


guitargeek:
Wants runtime 20 -> 21. Master is on 22. Superseded twice over.

PR #1552, opened 2025-09-22
Move MacOS x86 runners to MacOS 15

mcbarton:
The MacOS 13 x86 runners have just been deprecated (see https://github.com/actions/runner-images/issues/13046) and a MacOS 15 x86 runner has been created as a replacement (see https://github.com/actions/runner-images/issues/13045). This will be the final MacOS x86 Github runner.


vgvassilev (approved):
LGTM!

PR #1553, opened 2025-09-22
Add some testing infrastructure for STL starting with <cmath>

vgvassilev:


test/Hessian/BuiltinDerivatives.C
```diff
@@ -402,15 +402,20 @@ int main() {
...
+// CHECK-NEXT:        _r1 += _d_y.value * clad::custom_derivatives::std::log_pushforward(x, 1.F).pushforward;
+// CHECK-NEXT:        *_d_x += _r1;
+// CHECK-NEXT:        _d_dlog += _d_y.pushforward * d_x;
+// CHECK-NEXT:        *_d_d_x += dlog * _d_y.pushforward;
+// CHECK-NEXT:    }
+// CHECK-NEXT:    {
+// CHECK-NEXT:        double _r0 = _d_dlog * -(1. / (x * x));
+// CHECK-NEXT:        *_d_x += _r0;
+// CHECK-NEXT:    }
+// CHECK-NEXT:}
```
vgvassilev, line 402: @PetroZarytskyi, can you check if that's the code that you expect to be here after I added temporaries to the log custom derivatives?

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -1402,6 +1399,102 @@ CUDA_HOST_DEVICE void hypot_pullback(T x, T y, U d_z, T* d_x, T* d_y) {
...
+  t = powl_pushforward(x, exponent, /*d_x=*/0.L, /*d_exponent*/ 1.L);
+  *d_exponent += t.pushforward * d_y;
+}
+
+CUDA_HOST_DEVICE inline ValueAndPushforward<float, float>
+sqrtf_pushforward(float x, float d_x) noexcept {
+  return {sqrtf(x), (1.F / (2.F * sqrtf(x))) * d_x};
+}
+
+CUDA_HOST_DEVICE inline void sqrtf_pullback(float a, float d_y,
```
PetroZarytskyi: If a function has only one numerical (non-reference) parameter, we differentiate it in forward mode even inside the reverse mode. In other words, I don't think it's possible to make Clad use ``sqrtf_pullback``.

@vgvassilev, did you have any particular use for this in mind?

And just in case, same goes for  ``sqrtl_pullback``. ``cbrtf_pullback``, ``cbrtl_pullback``.

test/Hessian/BuiltinDerivatives.C
PetroZarytskyi: Yes, I believe the code looks reasonable.

test/Features/stl-cmath.cpp
```diff
@@ -0,0 +1,357 @@
...
+  // Hyperbolic
+  CHECK_ALL(sinh);
+  CHECK_ALL(cosh);
+  CHECK_ALL(tanh);
+  CHECK_ALL(asinh);
+
+  // x in [1,inf)
+  CHECK_RANGE(acosh, {1.0, 1.1, 1.5, 2.0, 3.0, 5.0, 10.0});
+  //CHECK_RANGE_F(acoshf, {1.01, 1.1, 1.5, 2.0, 3.0, 5.0, 10.0});
+  //CHECK_RANGE_LD(acoshl, {1.01, 1.1, 1.5, 2.0, 3.0, 5.0, 10.0});
```
PetroZarytskyi, line 350: Why are some types commented out?

vgvassilev: We don't support them yet. I am trying to capture the current state and then I can move forward with adding that support.

include/clad/Differentiator/BuiltinDerivatives.h
vgvassilev: Good point. I am working on a diagnostic first.

vgvassilev: That should be fixed now.

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -212,14 +212,6 @@ CUDA_HOST_DEVICE ValueAndPushforward<T, dT> abs_pushforward(T x, dT d_x) {
...
 }
 
-template <typename T, typename U>
-CUDA_HOST_DEVICE void abs_pullback(T x, U d_y, T* d_x) {
-  if (x >= 0)
-    *d_x += d_y;
-  else
-    *d_x -= d_y;
-}
-
```
PetroZarytskyi, line 222: Why wasn't this diagnosed in the other PR though?

vgvassilev: Because we had no use of it?

Issue #1557, opened 2025-09-23
Regression with `std::numeric_limits` and Apple Clang


guitargeek:
There are compilation problems of pullbacks with Apple Clang, as can be seen in the ROOT CI runs with Clad `master`:
https://github.com/root-project/root/actions/runs/17949296265/job/51043826899?pr=19961

I guess the fix should be to change something private to public?

@PetroZarytskyi 

```txt
2025-09-23T14:53:18.9756140Z error: 'infinity_pullback' is a private member of 'std::__libcpp_numeric_limits<double, true>'
2025-09-23T14:53:18.9757320Z /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1/limits:462:45: note: constrained by private inheritance here
2025-09-23T14:53:18.9758430Z class _LIBCPP_TEMPLATE_VIS numeric_limits : private __libcpp_numeric_limits<_Tp> {
2025-09-23T14:53:18.9759040Z                                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2025-09-23T14:53:18.9759450Z note: member is declared here
2025-09-23T14:53:18.9760020Z error: 'min_pullback' is a private member of 'std::__libcpp_numeric_limits<double, true>'
2025-09-23T14:53:18.9761130Z /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1/limits:462:45: note: constrained by private inheritance here
2025-09-23T14:53:18.9762220Z class _LIBCPP_TEMPLATE_VIS numeric_limits : private __libcpp_numeric_limits<_Tp> {
2025-09-23T14:53:18.9762820Z                                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2025-09-23T14:53:18.9763220Z note: member is declared here
2025-09-23T14:53:18.9763790Z error: 'max_pullback' is a private member of 'std::__libcpp_numeric_limits<unsigned int, true>'
2025-09-23T14:53:18.9764910Z /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1/limits:462:45: note: constrained by private inheritance here
2025-09-23T14:53:18.9766330Z class _LIBCPP_TEMPLATE_VIS numeric_limits : private __libcpp_numeric_limits<_Tp> {
[...]

PR #1561, opened 2025-09-24
Disable analyses for dynamically scheduled derivatives

### LINKED ISSUE #1559: Regression in gradient when using constant array parameters with pointer offset in inner function
@PetroZarytskyi 

Reproducer as a ROOT macro:
```c++

inline double flexibleInterp(double const *params, unsigned int n, double const *low, double const *high)
{
   double total = 1.;
      total += params[i] * (high[i] - 1.);
   }

   return total;
}

#pragma cling optimize(2)
double roo_codegen_4(double *params, const double *xlArr)
{
   return flexibleInterp(params, 2, xlArr, xlArr + 2);
}

#pragma clad ON
void gradient_request()
{
   clad::gradient(roo_codegen_4, "params");
}
#pragma clad OFF
// clang-format off

    0.1, 0.1, 1.
};

std::vector<double> auxConstantsVec = {
    0.95, 0.99, 
    1.05, 1.01
};

// clang-format on

// To run as a ROOT macro
void repro()
{

   auto func = [&](std::span<double> params) { return roo_codegen_4(params.data(), auxConstantsVec.data()); };
      return roo_codegen_4_grad_0(parametersVec.data(), auxConstantsVec.data(), out.data());
   };

   auto numDiff = [&](int i) {
      const double eps = 1e-6;
      double funcValDown = func(p);
      double funcValUp = func(p);
   };

   }
}
```

Output is:
```txt
0:
  numr : 0.05
  clad : 0
1:
  numr : 0.01
  clad : 0
2:
  numr : 0
  clad : -2
```

PetroZarytskyi:
After #1456, TBR (as well as other analyses) depend heavily on the DiffPlanner because ``m_AnalysisDC`` is generated there. This is an issue if we're dealing with dynamic derivatives. This has led to issue #1559, where TBR is labeled as enabled but essentially doesn't run (there is no DFC to traverse). Since TBR is technically enabled, but nothing is labeled as to-be-recorded, we don't generate the essential stores.
Of course, disabling analyses for dynamic derivatives is not a perfect solution. We need to decide whether we want to prioritize enabling them or getting rid of dynamic derivatives.

Fixes #1559.

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -391,6 +391,12 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
     FunctionDecl* derivative = this->FindDerivedFunction(request);
     if (!derivative) {
       alreadyDerived = false;
+      // FIXME: Our analyses are closely tied to the DiffPlanner. Dynamic
+      // derivatives don't have m_AnalysisDC. We should either disable
+      // dynamic scheduling or build m_AnalysisDC here.
+      request.EnableTBRAnalysis = false;
+      request.EnableVariedAnalysis = false;
+      request.EnableUsefulAnalysis = false;
 
       {
         // Store and restore the original function and its order.
--- a/test/Gradient/FunctionCalls.C
+++ b/test/Gradient/FunctionCalls.C
@@ -1053,6 +1053,42 @@ double fn29(double *x) {
 // CHECK-NEXT:     }
 // CHECK-NEXT: }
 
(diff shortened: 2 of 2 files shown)


vgvassilev:
I agree with the assessment. We want both and it's not a priority. The dynamically scheduled derivatives are suboptimal and it is a mid-term priority to schedule them statically.

guitargeek:
Tests in ROOT with this PR started:
  * https://github.com/root-project/root/pull/19961

I hope that with this PR, the only regression left will be:
  * https://github.com/vgvassilev/clad/issues/1557

Issue #1562, opened 2025-09-24
Derivative part of StmtDiff is built inconsistently in the reverse mode


PetroZarytskyi:
1) For a ``Stmt`` (and not ``Expr``), we always consider the derivative to be the corresponding stmt in the reverse sweep, for example, if the original stmt looks like
```
for (int i=0; i < 3; ++i) {...} // for forward pass
```
the derivative stmt would look like
```
for (; _t0; --_t0) {...} // for reverse pass
```
2) For Lvalue and/or pointer ``Expr``, the derivative is always a reference to the corresponding derivative Lvalue. For example, the derivative of ``x`` is always ``_d_x`` (the derivative of ``x = y`` is also ``_d_x``). This infrastructure is used for handling references/pointer/objects. Note: derivative increment stmts like ``_d_y += _d_x;`` are never returned as the derivatives of any stmts, they are only dumped directly to the block.

3) For Rvalue ``Expr``, we sometimes return the corresponding type zero, e.g., the derivative of ``1`` is considered to be ``0``. However, sometimes the derivative is returned, e.g., the derivative of ``1.`` is just ``nullptr``.

Then, when we want to initialize something, we don't know whether the underlying object is case 2) or 3). We mostly differentiate between them based on whether the derivative is ``nullptr`` or not, and if the derivative is ``nullptr``, we call ``getZeroInit()``. This is a recurring pattern in Clad. 
But then, even when an Lvalue is used with a cast to Rvalue, we sometimes call ``getZeroInit()`` to get a simpler init. For example, the derivative of ``x`` is ``_d_x``, but if ``x`` is used as an Rvalue, we replace ``_d_x`` with ``0.``, which introduces even more checks.
I believe we should sink ``getZeroInit()`` into the visitation process. In other words, complete 3) to always return the corresponding zero. We can implement the conversions from some Lvalues to zeros (like ``_d_x`` to ``0.``) by returning them as part of ``ImplicitCastExpr`` representing ``LValueToRValue`` casts. Ideally, we should get to a point when we can blindly use ``getExpr_dx()`` for initialization with no analysis.

PR #1563, opened 2025-09-24
Build adjoint init lists during differentiation instead of getZeroInit

### LINKED ISSUE #1560: Aggregate and array types are initialized incorrectly in the reverse mode
In the reverse mode, we always initialize aggregate/array type adjoints with zero init lists generated by ``getZeroInit()``. Such init lists lack context and lead to errors. For example
```
struct Session {
   double x;
   float* ptr;
};

double fn(double x, float y) {
  Session s = {x, &y};
  return s.x + *s.ptr;
}
```
Is differentiated as
```
void fn_grad(double x, float y, double *_d_x, float *_d_y) {
    Session _d_s = {0., nullptr};     // This should be `{0., &*_d_y}`
    Session s = {x, &y};
    {
        _d_s.x += 1;
        *_d_s.ptr += 1;       // We dereference a nullptr
    }
    *_d_x += _d_s.x;
}
```
This is currently a blocker for #1413, where instead of generating
```
std::reference_wrapper<double> _d_x[]{_d_params[3], _d_params[2], _d_params[1], _d_params[0]};
```
We attempt to generate
```
std::reference_wrapper<double> _d_x[] = {0.};
```
Obviously, it doesn't work because ``std::reference_wrapper`` expects Lvalues, it cannot be default/zero initialized.
We should only replace adjoints with zero-literals when we're sure that it's safe, when they are Rvalue builtin (or real) types.

PetroZarytskyi:
The problem addressed in this PR is explained in detail in #1560. For short, when differentiating
```
Session s = {x, &y};
```
we initialize ``_d_s`` using ``getZeroInit()`` and get
```
Session _d_s = {0., nullptr}; // should be {0., &*_d_y}
```
This is quite a fundamental issue because we cannot fix ``getZeroInit()`` to have access to ``_d_y`` (and know anything about how ``y`` is differentiated in general). The only way to fix this is perhaps to make ``RMV::VisitInitListExpr`` return ``{0., &*_d_y}`` as the derivative.
The solution in this PR is based on the one described in #1562. This PR addresses the primary points in the issue, but doesn't remove all checks (mostly not to overload the PR with unrelated changed tests).

Fixes #1560.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -3304,7 +3304,14 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
   ReverseModeVisitor::VisitImplicitCastExpr(const ImplicitCastExpr* ICE) {
     // Casts should be handled automatically when the result is used by
     // Sema::ActOn.../Build...
-    return Visit(ICE->getSubExpr(), dfdx());
+    StmtDiff result = Visit(ICE->getSubExpr(), dfdx());
+    // In the forward pass, builtin-type derivatives are always zero.
+    // We can replace rvalue builtin types with zeros for simplicity,
+    // e.g., `double _d_x = 0.` looks simpler than `double _d_x = _d_y`.
+    if (ICE->getType()->isBuiltinType() &&
+        ICE->getCastKind() == CK_LValueToRValue)
+      result.updateStmtDx(getZeroInit(ICE->getType()));
+    return result;
   }
 
   StmtDiff ReverseModeVisitor::VisitImplicitValueInitExpr(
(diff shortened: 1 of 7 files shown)


test/Arrays/ArrayInputsReverseMode.C
```diff
@@ -158,7 +158,7 @@ double func4(double x) {
 //CHECK: void func4_grad(double x, double *_d_x) {
 //CHECK-NEXT:     int _d_i = 0;
 //CHECK-NEXT:     int i = 0;
-//CHECK-NEXT:     double _d_arr[3] = {0};
+//CHECK-NEXT:     double _d_arr[3] = {};
```
vgvassilev: Is it intended to create a non-zero initialized array here?

PetroZarytskyi: No, just to zero-initialize it. Is there a difference between ``{0}`` and ``{}`` in this context? I can also make it ``{0}`` again. The code for generating ``{}`` was just a little simpler.

PetroZarytskyi: Just in case: this logic is only applied to const arrays of builtin types (pointers/references don't count).

vgvassilev: The {0} means initialize all array elements with 0.

PetroZarytskyi: Right. I believe {} does the same, doesn't it?

vgvassilev: No. Maybe in c++23: https://en.cppreference.com/w/c/language/array_initialization.html

PetroZarytskyi: Reverted.

vgvassilev:
Can you make a set of commits that you want to land? Alternatively I can squash all.

PR #1567, opened 2025-09-25
Move derivative call building to a separate function

PetroZarytskyi:
Every time a call to a derivative is built (in all modes), we check if the derivative is a method, and then decide how to handle the method base. We have a function ``BuildCallExprToFunction``, which is supposed to be able to cover both function and method calls. However, it can only build calls to methods called on ``this``. This PR addresses this limitation and uses this function to simplify the recurring pattern we have across modes.
This PR also sinks the base derivatives in RMV::VisitCallExpr into the list of differentiated arguments for more consistent handling.


lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -339,7 +339,9 @@ DerivativeAndOverload HessianModeVisitor::Derive() {
       if (m_DiffReq.Mode == DiffMode::hessian_diagonal) {
         const size_t HessianMatrixStartIndex = i;
         // Call the derived function for second derivative.
-        Expr* call = BuildCallExprToFunction(secDerivFuncs[i], DeclRefToParams);
+        Expr* call =
+            BuildCallExprToFunction(secDerivFuncs[i], DeclRefToParams, nullptr,
```
vgvassilev:
```suggestion
            BuildCallExprToFunction(secDerivFuncs[i], DeclRefToParams, /*CUDAExecConfig=*/nullptr,
```

lib/Differentiator/HessianModeVisitor.cpp
```diff
@@ -367,7 +369,9 @@ DerivativeAndOverload HessianModeVisitor::Derive() {
           DeclRefToParams.push_back(SliceExpr);
           columnIndex += indArgSize;
         }
-        Expr* call = BuildCallExprToFunction(secDerivFuncs[i], DeclRefToParams);
+        Expr* call =
+            BuildCallExprToFunction(secDerivFuncs[i], DeclRefToParams, nullptr,
```
vgvassilev:
```suggestion
            BuildCallExprToFunction(secDerivFuncs[i], DeclRefToParams, /*CUDAExecConfig=*/nullptr,
```

lib/Differentiator/VectorForwardModeVisitor.cpp
```diff
@@ -323,7 +323,7 @@ VectorForwardModeVisitor::CreateVectorModeOverload(FunctionDecl* derivative) {
     addToCurrentBlock(BuildDeclStmt(vectorModeVD));
   }
 
-  Expr* callExpr = BuildCallExprToFunction(derivative, callArgs,
+  Expr* callExpr = BuildCallExprToFunction(derivative, callArgs, nullptr,
```
vgvassilev:
```suggestion
  Expr* callExpr = BuildCallExprToFunction(derivative, callArgs, /*CUDAExecConfig=*/nullptr,
```

PR #1568, opened 2025-09-25
Improve the diagnostics for propagator signature mismatch.

vgvassilev:


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -961,77 +962,112 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
...
+      IdentifierInfo* II = &C.Idents.get(Name);
+      DeclarationNameInfo DNInfo(II, utils::GetValidSLoc(S));
+      LookupResult Found(S, DNInfo, Sema::LookupOrdinaryName);
+      S.LookupQualifiedName(Found, DC);
+      return Found;
+    };
+
+    std::string Name = R.ComputeDerivativeName();
+    LookupResult Found = LookupPropagator(Name);
+    // FIXME: This is a hack to reuse the builtin derivatives for vector mode.
```
PetroZarytskyi, line 992: I perceived it rather as a feature than as a hack. Vector forward mode is the same forward mode, but with multiple derivatives computed in one run. It mostly works by the same rules. In my understanding, it made perfect sense to reuse the derivatives if possible. Do you think we should rewrite the derivatives for vector mode?

vgvassilev: I believe we could probably have cases where vector operations can be expressed more efficiently for some derivatives?

PR #1570, opened 2025-09-26
Support for-loops without increments in the reverse mode.

### LINKED ISSUE #1436: Clad crashes if the loop increment expression is empty

PetroZarytskyi:
Apart from accounting for the absence of increment in for-loops, this PR removes the part of the code that is supposed to wrap condition derivative in a lambda if it contains statements. It doesn't seem possible to cover this case. Currently, while differentiating expressions, we only generate expressions, not statements. In case this changes, it will be easy to spot because ``cast<Expr>(S)`` will simply crash.

Fixes #1436.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1103,41 +1103,24 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     if (FS->getCond())
       std::tie(condDiff, condExprDiff) = DifferentiateSingleExpr(FS->getCond());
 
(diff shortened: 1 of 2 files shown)


vgvassilev:
> Apart from accounting for the absence of increment in for-loops, this PR removes the part of the code that is supposed to wrap condition derivative in a lambda if it contains statements. It doesn't seem possible to cover this case. Currently, while differentiating expressions, we only generate expressions, not statements. In case this changes, it will be easy to spot because `cast<Expr>(S)` will simply crash.
> 
> Fixes 1436.

You need a # upfront.

vgvassilev (approved):
LGTM!

PR #1575, opened 2025-09-27
Initialize builtin type adjoints with init derivatives

PetroZarytskyi:
Currently, when differentiating buitin type variables in the reverse mode, we initialize them using ``getZeroInit()``.
After #1563, differentiating builtin rvalue expression types, like ``1.`` or ``x + y`` (numericals), gives a zero of the corresponding type as the derivative. This means that we can initialize adjoint variables of such types with the derivatives of the original inits. This both makes ``VarDecl`` differentiation more consistent and makes adjoint decls better reflect the original code, for example
```
double prod = 1;
```
used to be differentiated as 
```
double _d_prod = 0.; // zero of the type of the variable
double prod = 1;
```
and now, it's simply
```
double _d_prod = 0; // zero of the same type as the original initializer `1`
double prod = 1;
```

### DIFF
--- a/test/Misc/RunDemos.C
+++ b/test/Misc/RunDemos.C
@@ -109,7 +109,7 @@
 //CHECK_FLOAT_SUM:    unsigned int _d_i = 0U;
 //CHECK_FLOAT_SUM:    unsigned int i = 0U;
 //CHECK_FLOAT_SUM:    clad::tape<float> _t1 = {};
-//CHECK_FLOAT_SUM:    float _d_sum = 0.F;
+//CHECK_FLOAT_SUM:    float _d_sum = 0.;
 //CHECK_FLOAT_SUM:    float sum = 0.;
 //CHECK_FLOAT_SUM:    unsigned {{int|long|long long}} _t0 = {{0U|0UL|0ULL}};
 //CHECK_FLOAT_SUM:    for (i = 0; i < n; i++) {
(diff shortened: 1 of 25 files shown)


test/Misc/RunDemos.C
```diff
@@ -109,7 +109,7 @@
 //CHECK_FLOAT_SUM:    unsigned int _d_i = 0U;
 //CHECK_FLOAT_SUM:    unsigned int i = 0U;
 //CHECK_FLOAT_SUM:    clad::tape<float> _t1 = {};
-//CHECK_FLOAT_SUM:    float _d_sum = 0.F;
+//CHECK_FLOAT_SUM:    float _d_sum = 0.;
 //CHECK_FLOAT_SUM:    float sum = 0.;
 //CHECK_FLOAT_SUM:    unsigned {{int|long|long long}} _t0 = {{0U|0UL|0ULL}};
```
vgvassilev, line 114: I really want to get rid of these `{{0U|0UL|0ULL}}` things..

PetroZarytskyi: This loop iterator is initialized with getZeroInit separately. It's a purely Clad-generated variable. If you prefer, we can just initialize it with an int 0, it's not hard to do

vgvassilev: Ok, if that's outside of the scope of this PR we can track it in an issue for now.

PR #1580, opened 2025-09-29
Differentiate method base before its arguments

PetroZarytskyi:
Currently, in ``RMV::VisitCallExpr``, we differentiate the base after the arguments. Our reverse mode differentiation is designed in such a way that expressions should be differentiated in the same order they are executed.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1883,27 +1883,12 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     // The set of arguments to be used in a ``reverse_forw`` after the original
     // args.
     llvm::SmallVector<Expr*, 16> revForwAdjointArgs{};
-    for (std::size_t i = static_cast<std::size_t>(isMethodOperatorCall),
-                     e = CE->getNumArgs();
-         i != e; ++i) {
-      const Expr* arg = CE->getArg(i);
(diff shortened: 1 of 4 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1938,12 +1923,28 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
-          CallArgDx.insert(CallArgDx.begin(), baseDerivative);
-          revForwAdjointArgs.insert(revForwAdjointArgs.begin(), baseDerivative);
+          CallArgDx.push_back(baseDerivative);
+          revForwAdjointArgs.push_back(baseDerivative);
         }
       }
     }
 
+    for (std::size_t i = static_cast<std::size_t>(isMethodOperatorCall),
+                     e = CE->getNumArgs();
```
```suggestion
    for (std::size_t skip_this = static_cast<std::size_t>(isMethodOperatorCall), i = skip_this,
                     e = CE->getNumArgs();
```
and below you can use that variable in `PVD`...

PetroZarytskyi: Done. Is that what you had in mind?

vgvassilev (approved):
Lgtm!

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: Yeah.

PR #1583, opened 2025-09-30
Diagnose user-provided single argument pullbacks.

vgvassilev:
This patch removes the unused pullbacks, too.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -1215,13 +1215,13 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
         return true;
 
       request.Function = FD;
+      request.CallContext = E;
       bool canUsePushforwardInRevMode =
           m_TopMostReq->Mode == DiffMode::reverse &&
           utils::canUsePushforwardInRevMode(FD);
 
-      // FIXME: hessians require second derivatives,
-      // i.e. apart from the pushforward, we also need
-      // to schedule pushforward_pullback.
+      // FIXME: hessians require second derivatives, i.e. apart from the
+      // pushforward, we also need to schedule pushforward_pullback.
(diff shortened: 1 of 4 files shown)


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1287,9 +1286,43 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
...
+      if (canUsePushforwardInRevMode &&
+          m_TopMostReq->Mode == DiffMode::reverse) {
+        DiffRequest R = request;
+        R.BaseFunctionName = utils::ComputeEffectiveFnName(R.Function);
+        R.Mode = DiffMode::pullback;
+        if (LookupCustomDerivativeDecl(R)) {
+          unsigned warnId = m_Sema.Diags.getCustomDiagID(
+              DiagnosticsEngine::Warning,
+              "unused function '%0_pullback'; '%0' is a single-argument "
+              "function and only pushforward is required");
```
PetroZarytskyi: We should be a little careful about the phrasing here. Not all single-argument functions are differentiated as pushforward, only if the argument (and the return value) is of a real non-reference type. For example, that one argument could be an array, and we would still need a pullback.
Probably this is not worth explaining in detail, maybe just phrase it as "single real argument function"?

PetroZarytskyi (approved):
LGTM! This is a great change, I think such things should make Clad way easier to use.

lib/Differentiator/DiffPlanner.cpp
vgvassilev: Here "real" can have a different meaning. Can you suggest something else that conveys the intent?

PetroZarytskyi: Single numerical argument function?

vgvassilev: I added “real domain”.

PR #1589, opened 2025-10-01
Add custom derivatives for Thrust prefix sum operations

a-elrawy:
This PR adds custom derivatives for `thrust::inclusive_scan` and `thrust::exclusive_scan` to support reverse-mode automatic differentiation of common parallel scan patterns in Thrust.


vgvassilev (approved):
LGTM!

PR #1598, opened 2025-10-06
Introduce a single clad:Tag and reimplement other tags with `using`

PetroZarytskyi:
We use tags to distinguish between overloads of some derivatives that cannot be distinguished based on parameter types. We do it to ``constructor_pushforward`` and ``constructor_reverse_forw``.  However, there is no reason to have separate tags for ``constructor_pushforward`` and ``constructor_reverse_forw``, because different modes always have different names. This PR replaces both of these tags with a single ``clad::Tag`` and reimplaments the old tags as aliases to the new one in order to have backwards compatibility.


vgvassilev (approved):
LGTM!

PR #1599, opened 2025-10-06
Added Generic functor support for transform

a-elrawy:
- Added support for user functors in thrust::transform (unary/binary) via operator_call_pullback. 
- Preserved builtin ops (negate, identity, plus, minus, multiplies, divides).
- Added new complex tests using custom operations.


vgvassilev (approved):
LGTM!

PR #1601, opened 2025-10-07
Add support for `elidable_reverse_forw` to `constructor_reverse_forw`

PetroZarytskyi:
This PR is a follow-up to https://github.com/vgvassilev/clad/pull/1586 that introduced the `elidable_reverse_forw` attribute to simplify `reverse_forw` of functions. This PR adds support for this attribute also for `constructor_reverse_forw`.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -3078,10 +3078,8 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     llvm::SmallVector<Stmt*, 16> inits;
     llvm::SmallVector<Decl*, 4> decls;
     llvm::SmallVector<Decl*, 4> declsDiff;
-    llvm::SmallVector<Decl*, 4> classDeclsDiff;
+    llvm::SmallVector<Decl*, 4> declsToZeroInit;
     llvm::SmallVector<Stmt*, 4> memsetCalls;
-    // Need to put array decls inlined.
-    llvm::SmallVector<Decl*, 4> localDeclsDiff;
     // reverse_mode_forward_pass does not have a reverse pass so declarations
     // don't have to be moved to the function global scope.
     bool promoteToFnScope =
@@ -3165,11 +3163,9 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
         if (VDDiff.getDecl_dx()) {
           const CXXRecordDecl* RD = VD->getType()->getAsCXXRecordDecl();
           bool isNonAggrClass = RD && !RD->isAggregate();
-          if (isa<VariableArrayType>(VD->getType()))
-            localDeclsDiff.push_back(VDDiff.getDecl_dx());
-          else if (isNonAggrClass) {
-            classDeclsDiff.push_back(VDDiff.getDecl_dx());
-          } else {
+          if (isa<VariableArrayType>(VD->getType()) || isNonAggrClass)
+            declsToZeroInit.push_back(VDDiff.getDecl_dx());
+          else {
             VarDecl* VDDerived = VDDiff.getDecl_dx();
             declsDiff.push_back(VDDerived);
             if (Stmt* memsetCall = CheckAndBuildCallToMemset(
@@ -3194,22 +3190,6 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     Stmt* DSClone = nullptr;
     if (!decls.empty())
       DSClone = BuildDeclStmt(decls);
-
-    if (!localDeclsDiff.empty()) {
-      Stmt* localDSDIff = BuildDeclStmt(localDeclsDiff);
-      addToCurrentBlock(
-          localDSDIff,
-          clad::rmv::forward); // Doesnt work for arrays decl'd in loops.
-      for (Decl* decl : localDeclsDiff)
-        if (const auto* VAT =
-                dyn_cast<VariableArrayType>(cast<VarDecl>(decl)->getType())) {
-          std::array<Expr*, 2> args{};
-          args[0] = BuildDeclRef(cast<VarDecl>(decl));
-          args[1] = Clone(VAT->getSizeExpr());
-          Stmt* initCall = GetCladZeroInit(args);
-          addToCurrentBlock(initCall, direction::forward);
-        }
-    }
     if (!declsDiff.empty()) {
       Stmt* DSDiff = BuildDeclStmt(declsDiff);
       Stmts& block =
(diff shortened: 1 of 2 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4315,6 +4326,21 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+    if (callsite->getNumArgs() == 1)
+      return args[0];
+    if (callsite->isListInitialization() || !isDeclInit)
+      return S.ActOnInitList(noLoc, args, noLoc).get();
+    // ParenList is empty -- default initialisation.
+    // Passing empty parenList here will silently cause 'most vexing
+    // parse' issue.
+    if (callsite->getNumArgs() == 0)
+      return nullptr;
+    return S.ActOnParenListExpr(noLoc, noLoc, args).get();
```
```suggestion
    unsigned NArgs = callsite->getNumArgs();
    if (!NArgs)
      return nullptr;
    if (NArgs == 1)
      return args[0];
    if (callsite->isListInitialization() || !isDeclInit)
      return S.ActOnInitList(noLoc, args, noLoc).get();
    // ParenList is empty -- default initialisation.
    // Passing empty parenList here will silently cause 'most vexing
    // parse' issue.
    return S.ActOnParenListExpr(noLoc, noLoc, args).get();
```

include/clad/Differentiator/STLBuiltins.h
```diff
@@ -758,9 +753,8 @@ void constructor_pullback(const ::std::shared_ptr<T>& p, U*,
 // std::weak_ptr<T> custom derivatives...
 template <typename T, typename U>
 clad::ValueAndAdjoint<::std::weak_ptr<T>, ::std::weak_ptr<T>>
-constructor_reverse_forw(clad::Tag<::std::weak_ptr<T>>, U p, U d_p) {
-  return {::std::weak_ptr<T>(p), ::std::weak_ptr<T>(d_p)};
-}
+    elidable_reverse_forw
+    constructor_reverse_forw(clad::Tag<::std::weak_ptr<T>>, U p, U d_p);
```
vgvassilev: Isn't that weird indentation?

PetroZarytskyi: I also think it is, but clang-format goes a little crazy with attributes. Do you have an idea how to fix it?

vgvassilev: Maybe move the attribute before the `(`?

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: Done except I kept ``if (!NArgs)`` where it was. The reason is that if we have list initialization with 0 args, we still want to create a list init (the previous case).

include/clad/Differentiator/STLBuiltins.h
PetroZarytskyi: Fixed.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3260,9 +3266,14 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
         const auto* CE = dyn_cast<CXXConstructExpr>(init->IgnoreImplicit());
-        bool copyInit =
-            CE && (CE->getNumArgs() == 0 ||
-                   isa<DeclRefExpr>(CE->getArg(0)->IgnoreImplicit()));
+        bool copyInit = CE && CE->getNumArgs() == 0;
+        if (!copyInit && CE) {
+          if (const auto* DRE =
+                  dyn_cast<DeclRefExpr>(CE->getArg(0)->IgnoreImplicit()))
+            if (cast<VarDecl>(decl)->getNameAsString() ==
+                "_d_" + cast<VarDecl>(DRE->getDecl())->getNameAsString())
```
vgvassilev: Can we move these into local variables?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4315,6 +4326,22 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+    unsigned NArgs = callsite->getNumArgs();
+    if (NArgs == 1)
+      return args[0];
+    if (callsite->isListInitialization() || !isDeclInit)
+      return S.ActOnInitList(noLoc, args, noLoc).get();
+    // ParenList is empty -- default initialisation.
+    // Passing empty parenList here will silently cause 'most vexing
+    // parse' issue.
+    if (!NArgs)
+      return nullptr;
```
vgvassilev, line 4333: Why this check cannot go first?

PetroZarytskyi: I explained above. If we have list initialization with 0 args, we still want to create a list init (the previous case). For instance, `{}` might be a valid argument to a call expr, and passing `nullptr` will crash it. List init should generally stay list init.

vgvassilev: Ok, understood. Sorry.

lib/Differentiator/ReverseModeVisitor.cpp
PetroZarytskyi: This is actually a good idea. I merged these two sets in the last commit. Let me know if that's what you had in mind.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3265,20 +3238,44 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+          }
+          if (const auto* CE =
+                  dyn_cast<CXXConstructExpr>(init->IgnoreImplicit())) {
+            copyInit = CE && CE->getNumArgs() == 0;
+            if (!copyInit && CE) {
+              if (const auto* DRE =
+                      dyn_cast<DeclRefExpr>(CE->getArg(0)->IgnoreImplicit()))
+                if (cast<VarDecl>(decl)->getNameAsString() ==
+                    "_d_" + cast<VarDecl>(DRE->getDecl())->getNameAsString())
+                  copyInit = true;
```
vgvassilev, line 3258:
```suggestion
              if (const auto* DRE =
                      dyn_cast<DeclRefExpr>(CE->getArg(0)->IgnoreImplicit()))
                if (cast<VarDecl>(decl)->getName() ==
                    "_d_" + cast<VarDecl>(DRE->getDecl())->getName())
                  copyInit = true;
```

vgvassilev: What relationship are we trying to establish here?

PetroZarytskyi: We're trying to detect our copy-init pattern
```
std::vector<double> v = ...;
std::vector<double> _d_v(v);      // <-- we're trying to find this
```
It's not enough to check that it's initialized with a copy of anything.

PetroZarytskyi: Also, since `getName()` returns `StringRef`, the expression  `"_d_" + ...->getName()` will have the type of `Twine`, which is not comparable with `StringRef`. We could convert it to `StringRef` with
```
llvm::SmallVector<char, 64> Buffer;
if (cast<VarDecl>(decl)->getName() == ("_d_" + cast<VarDecl>(DRE->getDecl())->getName()).toStringRef(Buffer))
```
but it somewhat defeats the purpose of simplifying the code.

vgvassilev: Ok, thanks!

vgvassilev (approved):
LGTM!

PR #1602, opened 2025-10-07
Add support for conversion operators

PetroZarytskyi:
We build all names of conversion operators using the prefix `conversion_operator`, for example, `conversion_operator_pullback`. In order to differentiate between conversion operators to different types, we add a `clad::Tag` parameter to the beginning of the signature.

This PR is an alternative approach to #1590.


vgvassilev (approved):
LGTM!

PR #1607, opened 2025-10-09
Update MacOS arm jobs in ci to MacOS 26

mcbarton:


.github/workflows/ci.yml
```diff
@@ -22,33 +22,33 @@ jobs:
       matrix:
 
         include:
-          - name: osx14-arm-clang-runtime14
-            os: macos-14
+          - name: osx26-arm-clang-runtime14
```
mcbarton: @vgvassilev Is this job needed? Its the only MacOS 26 job failing, and you have no MacOS arm jobs for runtime15 and runtime16 for some reason.

vgvassilev: Runtime-14 is rather old. Maybe we can remove it.

mcbarton: The job has been removed and now the ci passes for the MacOS 26 arm jobs.

PR #1611, opened 2025-10-11
Add custom derivatives for thrust::sort_by_key

a-elrawy:
Implement reverse-mode support for `thrust::sort_by_key` (with/without comparator)

### DIFF
--- a/include/clad/Differentiator/ThrustDerivatives.h
+++ b/include/clad/Differentiator/ThrustDerivatives.h
@@ -4,6 +4,7 @@
 #include <cstddef>
 #include <iterator>
 #include <thrust/adjacent_difference.h>
+#include <thrust/copy.h>
 #include <thrust/count.h>
 #include <thrust/device_ptr.h>
 #include <thrust/fill.h> // NOLINT(misc-include-cleaner)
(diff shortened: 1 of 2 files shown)


include/clad/Differentiator/ThrustDerivatives.h
```diff
@@ -895,6 +898,145 @@ adjacent_difference_reverse_forw(InputIt first, InputIt last, OutputIt result,
   return {::thrust::adjacent_difference(first, last, result, op), {}};
 }
 
+namespace detail {
+inline ::std::vector<::thrust::device_vector<::std::size_t>>&
+permutation_stack() {
+  static ::std::vector<::thrust::device_vector<::std::size_t>> stack;
```
vgvassilev, line 904: Why do we need this `static` state here? Can you point me to the implementation of `sort_by`?

a-elrawy: In the current implementation reverse_forw sorts keys in-place (via thrust::sort_by_key), so the original order (the exact permutation) is lost. 

The implementation should be in https://github.com/NVIDIA/cccl/blob/main/thrust/thrust/detail/sort.inl

vgvassilev: Okay, I do not see any static members. Do we have a way to avoid this `static` definition here?

cc: @PetroZarytskyi

PetroZarytskyi: This is a problem we've already had in STL. We made some steps to solve it with ``clad::restore_tracker``, but it's still not usable for most OOP. We really need to solve this, but perhaps we shouldn't block this PR because of this.

vgvassilev: Can we open an issue for that to track it and move forward with this PR?

PR #1612, opened 2025-10-11
Add support for OOP and enable respective numerical tests

ovdiiuv:
This PR adds basic support for OOP in VA and enables it in a lot of numerical tests.

### DIFF
--- a/lib/Differentiator/ActivityAnalyzer.cpp
+++ b/lib/Differentiator/ActivityAnalyzer.cpp
@@ -2,6 +2,7 @@
 #include "AnalysisBase.h"
 
 #include "clang/AST/Decl.h"
+#include "clang/AST/DeclCXX.h"
 #include "clang/AST/Expr.h"
 #include "clang/AST/ExprCXX.h"
 #include "clang/AST/Stmt.h"
@@ -12,6 +13,7 @@
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/Support/Casting.h"
 
+#include <cstddef>
 #include <memory>
 #include <set>
 
@@ -28,10 +30,22 @@ void VariedAnalyzer::Analyze() {
 
   m_BlockData[m_CurBlockID] = std::make_unique<VarsData>();
 
+  const auto* MD = dyn_cast<CXXMethodDecl>(m_DiffReq.Function);
+  if (MD && !MD->isStatic()) {
+    VarData& thisData = getCurBlockVarsData()[nullptr];
+    thisData = VarData(MD->getThisType(), /*forceInit=*/true);
+    if (!isa<CXXConstructorDecl>(m_DiffReq.Function))
+      setIsRequired(&thisData);
+  }
+
   for (const auto* i : m_DiffReq.getVariedDecls()) {
     addVar(i, /*forceInit=*/true);
     setIsRequired(getVarDataFromDecl(i));
   }
+
+  if (const auto* CD = dyn_cast<CXXConstructorDecl>(m_DiffReq.Function))
+    for (auto* CI : CD->inits())
+      TraverseStmt(CI->getInit());
 
   auto paramsRef = m_DiffReq.Function->parameters();
   // If parameter was not marked as varied, add it's VarData and mark
@@ -155,8 +169,50 @@ bool VariedAnalyzer::TraverseConditionalOperator(ConditionalOperator* CO) {
   TraverseStmt(CO->getFalseExpr());
   return false;
 }
+bool VariedAnalyzer::TraverseCXXOperatorCallExpr(
+    clang::CXXOperatorCallExpr* CE) {
+  FunctionDecl* FD = CE->getDirectCallee();
+  const auto* MD = dyn_cast<CXXMethodDecl>(FD);
+  bool isMethodOperatorCall = MD && isa<CXXOperatorCallExpr>(CE);
+  Expr* baseOriginalE = CE->getArg(0);
+  bool hasVariedArg = false;
+  bool variedBefore = m_Varied;
+
+  for (std::size_t i = static_cast<std::size_t>(isMethodOperatorCall),
+                   e = CE->getNumArgs();
+       i != e; ++i) {
+    Expr* arg = CE->getArg(i);
+    const auto* PVD =
+        FD->getParamDecl(i - static_cast<unsigned long>(isMethodOperatorCall));
+
+    m_Varied = false;
+    TraverseStmt(arg);
+    if (m_Varied) {
+      hasVariedArg = true;
+      markExpr(arg);
+      m_DiffReq.addVariedDecl(PVD);
+    }
+  }
+
+  if (hasVariedArg || variedBefore) {
+    m_Varied = true;
+    m_Marking = true;
+  }
+
+  TraverseStmt(baseOriginalE);
+
+  m_Varied = false;
+  m_Marking = false;
+
+  m_Varied = hasVariedArg || variedBefore || m_DiffReq.isVaried(baseOriginalE);
+  return false;
+}
 
 bool VariedAnalyzer::TraverseCallExpr(CallExpr* CE) {
+  Expr* callee = CE->getCallee();
+  if (isa<CXXPseudoDestructorExpr>(callee))
+    return false;
+
   bool variedBefore = m_Varied;
   bool hasVariedArg = false;
   FunctionDecl* FD = CE->getDirectCallee();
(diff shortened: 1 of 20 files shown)


lib/Differentiator/ActivityAnalyzer.cpp
```diff
@@ -28,11 +30,29 @@ void VariedAnalyzer::Analyze() {
...
+      setIsRequired(&thisData);
+  }
+
   for (const auto* i : m_DiffReq.getVariedDecls()) {
     addVar(i, /*forceInit=*/true);
     setIsRequired(getVarDataFromDecl(i));
   }
 
+  if (isa<CXXConstructorDecl>(m_DiffReq.Function)) {
+    const auto* CD = dyn_cast<CXXConstructorDecl>(m_DiffReq.Function);
```
PetroZarytskyi: We can just do
```
if (const auto* CD = dyn_cast<CXXConstructorDecl>(m_DiffReq.Function)) {
```

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -517,6 +518,14 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
...
       }
 
+      // Prior to Clang 16 some functions were not considered built-in on the
+      // AST level, like std::move, which led us to differentianting their
+      // bodies.
+#if CLANG_VERSION_MAJOR < 16
+      std::string FDName = FD->getNameAsString();
+      if (FDName == "move" || FDName == "forward" || FDName == "capacity")
+        return {};
+#endif
```
PetroZarytskyi: We shouldn't do it just based on the name, at least check ``->isImplicit`` or that it's in std.

test/Gradient/Loops.C
```diff
@@ -2614,24 +2614,24 @@ struct tmp {
...
   }
-  float forward(const float &x) const {
+  float forw(const float &x) const {
     return x + z;
   }
 };
 struct layer {
   std::vector<tmp> w;
-  float forward(const float &inp) const {
+  float forw(const float &inp) const {
```
PetroZarytskyi: I don't think we should ban custom functions named "forward". This is the reason for the comment I left above.

ovdiiuv: Addressed.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -517,6 +519,15 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
         return result;
       }
 
+      // Prior to Clang 16 some functions were not considered built-in on the
+      // AST level, like std::move, which led us to differentianting their
+      // bodies.
+#if CLANG_VERSION_MAJOR < 16
+      std::string FDName = FD->getNameAsString();
+      if (clang::AnalysisDeclContext::isInStdNamespace(FD) &&
+          (FDName == "move" || FDName == "forward" || FDName == "capacity"))
```
vgvassilev: Why these functions have to be special?

ovdiiuv: Prior to clang 16 those functions were not treated as builtins and hence had actual bodies which we attempted to differentiate. For more details see
https://reviews.llvm.org/D123345

vgvassilev: Can you point where in clang are now treated as builtins?

ovdiiuv: Well, if you dump the ast, those functions will be marked "built-in", instead of instantiation.

ovdiiuv:
> > This PR adds basic support for OOP in VA and enables it in a lot of numerical tests.
> 
> How many tests are still left to enable?

I still have Root tests fail, and one test in Gradient folder. Around five in total.

vgvassilev:
> > > This PR adds basic support for OOP in VA and enables it in a lot of numerical tests.
> > 
> > 
> > How many tests are still left to enable?
> 
> Only hessian tests.

We probably need a dedicated line that only runs with enabled va. We also need a test for va just like we have for tbr.

vgvassilev:
I do not see `test/Analyses/VA.cpp` similar to `test/Analyses/TBR.cpp`. Also I'd expect a separate `RUN:` line for the `VA` tests, do we always want to do run the va tests with `-Xclang -disable-tbr`?

ovdiiuv:
> I do not see `test/Analyses/VA.cpp` similar to `test/Analyses/TBR.cpp`. Also I'd expect a separate `RUN:` line for the `VA` tests, do we always want to do run the va tests with `-Xclang -disable-tbr`?

It is under `test/Analyses/ActivityReverse.C`. 

We only run VA with `-disable-tbr` numerically. I saw VA could be affected by TBR in context of `reverse_forw`.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1347,15 +1356,15 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
         utils::hasMemoryTypeParams(request.Function) && request->isDefined() &&
         E->getDirectCallee();
     bool shouldUseRestoreTracker =
-        utils::shouldUseRestoreTracker(request.Function);
+        utils::shouldUseRestoreTracker(request.Function) && !nonDiff;
```
PetroZarytskyi: I don't think the restore tracker is related to non-differentiability. Non-diff functions (that don't require a pullback) can easily have side effects and require storing. Was this motivated by some tests?

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1347,15 +1356,15 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
...
     if (!(LookupCustomDerivativeDecl(request) || nonDiff) || requestTBR) {
       clang::CFG::BuildOptions Options;
       std::unique_ptr<AnalysisDeclContext> AnalysisDC =
           std::make_unique<AnalysisDeclContext>(
               /*AnalysisDeclContextManager=*/nullptr, request.Function,
               Options);
 
-      if (m_TopMostReq->EnableVariedAnalysis) {
+      if (request.EnableVariedAnalysis && request->isDefined()) {
```
PetroZarytskyi, line 1361: This is ugly, but it's not the fault of this PR. We need to make the traversal more flexible. Something from the scope of #1628.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -247,6 +247,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
 
   DerivativeAndOverload ReverseModeVisitor::Derive() {
     assert(m_DiffReq.Function && "Must not be null.");
+    // m_DiffReq->dump();
```
PetroZarytskyi: Let's remove this

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1192,8 +1192,18 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
...
 
+      request.VerboseDiags = false;
+      request.EnableTBRAnalysis = m_TopMostReq->EnableTBRAnalysis;
+      request.EnableVariedAnalysis = m_TopMostReq->EnableVariedAnalysis;
+      request.EnableUsefulAnalysis = m_TopMostReq->EnableUsefulAnalysis;
+      request.CallContext = E;
+
       const auto* MD = dyn_cast<CXXMethodDecl>(FD);
+      auto* mm = dyn_cast<CXXMemberCallExpr>(E);
+      // mm->dump();
```
PetroZarytskyi: Let's remove the commented out line and rename the variable to something like MCE to follow the usual style

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1206,6 +1217,10 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
       if (!(MD && MD->isInstance()) && !hasPointerOrRefReturn &&
           allArgumentsAreLiterals(E->arguments(), m_ParentReq))
         nonDiff = true;
```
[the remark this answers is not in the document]
vgvassilev: This issue does not seem to be fixed.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1207,6 +1216,10 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
...
+      else if (!hasPointerOrRefReturn && MCE && MD->isInstance() &&
+               allArgumentsAreLiterals(E->arguments(), m_ParentReq) &&
+               !m_ParentReq->isVaried(MCE->getImplicitObjectArgument()))
+        nonDiff = true;
```
vgvassilev: We don't have a test for this. Can you add one?

PR #1615, opened 2025-10-13
Elide memory function reverse_forw

PetroZarytskyi:
Currently, we have special behaviour for memory functions in the reverse mode: `malloc`, `calloc`, `realloc`, `memset`, `cudaMalloc`. However, we can now achieve the same behaviour by marking these functions with the attribute `elidable_reverse_forw` instead of hard-coding them.

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -142,6 +142,9 @@ template <typename T>
 void cudaMalloc_pullback(T** devPtr, size_t sz, cudaError_t d_ret, T** d_devPtr,
                          size_t* d_sz) __attribute__((host));
 
+cudaError_t cudaFree_reverse_forw(void* ptr, void* d_ptr) elidable_reverse_forw;
+
+void cudaFree_pullback(void* ptr, cudaError_t d_ret, void* d_ptr);
 #endif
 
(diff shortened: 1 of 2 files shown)


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1243,12 +1243,10 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
         return true;
       }
 
-      // FIXME: Generalize this to other functions that we don't need
-      // pullbacks of.
+      // FIXME: Use elidable_reverse_forw
```
vgvassilev, line 1246: Why not do it altogether now?

PetroZarytskyi: Well, I almost did it with begin/end but there was a broken corner case with std::initializer_list. In short, we replace std::initializer_list with clad::array, but std::begin instantiated for std::initializer_list doesn't work with clad::array. We need to reinstantiate it. I just opened this PR with what I had but I could also finish it in this PR.

PetroZarytskyi: Update: I fixed all of this, but doing this currently requires more workarounds than what we can remove. So I would say that this is currently not worth it.

vgvassilev: Makes sense. Can we track that into an issue?

PR #1616, opened 2025-10-14
Support loop checkpointing in the reverse mode

PetroZarytskyi:


test/Gradient/Loops.C
```diff
@@ -2735,6 +2735,56 @@ double fn44(double u, double v) {
 //CHECK-NEXT:        }
 //CHECK-NEXT:}
 
+double fn45(double x, double y) {
+  double sum = 0;
+  #pragma clad_checkpoint ON
```
vgvassilev:
```suggestion
  #pragma clad checkpoint loop
```

PetroZarytskyi: Done.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -732,6 +732,9 @@ namespace clad {
     // other cases, we have to use InitListExpr and change the constructor
     // style. Remove this once we generate constructors explicitly.
     bool m_TrackVarDeclConstructor = false;
+    /// A flag indicating if the Stmt we are currently inside a checkpointed
+    /// loop.
```
vgvassilev:
```suggestion
    /// A flag indicating if the Stmt is contained in a checkpointed loop.
```

test/Gradient/Loops.C
```diff
@@ -2732,6 +2732,121 @@ double fn44(double u, double v) {
 //CHECK-NEXT:        }
 //CHECK-NEXT:}
 
+double fn45(double x, double y) {
+  double sum = 0;
+  #pragma clad checkpoint loop
```
vgvassilev, line 2737: Can we add a test with this attached to a construct different from a loop?

test/Gradient/Loops.C
```diff
@@ -2732,6 +2732,121 @@ double fn44(double u, double v) {
...
+double fn46(double x, double y) {
+  double sum = 0;
+  #pragma clad checkpoint  // expected-warning {{expected 'loop' after 'checkpoint' in #pragma clad}}
+  for (int i = 0; i < 100; ++i) {
+    double t1 = x;
+    double t2 = y / t1;
+    sum += std::pow(t2, 2);
+  }
+  #pragma clad checkpoint other  // expected-warning {{expected 'loop' after 'checkpoint' in #pragma clad}}
+  while (false) {}
```
vgvassilev: What about `do-while`?

guitargeek:
To answer the question about the improved jitting time: the code didn't get shorter at all.

Here the example of one likelihood channel before and after the checkpointing with the pragma:

[out_old.C.txt](https://github.com/user-attachments/files/23207922/out_old.C.txt)
[out_new.C.txt](https://github.com/user-attachments/files/23207925/out_new.C.txt)

So it's probably indeed the templated tape class that makes things slow to compile.

include/clad/Differentiator/ReverseModeVisitor.h
PetroZarytskyi: Addressed.

test/Gradient/Loops.C
PetroZarytskyi: Added a test for both while and do-while.

test/Gradient/Loops.C
PetroZarytskyi: I added diagnostics of unused pragmas in the last commit.

test/Gradient/Loops.C
```diff
@@ -2793,6 +2793,8 @@ double fn46(double x, double y) {
   }
   #pragma clad checkpoint other  // expected-warning {{expected 'loop' after 'checkpoint' in #pragma clad}}
   while (false) {}
+
+  #pragma clad checkpoint loop  // expected-warning {{Unused clad checkpointing pragma, ignored.}}
```
vgvassilev: Can you reword it saying that the pragma works for loops only.

vgvassilev: And the diagnostic should probably be an error.

tools/ClangPlugin.cpp
```diff
@@ -209,6 +239,30 @@ void InitTimers();
...
+        request.m_CladLoopCheckpoints.emplace(*it, false);
+    }
+
+    static void diagnoseUnusedPragma(Sema& S, DiffRequest& request) {
+      for (const auto& pair : request.m_CladLoopCheckpoints) {
+        if (!pair.second) {
+          unsigned diagID = S.Diags.getCustomDiagID(
+              DiagnosticsEngine::Warning,
+              "Unused clad checkpointing pragma, ignored.");
+          S.Diag(pair.first, diagID);
```
vgvassilev, line 261:
```suggestion
          unsigned diagID = S.Diags.getCustomDiagID(
              DiagnosticsEngine::Error,
              "'#pragma clad checkpoint loop' is only allowed before a loop");
          S.Diag(pair.first, diagID) << pragma_source_location;
```

vgvassilev: There must be a way to get the pragma as string instead of spelling it out.

PetroZarytskyi: Done. I also moved the test to a separate file to verify the error.

tools/ClangPlugin.cpp
```diff
@@ -66,20 +69,47 @@ void InitTimers();
...
+            return;
+          }
+          CladLoopCheckpoints.insert(PragmaTok.getLocation());
+          return;
+        }
+        // Diagnose unknown clad pragma option
+        PP.Diag(
+            TokLoc,
+            PP.getDiagnostics().getCustomDiagID(
+                DiagnosticsEngine::Warning,
```
vgvassilev:
```suggestion
                DiagnosticsEngine::Error,
```

tools/ClangPlugin.cpp
```diff
@@ -66,20 +69,47 @@ void InitTimers();
...
         }
+        // Handle #pragma clad checkpoint loop
+        if (OptionName == "checkpoint") {
+          PP.Lex(PragmaTok);
+          // Ensure the next token is `loop`
+          if (PragmaTok.isNot(tok::identifier) ||
+              PragmaTok.getIdentifierInfo()->getName() != "loop") {
+            PP.Diag(PragmaTok.getLocation(),
+                    PP.getDiagnostics().getCustomDiagID(
+                        DiagnosticsEngine::Warning,
```
vgvassilev:
```suggestion
                        DiagnosticsEngine::Error,
```

vgvassilev (approved):
Lgtm!

tools/ClangPlugin.cpp
PetroZarytskyi: Done.

tools/ClangPlugin.cpp
PetroZarytskyi: Done. I also moved the respective tests to a separate file.

PR #1625, opened 2025-10-21
Generate constructor_reverse_forw automatically

### LINKED ISSUE #1577: Differentiate constructors in the forward pass properly
Even though we generate constructor pullbacks now, we still don't generate ``constructor_reverse_forw``, which should be used in the forward pass. To make matters worse, we don't warn the user that they need to provide a custom one. Also, constructors are often implicit (e.g., they arise when passing objects by value), and it might be completely not obvious to the user that they should write a custom derivative. That's exactly what happens in issue #1469.

It's worth mentioning that, when a custom ``constructor_reverse_forw`` is not provided, we sometimes use this pattern:
```
S s {...};
S _d_s(s);
clad::zero_init(_d_s);
```
However, the approach is only designed to cover iterable containers and purely numerical types like ``std::complex<double>``. If we deal with pointer fields, we just zero-initialize them (with ``clad::zero``). In other words, there's no guarantee that this approach actually gives us a correct answer. Writing an analysis to diagnose this would most likely take more effort than switching to a different system. Also note that this system only works when the constructor is used as a VarDecl initializer.
Once we fix the problem generally, we should probably get rid of this hack.

We have the following solutions to this problem:
1) Automatic ``constructors_reverse_forw``: would take more effort to implement, but would give us a very general solution. It would look somewhat like this:
```
class S {
  double m_x, m_y;
  S(double x, double y) {
    m_x = x;
    m_y = y;
  }
};
```
```
clad::ValueAndAdjoint<S, S>
S::constructor_reverse_forw(double x, double y, double _d_x, double _d_y) {
  S* _this = malloc(sizeof(S));
  S* _d_this = malloc(sizeof(S));
  _this->m_x = x;
  _d_this->m_x = _d_x;
  _this->m_y = y;
  _d_this->m_y = _d_y;
  return {*_this, *_d_this};
}
```
The malloc arises because we're trying to generate a function that has traits of constructors (raw uninitialized memory allocation). This is the exact same approach that we use with ``constructor_pullback``.

2) If reverse_forw is not provided, we use a pair of constructors. E.g. with copy constructors
<copy_constructor>(s) -> 
<copy_constructor>(s), <copy_constructor>(_d_s)

This would probably cover 95% of the cases and make them look very good, but the remaining 5% would fail silently and be only solvable with custom derivatives.

3) Produce a warning/error to force the user to write a custom derivative. This is a very awkward solution and can only be used as a temporary measure.

PetroZarytskyi:
This PR adds support for automatic constructor_reverse_forw. This solves the problem described in detail in issue #1577. By default, we generate constructor_reverse_forw for every constructor, which is the option that works in the general case. If possible, we detect that the constructor is elidable and simplify the code (the user can do it manually by adding the `elidable_reverse_forw` to the custom derivative).

Fixes #1577.

### DIFF
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -358,6 +358,44 @@ namespace clad {
       return true;
     }
 
+    bool isElidableConstructor(const clang::CXXConstructorDecl* CD,
+                               const clang::ASTContext& C) {
+      const CXXRecordDecl* RD = CD->getParent();
+      if (CD->isCopyOrMoveConstructor() && CD->isTrivial())
+        return true;
+      if (RD->isAggregate())
+        return true;
+      if (utils::unwrapIfSingleStmt(CD->getBody()))
+        return false;
+      bool isZeroOrCopyCtor = true;
+      for (CXXCtorInitializer* CI : CD->inits()) {
+        if (CI->isMemberInitializer()) {
+          QualType memberTy = CI->getMember()->getType();
+          if (memberTy->isPointerType() || memberTy->isLValueReferenceType())
+            continue;
+        }
+        Expr* init = CI->getInit()->IgnoreImplicit();
+        Expr::EvalResult value;
+        bool isZero = false;
+        if (clad_compat::Expr_EvaluateAsConstantExpr(init, value, C)) {
+          const auto* val = &value.Val;
+          if (val->isArray() && val->hasArrayFiller())
+            val = &val->getArrayFiller();
+          if (val->isInt())
+            isZero = val->getInt() == 0;
+          else if (val->isFloat())
+            isZero = val->getFloat().isZero();
+        }
+
+        if (!(isa<DeclRefExpr>(init) || isa<CXXConstructExpr>(init) ||
+              isZero)) {
+          isZeroOrCopyCtor = false;
+          break;
+        }
+      }
+      return isZeroOrCopyCtor;
+    }
+
     void getRecordDeclFields(
         const clang::RecordDecl* RD,
         llvm::SmallVectorImpl<const clang::FieldDecl*>& fields) {
--- a/lib/Differentiator/ReverseModeForwPassVisitor.cpp
+++ b/lib/Differentiator/ReverseModeForwPassVisitor.cpp
@@ -11,6 +11,7 @@
 #include "clang/AST/Decl.h"
 #include "clang/AST/DeclCXX.h"
 #include "clang/AST/Expr.h"
+#include "clang/AST/OperationKinds.h"
 #include "clang/Basic/LLVM.h"
 
 #include <algorithm>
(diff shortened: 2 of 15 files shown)


lib/Differentiator/CladUtils.cpp
```diff
@@ -355,6 +355,44 @@ namespace clad {
...
+      bool isZeroOrCopyConstr = true;
+      for (CXXCtorInitializer* CI : CD->inits()) {
+        if (CI->isMemberInitializer()) {
+          QualType memberTy = CI->getMember()->getType();
+          if (memberTy->isPointerType() || memberTy->isLValueReferenceType())
+            continue;
+        }
+        Expr* init = CI->getInit()->IgnoreImplicit();
+        Expr::EvalResult value;
+        bool isNull = false;
```
```suggestion
        bool isZero = false;
```
Do you mean default initialized?

lib/Differentiator/CladUtils.cpp
```diff
@@ -355,6 +355,44 @@ namespace clad {
...
+    bool isElidableConstructor(const clang::CXXConstructorDecl* CD,
+                               const clang::ASTContext& C) {
+      const CXXRecordDecl* RD = CD->getParent();
+      if (CD->isCopyOrMoveConstructor() && CD->isTrivial())
+        return true;
+      if (RD->isAggregate())
+        return true;
+      if (utils::unwrapIfSingleStmt(CD->getBody()))
+        return false;
+      bool isZeroOrCopyConstr = true;
```
```suggestion
      bool isZeroOrCopyCtor = true;
```

lib/Differentiator/ReverseModeForwPassVisitor.cpp
```diff
@@ -67,15 +68,43 @@ DerivativeAndOverload ReverseModeForwPassVisitor::Derive() {
     beginBlock();
     beginBlock(direction::reverse);
 
+    // If we the differentiated function is a constructor, generate `this`
+    // object and differentiate its inits.
+    Stmt* constrReturnStmt = nullptr;
```
vgvassilev:
```suggestion
    Stmt* ctorReturnStmt = nullptr;
```

lib/Differentiator/CladUtils.cpp
PetroZarytskyi: Yes. The idea is that not all constructors represent zero-initialization, some fields might be non-zero by default. This logic is there to check if the constructor is truly default in the sense that we want.

vgvassilev: Then that'd mean isDefaultInit or similar...

PetroZarytskyi: But wouldn't it mean that default non-zero values should be `isDefaultInit = true`? Like
```
class name {
    double val = 12;
...
};
```
We're interested in detecting zero default values.

vgvassilev: Ok, then let's keep it as it is.

test/Gradient/Constructors.C
```diff
@@ -153,24 +190,31 @@ double fn2(double u, double v) {
...
-// CHECK-NEXT:      S2 s2(v);
-// CHECK-NEXT:      S2 _d_s2(s2);
-// CHECK-NEXT:      clad::zero_init(_d_s2);
-// CHECK-NEXT:      S3 s3(v);
-// CHECK-NEXT:      S3 _d_s3(s3);
-// CHECK-NEXT:      clad::zero_init(_d_s3);
-// CHECK-NEXT:      S5 s5(u);
-// CHECK-NEXT:      S5 _d_s5(s5);
-// CHECK-NEXT:      clad::zero_init(_d_s5);
+// CHECK-NEXT:      clad::ValueAndAdjoint<S1, S1> _t0 = S1::constructor_reverse_forw(clad::Tag<S1>(), u, 0.);
```
vgvassilev, line 206: Why these changes should not be considered as a regression? Is it because we stopped relying on the copy constructor?

PetroZarytskyi: We had a discussion that the copy + zero_init approach is not guaranteed to work. #1577 is the issue you asked me to open to capture it. 
We don't generate ``clad::zero``, and we don't have a way to tell whether a given type would actually work with this system. All the uses in custom derivatives rely on specific properties of STL containers. In particular, containers of real types. For example, there is no way to define ``clad::zero`` to make ``std::vector<double*>`` the same way.  Moreover, all STL cases can be covered with ``elidable_reverse_forw``, which gives even prettier results.
So now we have 2 options: the nice one that works in 95% of the cases (``elidable_reverse_forw``) and the generic  constructor_reverse_forw. However, if necessary, we can have a third one, which would be somewhere in between.

PetroZarytskyi: One more reason I wanted to get rid of the copy-init approach is that it only works for constructors used as var decl inits. For example, we don't do it if a constructor is used as an argument or as part of some other expr. If we keep it, and we want to avoid scheduling the constructor_reverse_forw, we'll have to consider the context of the constructor in the diffplanner.

vgvassilev: Sounds good, do we have any performance comparisons?

PR #1630, opened 2025-10-26
 Add BoW logistic regression Thrust demo

a-elrawy:
Adds `demos/CUDA/BoWLogisticRegression.cu` implementing a minimal logistic regression demo using Thrust and clad AD:

- Single-example logistic loss and L2-regularized 2-doc batch loss.
- Runs SGD over a 4-document mini-batch; prints loss, gradient header, iteration losses, and final accuracy.


vgvassilev (approved):
LGTM!

Issue #1629, opened 2025-10-25
No overload exists for std::forward


ovdiiuv:
Consider

```cpp
double f(double x){
   std::pair<double, double> p(x,1);
   return 0;
}
```

in clang16+ its derivative is:

```cpp
__attribute__((exclude_from_explicit_instantiation)) __attribute__((abi_tag("ne180100"))) static constexpr void constructor_pullback(double &__u1, int &&__u2, std::pair<double, double> *_d_this, double *_d___u1, int *_d___u2) noexcept((is_nothrow_constructible<first_type, double &>::value && is_nothrow_constructible<second_type, int>::value)) {
    std::pair<double, double> *_this = (std::pair<double, double> *)malloc(sizeof(std::pair<double, double>));
    clad::ValueAndAdjoint<double &, double &> _t0 = clad::custom_derivatives::std::forward_reverse_forw(__u1, *_d___u1);
    _this->first = _t0.value;
    _this->second = std::forward<int>(__u2);
    _d_this->second = 0.;
    {
        _t0.adjoint += _d_this->first;
        _d_this->first = 0.;
    }
    free(_this);
}
void f_grad(double x, double *_d_x) {
              . . .
}
```

notice that we are missing another `clad::ValueAndAdjoint` for `__u2`. That is because no suitable overload is found in STLBuiltins, since `__u2` is an lvalue and the type inside of ValueAndAdjoint is deduced to `int &&`.

PR #1635, opened 2025-11-01
Wrap argumets when visiting std::forward into std::move

ovdiiuv:
Previously when differentianting `std::forward` we would always choose one overload:
`ValueAndAdjoint<T&, T&> forward_reverse_forw(T& t, T& dt)`. We want to preserve the value categories of the arguments, so now we wrap them in `std::move` and manually find the right oveload of `forward_reverse_forw`.

Fixes: #1629

### DIFF
--- a/include/clad/Differentiator/STLBuiltins.h
+++ b/include/clad/Differentiator/STLBuiltins.h
@@ -872,14 +872,12 @@ template <typename... Args> auto make_tuple_pushforward(Args... args) noexcept {
 
 // std::forward custom derivatives
 template <class T>
-clad::ValueAndAdjoint<T&&, T&&> forward_reverse_forw(T&& t, T&& dt) {
-  return {::std::forward<T>(t), ::std::forward<T>(dt)};
-}
+clad::ValueAndAdjoint<T&&, T&&>
+    elidable_reverse_forw forward_reverse_forw(T&& t, T&& dt);
 
 template <class T>
-clad::ValueAndAdjoint<T&, T&> forward_reverse_forw(T& t, T& dt) {
-  return {t, dt};
-}
+clad::ValueAndAdjoint<T&, T&> elidable_reverse_forw forward_reverse_forw(T& t,
+                                                                         T& dt);
 
 template <class T>
 constexpr void forward_pullback(T&& t, T dy, T* dt) noexcept {
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -36,7 +36,6 @@
 #include "clang/Basic/LLVM.h" // isa, dyn_cast
 #include "clang/Basic/Specifiers.h"
 #include "clang/Basic/TokenKinds.h"
-#include "clang/Basic/Version.h"
 #include "clang/Sema/Lookup.h"
 #include "clang/Sema/Overload.h"
 #include "clang/Sema/Scope.h"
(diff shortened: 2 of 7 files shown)


test/Gradient/STLCustomDerivatives.C
```diff
@@ -1134,3 +1140,17 @@ int main() {
...
+// CHECK: void fn24_grad(double x, double *_d_x) {
+// CHECK-NEXT:    double _t0 = x;
+// CHECK-NEXT:    std::pair<double, double> p(x, 1);
+// CHECK-NEXT:    std::pair<double, double> _d_p(p);
+// CHECK-NEXT:    clad::zero_init(_d_p);
+// CHECK-NEXT:    _d_p.first += 1;
+// CHECK-NEXT:    {
+// CHECK-NEXT:        x = _t0;
+// CHECK-NEXT:        int _r0 = 0;
+// CHECK-NEXT:        std::pair<double, double>::constructor_pullback(x, 1, &_d_p, _d_x, &_r0);
```
PetroZarytskyi, line 1168: I'm assuming the goal of the PR is to generate the constructor_pullback correctly. Can you include that in the test?

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -518,7 +518,12 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
         }
         return result;
       }
-
+#if CLANG_VERSION_MAJOR < 16
+      std::string FDName = FD->getNameAsString();
+      if (clang::AnalysisDeclContext::isInStdNamespace(FD) &&
+          (FDName == "move" || FDName == "forward" || FDName == "capacity"))
+        return {};
+#endif
```
PetroZarytskyi: Can we do this as an early return in ``DiffCollector::VisitCallExpr``?

include/clad/Differentiator/STLBuiltins.h
```diff
@@ -872,12 +872,14 @@ template <typename... Args> auto make_tuple_pushforward(Args... args) noexcept {
...
+clad::ValueAndAdjoint<T&&, T&&>
+    elidable_reverse_forw forward_reverse_forw(T&& t, T&& dt) {
   return {::std::forward<T>(t), ::std::forward<T>(dt)};
 }
 
 template <class T>
-clad::ValueAndAdjoint<T&, T&> forward_reverse_forw(T& t, T& dt) {
+clad::ValueAndAdjoint<T&, T&>
+    elidable_reverse_forw forward_reverse_forw(T& t, T& dt) {
   return {t, dt};
```
PetroZarytskyi: If you add the `elidable_reverse_forw` attribute, it means the body will not be used. We should do
```
... forward_reverse_forw(...) elidable_reverse_forw;
```
Same above.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1740,8 +1742,18 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
     }
 
-    auto NArgs = FD->getNumParams();
     SourceLocation Loc = CE->getExprLoc();
+    bool shouldWrapInStdMove = false;
+    bool isCastSem = false;
+    if (clang::AnalysisDeclContext::isInStdNamespace(FD) &&
+        FDName == "forward") {
+      isCastSem = true;
+      // get back here
```
PetroZarytskyi: Either remove this comment or turn it into a FIXME

PR #1636, opened 2025-11-02
Fix: Ensure clad::gradient and clad::differentiate produce consistent results under numerical edge cases

Vedant2005goyal:
This commit implements a backward data-flow **activity analysis** in the `ReverseModeVisitor`.
The analysis starts at the `return` statement and traces all variable dependencies backward to find all "active" variables that contribute to the final result. In `VisitDeclStmt`, if a `VarDecl` is found to be **inactive** (like `a` in the example), its differentiation is now skipped. We only generate its original (forward-pass) code. This correctly prevents the derivative for `std::acos` from being generated, resolving the `NaN` error and making `clad::gradient`'s output consistent with `clad::differentiate`.
The test case from the issue now correctly outputs:
60,50,60
And the dumped gradient function is clean:
```cpp
void func_grad(double C, double A, double *_d_C, double *_d_A) {
    double a = std::acos(-C / A);
    {
        *_d_C += 2 * 1 * A * C;
        *_d_C += 2 * C * 1 * A;
        *_d_A += 2 * C * C * 1;
    }
}

**Note**: This patch is compatible with the clad codebase, which relies on Clang APIs from versions 11.x through 21.x (e.g., Clang 18). It will not compile against Clang 22 or newer due to breaking API changes in the Clang frontend that this clad code depends on.

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -42,12 +42,36 @@ namespace clad {
   class ErrorEstimationHandler;
   class ExternalRMVSource;
   class MultiplexExternalRMVSource;
+  class ActiveVariableAnalyzer {
+  private:
+    std::set<const clang::VarDecl*> m_ActiveVars;
+    std::set<const clang::Expr*> m_ActiveExprs;
+
+    void markActive(const clang::Expr* E);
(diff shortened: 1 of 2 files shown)


vgvassilev:
Can you add a test for this change?

PR #1638, opened 2025-11-04
Do not treat tensors in a special way

### LINKED ISSUE #1526: Do not treat tensors in a special way
In #1462 we introduced a tensor-specific code which is supposed to copy a tensor. However, in #1513 we tried to address that but only partially leading the main use-case to still crash.

@Rohan-T144, can you provide us with more details about the crash?

PetroZarytskyi:
Copy-initialization of adjoints has proven to be more reliable than default initialization. However, when handling non-differentiable parameters (unlike other variables), we use this strategy only with tensor-like types. This PR generalizes this approach to all copiable record types.

Fixes #1526


vgvassilev (approved):
LGTM!

PR #1639, opened 2025-11-04
Fix: Ensure clad::gradient and clad::differentiate produce consistent results under numerical edge cases 

Vedant2005goyal:
This patch implements a backward data-flow activity analysis in the ReverseModeVisitor.
The analysis starts at the return statement and traces all variable dependencies backward to find all "active" variables that contribute to the final result. In VisitDeclStmt, if a VarDecl is found to be inactive (like a in the example), its differentiation is now skipped. We only generate its original (forward-pass) code. This correctly prevents the derivative for std::acos from being generated, resolving the NaN error and making clad::gradient's output consistent with clad::differentiate.
Changes made in: 
ReverseModeVisitor.h 
ReverseModeVisitor.cpp
and also added a "test/Issue-1108.cpp" as a new regression test to fix this issue as requested by @vgvassilev 

While this fixes the original issue, it's causing a small number of other tests to fail (e.g., Analyses/ActivityReverse.cpp, ErrorEstimation/LoopsAndArrays.C). They are failing with Assertion failed: (detail::isPresent(Val) && "dyn_cast on a non-existent value"). I have tried to fix them as well but the failure persists and This suggests my logic in VisitDeclStmt is still missing something or is interacting with the rest of the visitor in a way I don't understand.
Could you provide some guidance on what I might be missing?
Thanks

### DIFF
--- a/test/Issue-1108.cpp
+++ b/test/Issue-1108.cpp
@@ -1,0 +1,63 @@
+// RUN: %cladclang %s -o %t && %t | %filecheck %s
+
+// CHECK: clad::gradient results:
+// CHECK-NEXT: 8 2
+// CHECK-NEXT: clad::differentiate results:
+// CHECK-NEXT: 8 2
+
+#include <iostream>
+#include <cmath> // Include for std::acos
+#include "clad/Differentiator/Differentiator.h"
+
+/**
+ * @brief The function to be differentiated.
+ * The function being differentiated is effectively f(C, A) = 2 * C^2 * A,
+ * as the 'a' variable is not used in the return value.
+ *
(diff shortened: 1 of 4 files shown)


test/Issue-1108.cpp
```diff
@@ -0,0 +1,63 @@
+// RUN: %cladclang %s -o %t && %t | %filecheck %s
```
vgvassilev, line 1: That test should probably go under test/Regressions.

Vedant2005goyal: I will do it

Vedant2005goyal: Thanks for replying @vgvassilev
I will surely do it

PR #1640, opened 2025-11-05
Don't allow non-differentiable non-const array parameters in forward mode

### LINKED ISSUE #1035: Unnecessary warning when differentiating a function that mutates its array arg

PetroZarytskyi:
All parameters w.r.t. which we don't differentiate still require adjoints. Since the user doesn't provide it, we have to construct it ourselves. However, if the parameter is a pointer array (e.g., `double*`), we don't know its size and, therefore, cannot initialize the adjoint.
If the array is const, it's guaranteed that it doesn't require an adjoint. If it's a const array (e.g., `double[3]`), we can initialize it. In other cases, we have to produce an error. We've been doing the same in the reverse mode for a while.

Fixes #1035.

### DIFF
--- a/benchmark/BenchmarkedFunctions.h
+++ b/benchmark/BenchmarkedFunctions.h
@@ -41,7 +41,7 @@ inline double product(double p[], int n) {
 }
 
 ///\returns the weighted sum of the elements in \p
-inline double weightedSum(double p[], double w[], int n) {
+inline double weightedSum(const double p[], const double w[], int n) {
   double sum = 0;
   for (int i = 0; i < n; i++)
     sum += p[i] * w[i];
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -317,16 +317,34 @@ void BaseForwardModeVisitor::GenerateSeeds(const clang::FunctionDecl* dFD) {
     // relation among them, thus it is safe (correct) to use the corresponding
     // non-reference type for creating the derivatives.
     QualType dParamType = param->getType().getNonReferenceType();
-    // We do not create derived variable for array/pointer parameters.
-    if (!utils::IsDifferentiableType(dParamType) ||
-        utils::isArrayOrPointerType(dParamType))
-      continue;
     Expr* dParam = nullptr;
-    if (dParamType->isRealType()) {
+    if (!utils::IsDifferentiableType(dParamType))
+      continue;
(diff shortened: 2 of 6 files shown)


benchmark/BenchmarkedFunctions.h
```diff
@@ -41,7 +41,7 @@ inline double product(double p[], int n) {
 }
 
 ///\returns the weighted sum of the elements in \p
-inline double weightedSum(double p[], double w[], int n) {
+inline double weightedSum(const double p[], const double w[], int n) {
```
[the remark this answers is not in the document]
PetroZarytskyi: We shouldn't change demos

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -317,16 +317,34 @@ void BaseForwardModeVisitor::GenerateSeeds(const clang::FunctionDecl* dFD) {
...
+    } else if (utils::isArrayOrPointerType(dParamType)) {
+      // We cannot initialize a pointer array adjoint ourselves.
+      // Produce an error.
+      if (param != m_IndependentVar &&
+          !utils::GetValueType(dParamType).isConstQualified()) {
+        diag(DiagnosticsEngine::Error, param->getLocation(),
+             "Non-differentiable non-const pointer and array parameters "
+             "are not supported. Please differentiate w.r.t. '%0' or mark "
+             "it const.",
+             {param->getNameAsString()});
```
vgvassilev, line 345:
```suggestion
        diag(DiagnosticsEngine::Error, param->getLocation(),
             "non-differentiable non-const pointer or array parameters are not supported; "
     "differentiate with respect to '%0' or mark it 'const'") << param;
```

We can create a fixit, too. Please take a look at #1596 how we should be doing diagnostics.

PetroZarytskyi: I added a fixme. I will address that in #1596

PR #1641, opened 2025-11-06
Add basic support for OpenMP in reverse mode

Errant404:
This PR introduces basic support for reverse mode in OpenMP.
The general approach is to privatize the tape for each thread using omp threadprivate, and then employ a custom omp for task scheduling so that each thread can replay its own temporary variables in reverse mode.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -3376,6 +3376,8 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
         auto* vDecl = cast<VarDecl>(decl);
         if (Expr* init = vDecl->getInit()) {
           if (promoteToFnScope) {
+            if (isInsideOMPBlock)
+              MarkDeclThreadPrivate(vDecl);
             auto* declRef = BuildDeclRef(vDecl);
             auto* assignment = BuildOp(BO_Assign, declRef, init);
             addToCurrentBlock(assignment, direction::forward);
--- a/test/Gradient/OpenMP.C
+++ b/test/Gradient/OpenMP.C
@@ -750,6 +750,72 @@ void fn19(const double *x, int start, int end, double *y) {
 // CHECK-NEXT:          }
 // CHECK-NEXT:  }
 
+struct Box20 {
+  double v;
+  Box20() : v(0.0) {}
+  Box20(double seed) : v(seed) {}
+};
+
+double fn20(const double* x, int n) {
+  double total = 0.0;
+  #pragma omp parallel for reduction(+:total)
+  for (int i = 0; i < n; ++i) {
+    Box20 b(1.0);
+    b.v = x[i];
+    total += b.v * b.v;
+  }
+  return total;
+}
+
+// CHECK:  void fn20_grad(const double *x, int n, double *_d_x, int *_d_n) {
+// CHECK-NEXT:      static clad::tape<Box20> _t0 = {};
+// CHECK-NEXT:      #pragma omp threadprivate(_t0);
+// CHECK-NEXT:      static clad::tape<Box20> _t1 = {};
+// CHECK-NEXT:      #pragma omp threadprivate(_t1);
+// CHECK-NEXT:      static Box20 b = {};
+// CHECK-NEXT:      #pragma omp threadprivate(b);
+// CHECK-NEXT:      static Box20 _d_b{};
+// CHECK-NEXT:      #pragma omp threadprivate(_d_b);
+// CHECK-NEXT:      double _d_total = 0.;
+// CHECK-NEXT:      double total = 0.;
+// CHECK-NEXT:      #pragma omp parallel reduction(+: total)
+// CHECK-NEXT:          {
+// CHECK-NEXT:              int _t_chunklo0 = 0;
+// CHECK-NEXT:              int _t_chunkhi0 = 0;
+// CHECK-NEXT:              clad::GetStaticSchedule(0, n - 1, 1, &_t_chunklo0, &_t_chunkhi0);
+// CHECK-NEXT:              for (int i = _t_chunklo0; i <= _t_chunkhi0; i += 1) {
+// CHECK-NEXT:                  clad::push(_t0, std::move(_d_b));
+// CHECK-NEXT:                  clad::push(_t1, std::move(b)) , b = 1.;
+// CHECK-NEXT:                  _d_b = 0.;
+// CHECK-NEXT:                  b.v = x[i];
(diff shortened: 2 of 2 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1472,10 +1479,18 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
 
       if (!it->second)
         return StmtDiff(clonedDRE);
+
+      clang::Expr* dExpr = it->second;
+      if (auto dVarDRE = dyn_cast<DeclRefExpr>(dExpr)) {
```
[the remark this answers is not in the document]
vgvassilev: Looks like this is not fixed. Can you make sure things are fixed before you mark them “resolved”?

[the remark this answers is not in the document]
Errant404: Sorry, I just marked it as fixed after fixing it locally.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4603,4 +4634,22 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+    Vars.push_back(declRef);
+    auto* TPDecl =
+        CLAD_COMPAT_CLANG19_SemaOpenMP(m_Sema).CheckOMPThreadPrivateDecl(
+            decl->getLocation(), Vars);
+    cast<VarDecl>(decl)->setInit(Init);
+    // Add the threadprivate declaration to the current context
+    m_Sema.CurContext->addDecl(TPDecl);
+    // Create a DeclStmt and add it to the global block for proper AST
+    // structure
+    Stmt* TPStmt = new (m_Context) DeclStmt(DeclGroupRef(TPDecl), noLoc, noLoc);
```
PetroZarytskyi: Does `VisitorBase::BuildDeclStmt` work here?

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -696,6 +729,8 @@ namespace clad {
     /// Builds and returns the sequence of derived function parameters.
     void BuildParams(llvm::SmallVectorImpl<clang::ParmVarDecl*>& params);
 
+    void MarkDeclThreadPrivate(clang::Decl* decl);
```
PetroZarytskyi: I'm under the impression that this function only makes sense for `clang::VarDecl`. If so, can we change its signature accordingly? Also, in this case, to avoid cast errors, we should call it inside of `RMV::DifferentiateVarDecl` (perhaps in the end) instead of `RMV::VisitDeclStmt`.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1482,10 +1490,18 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
 
       if (!it->second)
         return StmtDiff(clonedDRE);
+
+      clang::Expr* dExpr = it->second;
+      if (auto* dVarDRE = dyn_cast<DeclRefExpr>(dExpr)) {
+        auto* dVar = cast<VarDecl>(dVarDRE->getDecl());
+        if (dVar->getDeclContext() != m_Sema.CurContext)
+          dExpr = BuildDeclRef(dVar, DRE->getQualifier());
+      }
```
PetroZarytskyi, line 1540: This fix is fine for this PR, but it arises from a broader problem that we reuse the adjoint declref. We should instead store the decl of the adjoint in `m_Variables` and rebuild the declref every time. Please just add a fixme saying something similar to what I explained in this comment.

include/clad/Differentiator/ReverseModeVisitor.h
Errant404: What you said in the earlier part is correct, but `#pragma omp threadprivate` must be written after the variable declaration to be correct. In other words, this part of the AST must be placed later in `m_Globals`.

PetroZarytskyi: Makes sense.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4603,4 +4644,22 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       params.push_back(dPVD);
     }
   }
+  void ReverseModeVisitor::MarkDeclThreadPrivate(VarDecl* decl) {
+    auto* Init = decl->getInit();
+    // set to null to pass CheckOMPThreadPrivateDecl
+    decl->setInit(nullptr);
+    auto* declRef = BuildDeclRef(decl);
+    llvm::SmallVector<Expr*, 1> Vars;
+    Vars.push_back(declRef);
```
PetroZarytskyi:
```diff
-    llvm::SmallVector<Expr*, 1> Vars;
-    Vars.push_back(declRef);
+    llvm::SmallVector<Expr*, 1> Vars{declRef};
```

test/Gradient/OpenMP.C
```diff
@@ -0,0 +1,773 @@
...
+  auto fn10_grad = clad::gradient(fn10);
+  auto fn11_grad = clad::gradient(fn11);
+  auto fn12_grad = clad::gradient(fn12);
+  auto fn13_grad = clad::gradient(fn13);
+  auto fn14_grad = clad::gradient(fn14);
+  auto fn15_grad = clad::gradient(fn15);
+  auto fn17_grad = clad::gradient(fn16);
+  auto fn20_grad = clad::gradient(fn17);
+  auto fn21_grad = clad::gradient(fn18);
+  auto fn25_grad = clad::gradient(fn19);
```
PetroZarytskyi: Can we add numerical checks? I.e., execute the gradients on some values. You can include `"../TestUtils.h"` and `"../PrintOverloads.h"` to use macros `INIT_GRADIENT` and `TEST_GRADIENT`. This way, you won't need to write them by hand, and it will be quicker

.github/workflows/ci.yml
```diff
@@ -944,7 +945,7 @@ jobs:
       if: ${{ success() && (matrix.coverage == true) }}
       uses: codecov/codecov-action@v5
       with:
-        file: ./coverage.info
+        files: ./coverage.info
```
vgvassilev, line 623: Can we undo this change? Why is this needed?

Errant404: See https://github.com/codecov/codecov-action/tree/v5/?tab=readme-ov-file#migration-guide
image

vgvassilev: Ok, okay, thank you!

Errant404:
It looks like the code has been covered by testing.

PR #1644, opened 2025-11-11
Store adjoints as decls in m_Variables in the reverse mode

PetroZarytskyi:
Currently, we track the correspondance of the original variables and their adjoints in a map `m_Variables` as `x` --> `_d_x`. However, `x` is stored a decl and `_d_x` is stored as an expr. The latter results in us reusing the same `DeclRefExpr` of `_d_x`. Apart from that, it makes it hard to get the actual declaration of `_d_x`.

### DIFF
--- a/include/clad/Differentiator/BaseForwardModeVisitor.h
+++ b/include/clad/Differentiator/BaseForwardModeVisitor.h
@@ -172,6 +172,10 @@ protected:
       llvm::SmallVectorImpl<clang::Expr*>& clonedArgs,
       llvm::SmallVectorImpl<clang::Expr*>& derivedArgs);
 
+  /// Map used to keep track of variable declarations and match them
+  /// with their derivatives.
+  std::unordered_map<const clang::ValueDecl*, clang::Expr*> m_Variables;
+
 private:
   /// Prepares the derivative function parameters.
   void
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -14,6 +14,7 @@
 #include "clad/Differentiator/ReverseModeVisitorDirectionKinds.h"
 #include "clad/Differentiator/VisitorBase.h"
 
+#include "clang/AST/Decl.h"
 #include "clang/AST/DeclCXX.h"
 #include "clang/AST/Expr.h"
 #include "clang/AST/ExprCXX.h"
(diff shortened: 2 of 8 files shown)


include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -85,6 +85,9 @@ namespace clad {
     clang::Expr* m_CurrentBreakFlagExpr;
 
     clang::Expr* m_RestoreTracker = nullptr;
+    /// Map used to keep track of variable declarations and match them
+    /// with their derivatives.
+    std::unordered_map<const clang::ValueDecl*, clang::VarDecl*> m_Variables;
```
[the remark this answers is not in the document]
PetroZarytskyi: @vgvassilev Should we avoid protected members?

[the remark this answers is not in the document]
vgvassilev: Probably that's to late to consider... Let's ignore this and maybe adjust the config file.

include/clad/Differentiator/BaseForwardModeVisitor.h
```diff
@@ -172,6 +172,10 @@ class BaseForwardModeVisitor
       llvm::SmallVectorImpl<clang::Expr*>& clonedArgs,
       llvm::SmallVectorImpl<clang::Expr*>& derivedArgs);
 
+  /// Map used to keep track of variable declarations and match them
+  /// with their derivatives.
+  std::unordered_map<const clang::ValueDecl*, clang::Expr*> m_Variables;
```
vgvassilev, line 177: What is this files used for? In principle we cannot reuse expressions (or decl-ref expressions) as clang builds a separate whenever they are needed. Can we drop this in favor of some utility function that builds a new expression?

vgvassilev: ping.

PetroZarytskyi: This was in ``VisitorBase``. In other words, it was this way in all visitors. I moved it to ``ReverseModeVisitor`` and ``BaseForwardModeVisitor`` and fixed it in the reverse mode the way you're describing.

vgvassilev (approved):
LGTM!

guitargeek:
Landed as 861aac5c (#1896).

PR #1646, opened 2025-11-12
Enable remaining XFAIL tests on i586

PetroZarytskyi:


vgvassilev:
Thanks -- this seems already fixed by other recent work in the master.

PR #1649, opened 2025-11-13
Add support for init lists of different length in clad::move

PetroZarytskyi:
After #1626, we started initializing arrays inside loops with clad::move. However, since the size of the array was deduced from the size of the init list, this change silently dropped support for decls like `double arr[10] = {0}`.


vgvassilev (approved):
LGTM!

PR #1653, opened 2025-11-18
Replace custom error estimation models with custom `getErrorVal` functions

PetroZarytskyi:
Currently, if a user wants to define a custom error estimation model, they have to derive a class from `FPErrorEstimationModel` and override `clang::Expr* AssignError(StmtDiff refExpr, const std::string& name)`. Then, they need to attach it to the Clad plugin by writing their own CMake files and loading their compiled plugin in the invocation. Although we have demos that show how to do this, the process still seems unnecessarily complex.

The only custom functionality that `AssignError` provides is the freedom of how to compute the error. It takes `x`, `_d_x`, and `"x"` as parameters. We can achieve the same functionality by accepting a custom `getErrorVal` function. This way, the user wouldn't have to generate AST and work with Clad's codebase directly. This is more consistent with the way we model custom user-defined behaviour in other modes, i.e., custom derivatives. Moreover, it makes the implementation of error estimation a lot simpler.

Also, this PR removes `m_ErrorEstHandler` and `m_EstModel` in the `DerivativeBuilder` that block implementing static scheduling in error estimation.

### DIFF
--- a/lib/Differentiator/EstimationModel.cpp
+++ b/lib/Differentiator/EstimationModel.cpp
@@ -9,8 +9,10 @@
 #include "clang/AST/DeclarationName.h"
 #include "clang/AST/Expr.h"
 #include "clang/AST/OperationKinds.h"
+#include "clang/AST/Type.h"
 #include "clang/Sema/DeclSpec.h"
 #include "clang/Sema/Lookup.h"
+#include "clang/Sema/TemplateDeduction.h"
 
 #include "llvm/ADT/APFloat.h"
 #include "llvm/ADT/SmallVector.h"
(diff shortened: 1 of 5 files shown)


lib/Differentiator/EstimationModel.cpp
```diff
@@ -13,31 +13,57 @@ using namespace clang;
...
+void FPErrorEstimationModel::LookupCustomErrorFunction() {
+  CXXScopeSpec SS;
+  NamespaceDecl* cladNS =
+      utils::LookupNSD(m_Sema, "clad", /*shouldExist=*/true);
+  SS.Extend(m_Context, cladNS, noLoc, noLoc);
+  IdentifierInfo* II = &m_Context.Idents.get("getErrorVal");
+  DeclarationNameInfo DNInfo(DeclarationName(II), utils::GetValidSLoc(m_Sema));
+  LookupResult R(m_Sema, DNInfo, Sema::LookupOrdinaryName);
+  m_Sema.LookupQualifiedName(R, cladNS);
+  if (R.empty())
```
vgvassilev, line 39: We need a proper diagnostic explaining what needs to be done. Perhaps we will need a default implementation but let's see if the diagnostics can be done in a way that we don't need default implementation.

PR #1659, opened 2025-11-26
Test error estimation numerically

PetroZarytskyi:
Currently, we only check the code produced in error estimation. The produced errors are never checked. This makes it hard to understand whether certain changes break error estimation or not. This PR introduces numerical checks for all error estimation tests.
One of the reasons errors were not tested was that they are usually of order `10^-7`. This is because all errors get multiplied by `std::numeric_limits<float>::epsilon()`, which is equal to `2^-23 = 1.19e-7`. This makes it inconvenient to compare the results with theoretical predictions. This PR adds `TEST_ERROR_ESTIMATION` that normalizes the error by `std::numeric_limits<float>::epsilon()` before printing it. Since dividing by such small numbers could impact precission, the normalization is done as `std::ldexp(error, std::numeric_limits<PrecissionType>::digits - 1)`. `std::ldexp` lets us change the exponent of the float directly, and multiply the number by `2^23` without changing the mantissa. `PrecissionType` is templated so that we can test other floating types in the future.

### DIFF
--- a/test/ErrorEstimation/Assignments.C
+++ b/test/ErrorEstimation/Assignments.C
@@ -1,7 +1,9 @@
 // RUN: %cladclang -I%S/../../include -oAssignments.out %s 2>&1 | %filecheck %s
-// RUN: ./Assignments.out
+// RUN: ./Assignments.out | %filecheck_exec %s
+// XFAIL: valgrind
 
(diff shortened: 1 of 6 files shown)


vgvassilev (approved):
LGTM!

PR #1661, opened 2025-11-29
Fix #1539 Use parallelism in clad::restore_tracker 

### LINKED ISSUE #1539: Use parallelism in clad::restore_tracker
``clad::restore_tracker`` is a class for low-level storing/restoring variables inside ``reverse_forw`` in the reverse mode.
The main function for restoring looks like this:
```
  void restore() {
    for (...& pair : m_data) {
      // restoring
    }
    m_data.clear();
  }
```
Since ``clad::restore_tracker`` should only be used for storing basic types (no arrays/objects), no restored memory intervals should overlap. Therefore, we should make the loop parallel.

Lucky-Lodhi2004:
#### Description

This PR fixes issue #1539. 
I have used c++ 17 feature `std::execution::par` to make execution of loop parellel. For efficient execution of threads I have used separate array. This was existing function so no tests were written.

#### Any external dependency introduced?

I have used standard c++ includes so no new dependency

### DIFF
--- a/include/clad/Differentiator/RestoreTracker.h
+++ b/include/clad/Differentiator/RestoreTracker.h
@@ -43,25 +43,26 @@ public:
   }
   // Set all stored addresses to the corresponsing values bitwise.
   void restore() {
-    // Copy pointers to a vector for efficient parallel iteration
+    // store the map to a vector of pointers for efficient execution
     std::vector<std::pair<Address, RawMemory*>> work;
     work.reserve(m_data.size());
     for (auto& kv : m_data) {
         work.emplace_back(kv.first, &kv.second);
     }
 
-    // Parallel restore
+    // Parallel restore 
     std::for_each(std::execution::par, work.begin(), work.end(),
         [](auto& entry) {
-            Address addr = entry.first;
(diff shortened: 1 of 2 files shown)


include/clad/Differentiator/RestoreTracker.h
```diff
@@ -41,10 +43,23 @@ class restore_tracker {
...
-      std::memcpy(pair.first, buffer.data(), buffer.size());
+    // Copy pointers to a vector for efficient parallel iteration
+    std::vector<std::pair<Address, RawMemory*>> work;
+    work.reserve(m_data.size());
+    for (auto& kv : m_data) {
+        work.emplace_back(kv.first, &kv.second);
     }
+
+    // Parallel restore
+    std::for_each(std::execution::par, work.begin(), work.end(),
```
parth-07, line 54: Can't we directly use `for_each` on `m_data` instead of copying the elements to a vector?

Lucky-Lodhi2004: Seems like github ci doesn't support <execution> (c++ 17)

Issue #958, opened 2024-06-23
Unnecessary identifier in VarDecl that starts with "_"


ovdiiuv:
Differentiating
```
double fn(double x){
    double __b = x;
    return __b;
}
```

produces:

```
double fn_darg0(double x) {
    double _d_x = 1;
    double _d___b = _d_x;
    double __b0 = x;
    return _d___b;
}
```

PR #1668, opened 2025-12-03
Removed unnecessary underscores in generated variable names

Vedant2005goyal:
This PR closes issue https://github.com/vgvassilev/clad/issues/958. I updated the "ReverseModeVisitor.cpp" to strip leading underscores from original variables preventing messy prefixes.
Ex: _d___val -> _d_val.
Added a test by name CleanName.C in Features, also made changes in the tests in the test suite so that they now rely on new clean names.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -433,9 +433,9 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
           // If the type is not a tensor, we can use zero initialization.
           initExpr = getZeroInit(VDDerivedType);
         }
-        auto* VDDerived =
-            BuildGlobalVarDecl(VDDerivedType, "_d_" + param->getNameAsString(),
-                               initExpr, isDirectInit);
+        auto* VDDerived = BuildGlobalVarDecl(
+            VDDerivedType, "_d_" + param->getName().ltrim('_').str(), initExpr,
+            isDirectInit);
         m_Variables[param] = BuildDeclRef(VDDerived);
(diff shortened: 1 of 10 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -432,9 +432,10 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
           // If the type is not a tensor, we can use zero initialization.
           initExpr = getZeroInit(VDDerivedType);
         }
-        auto* VDDerived =
-            BuildGlobalVarDecl(VDDerivedType, "_d_" + param->getNameAsString(),
-                               initExpr, isDirectInit);
+        llvm::StringRef Name = param->getName();
+        std::string CleanName = Name.ltrim('_').str();
+        auto* VDDerived = BuildGlobalVarDecl(VDDerivedType, "_d_" + CleanName,
+                                             initExpr, isDirectInit);
```
```suggestion
        auto* VDDerived = BuildGlobalVarDecl(VDDerivedType, "_d_" + Name.ltrim('_'),
                                             initExpr, isDirectInit);
```

test/Features/CleanName.C
```diff
@@ -0,0 +1,34 @@
+// RUN: %cladclang -std=c++17 -I%S/../../include %s -o %t
```
vgvassilev: Does that test actually increase coverage? I think this is pretty well covered by the other tests maybe we do not need it.

Vedant2005goyal: I've removed the test

vgvassilev (approved):
LGTM. In future we want to have more meaningful identifiers which capture some of the type information or the previous name of the forward variable. That would make the code more readable.

Issue #1666, opened 2025-12-02
[cmake] Improve the compiler swap for the clad benchmarks


vgvassilev:
When running benchmarks clad changes the compiler with the compiler it was built against in:
https://github.com/vgvassilev/clad/blob/44d4794e955f5613cef9075c35ca7f3b09b598d6/CMakeLists.txt#L329-L352

We should investigate that if we create a new target for the benchmarks we won't have to store and restore the compilers or the compilation flags. Something like:
```cmake
function(clad_apply_test_compiler target)
    target_compile_options(${target} PRIVATE
        -Wno-class-memaccess
        -Wno-undefined-inline
        -fno-lifetime-dse
    )

    target_compile_options(${target} PRIVATE
        $<$<NOT:$<CXX_COMPILER_ID:Clang>>:-flags>
    )

    target_compile_definitions(${target} PRIVATE CLAD_TESTING=1)

    set_property(TARGET ${target} PROPERTY CXX_COMPILER ${LLVM_TOOLS_BINARY_DIR}/clang++)
endfunction()
```

Shreyansh2025:
Hi @vgvassilev, I'm Shreyansh, a 3rd-year CS student. I've analyzed the current global compiler swap in CMakeLists.txt and the previous PR attempts. I'd like to implement the clad_apply_test_compiler function approach you suggested to keep the benchmark targets isolated without global state changes. May I take this?

shishir-kuet:
Hi @vgvassilev, I’d like to work on this if it’s still available. I plan to replace the global compiler switching with a target-based approach using a helper function applied to benchmark targets, as suggested in the issue. Happy to proceed or coordinate if someone is already working on it.

PR #1671, opened 2025-12-10
Fixes: #1666 Created new targets for benchmarks

Aditi571:
This PR aims to resolve the issue #1666 by changing the target flags for benchmarks only instead of changing compilers or the compilation flags globally.

**Key changes :** 
1. cmake/ClangBenchmark.cmake : Contains the code of function that forces a given target to be built with Clang.
2. benchmark/CMakeLists.txt : Applied Clang-specific flags and definitions for all benchmarks.
3. CMakeLists.txt : Removed the code of changing compilers or the compilation flags globally for benchmarks and included ClangBenchmark.


Aditi571:
Hi @vgvassilev , I’ve attempted an implementation of the benchmark-related hack to avoid the global compiler switch. I’d appreciate it if you could take a look and share any feedback or suggestions for improvement.

guitargeek:
It's still relevant as #1666 is not closed yet, but there is a more up-to-date PR #1798 that attempts to close the same issue. I propose to close the PR here and move the discussion to #1666 or #1798. @Aditi571, please feel free to open a fresh PR if you think you can cleanly solve the issue on current `master`!

PR #1672, opened 2025-12-10
Improve the safety of numerical differentiation

PetroZarytskyi:
This PR addresses 3 problems:
1) Numerical differentiation is allowed on functions that don't have reference parameters. Such restriction is not sufficient because our num-diff infrastructure only supports functions with real parameters and return type, the rest is undefined behaviour. This PR restricts num-diff to real functions in all modes and adds diagnostics if the function is not viable.
2) Numerical differentiation is only used on undefined functions. However, our diagnostics only says `"function was not differentiated because clad failed to differentiate it..."`. This PR changes the warning to explicit `"attempted differentiation of function without definition..."`.
3) All of the diagnostics was moved from `BFMV::VisitCallExpr` and `RMV::VisitCallExpr` to a central point in `DerivativeBuilder::Derive`.
4) We didn't produce diagnostics for undefined functions that are not eligible for numerical differentiation.

### DIFF
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -16,6 +16,7 @@
 #include "clang/AST/RecursiveASTVisitor.h"
 #include "clang/AST/StmtVisitor.h"
 #include "clang/Basic/Diagnostic.h"
+#include "clang/Lex/Preprocessor.h"
 #include "clang/Sema/Sema.h"
 
 #include <array>
@@ -196,6 +197,15 @@ namespace clad {
     /// \param[in] Request The request to be processed.
     /// \returns The derivative function if found, nullptr otherwise.
     clang::FunctionDecl* HandleNestedDiffRequest(DiffRequest& request);
+
+    /// Emits diagnostic messages on differentiation (or lack thereof) for
+    /// an function without a definition.
+    ///
+    /// \param[in] \c FD - The function declaration.
+    /// \param[in] \c srcLoc Any associated source location information.
+    void diagnoseUndefinedFunction(const clang::FunctionDecl* FD,
+                                   clang::SourceLocation srcLoc,
+                                   bool numDiffViable);
   };
 
 } // end namespace clad
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -425,6 +425,37 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
     return derivative;
   }
 
+  void
+  DerivativeBuilder::diagnoseUndefinedFunction(const clang::FunctionDecl* FD,
+                                               SourceLocation srcLoc,
+                                               bool numDiffViable) {
+    bool NumDiffEnabled =
+        !m_Sema.getPreprocessor().isMacroDefined("CLAD_NO_NUM_DIFF");
+    diag(DiagnosticsEngine::Warning, srcLoc,
+         "attempted differentiation of function %0 without definition "
+         "and no suitable overload was found in "
+         "namespace 'custom_derivatives'")
+        << FD << srcLoc;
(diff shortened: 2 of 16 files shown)


include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -16,6 +16,7 @@
 #include "clang/AST/RecursiveASTVisitor.h"
 #include "clang/AST/StmtVisitor.h"
 #include "clang/Basic/Diagnostic.h"
+#include "clang/Lex/Preprocessor.h"
 #include "clang/Sema/Sema.h"
```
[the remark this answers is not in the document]
PetroZarytskyi: That's not true, it doesn't compile without it.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -425,6 +425,37 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
     return derivative;
   }
 
+  void
+  DerivativeBuilder::diagnoseUndefinedFunction(const clang::FunctionDecl* FD,
+                                               SourceLocation srcLoc,
+                                               bool numDiffViable) {
+    bool NumDiffEnabled =
+        !m_Sema.getPreprocessor().isMacroDefined("CLAD_NO_NUM_DIFF");
```
vgvassilev, line 433: Isn’t it better to define that as an invocation argument and a clad::option rather than having it as a macro?

PetroZarytskyi: I totally agree. Do you think we should address this in the same PR? Perhaps we could do it together making num diff off by default.

vgvassilev: Sure...

test/CUDA/GradientKernels.cu
```diff
@@ -476,7 +474,8 @@ double fn_memory(double *out, double *in) {
   cudaMemcpy(out_host, out, 10 * sizeof(double), cudaMemcpyDeviceToHost);
   double res = 0;
   for (int i=0; i < 10; ++i) {
-    printf("Writing result of out[%d]\n", i);
+    printf("Writing result of out[%d]\n", i); // expected-warning {{attempted differentiation of function 'printf' without definition and no suitable overload was found in namespace 'custom_derivatives'}}
```
vgvassilev, line 477: How we should be expressing these as non-differentiable to avoid such warnings?

test/FirstDerivative/DiffInterface.C
```diff
@@ -171,8 +171,8 @@ int main () {
   float one = 1.0;
   clad::differentiate(f_2, one); // expected-error {{failed to parse the parameters, must be string or numeric literal}}
 
-  clad::differentiate(f_no_definition, 0); // expected-error {{attempted differentiation of function 'f_no_definition', without definition}}
-
+  clad::differentiate(f_no_definition, 0); // expected-warning {{attempted differentiation of function 'f_no_definition' without definition and no suitable overload was found in namespace 'custom_derivatives'}}
```
vgvassilev, line 174: In a separate PR we should probably change the diagnostic saying, we cannot differentiate this, considering it as non-differentiable. If you want to provide a custom derivative, use that signature <and we create the expected signature> that people can paste in their code.

test/CUDA/GradientKernels.cu
PetroZarytskyi: Right now, the only way is to define a custom derivative

PetroZarytskyi: In other words, in the reverse mode, the user can define an empty pullback. In the forward mode, currently, the only way is defining a pushforward that calls the original function internally.

vgvassilev: Let's track this as a separate issue and move forward.

PetroZarytskyi: I opened #1698.

PR #1674, opened 2025-12-11
[Derivatives] Implement forward-mode derivative for std::comp_ellint_1

### LINKED ISSUE #1673: [Derivatives] Missing symbolic derivative for C++17 std::comp_ellint_1
### Description
Clad currently lacks the symbolic derivative rule for the C++17 special math function `std::comp_ellint_1` (Complete Elliptic Integral of the First Kind).

### Current Behavior
When differentiating code that uses this function, Clad fails to find a suitable overload and falls back to numerical differentiation. This results in the following warning during compilation:

> warning: falling back to numerical differentiation for 'comp_ellint_1' since no suitable overload was found

### Expected Behavior
Clad should support the analytical derivative of $K(k)$ using the standard recurrence relations involving the Second Kind Integral ($E(k)$).

The formula for the derivative is:
$$\frac{d}{dk} K(k) = \frac{E(k) - (1-k^2)K(k)}{k(1-k^2)}$$

### Proposed Solution
I intend to implement the forward-mode pushforward for this function in `BuiltinDerivatives.h`, utilizing `std::comp_ellint_2` for the $E(k)$ term.

ayush4874:
## Summary
This PR implements the exact symbolic differentiation rule for `std::comp_ellint_1` (Complete Elliptic Integral of the First Kind), resolving the fallback to numerical differentiation.

Fixes #1673

## Mathematical Context
The derivative of the First Kind Elliptic Integral $K(k)$ is defined using the Second Kind Integral $E(k)$ as:

$$\frac{dK(k)}{dk} = \frac{E(k) - (1-k^2)K(k)}{k(1-k^2)}$$

where:
* $K(k)$ maps to `std::comp_ellint_1(k)`
* $E(k)$ maps to `std::comp_ellint_2(k)`

## Implementation Details
* **File Modified:** `clad/Differentiator/BuiltinDerivatives.h`
* **Changes:**
    * Implemented `comp_ellint_1_pushforward` in the `clad::custom_derivatives` namespace.
    * Added the corresponding `using` declaration to register the function in the global lookup scope.
    * The implementation handles the C++17 standard library calls for elliptic integrals.

## Verification
I tested the implementation locally by differentiating a wrapper function for `std::comp_ellint_1`.

**1. Compilation Check:**
The compilation warning `falling back to numerical differentiation` is **gone**, confirming Clad is now using the symbolic rule.

**2. Numerical Accuracy:**
I verified the derivative output at $k = 0.5$:
* **Analytical Value:** `0.541733...`
* **Clad Output:** `0.541732`

This confirms the implementation produces the correct gradients for physics applications.

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -1213,6 +1213,85 @@ CUDA_HOST_DEVICE void beta_pullback(T x, T y, U d_z, T* d_x, T* d_y) {
 }
 #endif
 
+#if __cplusplus >= 201703L && (defined(__cpp_lib_math_special_funcs) ||        \
+                               defined(__STDCPP_MATH_SPEC_FUNCS__))
+template <typename T, typename dT,
+          typename T_out = decltype(::std::comp_ellint_1(T())),
+          typename dT_out = typename AdjOutType<T_out, dT>::type>
+CUDA_HOST_DEVICE ValueAndPushforward<T_out, dT_out>
+comp_ellint_1_pushforward(T k, dT d_k) {
+  T_out K = ::std::comp_ellint_1(k);
+  T_out E = ::std::comp_ellint_2(k);
+  T_out k_sq = k * k;
+  T_out one = 1.0;
+  T_out derivative = 0.0;
+  if (k_sq < 1.0 && k != 0.0)
+    derivative = (E - (one - k_sq) * K) / (k * (one - k_sq));
+  return {K, static_cast<dT_out>(d_k) * derivative};
+}
+
+template <typename T, typename dT,
+          typename T_out = decltype(::std::comp_ellint_2(T())),
+          typename dT_out = typename AdjOutType<T_out, dT>::type>
+CUDA_HOST_DEVICE ValueAndPushforward<T_out, dT_out>
+comp_ellint_2_pushforward(T k, dT d_k) {
+  T_out K = ::std::comp_ellint_1(k);
+  T_out E = ::std::comp_ellint_2(k);
+  T_out derivative = 0.0;
+  if (k != 0.0)
+    derivative = (E - K) / k;
+  return {E, static_cast<dT_out>(d_k) * derivative};
+}
+
+template <typename T1, typename T2, typename dT1, typename dT2,
+          typename T_out = decltype(::std::comp_ellint_3(T1(), T2())),
+          typename dT_out = typename AdjOutType<T_out, dT1>::type>
+CUDA_HOST_DEVICE ValueAndPushforward<T_out, dT_out>
+comp_ellint_3_pushforward(T1 k, T2 nu, dT1 d_k, dT2 d_nu) {
+  T_out K = ::std::comp_ellint_1(k);
+  T_out E = ::std::comp_ellint_2(k);
+  T_out Pi = ::std::comp_ellint_3(k, nu);
+  T_out k2 = k * k;
+  T_out one = 1.0;
+  T_out grad_k = 0.0;
+  if (k2 != static_cast<T_out>(nu) && k != 0.0 && k2 < 1.0) {
+    T_out term_k = E / (one - k2);
+    grad_k = (k / (k2 - static_cast<T_out>(nu))) * (term_k - Pi);
+  }
+  T_out grad_nu = 0.0;
+  if (nu != 0.0 && nu != 1.0 && k2 != static_cast<T_out>(nu)) {
+    T_out p2 = ((k2 - static_cast<T_out>(nu)) / static_cast<T_out>(nu)) * K;
+    T_out p3 = ((static_cast<T_out>(nu) * static_cast<T_out>(nu) - k2) /
+                static_cast<T_out>(nu)) *
+               Pi;
+    grad_nu = (one / (2.0 * (static_cast<T_out>(nu) - one) *
+                      (k2 - static_cast<T_out>(nu)))) *
+              (E + p2 + p3);
+  }
+  return {Pi, (static_cast<dT_out>(d_k) * grad_k) +
+                  (static_cast<dT_out>(d_nu) * grad_nu)};
+}
+
+template <typename T1, typename T2, typename T3>
+CUDA_HOST_DEVICE void comp_ellint_3_pullback(T1 k, T2 nu, T3 d_out, T1* d_k,
+                                             T2* d_nu) {
+  auto K = ::std::comp_ellint_1(k);
+  auto E = ::std::comp_ellint_2(k);
+  auto Pi = ::std::comp_ellint_3(k, nu);
+  auto k2 = k * k;
+  auto one = 1.0;
+  if (k2 != nu && k != 0.0 && k2 < 1.0) {
+    auto term_k = E / (one - k2);
+    *d_k += d_out * ((k / (k2 - nu)) * (term_k - Pi));
+  }
+  if (nu != 0.0 && nu != one && k2 != nu) {
+    auto p2 = ((k2 - nu) / nu) * K;
+    auto p3 = ((nu * nu - k2) / nu) * Pi;
+    *d_nu += d_out * ((one / (2.0 * (nu - one) * (k2 - nu))) * (E + p2 + p3));
+  }
+}
+#endif
+
 } // namespace std
 
 CUDA_HOST_DEVICE inline ValueAndPushforward<float, float>
@@ -1487,6 +1566,14 @@ using std::sqrt_pushforward;
 #if __cplusplus >= 201703L
 using std::beta_pullback;
 using std::beta_pushforward;
+#endif
+
+#if __cplusplus >= 201703L && (defined(__cpp_lib_math_special_funcs) ||        \
+                               defined(__STDCPP_MATH_SPEC_FUNCS__))
+using std::comp_ellint_1_pushforward;
+using std::comp_ellint_2_pushforward;
+using std::comp_ellint_3_pullback;
+using std::comp_ellint_3_pushforward;
 #endif
 
 namespace class_functions {
--- a/test/Features/stl-cmath.cpp
+++ b/test/Features/stl-cmath.cpp
@@ -99,9 +99,9 @@
 // D assoc_laguerre / f / l    (C++17) associated Laguerre polynomials
 // D assoc_legendre/ f / l     (C++17) associated Legendre polynomials
 // DS beta/ betaf/ betal        (C++17) beta function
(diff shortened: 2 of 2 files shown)


include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -1323,6 +1349,7 @@ using std::min_pushforward;
 using std::pow_pullback;
 using std::pow_pushforward;
 using std::sqrt_pushforward;
+using std::comp_ellint_1_pushforward;
```
vgvassilev, line 1573: We will need a pullback for this and also tests.

ayush4874: Thanks for the review!

I replaced the manual pullback with `comp_ellint_1_darg0` (derivative w.r.t arg 0). This allows Clad to automatically derive the reverse-mode pullback, resolving the signature ambiguity.

I also added a regression test (`clad/test/Differentiator/EllipticTest.cpp`) and verified the gradients against the analytical values locally.

**Verification Output:**
```text
Forward Derivative: 0.541732
Reverse Derivative: 0.541732

File: include/clad/Differentiator/BuiltinDerivatives.h [hunk 2]
```diff
@@ -1059,6 +1059,52 @@ CUDA_HOST_DEVICE void hypot_pullback(T x, T y, U d_z, T* d_x, T* d_y) {
   *d_y += (y / h) * d_z;
 }
 
+// -------------------------------------------------------------------------
+// [Ayush GSoC 2026 Contribution]
+// Feature: Added support for Complete Elliptic Integral of the First Kind
+// Target: std::comp_ellint_1(k)
+// Formula: dK/dk = (E(k) - (1-k^2)*K(k)) / (k * (1-k^2))
```
vgvassilev: Remove the redundant comments.

test/Differentiator/EllipticTest.cpp
```diff
@@ -0,0 +1,33 @@
+// RUN: %clangxx -fplugin=%clad_plugin -std=c++17 %s -o %t
```
vgvassilev: This test should be part of `test/Features/stl-cmath.cpp`

ayush4874: Done. I've moved the test case into `test/Features/stl-cmath.cpp` and removed the verbose comments from the header file.

test/Features/stl-cmath.cpp
```diff
@@ -295,6 +295,10 @@ DEFINE_FUNCTIONS(atanh) // x in [-1,1]
 //
 DEFINE_FUNCTIONS(erf)  // x in (-inf,+inf)
 
+double comp_ellint_1_wrapper(double k) {
```
vgvassilev: You need to make this look the same as the other functions.

ayush4874: Done. Since `comp_ellint_1` is a C++17 addition and does not have a legacy C equivalent (like `comp_ellint_1f`), I could not use the `DEFINE_FUNCTIONS` macro directly.

However, I have manually defined a templated wrapper `f_comp_ellint_1` that follows the exact same naming convention and pattern as the macro, ensuring consistency with the rest of the file.

test/Features/stl-cmath.cpp
```diff
@@ -295,9 +295,9 @@ DEFINE_FUNCTIONS(atanh) // x in [-1,1]
 //
 DEFINE_FUNCTIONS(erf)  // x in (-inf,+inf)
 
-double comp_ellint_1_wrapper(double k) {
-  return std::comp_ellint_1(k);
-}
+// Manually defined to match DEFINE_FUNCTIONS pattern (comp_ellint_1f does not exist)
```
vgvassilev: Still does not look as the surroundings. Please make sure you understand this file and its comments.

ayush4874: My apologies for the oversight. I realized upon closer inspection that `DEFINE_FUNCTIONS` generates suffix variants (`f_...f`, `f_...l`) that my previous template missed.

I have now added `f_comp_ellint_1f` and `f_comp_ellint_1l` manually to fully replicate the macro's behavior as requested.

test/Features/stl-cmath.cpp
```diff
@@ -295,6 +295,12 @@ DEFINE_FUNCTIONS(atanh) // x in [-1,1]
 //
 DEFINE_FUNCTIONS(erf)  // x in (-inf,+inf)
 
+template <typename T>
+T f_comp_ellint_1(T x) { return std::comp_ellint_1(x); }
+
+inline float f_comp_ellint_1f(float x) { return std::comp_ellint_1(x); }
+inline long double f_comp_ellint_1l(long double x) { return std::comp_ellint_1(x); }
```
vgvassilev: Why we cannot use DEFINE_FUNCTIONS here? Note this is missing the pullbacks and updating the section with the proper comment `// ---- special fns`. Also misses updating the table in the beginning of the file. Misses the range...

As I've said -- needs to look like the other functions tests in the same file.

ayush4874: I have updated the file to fully match the surrounding style and structure as requested:

1. Added `comp_ellint_1` to the feature table at the top of the file.
2. Moved the implementation to the "Special Functions" section with the `// k in (-1, 1)` range comment.
3. Defined `comp_ellint_1f` and `comp_ellint_1l` wrappers so I can use `DEFINE_FUNCTIONS(comp_ellint_1)` directly.
4. Updated the test case in `main()` to use the standard `CHECK_ALL_RANGE` macro.

ayush4874: @vgvassilev it's ready for review, can you have a look?

vgvassilev: Why adding only this special function and not the rest?

ayush4874: I initially isolated `comp_ellint_1` to ensure the testing pattern for special functions was correct before expanding.

Since `comp_ellint_2` (Complete Elliptic Integral of the Second Kind) is actually required for the derivative of `comp_ellint_1,` it is logical to group them. I will update this PR to include `std::comp_ellint_2` and `std::comp_ellint_3` so they can be reviewed together. I'll push those changes shortly.

ayush4874: Done, can you have a look?

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -1059,6 +1059,59 @@ CUDA_HOST_DEVICE void hypot_pullback(T x, T y, U d_z, T* d_x, T* d_y) {
   *d_y += (y / h) * d_z;
 }
 
+CUDA_HOST_DEVICE inline double comp_ellint_1_darg0(double k) {
```
vgvassilev: Do we need the `_darg` versions here?

ayush4874: You're right, we don't need them. I've removed the `_darg` helpers and moved the derivative logic directly inside the `pushforward` templates.

test/Features/stl-cmath.cpp
```diff
@@ -295,6 +300,40 @@ DEFINE_FUNCTIONS(atanh) // x in [-1,1]
 //
 DEFINE_FUNCTIONS(erf)  // x in (-inf,+inf)
 
+//------------------------ Elliptic integrals -----------------------------
+//
+// Helper wrappers to enable DEFINE_FUNCTIONS for comp_ellint_1
+inline float comp_ellint_1f(float x) { return std::comp_ellint_1(x); }
+inline long double comp_ellint_1l(long double x) { return std::comp_ellint_1(x); }
+
+DEFINE_FUNCTIONS(comp_ellint_1)
```
vgvassilev: Can you add the domain of x?

test/Features/stl-cmath.cpp
```diff
@@ -295,6 +300,40 @@ DEFINE_FUNCTIONS(atanh) // x in [-1,1]
...
+  auto d_ellint3_k = clad::differentiate(std::comp_ellint_3, 0);
+  double res_k = d_ellint3_k.execute(k, nu);
+
+  // Test differentiation w.r.t 'nu' (Arg 1)
+  auto d_ellint3_nu = clad::differentiate(std::comp_ellint_3, 1);
+  double res_nu = d_ellint3_nu.execute(k, nu);
+  
+  // Simple check to ensure we got numbers back (prevents unused var warning)
+  (void)res_k; 
+  (void)res_nu;
```
vgvassilev: Here we should compare against the numerical, right? See what the CHECK macro does.

test/Features/stl-cmath.cpp
ayush4874: Done. I've added comments specifying the domain is k in (-1, 1) for these integrals.

test/Features/stl-cmath.cpp
ayush4874: You are right. I updated the test to compare the symbolic result against a Finite Difference numerical approximation. The test will now fail (return 1) if the derivative deviates from the numerical reference.

test/Features/stl-cmath.cpp
vgvassilev: look how was done for the other functions.

test/Features/stl-cmath.cpp
```diff
@@ -306,31 +306,43 @@ DEFINE_FUNCTIONS(erf)  // x in (-inf,+inf)
...
-  
+  double h = 1e-8;
+  double tol = 1e-5;
+
   // Test differentiation w.r.t 'k' (Arg 0)
   auto d_ellint3_k = clad::differentiate(std::comp_ellint_3, 0);
-  double res_k = d_ellint3_k.execute(k, nu);
+  double sym_k = d_ellint3_k.execute(k, nu);
+  // Numerical approximation: (f(k+h) - f(k-h)) / 2h
+  double num_k = (std::comp_ellint_3(k + h, nu) - std::comp_ellint_3(k - h, nu)) / (2 * h);
```
vgvassilev: There is an api in this file that gets you the numerical part.

ayush4874: I see the pattern now. I updated the tests to match `atan2` and `pow` by using the `f_` wrappers.

For `comp_ellint_3,` I fixed the second arg in the wrapper so I can just use `CHECK_ALL_RANGE,` which handles the numerical verification automatically.

vgvassilev (approved):
LGTM!

test/Features/stl-cmath.cpp
vgvassilev, line 6: We need to update this table at the relevant place, too.

ayush4874: Updated the table. The elliptic integrals were already listed under the Mathematical special functions section, but I changed their classification from D to DS to match beta.

PR #1678, opened 2025-12-11
Fixes #1666

### LINKED ISSUE #1666: [cmake] Improve the compiler swap for the clad benchmarks

Vedant2005goyal:
I have removed the global compiler change logic (ie the store and restore of compilers and flags) from the root CMakeLists.txt and rather the compiler configuration  is isolated now within the directory specific CMakeLists.txt files for benchmarks, unittests and test


CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ endforeach()
 
 # Generate CMakeArgs.txt file with source, build dir and command line args
 write_file("${CMAKE_CURRENT_BINARY_DIR}/CMakeArgs.txt"
-  "-S${CMAKE_SOURCE_DIR} -B${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_ARGS}")
+           "-S${CMAKE_SOURCE_DIR} -B${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_ARGS}")
```
vgvassilev, line 24: Seems like a stray change.

CMakeLists.txt
```diff
@@ -231,17 +231,17 @@ endif()
...
 if( CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR AND NOT MSVC_IDE )
   message(FATAL_ERROR "In-source builds are not allowed. CMake would overwrite "
-"the makefiles distributed with LLVM. Please create a directory and run cmake "
-"from there, passing the path to this source directory as the last argument. "
-"This process created the file `CMakeCache.txt' and the directory "
-"`CMakeFiles'. Please delete them.")
+      "the makefiles distributed with LLVM. Please create a directory and run cmake "
+      "from there, passing the path to this source directory as the last argument. "
+      "This process created the file `CMakeCache.txt' and the directory "
+      "`CMakeFiles'. Please delete them.")
```
vgvassilev, line 244: Please undo the formatting on the lines that are not really changed.

benchmark/CMakeLists.txt
```diff
@@ -1,29 +1,55 @@
...
-CB_ADD_GBENCHMARK(AlgorithmicComplexity AlgorithmicComplexity.cpp)
-CB_ADD_GBENCHMARK(ArrayExpressionTemplates ArrayExpressionTemplates.cpp)
-if (CLAD_ENABLE_ENZYME_BACKEND)
-  CB_ADD_GBENCHMARK(EnzymeCladComparison EnzymeCladComparison.cpp)
+set(CMAKE_CXX_COMPILER ${LLVM_TOOLS_BINARY_DIR}/clang++)
+function(diff_compiler_test target)
+  target_compile_options(
+    ${target} PRIVATE  -Wno-undefined-inline)
+
+  target_compile_definitions(${target} PRIVATE CLAD_TESTING=1)
```
vgvassilev, line 9: Why we have `CLAD_TESTING` here defined?

benchmark/CMakeLists.txt
```diff
@@ -1,29 +1,55 @@
 set(CTEST_BUILD_NAME ${ROOT_ARCHITECTURE}-${CMAKE_BUILD_TYPE})
 enable_testing()
 
-CB_ADD_GBENCHMARK(Simple Simple.cpp)
-CB_ADD_GBENCHMARK(AlgorithmicComplexity AlgorithmicComplexity.cpp)
-CB_ADD_GBENCHMARK(ArrayExpressionTemplates ArrayExpressionTemplates.cpp)
-if (CLAD_ENABLE_ENZYME_BACKEND)
-  CB_ADD_GBENCHMARK(EnzymeCladComparison EnzymeCladComparison.cpp)
+set(CMAKE_CXX_COMPILER ${LLVM_TOOLS_BINARY_DIR}/clang++)
```
vgvassilev, line 4: Why do we need to change this explicitly here?

benchmark/CMakeLists.txt
```diff
@@ -1,29 +1,55 @@
...
+  cb_add_gbenchmark(EnzymeCladComparison EnzymeCladComparison.cpp)
 endif(CLAD_ENABLE_ENZYME_BACKEND)
-CB_ADD_GBENCHMARK(VectorModeComparison VectorModeComparison.cpp)
-CB_ADD_GBENCHMARK(MemoryComplexity MemoryComplexity.cpp)
-CB_ADD_GBENCHMARK(Multithreading Multithreading.cpp)
-CB_ADD_GBENCHMARK(Hessians Hessians.cpp)
+cb_add_gbenchmark(VectorModeComparison VectorModeComparison.cpp)
+cb_add_gbenchmark(MemoryComplexity MemoryComplexity.cpp)
+cb_add_gbenchmark(Multithreading Multithreading.cpp)
+cb_add_gbenchmark(Hessians Hessians.cpp)
```
vgvassilev: Undo the syntax changes here.

benchmark/CMakeLists.txt
Vedant2005goyal: Since we have removed the logic of altering the compiler globally we need to set up the compiler locally so that the benchmark continues to build with the different compiler

vgvassilev: The point of the fix was to avoid that altogether, right? We want to have a different compiler per target.

Vedant2005goyal: ok i got it, now i will try to work in that direction.

PR #1685, opened 2025-12-24
Fixes #1074

### LINKED ISSUE #1074: Use `VisitorBase::getZeroInit()` or `ConstantFolder::synthesizeLiteral()`  for assignments and initializations of variables

Vedant2005goyal:
I have configured the clad both reverse mode and the forward mode to use  'getZeroInit()' for initialization and assignments of variable. I have changed the test suite to align with the changes made. While running the tests locally with ASAN enabled  i encountered **use-after-poison** crash in clad/lib/Analyses/UsefulAnalyzer.cpp. And the fix of that problem (in the UsefulAnalyzer.cpp) is also made in this PR.
Below is the stack trace of the crash which has been resolved:-

Screenshot 2025-12-24 at 11 59 12 PM


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3408,7 +3415,24 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+      if (auto* DS = dyn_cast<DeclStmt>(DSClone)) {
+        for (auto* D : DS->decls()) {
+          if (auto* VD = dyn_cast<VarDecl>(D)) {
+            if (VD && VD->hasInit()) {
+              Expr* Init = VD->getInit();
+              if (Init && VD->getType()->isRealFloatingType()) {
+                if (auto* IL =
+                        dyn_cast<IntegerLiteral>(Init->IgnoreParenImpCasts())) {
+                  if (IL->getValue() == 0)
+                    VD->setInit(getZeroInit(VD->getType()));
```
vgvassilev, line 3459: Shouldn't this happen in DSClone when building the declaration statement?

Vedant2005goyal: Thanks for the review!
That makes sense handling the getZeroInit during the declaration is cleaner and more efficient way than patching in the later stages.
I will make the changes accordingly and push the update.

vgvassilev:
Why did we close these issues?

PR #1686, opened 2025-12-25
Fix #1376 - Allow redefining builtin custom derivatives

### LINKED ISSUE #1376: Allow redefining builtin custom derivatives
Consider,

```cpp

extern "C" int printf(const char*,...);

namespace clad::custom_derivatives::std {
CUDA_HOST_DEVICE ValueAndPushforward<T, dT> exp_pushforward(T x, dT d_x) {
  return {::std::exp(x), ::std::exp(x) * d_x};
}
}
double differentiable_code(double x) {
    return std::exp(x);
}

int main() {
  auto df = clad::gradient(differentiable_code);
  // Print the generated code to standard output.
  df.dump();
  // Declare the necessary variables.
  double x, d_x = 0;
  // Finally call execute on the generated code.
  df.execute(x, &d_x);
}
```

Currently this fails with:
```
<source>:7:45: error: redefinition of 'exp_pushforward'
    7 | CUDA_HOST_DEVICE ValueAndPushforward<T, dT> exp_pushforward(T x, dT d_x) {
      |                                             ^
/opt/compiler-explorer/clang-plugins/clad-trunk-clang-20.1.0/include/clad/Differentiator/BuiltinDerivatives.h:525:45: note: previous definition is here
  525 | CUDA_HOST_DEVICE ValueAndPushforward<T, dT> exp_pushforward(T x, dT d_x) {
      |                                             ^
```

We should allow users to provide their own logic.

Lucky-Lodhi2004:
I made changes for only one derivative i.e. `exp_pushforward`. Changes for other derivatives can be made after discussing from maintainers. 

Users can redefine builtin derivatives after my changes. 

```
 ~/repos/llvm-project/build/bin/clang++ -std=c++17 -I ./include/ -fplugin=./build/lib/clad.so example.cpp -o example
./example
The code is: 
void differentiable_code_grad(double x, double *_d_x) {
    {
        double _r0 = 0.;
        _r0 += 1 * clad::custom_derivatives::std::exp_pushforward(x, 1.).pushforward;
        *_d_x += _r0;
    }
}
```

#### Changes

Added control macro for the definition of std::exp_pushforward. 
here 
```C++
// 2.1 exp, expf, expl
// Allow disabling built-in derivative if user-defined derivative provided.
#ifndef CLAD_NO_BUILTIN_EXP
CUDA_HOST_DEVICE ValueAndPushforward<T, dT> exp_pushforward(T x, dT d_x) {
  return {::std::exp(x), ::std::exp(x) * d_x};
}
#endif
```

and here:
```C++
#ifndef CLAD_NO_BUILTIN_EXP
using std::exp_pushforward;  //disable built-in exp derivative
#endif
```

Similar macro for all the functions can be used (if maintainers approve).

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -510,6 +510,7 @@ CUDA_HOST_DEVICE ValueAndPushforward<T, dT> exp_pushforward(T x, dT d_x) {
   return {::std::exp(x), ::std::exp(x) * d_x};
 }
 #endif
+
 // pushforward for expf, expl
 template <typename T, typename dT>
 CUDA_HOST_DEVICE ValueAndPushforward<T, dT> expf_pushforward(T x, dT d_x) {
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -503,11 +503,13 @@ CUDA_HOST_DEVICE void fdiml_pullback(T x, T y, U d_z, T* d_x, T* d_y) {
 // 2. Exponential Functions
 
 // 2.1 exp, expf, expl
+// Allow disabling built-in derivative if user-defined derivative provided.
+#ifndef CLAD_NO_BUILTIN_EXP
 template <typename T, typename dT>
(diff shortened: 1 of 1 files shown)


guitargeek:
Issue #1376 closed by cfdfcf7f with a proper `clad::custom_derivatives::overrides` namespace. The PR's `#ifndef CLAD_NO_BUILTIN_EXP` one-function hack is superseded.

PR #1688, opened 2025-12-29
Parallelism in clad::restore_tracker using multithreading

Vedant2005goyal:
Enforced parallelism in restore_tracker by creating safe multi threads to carry out the process of restoring the stored primals without disturbing other threads and i have used async to manage threads automatically without any human intervention
#Fixes 1539

### DIFF
--- a/include/clad/Differentiator/RestoreTracker.h
+++ b/include/clad/Differentiator/RestoreTracker.h
@@ -4,7 +4,9 @@
 #include <cassert>
 #include <cstdint>
 #include <cstring>
+#include <iterator>
 #include <map>
+#include <thread>
 #include <utility>
 #include <vector>
 
@@ -40,11 +42,38 @@ public:
     m_data.emplace((char*)&val, std::move(buffer));
   }
   // Set all stored addresses to the corresponsing values bitwise.
-  void restore() {
-    for (std::pair<const Address, RawMemory>& pair : m_data) {
-      std::vector<uint8_t>& buffer = pair.second;
-      std::memcpy(pair.first, buffer.data(), buffer.size());
+  template <typename = void> void restore() {
+    size_t total_items = m_data.size();
+    if (total_items == 0)
+      return;
+
+    size_t avail_threads = std::thread::hardware_concurrency();
+    if (avail_threads == 0)
+      avail_threads = 1;
+    std::vector<std::thread> workers;
+    workers.reserve(avail_threads);
+
+    size_t task_per_worker = total_items / avail_threads;
+    size_t left_task = total_items % avail_threads;
+
+    auto block_start = m_data.begin();
+    for (size_t i = 0; i < avail_threads; i++) {
+      size_t curr_block_size = task_per_worker + (i < left_task ? 1 : 0);
+      if (curr_block_size == 0)
+        continue;
+
+      auto block_end = block_start;
+      std::advance(block_end, curr_block_size);
+
+      workers.emplace_back([block_start, block_end]() {
+        for (auto it = block_start; it != block_end; it++)
+          std::memcpy(it->first, it->second.data(), it->second.size());
+      });
+      block_start = block_end;
     }
+    for (auto& t : workers)
+      if (t.joinable())
+        t.join();
     m_data.clear();
   }
 };


include/clad/Differentiator/RestoreTracker.h
```diff
@@ -40,11 +42,38 @@ class restore_tracker {
...
+  template <typename = void> void restore() {
+    size_t total_items = m_data.size();
+    if (total_items == 0)
+      return;
+
+    size_t avail_threads = std::thread::hardware_concurrency();
+    if (avail_threads == 0)
+      avail_threads = 1;
+    std::vector<std::thread> workers;
+    workers.reserve(avail_threads);
```
parth-07, line 54: Would it be possible to directly use `llvm::parallelFor` here?

include/clad/Differentiator/RestoreTracker.h
```diff
@@ -40,11 +42,38 @@ class restore_tracker {
...
-  void restore() {
-    for (std::pair<const Address, RawMemory>& pair : m_data) {
-      std::vector<uint8_t>& buffer = pair.second;
-      std::memcpy(pair.first, buffer.data(), buffer.size());
+  template <typename = void> void restore() {
+    size_t total_items = m_data.size();
+    if (total_items == 0)
+      return;
+
+    size_t avail_threads = std::thread::hardware_concurrency();
```
vgvassilev, line 50: In practice this spawns `hardware_concurrency()` threads to copy one scalar. Since 5bea5283 a tracker inside a loop holds only that iteration's records: test/Gradient/ReverseForw.C shows one whose entire content is `_tracker0.store(*p)`, a single eight-byte value, and `restore()` runs twice per call, before the pullback and after it. Creating and joining an OS thread costs tens of microseconds; the `memcpy` it is replacing costs a few nanoseconds. There is no size threshold, so the small case pays the full cost, and the small case is every case clad currently generates. benchmark/ReplayFreePullbacks.cpp already drives this exact protocol in a loop and would give a number. If a number ever justifies parallel restore, the shape to measure is a pool created once, not a thread per call.

vgvassilev (changes requested):
Parallelising this loop makes a class the codebase already treats as unsafe to share less safe, not more. lib/Differentiator/CladUtils.cpp carries `// FIXME: clad::restore_tracker is not thread-safe.` directly above the check that turns `reverse_forw` off for CUDA. Meanwhile clad does emit parallel reverse sweeps: test/Gradient/OpenMP.C shows the generated gradient wrapping the reverse loop in `#pragma omp parallel`. A variable declared outside a parallel region is shared by the threads inside it, and the tracker for a call outside a loop is created by `GlobalStoreImpl`, which, unlike `MakeCladTapeFor`, does not mark the declaration `threadprivate`. If two OpenMP threads reach the same tracker on master, they lose an update. With this change, one thread's `m_data.clear()` can run while another thread's workers are still dereferencing iterators into that same map, which is a use-after-free rather than a lost write. That is the shape of issue #2114: a reverse sweep parallelised over shared state with nothing synchronising it, giving a wrong gradient and no diagnostic. Being honest about what I checked: I read the emission code and the OpenMP test, but I could not find an existing test that puts a `reverse_forw` call inside an OpenMP loop, so I cannot say that combination is reachable today. Nothing in clad prevents it, though, and this change would make the consequence worse. Before parallel restore can be considered, the ownership question has to be settled: who owns a tracker, and can two threads ever be inside one at the same time.

Separately, the parallelism inside a single `restore()` is only sound while no two stored byte ranges overlap. Issue #1539 states that as a given, but `store()` is a template over any `T` and nothing enforces it. If that invariant is what makes the change safe, it belongs in `store()` as an assert, not in an issue description.

The branch is also 313 commits behind master and the protocol it edits has changed. 5bea5283 made `restore()` deliberately non-destructive, and the generated code now calls it twice around every pullback: once so the pullback's forward replay starts from the pre-call state, once because that replay has just re-mutated it. The `m_data.clear()` at the end of this `restore()` would make the second call a no-op, so every reverse-sweep consumer ahead of the call reads post-call values. That is a silently wrong gradient. 3adaa8aa added a `clear()` member the generated code calls, and cf649273 added an `#ifdef __CUDACC__` branch; that branch's `restore()` is `__host__ __device__`, so whatever lands has to keep a device path that does not touch `std::thread`. The rebase is a rewrite, not a merge. CI green here means less than it looks: the last real build on this head ran on 2026-05-09, three months before 5bea5283, and the branch is now `dirty`.

On the `llvm::parallelFor` question above, it is not reachable from this file. RestoreTracker.h is part of clad's runtime, pulled into the user's own translation unit through Differentiator.h, and that build does not link LLVM. `<thread>` has a smaller version of the same problem: it puts a pthread requirement on every clad user's build. clad's own lit does not pass `-pthread` for `%cladclang` (test/lit.cfg), only for the CUDA substitution. That happens to work on glibc 2.34 and later, where libpthread was folded into libc, and breaks on older toolchains, which clad still supports down to LLVM 12.

I could not find a test that would fail if this were reverted. The generated code is identical either way, so the FileCheck expectations do not see it, and a correct parallel restore and a correct serial restore produce the same numbers, so the CHECK-EXEC lines do not either. Codecov's "all modified lines covered" means those lines run, not that anything observes what they do.

This overlaps #1821, which rewrites the same `restore()` for a different reason. They conflict textually and they disagree about what `restore()` is for. #1821 is the one that can be argued on its merits once rebased, and it removes the per-record allocation that is the only plausible reason this loop would ever be worth parallelising. If it lands first, the question here becomes much easier to answer with a measurement.

vgvassilev:
I believe the best solution to #1539 might be described in #2141. Feel free to reopen if I missed something.

PR #1691, opened 2026-01-01
Fix: Add static_assert to prevent segfault in execute() with array arguments (#1690)

### LINKED ISSUE #1690: [Segfault] Passing array to execute() triggers runtime crash due to unsafe padding
### Description
Passing a raw array (e.g., `double grads[2]`) to `grad.execute(...)` causes a segmentation fault at runtime.

It seems `execute_with_default_args` fails to unpack the array into individual pointers. Instead of detecting the argument count mismatch, it assumes the remaining arguments are missing and pads them with `nullptr`. The generated gradient code then attempts to write to these null addresses.

### Reproduction
```cpp

double foo(double x, double y) { return x * y; }

int main() {
    auto grad = clad::gradient(foo);
    double x = 3.0, y = 4.0;
    
    // This triggers the crash
    double grads[2] = {0, 0};
    grad.execute(x, y, grads); 
    
    return 0;
}

### PR DESCRIPTION
**Pull request by ayush4874**
### Description
Fixes a segmentation fault that occurs when passing a raw array (e.g., `double[2]`) to `clad::gradient::execute`.

Previously, `execute_with_default_args` failed to detect the argument count mismatch when an array was passed. It would pad the missing arguments with `nullptr`, causing the generated code to write to null addresses at runtime.

### Changes
- Added a `static_assert` in `execute_with_default_args` (both standalone and member function versions) to enforce explicit pointer arguments.
- This converts the runtime crash into a readable compile-time error.

### Verification
- Added a regression test in `test/Diagnostics/ArrayExecuteSafety.cpp` using `FileCheck` to verify that the static assertion triggers correctly.
- Confirmed the test passes locally.

Fixes #1690

### DIFF
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -446,7 +446,7 @@ CUDA_HOST_DEVICE void push(tape<T[N], SBO_SIZE, SLAB_SIZE>& to, const U& val) {
                             return_type_t<F>>::type constexpr CUDA_HOST_DEVICE
     execute(Args&&... args) const {
       #if __cplusplus >= 201703L
-      static_assert((!std::is_array<typename std::remove_reference<Args>::type>::value && ...),
+      static_assert((!std::is_array<std::remove_reference_t<Args>>::value && ...),
                     "Clad: Mixed scalar/array arguments are not supported.");
       #endif
       return static_cast<return_type_t<F>>(0);
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -399,7 +399,7 @@ CUDA_HOST_DEVICE void push(tape<T[N], SBO_SIZE, SLAB_SIZE>& to, const U& val) {
                             return_type_t<F>>::type constexpr CUDA_HOST_DEVICE
     execute(Args&&... args) const {
       #if __cplusplus >= 201703L
-      static_assert((!std::is_array<typename std::remove_reference<Args>::type>::value && ...),
+      static_assert((!std::is_array<std::remove_reference_t<Args>>::value && ...),
                     "Clad: Mixed scalar/array arguments are not supported.");
       #endif
       if (!m_Function)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: include/clad/Differentiator/Differentiator.h [hunk 1]
```diff
@@ -292,6 +300,14 @@ CUDA_HOST_DEVICE void push(tape<T[N], SBO_SIZE, SLAB_SIZE>& to, const U& val) {
   execute_with_default_args(list<Rest...>, ReturnType C::*f, Obj&& obj,
                             list<fArgTypes...>,
                             Args&&... args) -> return_type_t<decltype(f)> {
+
+    // We cannot safely pad missing arguments with nullptr because the generated 
+    // code might try to write to them (segfault). Force explicit arguments.
+    static_assert(sizeof...(Rest) == 0,
```
vgvassilev: It seems this diagnostic remains untested.

ayush4874: I had to move this check. It was causing widespread regressions (about 80 failures) in the existing tests because it was triggering too aggressively inside `execute_with_default_args`.

Instead, I moved the `static_assert` up into the main `execute` wrapper and guarded it with `if constexpr(is_array)`. This ensures existing scalar calls remain unaffected while still strictly enforcing the rule for arrays.

I verified locally that the diagnostic still triggers correctly if mixed arguments are passed:

```cpp
include/clad/Differentiator/Differentiator.h:789:22: error: static assertion failed due to requirement 'sizeof...(Rest) == 0': Clad: Passing an array requires it to be the only argument. Mixed arguments are not supported.
```
I have also added `test/Arrays/ArrayExecuteTest.cpp` to verify that valid array execution works correctly.

test/Regressions/issue-1690.cpp
```diff
@@ -0,0 +1,17 @@
+// Regression test for array argument crash in execute()
```
vgvassilev: Apologies for the delay. Can you take a look how the regression tests are named in `test/Regression` and follow that pattern?

ayush4874: Done. I renamed the test to `test/Regressions/issue-1690.cpp` to match the project's conventions.

PR #1692, opened 2026-01-02
Fixes XFail test (ArrayErrors.C)

### LINKED ISSUE #1540: Fix xfailed tests with valgrind.

Vedant2005goyal:
This PR resolves the problem where error diagnostics in Hessian Mode were pointing to the function declaration instead of the call site when array indices were missing. This fix enable the "Hessian/ArrayErrors.C" which was previously marked as XFail.

Fixes #1540

### DIFF
--- a/test/Hessian/ArrayErrors.C
+++ b/test/Hessian/ArrayErrors.C
@@ -2,14 +2,13 @@
 
 #include "clad/Differentiator/Differentiator.h"
 
-//XFAIL:*
-double f(int a, const double *b) {
+double f(double a, const double *b) {
   return b[0] * a + b[1] * a + b[2] * a;
 }
 
 int main() {
-    clad::hessian(f, 1); // expected-error {{hessian mode differentiation w.r.t. array or pointer parameters needs explicit declaration of the indices of the array using the args parameter; did you mean 'clad::hessian(f, "b[0:<last index of b>]")'}}
+    clad::hessian(f, 1); // expected-error {{hessian mode differentiation w.r.t. array or pointer parameters needs explicit declaration of the indices of the array using the args parameter; did you mean 'clad::hessian}}
     clad::hessian(f, "a");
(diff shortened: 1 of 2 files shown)


test/Hessian/ArrayErrors.C
```diff
@@ -2,14 +2,14 @@
...
+double f(double a, const double *b) {
   return b[0] * a + b[1] * a + b[2] * a;
 }
 
 int main() {
-    clad::hessian(f, 1); // expected-error {{hessian mode differentiation w.r.t. array or pointer parameters needs explicit declaration of the indices of the array using the args parameter; did you mean 'clad::hessian(f, "b[0:<last index of b>]")'}}
+    clad::hessian(f); // expected-error {{hessian mode differentiation w.r.t. array or pointer parameters needs explicit declaration of the indices of the array using the args parameter}}
+    clad::hessian(f, 1); // expected-error {{hessian mode differentiation w.r.t. array or pointer parameters needs explicit declaration of the indices of the array using the args parameter}}
     clad::hessian(f, "a");
-    clad::hessian(f, "a, b"); // expected-error {{hessian mode differentiation w.r.t. array or pointer parameters needs explicit declaration of the indices of the array using the args parameter; did you mean 'clad::hessian(f, "a, b[0:<last index of b>]")'}}
```
vgvassilev, line 13: Why did we lose that wording?

Vedant2005goyal: i removed that to simplify the check and to enusure that the test passes easily but if you prefer to have it i will accomodate that too.

vgvassilev: If that wording is checked in some other test then that's fine as is but somehow we lost testing coverage, too..

Vedant2005goyal: test coverage was lost due to the ternary operator i have used and i have replaced that logic with the if block and also added the missing terms in the check

PR #1695, opened 2026-01-06
Rework lambdas in Clad

ovdiiuv:
This PR adds basic support for lambda pullbacks inside derived functions. It is not complete and still requires some work, but implements core features.

TODO list:
- Fix clang version API mismatch.
- Make m_Globals emit to the lambda not to the outer function.
- Replace const_cast inside of RMV::VisitLambdaExpr with proper lambda cloning.
- Remove const_cast of a DiffRequest inside of buildLambda.
- Probably rework how we fill the body of a lambda inside buildLambda.
- Fix VA from failing.
- Address minor issues left as comments.

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -22,10 +22,10 @@
 #include "clang/AST/StmtVisitor.h"
 #include "clang/Basic/Diagnostic.h"
 #include "clang/Basic/SourceLocation.h"
+#include "clang/Basic/Version.h"
 #include "clang/Sema/Sema.h"
-
-#include <llvm/ADT/ArrayRef.h>
-#include <llvm/ADT/SmallVector.h>
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/SmallVector.h"
 
 #include <array>
 #include <limits>
@@ -90,7 +90,7 @@ namespace clad {
 
     unsigned outputArrayCursor = 0;
     unsigned numParams = 0;
-    clang::Expr* m_Pullback = nullptr;
+    llvm::SmallVector<clang::Expr*, 1> m_Pullback;
     const char* funcPostfix() const {
       if (m_DiffReq.Mode == DiffMode::jacobian)
         return "_jac";
@@ -386,6 +386,10 @@ namespace clad {
     StmtDiff VisitForStmt(const clang::ForStmt* FS);
     StmtDiff VisitIfStmt(const clang::IfStmt* If);
     StmtDiff VisitImplicitCastExpr(const clang::ImplicitCastExpr* ICE);
+
+#if CLANG_VERSION_MAJOR > 16
+    StmtDiff VisitLambdaExpr(const clang::LambdaExpr* LE);
+#endif // CLANG_VERSION_MAJOR
     StmtDiff
     VisitCXXFunctionalCastExpr(const clang::CXXFunctionalCastExpr* FCE);
     StmtDiff VisitCStyleCastExpr(const clang::CStyleCastExpr* CSCE);
@@ -696,8 +700,21 @@ namespace clad {
     ///\paramp[in] source An external RMV source
     void AddExternalSource(ExternalRMVSource& source);
 
+    clang::QualType GetLambdaDerivativeType(const clang::LambdaExpr* LE) {
+      clang::FunctionDecl* FD = LE->getCallOperator();
+      llvm::SmallVector<const clang::ValueDecl*, 4> diffParams{};
+      for (const auto* param : FD->parameters())
+        diffParams.push_back(param);
+
+      return utils::GetDerivativeType(m_Sema, FD, DiffMode::pullback,
+                                      diffParams,
+                                      /*forCustomDerv=*/false,
+                                      /*shouldUseRestoreTracker=*/false);
+    }
+    clang::Expr* buildDerivedLambda(const clang::LambdaExpr* LE);
     /// Builds and returns the sequence of derived function parameters.
-    void BuildParams(llvm::SmallVectorImpl<clang::ParmVarDecl*>& params);
+    void BuildParams(llvm::SmallVectorImpl<clang::ParmVarDecl*>& params,
+                     const clang::LambdaExpr* LE = nullptr);
 
     /// Stores data required for differentiating a switch statement.
     struct SwitchStmtInfo {
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -20,6 +20,8 @@
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/ASTLambda.h"
 #include "clang/AST/Attr.h"
+#include "clang/AST/Decl.h"
+#include "clang/AST/DeclBase.h"
 #include "clang/AST/DeclCXX.h"
 #include "clang/AST/DeclTemplate.h"
 #include "clang/AST/DeclarationName.h"
@@ -30,17 +32,22 @@
 #include "clang/AST/TemplateBase.h"
 #include "clang/AST/Type.h"
 #include "clang/Analysis/AnalysisDeclContext.h"
+#include "clang/Basic/ExceptionSpecificationType.h"
 #include "clang/Basic/LLVM.h" // for clang::isa
+#include "clang/Basic/Lambda.h"
 #include "clang/Basic/OperatorKinds.h"
 #include "clang/Basic/SourceLocation.h"
 #include "clang/Basic/TargetInfo.h"
 #include "clang/Basic/TokenKinds.h"
 #include "clang/Basic/TypeTraits.h"
 #include "clang/Basic/Version.h"
+#include "clang/Sema/DeclSpec.h"
 #include "clang/Sema/Lookup.h"
 #include "clang/Sema/Overload.h"
 #include "clang/Sema/Ownership.h"
+#include "clang/Sema/ParsedAttr.h"
 #include "clang/Sema/Scope.h"
+#include "clang/Sema/ScopeInfo.h"
 #include "clang/Sema/Sema.h"
 #include "clang/Sema/SemaInternal.h"
 #include "clang/Sema/Template.h"
@@ -265,9 +272,9 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     // added by the plugins yet.
     if (m_DiffReq.Mode == DiffMode::reverse) {
       if (returnTy->isRealType())
-        m_Pullback =
-            ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context,
-                                              /*val=*/1);
+        m_Pullback.push_back(ConstantFolder::synthesizeLiteral(m_Context.IntTy,
+                                                               m_Context,
+                                                               /*val=*/1));
       else if (!returnTy->isVoidType()) {
         diag(DiagnosticsEngine::Warning, m_DiffReq.Function->getBeginLoc(),
              "clad::gradient only supports differentiation functions of real "
@@ -1313,7 +1320,9 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       return {nullptr, nullptr};
     const Expr* value = RS->getRetValue();
     QualType type = value->getType();
-    auto* dfdf = m_Pullback;
+    Expr* dfdf = nullptr;
+    if (!m_Pullback.empty())
+      dfdf = m_Pullback.back();
     if (dfdf && (isa<FloatingLiteral>(dfdf) || isa<IntegerLiteral>(dfdf)) &&
         type->isScalarType()) {
       ExprResult tmp = dfdf;
@@ -1469,17 +1478,6 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     // Check if referenced Decl was "replaced" with another identifier inside
     // the derivative
     if (auto* VD = dyn_cast<VarDecl>(cast<DeclRefExpr>(clonedDRE)->getDecl())) {
-      // If current context is different than the context of the original
-      // declaration (e.g. we are inside lambda), rebuild the DeclRefExpr
-      // with Sema::BuildDeclRefExpr. This is required in some cases, e.g.
-      // Sema::BuildDeclRefExpr is responsible for adding captured fields
-      // to the underlying struct of a lambda.
-      if (VD->getDeclContext() != m_Sema.CurContext) {
-        auto* ccDRE = dyn_cast<DeclRefExpr>(clonedDRE);
-        NestedNameSpecifier* NNS = DRE->getQualifier();
-        auto* referencedDecl = cast<VarDecl>(ccDRE->getDecl());
-        clonedDRE = BuildDeclRef(referencedDecl, NNS, DRE->getValueKind());
-      }
       // This case happens when ref-type variables have to become function
       // global. Ref-type declarations cannot be moved to the function global
       // scope because they can't be separated from their inits.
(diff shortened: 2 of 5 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -250,6 +250,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
   }
 
   DerivativeAndOverload ReverseModeVisitor::Derive() {
+    // m_DiffReq->dump();
```
vgvassilev: Stray comments.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -698,8 +710,139 @@ namespace clad {
...
+                        /* NumExceptions */ 0,
+                        /* NoexceptExpr */ nullptr,
+                        /* ExceptionSpecTokens */ nullptr,
+                        /* DeclsInPrototype */ std::nullopt,
+                        /* LocalRangeBegin */ noLoc,
+                        /* LocalRangeEnd */ noLoc,
+                        /* Declarator */ D,
+                        /* TrailingReturnType */ clang::ParsedType(),
+                        /* TrailingReturnTypeLoc */ clang::SourceLocation()),
+                    /* EndLoc */ clang::SourceLocation());
```
vgvassilev:
```suggestion
                        /*hasProto=*/ true,
                        /*isAmbiguous=*/ false,
                        /*LParenLoc=*/ noLoc,
                        /*Params=*/ ParamInfoLambda.data(),
                        /*NumParams=*/ ParamInfoLambda.size(),
                        /*EllipsisLoc=*/ clang::SourceLocation(),
                        /*RParenLoc=*/ noLoc,
                        /*RefQualifierIsLValueRef=*/ true,
                        /*RefQualifierLoc=*/ clang::SourceLocation(),
                        /*MutableLoc=*/ clang::SourceLocation(),
                        /*ESpecType=*/ clang::EST_None,
                        /*ESpecRange=*/ clang::SourceRange(),
                        /*Exceptions=*/ nullptr,
                        /*ExceptionRanges=*/ nullptr,
                        /*NumExceptions=*/ 0,
                        /*NoexceptExpr=*/ nullptr,
                        /*ExceptionSpecTokens=*/ nullptr,
                        /*DeclsInPrototype=*/ std::nullopt,
                        /*LocalRangeBegin=*/ noLoc,
                        /*LocalRangeEnd=*/ noLoc,
                        /*Declarator=*/ D,
                        /*TrailingReturnType=*/ clang::ParsedType(),
                        /*TrailingReturnTypeLoc=*/ clang::SourceLocation()),
                    /*EndLoc=*/ clang::SourceLocation());
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1732,6 +1724,22 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     return result;
   }
 
+#if CLANG_VERSION_MAJOR > 16
+  StmtDiff ReverseModeVisitor::VisitLambdaExpr(const LambdaExpr* LE) {
+    Expr* lambdaE = buildLambda(LE, [&]() {
+      StmtDiff BodyDiff = Visit(LE->getCallOperator()->getBody());
+      // will do for now
+      for (auto* S : cast<CompoundStmt>(BodyDiff.getStmt())->body())
+        addToCurrentBlock(S);
```
vgvassilev: Can we write a test hitting this one?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1745,6 +1753,38 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+              if (lookupVD->getNameAsString() ==
+                  "_d_" + VD->getNameAsString()) {
+                CXXScopeSpec SS;
+                LambdaCallOpExpr = DeclRefExpr::Create(
+                    m_Context, NestedNameSpecifierLoc(), SourceLocation(),
+                    lookupVD,
+                    /*RefersToEnclosingVariableOrCapture=*/false, noLoc,
+                    lookupVD->getType(), VK_LValue);
+              }
+        }
```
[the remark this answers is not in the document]
vgvassilev: That seems a reasonable suggestion -- maybe it did not get it right but we should get rid of the else...

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1745,6 +1753,38 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       return StmtDiff(Clone(CE));
     }
 
+    bool isLambdaCE = false;
```
vgvassilev: Can we somehow get rid of `isLambdaCE` state we are dragging around.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3136,6 +3187,14 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
         }
 
         decls.push_back(VDDiff.getDecl());
+        if (isLambdaDS) {
+          // uncomment at some point
+          // m_Sema.FinalizeDeclaration(VDDiff.getDecl_dx());
+          // m_Sema.FinalizeDeclaration(VDDiff.getDecl());
```
vgvassilev: Why we can't use that code? Maybe clarify in the comment.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3136,6 +3187,14 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+          m_Sema.PushOnScopeChains(VDDiff.getDecl_dx(), getCurrentScope(),
+                                   true);
+          m_Sema.PushOnScopeChains(VDDiff.getDecl(), getCurrentScope(), true);
```
```suggestion
          m_Sema.PushOnScopeChains(VDDiff.getDecl_dx(), getCurrentScope(),
                                   /*XXX=*/true);
          m_Sema.PushOnScopeChains(VDDiff.getDecl(), getCurrentScope(), /*XXX=*/true);
```

Please add the name of the param here.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4567,7 +4630,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     bool HasRet = false;
     QualType dRetTy = FD->getReturnType().getNonReferenceType();
     dRetTy = utils::getNonConstType(dRetTy, m_Sema);
-    if (m_DiffReq.Mode == DiffMode::pullback && !dRetTy->isVoidType() &&
+    if ((m_DiffReq.Mode == DiffMode::pullback || LE) && !dRetTy->isVoidType() &&
```
vgvassilev, line 4710: Why do we need to special case this LE here?

ovdiiuv: because we want to treat lambdas as pullbacks. we could remove it but that would require more const_casts to temporarily set the DiffMode.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -698,8 +700,140 @@ namespace clad {
...
+        diffParams.push_back(param);
+
+      return utils::GetDerivativeType(
+          m_Sema, FD, DiffMode::pullback, diffParams,
+          /*forCustomDerv=*/false,
+          /*shouldUseRestoreTracker=*/false);
+    }
+
+    template <typename F>
+    clang::Expr* buildLambda(const clang::LambdaExpr* LE, F&& func) {
```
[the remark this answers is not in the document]
vgvassilev: Seems useful to address...

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -698,8 +710,139 @@ namespace clad {
...
+
+      return utils::GetDerivativeType(m_Sema, FD, DiffMode::pullback,
+                                      diffParams,
+                                      /*forCustomDerv=*/false,
+                                      /*shouldUseRestoreTracker=*/false);
+    }
+
+#if CLANG_VERSION_MAJOR > 16
+    template <typename F>
+    clang::Expr* buildLambda(const clang::LambdaExpr* LE, F&& func) {
```
vgvassilev: We should move the definition to the cpp file.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -698,8 +710,139 @@ namespace clad {
...
+      clang::sema::LambdaScopeInfo* LSI = m_Sema.getCurLambda();
+      LSI->CallOperator->setType(dFnType);
+
+      m_Sema.PushDeclContext(getCurrentScope(), LSI->CallOperator);
+
+      LSI->Lambda->setDeclContext(DC);
+      // DC->addDecl(LSI->Lambda);
+      m_Derivative = LSI->CallOperator;
+      // should be easy remove
+      const_cast<DiffRequest&>(m_DiffReq).Function = LE->getCallOperator();
```
vgvassilev: Can we open an issue about this -- ideally this should be scheduled statically in a separate DiffRequest...

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1732,6 +1725,22 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+      // will do for now
+      for (auto* S : cast<CompoundStmt>(BodyDiff.getStmt())->body())
+        addToCurrentBlock(S);
+      for (auto* S : cast<CompoundStmt>(BodyDiff.getStmt_dx())->body())
+        addToCurrentBlock(S);
+    });
+
+    m_Pullback.pop_back();
+    // FIXME: Clone lambda properly.
+    return {const_cast<Expr*>(cast<Expr>(LE)), lambdaE};
```
vgvassilev, line 1856: Can you open an issue to track this?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2909,9 +2961,21 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+    else {
+      QualType AutoQT = m_Context.getAutoDeductType();
+      TypeSourceInfo* TSI = m_Context.getTrivialTypeSourceInfo(AutoQT, noLoc);
+      IdentifierInfo* Id = &m_Context.Idents.get(VD->getNameAsString());
+      VDClone = VarDecl::Create(m_Context, m_Sema.getCurLexicalContext(), noLoc,
+                                noLoc, Id, AutoQT, TSI, SC_None);
+      m_Sema.AddInitializerToDecl(VDClone, initDiff.getExpr(),
+                                  /*DirectInit=*/false);
+      m_Sema.FinalizeDeclaration(VDClone);
+      m_Sema.PushOnScopeChains(VDClone, getCurrentScope(),
```
vgvassilev: This seems duplicated code, can you move it into a utility function? Maybe you "just" need `VisitorBase::BuildVarDecl`?

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: Okay, this will go towards the point I made earlier about statically scheduling the lambdas.

lib/Differentiator/ReverseModeVisitor.cpp
ovdiiuv: sure, will open every issue that is mentioned in the PR comment.

include/clad/Differentiator/ReverseModeVisitor.h
```diff
@@ -698,8 +702,22 @@ namespace clad {
...
+      for (const auto* param : FD->parameters())
+        diffParams.push_back(param);
+
+      return utils::GetDerivativeType(m_Sema, FD, DiffMode::pullback,
+                                      diffParams,
+                                      /*forCustomDerv=*/false,
+                                      /*shouldUseRestoreTracker=*/false);
+    }
+    template <typename F>
+    clang::Expr* buildLambda(const clang::LambdaExpr* LE, F func);
```
vgvassilev:
```suggestion
```

ovdiiuv: we cant make it file static, if that is what you mean.

vgvassilev: Why not?

ovdiiuv: we set m_DerivativeFnScope inside, for example.

vgvassilev: Ah, okay, sorry overlooked that. We can’t have an out of line template. Please move the definition back in the header.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1732,6 +1730,135 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+    auto* DC =
+        const_cast<DeclContext*>(LE->getCallOperator()->getDeclContext());
+    llvm::SaveAndRestore<DeclContext*> SaveContext(m_Sema.CurContext);
+    llvm::SaveAndRestore<FunctionDecl*> SaveDerivative(m_Derivative);
+
+    // should be easy remove
+    auto*& FuncRef = const_cast<const FunctionDecl*&>(m_DiffReq.Function);
+
+    llvm::SaveAndRestore<const FunctionDecl*> SaveReq(FuncRef);
+    llvm::SaveAndRestore<Scope*> SaveScope(getCurrentScope());
```
PetroZarytskyi: Do we need this considering we call ``endScope`` in the end of the function?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1732,6 +1730,135 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+    m_Sema.PushDeclContext(getCurrentScope(), LSI->CallOperator);
+
+    LSI->Lambda->setDeclContext(DC);
+
+    m_Derivative = LSI->CallOperator;
+    // FIXME: Should be easy remove
+    const_cast<DiffRequest&>(m_DiffReq).Function = LE->getCallOperator();
+
+    BuildParams(params, LE);
+    m_Derivative->setBody(new (m_Context) CompoundStmt(noLoc));
```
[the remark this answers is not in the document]
PetroZarytskyi: We should avoid creating stmts with new. We have plenty of utility functions. For example, you can do ``MakeCompoundStmt({})``.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1732,6 +1730,135 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+                 clang::DeclaratorContext::LambdaExpr);
+
+    QualType dFnType = GetLambdaDerivativeType(LE);
+
+    auto* DC =
+        const_cast<DeclContext*>(LE->getCallOperator()->getDeclContext());
+    llvm::SaveAndRestore<DeclContext*> SaveContext(m_Sema.CurContext);
+    llvm::SaveAndRestore<FunctionDecl*> SaveDerivative(m_Derivative);
+
+    // should be easy remove
```
PetroZarytskyi: Perhaps this should be a fixme.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1732,6 +1730,135 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+    endScope();
+    endScope();
+    endScope();
+    return lambda;
+  }
+
+  StmtDiff ReverseModeVisitor::VisitLambdaExpr(const LambdaExpr* LE) {
+    Expr* lambdaE = buildLambda(LE, [&]() {
+      StmtDiff BodyDiff = Visit(LE->getCallOperator()->getBody());
+      // will do for now
```
PetroZarytskyi: Can you briefly mention our thoughts about globals as a fixme?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1745,6 +1872,37 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       return StmtDiff(Clone(CE));
     }
 
+    const auto* MD = dyn_cast<CXXMethodDecl>(FD);
+    Expr* LambdaCallOpExpr = nullptr;
+    if (MD) {
+      if (isLambdaCallOperator(MD)) {
+        const auto* CallE = CE->getArg(0)->IgnoreParenImpCasts();
+        // handle direct calls later
```
PetroZarytskyi: Can you mark this as a fixme?

test/Gradient/Lambdas.C
```diff
@@ -12,56 +13,59 @@ double f1(double i, double j) {
...
-// CHECK-NEXT:             *_d_j += _r0;
-// CHECK-NEXT:         }
-// CHECK-NEXT:     }
+// CHECK-NEXT: }
 
 double f2(double i, double j) {
   auto _f = [] (double t, double k) {
-    return t + k;
+    double a = t*k;
+    return a;
```
PetroZarytskyi, line 37: Why is the primal test changed? I think it would make more sense to leave the test as it was and add this new version as a separate function `f3`

ovdiiuv: it was not any different from previous one. it would make more sense to add tests when globals issue is solved.

include/clad/Differentiator/ReverseModeVisitor.h
ovdiiuv: static scheduling doesn't seem like an option. we cannot derive lambda first -- VarDecls of captures have to be cloned prior. if we derive the outer funtions first, then we need to introduce a placeholder to later replace it with the derived lambda. and the biggest concern is proper derived lambda lookup, it has to not only consider signatures, but scopes.

PR #1704, opened 2026-01-11
Fix #1701 : lambda-local globals leaking into outer derivative function 

### LINKED ISSUE #1701: FIx global variables being emmited to the outer function when visiting lambda
When differentiating
```cpp
void f2(double i, double j) {
  auto _f = [] () {
    {
      double a = 1;
    }
  };
}
```

we get

```cpp
void f2_grad(double i, double j, double *_d_i, double *_d_j) {
    double _d_a = 0.;
    double a = 0.;
    auto _f0 = []() {
        {
            double a = 1;
        }
    };
    auto _d__f = []() {
        {
            a = 1;
        }
        {
        }
    };
}
```

`double a = 0.;` declaration is in the wrong place. The fix is to make `m_Globals` a stack.

SrijanRoy123-github:
Fixes: #1701

## Summary
When differentiating functions that contain lambdas, lambda-local variables (e.g., `a` / `_d_a`) could be emitted into the *outer* derivative function body. This change makes the “globals” storage scoped (stack-based), so lambda-specific globals are emitted in the correct helper/pullback context instead of leaking into the parent derivative function.

## Tests
- Added regression test: `test/Misc/LambdaGlobals.cpp`
- Ran: `cmake --build "$CLAD_ROOT/build" --target check-clad`
- Ran: `cmake --build "$CLAD_ROOT/build" --target check-clad-misc`

Screenshot 2026-01-12 at 12 22 27 AM

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -77,7 +77,7 @@ namespace clad {
     /// A sequence of DeclStmts containing "tape" variable declarations
     /// that will be put immediately in the beginning of derivative function
     /// block.
-    ///Stmts m_Globals;
+    /// Stmts m_Globals;
     std::vector<Stmts> m_Globals = {{}};
 
     Stmts& globals() { return m_Globals.back(); }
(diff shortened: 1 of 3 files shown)


vgvassilev:
Please make sure you run the tests locally before pushing.

PR #1707, opened 2026-01-13
Check if valgrind still fails on clang 21

vgvassilev:


vgvassilev:
Yes, and it fails badly.

PR #1708, opened 2026-01-15
Corrected the wrong scope of the variables declared inside the Lambda function

### LINKED ISSUE #1701: FIx global variables being emmited to the outer function when visiting lambda

Vedant2005goyal:
Rather than changing the data structure of the m_Globals to stack which has been specified by the author of the issue (since it breaks the memory layout of the class) I have opted to specifically fix the variable scope problem for the Lambda functions only, by swapping the variables  in order to temporarily isolate the lambda variable declaration from the outer function and inject them in the correct scope only.
```cpp
void f2_grad(double i, double j, double *_d_i, double *_d_j) {
    auto _f0 = []() {
        {
            double a = 1;
        }
    };
    auto _d__f = []() {
        double _d_a = 0.;
        double a = 0.;
        {
            a = 1;
        }
        {
        }
    };
}
```
Fixes #1701

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1823,9 +1823,11 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
             getCurrentScope(), DS));
 
     m_DerivativeFnScope = getCurrentScope();
+    Stmts OuterGlobals;
+    std::swap(m_Globals, OuterGlobals);
+
     beginBlock();
 
-    // FIXME: Fix m_Globals being emmitted to the outer function.
     StmtDiff BodyDiff = Visit(LE->getCallOperator()->getBody());
     for (auto* S : cast<CompoundStmt>(BodyDiff.getStmt())->body())
       addToCurrentBlock(S);
@@ -1833,6 +1835,19 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       addToCurrentBlock(S);
 
     CompoundStmt* DerivedBody = endBlock();
+
+    if (!m_Globals.empty()) {
+      llvm::SmallVector<Stmt*, 32> NewBodyStmts;
+      NewBodyStmts.append(m_Globals.begin(), m_Globals.end());
+      NewBodyStmts.append(DerivedBody->body().begin(),
+                          DerivedBody->body().end());
+
+      DerivedBody = CompoundStmt::Create(
+          m_Sema.getASTContext(), NewBodyStmts, FPOptionsOverride(),
+          DerivedBody->getLBracLoc(), DerivedBody->getRBracLoc());
+    }
+    m_Globals.clear();
+    std::swap(m_Globals, OuterGlobals);
 
     Expr* lambda =
         m_Sema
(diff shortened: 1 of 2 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1834,6 +1836,19 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
 
     CompoundStmt* DerivedBody = endBlock();
 
+    if (!m_Globals.empty()) {
+      llvm::SmallVector<Stmt*, 32> NewBodyStmts;
+      NewBodyStmts.append(m_Globals.begin(), m_Globals.end());
+      NewBodyStmts.append(DerivedBody->body().begin(),
+                          DerivedBody->body().end());
```
vgvassilev, line 1843: We will need tests. Can you also look why valgrind fails for this code quite badly: https://github.com/vgvassilev/clad/pull/1707

Vedant2005goyal: Hi @vgvassilev 
I have identified the valgrind fail and that was due to a logical problem in the **VisitLambdaExpr** where the pop_back() operation was done without checking whether the list was empty or not and i have added a guard for that and it resolves it in my latest push.
I have earlier added regression test to increase coverage but it failed many tests due to the clang version problem which is caused by the 'if guard in the **buildDerivedLambda**'

Image

vgvassilev: This should probably be a separate pull request.

Vedant2005goyal: Just for clarification i want to know whether you are referring to seperate PR for the clang version thing and if yes then i will be happy to contribute to that, but for the time being should we wait for it to be merged first before proceeding with this current one?

vgvassilev: I meant the valgrind fix.

Vedant2005goyal: Ok I will make separate PR for that

Vedant2005goyal: Hi @vgvassilev 
I have seperated the PR's, if there is anything needed to be done further, Please let me know.

vgvassilev: Thank you, it has fixed the valgrind issues with llvm21 coming from lambda functions. There are a few new ones that need to be addressed but that's probably outside of the scope of this PR. To move forward this PR we need to add tests covering the newly added lines.

Vedant2005goyal: Thank you @vgvassilev for reviewing the PR.

I have added the regression test covering the new lambda scope logic. Since this specific lambda behavior relies on newer Clang features, I wrapped the test case in a #if CLANG_VERSION_MAJOR > 16 guard.

vgvassilev: You can do `// UNSUPPORTED: clang-13, ..., clang-15

Vedant2005goyal: Use of //UNSUPPORTED was a much cleaner option! I have pushed the update.

PR #1709, opened 2026-01-16
Fixes the valgrind fails 

Vedant2005goyal:
valgrind fails which are caused because of pull_back operation in the VisitLambdaExpr done on list without checking whether it is empty or not


vgvassilev (approved):
Lgtm!

PR #1714, opened 2026-01-29
Fix C++11 build failure by adding compatibility macro for relaxed constexpr (Fixes #1713)

### LINKED ISSUE #1713: [Bug] Build failure with -std=c++11 due to C++14 relaxed constexpr usage
**Problem:**
Building CLAD with `-std=c++11` fails because several headers use "relaxed constexpr" features that were only introduced in C++14.

**Errors Observed:**
1. Loops: `for` loops inside `constexpr` functions (e.g., `ArrayRef.h:52`).
2. Multiple Statements: Functions containing `assert`, `if`, and multiple `return` statements
3. Implicit Const: Warnings regarding `constexpr` member functions not being implicitly `const` in C++14.

**Proposed Solution:**
Introduce a compatibility macro (e.g., `CLAD_CONSTEXPR_CXX14`) that expands to:
- `constexpr` for C++14 and newer.
- `inline` for C++11 (stripping the constexpr requirement to allow compilation).

link to error on godbolt (using clad v1.10): https://godbolt.org/z/5oPhnM73q

leetcodez:
**The Problem**
I noticed that building CLAD with `-std=c++11` fails right now. The issue is that `ArrayRef.h` and `Differentiator.h` are using some "relaxed constexpr" features (like loops, `if` statements, and multiple returns) that are only allowed in C++14 and newer.

**The Fix**
I added a simple compatibility macro called `CLAD_CONSTEXPR_CXX14`.
- For C++14+, it acts just like `constexpr`.
- For C++11, it falls back to `inline`. This keeps the code efficient but allows it to actually compile under the stricter C++11 rules.

**What I Changed**
1. Added the macro definition to `Differentiator.h` and `ArrayRef.h`.
2. Swapped out `constexpr` for the macro in the crashing functions (mostly in `ArrayRef::operator=` and `slice`).
3. Fixed a small issue where an `auto` return type was missing its trailing return type (another C++11 requirement).

**Verification**
I verified this locally by running `clang++ -std=c++11 ...` against a demo file, and it builds cleanly now without the constexpr errors.

### DIFF
--- a/include/clad/Differentiator/ArrayRef.h
+++ b/include/clad/Differentiator/ArrayRef.h
@@ -35,7 +35,7 @@ public:
       : m_arr(a.ptr()), m_size(a.size()) {}
 
   /// Operator for conversion from array_ref<T> to T*.
-  constexpr CUDA_HOST_DEVICE operator T*() { return m_arr; }
+  CLAD_CONSTEXPR_CXX14 CUDA_HOST_DEVICE operator T*() { return m_arr; }
   /// Operator for conversion from array_ref<T> to const T*.
   constexpr CUDA_HOST_DEVICE operator const T*() const { return m_arr; }
 
@@ -47,7 +47,8 @@ public:
     return *this;
   }
 
-  constexpr CUDA_HOST_DEVICE array_ref<T>& operator=(const array_ref<T>& a) {
+  CLAD_CONSTEXPR_CXX14 CUDA_HOST_DEVICE array_ref<T>&
+  operator=(const array_ref<T>& a) {
     if (this == &a)
       return *this;
     assert(m_size == a.size());
@@ -67,18 +68,20 @@ public:
   /// Returns the size of the underlying array
   constexpr CUDA_HOST_DEVICE std::size_t size() const { return m_size; }
   constexpr CUDA_HOST_DEVICE PUREFUNC T* ptr() const { return m_arr; }
-  constexpr CUDA_HOST_DEVICE PUREFUNC T*& ptr_ref() { return m_arr; }
+  CLAD_CONSTEXPR_CXX14 CUDA_HOST_DEVICE PUREFUNC T*& ptr_ref() { return m_arr; }
   /// Returns an array_ref to a part of the underlying array starting at
   /// offset and having the specified size
-  constexpr CUDA_HOST_DEVICE array_ref<T> slice(std::size_t offset,
-                                                std::size_t size) {
+  CLAD_CONSTEXPR_CXX14 CUDA_HOST_DEVICE array_ref<T> slice(std::size_t offset,
+                                                           std::size_t size) {
     assert((offset >= 0) && (offset + size <= m_size) &&
            "Window is outside array. Please provide an offset and size "
            "inside the array size.");
     return array_ref<T>(&m_arr[offset], size);
   }
   /// Returns the reference to the underlying array
-  constexpr CUDA_HOST_DEVICE PUREFUNC T& operator*() { return *m_arr; }
+  CLAD_CONSTEXPR_CXX14 CUDA_HOST_DEVICE PUREFUNC T& operator*() {
+    return *m_arr;
+  }
 
   // Arithmetic overloads
   /// Divides the arrays element wise
@@ -254,7 +257,8 @@ public:
     template <typename T>
     constexpr CUDA_HOST_DEVICE array_ref(const array_ref<T>& other)
         : m_arr(other.ptr()), m_size(other.size()) {}
-    template <typename T> constexpr CUDA_HOST_DEVICE operator array_ref<T>() {
+    template <typename T>
+    CLAD_CONSTEXPR_CXX14 CUDA_HOST_DEVICE operator array_ref<T>() {
       return array_ref<T>((T*)(m_arr), m_size);
     }
     [[nodiscard]] constexpr CUDA_HOST_DEVICE void* ptr() const { return m_arr; }
--- a/include/clad/Differentiator/CladConfig.h
+++ b/include/clad/Differentiator/CladConfig.h
@@ -3,6 +3,12 @@
 #ifndef CLAD_CONFIG_H
 #define CLAD_CONFIG_H
 
+#if __cplusplus >= 201402L
+#define CLAD_CONSTEXPR_CXX14 constexpr
+#else
+#define CLAD_CONSTEXPR_CXX14 inline
+#endif
+
 #include <cstdlib>
 #include <memory>
 
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -259,7 +259,7 @@ T& back(
   template <bool EnablePadding, class... Rest, class F, class... Args,
             class... fArgTypes,
             typename std::enable_if<EnablePadding, bool>::type = true>
-  constexpr CUDA_HOST_DEVICE return_type_t<F>
+  CLAD_CONSTEXPR_CXX14 CUDA_HOST_DEVICE return_type_t<F>
   execute_with_default_args(list<Rest...>, F f, list<fArgTypes...>,
                             CUDA_ARGS CUDA_REST_ARGS Args&&... args) {
 #if defined(__CUDACC__) && !defined(__CUDA_ARCH__)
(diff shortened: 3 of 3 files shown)


include/clad/Differentiator/ArrayRef.h
```diff
@@ -7,6 +7,14 @@
 #include <assert.h>
 #include <type_traits>
 
+// --- BACKWARD COMPATIBILITY MACRO ---
+#if __cplusplus >= 201402L
+    #define CLAD_CONSTEXPR_CXX14 constexpr
+#else
+    #define CLAD_CONSTEXPR_CXX14 inline
+#endif
+// ------------------------------------
```
vgvassilev: That should go in a common place such as Config.h.

include/clad/Differentiator/ArrayRef.h
```diff
@@ -47,7 +55,8 @@ template <typename T> class array_ref {
     return *this;
   }
 
-  constexpr CUDA_HOST_DEVICE array_ref<T>& operator=(const array_ref<T>& a) {
+  // FIXED: Loop inside constexpr crashes C++11
```
vgvassilev: Please remove these redundant comments.

include/clad/Differentiator/ArrayRef.h
leetcodez: I have moved the compatibility macro to CladConfig.h

include/clad/Differentiator/ArrayRef.h
leetcodez: I have removed the redundant comments to make the code more clean

include/clad/Differentiator/ArrayRef.h
```diff
@@ -47,7 +47,8 @@ template <typename T> class array_ref {
     return *this;
   }
 
-  constexpr CUDA_HOST_DEVICE array_ref<T>& operator=(const array_ref<T>& a) {
+  CLAD_CONSTEXPR_CXX14 CUDA_HOST_DEVICE array_ref<T>&
+  operator=(const array_ref<T>& a) {
```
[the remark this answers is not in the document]
vgvassilev: Can we address such issues in separate PRs?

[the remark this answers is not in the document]
leetcodez: allright!

leetcodez:
Sorry for the late commit I was in the hospital for a week.
I've addressed the C++11 compatibility issues and ensured the code is properly formatted. You can expect more regular contributions from my side moving forward :)

include/clad/Differentiator/CladConfig.h
```diff
@@ -91,10 +97,7 @@ template <typename T> __host__ __device__ T* clad_addressof(T& r) {
 #endif
 }
 #else
-template<typename T>
-T* clad_addressof(T& r) {
-  return std::addressof(r);
-}
+template <typename T> T* clad_addressof(T& r) { return std::addressof(r); }
```
vgvassilev: Why did we change this? I do not think we touched it?

include/clad/Differentiator/Differentiator.h
```diff
@@ -339,9 +339,15 @@ CUDA_HOST_DEVICE void push(tape<T[N], SBO_SIZE, SLAB_SIZE>& to, const U& val) {
...
-        ;
+      m_Code = (char*)malloc(length + 1);
+      if (m_Code)
+        memcpy((void*)m_Code, code, length + 1);
+      else
+#ifdef __CUDACC__
+        printf("stderr: Error: Failed to allocate memory for m_Code\n");
+#else
+        fprintf(stderr, "Error: Failed to allocate memory for m_Code\n");
+#endif
```
vgvassilev: Why did we change this?

include/clad/Differentiator/Differentiator.h
```diff
@@ -396,8 +402,9 @@ CUDA_HOST_DEVICE void push(tape<T[N], SBO_SIZE, SLAB_SIZE>& to, const U& val) {
 
     template <typename... Args, class FnType = CladFunctionType>
     typename std::enable_if<!std::is_same<FnType, NoFunction*>::value,
-                            return_type_t<F>>::type constexpr CUDA_HOST_DEVICE
-    execute(Args&&... args) const {
+                            return_type_t<F>>::type
+        CLAD_CONSTEXPR_CXX14 CUDA_HOST_DEVICE
+        execute(Args&&... args) const {
```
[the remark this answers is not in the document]
vgvassilev: Is this warning valid?

include/clad/Differentiator/Differentiator.h
```diff
@@ -445,12 +452,13 @@ CUDA_HOST_DEVICE void push(tape<T[N], SBO_SIZE, SLAB_SIZE>& to, const U& val) {
     }
 
     template <typename... Args>
-    constexpr CUDA_HOST_DEVICE auto operator()(Args&&... args) const {
+    constexpr CUDA_HOST_DEVICE auto operator()(Args&&... args) const
+        -> decltype(this->execute(std::forward<Args>(args)...)) {
```
vgvassilev, line 468: Did we really have to change this?

include/clad/Differentiator/Differentiator.h
```diff
@@ -801,4 +809,4 @@ CUDA_HOST_DEVICE void push(tape<T[N], SBO_SIZE, SLAB_SIZE>& to, const U& val) {
 
 // Enable clad after the header was included.
 // FIXME: The header inclusion should be made automatic if the pragma is seen.
-#pragma clad ON
+#pragma clad ON
```
vgvassilev: That should not have been changed.

include/clad/Differentiator/CladConfig.h
leetcodez: This was an automatic change applied by clang-format. I will revert it if you like the multi line style

include/clad/Differentiator/Differentiator.h
leetcodez: The previous implementation used a while loop to copy the string which I replaced with memcpy for better safety. I added this check to handle the case where malloc fails
I can remove the printf block if you like

include/clad/Differentiator/Differentiator.h
leetcodez: > Did we really have to change this?

C++11 does not have automatic return type deduction and hence we have to explicitly state it.
Without the -> decltype this code fails to compile in C++11 because the compiler cannot deduce the return type.

include/clad/Differentiator/Differentiator.h
leetcodez: I must have accidently deleted the newline at the end, I will fix this in the update

include/clad/Differentiator/Differentiator.h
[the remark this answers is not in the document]
leetcodez: Yes, the warning is valid. Since Args&& is a forwarding reference, we need std::forward to preserve the value category of the arguments. Without it, args are treated as l-values inside the function
I will update the call to use std::forward<Args>(args)....

include/clad/Differentiator/CladConfig.h
vgvassilev: We should not apply clang-format to code that was not touched.

include/clad/Differentiator/Differentiator.h
vgvassilev: That’s not part of the intended changes of the pr. You can’t just randomly change code like that.

include/clad/Differentiator/CladConfig.h
leetcodez: Understood. I will revert CladConfig.h to its original state to ensure no unrelated formatting changes are included.

include/clad/Differentiator/Differentiator.h
leetcodez: Got it. I will revert this block to the original implementation to keep this PR strictly focused on C++11 compatibility.

vgvassilev: The memcpy part can go in a separate PR.

include/clad/Differentiator/Differentiator.h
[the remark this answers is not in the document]
vgvassilev: Is that resolved? Can you rebase this PR?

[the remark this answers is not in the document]
leetcodez: Yes, it was resolved in the last commit, The rebase is also done!

PR #1716, opened 2026-01-31
Implement Multilayer Disk Offloading feature to the tape.

Vedant2005goyal:
Multilayer Storage feature offloads inactive data to the disk when the RAM gets full enabling differentiation of larger workloads which would otherwise lead to run out of memory errorand its working is similar to the paging in OS, here also the least recently used data is offloaded to the disk and fetched from the disk only when it is needed. Users can opt-in to disk offloading by setting the new template parameter to `true`. 
**Note:** By default the maximum slab size in RAM is set to 1000.

### DIFF
--- a/benchmark/MemoryComplexity.cpp
+++ b/benchmark/MemoryComplexity.cpp
@@ -1,7 +1,9 @@
 #include "benchmark/benchmark.h"
 
 #include "clad/Differentiator/Differentiator.h"
-
+#include "clad/Differentiator/Tape.h"
+#include <cstddef>
+#include <cstdint>
 namespace {
   struct MemoryManager : public benchmark::MemoryManager {
     size_t cur_num_allocs = 0;
@@ -60,21 +62,29 @@ void operator delete(void* p) noexcept {
   free(p);
 }
 
-template <typename T, std::size_t SBO_SIZE = 64, std::size_t SLAB_SIZE = 1024>
-void func(clad::tape<T, SBO_SIZE, SLAB_SIZE>& t, T x, int n) {
+template <typename T, std::size_t SBO_SIZE = 64, std::size_t SLAB_SIZE = 1024,
+          bool DiskOffload = false>
+void func(clad::tape_impl<T, SBO_SIZE, SLAB_SIZE, /*is_Multithread=*/false,
+                          DiskOffload>& t,
+          T x, int n) {
   for (int i = 0; i < n; i++)
-    clad::push<T, SBO_SIZE, SLAB_SIZE>(t, x);
+    clad::push(t, x);
 
-  for (int i = 0; i < n; i++)
-    benchmark::DoNotOptimize(clad::pop<T, SBO_SIZE, SLAB_SIZE>(t));
+  for (int i = 0; i < n; i++) {
+    benchmark::DoNotOptimize(t.back());
+    t.pop_back();
+  }
 }
 
 static void BM_TapeMemory(benchmark::State& state) {
   int block = state.range(0);
   AddBMCounterRAII MemCounters(*mm.get(), state);
   for (auto _ : state) {
-    clad::tape<double> t;
-    func<double>(t, 1, block * 2 + 1);
+    // Explicitly using false for DiskOffload to test baseline
+    clad::tape_impl<double, 64, 1024, /*is_Multithread=*/false,
+                    /*DiskOffload=*/false>
+        t;
+    func<double, 64, 1024, /*DiskOffload=*/false>(t, 1, block * 2 + 1);
   }
 }
 BENCHMARK(BM_TapeMemory)->RangeMultiplier(2)->Range(0, 4096);
@@ -84,8 +94,11 @@ static void BM_TapeMemory_Templated(benchmark::State& state) {
   int block = state.range(0);
   AddBMCounterRAII MemCounters(*mm.get(), state);
   for (auto _ : state) {
-    clad::tape<double, SBO_SIZE, SLAB_SIZE> t;
-    func<double, SBO_SIZE, SLAB_SIZE>(t, 1, block * 2 + 1);
+    clad::tape_impl<double, SBO_SIZE, SLAB_SIZE, /*is_Multithread=*/false,
+                    /*DiskOffload=*/false>
+        t;
+    func<double, SBO_SIZE, SLAB_SIZE, /*DiskOffload=*/false>(t, 1,
+                                                             block * 2 + 1);
   }
 }
 
@@ -98,8 +111,27 @@ static void BM_TapeMemory_Templated(benchmark::State& state) {
 REGISTER_TAPE_BENCHMARK(64, 1024);
 REGISTER_TAPE_BENCHMARK(32, 512);
 
-#include "BenchmarkedFunctions.h"
+// This explicitly tests the case where DiskOffload = true
+template <std::size_t SBO_SIZE, std::size_t SLAB_SIZE>
+static void BM_Multilayer_Storage(benchmark::State& state) {
+  int64_t block = state.range(0);
+  AddBMCounterRAII MemCounters(*mm, state);
+  for (auto _ : state) {
+    // Set DiskOffload = true here
+    clad::tape_impl<double, SBO_SIZE, SLAB_SIZE, /*is_Multithread=*/false,
+                    /*DiskOffload=*/true>
+        t;
+    func<double, SBO_SIZE, SLAB_SIZE, /*DiskOffload=*/true>(t, 1,
+                                                            block * 2 + 1);
+  }
+}
 
+BENCHMARK_TEMPLATE(BM_Multilayer_Storage, 64, 1024)
+    ->RangeMultiplier(2)
+    ->Range(0, 4096)
+    ->Name("BM_Multilayer_Storage/SBO_64_SLAB_1024_DISK");
+
+#include "BenchmarkedFunctions.h"
 static void BM_ReverseGausMemoryP(benchmark::State& state) {
   auto dfdp_grad = clad::gradient(gaus, "p");
   unsigned dim = state.range(0);
@@ -118,5 +150,39 @@ static void BM_ReverseGausMemoryP(benchmark::State& state) {
 }
 BENCHMARK(BM_ReverseGausMemoryP)->RangeMultiplier(2)->Range(0, 4096);
 
-// Define our main.
+const size_t TARGET_ELEMENTS = 20000;
+
+static void BM_CrashTest_OS_Paging(benchmark::State& state) {
+  AddBMCounterRAII MemCounters(*mm, state);
+  for (auto _ : state) {
+    clad::tape_impl<double, 64, 1024, /*is_Multithread=*/false,
+                    /*DiskOffload=*/false>
+        t;
+
+    for (size_t i = 0; i < TARGET_ELEMENTS; ++i) {
+      try {
+        clad::push(t, 1.0);
+      } catch (std::bad_alloc& e) {
+        state.SkipWithError("OS ran out of memory!");
+        break;
+      }
+    }
+  }
+}
+
+BENCHMARK(BM_CrashTest_OS_Paging)->Iterations(1);
+
+static void BM_CrashTest_Clad_Offload(benchmark::State& state) {
+  AddBMCounterRAII MemCounters(*mm, state);
+  for (auto _ : state) {
+    clad::tape_impl<double, 64, 1310720, /*is_Multithread=*/false,
+                    /*DiskOffload=*/true>
+        t;
+
+    for (size_t i = 0; i < TARGET_ELEMENTS; ++i)
+      clad::push(t, 1.0);
+  }
+}
+BENCHMARK(BM_CrashTest_Clad_Offload)->Iterations(1);
+
 BENCHMARK_MAIN();
--- a/include/clad/Differentiator/Tape.h
+++ b/include/clad/Differentiator/Tape.h
@@ -4,10 +4,14 @@
 #include "clad/Differentiator/CladConfig.h"
 #include <cassert>
 #include <cstddef>
+#include <cstdint>
 #include <cstdio>
+#include <fstream>
+#include <ios>
 #include <iterator>
 #include <memory>
 #include <new>
+#include <string>
 #include <type_traits>
 #include <utility>
 #ifndef __CUDACC__
@@ -16,8 +20,54 @@
 
 namespace clad {
 
+namespace detail {
+
+/// Manages offloading of data to disk when RAM capacity is exceeded.
+/// Handles files I/O operations for reading and writing slabs.
+template <typename T, std::size_t SLAB_SIZE> struct DiskManager {
+#ifndef __CUDA_ARCH__
+  std::fstream file;
+  std::string filename;
+  DiskManager() {
+    filename = "clad_tape_" + std::to_string((uintptr_t)this) + ".tmp";
+    file.open(filename, std::ios::in | std::ios::out | std::ios::binary |
+                            std::ios::trunc);
+  }
+  ~DiskManager() {
+    if (file.is_open())
+      file.close();
+    std::remove(filename.c_str());
+  }
+  std::size_t write_slab(const T* data) {
+    file.seekp(0, std::ios::end);
+    std::size_t offset = file.tellp();
+    const void* raw_data = static_cast<const void*>(data);
+    file.write(static_cast<const char*>(raw_data), SLAB_SIZE * sizeof(T));
+    return offset;
+  }
+  void read_slab(T* dest, std::size_t offset) {
+    file.seekg(offset, std::ios::beg);
+    void* raw_dest = static_cast<void*>(dest);
+    file.read(static_cast<char*>(raw_dest), SLAB_SIZE * sizeof(T));
+  }
+#else
+  CUDA_HOST_DEVICE DiskManager() {}
+  CUDA_HOST_DEVICE ~DiskManager() {}
+  CUDA_HOST_DEVICE std::size_t write_slab(const T* data) { return 0; }
+  CUDA_HOST_DEVICE void read_slab(T* dest, std::size_t offset) {}
+#endif
+};
+
+struct NoOpMutex {
+  void lock() {}
+  void unlock() {}
+  static bool try_lock() { return true; }
+};
+
+} // namespace detail
+
 template <typename T, std::size_t SBO_SIZE, std::size_t SLAB_SIZE,
-          bool is_multithread>
+          bool is_multithread, bool DiskOffload>
 class tape_impl;
 
 /// A forward iterator for traversing elements in `clad::tape_impl`.
@@ -26,9 +76,10 @@ class tape_impl;
 /// - Increment (`++`)
 /// - Equality and inequality comparisons
 template <typename T, std::size_t SBO_SIZE = 64, std::size_t SLAB_SIZE = 1024,
-          bool is_multithread = false>
+          bool is_multithread = false, bool DiskOffload = false>
 class tape_iterator {
-  using tape_t = clad::tape_impl<T, SBO_SIZE, SLAB_SIZE, is_multithread>;
+  using tape_t =
+      clad::tape_impl<T, SBO_SIZE, SLAB_SIZE, is_multithread, DiskOffload>;
   tape_t* m_tape;
   std::size_t m_index;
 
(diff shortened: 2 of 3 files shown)


benchmark/MemoryComplexity.cpp
```diff
@@ -60,21 +60,26 @@ void operator delete(void* p) noexcept {
   free(p);
 }
 
-template <typename T, std::size_t SBO_SIZE = 64, std::size_t SLAB_SIZE = 1024>
-void func(clad::tape<T, SBO_SIZE, SLAB_SIZE>& t, T x, int n) {
+template <typename T, std::size_t SBO_SIZE = 64, std::size_t SLAB_SIZE = 1024,
+          bool DiskOffload = false>
+void func(clad::tape_impl<T, SBO_SIZE, SLAB_SIZE, false, DiskOffload>& t, T x,
```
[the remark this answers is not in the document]
vgvassilev: Can we resolve these issues?

benchmark/MemoryComplexity.cpp
```diff
@@ -60,21 +60,26 @@ void operator delete(void* p) noexcept {
...
+          int n) {
   for (int i = 0; i < n; i++)
-    clad::push<T, SBO_SIZE, SLAB_SIZE>(t, x);
+    t.emplace_back(x);
```
vgvassilev: Why did we change that? Nominally, `clad::push` is the code that's generated.

benchmark/MemoryComplexity.cpp
```diff
@@ -60,21 +60,26 @@ void operator delete(void* p) noexcept {
...
 }
 
 static void BM_TapeMemory(benchmark::State& state) {
   int block = state.range(0);
   AddBMCounterRAII MemCounters(*mm.get(), state);
   for (auto _ : state) {
-    clad::tape<double> t;
-    func<double>(t, 1, block * 2 + 1);
+    // Explicitly using false for DiskOffload to test baseline
+    clad::tape_impl<double, 64, 1024, false, false> t;
```
vgvassilev:
```suggestion
    clad::tape_impl<double, 64, 1024, /*NAME_OF_PARAM=*/false, /*???=*/false> t;
```

Please update all occurrences of this pattern for the newly added code.

benchmark/MemoryComplexity.cpp
```diff
@@ -84,8 +89,8 @@ static void BM_TapeMemory_Templated(benchmark::State& state) {
   int block = state.range(0);
   AddBMCounterRAII MemCounters(*mm.get(), state);
   for (auto _ : state) {
-    clad::tape<double, SBO_SIZE, SLAB_SIZE> t;
-    func<double, SBO_SIZE, SLAB_SIZE>(t, 1, block * 2 + 1);
+    clad::tape_impl<double, SBO_SIZE, SLAB_SIZE, false, false> t;
```
vgvassilev: Might be a better idea to have a different typedef -- something like `disk_offloadable_tape` or similar.

include/clad/Differentiator/Tape.h
```diff
@@ -67,23 +72,12 @@ class tape_iterator {
...
-  struct Slab {
-    // std::aligned_storage_t<sizeof(T), alignof(T)> raw_data[SLAB_SIZE];
-    // For now use the implementation below as above implementation is not
-    // supported by c++11
+  struct RAMStorage {
     alignas(T) char raw_data[SLAB_SIZE * sizeof(T)];
-    Slab* prev;
-    Slab* next;
-    CUDA_HOST_DEVICE Slab() : prev(nullptr), next(nullptr) {}
+    CUDA_HOST_DEVICE RAMStorage() {}
```
vgvassilev, line 136: Can the user specialize these? It would be great if the user can write a "trait" class that specializes the storage kinds for further optimizations.

include/clad/Differentiator/Tape.h
```diff
@@ -93,16 +85,77 @@ class tape_impl {
...
+    }
+    std::size_t write_slab(const T* data) {
+      file.seekp(0, std::ios::end);
+      std::size_t offset = file.tellp();
+      file.write(reinterpret_cast<const char*>(data), SLAB_SIZE * sizeof(T));
+      return offset;
+    }
+    void read_slab(T* dest, std::size_t offset) {
+      file.seekg(offset, std::ios::beg);
+      file.read(reinterpret_cast<char*>(dest), SLAB_SIZE * sizeof(T));
```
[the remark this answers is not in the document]
vgvassilev: IF we are forcing `char*` why do we take a `T*`?

include/clad/Differentiator/Tape.h
```diff
@@ -123,6 +187,50 @@ class tape_impl {
 #endif
   }
 
+  void check_and_evict(std::true_type) { // Called when DiskOffload = true
```
vgvassilev: That policy could be a separate configurable type trait, too.

include/clad/Differentiator/Tape.h
```diff
@@ -170,44 +280,31 @@ class tape_impl {
...
-      // Construct element in-place
+      ensure_loaded(m_tail, std::integral_constant<bool, DiskOffload>{});
       ::new (const_cast<void*>(static_cast<const volatile void*>(
           m_tail->elements() + offset))) T(std::forward<ArgsT>(args)...);
     }
     m_size++;
   }
 
   CUDA_HOST_DEVICE std::size_t size() const { return m_size; }
-
```
vgvassilev, line 181: Please undo such changes. Perhaps you should configure your editor. The PRs should touch only the lines that are relevant.

include/clad/Differentiator/Tape.h
```diff
@@ -170,44 +269,31 @@ class tape_impl {
...
   }
 
   CUDA_HOST_DEVICE const_reference back() const {
     assert(m_size);
-    std::size_t index = m_size - 1;
-    if (index < SBO_SIZE)
-      return *(sbo_elements() + index);
-    index = (index - SBO_SIZE) % SLAB_SIZE;
-    return *(m_tail->elements() + index);
+    return const_cast<tape_impl*>(this)->operator[](m_size - 1);
```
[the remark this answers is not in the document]
vgvassilev: That's one of the cases where we can do `// NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)` because it is safe.

include/clad/Differentiator/Tape.h
```diff
@@ -170,44 +280,31 @@ class tape_impl {
           m_tail = m_tail->next;
       }
 
-      // Construct element in-place
+      ensure_loaded(m_tail, std::integral_constant<bool, DiskOffload>{});
```
vgvassilev: If that's the prefetching mechanism, we should also figure out how to make it configurable.

vgvassilev (commented):
This is heading in a good direction. Please address my comments on style, formatting, etc. The ones that are related to type traits and user configurability can be addressed in a subsequent PR.

benchmark/MemoryComplexity.cpp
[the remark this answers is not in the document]
Vedant2005goyal: Yes I have added the header file for tape in the benchmark file

include/clad/Differentiator/Tape.h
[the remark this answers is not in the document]
Vedant2005goyal: Using T* provides abstraction to low level byte calculation. It ensures the caller doesn't have to manually compute sizeof(T) * SLAB_SIZE or handle the cast at every call site. Rather than using reinterpret cast i have switched to static_cast .

benchmark/MemoryComplexity.cpp
Vedant2005goyal: The reason behind it was that the clad::push signature currently don't include DiskOffload parameter so using it would lead to an error.

include/clad/Differentiator/Tape.h
Vedant2005goyal: Sorry for that, that's because of auto formatter in my IDE, i will undo them

vgvassilev:
> Okay, I see. Do we have any real world cases where the typical RAM + swap size is not sufficient?

Some proprietary infrastructure I am aware of needs 150GB of tape with CodiPack.

include/clad/Differentiator/Tape.h
```diff
@@ -93,16 +87,88 @@ class tape_impl {
...
+
+    CUDA_HOST_DEVICE DiskStorage() { allocate(); }
+    CUDA_HOST_DEVICE ~DiskStorage() { deallocate(); }
+
+    DiskStorage(const DiskStorage&) = delete;
+    DiskStorage& operator=(const DiskStorage&) = delete;
+
+    void allocate() {
+      if (!data_ptr)
+        data_ptr = static_cast<T*>(::operator new(SLAB_SIZE * sizeof(T)));
```
parth-07, line 156: Is there any particular reason for using `::operator new` instead of just `new` here?

parth-07:
> Some proprietary infrastructure I am aware of needs 150GB of tape with CodiPack.

@vgvassilev Thank you for providing this context.

> The ciritical limitation I aim to address is Capacity not efficiency. To verify this i ran a test with workload size of 16GB of data, the results were as follow:

@Vedant2005goyal Thank you for doing this test and sharing the results. I understand the context and the use-case better now. I will need some time to properly review this change. Can you please also push BM_CrashTest_OS_Paging benchmark that you wrote as part of this PR for now?

benchmark/MemoryComplexity.cpp
vgvassilev: Then how this feature will be used by clad when generating code?

Vedant2005goyal: I have to update  clad::push function signature to support new parameter, once i do that I will revert this change in the benchmark to use clad::push again

include/clad/Differentiator/Tape.h
```diff
@@ -93,16 +87,88 @@ class tape_impl {
...
   Slab* m_tail = nullptr;
   std::size_t m_size = 0;
   std::size_t m_capacity = SBO_SIZE;
 
+  struct DiskManager {
+#ifndef __CUDA_ARCH__
+    std::fstream file;
+    std::string filename;
+    DiskManager() {
+      filename = "clad_tape_" + std::to_string((uintptr_t)this) + ".tmp";
```
parth-07: What do you think about storing this file in `/tmp` directory?

include/clad/Differentiator/Tape.h
```diff
@@ -93,16 +87,88 @@ class tape_impl {
...
+    CUDA_HOST_DEVICE DiskManager() {}
+    CUDA_HOST_DEVICE ~DiskManager() {}
+    CUDA_HOST_DEVICE std::size_t write_slab(const T* data) { return 0; }
+    CUDA_HOST_DEVICE void read_slab(T* dest, std::size_t offset) {}
+#endif
+  };
+
+  std::unique_ptr<DiskManager> m_DiskManager;
+  std::size_t m_ActiveSlabs = 0;
+  std::size_t m_MaxRamSlabs = 1000;
```
parth-07: We need these members only when the `DiskOffload` is true, right? If so, would it be possible for these members to only be present when `DiskOffload` is true?

include/clad/Differentiator/Tape.h
```diff
@@ -280,15 +376,20 @@ class tape_impl {
...
-
     for (std::size_t i = 0; i < SBO_SIZE && count > 0; ++i, --count)
       destroy_element(&sbo_elements()[i]);
 
     Slab* slab = m_head;
     while (slab) {
-      T* elems = slab->elements();
-      for (size_t i = 0; i < SLAB_SIZE && count > 0; ++i, --count)
-        destroy_element(elems + i);
+      if (!DiskOffload) {
```
parth-07: Can we use `if constexpr` here?

include/clad/Differentiator/Tape.h
```diff
@@ -123,6 +189,50 @@ class tape_impl {
...
+            v->deallocate();
+            v->is_on_disk = true;
+            m_ActiveSlabs--;
+            break;
+          }
+          v = v->next;
+        }
+      }
+      slab->allocate();
+      m_DiskManager->read_slab(slab->elements(), slab->disk_offset);
```
parth-07: Once we write to a file, we never really remove the content from the file until the file is deleted, right?

include/clad/Differentiator/Tape.h
Vedant2005goyal: Using `new` operator would perform both memory allocation and object creation whereaso `::operator new `performs only raw memory allocation and using it leads to reduced overheads by creating objects only when `push` is called.

include/clad/Differentiator/Tape.h
Vedant2005goyal: Won't moving files specifically to the `/tmp` directory can lead to problems since the path to `\tmp` is different for different system?

include/clad/Differentiator/Tape.h
Vedant2005goyal: Yes using `constexpr` would be a good choice. I will make the change.

include/clad/Differentiator/Tape.h
Vedant2005goyal: Yes, exactly. We don't remove the file once the data is read back to RAM but that file can now be overwritten. Explicitly deleting the file from the disk after reading data would lead to unnecessary overhead since OS will automatically delete the file when `Tape` is destroyed.

include/clad/Differentiator/Tape.h
Vedant2005goyal: That's a great point. Keeping them when `DiskOffload` is false is an unnecessary overhead,  I will refactor the code to ensure they are completely compiled only when `DiskOffload` is true

manorit2001:
Hi @Vedant2005goyal,

> The peak memory footprint for the OS one was ~21GB and that of clad was ~1GB. I have run this test on Macbook with 8GB of memory. The reason why the OS benchmark did not caused the system to crash is that Mac compresses the pages in RAM to accomodate more and this lead to no crash. But for the production like environment (like EC2 provided by AWS) the swap size is fixed and this feature of compression of pages in RAM is not there most of the times, so such systems would trigger OOM kill. Regarding the speed metric, OS benchmarks reveals that they run faster than Clad but this difference in time taken can be minimized by increasing the slab size which by default is 1024 bytes, I have reduced the difference by increasing the slab size to ~1MB (earlier it took more than 5minutes but later with increased slab size it took only ~33s).To be honest i dont have much idea about "Madvise" but i am assuming that it's a sort of swapping management tool, according to me it would be able to improve the efficiency within current space at max but it can not prevent OOM I see this PR as a "Safety Net." While OS paging is faster for workloads that fit in Swap, Clad's manual offloading is essential for massive simulations that exceed the system's total limit, since the diskoffloading feature can be turned off users with smaller workloads can still strictly rely on the OS for maximum performance

Not to block this PR or anything but just trying to understand the motivation, am wondering all this paging is usually an OS concept and with the usecases that you are telling, OS can be made to work in the way that you desire. 

Looking at what you have told and with your implementation, it seems like nearly 1-2GB of working ram is what clad requires to be able to perform, systemd has ways to limit the ram usage of one process (systemd-run --scope -p MemoryLimit=1536M XXX, there are more parameters as well to specify the peak memory usage where kernel should start swapping things out along with max ram), along with this, if we end up using the compressed ram pages with either zram/zwap ( zwap being the newer one https://linuxblog.io/zswap-better-than-zram/ ), am wondering we should be able to handle things right..

For reference you can look at some things that I had figured out in the past when it's not exactly possible to go and modify the source code of applications ( i was working with android studio back then ) -> https://medium.com/@manorit2001/linux-less-ram-want-more-a-guide-to-increasing-your-ram-in-linux-5d5efee6ae2a.

PS: Am not sure why would EC2 have such restrictions.. ( i haven't used it myself but can't seem to understand it.. ) if you have so much disc space to be able to hold 150GB of data then technically you should be able to create a swap out of it as well, doing a quick LLM search doesn't seem to be telling such restrictions but since it's an LLM I still don't know if it's true or not..

Vedant2005goyal:
> Hi @Vedant2005goyal,
> 
> > The peak memory footprint for the OS one was ~21GB and that of clad was ~1GB. I have run this test on Macbook with 8GB of memory. The reason why the OS benchmark did not caused the system to crash is that Mac compresses the pages in RAM to accomodate more and this lead to no crash. But for the production like environment (like EC2 provided by AWS) the swap size is fixed and this feature of compression of pages in RAM is not there most of the times, so such systems would trigger OOM kill. Regarding the speed metric, OS benchmarks reveals that they run faster than Clad but this difference in time taken can be minimized by increasing the slab size which by default is 1024 bytes, I have reduced the difference by increasing the slab size to ~1MB (earlier it took more than 5minutes but later with increased slab size it took only ~33s).To be honest i dont have much idea about "Madvise" but i am assuming that it's a sort of swapping management tool, according to me it would be able to improve the efficiency within current space at max but it can not prevent OOM I see this PR as a "Safety Net." While OS paging is faster for workloads that fit in Swap, Clad's manual offloading is essential for massive simulations that exceed the system's total limit, since the diskoffloading feature can be turned off users with smaller workloads can still strictly rely on the OS for maximum performance
> 
> Not to block this PR or anything but just trying to understand the motivation, am wondering all this paging is usually an OS concept and with the usecases that you are telling, OS can be made to work in the way that you desire.
> 
> Looking at what you have told and with your implementation, it seems like nearly 1-2GB of working ram is what clad requires to be able to perform, systemd has ways to limit the ram usage of one process (systemd-run --scope -p MemoryLimit=1536M XXX, there are more parameters as well to specify the peak memory usage where kernel should start swapping things out along with max ram), along with this, if we end up using the compressed ram pages with either zram/zwap ( zwap being the newer one https://linuxblog.io/zswap-better-than-zram/ ), am wondering we should be able to handle things right..
> 
> For reference you can look at some things that I had figured out in the past when it's not exactly possible to go and modify the source code of applications ( i was working with android studio back then ) -> https://medium.com/@manorit2001/linux-less-ram-want-more-a-guide-to-increasing-your-ram-in-linux-5d5efee6ae2a.
> 
> PS: Am not sure why would EC2 have such restrictions.. ( i haven't used it myself but can't seem to understand it.. ) if you have so much disc space to be able to hold 150GB of data then technically you should be able to create a swap out of it as well, doing a quick LLM search doesn't seem to be telling such restrictions but since it's an LLM I still don't know if it's true or not..

Hi @manorit2001 
Thanks you for stimulating the discussion and for the detailed insights.
You are absolutely right that in a fully controllable Linux environment, a skilled sysadmin could tune zswap, systemd scopes, and swap partitions to handle this. However `Clad` is a library not a standalone application and this distinction creates the need for manual `DiskOffloading`:
Clad's is used majorly by researchers, physicists and ML enthusiasts and they mostly run simulations on `HPC` and these users don't have root privileges by default(although recent efforts are made to enable only some flagship users with root privileges `https://www.computer.org/csdl/proceedings-article/cluster-workshops/2024/834500a188/21EtR2uPFHG` but the work is in very early stages). Seconldy the users which dont use clad on HPC clusers we still want clad to be platform agnostic and relying on Linux specific features like (systemd, Zswap) would leave other environments stranded. We can not expect the physicist, ML researcher or any other user to make so many changes in OS kernel to make things work and they wont be wiling to do this either, i dont think that would be a viable option so that's why we need a concrete solution to this, If they have enough disk space, your computation will finish although at expense of bit slower performance(i have mentioned earlier also that this performance gap can be minimized by making chagnes in the slab size and all) but with increased convenience.
Lastly, as i have mentioned earlier it is a functionality which can be used by user according to his workloads

manorit2001:
I see, thank you for explaining it well. As I had mentioned, no intentions of blocking the PR ( i haven't even worked on clad as such ). For reference, I come from an OS background and @parth-07 and me started having some discussion on this PR recently so was just wondering if there is some easier way to incorporate this which would be OS based ( as I personally feel that if you say that physicist, ML researcher or any other user might not be willing to do these many changes, they might just end up on some similar problem on some other project that might not be clad ), and it's not always possible that we can just go and change the codebase of any project to add some paging support ( also if each and every project goes and adds the paging support then do we even need an OS at that point to handle this job for us? ).

But given I haven't worked on clad or have no background of it, I'd let you guys only decide what's good for it. Thanks for sharing your insights, it's a fruitful discussion nonetheless to think on those problems and being platform agnostic also is important, I might be able to tell Linux related stuff but for sure we might have windows/osx users who might struggle if we end up using clad with that and then the only solution is your PR only in that case ig, so kudos for the PR. Wish there'd be a better way though ( just blabbering, ignore me )

benchmark/MemoryComplexity.cpp
```diff
@@ -118,5 +148,35 @@ static void BM_ReverseGausMemoryP(benchmark::State& state) {
...
+const size_t TARGET_ELEMENTS = 20000;
+
+static void BM_CrashTest_OS_Paging(benchmark::State& state) {
+  AddBMCounterRAII MemCounters(*mm, state);
+  for (auto _ : state) {
+    clad::tape_impl<double, 64, 1024, false, false> t;
+
+    for (size_t i = 0; i < TARGET_ELEMENTS; ++i) {
+      try {
+        clad::push(t, 1.0);
```
vgvassilev, line 164: Can we avoid relying on exceptions? Could we make clad::push return 0/1 if the push was not successful? I am also wondering if we should add another parameter to the tape limiting its maximum size. In such cases that would simplify benchmarking because we can set small limits to check things like this.

include/clad/Differentiator/Tape.h
```diff
@@ -93,16 +87,88 @@ class tape_impl {
...
+
+private:
   alignas(T) char m_static_buffer[SBO_SIZE * sizeof(T)];
 
   Slab* m_head = nullptr;
   Slab* m_tail = nullptr;
   std::size_t m_size = 0;
   std::size_t m_capacity = SBO_SIZE;
 
+  struct DiskManager {
```
vgvassilev: Here is one more specialization request/idea: in order to avoid destroying our disks when running CI or tests our benchmarks should specify a `RAMDiskManager` where no actual disk writing happens...

include/clad/Differentiator/Tape.h
```diff
@@ -217,16 +317,18 @@ class tape_impl {
...
   }
 
-  /// Remove the last value from the tape.
   CUDA_HOST_DEVICE void pop_back() {
     assert(m_size);
     m_size--;
     if (m_size < SBO_SIZE)
       destroy_element(sbo_elements() + m_size);
     else {
+      ensure_loaded(m_tail, std::integral_constant<bool, DiskOffload>{});
```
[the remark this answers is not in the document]
vgvassilev: Is that legit and now handled?

include/clad/Differentiator/Tape.h
Vedant2005goyal: I have implemented by inheriting a special block by name `detail` bcoz better alternatives like `attribute: [[no_unique_address]]` works only for the C++ newer versions.

include/clad/Differentiator/Tape.h
parth-07: Oh, I see. Thank you for the explanation.

include/clad/Differentiator/Tape.h
parth-07: Okay, I see. It seems good to me.

include/clad/Differentiator/Tape.h
parth-07: Got it. Thanks.

parth-07:
Thank you @manorit2001 for sharing your inputs. While different approaches to efficiently use memory can be incredibly helpful, they are often OS-specific :'). Perhaps it would be helpful to extend Clad documentation with OS-specific tips on how to efficiently compute large gradients for the more enthusiastic users.

parth-07 (approved):
LGTM.

This seems to be a good initial patch for adding a new core functionality. As a separate follow-up PR, we should extend the lit testing framework with a new configuration that run all the tests with `DiskOffload` as true.

include/clad/Differentiator/Tape.h
```diff
@@ -169,45 +187,31 @@ class tape_impl {
...
 
   CUDA_HOST_DEVICE const_reference back() const {
     assert(m_size);
-    std::size_t index = m_size - 1;
-    if (index < SBO_SIZE)
-      return *(sbo_elements() + index);
-    index = (index - SBO_SIZE) % SLAB_SIZE;
-    return *(m_tail->elements() + index);
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
+    return const_cast<tape_impl*>(this)->operator[](m_size - 1);
```
vgvassilev: Very likely this is the regression introducer, however, I really want to undo all changes that do not contribute to the final state and have a minimal set of changes..

benchmark/MemoryComplexity.cpp
```diff
@@ -98,8 +110,26 @@ static void BM_TapeMemory_Templated(benchmark::State& state) {
 REGISTER_TAPE_BENCHMARK(64, 1024);
 REGISTER_TAPE_BENCHMARK(32, 512);
 
-#include "BenchmarkedFunctions.h"
```
vgvassilev, line 101: This should not have moved below.

include/clad/Differentiator/Tape.h
```diff
@@ -16,8 +20,52 @@
 
 namespace clad {
 
+namespace detail {
+
+template <typename T, std::size_t SLAB_SIZE> struct DiskManager {
```
vgvassilev, line 27: We need doxygen style documentation here.

include/clad/Differentiator/Tape.h
```diff
@@ -72,55 +121,163 @@ class tape_iterator {
...
 /// slabs if capacity exceeds SBO.
 template <typename T, std::size_t SBO_SIZE = 64, std::size_t SLAB_SIZE = 1024,
-          bool is_multithread = false>
+          bool is_multithread = false, bool DiskOffload = false>
 class tape_impl {
-  /// A block of contiguous storage allocated dynamically when SBO capacity is
-  /// exceeded.
-  struct Slab {
+
+  struct RAMStorage {
```
vgvassilev, line 130: We need documentation.

include/clad/Differentiator/Tape.h
```diff
@@ -72,55 +121,163 @@ class tape_iterator {
...
   Slab* m_tail = nullptr;
   std::size_t m_size = 0;
-  std::size_t m_capacity = SBO_SIZE;
-
+  std::size_t m_Capacity = SBO_SIZE;
 #ifndef __CUDACC__
   mutable std::mutex m_TapeMutex;
 #endif
 
+  struct DiskInfo {
```
vgvassilev, line 195: We need documentation.

include/clad/Differentiator/Tape.h
```diff
@@ -72,55 +121,163 @@ class tape_iterator {
...
+    std::unique_ptr<detail::DiskManager<T, SLAB_SIZE>> m_DiskManager;
+    std::size_t m_ActiveSlabs = 0;
+    std::size_t m_MaxRamSlabs = 1000;
```
vgvassilev: Why not a power of two?

include/clad/Differentiator/Tape.h
```diff
@@ -72,55 +121,163 @@ class tape_iterator {
...
+    DiskInfo() = default;
+  };
+  struct Empty {};
+
+  using DiskInfoType = std::conditional_t<DiskOffload, DiskInfo, Empty>;
+
+  struct InternalState : public DiskInfoType {
```
vgvassilev: Is this used anywhere?

include/clad/Differentiator/Tape.h
```diff
@@ -72,55 +121,163 @@ class tape_iterator {
...
+    if (info.m_ActiveSlabs >= info.m_MaxRamSlabs) {
+      Slab* candidate = m_head;
+      while (candidate && (candidate->is_on_disk || candidate == m_tail))
+        candidate = candidate->next;
+      if (candidate) {
+        if (!info.m_DiskManager)
+          info.m_DiskManager.reset(new detail::DiskManager<T, SLAB_SIZE>());
+        candidate->disk_offset =
+            info.m_DiskManager->write_slab(candidate->elements());
+        candidate->deallocate();
```
vgvassilev, line 245: Can we add a fixme here, saying that we probably should not deallocate the slab but use a pool where old ones would be recycled?

include/clad/Differentiator/Tape.h
Vedant2005goyal: I missed that now i have made it 2^10

include/clad/Differentiator/Tape.h
Vedant2005goyal: Done

include/clad/Differentiator/Tape.h
Vedant2005goyal: I have removed it in my latest push since Internal state is just a wrapper

benchmark/MemoryComplexity.cpp
Vedant2005goyal: Originally it was below in the file and moving it up crashes the benchmarks test

include/clad/Differentiator/Tape.h
```diff
@@ -72,55 +123,166 @@ class tape_iterator {
...
   CUDA_HOST_DEVICE T* sbo_elements() {
 #if __cplusplus >= 201703L
     return std::launder(reinterpret_cast<T*>(m_static_buffer));
 #else
     return reinterpret_cast<T*>(m_static_buffer);
 #endif
-  }
 
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+    return reinterpret_cast<T*>(m_static_buffer);
```
vgvassilev, line 213: How is this not deadcode?

include/clad/Differentiator/Tape.h
```diff
@@ -72,55 +123,166 @@ class tape_iterator {
...
+  }
   CUDA_HOST_DEVICE const T* sbo_elements() const {
 #if __cplusplus >= 201703L
     return std::launder(reinterpret_cast<const T*>(m_static_buffer));
 #else
     return reinterpret_cast<const T*>(m_static_buffer);
 #endif
+
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+    return reinterpret_cast<const T*>(m_static_buffer);
```
vgvassilev, line 223: Likewise.

benchmark/MemoryComplexity.cpp
```diff
@@ -60,21 +62,29 @@ void operator delete(void* p) noexcept {
...
   int block = state.range(0);
   AddBMCounterRAII MemCounters(*mm.get(), state);
   for (auto _ : state) {
-    clad::tape<double> t;
-    func<double>(t, 1, block * 2 + 1);
+    // Explicitly using false for DiskOffload to test baseline
+    clad::tape_impl<double, 64, 1024, /*is_Multithread=*/false,
+                    /*DiskOffload=*/false>
+        t;
+    func<double, 64, 1024, false>(t, 1, block * 2 + 1);
```
vgvassilev: Please revisit all functions that pass `false` or constants and add the usual `/*param_name=*/` comment.

benchmark/MemoryComplexity.cpp
```diff
@@ -118,5 +148,35 @@ static void BM_ReverseGausMemoryP(benchmark::State& state) {
...
+        state.SkipWithError("OS ran out of memory!");
+        break;
+      }
+    }
+  }
+}
+
+BENCHMARK(BM_CrashTest_OS_Paging)->Iterations(1);
+
+static void BM_CrashTest_Clad_Offload(benchmark::State& state) {
```
vgvassilev, line 175: I am wondering how we should construct a benchmark that tests the latency of prefetching capability. That is, I'd like to test what's the efficiency of the prefetch with different strategies. At least we could compare against a separate thread that runs in the background...

vgvassilev (commented):
This is almost good to go.

benchmark/MemoryComplexity.cpp
Vedant2005goyal: Done

include/clad/Differentiator/Tape.h
Vedant2005goyal: Fixed it

benchmark/MemoryComplexity.cpp
Vedant2005goyal: Agreed but for that we need to make to introduce prefetching capability since currently it is relyin on purely synchronous I/O. Once we add that logic i think then we can add the benchmark.

vgvassilev: Ok, let's keep track of this in a new issue in our bug tracker.

vgvassilev (approved):
LGTM!

PR #1718, opened 2026-02-01
Adds benchmark to compare Tapenade and Clad performance

Vedant2005goyal:
This PR adds a benchmark for comparing the peformance of Tapenade and Clad when performing Disk OffLoading. The Tapenade is fetched only at the build time and in order to. use it the user have to pass the flag:
``-DCLAD_ENABLE_TAPENADE_BENCHMARK=ON`` 
in the Cmake build.
**Note:** Initial build can take some time as fetching the Tapenade is a long process.

### DIFF
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -7,6 +7,33 @@ CB_ADD_GBENCHMARK(ArrayExpressionTemplates ArrayExpressionTemplates.cpp)
 if (CLAD_ENABLE_ENZYME_BACKEND)
   CB_ADD_GBENCHMARK(EnzymeCladComparison EnzymeCladComparison.cpp)
 endif(CLAD_ENABLE_ENZYME_BACKEND)
+
+option(CLAD_ENABLE_TAPENADE_BENCHMARK "Enable Tapenade comparison benchmarks" OFF)
+
+if (CLAD_ENABLE_TAPENADE_BENCHMARK)
+  include(FetchContent)
+  FetchContent_Declare(
+    tapenade_kit
+    GIT_REPOSITORY https://gitlab.inria.fr/tapenade/tapenade.git
+    GIT_TAG        develop
+    GIT_SHALLOW    TRUE
+    GIT_PROGRESS   TRUE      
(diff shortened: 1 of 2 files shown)


benchmark/CMakeLists.txt
```diff
@@ -7,6 +7,33 @@ CB_ADD_GBENCHMARK(ArrayExpressionTemplates ArrayExpressionTemplates.cpp)
...
 endif(CLAD_ENABLE_ENZYME_BACKEND)
+
+option(CLAD_ENABLE_TAPENADE_BENCHMARK "Enable Tapenade comparison benchmarks" OFF)
+
+if (CLAD_ENABLE_TAPENADE_BENCHMARK)
+  include(FetchContent)
+  FetchContent_Declare(
+    tapenade_kit
+    GIT_REPOSITORY https://gitlab.inria.fr/tapenade/tapenade.git
+    GIT_TAG        develop
```
vgvassilev, line 18: We need the last release.

benchmark/CMakeLists.txt
```diff
@@ -7,6 +7,33 @@ CB_ADD_GBENCHMARK(ArrayExpressionTemplates ArrayExpressionTemplates.cpp)
 if (CLAD_ENABLE_ENZYME_BACKEND)
   CB_ADD_GBENCHMARK(EnzymeCladComparison EnzymeCladComparison.cpp)
 endif(CLAD_ENABLE_ENZYME_BACKEND)
+
+option(CLAD_ENABLE_TAPENADE_BENCHMARK "Enable Tapenade comparison benchmarks" OFF)
```
vgvassilev, line 11: We don't need to have a separate option -- we already depend on things from the internet (as fetching gbench)

benchmark/TapenadeVsclad.cpp
```diff
@@ -0,0 +1,56 @@
+#include "clad/Differentiator/Tape.h"
```
vgvassilev, line 1: This should be a part of the already existing on-disk benchmarks. If needed we can split all of them into a separate file.

PR #1723, opened 2026-02-11
Fix ConstantFolder::synthesizeLiteral fallback for unsupported types (Fixes #1073)

### LINKED ISSUE #1073: Handle more type cases in `ConstantFolder::synthesizeLiteral`
Currently any type that is not an integral type, a floating type, a boolean type or a pointer type is assigned an int value `Expr` in `ConstantFolder::synthesizeLiteral` , which could cause compilation errors. We should expand this function to handle complex types, typedefs, structs, class members etc.

Maanvi212006:
Previously, ConstantFolder::synthesizeLiteral silently synthesized an `int` literal for unsupported types such as structs, classes, and complex types. 
This could lead to invalid ASTs and compilation errors downstream.

This change removes the unsafe fallback and instead fails fast by returning nullptr for unsupported types, making constant folding behavior explicit and correct.

- Handles: Enums, pointers, booleans, integral types, floating-point types
- Unsupported types now return nullptr
- Assertion added to catch unsupported types in debug builds

No behavior is changed for supported types. This is a minimal, safe fix.

### DIFF
--- a/lib/Differentiator/ConstantFolder.cpp
+++ b/lib/Differentiator/ConstantFolder.cpp
@@ -173,7 +173,7 @@ namespace clad {
       Result = clad::synthesizeLiteral(QT, C, APVal);
     } else {
       // Unsupported type: do not silently convert to int
-      return nullptr;
+      Result =  nullptr;
     }
     assert(Result && "Unsupported type passed to ConstantFolder::synthesizeLiteral");
     return Result;
--- a/lib/Differentiator/ConstantFolder.cpp
+++ b/lib/Differentiator/ConstantFolder.cpp
@@ -172,11 +172,10 @@ namespace clad {
       llvm::APFloat APVal(C.getFloatTypeSemantics(QT), val);
       Result = clad::synthesizeLiteral(QT, C, APVal);
     } else {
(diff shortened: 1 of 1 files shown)


lib/Differentiator/ConstantFolder.cpp
```diff
@@ -172,11 +172,10 @@ namespace clad {
       llvm::APFloat APVal(C.getFloatTypeSemantics(QT), val);
       Result = clad::synthesizeLiteral(QT, C, APVal);
     } else {
-      // FIXME: Handle other types, like Complex, Structs, typedefs, etc.
-      // typecasting may be needed right now
-      Result = ConstantFolder::synthesizeLiteral(C.IntTy, C, val);
+      // Unsupported type: do not silently convert to int
+      return nullptr;
```
parth-07: Instead of returning `nullptr`, shouldn't we set `Result` to `nullptr` so that we reach the assert at line 178?

Maanvi212006: @parth-07 Thanks for the feedback!
I did the necessary changes

PR #1726, opened 2026-02-13
Use Memcpy allocation with error handling (Follow-up to PR  #1714)

leetcodez:
This is a follow-up to the discussion in PR #1714  regarding C++11 compatibility, as requested by @vgvassilev 

> The memcpy part can go in a separate PR.

**Changes**

1. Replaced Manual Loop: Switched from while (*temp++ = *code++) to std::memcpy optimized copying.
2.Added Error Handling: Introduced a check for malloc failure.

### DIFF
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -339,16 +339,11 @@ CUDA_HOST_DEVICE void push(tape<T[N], SBO_SIZE, SLAB_SIZE>& to, const U& val) {
 #endif
       size_t length = GetLength(code);
       m_Code = (char*)malloc(length + 1);
-
-      if (m_Code) {
+      if (m_Code)
         memcpy((void*)m_Code, code, length + 1);
-      } else {
-#ifdef __CUDACC__
-        printf("stderr: Error: Failed to allocate memory for m_Code\n");
-#else
+      else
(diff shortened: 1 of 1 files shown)


include/clad/Differentiator/Differentiator.h
```diff
@@ -338,10 +338,17 @@ CUDA_HOST_DEVICE void push(tape<T[N], SBO_SIZE, SLAB_SIZE>& to, const U& val) {
...
+      m_Code = (char*)malloc(length + 1);
+
+      if (m_Code) {
+        memcpy((void*)m_Code, code, length + 1);
+      } else {
+#ifdef __CUDACC__
+        printf("stderr: Error: Failed to allocate memory for m_Code\n");
+#else
+        fprintf(stderr, "Error: Failed to allocate memory for m_Code\n");
+#endif
```
vgvassilev: Can we merge these branches?

leetcodez: Everything is done on my end, please go ahead and merge.

vgvassilev: I mean can we use the fprintf unconditionally for cuda and the rest?

leetcodez: You were right.
I removed the #ifdef  and let the CI runners test it. It turns out fprintf compiles unconditionally without any issues, even on the CUDA builders.

PR #1733, opened 2026-02-18
Add derivative support for std::beta and introduce digamma function utility

leetcodez:
Added a digamma utility to into BuiltInDerivatives.h It handles the math using a standard asymptotic expansion and uses a recurrence shift for smaller values of x to ensure the floating-point results stay accurate.
implemented beta_pushforward and beta_pullback.

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -1071,6 +1071,11 @@ template <typename T> CUDA_HOST_DEVICE inline T clad_beta_primal(T x, T y) {
   return ::std::tgamma(x) * ::std::tgamma(y) / ::std::tgamma(x + y);
 #endif
 }
+// Computes the digamma function psi(x) using a standard asymptotic expansion
+// NOTE: This is a numerical approximation
+// Reference: Wolfram MathWorld, Digamma Function
+// https://mathworld.wolfram.com/DigammaFunction.html
+
 template <typename T> CUDA_HOST_DEVICE inline T clad_digamma(T x) {
   if (x <= 0.0) {
     if (x == ::std::floor(x))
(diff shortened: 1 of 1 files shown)


test/FirstDerivative/BuiltinDerivatives.C
```diff
@@ -515,6 +515,14 @@ double f_custom_min(double x, double y) { return std::min(x, y, std::greater<dou
 // CHECK-NEXT:      return _t0.pushforward;
 // CHECK-NEXT:  }
 
+double f_beta(double x, double y) { return std::beta(x, y); }
```
vgvassilev, line 529: We need to also properly update `test/Features/stl-cmath.cpp`

leetcodez: done and done

test/Features/stl-cmath.cpp
```diff
@@ -123,6 +123,14 @@
 #include <cmath>
 #include <iostream>
 #include <iomanip>
+#if !defined(__cpp_lib_math_special_functions)
+namespace std {
+  template <typename T>
+  inline T beta(T x, T y) {
+    return std::tgamma(x) * std::tgamma(y) / std::tgamma(x + y);
```
vgvassilev, line 130: We need to update the description on the top of this file accordingly.

leetcodez: I have updated the file description!

vgvassilev:
@guitargeek, can you take a look at this PR?

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -1062,6 +1062,112 @@ CUDA_HOST_DEVICE void hypot_pullback(T x, T y, U d_z, T* d_x, T* d_y) {
...
+  if (d_y)
+    pushforward += b * (clad_digamma(y) - psi_xy) * d_y;
+  return {b, pushforward};
+}
+
+template <typename T, typename U>
+CUDA_HOST_DEVICE void beta_pullback(T x, T y, U d_z, T* d_x, T* d_y) {
+  T b = clad_beta_primal(x, y);
+  T psi_xy = clad_digamma(x + y);
+  if (d_x)
```
parth-07, line 1164: Can `d_x` and `d_y` be `nullptr`?

leetcodez: Yes, If you only request the derivative with respect to one of the arguments, clad passes a nullptr for the inactive variable's adjoint. 
The if checks are there to make sure we don't segfault when trying to write to them. I just followed the same pattern used by the other _pullback functions in this file to keep things consistent.

parth-07: Oh, I see. Thanks. It seems that we are not properly adding these if-checks for many pre-existing built-in derivatives functions.

PR #1734, opened 2026-02-20
Implements Dual mode tape memory manager and custom_tape specialization hook

Vedant2005goyal:
This PR upgrades the Clad's tape architecture to a Dual mode memory offloading system the dual modes are namely RamDisk mode and the Disk offload mode.
Ram disk mode utility uses `mmap` to create zero wear Ram disks without relying on the physical system paths. 
Disk offload mode utility offloads data to disk only when the RAM gets full.
It also introduces TapeTraits and a custom_tape alias. This allows users to inject their own custom tape implementations (like a RAM disk tape) via template specialization without modifying Clad's core code generation.

### DIFF
--- a/include/clad/Differentiator/Tape.h
+++ b/include/clad/Differentiator/Tape.h
@@ -6,16 +6,25 @@
 #include <cstddef>
 #include <cstdint>
 #include <cstdio>
-#include <fstream>
-#include <ios>
+#include <cstdlib> // Added for mkstemp
 #include <iterator>
 #include <memory>
 #include <new>
 #include <string>
 #include <type_traits>
 #include <utility>
+
+#ifndef __CUDA_ARCH__
+#include <cstring>
 #ifndef __CUDACC__
 #include <mutex>
+#endif
+#ifdef _WIN32
+#include <windows.h>
+#else
+#include <sys/mman.h>
+#include <unistd.h>
+#endif
 #endif
 
 namespace clad {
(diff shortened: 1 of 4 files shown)


include/clad/Differentiator/Tape.h
```diff
@@ -57,6 +57,24 @@ template <typename T, std::size_t SLAB_SIZE> struct DiskManager {
...
+template <typename T, std::size_t SLAB_SIZE>
+struct RamDiskManager : public DiskManager<T, SLAB_SIZE> {
+#ifndef __CUDA_ARCH__
+  RamDiskManager(const std::string& custom_dir = "")
+      : DiskManager<T, SLAB_SIZE>(
+#if defined(__linux__)
+            custom_dir.empty() ? "/dev/shm/" : custom_dir // Linux RAM disk path
+#else
+            custom_dir.empty() ? "" : custom_dir // Fallback for Windows/macOS
+#endif
```
vgvassilev: We can invert the condition and this way move the path only in the `ifdef`s.

test/Features/Tape-RamDisk.cpp
```diff
@@ -0,0 +1,47 @@
...
+};
+
+namespace clad {
+  template <>
+  struct TapeTraits<float, 64, 1024, false, true> {
+    using Type = MockTape; 
+  };
+}
+
+int main() {
```
vgvassilev, line 32: Can we move the benchmarks using ramdisks now?

include/clad/Differentiator/Tape.h
```diff
@@ -518,6 +536,22 @@ class tape_impl {
       (*arr)[i].~ElTy();
   }
 };
+/// Default trait that resolves to the standard clad::tape_impl.
+/// Users can specialize this struct to inject their own custom tape
+/// implementations.
+template <typename T, std::size_t SBO_SIZE = 64, std::size_t SLAB_SIZE = 1024,
+          bool is_multithread = false, bool DiskOffload = false>
+struct TapeTraits {
```
parth-07: How can a user specify a `TapeTraits` specialization that affects all the types?

We definitely need to add a lit configuration that extends running all tests with `DiskManager` and update the Clad docs.

Vedant2005goyal: Thanks @parth-07  for the review.
> How can a user specify a `TapeTraits` specialization that affects all the types?

User can achieve this by using partial template specialization, For example:
```cpp
template <typename T, std::size_t SBO_SIZE, std::size_t SLAB_SIZE, bool is_multithread>
  struct TapeTraits<T, SBO_SIZE, SLAB_SIZE, is_multithread, true /* DiskOffload */> {
    using Type = MyCustomGlobalTape;
};
```


> We definitely need to add a lit configuration that extends running all tests with `DiskManager` and update the Clad docs.

Totally agree! I will add a lit test to ensure total coverage and i thought it would be better to make the changes in official documentation in follow up PR, Does that sound good to keep this PR focussed on the core implementation?

benchmark/CMakeLists.txt
```diff
@@ -17,7 +17,6 @@ FetchContent_Declare(
   # the 'disk offloading' feature is not yet available in the stable release of tapenade.
   GIT_TAG        6f8cb04a4168cac5f6242cdeb65f6816c413f4ee 
   
-  GIT_SHALLOW    TRUE
```
vgvassilev: Why did we change this?

include/clad/Differentiator/Tape.h
```diff
@@ -6,49 +6,161 @@
...
 template <typename T, std::size_t SLAB_SIZE> struct DiskManager {
 #ifndef __CUDA_ARCH__
-  std::fstream file;
-  std::string filename;
-  DiskManager() {
-    filename = "clad_tape_" + std::to_string((uintptr_t)this) + ".tmp";
-    file.open(filename, std::ios::in | std::ios::out | std::ios::binary |
-                            std::ios::trunc);
+  std::size_t capacity;
+  std::size_t current_offset_;
```
vgvassilev: Can we remove the trailing underscores and use the usual coding conventions?

benchmark/CMakeLists.txt
Vedant2005goyal: Since the develop branch changes with time the tip of the branch was now no longer pointing to the specific commit hash so disabling shallow makes sure we are successfully fetching the required hash

include/clad/Differentiator/Tape.h
Vedant2005goyal: Ok i will strip out the trailing underscores

benchmark/CMakeLists.txt
vgvassilev: I don't understand that but let's have that in a separate PR.

include/clad/Differentiator/Tape.h
```diff
@@ -296,6 +421,24 @@ class tape_impl {
       tape_iterator<T, SBO_SIZE, SLAB_SIZE, is_multithread, DiskOffload>;
   using const_iterator =
       tape_iterator<const T, SBO_SIZE, SLAB_SIZE, is_multithread, DiskOffload>;
+
+  void setmax_ram_slabs(std::size_t max_slabs, std::true_type) {
```
vgvassilev: These benchmark-specific things should happen in a derived class implemented in the benchmark itself.

Vedant2005goyal: Done!

test/Features/Tape-RamDisk.cpp
```diff
@@ -0,0 +1,65 @@
...
+  // Test standard tape behavior
+  clad::custom_tape<double, 64, 1024, false, true> default_tape;
```
vgvassilev, line 34: We need to figure out a way for the user to change the kind of tape clad generates as part of the derivative code.

include/clad/Differentiator/Tape.h
parth-07: > Does that sound good to keep this PR focussed on the core implementation?

Yes, it sounds perfect.

include/clad/Differentiator/Tape.h
```diff
@@ -196,6 +319,7 @@ class tape_impl {
     std::unique_ptr<detail::DiskManager<T, SLAB_SIZE>> m_DiskManager;
     std::size_t m_ActiveSlabs = 0;
     std::size_t m_MaxRamSlabs = 1024;
+    bool m_use_file_offload = false;
```
vgvassilev, line 322: We should follow the naming conventions above

include/clad/Differentiator/Tape.h
```diff
@@ -204,6 +328,11 @@ class tape_impl {
   // NOLINTNEXTLINE(readability-identifier-naming)
   DiskInfoType m_state;
 
+  DiskInfo& getDiskInfo() {
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+    return *reinterpret_cast<DiskInfo*>(&m_state);
```
vgvassilev: Can we avoid reinterpret casts?

lib/Differentiator/VisitorBase.cpp
```diff
@@ -490,6 +490,9 @@ namespace clad {
 
   TemplateDecl* VisitorBase::GetCladTapeDecl() {
     static TemplateDecl* Result = nullptr;
+    if (Result)
+      return Result;
+    Result = utils::LookupTemplateDeclInCladNamespace(m_Sema, "custom_tape");
```
vgvassilev: You are getting closer. It should check if you have such a name of a typedef called say "custom_tape" and then take the type it points to...

test/Features/Tape-RamDisk.cpp
```diff
@@ -0,0 +1,65 @@
...
+  float back() const { return 99.9f; }
+  void pop_back() {}
+};
+
+namespace clad {
+  template <>
+  struct TapeTraits<float, 64, 1024, false, true> {
+    using Type = MockTape; 
+  };
+}
```
vgvassilev, line 24:
```suggestion
namespace clad {
  using custom_tape = MockTape;
}
```

We will need a better naming but whatever. I think you are confusing youself because in your tests you define the tape yourself and test something with it. I believe that's more for the benchmarks and unittests. In the simple case I am talking about you need to force clad to generate that tape references conditionally on if that typedef is defined.

test/Analyses/TBR.cpp
```diff
@@ -48,8 +48,8 @@ double f2(double val) {
 //CHECK: void f2_grad(double val, double *_d_val) {
 //CHECK-NEXT:     int _d_i = 0;
 //CHECK-NEXT:     int i = 0;
-//CHECK-NEXT:     clad::tape<bool> _cond0 = {};
-//CHECK-NEXT:     clad::tape<unsigned {{int|long}}> _t1 = {};
+//CHECK-NEXT:     clad::custom_tape<bool> _cond0 = {};
+//CHECK-NEXT:     clad::custom_tape<unsigned {{int|long}}> _t1 = {};
```
vgvassilev: No test should change the default tape but only the one that you added.

.github/workflows/ci.yml
```diff
@@ -67,7 +67,7 @@ jobs:
         include:
           # --- LINUX: Last 10 Runtimes (12-21) ---
           # Using Ubuntu 22.04 for older runtimes, 24.04 for newer ones
-          - { name: ubu22-gcc10-runtime11, os: ubuntu-22.04, compiler: gcc-10, clang-runtime: '11' }
+          - { name: ubu22-gcc10-runtime11, os: ubuntu-22.04, compiler: gcc-10, clang-runtime: '15' }
```
vgvassilev: Why did we change this?

.github/workflows/ci.yml
```diff
@@ -624,4 +624,4 @@ jobs:
         files: ./coverage.info
         fail_ci_if_error: true
         verbose: true
-        token: ${{ secrets.CODECOV_TOKEN }}
+        token: ${{ secrets.CODECOV_TOKEN }}
```
vgvassilev: Please revert this change.

include/clad/Differentiator/Tape.h
```diff
@@ -353,15 +478,11 @@ class tape_impl {
...
   CUDA_HOST_DEVICE std::size_t size() const { return m_size; }
-
   CUDA_HOST_DEVICE iterator begin() { return iterator(this, 0); }
-
   CUDA_HOST_DEVICE const_iterator begin() const {
     return const_iterator(this, 0);
   }
-
   CUDA_HOST_DEVICE iterator end() { return iterator(this, m_size); }
-
```
vgvassilev, line 364: Please undo such whitespace changes

Vedant2005goyal: Done!

vgvassilev (approved):
This is great. LGTM, thank you, @Vedant2005goyal!

PR #1741, opened 2026-03-03
Fixes the CI failure caused by fetching the wrong commit hash.

Vedant2005goyal:
The CI currently fails with a "fatal: unable to read tree" error when attempting to check out the specific Tapenade commit hash. Because the targeted commit is no longer at the tip of the Tapenade `develop` branch, `GIT_SHALLOW TRUE` (which performs a `--depth 1` clone) fails to download the necessary git history to resolve the hash. So the fix involves disabling git shallow


vgvassilev (approved):
This will make the ci a bit slower but probably that is fine. Maybe we should monitor when that code is released and rely on a particular release. LGTM!

PR #1747, opened 2026-03-04
Template deduction failure with implicitly casted custom derivative args

### LINKED ISSUE #1043: Template custom derivatives do not get selected by overload resolution due to lost type conversions in `ImplicitCastExpr` nodes
Template custom derivatives do not get selected, when they should get selected, due to lost type conversions in `ImplicitCastExpr` nodes.

Reproducible example:

```cpp

double sum(float a, float b) {
    return a + b;
}

namespace clad {
        template <typename T>
        void sum_pullback(T a, T b, double d_y, T *d_a, T *d_b) {
            *d_a += d_y;
            *d_b += d_y;
        }
    }
}

    double res = sum(u, 7);
    return res;
}

int main() {
}
```

Ideally, `template<typename T> void sum_pullback(...)` should get selected as the custom derivative for the call `sum(u, 7)`. However, overload resolution does not find compatible custom pullback for the call `sum(u, 7)`. Why does this happen?

`7` in `sum(u, 7)` is represented as:

```
 |     `-ImplicitCastExpr 0x5eedadcbf030 <col:25> float <IntegralToFloating>
 |       `-IntegerLiteral 0x5eedadcbef50 <col:25> int 7
```

However, Clad ignores the type-conversion by `ImplicitCastExpr` and passes just `IntegerLiteral` for the `sum_pullback` call. Adjoint variable is still of type `double *`. Therefore, there is an inconsistency in the type of `T` for the custom pullback: `void sum_pullback(T a, T b, double d_y, T *d_a, T *d_b)`. Due to the inconsistency in the type of `T`, template instantiation fails. 

Perhaps, we should not ignore type-conversion by `ImplicitCastExpr` nodes.

leetcodez:
Fixes  #1043  
When passing an implicitly casted argument (like an int literal 7 to a float parameter), template deduction for custom derivatives was failing. DifferentiateCallArg was silently stripping the ImplicitCastExpr node, causing the matcher to see the wrong types and silently fall back to numerical differentiation

The Fix:
Updated ReverseModeVisitor::DifferentiateCallArg to intercept and reconstruct numeric casts using Sema::ImpCastExprToType before adding them to CallArgs. This preserves the exact types needed for template deduction while safely ignoring complex casts (like ArrayToPointerDecay) that would otherwise crash the LLVM IR generator.

Testing:
1. Added test/Gradient/ImplicitCastCustomDerivatives.C to verify Clad correctly matches the custom pullback namespace/function.
2. Added the -DCLAD_NO_NUM_DIFF flag to the test runner so it will fail rather than silently falling back if this behavior ever regresses.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2152,10 +2152,27 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       StmtDiff argDiff =
           DifferentiateCallArg(arg, PVD, PreCallStmts, /*isNonDiff=*/nonDiff,
                                isa<CUDAKernelCallExpr>(CE));
-      CallArgs.push_back(argDiff.getExpr());
+
+      Expr* finalArg = argDiff.getExpr();
+      if (const auto* ICE = dyn_cast<ImplicitCastExpr>(arg)) {
+        auto CK = ICE->getCastKind();
+        if (CK == CK_IntegralToFloating || CK == CK_FloatingCast ||
+            CK == CK_IntegralCast || CK == CK_FloatingToIntegral) {
+          if (finalArg->isGLValue()) {
+            ExprResult rvalueRes = m_Sema.DefaultLvalueConversion(finalArg);
+            if (rvalueRes.isUsable())
+              finalArg = rvalueRes.get();
+          }
+          ExprResult castExpr = m_Sema.ImpCastExprToType(
+              finalArg, ICE->getType(), CK, ICE->getValueKind());
+          if (castExpr.isUsable())
+            finalArg = castExpr.get();
+        }
+      }
+      CallArgs.push_back(finalArg);
 
       if (elideReverseForw && PVD->getType()->isIntegerType())
-        revForwAdjointArgs.push_back(argDiff.getExpr());
+        revForwAdjointArgs.push_back(finalArg);
       else
         revForwAdjointArgs.push_back(argDiff.getRevSweepAsExpr());
 
--- a/test/Regressions/issue-1043.cpp
+++ b/test/Regressions/issue-1043.cpp
@@ -1,0 +1,31 @@
+// RUN: %cladclang -I%S/../../include -DCLAD_NO_NUM_DIFF %s 2>&1 | %filecheck %s
+
+#include "clad/Differentiator/Differentiator.h"
+
+double sum(float a, float b) { 
+  return a + b; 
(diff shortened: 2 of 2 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3379,12 +3396,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
 
   StmtDiff
   ReverseModeVisitor::VisitImplicitCastExpr(const ImplicitCastExpr* ICE) {
-    // Casts should be handled automatically when the result is used by
-    // Sema::ActOn.../Build...
     StmtDiff result = Visit(ICE->getSubExpr(), dfdx());
-    // In the forward pass, builtin-type derivatives are always zero.
-    // We can replace rvalue builtin types with zeros for simplicity,
-    // e.g., `double _d_x = 0.` looks simpler than `double _d_x = _d_y`.
```
vgvassilev: Why these comments were deleted?

leetcodez: That must have been an accidental keystroke while I was scrolling through the file. I didn't even realize I had deleted them. I've restored them in the latest commit!

test/Gradient/ImplicitCastCustomDerivatives.C
```diff
@@ -0,0 +1,33 @@
...
+}
+
+double fn(double u) {
+  //Passing '7' (int) to 'float b' triggers the ImplicitCastExpr.
+  //CHECK: clad::custom_derivatives::sum_pullback
+  return sum(u, 7);
+}
+
+int main() {
+  //Trigger AST generation for FileCheck
```
vgvassilev:
```suggestion
```

test/Gradient/ImplicitCastCustomDerivatives.C
```diff
@@ -0,0 +1,33 @@
+// RUN: %cladclang %s -I%S/../../include -oImplicitCastCustomDerivatives.out 2>&1 | %filecheck %s
```
vgvassilev: We should follow the test/Regression/ pattern and name this issue-1043

leetcodez: I have updated the pattern according to the regression test files and changed the name.

test/Gradient/ImplicitCastCustomDerivatives.C
leetcodez: I have removed this comment as suggested

PR #1748, opened 2026-03-04
Add support for differentiation of functions using `std::hermite` and its variants

EuclidDivisionLemma:
Add support for differentiation of functions using `std::hermite`, `std::hermitef` and `std::hermitel` in both forward and backward modes with tests.

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -1203,7 +1203,7 @@ hermitel_pushforward(unsigned int n, long double x, unsigned int,
 template <typename T, typename U>
 CUDA_HOST_DEVICE void hermite_pullback(unsigned int n, T x, U d_z,
                                        unsigned int*, T* d_x) {
-  *d_x += 2 * n * ::std::hermite(n - 1, x) * d_z;
+  *d_x += 2 * n * clad_hermite_primal(n - 1, x) * d_z;
 }
 
 template <typename T>
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -1167,10 +1167,26 @@ CUDA_HOST_DEVICE void beta_pullback(T x, T y, U d_z, T* d_x, T* d_y) {
     *d_y += b * (clad_digamma(y) - psi_xy) * d_z;
 }
 
+template <typename T> T clad_hermite_primal(unsigned int n, T x) {
+#if defined(__cpp_lib_math_special_functions)
+  return ::std::hermite(n, x);
+#else
+
+  if (n == 0)
+    return 1;
+  else if (n == 1)
+    return 2 * x;
+  else
+    return (2 * x * clad_hermite_primal(n - 1, x)) -
+           (2 * (n - 1) * clad_hermite_primal(n - 2, x));
+#endif
+}
+
 template <typename T>
 CUDA_HOST_DEVICE ValueAndPushforward<T, T>
 hermite_pushforward(unsigned int n, T x, unsigned int, T d_x) {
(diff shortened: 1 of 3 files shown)


test/FirstDerivative/BuiltinDerivatives.C
```diff
@@ -14,6 +14,16 @@ namespace std {
   inline double beta(double x, double y) {
     return std::tgamma(x) * std::tgamma(y) / std::tgamma(x + y);
   }
+
+  template <typename T> T hermite(unsigned int n, T x) {
```
vgvassilev, line 18: I thought that's available in the standard. If so, we will need a pullback and a pushforward instead of the definition.

EuclidDivisionLemma: I already added `pullback` and `pushforward`. But on macOS, there doesn't seem to be a function called `std::hermite` as the earlier test failures indicate.

include/clad/Differentiator/BuiltinDerivatives.h
```diff
@@ -1166,6 +1166,57 @@ CUDA_HOST_DEVICE void beta_pullback(T x, T y, U d_z, T* d_x, T* d_y) {
   if (d_y)
     *d_y += b * (clad_digamma(y) - psi_xy) * d_z;
 }
+
+template <typename T> T clad_hermite_primal(unsigned int n, T x) {
```
vgvassilev, line 1170: Why do we need to this function, just call hermite and wrap that with the relevant c++ standard version checks.

EuclidDivisionLemma: Before the function was introduced, the checks failed with errors saying that `std::hermite` could not be found. We can't have the preprocessor exclude the code by adding a `__cpp_lib_math_special_functions` check because the `hermite_pushforward` and `hermite_pullback` are called by the tests. If we make the preprocessor exclude the pushforward and pullback functions, `FileCheck` gives an error. Now all the checks are successful.

EDIT: It seems that something similar is happening with `std::beta`

vgvassilev: Sure, we just need to wrap the relevant tests with the same ifdefs and add another RUN line with the proper flags.

EuclidDivisionLemma: Do you mean LLVM Lit's `UNSUPPORTED` or the `--std=c++23` flag passed to clang?

vgvassilev: No I meant we can do something line
```cpp
#ifdef __cplusplus > X
...
#endif
```

Then in the tests we add a similar ifdef and we add an extra `//RUN: ... -std=c++23` to go through that ifdef separately.

EuclidDivisionLemma: I might be wrong. But `clang`  on apple platform doesn't seem to support mathematical special functions. The tests passed on all other platforms before adding a custom definition like what is being done with `std::beta`
[https://developer.apple.com/xcode/cpp/](https://developer.apple.com/xcode/cpp/)

vgvassilev: Okay, another ifdef we should add.

EuclidDivisionLemma: Around which function do you propose to add the `ifdef`?  It can't surely be `clad_hermite_primal` because the very point of the function is to provide a definition for it on platforms that don't have one.  

If you are suggesting to remove `clad_hermite_primal` entirely, I don't think it would work as `FileCheck` doesn't respect ifdefs.

Apologies if I'm wrong.

vgvassilev: You are right that `FileCheck` does not respect ifdefs but there is a way to do this. I currently don't have bandwidth to help too much with that -- you can ask an llm and let me know if you are stuck.

PR #1749, opened 2026-03-04
Fixed nested hessian context loss and reverse mode memory OOB error for nested arrays

### LINKED ISSUE #1737: `clad::hessian` fails for nested functions
Example as a ROOT macro, including extensive validation:
```c++

#define DO_HESSIAN

double func_3(double *params)
{
   return params[0] * params[0] + params[1] * params[1] + params[0] * params[1];
}

double func_2(double *params)
{
   return func_3(params) - std::log(params[0]);
}

#pragma clad ON
void gradient_request()
{
   clad::gradient(func_2, "params");
#ifdef DO_HESSIAN
   clad::hessian(func_2, "params[0:2]");
#endif
}
#pragma clad OFF

std::vector<double> parametersVec = {0.5, -1, 2};

// To run as a ROOT macro
void ConstraintSum()
{

   auto func = [&](std::span<double> params) { return func_2(params.data()); };
      return func_2_grad(parametersVec.data(), out.data());
   };

   auto numDiff = [&](int i) {
      const double eps = 1e-6;
      double funcValDown = func(p);
      double funcValUp = func(p);
   };

   }

#ifdef DO_HESSIAN
   std::cout << "\n";

      return func_2_hessian(params.data(), out.data());
   };

   // Uses central differences:

      if (i == j) {
         const double f0 = func(p);

         const double fUp = func(p);

         const double fDown = func(p);

      } else {
         // f(x_i + eps, x_j + eps)
         const double fPP = func(p);

         // f(x_i + eps, x_j - eps)
         const double fPM = func(p);

         // f(x_i - eps, x_j + eps)
         const double fMP = func(p);

         // f(x_i - eps, x_j - eps)
         const double fMM = func(p);

      }
   };

      }
   }

      }
   }

      }
      std::cout << "\n";
   }

      }
      std::cout << "\n";
   }
#endif
}
```
The output is wrong:
```txt
Processing ConstraintSum.C...
0:
  numr : -2
  clad : -2
1:
  numr : -1.5
  clad : -1.5
2:
  numr : 0
  clad : 0

Hessian comparison (clad vs numeric vs diff):

[0,0] clad=4  num=6  diff=-2
[0,1] clad=0  num=1  diff=-1
[0,2] clad=0  num=0  diff=0
[1,0] clad=0  num=1  diff=-1
[1,1] clad=0  num=2  diff=-2
[1,2] clad=0  num=0  diff=0
[2,0] clad=0  num=0  diff=0
[2,1] clad=0  num=0  diff=0
[2,2] clad=0  num=0  diff=0

Raw Clad Hessian matrix:
4   0   0
0   0   0
0   0   0

Raw Numerical Hessian matrix:
6   1   0
1   2   0
0   0   0
```
This should work, because it's a basic requirement to implement combined likelihood fits.

Fermats-Last-Theorem:
This PR resolves the issue where Clad loses the original array dimension during Hessian calculation for nested function calls, which leads to incorrect derivatives as shown in Issue #1737 and causes OOB memory access in the reverse pass . It fixes #1737. 

When `diffplanner` computes `clad::hessian(f, "p[0:4]")`, it generates sub requests for each index. This causes the original array capacity to be lost. The forward pass then uses this wrong capacity to generate an array for this size. The reverse mode then writes cross term adjoints till the original capacity causing OOB errors.

This PR tracks total size through sub requests and generate temporary arrays of correct size. This resolves the OOB errors in the reverse pass. 

Added tests for NestedArrays.

### DIFF
--- a/include/clad/Differentiator/ParseDiffArgsTypes.h
+++ b/include/clad/Differentiator/ParseDiffArgsTypes.h
@@ -69,6 +69,8 @@ namespace clad {
     const clang::ValueDecl* param = nullptr;
     /// array index range associated with the parameter.
     IndexInterval paramIndexInterval;
+    /// track the true global capacity across sub-requests
+    std::size_t TotalCapacity = 0;
     /// Nested field information.
     llvm::SmallVector<std::string, 4> fields;
     // FIXME: Add support for differentiating with respect to array fields.
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -26,7 +26,9 @@
 #include "clang/AST/Type.h"
 #include "clang/Basic/LLVM.h"
 #include "clang/Basic/SourceLocation.h"
+#include "clang/Basic/Specifiers.h"
 #include "clang/Basic/TokenKinds.h"
+#include "clang/Basic/Version.h"
 #include "clang/Sema/Lookup.h"
 #include "clang/Sema/Overload.h"
 #include "clang/Sema/Scope.h"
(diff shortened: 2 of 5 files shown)


lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1117,9 +1119,89 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
...
+              llvm::APInt sizeAPI(
+                  m_Context.getTypeSize(m_Context.getSizeType()), arrSize);
+#if CLANG_VERSION_MAJOR >= 18
+              QualType arrType = m_Context.getConstantArrayType(
+                  elemType, sizeAPI, nullptr, clang::ArraySizeModifier::Normal,
+                  0);
+#else
+              QualType arrType = m_Context.getConstantArrayType(
+                  elemType, sizeAPI, nullptr, clang::ArrayType::Normal, 0);
+#endif
```
vgvassilev: We should use `clad_compat::getConstantArrayType` here.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1117,9 +1119,89 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
...
+
+              llvm::SmallVector<Expr*, 4> initExprs;
+              for (std::size_t k = 0; k < arrSize; ++k) {
+                Expr* idxExpr = ConstantFolder::synthesizeLiteral(
+                    m_Context.IntTy, m_Context, k);
+
+                Expr* subExpr = new (m_Context) ArraySubscriptExpr( // NOLINT
```
vgvassilev: The NOLINT(should-have-the-exact-check).

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1117,9 +1119,89 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
...
+
+      if ((!dArg || clad::utils::IsZeroOrNullValue(dArg)) &&
+          (arg->getType()->isPointerType() || arg->getType()->isArrayType())) {
+        const Expr* cleanArg = arg->IgnoreParenImpCasts();
+        if (const auto* DRE = dyn_cast<DeclRefExpr>(cleanArg)) {
+          if (const auto* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
+
+            std::size_t arrSize = 0;
+
+            /// try to get the size from the AST
```
vgvassilev: We don't need doxygen-style comments but we need full sentences.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1117,9 +1119,89 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
...
+                  initExprs.push_back(subDx);
+              }
+
+              auto* initList = new (m_Context) InitListExpr( // NOLINT
+                  m_Context, CE->getBeginLoc(), initExprs, CE->getEndLoc());
+              initList->setType(arrType);
+
+              TypeSourceInfo* tInfo =
+                  m_Context.getTrivialTypeSourceInfo(arrType);
+              auto* CLE = new (m_Context) CompoundLiteralExpr( // NOLINT
```
vgvassilev:
```suggestion
              // NOLINTNEXTLINE(...)
              auto* CLE = new (m_Context) CompoundLiteralExpr(
```

include/clad/Differentiator/ParseDiffArgsTypes.h
```diff
@@ -69,6 +69,8 @@ namespace clad {
     const clang::ValueDecl* param = nullptr;
     /// array index range associated with the parameter.
     IndexInterval paramIndexInterval;
+    /// track the true global capacity across sub-requests
+    std::size_t TotalCapacity = 0;
```
vgvassilev, line 73: Do we need this state here?

Fermats-Last-Theorem: In `DiffCollector::VisitCallExpr` we overwrite the `paramIndexInterval` for hessians which results in loss of the original size of the array which is why we need a state to retain it. The Fixme also notes this might lead to issues.
``` cpp
if (utils::isArrayOrPointerType(PVD->getType())) {
            // FIXME: We shouldn't synthesize Args strings.
            for (auto i = indexInterval.Start; i < indexInterval.Finish; ++i) {
              auto independentArgString =
                  PVD->getNameAsString() + "[" + std::to_string(i) + "]";
              forwRequest.Args = utils::CreateStringLiteral(
                  m_Sema.getASTContext(), independentArgString);
              forwRequest.UpdateDiffParamsInfo(m_Sema);
              if (!forwRequest.DVI.empty())
                forwRequest.DVI.back().TotalCapacity = indexInterval.Finish;
              LookupCustomDerivativeDecl(forwRequest);
              m_DiffRequestGraph.addNode(forwRequest, /*isSource=*/true);
            }
          }
```

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1117,9 +1119,95 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
...
+        if (const auto* DRE = dyn_cast<DeclRefExpr>(cleanArg)) {
+          if (const auto* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
+
+            std::size_t arrSize = 0;
+
+            // Try to get the size from the AST.
+            if (const auto* CAT =
+                    m_Context.getAsConstantArrayType(VD->getType())) {
+              arrSize = CAT->getSize().getZExtValue();
+            }
```
vgvassilev, line 1146: Can we test this code path?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1117,9 +1119,106 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
...
+
+            // If array size was found construct AST node for arg[k] for all k.
+            if (arrSize > 0) {
+              QualType elemType =
+                  QualType(arg->getType()->getPointeeOrArrayElementType(), 0);
+
+              llvm::APInt sizeAPI(
+                  m_Context.getTypeSize(m_Context.getSizeType()), arrSize);
+#if CLANG_VERSION_MAJOR >= 18
```
vgvassilev: When we use `clad_compat::getConstantArrayType` we don't need ifdefs.

lib/Differentiator/BaseForwardModeVisitor.cpp
Fermats-Last-Theorem: Added

lib/Differentiator/BaseForwardModeVisitor.cpp
Fermats-Last-Theorem: Sorry I missed that, fixed.

Issue #1721, opened 2026-02-07
Regression since Clad 2.1 that broke differentiating code emitted by SOFIE


guitargeek:
There is a regression that broke the proof-of-concept for differentiating though code generated by SOFIE:
https://github.com/root-project/root/pull/18332

The pattern seems to be a double-nested function and some mixture of `const` and non-`const` array inputs.

Reproducer:
```c++
#include <Math/CladDerivator.h>

#include <algorithm>
#include <cmath>
#include <vector>
#include <fstream>

inline void scalar_prod(float *output, int m, const float *a, const float *b)
{
   for (int i = 0; i < m; ++i) {
      output[i] = a[i] * b[0];
   }
}

void inner_func(float const *x, float const *theory_params, float *linear_3)
{

   float val_0[5] = {1., 1., 1., 1., 1.};
   float linear[5] = {0., 0., 0., 0., 0.};

   // This is somehow problematic
   scalar_prod(linear, 5, val_0, theory_params);

   // This would work!
   // for (int i = 0; i < 5; ++i) {
   //    linear[i] = val_0[i] * theory_params[0];
   // }

   for (int i = 0; i < 5; ++i) {
      linear_3[0] += linear[i];
   }
}

float my_func(float const *x, float const *theory_params)
{
   float out = 0.;
   inner_func(x, theory_params, &out);
   return out;
}

void cladrepro()
{
   std::vector<float> input1{5.0, 2.0, 1.0, -1.0, 1.0};
   std::vector<float> input2{0.0};

   auto func = [&](std::span<float> params) { return my_func(input1.data(), params.data()); };

   auto numDiff = [&](int i) {
      const float eps = 1e-4;
      std::vector<float> p{input2};
      p[i] = input2[i] - eps;
      float funcValDown = func(p);
      p[i] = input2[i] + eps;
      float funcValUp = func(p);
      return (funcValUp - funcValDown) / (2 * eps);
   };

   for (std::size_t i = 0; i < input2.size(); ++i) {
      std::cout << i << ":" << std::endl;
      std::cout << "  numr : " << numDiff(i) << std::endl;
   }

#if 1
   float grad_output[]{0., 0., 0., 0., 0.};
   //auto g_grad = clad::gradient<clad::opts::disable_tbr>(my_func, "theory_params");
   auto g_grad = clad::gradient(my_func, "theory_params");
   g_grad.execute(input1.data(), input2.data(), grad_output);
[...]

PR #1752, opened 2026-03-05
Fix 1721 const adjoint

fatfat123-arch:
Fixes a regression where reverse_forw could propagate const into adjoint pointer parameters (e.g. const float* _d_b), causing compilation failures for SOFIE-emitted patterns (#1721). The change ensures adjoint pointer params are writable (T*) when used as accumulators. Adds a regression test for #1721 and aligns FileCheck directives with the harness (CHECK-EXEC). ninja check-clad-regressions passes locally.

### DIFF
--- a/lib/Differentiator/ReverseModeForwPassVisitor.cpp
+++ b/lib/Differentiator/ReverseModeForwPassVisitor.cpp
@@ -12,6 +12,7 @@
 #include "clang/AST/DeclCXX.h"
 #include "clang/AST/Expr.h"
 #include "clang/AST/OperationKinds.h"
+#include "clang/AST/Type.h"
 #include "clang/Basic/LLVM.h"
 
 #include <algorithm>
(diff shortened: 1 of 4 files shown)


lib/Differentiator/ReverseModeForwPassVisitor.cpp
```diff
@@ -179,6 +185,12 @@ ReverseModeForwPassVisitor::BuildParams(DiffParams& diffParams) {
     if (it != std::end(diffParams)) {
       *it = newPVD;
       QualType dType = derivativeFnType->getParamType(dParamTypesIdx);
+
```
guitargeek: This pattern is repeated twice now:
```c+
      QualType dType = derivativeFnType->getParamType(dParamTypesIdx);

      if (dType->isPointerType() &&
          dType->getPointeeType().isConstQualified()) {
        QualType pointee = dType->getPointeeType().getUnqualifiedType();Expand commentComment on line R191Resolved
        dType = m_Context.getPointerType(pointee);
      }
```
What about factoring it out to a helper function in the anonymous namespace, that is defined in the same file? Then you can also use this helper function directly inline in `utils::BuildParmVarDecl`.

fatfat123-arch: Super happy to hear that and thank you for the review and the suggestion !!

Agreed on removing the duplication. I’ll factor this const-pointer adjustment into a helper in the anonymous namespace (same file) and reuse it in `utils::BuildParmVarDecl`, then push an update.

lib/Differentiator/ReverseModeForwPassVisitor.cpp
```diff
@@ -20,6 +20,15 @@
 using namespace clang;
 
 namespace clad {
+namespace {
+QualType getWritableAdjointParamType(ASTContext& Ctx, QualType T) {
```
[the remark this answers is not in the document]
fatfat123-arch: @guitargeek  I just pushed the last small follow-up (the clang-tidy include-cleaner fix) and re-ran the checks. This was the only additional change, so everything should run smoothly now.

lib/Differentiator/ReverseModeForwPassVisitor.cpp
```diff
@@ -20,6 +21,15 @@
...
```
vgvassilev: Can you take a look at what's in `CladUtils.cpp` and presumably use one of the utility functions there?

fatfat123-arch: Thank you for the suggestion! I just did it and pushed the changes.

I removed the local helper and switched to using the existing utility in CladUtils (`utils::getNonConstType`) for the adjoint parameter type normalization in `ReverseModeForwPassVisitor.cpp`.

I reran local checks (`check-clad-gradient`, `check-clad-regressions`, plus the #1721 targeted tests) and everything is green on my side.

PR #1755, opened 2026-03-06
Fixes the adjoint initialization for all copyable record types

### LINKED ISSUE #1750: Another regression since Clad 2.2 that broke differentiating the code emitted by TMVA SOFIE
Reproducer as a ROOT macro:
```c++

#include <vector>

extern "C" void sgemm_(const char *transa, const char *transb, const int *m, const int *n, const int *k,
                       const float *alpha, const float *A, const int *lda, const float *B, const int *ldb,
                       const float *beta, float *C, const int *ldc);

inline void My_Gemm_Call(float *output, int m, int n, int k, const float *A,
                         const float *B)
{
   char ct = 't';
   char cn = 'n';
   const int *lda = &m;
   const int *ldb = &k;
   const int *ldc = &m;
   float one = 1.;
   ::sgemm_(&cn, &cn, &m, &n, &k, &one, A, lda, B, ldb,
                                           &one, output, ldc);
}

inline void My_Gemm_Call_pullback(float *output, int m, int n, int k,
                                  const float *A, const float *B, float *_d_output,
                                  int *, int *, int *, float *_d_A, float *_d_B)
{
   _d_A[0] += _d_output[0] * B[0];
   _d_B[0] += _d_output[0] * A[0];
}

} // namespace clad::custom_derivatives

struct Session;
void doInfer(Session const &session, float const *tensor_x, float *tensor_theory_params, float *tensor_linear_3);
struct Session {
   std::vector<float> fTensor_linear = std::vector<float>(5);
   float *tensor_linear = fTensor_linear.data();

   Session() {}
};

void doInfer(Session const &session, float const *tensor_x, float *tensor_theory_params, float *tensor_linear_3)
{

   float *tensor_linear = session.tensor_linear;
   float one = 1;
   tensor_linear[0] = 1.0;

   My_Gemm_Call(tensor_linear, 1, 1, 1, &one, tensor_theory_params);
   tensor_linear_3[0] = one;
}

float my_func(Session const &session, float const *tensor_x, float *tensor_theory_params)
{
   float out = 0.;
   doInfer(session, tensor_x, tensor_theory_params, &out);
   return out;
}

void cladrepro2()
{
   std::vector<float> input1{5.0, 2.0, 1.0, -1.0, 1.0};
   std::vector<float> input2{0.0};

   Session s{};

   // Once instantiated the session object's infer method can be used
   // std::vector<float> out = s.infer(input1.data(), input2.data());

   auto func = [&](std::span<float> params) { return my_func(s, input1.data(), params.data()); };

   auto numDiff = [&](int i) {
      const float eps = 1e-4;
      std::vector<float> p{input2};
      p[i] = input2[i] - eps;
      float funcValDown = func(p);
      p[i] = input2[i] + eps;
      float funcValUp = func(p);
   };

   for (std::size_t i = 0; i < input2.size(); ++i) {
   }

   float grad_output[]{0., 0., 0., 0., 0.};
   // auto g_grad = clad::gradient<clad::opts::disable_tbr>(my_func, "tensor_theory_params");
   auto g_grad = clad::gradient(my_func, "tensor_theory_params");
   g_grad.execute(s, input1.data(), input2.data(), grad_output);
   std::fill(std::begin(grad_output), std::end(grad_output), 0);
   g_grad.execute(s, input1.data(), input2.data(), grad_output);

   std::cout << "  clad : " << grad_output[0] << std::endl;

   g_grad.dump();
}
```
The `sgemm_` comes from any BLAS library, so one needs a blas library in the `LD_LIBRARY_PATH` so it can get correctly autoloaded.
The output with Clad `master` is:
```txt
0:
  numr : 0
  clad : 1
```
The Clad derivative also needs to be zero here.

After some bisection, I figured out that the regression was introduced by commit a1f9b9d6.

Vedant2005goyal:
Commit [a1f9b9d](https://github.com/vgvassilev/clad/commit/a1f9b9d636659c8d2495fc488baa9e36c4d50fc8) broadened the adjoint initialization scope to all non aggregate copyable records. If a utility struct (like Session in the reproducer) contains a raw pointer, the newly instantiated adjoint variable shares the exact same memory address as the forward variable. Now during the reverese pass when the adjoint is Zeored out it overwrites the forward data too. So, the fix includes blocking copy init for the records containing the raw pointers or the references.
Fixes #1750

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -5,7 +5,6 @@
 //------------------------------------------------------------------------------
 
 #include "clad/Differentiator/ReverseModeVisitor.h"
-
 #include "ConstantFolder.h"
 
 #include "TBRAnalyzer.h"
@@ -430,8 +429,16 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
         // be used for the adjoint as well.
         const CXXRecordDecl* RD = VDDerivedType->getAsCXXRecordDecl();
         bool isNonAggrClass = RD && !RD->isAggregate();
+        auto hasDangerousFields = [](const CXXRecordDecl* Record) {
+          bool result = false;
+          for (const clang::FieldDecl* FD : Record->fields())
+            result = result || FD->getType()->isPointerType() ||
+                     FD->getType()->isReferenceType();
+          return result;
+        };
         bool isDirectInit = false;
-        if (isNonAggrClass && utils::isCopyable(RD)) {
+        if (isNonAggrClass && utils::isCopyable(RD) &&
+            !hasDangerousFields(RD)) {
           ParmVarDecl* newFuncParam = nullptr;
           for (auto* p : m_Derivative->parameters()) {
             if (p->getName() == param->getName()) {
--- a/test/Gradient/Pointers.C
+++ b/test/Gradient/Pointers.C
@@ -475,6 +475,15 @@ double nestedStruct(double x) {
 // CHECK-NEXT:      }
 // CHECK-NEXT:  }
 
+struct StructWithPtr {
+  double* ptr;
+};
+double pointerAliasing(double x) {
+  StructWithPtr s{&x};
+  return s.ptr[0] * 3.0; 
+}
+
+
 #define NON_MEM_FN_TEST(var)\
 res[0]=0;\
 var.execute(5,res);\
@@ -595,4 +604,9 @@ int main() {
 
   auto d_nestedStruct = clad::gradient(nestedStruct);
   NON_MEM_FN_TEST(d_nestedStruct); // CHECK-EXEC: 1.00
+
+  auto d_ptrAliasing = clad::gradient(pointerAliasing);
+  double d_x_1750 = 0;
+  d_ptrAliasing.execute(5.0, &d_x_1750);
+  printf("%.2f\n", d_x_1750); // CHECK-EXEC: 3.00
 }


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -430,8 +429,16 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+        auto hasDangerousFields = [](const CXXRecordDecl* Record) {
+          for (const clang::FieldDecl* FD : Record->fields())
+            if (FD->getType()->isPointerType() ||
+                FD->getType()->isReferenceType())
+              return true;
+          return false;
+        };
         bool isDirectInit = false;
-        if (isNonAggrClass && utils::isCopyable(RD)) {
+        if (isNonAggrClass && utils::isCopyable(RD) &&
```
vgvassilev, line 440: Why `isCopyable` fails here? I guess there is a default copy constructor but then the types will require deep copy. I wonder if that affects other uses of isCopyable.

guitargeek:
Great, this fixes the problem for me!

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: @Vedant2005goyal, ping.

lib/Differentiator/CladUtils.cpp
```diff
@@ -878,6 +878,10 @@ namespace clad {
     }
 
     bool isCopyable(const clang::CXXRecordDecl* RD) {
+      // for (const clang::FieldDecl* FD : RD->fields())
+      //   if (FD->getType()->isPointerType() ||
+      //   FD->getType()->isReferenceType())
+      //     return false;
```
vgvassilev: Do we need this commented out?

Vedant2005goyal: No we don't need them, sorry I just forgot to remove them

Issue #1736, opened 2026-02-27
Using `#pragma clad checkpoint loop` to optimize gradients breaks `clad::hessian()`


guitargeek:
This is the only blocker before we can start using Clad-generated Hessians in RooFit.

Reproducer as a ROOT macro:
```c++
// Gaussian.C

// auto-generated test macro
#include <RooFit/Detail/MathFuncs.h>
#include <Math/CladDerivator.h>

double roo_codegen_0(double *params, double const *obs, double const *xlArr)
{
   constexpr double inf = std::numeric_limits<double>::infinity();
   double RooNLLVarNewWeightSum = 0.0;
   double RooNLLVarNewResult = 0.0;
   const double t1 = (params[0] + params[1]);
   const double t3 = 1.5;
   double t2[]{params[2], t3};
   const double t4 = RooFit::Detail::MathFuncs::product(t2, 2);
   const double t6 = RooFit::Detail::MathFuncs::gaussianIntegral(-10, 10, t1, t4);
#pragma clad checkpoint loop
   for (int loopIdx1 = 0; loopIdx1 < obs[1]; loopIdx1++) {
      const double t5 = RooFit::Detail::MathFuncs::gaussian(obs[static_cast<int>(obs[0]) + loopIdx1], t1, t4);
      const double t7 = t5 / t6;
      RooNLLVarNewResult += RooFit::Detail::MathFuncs::nll(t7, obs[static_cast<int>(obs[2]) + loopIdx1], 0, 0);
   }
   return RooNLLVarNewResult;
}

#pragma clad ON
void gradient_request()
{
   clad::gradient(roo_codegen_0, "params");
   clad::hessian(roo_codegen_0, "params[0:2]");
}
#pragma clad OFF
// clang-format off

std::vector<double> parametersVec = {
    0, 1, 3
};

std::vector<double> observablesVec = {
    6, 54, 60, 54, 114, 54, -9.1, -8.1, -5.7, -5.5, 
    -5.3, -4.9, -4.5, -4.3, -3.9, -3.7, -3.5, -3.1, -2.9, -2.5, 
    -2.3, -1.7, -1.5, -1.3, -1.1, -0.9, -0.7, -0.5, -0.3, -0.1, 
    0.1, 0.3, 0.5, 0.7, 0.9, 1.1, 1.3, 1.5, 1.9, 2.1, 
[...]

PR #1756, opened 2026-03-06
Fix checkpoint loop pragma handling in hessian flows (#1736)

fatfat123-arch:
Fixes false-positive `#pragma clad checkpoint loop` diagnostics in hessian-related flows.

- I added a regression test: `test/Regressions/issue-1736.cpp`
- Also updated `tools/ClangPlugin.cpp` so dangling checkpoint pragma diagnostics are only emitted when appropriate in hessian/relevant flows, avoiding false positives during hessian scheduling

Passed locally:
- `ninja check-clad-regressions`
- `ninja check-clad-diagnostics`
- `lit -sv build/test`
- `lit -sv build/test/Regressions/issue-1736.cpp`

### DIFF
--- a/tools/ClangPlugin.cpp
+++ b/tools/ClangPlugin.cpp
@@ -17,6 +17,7 @@
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/Attr.h"
 #include "clang/AST/RecursiveASTVisitor.h"
+#include "clang/AST/Stmt.h"
 #include "clang/Basic/CodeGenOptions.h"
 #include "clang/Basic/LLVM.h" // isa, dyn_cast
 #include "clang/Basic/SourceLocation.h"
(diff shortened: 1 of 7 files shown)


tools/ClangPlugin.cpp
```diff
@@ -252,13 +255,39 @@ void InitTimers();
...
-        if (!pair.second) {
-          unsigned diagID = S.Diags.getCustomDiagID(
-              DiagnosticsEngine::Error,
-              "'#pragma clad checkpoint loop' is only allowed before a loop");
-          S.Diag(pair.first, diagID);
+        if (pair.second)
+          continue;
+
+        clang::SourceLocation AfterPragma =
+            clang::Lexer::getLocForEndOfToken(pair.first, 0, SM, LangOpts);
```
vgvassilev: There must be a better way than using a lexer for this. How does clang do it?

fatfat123-arch: Thank you for the feedback ! 

I used the lexer lookahead as a temporary minimal fix to validate the root cause quickly, but I agree we should align with Clang’s approach and avoid raw-token peeking here.

Another plan would be to bind `#pragma clad checkpoint loop` to the immediately following parsed statement and then treat it as valid only for `ForStmt` / `WhileStmt` / `DoStmt`, while keeping `diagnoseUnusedPragma` for unattached pragma.

Would you recommend implementing this through a dedicated pragma-handler/parser path, or through request-time AST association in Clad? I’d appreciate your guidance on which direction you suggest would be best.

vgvassilev: My preference is whatever is done in clang and looks like a lean and simple enough implementation.

fatfat123-arch: I updated it ! Now `#pragma clad checkpoint loop` is checked against the next parsed statement and accepted only for `for` / `while` / `do`; otherwise it emits the same diagnostic. 

Commit: 4d7b61e6

vgvassilev: Can you give the relevant clang code that does something similar?

fatfat123-arch: The relevant clang code for the similar behavior is:
- `clang/lib/Parse/ParseStmt.cpp`: `tok::annot_pragma_loop_hint` -> `ParsePragmaLoopHint(...)`
- `clang/lib/Parse/ParseStmt.cpp`: `ParsePragmaLoopHint(...)` parses the pragma, parses the immediately following statement, and prepends pragma attrs to that statement.
- `clang/lib/Sema/SemaStmtAttr.cpp`: `handleLoopHintAttr(...)` accepts only loop statements and otherwise emits `diag::err_pragma_loop_precedes_nonloop`.
- `clang/include/clang/Basic/DiagnosticSemaKinds.td`: `err_pragma_loop_precedes_nonloop`.

Issue #1700, opened 2026-01-09
Remove const_cast of a DiffRequest inside of `buildDerivedLambda`


ovdiiuv:
Inside of buildDerivedLambda there are lines:

```cpp
    auto*& FuncRef = const_cast<const FunctionDecl*&>(m_DiffReq.Function);
    llvm::SaveAndRestore<const FunctionDecl*> SaveReq(FuncRef);

                                           ...

    const_cast<DiffRequest&>(m_DiffReq).Function = LE->getCallOperator();
```

I guess making `const clang::FunctionDecl* Function = nullptr;` a stack could resolve an issue.

Issue #1699, opened 2026-01-09
Clone lambda expressions properly in the reverse mode


ovdiiuv:
Just like for other types of expressions we should clone lambdas and not reuse the original ones. This would remove const_cast from `RMV::VisitLambdaExpr`.

PR #1757, opened 2026-03-06
Fix LambdaExpr cloning and add nested request for lambda functions

Fermats-Last-Theorem:
`RMV::VisitLambdaExpr` used const_cast on the original LambdaExpr AST node reusing it. This PR Implements `StmtClone::VisitLambdaExpr` to properly clone the node.

`buildDerivedLambda` also used const_cast to mutate the `DiffRequest.Function` pointer. This PR implements a nested DiffRequest for the lambda function. Also adds tests for lambdas in global, function and lambda scope.

Fixes https://github.com/vgvassilev/clad/issues/1699
Fixes https://github.com/vgvassilev/clad/issues/1700

### DIFF
--- a/include/clad/Differentiator/DiffPlanner.h
+++ b/include/clad/Differentiator/DiffPlanner.h
@@ -286,6 +286,11 @@ public:
     bool VisitDeclRefExpr(clang::DeclRefExpr* DRE);
     bool VisitCXXConstructExpr(clang::CXXConstructExpr* e);
     bool shouldVisitImplicitCode() const { return true; }
+    /// Here we use TraverseLambdaExpr and not VisitLambdaExpr to ensure the
+    /// new nested DiffRequest is created before the visitor goes to the capture
+    /// or constructor initializers. If we use Visit they would be processed
+    /// under the parent DiffRequest which is not in the lambda scope.
+    bool TraverseLambdaExpr(clang::LambdaExpr* LE);
     bool TraverseFunctionDeclOnce(const clang::FunctionDecl* FD) {
       llvm::SaveAndRestore<bool> Saved(m_IsTraversingTopLevelDecl, false);
       if (m_Traversed.count(FD))
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1766,10 +1766,6 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     llvm::SaveAndRestore<DeclContext*> SaveContext(m_Sema.CurContext);
     llvm::SaveAndRestore<FunctionDecl*> SaveDerivative(m_Derivative);
 
-    // FIXME: Should be easy remove.
-    auto*& FuncRef = const_cast<const FunctionDecl*&>(m_DiffReq.Function);
-
-    llvm::SaveAndRestore<const FunctionDecl*> SaveReq(FuncRef);
     llvm::SaveAndRestore<Scope*> SaveFunctionScope(m_DerivativeFnScope);
     beginScope(Scope::LambdaScope | Scope::DeclScope |
                Scope::FunctionDeclarationScope | Scope::FunctionPrototypeScope);
@@ -1793,8 +1789,6 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     LSI->Lambda->setDeclContext(DC);
 
     m_Derivative = LSI->CallOperator;
-    // FIXME: Should be easy remove.
-    const_cast<DiffRequest&>(m_DiffReq).Function = LE->getCallOperator();
 
     BuildParams(params, LE);
     m_Derivative->setBody(MakeCompoundStmt({}));
@@ -1883,11 +1877,14 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
   }
 
   StmtDiff ReverseModeVisitor::VisitLambdaExpr(const LambdaExpr* LE) {
-    Expr* lambdaE = buildDerivedLambda(LE);
-    if (!m_Pullback.empty())
-      m_Pullback.pop_back();
-    // FIXME: Clone lambda properly.
-    return {const_cast<Expr*>(cast<Expr>(LE)), lambdaE};
+    DiffRequest LambdaReq = m_DiffReq;
+    LambdaReq.Function = LE->getCallOperator();
+    LambdaReq.Functor = LE->getLambdaClass();
+
+    ReverseModeVisitor NestedVisitor(m_Builder, LambdaReq);
+    Expr* lambdaE = NestedVisitor.buildDerivedLambda(LE);
+
+    return {cast<Expr>(Clone(LE)), lambdaE};
   }
 #endif // CLANG_VERSION_MAJOR
   StmtDiff ReverseModeVisitor::VisitCallExpr(const CallExpr* CE) {
--- a/test/Gradient/Lambdas.C
+++ b/test/Gradient/Lambdas.C
@@ -67,6 +67,111 @@ double f2(double i, double j) {
 // CHECK-NEXT:     }
 // CHECK-NEXT: }
 
+double f3(double i, double j) {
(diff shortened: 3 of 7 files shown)


include/clad/Differentiator/VisitorBase.h
```diff
@@ -128,6 +129,8 @@ namespace clad {
     clang::FunctionDecl* m_Derivative;
     /// The differentiation request that is being currently processed.
     const DiffRequest& m_DiffReq;
+    /// The function currently being differentiated
+    const clang::FunctionDecl* m_CurrentFunction = nullptr; // NOLINT
```
vgvassilev: Why do we need more state? Note here it can be either a function or a functor. We somehow need to figure out how to consolidate this in the DiffRequest.

Fermats-Last-Theorem: The issue with the way clad used to handle lambdas was that it was using `const_cast` on `m_DiffReq` to mutate the function context from the Lambda functor to the internal `operator()`. 
In #1700, ovdiiuv suggested that since lambdas and functors require current Function context changes making `m_DiffReq.Function` a stack could fix the issue. I do the same thing but I rely on `SaveAndRestore` as the implicit stack to change the function context. Since the current function context is not always the same as the original function context in DiffReq, I added `m_CurrentFunction` state in VisitorBase to track that context while traversing the AST. 
I could try to consolidate this in m_DiffReq by making `m_DiffReq.Function` mutable or making it a stack like ovdiiuv suggested.

vgvassilev: I understand that, however that’s not a fix but a workaround. Instead these things need to be decided in the DiffPlanner a stored in the DiffRequest. In a very few cases new members are required and I do not see that as one of them.

Fermats-Last-Theorem: Makes sense. So as I understand it, when dealing with functors/lambda one way to deal with this could be to make a new DiffRequest for the inner operator() function separate from the original DiffRequest for the functor.
Does this seem like a more appropriate fix?

vgvassilev: Yes, that might be the better fix indeed. Somehow a nested request for the lambda functions. That would also nest nicely -- we should add a test with a lambda on the global scope, lambda in a function scope and lambda inside a lambda.

Fermats-Last-Theorem:
Hi @vgvassilev. Can you please have a look at the new changes?
Now when DiffCollector crosses a lambda function boundary it creates a new `DiffRequest` for the internal function context and makes it active. I have also added tests for function/global/lambda scope.

include/clad/Differentiator/DiffPlanner.h
```diff
@@ -285,6 +285,7 @@ struct DiffRequest {
     bool VisitCallExpr(clang::CallExpr* E);
     bool VisitDeclRefExpr(clang::DeclRefExpr* DRE);
     bool VisitCXXConstructExpr(clang::CXXConstructExpr* e);
+    bool TraverseLambdaExpr(clang::LambdaExpr* LE);
```
vgvassilev: Why is this not Visit?

Fermats-Last-Theorem: That's because VisitLambdaExpr runs after the traversal has already started. If we used VisitLambdaExpr then things like capture list and constructor initilization would still happen in the original diffRequest. TraverseLambdaExpr ensures that the new diffRequest is made before anything internal is traversed.

vgvassilev: I am not sure I entirely understand that but we should capture this in a comment in the source code.

Fermats-Last-Theorem: I have added the comment explaining the reason.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -641,8 +641,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
 
     // First add the function itself as a parameter/argument
     // FIXME: We should not use const_cast to get the decl context here.
-    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
-    enzymeArgs.push_back(
+    enzymeArgs.push_back( // NOLINT(cppcoreguidelines-pro-type-const-cast)
```
vgvassilev: This seems changed unintentionally.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1883,11 +1876,14 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
-    if (!m_Pullback.empty())
-      m_Pullback.pop_back();
-    // FIXME: Clone lambda properly.
-    return {const_cast<Expr*>(cast<Expr>(LE)), lambdaE};
+    DiffRequest LambdaReq = m_DiffReq;
+    LambdaReq.Function = LE->getCallOperator();
+    LambdaReq.Functor = LE->getLambdaClass();
+
+    ReverseModeVisitor NestedVisitor(m_Builder, LambdaReq);
+    Expr* lambdaE = NestedVisitor.buildDerivedLambda(LE);
```
vgvassilev, line 1885: Nice!

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -642,7 +642,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     // First add the function itself as a parameter/argument
     // FIXME: We should not use const_cast to get the decl context here.
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
-    enzymeArgs.push_back(
+    enzymeArgs.push_back(
```
vgvassilev: That's still changed. Please revert this change.

test/Gradient/Lambdas.C
```diff
@@ -78,4 +183,24 @@ int main() {
...
+  di = 0, dj = 0;
+  df4.execute(1, 1, &di, &dj);
+  printf("%.2f %.2f\n", di, dj);              // CHECK-EXEC: 1.00 2.00
+
+  auto df5 = clad::gradient(_global_f);
+  double dt = 0; 
+  df5.execute(4, &dt);
+  printf("%.2f\n", dt);                       // CHECK-EXEC: 3.00
+
+  auto df6 = clad::gradient(f6);
```
vgvassilev, line 202: We should make sure that clad::differentiate works too.

test/Gradient/Lambdas.C
```diff
@@ -67,6 +67,111 @@ double f2(double i, double j) {
...
+// CHECK-NEXT:         *_d_i += 1;
+// CHECK-NEXT:         double _r0 = 0.;
+// CHECK-NEXT:         _d__f(j, 1, &_r0);
+// CHECK-NEXT:         *_d_j += _r0;
+// CHECK-NEXT:     }
+// CHECK-NEXT: }
+
+double f4(double i, double j) {
+  double c = 3.0;
+  double d = 4.0;
```
vgvassilev, line 100: Can we add a case where captures are used as part of the computation in a non-linear way?

Fermats-Last-Theorem: Sorry for the delayed reply. I tried to add a test which uses the captured variable in both forward and reverse mode. 
For reverse mode, clad currently generates the pullback with an empty capture list even with explicit captures so anything that is used in the lambda gets this error `variable 'c' cannot be implicitly captured`. (this is what I mentioned earlier)
For forward mode, explicit captures can be made but they can't be used inside the lambda scope because the captured pushforward variable is still pointing to the original variable. And we get, `reference to local variable 'c' declared in enclosing function 'f5'`. 
I could try to address the issues but should they be a part of this pr?

Fermats-Last-Theorem: I have added a test in forward mode which captures a variable as that was missing

vgvassilev: Can you open issues for fixing the captures that you reported did not work?

Fermats-Last-Theorem: I opened #1769

Issue #1754, opened 2026-03-05
`clad::pullback`?


guitargeek:
In the general case, one wants to be able to pass custom data classes to functions differentiate by Clad, like in this ROOT script example:
```c++
#include <Math/CladDerivator.h>

class MyStruct {
public:
    MyStruct() {}
    double factor = 5;
};

double func(MyStruct const &myStruct, double x)
{
    return myStruct.factor * x;
}

void repro() {
    clad::gradient(func, "x").dump();
}
```
That works well, and the dumped gradient is:
```c++
void func_grad_1(const MyStruct &myStruct, double x, double *_d_x) {
    MyStruct _d_myStruct = {};
    {
        _d_myStruct.factor += 1 * x;
        *_d_x += myStruct.factor * 1;
    }
}
```

One can see that a corresponding data structure for the backwards pass has to be constructed.

But what if the structure has no default constructor, e.g. this?
```c++
#include <Math/CladDerivator.h>

class MyStruct {
public:
    MyStruct(double f) : factor{f} {}
    double factor;
};

double func(MyStruct const &myStruct, double x)
{
    return myStruct.factor * x;
}

void repro() {
    clad::gradient(func, "x").dump();
}
````
This is currently failing:
```txt
/home/rembserj/code/root/root_src/repro.C:9:8: error: no matching constructor for initialization of 'MyStruct'
double func(MyStruct const &myStruct, double x)
       ^
/home/rembserj/code/root/root_src/repro.C:5:5: note: candidate constructor not viable: requires single argument 'f', but no arguments were provided
    MyStruct(double f) : factor{f} {}
    ^        ~~~~~~~~
/home/rembserj/code/root/root_src/repro.C:3:7: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 0 were provided
class MyStruct {
      ^~~~~~~~
/home/rembserj/code/root/root_src/repro.C:3:7: note: candidate constructor (the implicit move constructor) not viable: requires 1 argument, but 0 were provided
class MyStruct {
      ^~~~~~~~
```
The error is understandable. But actually the pullback should be possible to automatically be generated, because it will take the structure for the backwards pass as an additional `_d_myStruct` argument, which I can create before calling the pullback manually and then get the gradient like this.

Can we add an interface to Clad to generate the pullback of the primal function for this purpose? That would make this usecase of the non-default constructible data classes possible.

It's not super urgent because I have a workaround:
  * Add a dummy default constructor
[...]

PR #1760, opened 2026-03-07
Introduce clad::pullback API and dead adjoint elimination

leetcodez:
This PR introduces the clad::pullback API to the Reverse Mode AD engine. A pullback executes the reverse-mode automatic differentiation pass but intentionally omits the final gradient extraction step.

This is highly valuable for performance optimization when chaining derivatives, computing Vector-Jacobian Products, or interfacing with external Machine Learning frameworks (such as PyTorch). By skipping the final gradient extraction, we allow external systems to efficiently consume the intermediate adjoints directly.

(Note: The dead-adjoint elimination optimization for non-copyable/const structures discussed in #1754 has been moved to a dedicated follow-up PR per review feedback).

**Key Changes:**

- API Addition: Introduced clad::pullback template overloads in Differentiator.h (tagged with "P") and added user-facing Doxygen documentation.
- Core Engine Updates: Extended ReverseModeVisitor and DerivativeBuilder to support pullback generation, bypassing the hardcoded 1.0 seed injection when a pullback is requested.
- Type Resolution: Updated VisitorBase::GetDerivativeType to align the generated function blueprint with the dynamic seed parameter (_d_return).
- Memory Management: Ensured that tape memory allocations and the Tape Record phase are correctly preserved during pullback execution.

**Testing Performed:**

- Added/Updated FileCheck tests (Assignments.C, DiffInterface.C, Cladtorch.C, GradientKernels.cu) to verify correct AST generation and safe tape memory management for the new API.

### DIFF
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -713,6 +713,61 @@ T& back(
         derivedFn /* will be replaced by gradient*/, code, f);
   }
 
+  /// Generates the pullback of the specified function.
+  ///
+  /// A pullback executes the reverse-mode automatic differentiation pass but
+  /// omits the final gradient extraction step. This allows for the computation
+  /// of intermediate adjoints. It is highly efficient when chaining derivatives
+  /// or when interfacing with external Machine Learning frameworks.
+  ///
+  /// \param f The function or method to be differentiated.
+  /// \param args (Optional) A string literal with comma-separated names of
+  ///             independent variables (e.g. "x" or "x, y"). If not provided,
+  ///             the pullback is generated with respect to all arguments.
+  template <unsigned... BitMaskedOpts, typename ArgSpec = const char*,
+            typename F, typename DerivedFnType = GradientDerivedFnTraits_t<F>,
+            typename = typename std::enable_if<
+                !clad::HasOption(GetBitmaskedOpts(BitMaskedOpts...),
+                                 opts::immediate_mode) &&
+                !std::is_class<remove_reference_and_pointer_t<F>>::value>::type>
+  constexpr CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>,
+                         true> __attribute__((annotate("P"))) CUDA_HOST_DEVICE
+  pullback(F f, ArgSpec args = "",
+           DerivedFnType derivedFn = static_cast<DerivedFnType>(nullptr),
+           const char* code = "", bool CUDAkernel = false) {
+    return CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>, true>(
+        derivedFn /* will be replaced by pullback*/, code, nullptr, CUDAkernel);
+  }
+
+  template <unsigned... BitMaskedOpts, typename ArgSpec = const char*,
+            typename F, typename DerivedFnType = GradientDerivedFnTraits_t<F>,
+            typename = typename std::enable_if<
+                clad::HasOption(GetBitmaskedOpts(BitMaskedOpts...),
+                                opts::immediate_mode) &&
+                !std::is_class<remove_reference_and_pointer_t<F>>::value>::type>
+  constexpr CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>, true,
+                         true> __attribute__((annotate("P"))) CUDA_HOST_DEVICE
+  pullback(F f, ArgSpec args = "",
+           DerivedFnType derivedFn = static_cast<DerivedFnType>(nullptr),
+           bool CUDAkernel = false) {
+    return CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>, true, true>(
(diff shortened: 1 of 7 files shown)


include/clad/Differentiator/DiffPlanner.h
```diff
@@ -88,6 +88,8 @@ struct DiffRequest {
   /// If function appears in the call to clad::gradient/differentiate,
   /// the call must be updated and the first arg replaced by the derivative.
   bool CallUpdateRequired = false;
+  /// A flag specifying if the user directly requested a raw pullback.
+  bool IsPullback = false;
```
vgvassilev: Why do we need this state. This can be kept in the DiffRequest only.

include/clad/Differentiator/Differentiator.h
```diff
@@ -716,6 +713,51 @@ T& back(
...
+  /// \brief Creates a raw pullback function for the specified calable.
+  template <unsigned... BitMaskedOpts, typename ArgSpec = const char*,
+            typename F, typename DerivedFnType = GradientDerivedFnTraits_t<F>,
+            typename = typename std::enable_if<
+                !clad::HasOption(GetBitmaskedOpts(BitMaskedOpts...),
+                                 opts::immediate_mode) &&
+                !std::is_class<remove_reference_and_pointer_t<F>>::value>::type>
+  constexpr CladFunction<DerivedFnType, ExtractFunctorTraits_t<F>,
+                         true> __attribute__((annotate("P"))) CUDA_HOST_DEVICE
+  pullback(F f, ArgSpec args = "",
```
vgvassilev, line 735: We will need a pushforward interface, too.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -399,6 +392,8 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       // independent variables (args).
       for (const ParmVarDecl* param : m_NonIndepParams) {
         QualType paramTy = param->getType();
+        if (paramTy.getNonReferenceType().isConstQualified())
```
vgvassilev: That needs to go in a separate PR.

include/clad/Differentiator/Differentiator.h
leetcodez: Would you prefer I add the pushforward API and its tests in this PR, or should I open a dedicated follow-up PR

lib/Differentiator/ReverseModeVisitor.cpp
leetcodez: Got it, I'll make a seperate PR

vgvassilev (commented):
This looks very good. Can you make the relevant documentation changes and squash all in a single commit with a proper commit log and PR description?

test/Gradient/DiffInterface.C
```diff
@@ -147,5 +152,7 @@ int main () {
   f2_dX.execute(2, 3, &result[0]);
   printf("{%.2f}\n", result[0]); // CHECK-EXEC: {1.00}
 
+  auto f3_pb = clad::pullback(f_3, "x, y");
```
parth-07, line 159: We should probably add CHECKs to verify that pullback is working as expected.

test/Gradient/DiffInterface.C
```diff
@@ -93,7 +93,12 @@ double f_2 (const double& x, const double& y) {
     F.execute(0, 0, 0, __VA_ARGS__);                                           \
     printf("{%.2f, %.2f, %.2f}\n", result[0], result[1], result[2]);           \
   }
+double f_3(double x, double y) {
+   return 2 * x + y;
+}
 
+//CHECK: void f_3_pullback(double x, double y, double _d_y0) {
+//CHECK-NEXT: }
```
parth-07, line 105: Why is the pullback body empty?

leetcodez:
@parth-07 
The empty body issue is fixed
The Fix: 
-DiffPlanner.cpp was previously returning early and skipping the string parsing for DiffMode::pullback.
-I updated UpdateDiffParamsInfo with an && !Args check so it now correctly parses explicit user arguments (like "x, y") while safely bypassing internal, arg-less pullbacks.

AST Testing: The generated AST is now perfectly correct, and DiffInterface.C passes the //CHECK assertions locally on LLVM 18.
Execution Testing (Commented Out): I wrote the runtime execution logic in the test but left it commented out for now. Currently, CladFunction::execute_helper throws a DropArgs template deduction error when trying to handle the pullback's seed parameter and output pointers. I figured extending the execution wrapper should probably be tackled in a separate PR.

Note on CI Failures:
The CUDA/Thrust tests are failing with cudaErrorSystemDriverMismatch (hardware/driver issue on the runner).
Several tests (like ArrayInputsReverseMode.C) are dumping core exclusively on the ubu24-clang20-runtime21-debug runner due to an upstream assertion in Clang 21

leetcodez:
Hi @vgvassilev,
any updates on the PR review?

PR #1772, opened 2026-03-12
fix: use adjoint while move it

fogsong233:
I have currently implemented a fix with maximum compatibility by creating a copy. However, xvalues are still being used within the function; this incorrect usage of rvalues is a systemic issue throughout the entire workflow.

It appears that Clad is not correctly perceiving the move operation, which prevents the proper transformation of the adjoint. This suggests a potential design conflict between the semantics of a 'move' and the requirements of automatic differentiation

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -4669,6 +4669,22 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     // SomeClass _d_c = _t0.adjoint;
     // SomeClass c = _t0.value;
     // ```
+    // Copy/move constructors consume their source objects, so replaying them
+    // through constructor_reverse_forw can invalidate the same values that the
+    // pullback still needs to use. Keep them on the pre-9ed0b9e9 path.
+    if (CD->isCopyOrMoveConstructor()) {
+      if (RD->isAggregate())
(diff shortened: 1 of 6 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1718,7 +1718,14 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     }
     if (argDiff.getExpr_dx() && arg->isXValue() &&
         argDiff.getExpr_dx()->isLValue()) {
-      llvm::SmallVector<Expr*, 1> moveArg = {argDiff.getExpr_dx()};
+      Expr* revSweepAdjoint = Clone(argDiff.getExpr_dx());
```
vgvassilev: We will need a test for this.

fogsong233: Maybe we should also need change the exsiting test,which checks generated code as before

guitargeek (changes requested):
Needs rebase

vgvassilev:
Ok, then we will need a fix to move forward. Eg. we need to resolve this as either part of this PR or a separate PR.

fogsong233:
@vgvassilev  It seems that failed checks  is not related to this pr. One is run the previously benchmark, which fails as expected, another may be the problem of envs

Issue #1765, opened 2026-03-08
Resolve valgrind issues for clad on llvm21


vgvassilev:
[Here](https://github.com/vgvassilev/clad/pull/1707) the valgrind builds fail on llvm21 because there are some outstanding valgrind issues we need to fix in clad's codebase.

vgvassilev:
We don't need to deal with the memory leaks per se. There are other things with llvm 21 that are more severe.

PR #1778, opened 2026-03-13
fix issue #1765 - Resolve valgrind issues for clad on llvm21

Lucky-Lodhi2004:
In lib/Differentiator/ErrorEstimator.cpp, FunctionProtoType::ExtProtoInfo EPI was not initialized, I initialized the object in this PR.
now all fields in ExtProtoInfo are zero initialized.

### DIFF
--- a/lib/Differentiator/ErrorEstimator.cpp
+++ b/lib/Differentiator/ErrorEstimator.cpp
@@ -464,7 +464,7 @@ void ErrorEstimationHandler::LookupCustomErrorFunction() {
   if (R.empty())
     return;
 
-  FunctionProtoType::ExtProtoInfo EPI;
+  FunctionProtoType::ExtProtoInfo EPI{};
   QualType ConstCharPtr = C.getPointerType(C.getConstType(C.CharTy));
   QualType DoubleTy = C.DoubleTy;
   llvm::SmallVector<QualType, 3> FnTypes = {DoubleTy, DoubleTy, ConstCharPtr};


lib/Differentiator/ErrorEstimator.cpp
```diff
@@ -464,7 +464,7 @@ void ErrorEstimationHandler::LookupCustomErrorFunction() {
   if (R.empty())
     return;
 
-  FunctionProtoType::ExtProtoInfo EPI;
+  FunctionProtoType::ExtProtoInfo EPI{};
```
parth-07, line 467: Can you please describe how does this affect behavior? Wouldn't the list-initialization (`{}`) here just call the default constructor and thus have the exact same behavior as the original code?

vgvassilev: @Lucky-Lodhi2004, ping.

Lucky-Lodhi2004: yea i am looking into it.

Lucky-Lodhi2004: The valgrind warnings are likely caused by garbage initialization and paddings etc.

I have looked extensively regarging {} initialization of a struct in c++ and here's what i found. C++ rules out that if a structure has user defined constructor, both `EPI` and `EPI{}` will resolve to calling the user defined constructor. If user doesn't provide a constructor, the compiler calls default constructor. The default constructor will (or i should say may) then initialize all members to zero and clear any padding in memory.

I checked the exact definition of `ExtProtoInfo` structure in latest repository of clang, and here's what the definition looks like:

```C++
struct ExtProtoInfo {
    FunctionType::ExtInfo ExtInfo;
    Qualifiers TypeQuals;
    RefQualifierKind RefQualifier = RQ_None;
    ExceptionSpecInfo ExceptionSpec;
    const ExtParameterInfo *ExtParameterInfos = nullptr;
    SourceLocation EllipsisLoc;
    FunctionEffectsRef FunctionEffects;
    FunctionTypeExtraAttributeInfo ExtraAttributeInfo;

    LLVM_PREFERRED_TYPE(bool)
    unsigned Variadic : 1;
    LLVM_PREFERRED_TYPE(bool)
    unsigned HasTrailingReturn : 1;
    LLVM_PREFERRED_TYPE(bool)
    unsigned CFIUncheckedCallee : 1;
    LLVM_PREFERRED_TYPE(AArch64SMETypeAttributes)
    unsigned AArch64SMEAttributes : 9;

    ExtProtoInfo()
        : Variadic(false), HasTrailingReturn(false), CFIUncheckedCallee(false),
          AArch64SMEAttributes(SME_NormalFunction) {}

    ExtProtoInfo(CallingConv CC)
        : ExtInfo(CC), Variadic(false), HasTrailingReturn(false),
          CFIUncheckedCallee(false), AArch64SMEAttributes(SME_NormalFunction) {}

    ExtProtoInfo withExceptionSpec(const ExceptionSpecInfo &ESI) {
      ExtProtoInfo Result(*this);
      Result.ExceptionSpec = ESI;
      return Result;
    }
```

here we can see that user defined constructor is there, so EPI and EPI{} both resolve to user defined constructor. So you are right here. But this constructor only initializes : 

```C++
Variadic
HasTrailingReturn
CFIUncheckedCallee
AArch64SMEAttributes
```

The following are not initialized and may not have proper initialization : 

```C++
ExtInfo
TypeQuals
RefQualifier
ExceptionSpec
ExtParameterInfos
EllipsisLoc
FunctionEffects
ExtraAttributeInfo
```

Using {} may cause the compiler to initialize them with zero and clear any paddings, but this solution may not be applicable universally. So I propose the following solution :- 

```C++
FunctionProtoType::ExtProtoInfo EPI;
std::memset(&EPI, 0, sizeof(EPI));
EPI = FunctionProtoType::ExtProtoInfo();
```

i.e. setting bytes to zero and then setting constructor defaults. In this PR I am continuing with {} solution and I am making a separate PR for memset solution.

guitargeek:
Duplicate of #1797

PR #1794, opened 2026-03-23
Fix reverse-mode scheduling for integral-return helper calls (issue #1793)

### LINKED ISSUE #1793: Regression since Clad 2.1 that broke differentiating code emitted by CMS Combine
Similar to https://github.com/vgvassilev/clad/issues/1721, but with a reproducer from a CMS Combine usecase:
```c++

unsigned int rawBinNumber(double x, double const *boundaries, std::size_t nBoundaries)
{
   double const *end = boundaries + nBoundaries;
   double const *it = std::lower_bound(boundaries, end, x);
   // double const* it = boundaries;
   //  always return valid bin number
   while (boundaries != it && (end == it || end == it + 1 || x < *it)) {
      --it;
   }
   return it - boundaries;
}

double roo_codegen_0(double *params, const double *obs, const double *xlArr)
{
   double out = 0.;
   double t23[5]{1. + params[0], 1. + params[1], 1. + params[2], 1. + params[3], 1. + params[4]};
   for (int loopIdx1 = 0; loopIdx1 < 5; loopIdx1++) {
      const double t215 = t23[rawBinNumber(obs[loopIdx1], xlArr, 6)];
      out += t215;
   }
   return out;
}

#pragma clad ON
void gradient_request()
{
   clad::gradient(roo_codegen_0, "params");
}
#pragma clad OFF
// clang-format off

std::vector<double> parametersVec = {0, 0, 0, 0, 0};

std::vector<double> observablesVec = {
    237.5, 275, 350, 500, 800
};

std::vector<double> auxConstantsVec = {
    225, 250, 300, 400, 600, 1000
};

// clang-format on

// To run as a ROOT macro
void cladrepro()
{

   auto func = [&](std::span<double> params) {
      return roo_codegen_0(params.data(), observablesVec.data(), auxConstantsVec.data());
   };
      return roo_codegen_0_grad_0(parametersVec.data(), observablesVec.data(), auxConstantsVec.data(), out.data());
   };

   auto numDiff = [&](int i) {
      const double eps = 1e-6;
      double funcValDown = func(p);
      double funcValUp = func(p);
   };

   }
}
```

The output is:
```txt
root [0]
Processing clad_regression.C...
In file included from input_line_8:1:
/home/rembserj/tmp/clad_regression.C:3:14: error: cannot initialize a variable of type 'double *' with an rvalue of type 'const double *'
unsigned int rawBinNumber(double x, double const *boundaries, std::size_t nBoundaries)
             ^

```

FYI @fatfat123-arch, @vgvassilev 

Thanks to @maxgalli for originally spotting the problem.

fatfat123-arch:
This PR : 

- mark nested reverse-mode call requests as non-differentiable when the callee returns an integral/enumeration type and has no memory-type parameters
- avoid scheduling unnecessary pullbacks for helper index functions like the CMS  pattern
- add regression 
- all tests run green locally 

Fixes #1793

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -1191,6 +1191,19 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
       if (!utils::hasMemoryTypeParams(FD) && hasPointerOrRefReturn &&
           m_TopMostReq->Mode == DiffMode::reverse)
         nonDiff = true;
+      // Skip reverse-mode scheduling for integral-return helper calls that
+      // cannot accumulate through memory arguments. Keep this narrow to avoid
+      // suppressing diagnostics on variadic/non-helper calls.
+      const bool HasPointerOrReferenceParam = std::any_of(
+          FD->parameters().begin(), FD->parameters().end(),
+          [](const ParmVarDecl* PVD) {
+            QualType ParamType = PVD->getType();
+            return ParamType->isPointerType() || ParamType->isReferenceType();
+          });
+      if (m_TopMostReq->Mode == DiffMode::reverse && !FD->isVariadic() &&
+          HasPointerOrReferenceParam && !utils::hasMemoryTypeParams(FD) &&
+          returnType->isIntegralOrEnumerationType())
+        nonDiff = true;
 
       if (nonDiff && m_TopMostReq->Mode != DiffMode::reverse)
         return true;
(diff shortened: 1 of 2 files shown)


guitargeek (approved):
Thanks! Works for me, and I think the idea of not differentiating these integer-returning functions is exactly right.

fatfat123-arch:
Quick clarification: my earlier CUDA-unrelated comment was based on the temporary debug run, where the issue-1793 regression dominated the failures. After restoring the real fix and rerunning on latest base, the failing pair was CUDA/GradientKernels.cu and Regressions/issue-1682.cpp, caused by the same over-broad condition in DiffPlanner. I narrowed that condition in commit 5a24a27d, and the full CI is now green !

PR #1797, opened 2026-03-24
fix #1765 - Resolve valgrind issues for clad on llvm21

### LINKED ISSUE #1765: Resolve valgrind issues for clad on llvm21

Lucky-Lodhi2004:
I have made another PR regarding this issue here - https://github.com/vgvassilev/clad/pull/1778 , but it uses {} for struct object initialization, which may not work universally as discussed here - https://github.com/vgvassilev/clad/pull/1778 , thus the proper way is to set bytes to zero using memset then apply the default constructor like this : 

```C++
FunctionProtoType::ExtProtoInfo EPI;
std::memset(&EPI, 0, sizeof(EPI));
EPI = FunctionProtoType::ExtProtoInfo();
  ```

The valgrind issues are likely caused by garbage initialization and/or padding in memory. I think this will fix this issue. I am now checkig other structs for this fix.

### CODE REVIEW

File: lib/Differentiator/ErrorEstimator.cpp [hunk 1]
```diff
@@ -465,6 +466,8 @@ void ErrorEstimationHandler::LookupCustomErrorFunction() {
     return;
 
   FunctionProtoType::ExtProtoInfo EPI;
+  std::memset(&EPI, 0, sizeof(EPI));
```
vgvassilev: Why do we need this memset? We should do without it.

vgvassilev (commented):
To test what works is to move the ci to run against llvm 22?

lib/Differentiator/ErrorEstimator.cpp
Lucky-Lodhi2004: memset is needed in order to initialize all members to zero because valgrind probably fails because of garbage initialization/padding. This was discussed in this PR https://github.com/vgvassilev/clad/pull/1778

We can do this without memset too, by using static zero initialization. i.e. 

```C++
static const FunctionProtoType::ExtProtoInfo DefaultEPI{};
FunctionProtoType::ExtProtoInfo EPI = DefaultEPI;
```

I am making another commit by applying above change. Whether this works can be tested by running CI with LLVM 21+ (21 or 22).

Other fix without memset is manually zero initializing left out members like `ExtInfo`, `TypeQuals` `,RefQualifier`, `ExceptionSpec`, `ExtParameterInfos` ,`EllipsisLoc`,  `FunctionEffects`, `ExtraAttributeInfo`. This doesn't look good.

So I am going with static zero initialization. This can be tested on LLVM 21 or 22.

guitargeek:
Master resolved #1765 correctly instead: diagnosed it as a **clang/Memcheck false  positive** (llvm/llvm-project#194147) and added `test/clad-valgrind-llvm22.supp`, which suppresses exactly `FunctionProtoType::FunctionProtoType(...ExtProtoInfo&)`.

PR #1800, opened 2026-03-25
Fix forward-mode differentiation for std::thread callables (issue #1795)

### LINKED ISSUE #1795: [Feature] Add forward-mode differentiation support for std::thread
## Summary

Clad fails to generate valid forward-mode derivative code for functions that contain `std::thread` launches.

## Minimal reproducer

```cpp
#include <iostream>
#include <thread>

void add_square(double x, double& out) {
    out = x * x;
}

    double rx = 0.0, ry = 0.0;
    std::thread t1(add_square, x, std::ref(rx));
    std::thread t2(add_square, y, std::ref(ry));
    t1.join();
    t2.join();
    return rx + ry;
}

int main() {
    auto df = clad::differentiate(f, "x");
    std::cout << "f'(3,4) = " << df.execute(3.0, 4.0) << "\n";
    // Expected: 6.0
}
```
## Current behavior

Clad emits invalid generated code when differentiating `f`, leading to compile-time errors for the `std::thread` construction.

## Expected behavior

Forward-mode differentiation should handle `std::thread` launches without crashing and correctly propagate derivatives through the launched callable.

For the example above:

- `f(x, y) = x^2 + y^2`
- `∂f/∂x = 2x`
- so `df.execute(3.0, 4.0)` should return `6.0`

## Possible direction

Extend the forward-mode visitor / call handling path to recognize `std::thread` construction as a callable launch site, then generate the correct pushforward call for the launched function and its arguments.

fatfat123-arch:
This PR : 
- handle `std::thread` constructor differentiation by differentiating the launched callable and using its pushforward in the derivative initialization path
- unwrap `std::ref`/`std::cref` thread arguments before argument differentiation so derivative references map to the underlying objects
- add regression test `test/Regressions/issue-1795.cpp` validating `clad::differentiate(f, "x")` returns `6` for the reproducer from #1795
- local tests run green !

Fixes #1795

### DIFF
--- a/test/Regressions/issue-1795.cpp
+++ b/test/Regressions/issue-1795.cpp
@@ -1,0 +1,67 @@
+// RUN: %cladclang -std=c++20 -I%S/../../include %s -o %t
+// RUN: %t | %filecheck_exec %s
+// UNSUPPORTED: clang-10, clang-11, clang-12, clang-13, clang-14, clang-15, clang-16
+// XFAIL: valgrind
+
+#include <iostream>
+#include <functional>
+#include <thread>
+
+#include "clad/Differentiator/Differentiator.h"
+
+void add_square(double x, double& out) {
+  out = x * x;
+}
+
+std::reference_wrapper<double> wrap(double& out) { return std::ref(out); }
+
+namespace custom {
+std::reference_wrapper<double> ref(double& out) { return std::ref(out); }
+} // namespace custom
+
+double f(double x, double y) {
+  double rx = 0.0;
+  double ry = 0.0;
+  std::thread t1(add_square, x, std::ref(rx));
+  std::thread t2(add_square, y, std::ref(ry));
+  t1.join();
+  t2.join();
+  return rx + ry;
+}
+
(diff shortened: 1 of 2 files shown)


test/Regressions/issue-1795.cpp
```diff
@@ -0,0 +1,32 @@
...
+  double ry = 0.0;
+  std::thread t1(add_square, x, std::ref(rx));
+  std::thread t2(add_square, y, std::ref(ry));
+  t1.join();
+  t2.join();
+  return rx + ry;
+}
+
+int main() {
+  auto df = clad::differentiate(f, "x");
```
vgvassilev, line 49: We also need to make sure that `clad::gradient` works.

fatfat123-arch: Thanks you for the suggestion!  I added a gradient check in the regression too for the next commit: auto gf = clad::gradient(g); gf.execute(3.0, 4.0, &dx, &dy); with CHECK-EXEC: grad g(3,4) = (6, 8). So the test now validates both the std::thread forward-mode path and a working gradient call.

Issue #1799, opened 2026-03-25
Falso postive for `'#pragma clad checkpoint loop' is only allowed before a loop` error


guitargeek:
I think there is a case where there is a false positive of this error, possibly introduced by commit 01d13363ab526066d379ec916c76806f731d90e1.

Reproducer:
```c++
// macro.C

#include <Math/CladDerivator.h>

double gauss_shifted_mean(double *params, double const *obs)
{
   double x[]{obs[0], params[0] - params[1], params[2]};
   const double arg = x[0] - x[1];
   const double sig = x[2];
   return std::exp(-0.5 * arg * arg / (sig * sig));
}

double gaussian_numeric_int(double *params)
{
   double output = 0.0;
   double t6[1];
   {
      const int n = 1000; // number of sampling points
      double d = 4 - -4;
      double eps = d / n;
#pragma clad checkpoint loop
      for (int i = 0; i < n; ++i) {
         t6[0] = -4 + eps * i;
         double tmpA = gauss_shifted_mean(params, t6);
         t6[0] = -4 + eps * (i + 1);
         double tmpB = gauss_shifted_mean(params, t6);
         output += (tmpA + tmpB) * 0.5 * eps;
      }
   }
   return output;
}

#pragma clad ON
void gradient_request()
{
   clad::gradient(gaussian_numeric_int, "params");
}
#pragma clad OFF

std::vector<double> parametersVec = {0, 1, 4};

// To run as a ROOT macro
void macro()
{
   const std::size_t n = parametersVec.size();

   std::vector<double> gradientVec(n);

   auto func = [&](std::span<double> params) { return gaussian_numeric_int(params.data()); };
   auto grad = [&](std::span<double> params, std::span<double> out) {
      return gaussian_numeric_int_grad(parametersVec.data(), out.data());
   };

   grad(parametersVec, gradientVec);

   auto numDiff = [&](int i) {
      const double eps = 1e-6;
      std::vector<double> p{parametersVec};
      p[i] = parametersVec[i] - eps;
      double funcValDown = func(p);
      p[i] = parametersVec[i] + eps;
      double funcValUp = func(p);
      return (funcValUp - funcValDown) / (2 * eps);
   };

   for (std::size_t i = 0; i < parametersVec.size(); ++i) {
      std::cout << i << ":" << std::endl;
      std::cout << "  numr : " << numDiff(i) << std::endl;
      std::cout << "  clad : " << gradientVec[i] << std::endl;
   }
}
```
[...]

PR #1801, opened 2026-03-25
Fix false positive checkpoint-loop pragma in nested scopes (issue #1799)

fatfat123-arch:
This PR fixes the false positive in #1799:

`'#pragma clad checkpoint loop' is only allowed before a loop`

The issue happens when the pragma is inside a nested `{ ... }` block.  
In that case, pragma attachment could miss the nested `for`/`while`/`do` and wrongly emit the diagnostic.

Changes in this PR:
- fix pragma-to-loop attachment lookup for nested scopes
- keep the same loop-only rule (`for` / `while` / `do`)
- add a regression test: `test/Regressions/issue-1799.cpp`

All relevant tests run green locally:
- `lit -sv build/test/Regressions/issue-1799.cpp`
- `lit -sv build/test/Regressions/issue-1736.cpp`
- `ninja -C build -j4 check-clad-regressions`
- `ninja -C build -j4 check-clad-diagnostics`

### DIFF
--- a/test/Regressions/issue-1799.cpp
+++ b/test/Regressions/issue-1799.cpp
@@ -1,0 +1,61 @@
+// RUN: %cladclang -std=c++20 -I%S/../../include %s -o %t
+// RUN: %t | %filecheck_exec %s
+#include <cmath>
+#include <iostream>
+
+#include "clad/Differentiator/Differentiator.h"
+
+double gauss_shifted_mean(double* params, const double* obs) {
+  double x[] = {obs[0], params[0] - params[1], params[2]};
+  const double arg = x[0] - x[1];
+  const double sig = x[2];
+  return std::exp(-0.5 * arg * arg / (sig * sig));
+}
+
+double gaussian_numeric_int(double* params) {
+  double output = 0.0;
+  double t6[1];
+  {
+    const int n = 100;
+    const double d = 4 - -4;
+    const double eps = d / n;
+    #pragma clad checkpoint loop
+    for (int i = 0; i < n; ++i) {
+      t6[0] = -4 + eps * i;
+      const double tmpA = gauss_shifted_mean(params, t6);
+      t6[0] = -4 + eps * (i + 1);
+      const double tmpB = gauss_shifted_mean(params, t6);
+      output += (tmpA + tmpB) * 0.5 * eps;
+    }
+  }
+  return output;
+}
+
+double gauss_point(double* params, double x) {
+  double obs[1] = {x};
+  return gauss_shifted_mean(params, obs);
+}
+
+double gaussian_numeric_int_no_braces(double* params) {
+  double output = 0.0;
+  const int n = 16;
+  const double eps = (4 - -4) / static_cast<double>(n);
+  #pragma clad checkpoint loop
+  for (int i = 0; i < n; ++i)
+    output += gauss_point(params, -4 + eps * i);
+  return output;
+}
+
+#pragma clad ON
+void gradient_request() {
+  clad::gradient(gaussian_numeric_int, "params");
+  clad::gradient(gaussian_numeric_int_no_braces, "params");
+}
+#pragma clad OFF
+
+int main() {
+  gradient_request();
+  std::cout << "ok\n";
+  // CHECK-EXEC: ok
+  return 0;
+}
--- a/tools/ClangPlugin.cpp
+++ b/tools/ClangPlugin.cpp
@@ -239,24 +239,45 @@ void InitTimers();
       }
     }
 
+    class AttachedLoopStmtFinder
+        : public RecursiveASTVisitor<AttachedLoopStmtFinder> {
+      SourceLocation m_PragmaLoc;
+      SourceManager& m_SM;
+      Stmt* m_AttachedStmt = nullptr;
+      SourceLocation m_AttachedLoopLoc;
+
+    public:
+      AttachedLoopStmtFinder(SourceLocation pragmaLoc, SourceManager& SM)
+          : m_PragmaLoc(pragmaLoc), m_SM(SM) {}
+
+      bool VisitStmt(Stmt* S) {
+        SourceLocation beginLoc = S->getBeginLoc();
+        if (!beginLoc.isValid() ||
+            !m_SM.isBeforeInTranslationUnit(m_PragmaLoc, beginLoc))
+          return true;
+
+        if (!m_AttachedStmt || m_SM.isBeforeInTranslationUnit(
+                                   beginLoc, m_AttachedStmt->getBeginLoc())) {
+          m_AttachedStmt = S;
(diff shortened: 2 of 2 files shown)


tools/ClangPlugin.cpp
```diff
@@ -239,22 +239,69 @@ void InitTimers();
...
+          S = AS->getSubStmt();
+          continue;
+        }
+        if (const auto* LS = dyn_cast<LabelStmt>(S)) {
+          S = LS->getSubStmt();
+          continue;
+        }
         break;
       }
+      return false;
```
vgvassilev: I'd implement something like inside the function:
```cpp
struct LoopFinder : public RecursiveASTVisitor<LoopFinder> {
  HasLoop = false;
  bool VisitStmt(const Stmt* S) {
    if (isa<loop>(S)) {
       HasLoop = true;
       return false; // abort recursive visitation
    }
  return true;
  }
} LoopFinder;
LoopFinder.TraverseStmt(S);
return LoopFinder.HasLoop;
```

tools/ClangPlugin.cpp
```diff
@@ -239,22 +239,69 @@ void InitTimers();
...
+                                              SourceLocation pragmaLoc,
+                                              SourceManager& SM) {
+      if (!root)
+        return nullptr;
+
+      const Stmt* nextStmt = nullptr;
+      if (const auto* body = dyn_cast<CompoundStmt>(root))
+        nextStmt = getFirstStmtAfterPragma(body, pragmaLoc, SM);
+
+      for (const Stmt* child : root->children()) {
```
vgvassilev: You can use a RecursiveASTVisitor here too.

test/Regressions/issue-1799.cpp
```diff
@@ -0,0 +1,45 @@
+// RUN: %cladclang -std=c++20 -I%S/../../include %s -o %t
+// RUN: %t | %filecheck_exec %s
+// UNSUPPORTED: clang-10, clang-11, clang-12, clang-13, clang-14, clang-15, clang-16
```
vgvassilev: Why these are not supported?

fatfat123-arch: First version was a minimal quick fix to unblock the false positive safely. In the next commit I switched to RecursiveASTVisitor (as suggested) and removed the UNSUPPORTED guard in the regression test !

vgvassilev: Let's go to the second version -- I do not think we need to split this into two.

fatfat123-arch: I moved to the second version only and kept it as a single commit. Thank you for the feedback !

tools/ClangPlugin.cpp
```diff
@@ -239,22 +239,55 @@ void InitTimers();
...
+      FirstStmtAfterPragmaFinder(SourceLocation pragmaLoc, SourceManager& SM)
+          : m_PragmaLoc(pragmaLoc), m_SM(SM) {}
+
+      bool VisitCompoundStmt(CompoundStmt* body) {
+        for (Stmt* S : body->body()) {
+          SourceLocation beginLoc = S->getBeginLoc();
+          if (!m_SM.isBeforeInTranslationUnit(m_PragmaLoc, beginLoc))
+            continue;
+          if (!m_NextStmt || m_SM.isBeforeInTranslationUnit(
+                                 beginLoc, m_NextStmt->getBeginLoc()))
```
vgvassilev: In principle there are `WalkUp` interfaces that wold give you the previous stmt maybe...

tools/ClangPlugin.cpp
```diff
@@ -239,22 +239,55 @@ void InitTimers();
...
+        bool VisitStmt(Stmt* Cur) {
+          if (isa<ForStmt>(Cur) || isa<WhileStmt>(Cur) || isa<DoStmt>(Cur)) {
+            HasLoop = true;
+            return false;
+          }
+          return true;
+        }
+      } finder;
+      finder.TraverseStmt(const_cast<Stmt*>(S));
+      return finder.HasLoop;
```
vgvassilev:
```suggestion
    static bool isLoopStmt(Stmt* S) {
      struct LoopFinder : public RecursiveASTVisitor<LoopFinder> {
        bool HasLoop = false;
        bool VisitStmt(Stmt* Cur) {
          if (isa<ForStmt>(Cur) || isa<WhileStmt>(Cur) || isa<DoStmt>(Cur)) {
            HasLoop = true;
            return false;
          }
          return true;
        }
      } finder;
      finder.TraverseStmt(S);
      return finder.HasLoop;
```

tools/ClangPlugin.cpp
```diff
@@ -239,22 +239,55 @@ void InitTimers();
...
+        : public RecursiveASTVisitor<FirstStmtAfterPragmaFinder> {
+      SourceLocation m_PragmaLoc;
+      SourceManager& m_SM;
+      const Stmt* m_NextStmt = nullptr;
+
+    public:
+      FirstStmtAfterPragmaFinder(SourceLocation pragmaLoc, SourceManager& SM)
+          : m_PragmaLoc(pragmaLoc), m_SM(SM) {}
+
+      bool VisitCompoundStmt(CompoundStmt* body) {
```
vgvassilev: I do not understand that logic. Consider:
```cpp
int sum = 0;
for (unsigned i = 0; i < 10; ++i) sum++;
```

That won't produce a compound statement. This entire check here feels incorrect -- can you explain in a comment what we are trying to do here?

fatfat123-arch: The code was trying to find the first statement after the pragma and validate that it is loop-related, but it incorrectly relied on CompoundStmt structure. Thank you for the clear example, @vgvassilev! I now realize the current CompoundStmt-based check is too restrictive.

tools/ClangPlugin.cpp
```diff
@@ -239,22 +239,59 @@ void InitTimers();
...
+                                             SourceLocation pragmaLoc,
+                                             SourceManager& SM) {
+      Stmt* body = FD ? FD->getBody() : nullptr;
+      if (!body)
+        return {};
+
+      FirstStmtAfterPragmaFinder finder(pragmaLoc, SM);
+      finder.TraverseStmt(body);
+      Stmt* nextStmt = finder.getNextStmt();
+      if (isLoopStmt(nextStmt))
```
vgvassilev: Why do we need two visitors to figure this out -- we need just one, right?

fatfat123-arch: I split it into two visitors: one finds the statement immediately attached to the pragma, and the second checks whether that attached statement is loop-related. I kept them separate to avoid mixing responsibilities and to prevent accidentally matching a later unrelated loop in the function body. I can also try a single-visitor version while keeping the same strict attachment behavior !

PR #1802, opened 2026-03-25
Add forward-mode std::thread differentiation: parallel derivative threads and join() handling (#1795)

### LINKED ISSUE #1795: [Feature] Add forward-mode differentiation support for std::thread

JoeFilfli:
## Summary

Fixes forward-mode differentiation of `std::thread` (issue #1795).

**`VisitCXXConstructExpr`** — detects `std::thread` constructor, requests the callable's pushforward via `HandleNestedDiffRequest`, and builds a real parallel derivative thread:
```cpp
std::thread _d_t1{add_square_pushforward, x, ref(rx), _d_x, ref(_d_rx)};
```

**`VisitCallExpr`** — detects `std::thread::join()` and emits `_d_t1.join()` alongside the primal join, preventing `std::terminate()` at destruction.

Fixes #1795
```

### DIFF
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -65,8 +65,7 @@ bool IsRealNonReferenceType(QualType T) {
 // Returns true if RD is in std namespace, handles libc++ inline
 // namespaces like std::__1 correctly.
 static bool isInStdNamespace(const CXXRecordDecl* RD) {
-  for (const DeclContext* DC = RD->getDeclContext(); DC;
-       DC = DC->getParent()) {
+  for (const DeclContext* DC = RD->getDeclContext(); DC; DC = DC->getParent()) {
     if (const auto* NS = dyn_cast<NamespaceDecl>(DC))
       if (NS->isStdNamespace())
         return true;
@@ -1037,8 +1036,7 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
   if (const auto* MCE = dyn_cast<CXXMemberCallExpr>(CE)) {
     if (const auto* MD = dyn_cast<CXXMethodDecl>(FD)) {
       const CXXRecordDecl* RD = MD->getParent();
-      if (isInStdNamespace(RD) &&
-          RD->getName() == "thread" &&
+      if (isInStdNamespace(RD) && RD->getName() == "thread" &&
           FD->getName() == "join") {
 
         // Get base: t1 → {t1, _d_t1}
@@ -1059,7 +1057,7 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
       }
     }
   }
-  
+
   SourceLocation validLoc{CE->getBeginLoc()};
 
   // Calls to lambda functions are processed differently
@@ -2174,8 +2172,7 @@ BaseForwardModeVisitor::VisitCXXConstructExpr(const CXXConstructExpr* CE) {
 
   CXXConstructorDecl* CD = CE->getConstructor();
   CXXRecordDecl* RD = CD->getParent();
-  bool isStdThread = isInStdNamespace(RD) &&
-                     RD->getName() == "thread" &&
+  bool isStdThread = isInStdNamespace(RD) && RD->getName() == "thread" &&
                      CE->getNumArgs() >= 1;
 
   if (isStdThread) {
@@ -2196,7 +2193,8 @@ BaseForwardModeVisitor::VisitCXXConstructExpr(const CXXConstructExpr* CE) {
           utils::ComputeEffectiveFnName(callableFD);
       pushforwardFnRequest.VerboseDiags = false;
       pushforwardFnRequest.EnableTBRAnalysis = m_DiffReq.EnableTBRAnalysis;
-      pushforwardFnRequest.EnableVariedAnalysis = m_DiffReq.EnableVariedAnalysis;
+      pushforwardFnRequest.EnableVariedAnalysis =
+          m_DiffReq.EnableVariedAnalysis;
 
       FunctionDecl* pushforwardFD =
           m_Builder.HandleNestedDiffRequest(pushforwardFnRequest);
@@ -2205,7 +2203,8 @@ BaseForwardModeVisitor::VisitCXXConstructExpr(const CXXConstructExpr* CE) {
         // Original thread:    std::thread t1(add_square,    x, _t0.value)
         // Derivative thread:  std::thread _d_t1(add_square_pushforward,
         //                                       x, _t0.value,      ← orig args
-        //                                       _d_x, _t0.pushforward) ← deriv args
+        //                                       _d_x, _t0.pushforward) ← deriv
+        //                                       args
 
         llvm::SmallVector<Expr*, 8> derivThreadArgs;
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/Regressions/issue-1795.cpp [hunk 1]
```diff
@@ -0,0 +1,26 @@
...
+#include <iostream>
+#include <thread>
+#include "clad/Differentiator/Differentiator.h"
+
+void add_square(double x, double& out) {
+    out = x * x;
+}
+
+double f(double x, double y) {
+    double rx = 0.0, ry = 0.0;
```
parth-07, line 14: This is the differentiated code for this function:

```cpp
double f_darg0(double x, double y) {
    double _d_x = 1;
    double _d_y = 0;
    double _d_rx = 0.;
    double rx = 0.;
    double _d_ry = 0.;
    double ry = 0.;
    clad::ValueAndPushforward<reference_wrapper<double>, reference_wrapper<double> > _t0 = std::ref_pushforward(rx, _d_rx);
    std::thread _d_t1(add_square_pushforward(x, rx, _d_x, _d_rx) , std::thread());
    std::thread t1{add_square, x, _t0.value};
    clad::ValueAndPushforward<reference_wrapper<double>, reference_wrapper<double> > _t1 = std::ref_pushforward(ry, _d_ry);
    std::thread _d_t2(add_square_pushforward(y, ry, _d_y, _d_ry) , std::thread());
    std::thread t2{add_square, y, _t1.value};
    t1.join();
    t2.join();
    return _d_rx + _d_ry;
}
```

Do we need to call both the `add_square_pushforward` and `add_square`?

parth-07 (changes requested):

test/Regressions/issue-1795.cpp
JoeFilfli: You are correct .`add_square_pushforward` already  contains the original computation `out = x * x`, so launching `t1` separately runs it twice. The fix is to return only the derivative thread from `VisitCXXConstructExpr`. I will push a fix shortly.

parth-07: Now, the differentiated code for `std::thread` object is:

```cpp
std::thread _d_t1{add_square_pushforward, x, _t0.value, _d_x, _t0.pushforward};
std::thread t1;
```

What do you think about not generating `t1` at all?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -2131,6 +2170,65 @@ BaseForwardModeVisitor::VisitCXXConstructExpr(const CXXConstructExpr* CE) {
...
+  if (isStdThread) {
+    // In std::thread t1(add_square, x, ref(result))
+    // CE->getArg(0) is add_square — a DeclRefExpr pointing to FunctionDecl
+    const Expr* callableArg = CE->getArg(0)->IgnoreParenImpCasts();
+    const FunctionDecl* callableFD = nullptr;
+
+    if (const auto* DRE = dyn_cast<DeclRefExpr>(callableArg))
+      callableFD = dyn_cast<FunctionDecl>(DRE->getDecl());
+
+    if (callableFD) {
```
vgvassilev: Only a callable written as a plain function name reaches the pushforward path; everything else leaves silently. `IgnoreParenImpCasts` does not strip a `CXXConstructExpr`, a `LambdaExpr`, or the `&` in `&Obj::run`, so a `std::function`, a functor object, a lambda and a function-pointer variable all leave `callableFD` null. Control then reaches the generic constructor fallback further down -- the one whose own FIXME says it is incorrect -- which builds the derivative thread out of the derivative arguments and whatever it made of the callable. Nothing points a diagnostic at the user's line. A wrong derivative that compiles is worse here than a refusal, and #1891 reports an error for exactly these cases, which is the right answer until they work.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -2131,6 +2170,65 @@ BaseForwardModeVisitor::VisitCXXConstructExpr(const CXXConstructExpr* CE) {
...
+      // This is exactly what VisitCallExpr does for regular function calls
+      DiffRequest pushforwardFnRequest;
+      pushforwardFnRequest.Function = callableFD;
+      pushforwardFnRequest.Mode = DiffMode::pushforward;
```
vgvassilev: Naming the mode here breaks vector forward mode. `GetPushForwardMode()` is virtual and `VectorForwardModeVisitor` overrides it to return `vector_pushforward`; `VisitCallExpr` builds the same request through that call rather than naming `DiffMode::pushforward`. `VisitCXXConstructExpr` is not overridden, so `clad::jacobian`, or `clad::differentiate` with `clad::opts::vector_mode`, on a function holding a `std::thread` will request the scalar pushforward and then call it with vector-mode derivative arguments. The comment two lines up says this is exactly what `VisitCallExpr` does, so it reads as a slip rather than a decision.

```suggestion
      pushforwardFnRequest.Mode = GetPushForwardMode();
```

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1019,6 +1032,32 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
...
+        Expr* derivBase = baseDiff.getExpr_dx();
+        Expr* derivJoin =
+            m_Sema
+                .ActOnCallExpr(getCurrentScope(),
+                               utils::BuildMemberExpr(m_Sema, getCurrentScope(),
+                                                      derivBase, "join"),
+                               noLoc, {}, noLoc)
+                .get();
+
+        return StmtDiff(nullptr, derivJoin);
```
vgvassilev: Returning a null primal deletes the user's `t1.join()` from the generated code, which is safe only while the constructor above has neutered the primal thread. When the callable is not a plain function name the constructor falls through and `t1` is a real running thread -- and this still deletes its `join()`, so `~thread` runs on a joinable thread, which is `std::terminate()` by [thread.thread.destr]. The two decisions have to be tied together: drop the primal join only for a thread this pass actually rewrote, or refuse the constructor when the callable is not recognised so no such thread exists. `joinable()` and `detach()` are not handled at all and will take the ordinary member-call path.

vgvassilev:
This work is superseded by #1891 which adds support for std::thread in both modes.

PR #1806, opened 2026-03-30
Add nullptr check in array subscripts

### LINKED ISSUE #1804: Regression since Clad 2.2: `clad::hessian` fails for nested functions with `double const *` paramters
This follows up on https://github.com/vgvassilev/clad/issues/1737 with the next error I'm hitting now.

I think the problem is similar to https://github.com/vgvassilev/clad/issues/1793, which was already solved, so this issue should not be too hard.

Also, I *think* this is a regression, because I vaguely remember seeing it working before. I'm still bisecting.
*Edit:* the error first appears with a46878ff, but it doesn't mean that commit was responsible for the problem. Before that commit, the Hessian seemed to work, but gave wrong results. So I think a46878ff unveiled this problem, but is not responsible for it.

Reproducer as a ROOT macro:

```c++

#define DO_HESSIAN

double inner_func(double *params, double const *obs)
{
   double arg = (obs[0] - params[1]) / params[2];
   return arg * arg;
}

double outer_func(double *params, double const *obs)
{
   return inner_func(params, obs) + params[0];
}

#pragma clad ON
void gradient_request()
{
   clad::gradient(outer_func, "params");
#ifdef DO_HESSIAN
   clad::hessian(outer_func, "params[0:2]");
#endif
}
#pragma clad OFF
// clang-format off

    0.5, -1, 2
};

std::vector<double> observablesVec = {-9.9};

// clang-format on

// To run as a ROOT macro
void cladrepro()
{

   auto func = [&](std::span<double> params) { return outer_func(params.data(), observablesVec.data()); };
      return outer_func_grad_0(parametersVec.data(), observablesVec.data(), out.data());
   };

   auto numDiff = [&](int i) {
      const double eps = 1e-6;
      double funcValDown = func(p);
      double funcValUp = func(p);
   };

   }

#ifdef DO_HESSIAN
   std::cout << "\n";

      return outer_func_hessian_0(params.data(), observablesVec.data(), out.data());
   };

   // Uses central differences:

      if (i == j) {
         const double f0 = func(p);

         const double fUp = func(p);

         const double fDown = func(p);

      } else {
         // f(x_i + eps, x_j + eps)
         const double fPP = func(p);

         // f(x_i + eps, x_j - eps)
         const double fPM = func(p);

         // f(x_i - eps, x_j + eps)
         const double fMP = func(p);

         // f(x_i - eps, x_j - eps)
         const double fMM = func(p);

      }
   };

      }
   }

      }
   }

      }
      std::cout << "\n";
   }

      }
      std::cout << "\n";
   }
#endif
}
```

The output is:
```txt
Processing cladrepro.C...
error: cannot initialize a parameter of type 'double *' with an rvalue of type 'const double *'
<<< cling interactive line includer >>>:1:1: note: passing argument to parameter '_d_d_obs' here

^
error: cannot initialize a parameter of type 'double *' with an rvalue of type 'const double *'
<<< cling interactive line includer >>>:1:1: note: passing argument to parameter '_d_d_obs' here
error: cannot initialize a parameter of type 'double *' with an rvalue of type 'const double *'
<<< cling interactive line includer >>>:1:1: note: passing argument to parameter '_d_d_obs' here
error: cannot initialize a parameter of type 'double *' with an rvalue of type 'const double *'
<<< cling interactive line includer >>>:1:1: note: passing argument to parameter '_d_d_obs' here
error: cannot initialize a parameter of type 'double *' with an rvalue of type 'const double *'
<<< cling interactive line includer >>>:1:1: note: passing argument to parameter '_d_d_obs' here
```

Fermats-Last-Theorem:
This PR fixes the segfault described in the follow up of #1804. The issue was that the pushforward and pullback functions were trying to read/write the adjoints for inactive parameters with nullptr values, causing the segfault.
`nullptr` checks have been added in `VisitArraySubuscriptExpr` to resolve it.
All the affected tests have been modified.

Fixes #1804

### DIFF
--- a/test/Arrays/ArrayInputsReverseMode.C
+++ b/test/Arrays/ArrayInputsReverseMode.C
@@ -29,7 +29,8 @@ double addArr(const double *arr, int n) {
 //CHECK-NEXT:         i--;
 //CHECK-NEXT:         {
 //CHECK-NEXT:             double _r_d0 = _d_ret;
-//CHECK-NEXT:             _d_arr[i] += _r_d0;
+//CHECK-NEXT:             if (_d_arr)
+//CHECK-NEXT:               _d_arr[i] += _r_d0;
 //CHECK-NEXT:         }
 //CHECK-NEXT:     }
 //CHECK-NEXT: }
@@ -72,14 +73,16 @@ float func(float* a, float* b) {
 //CHECK-NEXT:         i--;
 //CHECK-NEXT:         {
 //CHECK-NEXT:             float _r_d1 = _d_sum;
-//CHECK-NEXT:             _d_a[i] += _r_d1;
+//CHECK-NEXT:             if (_d_a)
+//CHECK-NEXT:               _d_a[i] += _r_d1;
 //CHECK-NEXT:         }
 //CHECK-NEXT:         {
 //CHECK-NEXT:             a[i] = clad::pop(_t1);
 //CHECK-NEXT:             float _r_d0 = _d_a[i];
 //CHECK-NEXT:             _d_a[i] = 0.F;
 //CHECK-NEXT:             _d_a[i] += _r_d0 * b[i];
-//CHECK-NEXT:             _d_b[i] += a[i] * _r_d0;
+//CHECK-NEXT:             if (_d_b)
+//CHECK-NEXT:               _d_b[i] += a[i] * _r_d0;
 //CHECK-NEXT:         }
 //CHECK-NEXT:     }
 //CHECK-NEXT: }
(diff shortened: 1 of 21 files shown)


test/Arrays/ArrayInputsReverseMode.C
```diff
@@ -294,8 +300,10 @@ double inv_square(double *params) {
 //CHECK-NEXT:     double _t0 = (params[0] * params[0]);
 //CHECK-NEXT:     {
 //CHECK-NEXT:         double _r0 = _d_y * -(1 / (_t0 * _t0));
-//CHECK-NEXT:         _d_params[0] += _r0 * params[0];
-//CHECK-NEXT:         _d_params[0] += params[0] * _r0;
+//CHECK-NEXT:         if (_d_params)
+//CHECK-NEXT:           _d_params[0] += _r0 * params[0];
+//CHECK-NEXT:         if (_d_params)
+//CHECK-NEXT:           _d_params[0] += params[0] * _r0;
```
vgvassilev, line 306: We don't need two if-stmts here.

vgvassilev (commented):
I am missing the context here a bit. Why do we end up trying to increments an array that's a `nullptr`?

PR #1812, opened 2026-04-08
Implements GPU-Offload feature to the tape

Vedant2005goyal:
This PR completes the N-tier memory architecture (VRAM -> RAM -> Disk). In systems compatible with CUDA will first fill the VRAM and when it gets full the older slabs will be offloaded to the RAM and same pattern is follwed in RAM also. Benchmarking is also updated to benchmark these newly added features

Below are the benchmarks for the task which dont requires gpu
image

Below are the benchmark for the task which requires gpu and if in case the device dont have it fallbacks to use the RAM.
image


vgvassilev:
Did you take a look at the implementation of universal_vector?

vgvassilev (approved):
LGTM!

PR #1813, opened 2026-04-11
Implement analytical derivatives for Cylindrical Bessel functions (j, i, k)

leetcodez:
This PR adds exact analytical derivatives for the C++17 cylindrical Bessel function `std::cyl_bessel_j`, `std::cyl_bessel_i`, and `std::cyl_bessel_k` covering both Pushforward and Pullback modes.

To avoid repeatedly evaluating expensive special functions, the implementation uses shifted recurrence relations that reuse already computed values.
Rather than the standard derivative form:
`0.5 * (J_{n-1}(x) - J_{n+1}(x))`

I used optimized equivalents:
```
Jₙ′(x) = Jₙ₋₁(x) − (n/x)·Jₙ(x)
Iₙ′(x) = Iₙ₋₁(x) − (n/x)·Iₙ(x)
Kₙ′(x) = −Kₙ₋₁(x) − (n/x)·Kₙ(x)
```

**Edge cases:**
- cyl_bessel_k is intentionally left without a zero check, since it is mathematically singular at the origin (→ ∞). Avoiding extra branching keeps the fast path intact.
- In Pullback mode, primal values are recomputed inside the handlers so the recurrence-based derivatives work correctly during adjoint propagation.

**Testing**
- Added FileCheck tests in BuiltinDerivatives.C to ensure correct AST generation and interception.
- Added CHECK_ALL_RANGE tests in stl-cmath.cpp, comparing analytical derivatives against central finite-difference approximations over a wide domain.

### DIFF
--- a/test/Features/stl-cmath.cpp
+++ b/test/Features/stl-cmath.cpp
@@ -102,9 +102,9 @@
 // DS comp_ellint_1/ f / l      (C++17) complete elliptic integral (1st kind)
 // DS comp_ellint_2/ f / l      (C++17) complete elliptic integral (2nd kind)
 // DS comp_ellint_3/ f / l      (C++17) complete elliptic integral (3rd kind)
-// D cyl_bessel_i/ f / l       (C++17) modified cylindrical Bessel (regular)
-// D cyl_bessel_j/ f / l       (C++17) cylindrical Bessel functions (1st kind)
-// D cyl_bessel_k/ f / l       (C++17) modified cylindrical Bessel (irregular)
+// DS cyl_bessel_i/ f / l       (C++17) modified cylindrical Bessel (regular)
+// DS cyl_bessel_j/ f / l       (C++17) cylindrical Bessel functions (1st kind)
+// DS cyl_bessel_k/ f / l       (C++17) modified cylindrical Bessel (irregular)
 // D cyl_neumann/ f / l        (C++17) cylindrical Neumann functions
 // D ellint_1/ f / l           (C++17) incomplete elliptic integral (1st kind)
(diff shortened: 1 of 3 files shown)


test/Features/stl-cmath.cpp
```diff
@@ -125,10 +125,23 @@
...
+  inline float cyl_bessel_j(float nu, float x) { return 0.0f; }
+  inline long double cyl_bessel_j(long double nu, long double x) { return 0.0L; }
+
+  inline double cyl_bessel_i(double nu, double x) { return 0.0; }
+  inline float cyl_bessel_i(float nu, float x) { return 0.0f; }
+  inline long double cyl_bessel_i(long double nu, long double x) { return 0.0L; }
+
+  inline double cyl_bessel_k(double nu, double x) { return 0.0; }
+  inline float cyl_bessel_k(float nu, float x) { return 0.0f; }
+  inline long double cyl_bessel_k(long double nu, long double x) { return 0.0L; }
```
vgvassilev, line 144: Why do we need this?

leetcodez: This is a workaround for the macOS CI runners. Apple's libc++ doesn't implement the C++17 Bessel functions yet. If we don't mock them, the compiler throws an error before clad can build the AST

Unlike std::beta or digamma where we could implement a true mathematical fallback, writing a full numerical approximation for Bessel functions from scratch just to satisfy the Apple CI was out of scope, so I used the return 0.0;

leetcodez:
Hi @vgvassilev,
any updates on the PR review?

Issue #1815, opened 2026-04-30
Assertion `detail::isPresent(Val) && "dyn_cast on a non-existent value"' failed.


MaxSagebaum:
I had time again to look at the TRACE differentiation. The differentiated sources need to be extended which lead to a new problem. The reduced problem is:

```
#include "clad/Differentiator/Differentiator.h"
typedef enum { a } b;
typedef enum { c, d } e;
e f;
int *g() {
  switch (f) { case c: }
}
void h(b, e, char[], char[], int, bool, char, char *, va_list) { g(); }
char i, o, j;
int k;
void l(float) {
  va_list arg;
  h(a, d, &i, &o, k, 0, NULL, &j, arg);
}
float m;
void n() {
  l(m);
  clad::gradient(n);
}

```

The output I get:
```
test.cpp:6:24: warning: label at end of compound statement is a C++23 extension [-Wc++23-extensions]
    6 |   switch (f) { case c: }
      |                        ^
test.cpp:6:11: warning: enumeration value 'd' not handled in switch [-Wswitch]
    6 |   switch (f) { case c: }
      |           ^
test.cpp:7:1: warning: non-void function does not return a value [-Wreturn-type]
    7 | }
      | ^
test.cpp:13:25: warning: implicit conversion of NULL constant to 'char' [-Wnull-conversion]
   13 |   h(a, d, &i, &o, k, 0, NULL, &j, arg);
      |   ~                     ^~~~
      |                         '\0'
test.cpp:18:3: error: attempted to differentiate function with no parameters
   18 |   clad::gradient(n);
      |   ^~~~
clang++: /usr/include/llvm/Support/Casting.h:662: decltype(auto) llvm::dyn_cast(From *) [To = clang::CompoundStmt, From = clang::Stmt]: Assertion `detail::isPresent(Val) && "dyn_cast on a non-existent value"' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
[...]

PR #1816, opened 2026-05-01
Fix: For issue-1815

Vedant2005goyal:
This PR closes issue #1815. Earlier when differentiating a function containing an empty switch case null_ptr is passed to llvm::dyn_cast<CompoundStmt>, triggering the assertion failure. Fix include to safely handle empty statements. One more bug was that when the user tries to differentiate function with no parameters which leads to segmentation fault. Fix for that includes early return.


vgvassilev:
Can we add a test?

Vedant2005goyal:
> Can we add a test?

Done

vgvassilev:
Can we fix the failing bots?

vgvassilev:
Fixing clang-tidy would also make sense.

PR #1819, opened 2026-05-04
Update Clad ClangPlugin to fix issues in root PR #21903

SahilPatidar:


SahilPatidar:
Okay, I think the crash happens because the consumer is placed before the others, causing two issues: first, they don’t get initialized through the Multiplex consumer, and second, it invalidates the Multiplex iterator.

vgvassilev:
Hi @SahilPatidar, any news one this one?

SahilPatidar:
No, I started looking into Clang internals for our project. This is quite hard to manage because in cling we want the consumer to come before the codegen consumer, so we want to use `AddBeforeMainAction`. But to keep the `MultiplexConsumer` iterator valid while stealing the consumer, it currently needs to stay last in normal mode.

In incremental mode we skip stealing, so this problem does not happen there, but in normal mode it invalidates the MultiplexConsumer iterator while initializing the consumer.

I’m not sure if we can return AddBeforeMainAction conditionally from there. Otherwise, I can try allowing stealing while adding `DoNotAppendDelayed` when we are in incremental compilation mode.

vgvassilev:
> No, I started looking into Clang internals for our project. This is quite hard to manage because in cling we want the consumer to come before the codegen consumer, so we want to use `AddBeforeMainAction`. But to keep the `MultiplexConsumer` iterator valid while stealing the consumer, it currently needs to stay last in normal mode.
> 
> In incremental mode we skip stealing, so this problem does not happen there, but in normal mode it invalidates the MultiplexConsumer iterator while initializing the consumer.
> 
> I’m not sure if we can return AddBeforeMainAction conditionally from there. Otherwise, I can try allowing stealing while adding `DoNotAppendDelayed` when we are in incremental compilation mode.

I think that's the final push and let's try to make this work.

SahilPatidar:
Okay, I need to try allowing stealing but not allow delayed, if that works with Cling. Is it designed for incremental compilation as well?

vgvassilev:
> Okay, I need to try allowing stealing but not allow delayed, if that works with Cling. Is it designed for incremental compilation as well?

I think so.

SahilPatidar:
This time I just swapped the last and front consumers. The good thing is that Cling never calls the initialize function, so this should not affect the current Cling behavior.

SahilPatidar:
Okay, clang-repl is not working. The test I ran locally is failing because these changes conflict with Clad’s assumptions.
```cpp
sahilpatidar@Sahils-MacBook-Air clang-build % /opt/homebrew/Cellar/llvm/19.1.6/bin/clang-repl -Xcc -fplugin=/Users/sahilpatidar/Desktop/origin-clad/clad/build/lib/clad.dylib -Xcc -I/Users/sahilpatidar/Desktop/origin-clad/clad/test/Misc/../../include
clang-repl> double sq(double x) { return x*x; }
clang-repl> extern "C" int printf(const char*,...);
clang-repl> #include "clad/Differentiator/Differentiator.h"
clang-repl> auto dsq = clad::differentiate(sq, "x");
clang-repl> auto r1 = printf("dsq(1)=%f\n", dsq.execute(1));
dsq(1)=0.000000
clang-repl> %quit
```
But cling works with the same changes. As I mentioned before, cling never calls `Initialize` or `HandleTranslationUnit`, yet it still works just by adding the consumer before the other declarations.

root.exe output:
```cpp
dsq(1)=2.000000
```
So I think cling must be doing something internally that makes this work. Do you know what could be affecting this behavior?

I wanted to properly test the Clad plugin with the “stealing” approach, but the issue is that we cannot call HandleTranslationUnit because we add the `PCHGenerator` consumer, which should only run at the end of the session when creating the PCM. That is why we only call `getCodeGenerator()->HandleTranslationUnit()`.

So the only option seems to be separating `PCHGenerator` from the interactive session, and only keeping `ASTWriter` as the listener, if that is possible? Then we can probably use the Clad plugin without modifying it.

SahilPatidar:
Or we can create a wrapper consumer around `PCHGenerator` and `PCHContainerGenerator` that skips the `HandleTranslationUnit` calls during incremental inputs and calls it only once at finalization time. Since Cling expect to call `HandleTranslationUnit` for every input, this may help preserve the assumptions expected by the `PCH generator`. Do you think this approach is feasible?

vgvassilev:
Sounds quite feasible to me. If nothing works we can always check if the preprocessor has the `__CLING__` macro and take a different route in clad but probably worth investigating a more robust solution.

SahilPatidar:
So I tried two things: adding the PCH-related consumer to the wrapper consumer, and adding a `GeneratePCH` function in Interpreter. This function calls `IncrementalAction::GenPCH`, where I`ncrementalAction` has a pointer to the wrapper consumer. The build completed successfully.

The second change was allowing CladPlugin to steal consumers. After that, the build was also successful, but a few tests (around 3–4 from roottest-cling) failed. Two of them failed with an error related to copyArray:
```
IncrementalExecutor::executeFunction: symbol '_ZN5cling7runtime8internal9copyArrayIKcEEvPT_Pvm' unresolved while linking [cling interface function]!
1494: You are probably missing the definition of void cling::runtime::internal::copyArray<char const>(char const*, void*, unsigned long)
```
Another issue is that whenever we call Handle*-related functions in the consumer, we also need to call HandleTranslationUnit to execute delayed calls.

I also tried using the `__CLING__` macro, but isMacroDefined did not work. Since `CladPlugin` is inside an archive, `#if` checks also did not work. However, the `__CLING__` macro is present in `PreprocessorOptions.Macros`, so we probably need to check the options directly.

vgvassilev:
> So I tried two things: adding the PCH-related consumer to the wrapper consumer, and adding a `GeneratePCH` function in Interpreter. This function calls `IncrementalAction::GenPCH`, where I`ncrementalAction` has a pointer to the wrapper consumer. The build completed successfully.

That seems heading in a good direction.

> 
> The second change was allowing CladPlugin to steal consumers. After that, the build was also successful, but a few tests (around 3–4 from roottest-cling) failed. Two of them failed with an error related to copyArray:
> 
> ```
> IncrementalExecutor::executeFunction: symbol '_ZN5cling7runtime8internal9copyArrayIKcEEvPT_Pvm' unresolved while linking [cling interface function]!
> 1494: You are probably missing the definition of void cling::runtime::internal::copyArray<char const>(char const*, void*, unsigned long)
> ```

That's a signal that somehow we dropped these definitions and codegen never saw them. Probably we started producing an extra `cling::Transaction` that does not get sent to codegen?

> 
> Another issue is that whenever we call Handle*-related functions in the consumer, we also need to call HandleTranslationUnit to execute delayed calls.

That's guaranteed by cling, right? end/finalize transaction should do something like that. In fact that might explain the previous observation.

> 
> I also tried using the `__CLING__` macro, but isMacroDefined did not work. Since `CladPlugin` is inside an archive, `#if` checks also did not work. However, the `__CLING__` macro is present in `PreprocessorOptions.Macros`, so we probably need to check the options directly.

Does this direction work with the tests?

SahilPatidar:
> > Another issue is that whenever we call Handle*-related functions in the consumer, we also need to call HandleTranslationUnit to execute delayed calls.
>
>That's guaranteed by cling, right? end/finalize transaction should do something like that. In fact that might explain the previous observation.

No, that is different. When we call `HandleTopLevelDecl` normally, Cling’s `DeclCollector` transforms the decl and forwards it to codegen and clad consumer. Then, in `codegenTransaction`, we call `CodeGenerator->HandleTranslationUnit()`.

`DeclCollector` does not have `HandleTranslationUnit` because Cling did not want to call it in cladplugin, due to the plugin stealing logic and the delayed call assumption, which Cling never actually used.

But now, when we use the cladplugin stealing logic, every time we call `HandleTopLevelDecl`, Clad stores it in `AppendDelayed` and forwards all calls to the other consumer (eg.`DeclCollector`) only when `HandleTranslationUnit()` is called.

> Does this direction work with the tests?

I ran these tests and passed successfully:
```
roottest-cling
roottest-root-hist
roottest-root-math
roottest-root-tree
io-stdarray

And those were failing on CI:

gtest-math-mathcore-CladDerivatorTests
minuit2_testADMinim 
tutorial-math-fit-exampleFit3D
tutorial-math-fit-minuit2GausFit 
gtest-roofit-roofitcore-testRooMinimizer
gtest-roofit-roofitcore-testRooFuncWrapper
gtest-roofit-histfactory-testHistFactory
gtest-hist-hist-TFormulaHessianTests
test-stressroofit-codegen
```

SahilPatidar:
When we add this `copyArray` (since it is a template), this happens during transformation time when we call `DeclCollector::HandleTopLevelDecl`. copyArray needs template instantiation to generate the proper decl before passing it to codegen.

Do we pass that generated decl to codegen? when and where does that happen? I mean, in which phase?

vgvassilev:
> But now, when we use the cladplugin stealing logic, every time we call `HandleTopLevelDecl`, Clad stores it in `AppendDelayed` and forwards all calls to the other consumer (eg.`DeclCollector`) only when `HandleTranslationUnit()` is called.

This is the logic that gets confused. In repl mode that should not delay it and that's what happens now with cling master and clang-repl. Why it decides to delay it?

>When we add this copyArray (since it is a template), this happens during transformation time when we call DeclCollector::HandleTopLevelDecl. copyArray needs template instantiation to generate the proper decl before passing it to codegen.
>
>Do we pass that generated decl to codegen? when and where does that happen? I mean, in which phase?

Yes, normally it will go to codegen from the template instantiator but this probably gets also delayed.

SahilPatidar:
I see two things.

First, earlier we were not using the stealing logic. `DeclCollector` was directly attached to CI, and in cling master no consumer was passed through cladplugin, so there was no delayed forwarding problem.

Second, with the current Clad design, consumer callbacks can be delayed because cladplugin stores them in `AppendDelayed()` and forwards them only when `HandleTranslationUnit()` is called.

This becomes a problem only if AST transformation or new declaration generation happens after the main parsing loop. We now call `HandleTranslationUnit()` inside `ParseOrWrapTopLevelDecl`, while cling master does not.

Also, cladplugin assumes there is a stolen consumer/multiplexer, but cling master never calls `Initialize()`, so `CladPlugin::m_Multiplexer` stays `nullptr`, which may cause a segfault.

Because of this design, delayed callbacks from cladplugin are replayed only when `HandleTranslationUnit()` runs.

In `clang-repl`, we do not do anything after this loop.

here is proof after i did some testing:

after adding the extra HandleTranslationUnit() call, the copyArray related error is gone:
```cpp
// Process any TopLevelDecls generated by #pragma weak.
for (Decl* D : S.WeakTopLevelDecls()) {
  DeclGroupRef DGR(D);
  S.getASTConsumer().HandleTopLevelDecl(DGR);
}
// Execute delayed calls to DeclCollector for transformation
// (here the copyArray expression gets added)
m_CI->getASTConsumer().HandleTranslationUnit(
    getCI()->getASTContext());
LocalInstantiations.perform();
GlobalInstantiations.perform();
// here copyArray template instantiation happens.
// this again triggers HandleTopLevelDecl(),
// but cladplugin delays it through AppendDelayed().

// execute delayed calls created during template instantiation,
// so instantiated copyArray finally reaches codegen.
m_CI->getASTConsumer().HandleTranslationUnit(
    getCI()->getASTContext());
return llvm::Error::success();
```
so the issue seems to be:

* copyArray template instantiation happens during GlobalInstantiations.perform()
* that creates new declarations
* cladplugin delays those declarations through AppendDelayed()
* without another HandleTranslationUnit() call, those delayed declarations never reach codegen

SahilPatidar:
Multiple calls to `HandleTranslationUnit()` are not causing any issue since `DeclCollector` does not implement `HandleTranslationUnit()`. We handle this separately (in cling master) by directly calling `CodeGenerator->HandleTranslationUnit()`.

SahilPatidar:
Now we have two ways to try to solve this: either handle the delayed callbacks properly, or simply disable stealing for cling and just swap the consumer front and back (CladConsumer), which matches cling’s original assumption.

vgvassilev:
> Now we have two ways to try to solve this: either handle the delayed callbacks properly, or simply disable stealing for cling and just swap the consumer front and back (CladConsumer), which matches cling’s original assumption.

Maybe do the latter...

vgvassilev:
Looks like this works on this end -- can you try if that fixes the issue on the other end and we can move forward.

SahilPatidar:
Just doing some clean-up on my root-side changes, then I’ll push them once I’m done.

vgvassilev:
@SahilPatidar, any reason why this pr is still a draft?

vgvassilev (approved):
LGTM!

PR #1821, opened 2026-05-12
Optimizes RestoreTracker

Vedant2005goyal:
This PR optimizes the  `store()` and `restore()` operations in RestoreTracker. 
Original implementation uses `std::map()` to store the address and values and the the average time complexity for search is `O(logN)` and the new implementation replaces it with the `std::unordered_set()` which has the avereage time complexity for search `O(1)`. 
Additionally, the old implementation creates a new vector every time we store a new value and this causes unnecessary overhead so in the new implementation uses a single contiguous buffer to store all the values.
Unknown

### DIFF
--- a/include/clad/Differentiator/RestoreTracker.h
+++ b/include/clad/Differentiator/RestoreTracker.h
@@ -4,8 +4,7 @@
 #include <cassert>
 #include <cstdint>
 #include <cstring>
-#include <map>
-#include <utility>
+#include <unordered_set>
 #include <vector>
 
 namespace clad {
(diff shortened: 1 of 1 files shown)


include/clad/Differentiator/RestoreTracker.h
```diff
@@ -20,34 +19,66 @@ namespace clad {
...
-  std::map<const Address, RawMemory> m_data;
+  std::unordered_set<void*> m_Seen;
+  std::vector<uint8_t> m_Buffer;
 
 public:
-  // Store the value and the address of `val`.
+  restore_tracker() {
+    // Prior reservation of some buffer for small and medium workloads will save
+    // us from massive reallocation overhead
+    m_Buffer.reserve(4096);
```
vgvassilev: This reserve is a memory regression in the shape master now generates. Since 5bea5283 a tracker used inside a loop is pushed per iteration onto a `clad::tape` (`clad::push(_tracker0, clad::restore_tracker());` in test/Gradient/ReverseForw.C), and the tape holds every iteration's tracker until the matching pop in the reverse sweep. Each one would carry 4 KB of capacity for the eight bytes it actually stores, so a loop of a million iterations holds about 4 GB of live buffer where the current `std::map` holds one node per record. An empty `std::vector` allocates nothing, and growing it on the first store costs one allocation, which is what the map's first node costs too. Whatever the right default is, it should be picked from a measurement of the loop shape rather than set to a constant.

vgvassilev:
Closing this, and the reason is not a judgement on the idea.

Master moved underneath the branch. Three commits have landed on `RestoreTracker.h` since the branch point, and the one that matters is 5bea5283: `restore()` is now deliberately non-destructive, because generated code calls it twice around every pullback -- once so the pullback's forward replay starts from the pre-call state, and once because that replay has just re-mutated it. The `restore()` here clears `m_Buffer` and `m_Seen`, so the second call would do nothing and every reverse-sweep consumer ahead of the call would read post-call values. That is a wrong gradient with no diagnostic, and no test would catch it. On top of that, 3adaa8aa added a `clear()` member that generated code calls, so the rebased header would not compile, and cf649273 added an `#ifdef __CUDACC__` implementation that this patch does not touch. The rebase is a rewrite, not a merge.

The idea is worth keeping and is now tracked in #2141: one flat buffer of `address|size|bytes` instead of a map of vectors. Two notes there that came out of reading this patch. The dedup set is not needed -- a tracker holds one or two records, so a linear scan over the buffer beats any set, which is what the device implementation already does. And the `reserve(4096)` would be a memory regression under the current protocol, because a tracker inside a loop now lives on a `clad::tape` with one pushed per iteration, so four kilobytes of capacity per eight bytes of payload becomes gigabytes on a long loop.

Thank you for the measurement in the description -- it is what made the allocation half of the win visible, and it carries over to #2141.

Written by ai-vgvassilev.

vgvassilev:
This is likely impossible to rebase but we keep the good idea as #2141.

PR #1825, opened 2026-05-22
Add reverse mode support for omp parallel directive

Fermats-Last-Theorem:
Adds reverse mode support for #pragma omp parallel directive. Forward mode already had support for `VisitOMPParallelDirective` from #1491 but reverse mode didn't, so standalone #pragma omp parallel blocks couldn't be differentiated.
Added tests.

### DIFF
--- a/test/Gradient/OpenMP.C
+++ b/test/Gradient/OpenMP.C
@@ -944,6 +944,48 @@ double fn23(const double *x, int n) {
 // CHECK-NEXT:         }
 // CHECK-NEXT: }
 
+double fn24(const double *x, int n) {
+    double result = 0;
+    #pragma omp parallel reduction(+:result)
+    {
+        result += x[0] * x[1];
+        result += x[1] * x[2];
+        result += x[2] * x[3];
+    }
+    return result;
+}
+
+// CHECK:  void fn24_grad(const double *x, int n, double *_d_x, int *_d_n) {
+// CHECK-NEXT:      double _d_result = 0.;
+// CHECK-NEXT:      double result = 0;
+// CHECK-NEXT:      #pragma omp parallel reduction(+: result)
+// CHECK-NEXT:          {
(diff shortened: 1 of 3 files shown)


test/Gradient/OpenMP.C
```diff
@@ -816,6 +816,33 @@ double fn20(const double* x, int n) {
...
+
+// CHECK:  void fn21_grad(const double *x, int n, double *_d_x, int *_d_n) {
+// CHECK-NEXT:      double _d_result = 0.;
+// CHECK-NEXT:      double result = 0;
+// CHECK-NEXT:      #pragma omp parallel reduction(+: result)
+// CHECK-NEXT:          {
+// CHECK-NEXT:              result += x[0] * x[0];
+// CHECK-NEXT:          }
+// CHECK-NEXT:      _d_result += 1;
+// CHECK-NEXT:      #pragma omp parallel private(result) firstprivate(_d_result)
```
parth-07, line 968: This does not look correct. The parallel code here have data-race undefined-behavior. The parallel block is modifying `_d_x[0]` but it is not protected, and instead we are unnecessarily marking `result` and `_d_result` as private.

Fermats-Last-Theorem: Hi @parth-07, thanks for the review. `_d_x[0]` does have a data race and I had been looking into this. In the reverse body of a non loop parallel directive every thread writes to shared pointers so to avoid a data race I've put them in a named critical section (clad_reverse_guard). Named to avoid deadlocks with user critical sections. This can be improved using activity analysis to skip the critical when writes are thread disjoint. As for the private variables, those were not introduced by my PR, they were part of #1641.

parth-07: Hi @Fermats-Last-Theorem, Thank you for the detailed information about this.

vgvassilev (approved):
LGTM!

PR #1826, opened 2026-06-01
Add reverse-mode support for CUDA shared memory attributes

Vedant2005goyal:
This PR enables differentiation of CUDA kernels having `__shared__` attribute. It also enable differentiation support of dynamic shared memory `extern __shared__`.  `__syncthreads` and `atomicAdd` is also enabled during the reverse sweep of the reverse mode to ensure no race condition.

### DIFF
--- a/include/clad/Differentiator/VisitorBase.h
+++ b/include/clad/Differentiator/VisitorBase.h
@@ -378,7 +378,8 @@ namespace clad {
                                        clang::Expr* Init = nullptr,
                                        bool DirectInit = false,
                                        clang::TypeSourceInfo* TSI = nullptr,
-                                       clang::StorageClass SC = clang::SC_None);
+                                       clang::StorageClass SC = clang::SC_None,
+                                       const clang::VarDecl* OrigVD = nullptr);
     /// Creates a namespace declaration and enters its context. All subsequent
     /// Stmts are built inside that namespace, until
     /// m_Sema.PopDeclContextIsUsed.
--- a/include/clad/Differentiator/VisitorBase.h
+++ b/include/clad/Differentiator/VisitorBase.h
@@ -651,7 +651,8 @@ namespace clad {
                                        clang::Expr* Init = nullptr,
                                        bool DirectInit = false,
                                        clang::TypeSourceInfo* TSI = nullptr,
-                                       clang::StorageClass SC = clang::SC_None);
+                                       clang::StorageClass SC = clang::SC_None,
+                                       const clang::VarDecl* OrigVD = nullptr);
     /// Creates a namespace declaration and enters its context. All subsequent
     /// Stmts are built inside that namespace, until
     /// m_Sema.PopDeclContextIsUsed.
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -151,6 +151,10 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     if (!m_Context.getLangOpts().CUDA)
       return false;
     if (const auto* DRE = dyn_cast<DeclRefExpr>(E)) {
+      if (const auto* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
+        if (VD->hasAttr<clang::CUDASharedAttr>())
+          return true;
+      }
       if (const auto* PVD = dyn_cast<ParmVarDecl>(DRE->getDecl())) {
         if (m_DiffReq->hasAttr<clang::CUDAGlobalAttr>())
           // Check whether this param is in the global memory of the GPU
@@ -168,6 +172,12 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     } else if (const auto* ASE = dyn_cast<ArraySubscriptExpr>(E)) {
       const auto* base =
           dyn_cast<DeclRefExpr>(ASE->getBase()->IgnoreImpCasts());
+      if (const auto* VD = dyn_cast<VarDecl>(base->getDecl())) {
+        if (VD->hasAttr<clang::CUDASharedAttr>()) {
+          const auto* idx = ASE->getIdx();
+          return !clad::utils::isInjective(idx, m_DiffReq.m_AnalysisDC);
+        }
+      }
       if (const auto* PVD = dyn_cast<ParmVarDecl>(base->getDecl())) {
         const auto* idx = ASE->getIdx();
         if (m_DiffReq->hasAttr<clang::CUDAGlobalAttr>())
@@ -314,8 +324,14 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     DeclarationNameInfo DNI = utils::BuildDeclarationNameInfo(m_Sema, name);
     DeclWithContext result = m_Builder.cloneFunction(m_DiffReq.Function, *this,
                                                      DC, loc, DNI, dFnType);
+
     m_Derivative = result.first;
 
+    if (m_DiffReq.Function->hasAttr<clang::CUDAGlobalAttr>()) {
+      auto* GlobalAtt = clang::CUDAGlobalAttr::CreateImplicit(m_Context);
+      GlobalAtt->setImplicit(false);
+      m_Derivative->addAttr(GlobalAtt);
+    }
     // Function declaration scope
     beginScope(Scope::FunctionPrototypeScope | Scope::FunctionDeclarationScope |
                Scope::DeclScope);
@@ -1958,6 +1974,10 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     // simplest way to support begin/end functions of the former and not deal
     // with the type mismatch.
     std::string FDName = FD->getNameAsString();
+    if (FDName == "__syncthreads") {
+      addToCurrentBlock(Clone(CE), direction::reverse);
+      return StmtDiff(Clone(CE), nullptr);
+    }
     if (FDName == "begin" || FDName == "end") {
       const Expr* arg = nullptr;
       if (const auto* MCE = dyn_cast<CXXMemberCallExpr>(CE))
@@ -3052,9 +3072,20 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
         dummyInit = m_Sema.ActOnInitList(noLoc, args, noLoc).get();
       }
     }
+    bool isDynamicSharedMem =
+        VD->hasAttr<CUDASharedAttr>() && VD->getType()->isIncompleteArrayType();
+
+    if (isDynamicSharedMem) {
+      QualType ElemTy = cast<IncompleteArrayType>(VD->getType().getTypePtr())
+                            ->getElementType();
+      VDDerivedType = m_Context.getPointerType(ElemTy);
+      dummyInit = nullptr; // Now perfectly in scope!
+    }
 
     StorageClass SC = isInsideOMPBlock ? SC_Static : SC_None;
 
+    if (VD->getStorageClass() == clang::SC_Extern)
+      SC = clang::SC_Extern;
     // Build the adjoint VarDecl
     VarDecl* VDDerived = nullptr;
     if (m_DiffReq.shouldHaveAdjoint(VD) &&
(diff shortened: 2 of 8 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3052,18 +3072,49 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+                                         /* dummyInit = */ nullptr, false,
+                                         nullptr, SC_None);
+        } else {
+          VDDerived = BuildGlobalVarDecl(VDDerivedType, "_d_" + CleanName,
+                                         dummyInit, false, nullptr, SC);
+        }
+        if (!isDynamicSharedMem && VD->hasAttr<clang::CUDASharedAttr>()) {
+          auto* SharedAttr = clang::CUDASharedAttr::CreateImplicit(
+              m_Context, VD->getAttr<clang::CUDASharedAttr>()->getRange());
+          SharedAttr->setImplicit(false);
```
vgvassilev: Why do we create implicit attribute and then change that -- can't we use the usual way of creating attributes?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3160,19 +3211,63 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+
+    if (VD->hasAttr<clang::CUDASharedAttr>()) {
+      auto* SharedAttr = clang::CUDASharedAttr::CreateImplicit(
+          m_Context, VD->getAttr<clang::CUDASharedAttr>()->getRange());
+      SharedAttr->setImplicit(false);
+      VDClone->addAttr(SharedAttr);
+      VDClone->setStorageClass(VD->getStorageClass() == clang::SC_Extern
+                                   ? clang::SC_Extern
+                                   : clang::SC_Static);
+    }
```
vgvassilev: I believe that code should go in its own function.

lib/Differentiator/ReverseModeVisitor.cpp
Vedant2005goyal: Thanks for the review. Older versions of Clang have a problem with the standard Create() method, so using an implicit attribute is a workaround for the problem

lib/Differentiator/ReverseModeVisitor.cpp
Vedant2005goyal: Done

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3160,19 +3205,55 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+          m_Sema
+              .BuildCStyleCastExpr(
+                  noLoc, m_Context.getTrivialTypeSourceInfo(VDDerivedType),
+                  noLoc, parenOffsetPtr)
+              .get();
+
+      initDiff.updateStmtDx(finalInit);
+    }
+
+    CloneCUDASharedAttr(VD, VDClone);
```
vgvassilev: That seems should be a part of `BuildGlobalVarDecl`.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3266,6 +3347,49 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     Stmt* ReverseResult = utils::unwrapIfSingleStmt(RCS);
     return {StmtDiff(ForwardResult, ReverseResult), EDiff};
   }
+  void ReverseModeVisitor::CloneCUDASharedAttr(const clang::VarDecl* OriginalVD,
```
vgvassilev: This does not need to be a member function, a file static if at all, should be enough.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3266,6 +3347,49 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+      auto* SharedAttr = clang::CUDASharedAttr::CreateImplicit(
+          m_Context, OriginalVD->getAttr<clang::CUDASharedAttr>()->getRange());
+      SharedAttr->setImplicit(false);
+      TargetVD->addAttr(SharedAttr);
+      TargetVD->setStorageClass(
+          OriginalVD->getStorageClass() == clang::SC_Extern ? clang::SC_Extern
+                                                            : clang::SC_Static);
+    }
+  }
+  void ReverseModeVisitor::HandleCUDASharedMemoryDecl(
```
vgvassilev: This should be inside a new file called `ReverseModeVisitorCUDA.cpp`

vgvassilev:
@Vedant2005goyal, can you rebase this pr?

Vedant2005goyal:
> @Vedant2005goyal, can you rebase this pr?
Some of the changes merged after this PR are currently causing issues. It was working before, but after rebasing onto the master CI tests are failing. So I think those changes need to be addressed before rebasing the PR.

vgvassilev (commented):
The clang-tidy errors about `m_HasDynamicSharedMem` and `m_SharedMemSizeParam` are stale: those members are gone from the current version and the latest run is clean.

---
These comments come from `ai-vgvassilev`: written by an AI reviewer and posted after @vgvassilev read them.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3690,20 +3731,52 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
                                  initDiff.getExpr(), VD->isDirectInit(),
                                  VDCloneTSI, SC);
+    if (isDynamicSharedMem && VDDerived) {
+
+      llvm::SmallVector<Expr*, 0> args;
+      Expr* sizeCall = GetFunctionCall("get_dynamic_smem_size", "clad", args);
+
+      Expr* two = ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context,
+                                                    /*val=*/2);
+      Expr* halfSize = BuildOp(BO_Div, sizeCall, two);
```
vgvassilev, line 3741: Splitting the dynamic shared memory in half is a contract the caller has to know about, and nothing states it.

The primal keeps the low half and the adjoint gets the high half, so a kernel that used to be launched with N bytes of shared memory now needs 2N. A user who launches with N gets an adjoint pointer past the end of the allocation, and writes through it land on whatever is there. Shared memory is not bounds-checked, so this does not fault, it just corrupts.

Two things would help. Say it here, in a comment that explains the layout rather than the arithmetic. And consider whether clad can check it at runtime, since `get_dynamic_smem_size()` is already being called and comparing it against what the primal needs is cheap next to a kernel launch.

lib/Differentiator/ReverseModeVisitorCUDA.cpp
```diff
@@ -0,0 +1,46 @@
...
+    if (initCall)
+      memsetCalls.push_back(initCall);
+  } else {
+    CloneCUDASharedAttr(VD, VDForward);
+    Expr* derivedRef = BuildDeclRef(VDDerived);
+    Expr* zeroIdx =
+        ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
+    Expr* arraySub = BuildArraySubscript(derivedRef, {zeroIdx});
+    QualType elemType = VDDerived->getType()->getPointeeType();
+    Expr* assignZero = BuildOp(BO_Assign, arraySub, getZeroInit(elemType));
```
vgvassilev, line 42: Only the first element of the dynamic adjoint array is zeroed, so the rest starts as garbage.

Shared memory is not zero-initialised at kernel launch, and adjoints accumulate with `+=`. Every element after `[0]` therefore accumulates onto whatever the previous block left behind, and the gradient comes out wrong with no sign that anything went wrong.

The static branch above gets this right by calling `clad::zero_init` on the whole array. The dynamic case cannot do the same because the length is not known at compile time, but it is known at run time: it is the half-size already computed for the offset, divided by the element size. A loop over that, strided by `blockDim.x` so the threads share the work, would cover it.

The current test does not catch this because it reads one element. A test that reads two would.

test/CUDA/SharedMem.cu
```diff
@@ -0,0 +1,157 @@
+// RUN: %cladclang_cuda -I%S/../../include --cuda-path=%cudapath \
+// RUN:     --cuda-gpu-arch=%cudaarch %cudaldflags -oReverseModeSharedMem.out \
+// RUN:     -Xclang -verify %s 2>&1 | %filecheck %s
+//
+// RUN: %cudarun ./ReverseModeSharedMem.out | %filecheck_exec %s
+//
+// REQUIRES: cuda-runtime
```
vgvassilev, line 7: This test only runs where there is a GPU, and the part worth running most does not need one.

`REQUIRES: cuda-runtime` restricts the whole file to the self-hosted runners. Those are also the only rows that have ever run a `.cu` test, and when they do not report, nothing does. `test/Regressions/issue-1624.cu` sat broken for two months that way.

Since e7ed02c0 there is a `cuda-compile` feature, available wherever the CUDA headers are, which is now three hosted rows. The CHECK lines above are all about generated code and need only a compiler. Only the line that runs the binary needs a device:

```suggestion
// RUN: %if cuda-runtime %{ %cudarun ./ReverseModeSharedMem.out | %filecheck_exec %s %}
//
// REQUIRES: cuda-compile
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3921,42 +3982,35 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     Stmt* DSClone = nullptr;
     if (!decls.empty())
       DSClone = BuildDeclStmt(decls);
+
+    // TARGETED FIX: Only reorder AST for CUDA Shared Memory to pass
```
vgvassilev, line 3986: A comment saying a reordering exists to pass a checker does not say why the checker was unhappy.

`ASTIntegrity` complains when generated nodes are shared between two places in the tree, or used before they are declared. Moving the declaration statement into the forward block makes the complaint go away, but the next person cannot tell whether that fixed the cause or hid it, and "TARGETED FIX" tells them the author was not sure either.

Please say which rule was being broken and why the shared-memory case breaks it when the others do not. If the answer is that the adjoint declaration was being emitted after its first use, that is worth writing down, because it is a real ordering constraint rather than a workaround.

lib/Differentiator/ReverseModeVisitorCUDA.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+  if (!isDynamicSharedMem) {
+    CloneCUDASharedAttr(VD, VDDerived);
+    VDDerived->setStorageClass(clang::SC_Static);
+
+    CloneCUDASharedAttr(VD, VDForward);
+    VDForward->setStorageClass(clang::SC_Static);
+
+    llvm::SmallVector<Expr*, 1> args = {BuildDeclRef(VDDerived)};
+    Stmt* initCall = GetCladZeroInit(args);
```
vgvassilev, line 32: Every thread in the block runs this, and nothing here orders it against the first accumulation.

`clad::zero_init` on a `__shared__` adjoint is executed by all threads. That is harmless while they are only writing zeros, but a thread that reaches this line after another has already accumulated into the array will erase that contribution. The result is a gradient that changes between runs.

If a `__syncthreads()` is guaranteed between this and the first `+=`, this is fine and worth a comment saying so. If it is not, this needs one thread to zero, or all threads to zero followed by a barrier.

The test has a single-element array and does not vary the block size, so it would not show either way. A block spanning more than one warp would.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -155,6 +155,10 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     if (!m_Context.getLangOpts().CUDA)
       return false;
     if (const auto* DRE = dyn_cast<DeclRefExpr>(E)) {
+      if (const auto* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
+        if (VD->hasAttr<clang::CUDASharedAttr>())
```
vgvassilev, line 159: Problem: the atomics decision reads the __shared__ attribute off the adjoint's declaration, and the dynamic adjoint does not carry it. `_d_sharedMem1` is built as a plain `int*` local, so this check is false for it and the reverse sweep emits `_d_sharedMem1[0] += ...` where the static case emits `atomicAdd(&_d_sharedMem[0], ...)`; SharedMem.cu lines 33 and 62 show the two side by side. With more than one thread writing the same index, the dynamic case races and the gradient is wrong, and the one-thread test cannot see it.

How to address it: decide from the primal declaration rather than the adjoint's. The adjoint is created in DifferentiateVarDecl from the primal VarDecl, which has the attribute, so record that fact there (for example, keep the primal in m_Variables' AdjointInfo or clone the attribute onto the pointer decl the way HandleCUDASharedMemoryDecl does for the static case) and consult it here. Then add the two-element, multi-warp kernel to the test so the atomicAdd is checked and the number is right.

include/clad/Differentiator/VisitorBase.h
```diff
@@ -651,7 +651,8 @@ namespace clad {
                                        clang::Expr* Init = nullptr,
                                        bool DirectInit = false,
                                        clang::TypeSourceInfo* TSI = nullptr,
-                                       clang::StorageClass SC = clang::SC_None);
+                                       clang::StorageClass SC = clang::SC_None,
+                                       const clang::VarDecl* OrigVD = nullptr);
```
vgvassilev, line 655: Problem: this new parameter is never passed. Every BuildGlobalVarDecl call in the PR uses the six-argument form, and the shared attribute is cloned by hand in ReverseModeVisitorCUDA.cpp through CloneCUDASharedAttr instead. So the header now advertises a way to do it that nothing uses, next to a second way that does.

How to address it: keep one. Either drop OrigVD here and in VisitorBase.cpp and leave CloneCUDASharedAttr as the single path, or pass the primal VarDecl at the two creation sites in DifferentiateVarDecl and delete CloneCUDASharedAttr. The first is the smaller change.

vgvassilev (commented):
Problem: the September push fixed three of the five August findings in code (whole-array zeroing, the barrier after zeroing, the test gated on cuda-compile) and replaced the TARGETED FIX comment with a real explanation, but none of the five threads has a reply, the 2N layout is documented without a runtime check, and the two tests those threads asked for are still missing. The test reads one element with a one-thread block, so the zeroing fix, the barrier, and the race below are all unexercised. Two new findings are inline.

How to address it: reply on each August thread with what changed, and extend SharedMem.cu with a kernel that accumulates into two or more shared elements from a block wider than a warp, with CHECK-EXEC on the numbers. That one test covers the zeroing, the barrier, and the atomics finding below.

I did not build or run anything. CI is green on all 44 checks.

PR #1828, opened 2026-06-08
Handle member access on call results in reverse mode

### LINKED ISSUE #1298: Getting members of function calls isn't differentiated in the reverse mode
```
struct A {
  double val;
};

A objVal(double x) {
  return {x*x};
}

  return objVal(x).val;
}
```
is differentiated as
```
void objVal_pullback(double x, A _d_y, double *_d_x) {
    {
        *_d_x += _d_y.val * x;
        *_d_x += x * _d_y.val;
    }
}
void f_grad(double x, double y, double *_d_x, double *_d_y) {
}
```

MdSaifAliMolla:
When a function returns a record and reverse mode differentiates a member of that result, seed the corresponding member of the record adjoint before visiting the call. This makes expressions like objVal(x).val emit the needed pullback call instead of producing an empty gradient.

Add a regression test for the reported case and for a scaled use of the same expression, so the test also checks propagation of non-unit adjoints.

Fixes: #1298

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -3558,16 +3558,51 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
   }
 
   StmtDiff ReverseModeVisitor::VisitMemberExpr(const MemberExpr* ME) {
-    auto baseDiff = Visit(ME->getBase());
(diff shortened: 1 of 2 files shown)


vgvassilev:
Can we add some tests to increase coverage?

vgvassilev (approved):
LGTM. Thank you @MdSaifAliMolla!

PR #1831, opened 2026-06-09
[ci] Add LLVM 22 ASan matrix row.

vgvassilev:
Both rows consume recipe-flavor LLVM 22 builds from compiler-research/ci-workflows -- the llvm-asan and llvm-msan recipes -- and propagate the sanitizer to every clad target via LLVM_USE_SANITIZER, so the runtime clad links against matches the instrumented LLVM it was built with.

Sanitizer coverage isn't a fit for the system-flavor rows (apt-llvm.org ships a vanilla build), so this lands on top of the LLVM 22 port and only exercises the recipe-cache path.

### DIFF
--- a/tools/ClangPlugin.cpp
+++ b/tools/ClangPlugin.cpp
@@ -21,6 +21,22 @@
 #include "clang/Basic/CodeGenOptions.h"
 #include "clang/Basic/LLVM.h" // isa, dyn_cast
 #include "clang/Basic/SourceLocation.h"
+
+#ifdef _WIN32
+// <windows.h> defines function-like min/max macros that mangle
+// std::numeric_limits<>::max() in llvm/ADT/Sequence.h (included below);
+// NOMINMAX suppresses them. WIN32_LEAN_AND_MEAN trims unrelated Win32 surface.
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+#include <windows.h>
+#else
+#include <dlfcn.h>
+#endif
+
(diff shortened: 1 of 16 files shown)


tools/ClangPlugin.cpp
```diff
@@ -123,18 +139,26 @@ void InitTimers();
...
+          hm) {
+        char buf[MAX_PATH];
+        if (DWORD n = GetModuleFileNameA(hm, buf, MAX_PATH);
+            n > 0 && n < MAX_PATH)
+          CGOpts.PassPlugins.emplace_back(buf);
+      }
+#else
+      if (Dl_info info;
+          dladdr(reinterpret_cast<void*>(&InitTimers), &info) && info.dli_fname)
+        CGOpts.PassPlugins.emplace_back(info.dli_fname);
```
devajithvs, line 160: This change seems to cause failures in ROOT (LLVM22):

dladdr is resolved to libCling.so and then registered as an LLVM pass plugin.

In LLVM20 this is ignored. In LLVM22 this errors out with:

```
  error: unable to load plugin 'libCling.so':  'Plugin entry point not found in 'libCling.so'. Is this a legacy plugin?'
```

causing all rootcling dictionary generation steps to fail.

I have temporarily reverted this in https://github.com/root-project/root/pull/21658 with https://github.com/root-project/root/pull/21658/changes/0836c2023f313404c7ed7c50291dfe1b54f67cb4

Maybe we should handle this differently for `-DCLAD_BUILD_STATIC_ONLY=ON`

vgvassilev: Yes, that makes sense. Can you submit a pr.

devajithvs: Done! Here: https://github.com/vgvassilev/clad/pull/1854

PR #1833, opened 2026-06-10
Add support for omp critical directive in forward and reverse mode

Fermats-Last-Theorem:
Adds forward and reverse mode support for `pragma omp critical` directive. The critical directive doesn't outline its body in a captured region (unlike parallel), so it needs its own visitors and can't go through `VisitOMPExecutableDirective`. 
This PR adds the new forward and reverse mode visitors for the critical directive. 

OpenMP programs using critical directive can now be differentiated using clad. One simple example of such a program is this:
``` cpp
#pragma omp parallel for
for (int i = 0; i < n; i++) {
    #pragma omp critical
    {
        result += x[0] * x[1];
    }
} 
```


vgvassilev (approved):
LGTM!

PR #1835, opened 2026-06-13
Allow redefining builtin custom derivatives

### LINKED ISSUE #1376: Allow redefining builtin custom derivatives

MdSaifAliMolla:
Lookup now prioritizes custom_derivatives::shadow and falls back to builtin only when no shadow declaration exists. Shadow mismatch suppresses fallback to throw  user errors. 

Closes:  #1376.

### DIFF
--- a/test/Regressions/issue-1376.cpp
+++ b/test/Regressions/issue-1376.cpp
@@ -1,0 +1,33 @@
+// RUN: %cladclang -I%S/../../include %s -o %t 2>&1 | %filecheck %s
+// RUN: %t | %filecheck_exec %s
+
+#include "clad/Differentiator/Differentiator.h"
+
+extern "C" int printf(const char*, ...);
+
+namespace clad::custom_derivatives::overrides::std {
+template <typename T, typename dT>
+CUDA_HOST_DEVICE ValueAndPushforward<T, dT> exp_pushforward(T x, dT d_x) {
+  return {::std::exp(x), 2 * ::std::exp(x) * d_x};
+}
+}
+double differentiable_code(double x) {
+    return std::exp(x);
+}
+
+// CHECK: void differentiable_code_grad(double x, double *_d_x) {
+// CHECK-NEXT:     {
+// CHECK-NEXT:         double _r0 = 0.;
+// CHECK-NEXT:         _r0 += 1 * clad::custom_derivatives::overrides::std::exp_pushforward(x, 1.).pushforward;
+// CHECK-NEXT:         *_d_x += _r0;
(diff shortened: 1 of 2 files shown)


MdSaifAliMolla:
@vgvassilev is this approach ok? This will lead to significant refactoring(changing custom --> builtin ), so I needed to be sure.

vgvassilev:
Perhaps we can do the opposite and have a "overload/shadow" namespace that takes priority?

MdSaifAliMolla:
> Perhaps we can do the opposite and have a "overload/shadow" namespace that takes priority?

something like custom::overload::std:: ?

test/Regressions/issue-1376.cpp
```diff
@@ -0,0 +1,33 @@
+// RUN: %cladclang -I%S/../../include %s -o %t 2>&1 | %filecheck %s
+// RUN: %t | %filecheck_exec %s
+
+#include "clad/Differentiator/Differentiator.h"
+
+extern "C" int printf(const char*, ...);
+
+namespace clad::custom_derivatives::shadow::std {
```
vgvassilev:
```suggestion
namespace clad::custom_derivatives::overrides::std {
```

Seems better as shadow has different meaning in the C++ world.

vgvassilev (approved):
LGTM!

PR #1838, opened 2026-06-18
Replace static LookupResult caches with per-instance members in VisitorBase

### LINKED ISSUE #1837: Static LookupResult caches in VisitorBase cause stale AST pointers across translation units
## Summary

`GetCladTapePush()`, `GetCladTapePop()`, `GetCladTapeBack()`, and `GetCladZeroInit()` in `VisitorBase` each use a **function-local `static` variable** to cache their `LookupResult`:

```cpp
LookupResult& VisitorBase::GetCladTapePush() {
    static clad_compat::llvm_Optional<LookupResult> Result{};
    if (!Result)
      Result = LookupCladTapeMethod("push");
    return clad_compat::llvm_Optional_GetValue(Result);
}
```

`LookupResult` objects hold pointers into AST structures that are specific to a particular `Sema`/`ASTContext`. Because `static` locals persist for the entire process lifetime, these cached results survive across translation units.

### Impact

In **single-compilation** plugin use, this is benign (there is only one TU per process). However, in **interactive/Cling** environments where multiple translation units are processed within the same process, the cached `LookupResult` from TU #1 is reused for TU #2, where the AST pointers it holds are dangling. This can lead to:
- Use-after-free on AST nodes
- Silent wrong-code generation
- Hard-to-diagnose crashes in the Sema layer

### Suggested fix

Replace the four `static` caches with per-instance `llvm::Optional<LookupResult>` members on `VisitorBase`.

Files affected:
- `include/clad/Differentiator/VisitorBase.h`
- `lib/Differentiator/VisitorBase.cpp`

Elvand-Lie:
Fixes #1837

## Problem

`GetCladTapePush()`, `GetCladTapePop()`, `GetCladTapeBack()`, and `GetCladZeroInit()` in `VisitorBase` each use a function-local `static` variable to cache their `LookupResult`. Since `LookupResult` holds pointers into the current `ASTContext`/`Sema`, these statics become dangling when multiple translation units are processed in the same process (e.g., in Cling).

## Solution

Replace all four `static llvm::Optional<LookupResult>` locals with per-instance members on `VisitorBase`:

- `m_TapePushLookup`
- `m_TapePopLookup`
- `m_TapeBackLookup`
- `m_TapeZeroInitLookup`

Each lookup is performed lazily on first use and dies with the visitor instance, so there is no cross-TU contamination.

## Files changed

- `include/clad/Differentiator/VisitorBase.h` -- add four `llvm_Optional<LookupResult>` members, add `#include <vector>`
- `lib/Differentiator/VisitorBase.cpp` -- replace `static` caches with member access in all four functions

## Testing

This is an internal refactor that preserves external behavior. The stale-pointer bug requires a multi-TU interactive session (Cling) to reproduce, which is not covered by the existing test suite.

### DIFF
--- a/include/clad/Differentiator/VisitorBase.h
+++ b/include/clad/Differentiator/VisitorBase.h
@@ -21,6 +21,7 @@
 #include "clang/Basic/OperatorKinds.h"
 #include "clang/Basic/Specifiers.h"
 #include "clang/Sema/DeclSpec.h"
+#include "clang/Sema/Lookup.h"
 #include "clang/Sema/Ownership.h"
 #include "clang/Sema/ParsedAttr.h"
 #include "clang/Sema/Sema.h"
@@ -152,6 +153,7 @@ namespace clad {
     // expression to be of object type in the reverse mode as well.
     clang::Expr* m_ThisExprDerivative = nullptr;
 
+  private:
     /// Per-instance cache for tape lookup results. These are tied to the
     /// current Sema/ASTContext and must not be shared across translation units.
     clad_compat::llvm_Optional<clang::LookupResult> m_TapePushLookup;
@@ -159,6 +161,7 @@ namespace clad {
     clad_compat::llvm_Optional<clang::LookupResult> m_TapeBackLookup;
     clad_compat::llvm_Optional<clang::LookupResult> m_TapeZeroInitLookup;
 
+  protected:
     /// The currently visited statement. Useful for crash pretty-printing.
     const clang::Stmt* m_CurVisitedStmt = nullptr;
 
--- a/include/clad/Differentiator/VisitorBase.h
+++ b/include/clad/Differentiator/VisitorBase.h
@@ -34,6 +34,7 @@
 #include <cstddef>
 #include <stack>
 #include <unordered_map>
+#include <vector>
 
 namespace clang {
 class NestedNameSpecifier;
@@ -150,6 +151,13 @@ namespace clad {
     // FIXME: Fix this inconsistency, by making `this` pointer derivative
     // expression to be of object type in the reverse mode as well.
     clang::Expr* m_ThisExprDerivative = nullptr;
+
+    /// Per-instance cache for tape lookup results. These are tied to the
+    /// current Sema/ASTContext and must not be shared across translation units.
+    clad_compat::llvm_Optional<clang::LookupResult> m_TapePushLookup;
+    clad_compat::llvm_Optional<clang::LookupResult> m_TapePopLookup;
+    clad_compat::llvm_Optional<clang::LookupResult> m_TapeBackLookup;
+    clad_compat::llvm_Optional<clang::LookupResult> m_TapeZeroInitLookup;
 
     /// The currently visited statement. Useful for crash pretty-printing.
     const clang::Stmt* m_CurVisitedStmt = nullptr;
--- a/lib/Differentiator/VisitorBase.cpp
+++ b/lib/Differentiator/VisitorBase.cpp
@@ -831,7 +831,8 @@ namespace clad {
   Stmt* VisitorBase::GetCladZeroInit(llvm::MutableArrayRef<Expr*> args) {
     if (!m_TapeZeroInitLookup)
       m_TapeZeroInitLookup = LookupCladTapeMethod("zero_init");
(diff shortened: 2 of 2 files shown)


include/clad/Differentiator/VisitorBase.h
```diff
@@ -151,6 +153,15 @@ namespace clad {
     // expression to be of object type in the reverse mode as well.
     clang::Expr* m_ThisExprDerivative = nullptr;
 
+  private:
+    /// Per-instance cache for tape lookup results. These are tied to the
+    /// current Sema/ASTContext and must not be shared across translation units.
+    clad_compat::llvm_Optional<clang::LookupResult> m_TapePushLookup;
+    clad_compat::llvm_Optional<clang::LookupResult> m_TapePopLookup;
+    clad_compat::llvm_Optional<clang::LookupResult> m_TapeBackLookup;
+    clad_compat::llvm_Optional<clang::LookupResult> m_TapeZeroInitLookup;
```
vgvassilev, line 162: Four members for what is one cache, and every visitor gets its own copy of it. `DerivativeBuilder` is the object whose lifetime actually matches what you want here: `CladPlugin` owns exactly one, as a `unique_ptr`; it is constructed with the `Sema&` these lookups come from; and it already owns a per-builder side table of this shape in `m_GeneratedCode`. Every visitor holds `m_Builder`, so one cache there would be reached from the same call sites, would be filled once per compilation instead of once per visitor, and would die with the `Sema` that filled it without anything having to reset it. That is also what #1856 needs, which is the argument for solving it once rather than twice.

lib/Differentiator/VisitorBase.cpp
```diff
@@ -832,10 +829,10 @@ namespace clad {
...
-    static clad_compat::llvm_Optional<LookupResult> Result{};
-    if (!Result)
-      Result = LookupCladTapeMethod("zero_init");
-    LookupResult& init = clad_compat::llvm_Optional_GetValue(Result);
+    if (!m_TapeZeroInitLookup)
+      m_TapeZeroInitLookup = LookupCladTapeMethod("zero_init");
+    LookupResult& init =
+        clad_compat::llvm_Optional_GetValue(m_TapeZeroInitLookup);
     CXXScopeSpec CSS;
     CSS.Extend(m_Context, utils::GetCladNamespace(m_Sema), noLoc, noLoc);
```
vgvassilev, line 837: This line is why the patch does not yet fix the bug it describes. `utils::GetCladNamespace` keeps its `NamespaceDecl*` in a function-local static of its own in CladUtils.cpp, so in a second translation unit this `CXXScopeSpec` is extended with a namespace declaration belonging to the first one. The same call is inside `tryLookupCladMethod`, where it supplies the `DeclContext` that `LookupQualifiedName` searches — so the `LookupResult` this patch carefully rebuilds per visitor is rebuilt by searching a context from a dead `ASTContext`. #1856 is the patch that removes that static. Until one of the two carries both halves, neither one closes #1837, and the issue should not be marked fixed by this alone.

vgvassilev (commented):
This is one half of a fix and #1856 is the other half; neither closes #1837 on its own. The mechanism is on the `CSS.Extend` line below. The two patches touch different files and do not conflict with each other, though both are behind master now and need a rebase.

I would still land this one first. It is self-contained, and its cost can be bounded by reading rather than by measuring: each of the four names is looked up at most once per visitor, and a visitor is constructed once per differentiation request in `DerivativeBuilder`. `ReverseModeVisitor::MakeCladTapeFor` asks for push and pop once per taped value, so what this patch changes is once-per-process becoming once-per-derivative-function, not once-per-value. I have not measured it and cannot tell you what it costs in wall time, but it is one qualified name lookup in one namespace, bounded by the number of derivatives in the file. #1856 is a different trade, because `utils::GetCladNamespace` sits on the per-tape path there, and that distinction belongs on that PR rather than being decided for both at once.

No test here would fail with the change reverted, and I do not think the lit suite can host one, since every test runs one clad over one file in one process. Say that in the description rather than leave it implied. What would make the patch reviewable on reading alone is two claims a reader can check without building anything, and both belong in the commit message. First, that these four were the only statics of this shape in the file: they were, and after the patch nothing left in `lib/Differentiator/VisitorBase.cpp` caches an AST pointer in a static. Second, that the new members cannot outlive what they point into: `VisitorBase` is constructed from `DerivativeBuilder`, which holds the `Sema&` the lookups came from, so they cannot.

PR #1842, opened 2026-06-21
Fix TBRAnalyzer visiting the true branch twice in TraverseConditionalOperator 

leetcodez:
This PR fixes a silent bug in `TBRAnalyzer::TraverseConditionalOperator`. The analyzer was visiting `getTrueExpr()` twice and completely ignoring the false branch.

Because TBR is default-on for reverse mode, a variable that only appeared in a ternary's false branch and got overwritten later, was never recorded to the tape. The backward pass would read the corrupted value and silently display an incorrect derivative. You can see the intended logic by comparing it to `VariedAnalyzer`, which correctly visits the condition, the true branch, and the false branch.

This fix corrects the second traversal to `getFalseExpr()`, matching the correct behavior already seen in `VariedAnalyzer::TraverseConditionalOperator`.

• Added a regression test that checks gradients against both the true and false conditions. I verified on a clean build that this test correctly catches the broken 0 derivative before the fix and outputs the expected 6 afterward.
• Updated a few CHECK-NEXT lines in `test/Gradient/Assignments.C` that were failing after this fix because they were  relying on the broken false-branch tape omissions.

### DIFF
--- a/lib/Differentiator/TBRAnalyzer.cpp
+++ b/lib/Differentiator/TBRAnalyzer.cpp
@@ -247,7 +247,7 @@ bool TBRAnalyzer::TraverseConditionalOperator(clang::ConditionalOperator* CO) {
 
   auto thenBranch = std::move(m_BlockData[m_CurBlockID]);
   m_BlockData[m_CurBlockID] = std::move(elseBranch);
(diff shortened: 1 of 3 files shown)


vgvassilev (approved):
Good catch. Lgtm!

PR #1843, opened 2026-06-22
Fix CUDA global argument detection for aliased kernel parameters

### LINKED ISSUE #1623: Pullback call emitted in CUDA gradient
Example:

```

__device__ void add(int *a, int *b) { *b += *a; }

__global__ void fn(int *a) { 
    add(a, a);
}

int main() { 
}
```

Derivative:
```
__attribute__((device)) void add_pullback_0_1(int *a, int *b, int *_d_a, int *_d_b) {
    *b += *a;
    {
        int _r_d0 = *_d_b;
        atomicAdd(_d_a, _r_d0);
    }
}
void fn_grad(int *a, int *_d_a) {
    add(a, a);
    // pullback missing
}
```

MdSaifAliMolla:
closes #1623 and 
```        
           // FIXME: If we cannot deduce whether the argument is
          // differentiable, we should still add it to CUDAGlobalArgsIndexes.
          // i.e. remove `&& PVD`.
``` 
as they are related, I believe.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -1302,11 +1302,8 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
           if (!utils::isArrayOrPointerType(paramDecl->getType()) ||
               IsDifferentiableArg)
             request.DVI.push_back(paramDecl);
-          // FIXME: If we cannot deduce whether the argument is
-          // differentiable, we should still add it to CUDAGlobalArgsIndexes.
-          // i.e. remove `&& PVD`.
           // We know we should use atomic ops here
-          if (useCUDA && PVD && IsDifferentiableArg)
+          if (useCUDA && IsDifferentiableArg)
             request.CUDAGlobalArgsIndexes.push_back(i);
         }
       }
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2050,6 +2050,9 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     // take arg by reference, we can request a derivative w.r.t. to this arg
     // using the forward mode.
     bool asGrad = !utils::canUsePushforwardInRevMode(FD);
+    // Method operators have a base like methods do but it's included in the
+    // call arguments so we have to shift the indexing of call arguments.
+    bool isMethodOperatorCall = MD && isa<CXXOperatorCallExpr>(CE);
 
     // Build the DiffRequest
     DiffRequest pullbackRequest{};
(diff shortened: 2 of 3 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2078,9 +2081,23 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
-            pullbackRequest.CUDAGlobalArgsIndexes.push_back(i);
+              argIdx < CE->getNumArgs()) {
+            const Expr* arg = CE->getArg(argIdx)->IgnoreParenImpCasts();
+            const ParmVarDecl* callerPVD = nullptr;
+            if (const auto* DRE = dyn_cast<DeclRefExpr>(arg))
+              callerPVD = dyn_cast<ParmVarDecl>(DRE->getDecl());
+            // If we can identify the caller's param and it is an
+            // independent parameter (i.e. a CUDA global arg), or if
+            // we cannot identify it, conservatively mark it.
+            if (!callerPVD || m_DiffReq.HasIndependentParameter(callerPVD))
```
vgvassilev, line 2098: Marking an argument global because it could not be identified can put an atomic on thread-local memory. The new `addr_kernel` passes `&b` for a local `int b`, and this rule marks that parameter; the same rule now applies in `DiffPlanner.cpp` too, once `&& PVD` came off. `atomicAdd` is defined for global and shared memory, and a generic-address atomic on a thread's own stack is not something CUDA promises anything about. The test misses it only because `shouldUseCudaAtomicOps` fires on a `DeclRefExpr` or an `ArraySubscriptExpr`, and the writes to `_d_b` in this callee go through `*_d_b`. Write the callee as `b[0] += a[0]` and you get `atomicAdd(&_d_b[0], ...)` on a stack address -- which compiles, so the compile-only CI would not see it.

There is a smaller point about the predicate itself. For a `__device__` caller the question you want is whether the parameter is global memory, and `m_DiffReq.CUDAGlobalArgsIndexes` answers that; `HasIndependentParameter` answers whether it is differentiable. Those coincide when the caller is the `__global__` kernel and can come apart one level down, for a device function calling another device function.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1302,11 +1302,8 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
           if (!utils::isArrayOrPointerType(paramDecl->getType()) ||
               IsDifferentiableArg)
             request.DVI.push_back(paramDecl);
-          // FIXME: If we cannot deduce whether the argument is
-          // differentiable, we should still add it to CUDAGlobalArgsIndexes.
-          // i.e. remove `&& PVD`.
           // We know we should use atomic ops here
-          if (useCUDA && PVD && IsDifferentiableArg)
+          if (useCUDA && IsDifferentiableArg)
```
vgvassilev, line 1306: The same decision is made here from `E->getArg(i)` with no shift, while the two sites in `ReverseModeVisitor.cpp` now shift by one for a `CXXOperatorCallExpr`. `DiffCollector::VisitCallExpr` is reached for operator calls as well, since `CXXOperatorCallExpr` is a `CallExpr`, so for a `__device__` functor's `operator()` inside a kernel gradient this site reads the object where the other two read the first real argument. The two then build different `CUDAGlobalArgsIndexes`, which `DerivedFnInfo` compares and `ComputeDerivativeName` mangles into the name -- two names for one pullback, which is the shape of #1623. I did not build a reproducer for it. Either way the rule now lives in three copies that have to stay in step; a helper taking the call and the parameter index would make that structural rather than a convention.

vgvassilev (commented):
Non-aliased kernels do change behaviour here, but in one narrow way. Dropping `&& PVD` adds index 0 for `device_fn(in[index], val)`, so `device_fn_pullback_1` becomes `device_fn_pullback_0_1` in two tests that were already working. The bodies under both updated `CHECK` blocks are unchanged, so nothing those kernels compute moves; what moves is the mangled name and, with it, the `DerivedFnInfo` identity, so a caller that still produces the index set `{1}` would now get a second near-identical `__device__` pullback emitted alongside. The direction is the safe one -- the rule can only add indexes, and a missing atomic is a race while a spare one is usually just slower -- with one exception, in the comment on `ReverseModeVisitor.cpp`.

What the new test proves, and what it does not. `test/CUDA/GradientKernels.cu` is `REQUIRES: cuda-compile`, and on a runner holding only the CUDA headers `%cudaldflags` becomes `-fsyntax-only`, so both compile lines stop at Sema and the `%if cuda-runtime` line never runs. The three `CHECK-EXEC` numbers are therefore checked nowhere in CI. What is checked is the part that matters for #1623: the `CHECK` blocks match the dumped derivative, so a pullback whose mangled name disagrees with its call site fails the run; and the first line compiles the device side with `--cuda-device-only -Xclang -verify`, so a `__device__` pullback calling something that exists only on the host is a Sema error and fails too. I worked the three expected numbers by hand and they come out: `*a += *a` gives 2, `a[1] += a[0]` with a seed of ones gives `2, 1`, and `addr_kernel` leaves `a` alone so the seed passes through as 1.

Both halves of the change are observed. Revert the `ReverseModeVisitor.cpp` hunks and `alias_kernel_grad`'s `device_int_add_pullback_0_1(a, a, _d_a, _d_a);` line stops matching, which is exactly #1623's symptom. Revert the `DiffPlanner.cpp` line and the two `device_fn_pullback_0_1` checks stop matching.

That second one is a separate change and would be a cleaner commit on its own: it is the FIXME quoted in the description, its only effect on working code is the rename above, and splitting it out makes the bisect for either half clean. The pull request also no longer merges against master.

PR #1844, opened 2026-06-22
Fix reference-variable VarData set leak in TBR analysis.

vgvassilev:
TraverseDeclStmt builds the dependency set for a reference (or pointer) declaration by installing a fresh std::set into VarData::m_RefData. But addVar has already constructed a reference-typed VarData with an empty set, so writing the new set over that live unique_ptr without releasing it orphaned one set per analysis pass. LeakSanitizer reports it on any function with a reference local in a loop;
test/Regressions/issue-767.cpp carried `// XFAIL: valgrind` for exactly this (the issue it is named for, an assertion failure, is unrelated and already closed).

VarData is a tagged union over a bool and two std::unique_ptr members with a no-op union destructor, so each owning member's lifetime must be managed by hand, and the existing move operations already reached through inactive union members. Give it explicit lifetime helpers (destroyValue / moveValue) used by the destructor and move ops, and add resetAsRef, which releases the active member before installing the fresh empty set. TraverseDeclStmt now calls resetAsRef instead of writing over the live member. Drop the now-stale valgrind XFAIL so the regression test guards the fix.

### DIFF
--- a/lib/Differentiator/AnalysisBase.h
+++ b/lib/Differentiator/AnalysisBase.h
@@ -88,19 +88,16 @@ struct VarData {
 
   VarData() = default;
   VarData(const VarData& other) = delete;
-  VarData(VarData&& other) noexcept : m_Type(other.m_Type) {
-    *this = std::move(other);
-  }
+  VarData(VarData&& other) noexcept { moveValue(std::move(other)); }
   VarData& operator=(const VarData& other) = delete;
   VarData& operator=(VarData&& other) noexcept {
-    m_Type = other.m_Type;
-    if (m_Type == FUND_TYPE)
-      m_Val.m_FundData = other.m_Val.m_FundData;
-    else if (m_Type == OBJ_TYPE || m_Type == ARR_TYPE)
-      m_Val.m_ArrData = std::move(other.m_Val.m_ArrData);
-    else if (m_Type == REF_TYPE)
-      m_Val.m_RefData = std::move(other.m_Val.m_RefData);
-    other.m_Type = UNDEFINED;
+    if (this != &other) {
+      // Release the active union member before overwriting it: assigning
+      // straight into the member for the new m_Type would drop the old owning
+      // member on a type switch, and writes through an inactive union member.
+      destroyValue();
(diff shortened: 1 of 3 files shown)


lib/Differentiator/AnalysisBase.h
```diff
@@ -127,6 +119,48 @@ struct VarData {
...
+    else if (m_Type == REF_TYPE)
+      std::destroy_at(std::addressof(m_Val.m_RefData));
+    m_Type = UNDEFINED;
+  }
+  /// Placement-constructs the union member for other.m_Type (it is otherwise
+  /// inactive, so assignment would be UB) and clears `other`. *this must be
+  /// UNDEFINED on entry (fresh, or after destroyValue()).
+  // `other` is consumed member-wise (its active union member is moved out) and
+  // left UNDEFINED, rather than std::move'd as a whole.
+  // NOLINTNEXTLINE(cppcoreguidelines-rvalue-reference-param-not-moved)
```
vgvassilev, line 150: Hi @voyager-jhk, this seems like a false positive. `moveValue` takes `VarData&&` but moves from other's members (e.g. std::move(other.m_Val.m_RefData)), never std::move(other) wholesale, so the check fires. That's a seems a genuine false-positive for member-wise ownership transfer (the param is consumed — other is reset to UNDEFINED).

Can you open an issue and potentially fix that in clang-tidy?

voyager-jhk: Agreed, it's a clear false positive. The check seems to strictly look for wholesale moves and entirely misses member-wise moves.

I will open an issue on the for this.

voyager-jhk: Hi @vgvassilev I found this is a known behavior (#173923). The check supports member-wise moves, but it's disabled by default.
You can resolve this by enabling cppcoreguidelines-rvalue-reference-param-not-moved.AllowPartialMove: true in your .clang-tidy config.

voyager-jhk: https://github.com/llvm/llvm-project/issues/173923

PR #1850, opened 2026-06-23
[ci] Run the valgrind job against clang-20 and runtime 22.

vgvassilev:


guitargeek:
Already fully landed: every removal the PR proposes is already in.

PR #1851, opened 2026-06-24
Add AD for std::ref, std::cref, and reference_wrapper.

a-elrawy:
- Added forward and reverse customs for std::ref, std::cref, and std::reference_wrapper in STLBuiltins.h
- Added forward- and reverse-mode tests in test/ForwardMode/Concurrency.C and test/Gradient/Concurrency.C


vgvassilev:
Can we add the reverse mode analogs for these?

a-elrawy:
> Can we add the reverse mode analogs for these?

I was thinking of focusing first on forward mode for these concurrency primitives. Reverse mode adds significant extra complexity. I'll handle reverse later in another PRs.

PR #1853, opened 2026-06-24
Support single statement loop body in omp reverse mode

Fermats-Last-Theorem:
`DifferentiateCanonicalLoop` currently assumes that an `omp parallel for` body is always a compoundStmt. When the body is a single statement the `Visit(Body)` added the reverse statements outside the omp region. 
This PR differentiates single statement bodies inside it's own blocks. Single statement loops can now be properly differentiated:
``` cpp
double f(const double* x, int n) {
  double total = 0.0;
  #pragma omp parallel for reduction(+ : total)
  for (int i = 0; i < n; i++)
    total += x[i] * x[i];   
  return total;
}
```


vgvassilev (approved):
LGTM!

PR #1854, opened 2026-06-25
Skip backend pass registration when built as static library

devajithvs:
When clad is built with CLAD_BUILD_STATIC_ONLY=ON, only static archives are produced (libcladDifferentiator.a, libcladPlugin.a) and no clad.so

In ROOT, when using clad, dladdr is resolved to libCling.so and then registered as an LLVM pass plugin.

In LLVM20 this is ignored. In LLVM22 this errors out with:

```
  error: unable to load plugin 'libCling.so':
  'Plugin entry point not found in 'libCling.so'. Is this a legacy plugin?'
```

causing all rootcling dictionary generation steps to fail.

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -10,6 +10,10 @@ set(CLAD_PLUGIN_SRC
 
 llvm_add_library(cladPlugin STATIC ${CLAD_PLUGIN_SRC})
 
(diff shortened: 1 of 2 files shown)


devajithvs:
Needed for: https://github.com/root-project/root/pull/21658

vgvassilev (approved):
LGTM!

PR #1856, opened 2026-06-26
Remove static CladUtils AST caches

Elvand-Lie:
## Problem

Several `CladUtils.cpp` helpers cache Clang AST-owned declarations or types in function-local `static` variables:

- `GetCladNamespace()` cached `NamespaceDecl*`
- `GetRestoreTrackerType()` cached `QualType`
- `GetCladMatrixOfType()`, `GetCladArrayOfType()`, `GetCladArrayRefOfType()`, and `GetCladTagOfType()` cached `TemplateDecl*`

Those objects are tied to the current `Sema` / `ASTContext`. Reusing them across translation units in a long-lived process can leave stale AST pointers, similar to the `LookupResult` cache issue fixed in #1838.

## Solution

Remove those function-local static caches and perform the lookup for the current `Sema` when needed.

## Testing

- `git diff --check HEAD`

Local build was not run because this Windows environment does not have `clang`, `llvm-config`, `cmake`, or `clang-format` on PATH.

### DIFF
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -892,17 +892,36 @@ namespace clad {
       return true;
     }
 
+    // NOLINTBEGIN(cppcoreguidelines-avoid-non-const-global-variables)
+    namespace {
+    NamespaceDecl* CachedCladNS = nullptr;
+    QualType CachedRestoreTrackerTy;
+    TemplateDecl* CachedMatrixDecl = nullptr;
+    TemplateDecl* CachedArrayDecl = nullptr;
+    TemplateDecl* CachedArrayRefDecl = nullptr;
+    TemplateDecl* CachedTagDecl = nullptr;
+    } // namespace
+    // NOLINTEND(cppcoreguidelines-avoid-non-const-global-variables)
+
+    void ResetCladUtilsASTCache(ASTContext&) {
+      CachedCladNS = nullptr;
+      CachedRestoreTrackerTy = QualType();
+      CachedMatrixDecl = nullptr;
+      CachedArrayDecl = nullptr;
+      CachedArrayRefDecl = nullptr;
+      CachedTagDecl = nullptr;
+    }
+
     NamespaceDecl* GetCladNamespace(Sema& S) {
-      static NamespaceDecl* Result = nullptr;
-      if (Result)
-        return Result;
+      if (CachedCladNS)
+        return CachedCladNS;
       DeclarationName CladName = &S.getASTContext().Idents.get("clad");
       LookupResult CladR(S, CladName, noLoc, Sema::LookupNamespaceName,
                          CLAD_COMPAT_Sema_ForVisibleRedeclaration);
       S.LookupQualifiedName(CladR, S.getASTContext().getTranslationUnitDecl());
       assert(!CladR.empty() && "cannot find clad namespace");
-      Result = cast<NamespaceDecl>(CladR.getFoundDecl());
-      return Result;
+      CachedCladNS = cast<NamespaceDecl>(CladR.getFoundDecl());
+      return CachedCladNS;
     }
 
     Expr* getZeroInit(QualType T, Sema& S) {
@@ -945,9 +964,8 @@ namespace clad {
     }
 
     clang::QualType GetRestoreTrackerType(clang::Sema& S) {
-      static QualType T;
-      if (!T.isNull())
-        return T;
+      if (!CachedRestoreTrackerTy.isNull())
+        return CachedRestoreTrackerTy;
       NamespaceDecl* CladNS = GetCladNamespace(S);
       CXXScopeSpec CSS;
       CSS.Extend(S.getASTContext(), CladNS, noLoc, noLoc);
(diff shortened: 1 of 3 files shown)


vgvassilev (commented):
That patch essentially makes things slower. While it is right in its core that we can have multiple instances of clad in the same process it is not yet a practical use-case. An alternative approach would be to provide some per plugin instance cache area which is somehow shared with the rest of the infrastructure. I think overall it is not worth the effort and the complexity right now.

Elvand-Lie:
@vgvassilev thanks again for the review. The first version was trying to avoid reusing AST-owned cached declarations/types across ASTContexts. After testing and looking at the hot path more carefully, I see why removing the caches entirely was the wrong tradeoff: it avoided stale pointers by making normal differentiation redo namespace/type/template lookups.                        
                                                                                                                                                                                                    
I reworked the PR to keep the cache fast path and reset the cached AST-owned values from `CladPlugin::Initialize(ASTContext&)`. So the patch now keeps the cached `NamespaceDecl*`, `QualType`, and `TemplateDecl*` values, but clears them when a new plugin ASTContext is initialized.                                                                                                          
                                                                                                                                                                                                    
  Validation:                                                                                                                                                                                       
  - local WSL Linux build of `clad`                                                                                                                                                                 
  - targeted lit suites: Arrays, ForwardMode, Gradient                                                                                                                                              
  - full local `check-clad`: 181 discovered, 158 passed, 23 unsupported, 0 failed                                                                                                                   
  - GitHub CI is passing across the full matrix                                                                                                                                                     

I think this version addresses the practical stale-cache concern while preserving the existing performance model. If this still feels too narrow or not worth carrying right now, I am fine closing the PR.

lib/Differentiator/CladUtils.cpp
```diff
@@ -892,17 +892,36 @@ namespace clad {
       return true;
     }
 
+    // NOLINTBEGIN(cppcoreguidelines-avoid-non-const-global-variables)
```
vgvassilev, line 895: This suppresses the check rather than answering it. The project's rule is no mutable globals for cross-call state, and `cppcoreguidelines-avoid-non-const-global-variables` is that rule enforcing itself, so a NOLINT here is an argument that has to be made in the review rather than silenced in the source.

There is also something given up that the diff does not show. A function-local static is initialized under a guard — see [stmt.dcl], https://eel.is/c++draft/stmt.dcl — and a namespace-scope pointer assigned lazily on first use is not. clad runs inside a single-threaded compiler today so nothing breaks, but it means the replacement is weaker than what it replaced before the lifetime question is even reached.

lib/Differentiator/CladUtils.cpp
```diff
@@ -892,17 +892,36 @@ namespace clad {
...
+    NamespaceDecl* CachedCladNS = nullptr;
+    QualType CachedRestoreTrackerTy;
+    TemplateDecl* CachedMatrixDecl = nullptr;
+    TemplateDecl* CachedArrayDecl = nullptr;
+    TemplateDecl* CachedArrayRefDecl = nullptr;
+    TemplateDecl* CachedTagDecl = nullptr;
+    } // namespace
+    // NOLINTEND(cppcoreguidelines-avoid-non-const-global-variables)
+
+    void ResetCladUtilsASTCache(ASTContext&) {
```
vgvassilev, line 906: The parameter is unnamed and never used. The signature says the cache is keyed on an `ASTContext` and the body says it is keyed on nothing: after this returns, the next caller refills all six from whatever `Sema` it happens to be holding. Either the cache knows which context it belongs to, or clearing it is only correct when contexts never overlap — and that assumption is not written down anywhere a future reader would find it.

vgvassilev (commented):
The rework answers the speed half of the earlier review but not the ownership half, and as it stands the patch still does not fix the thing it is about. Nothing ties the cached declarations to a context. `CladPlugin::Initialize` clears them, and the next caller refills them from whichever `Sema` happens to ask first, so two compiler instances alive at the same time still hand each other's declarations out in between resets — and that is exactly the case the description is about. `CompilerInstance` calls `ASTConsumer::Initialize` from `setASTContext` and `setASTConsumer`, so it runs once per compiler instance: in an incremental session there is one, the reset fires once, and nothing is different from today.

The per-instance cache area asked for already exists in the tree, and it is `DerivativeBuilder`. `CladPlugin` owns exactly one, as a `unique_ptr`; it is constructed with the `Sema&` these lookups come from; it already owns a side table of this exact shape in `m_GeneratedCode`; and every caller of these helpers — `ReverseModeVisitor`, `VectorForwardModeVisitor`, `ReverseModeForwPassVisitor`, `VisitorBase` — reaches it through `m_Builder`. A cache there needs no reset hook at all, because it dies with the `Sema` that filled it. The helpers are free functions taking `Sema&`, so they would need the cache handed to them alongside it, or thin accessors on the builder; either is a smaller surface than a global plus a hook somebody has to remember to call. It also subsumes #1838, whose four per-visitor members are the same question asked in the other file: one cache on the builder would serve both, and would be filled once per compilation rather than once per visitor. For the pair of patches I would land #1838 as it stands once rebased, since it is self-contained and its cost is bounded, and redo this one on the builder.

The title and the description no longer describe the patch. Both say the caches are removed and the lookup redone for the current `Sema`; the patch keeps all six and promotes them from function-local statics to file-scope globals. Whatever the patch settles on, rewrite both before it lands — the title becomes the commit message, and someone reading `git log` later will go looking for a removal that is not there.

No test here would fail with the change reverted, and I do not think the lit suite can host one, since each test is one clad over one file in one process. Saying that plainly is fine. What would make it reviewable on reading alone is one claim I could not check myself: that nothing reaches `GetCladNamespace` or the template helpers before `CladPlugin::Initialize` has run for that context. That question stops existing if the cache is owned rather than reset, which is one more argument for the builder.

Issue #1855, opened 2026-06-26
Error with stacktrace


MaxSagebaum:
After fixing the last issue. https://github.com/vgvassilev/clad/issues/1815 Another one was there. I posted this but it seems to be missed. So I open a new one here.

The reduced problem is:

```
#include "clad/Differentiator/Differentiator.h"
typedef enum { a } b;
typedef enum { c, d } e;
e f;
int *g() {
  switch (f) { case c: }
}
void h(b, e, char[], char[], int, bool, char, char *, va_list) { g(); }
char i, o, j;
int k;
void l(float) {
  va_list arg;
  h(a, d, &i, &o, k, 0, NULL, &j, arg);
}
float m;
void n() {
  l(m);
  clad::gradient(n);
}

```

Command line:
```
/bin/clang++ -fno-crash-diagnostics -fplugin=/home/msagebaum/Applications/clad/inst/lib64/clad.so -I/home/msagebaum/Applications/clad/inst/include -Xclang -plugin-arg-clad -Xclang -fgenerate-source-file -DCLAD_NO_NUM_DIFF test.cpp -c -o test.o
```

The output I get:
```
--- Begin Stmt Dump ---
code
--- End Stmt Dump ---
 #0 0x00007f76906cc569 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/lib64/libLLVM.so.21.1+0x50cc569)
 #1 0x00007f76906cc2d9 llvm::sys::CleanupOnSignal(unsigned long) (/lib64/libLLVM.so.21.1+0x50cc2d9)
 #2 0x00007f769063b955 (/lib64/libLLVM.so.21.1+0x503b955)
 #3 0x00007f769063bba6 (/lib64/libLLVM.so.21.1+0x503bba6)
 #4 0x00007f768b027290 __restore_rt (/lib64/libc.so.6+0x1a290)
 #5 0x00007f768a6e474a clad::AnalysisBase::getIDSequence(clang::Expr const*, clang::VarDecl const*&, llvm::SmallVectorImpl<llvm::FoldingSetNodeID>&) (/home/msagebaum/Applications/clad/inst/lib64/clad.so+0x8e474a)
 #6 0x00007f768a4d4041 clad::TBRAnalyzer::setIsRequired(clang::Expr const*, bool) (/home/msagebaum/Applications/clad/inst/lib64/clad.so+0x6d4041)
[...]

PR #1857, opened 2026-06-26
Guard TBR analysis for untracked pointers

Elvand-Lie:
## Summary
- guard AnalysisBase::getIDSequence when pointer/reference analysis data is missing
- add regression for reverse-mode calls with an untracked global pointer argument

Refs #1855

## Testing
- git diff --check
- Not run: local clang/cmake are not on PATH

### DIFF
--- a/lib/Differentiator/AnalysisBase.cpp
+++ b/lib/Differentiator/AnalysisBase.cpp
@@ -169,6 +169,12 @@ bool AnalysisBase::getIDSequence(const clang::Expr* E, const VarDecl*& VD,
       QualType VDType = VD->getType();
       if (VDType->isLValueReferenceType() || VDType->isPointerType()) {
         VarData* refData = getVarDataFromDecl(VD);
+        // Globals and other untracked declarations have no VarData.
+        if (!refData) {
+          IDSequence.clear();
+          VD = nullptr;
+          return false;
+        }
         if (refData->m_Type == VarData::REF_TYPE) {
           std::set<const VarDecl*>& vars = *refData->m_Val.m_RefData;
           if (vars.size() == 1) {
--- a/test/Regressions/issue-1855.cpp
+++ b/test/Regressions/issue-1855.cpp
@@ -1,0 +1,16 @@
+// RUN: %cladclang -fsyntax-only -std=c++17 -I%S/../../include %s
+
+#include "clad/Differentiator/Differentiator.h"
+
+int* global_ptr;
+
+void use(int*) {}
+
+double fn(double x) {
+  use(global_ptr);
+  return x * x;
+}
+
+void test() {
+  auto grad_fn = clad::gradient(fn);
+}


vgvassilev (commented):
There are a lot of untested branches. I believe we can find a smaller fix to the underlying issue.

Elvand-Lie:
@vgvassilev thanks for the review. I reworked this based on your feedback and narrowed the PR to the specific TBR 
  crash path.
                                                                                                                    
  My first pass tried to harden several related AnalysisBase/TBR paths together, but after setting up a local CLAD/ 
  LLVM build and reducing the failure, the stack trace pointed to getIDSequence handling a pointer/reference        
  declaration with no VarData, such as an untracked global. This version only handles that case and falls back to   
  the existing conservative dependency handling.                                                                    
                                                                                                                    
  I replaced the regression with a smaller reverse-mode test for this path. Locally I ran git diff --check, rebuilt 
  clad, and ran the new regression with the rebuilt plugin.                                                         
                                                                                                                    
  I also checked the full reproducer. This patch gets past the TBR/getIDSequence crash, but the full case still     
  exposes a separate crash after the no-parameter-gradient diagnostic, so I changed the PR to Refs #1855 rather than
  claiming it fully closes the issue. If you have time, I’d appreciate your thoughts on whether this is the right   
  scope.

vgvassilev (approved):
LGTM!

vgvassilev:
> @vgvassilev thanks for the review. I reworked this based on your feedback and narrowed the PR to the specific TBR crash path.
> 
> My first pass tried to harden several related AnalysisBase/TBR paths together, but after setting up a local CLAD/ LLVM build and reducing the failure, the stack trace pointed to getIDSequence handling a pointer/reference declaration with no VarData, such as an untracked global. This version only handles that case and falls back to the existing conservative dependency handling.
> 
> I replaced the regression with a smaller reverse-mode test for this path. Locally I ran git diff --check, rebuilt clad, and ran the new regression with the rebuilt plugin.
> 
> I also checked the full reproducer. This patch gets past the TBR/getIDSequence crash, but the full case still exposes a separate crash after the no-parameter-gradient diagnostic, so I changed the PR to Refs #1855 rather than claiming it fully closes the issue. If you have time, I’d appreciate your thoughts on whether this is the right scope.

Thanks for explaining. I suspect @Vedant2005goyal has some thoughts already.

Vedant2005goyal:
> > @vgvassilev thanks for the review. I reworked this based on your feedback and narrowed the PR to the specific TBR crash path.
> > My first pass tried to harden several related AnalysisBase/TBR paths together, but after setting up a local CLAD/ LLVM build and reducing the failure, the stack trace pointed to getIDSequence handling a pointer/reference declaration with no VarData, such as an untracked global. This version only handles that case and falls back to the existing conservative dependency handling.
> > I replaced the regression with a smaller reverse-mode test for this path. Locally I ran git diff --check, rebuilt clad, and ran the new regression with the rebuilt plugin.
> > I also checked the full reproducer. This patch gets past the TBR/getIDSequence crash, but the full case still exposes a separate crash after the no-parameter-gradient diagnostic, so I changed the PR to Refs #1855 rather than claiming it fully closes the issue. If you have time, I’d appreciate your thoughts on whether this is the right scope.
> 
> Thanks for explaining. I suspect @Vedant2005goyal has some thoughts already.

Thanks, @Elvand-Lie, for the patch.
The secondary crash you’re observing is caused by the missing return value in g(). I already have a patch that addresses this by introducing a safe fallback, preventing the crash in that case. I'll be opening a seperate PR for that shortly

PR #1859, opened 2026-06-30
Support differentiation of lambdas with active captures

### LINKED ISSUE #1053: Change lambda function differentiation to produce valid source code and work with lambda functions with captures (fwd mode)
currently, if a function contains a lambda expression, Clad treats it as a special case of a functor. there's no general support for lambdas with captures yet. however, I believe at this point we should stop treating lambda functions as functors but instead generate separate lambda functions representing the pushforward of the initial lambda. this way Clad wouldn't need to modify the lambda class that is impossible for the user to manipulate.

pushforwards of the functions that accept lambdas as their arguments would then need to accept both the initial lambda `f` and its pushforward `d_f`.

this ensures valid source code generation.

### LINKED ISSUE #1054: Change lambda function differentiation to produce valid source code and work with lambda functions with captures (reverse mode)

### LINKED ISSUE #1769: Active Lambda Captures fail to work in Forward mode and Gradient
Clad currently fails to handle accumulation of gradients into adjoint variables both explicit and implicit. This basically means you can't use the captured variables inside the lambda body. To reproduce add this test in Lambdas.C either ForwardMode or Gradient.
```cpp
double f(double x) {
    double c = 3.0;
    auto _f = [c](double _x) {
        return _x * _x * c;
    };
    return _f(x);
}
```

Fermats-Last-Theorem:
Enables differentiation of lambdas with captures in clad. For each captured variable, the derivative lambda has an extra parameter pair. Captured Value and derivative for forward mode. Captured Value and a pointer to its adjoint for reverse mode.

Fixes #1053.
Fixes #1054.
Fixes #1769.

### DIFF
--- a/include/clad/Differentiator/VisitorBase.h
+++ b/include/clad/Differentiator/VisitorBase.h
@@ -26,6 +26,7 @@
 #include "clang/Sema/Sema.h"
 
 #include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/PrettyStackTrace.h"
 
@@ -34,9 +35,11 @@
 #include <cstddef>
 #include <stack>
 #include <unordered_map>
+#include <utility>
 
 namespace clang {
 class NestedNameSpecifier;
+class LambdaExpr;
 } // namespace clang
 
 namespace llvm {
@@ -139,6 +142,10 @@ namespace clad {
     /// See the example inside ForwardModeVisitor::VisitDeclStmt.
     std::unordered_map<const clang::VarDecl*, clang::VarDecl*>
         m_DeclReplacements;
+    /// Maps a (lambda, by value captured variable) to its creation snapshot
+    llvm::DenseMap<std::pair<const clang::LambdaExpr*, const clang::VarDecl*>,
+                   clang::VarDecl*>
+        m_LambdaCaptureSnapshots;
     /// A stack of all the blocks where the statements of the gradient function
     /// are stored (e.g., function body, if statement blocks).
     std::vector<Stmts> m_Blocks;
@@ -210,6 +217,8 @@ namespace clad {
       return S.ActOnCallExpr(V.getCurrentScope(), lambda, noLoc, {}, noLoc)
           .get();
     }
+    /// Resolve a captured variable to its clone in the enclosing derivative.
+    clang::VarDecl* getEnclosingCaptureClone(const clang::VarDecl* capVD);
     /// For a qualtype QT returns if it's type is Array or Pointer Type
     static bool isArrayOrPointerType(const clang::QualType QT) {
       return utils::isArrayOrPointerType(QT);
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -14,6 +14,7 @@
 #include "clad/Differentiator/DiffPlanner.h"
 #include "clad/Differentiator/ErrorEstimator.h"
 #include "clad/Differentiator/ParseDiffArgsTypes.h"
+#include "clad/Differentiator/PushForwardModeVisitor.h"
 #include "clad/Differentiator/VisitorBase.h"
 
 #include "clang/AST/ASTContext.h"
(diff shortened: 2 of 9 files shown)


Fermats-Last-Theorem:
@vgvassilev This PR should now be ready for review. It now builds the primal closure using `buildClonedLambda` from #1875. 
Additionally, the Kokkos reverse mode work (#783) needs active lambda captures before it can be implemented.

vgvassilev:
> @vgvassilev This PR should now be ready for review. It now builds the primal closure using `buildClonedLambda` from #1875. Additionally, the Kokkos reverse mode work (#783) needs active lambda captures before it can be implemented.

I have some version of that with some overall design improvements. Let me land this first and we can adapt that to these changes.

Fermats-Last-Theorem:
> I have some version of that with some overall design improvements. Let me land this first and we can adapt that to these changes.

Sounds good. The CI is finally green as well.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -315,6 +316,143 @@ void BaseForwardModeVisitor::SetupDerivativeParameters(
...
+    paramTypes.push_back(utils::GetParameterDerivativeType(
+        m_Sema, DiffMode::pushforward, valTy));
+  }
+  return m_Context.getFunctionType(FPT->getReturnType(), paramTypes,
+                                   FPT->getExtProtoInfo());
+}
+
+#if CLANG_VERSION_MAJOR > 16
+clang::Expr*
+BaseForwardModeVisitor::buildPushforwardLambda(const LambdaExpr* LE) {
```
vgvassilev, line 346: Problem: this reproduces the lambda-introduction sequence by hand (PushLambdaScope, ActOnLambdaExpressionAfterIntroducer, the Declarator, ActOnStartOfLambdaDefinition, three endScope calls). On master that sequence is VisitorBase::LambdaBuilder. On top of that, the per-capture type pairs are computed four times: GetLambdaPushforwardType above, the parameter loop here, GetLambdaDerivativeType in ReverseModeVisitor.h, and BuildParams in ReverseModeVisitor.cpp.

How to address it: after the rebase, use LambdaBuilder here, and put the capture pairs in one helper, say captureParamTypes(LE, mode), that both modes call for the type and for the parameters.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1055,6 +1193,78 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
...
+        if ((lambdaVD = dyn_cast<VarDecl>(DRE->getDecl())))
+          if (const Expr* vinit = lambdaVD->getInit()) {
+            if (const auto* EWC = dyn_cast<ExprWithCleanups>(vinit))
+              vinit = EWC->getSubExpr();
+            calledLambda =
+                dyn_cast_or_null<LambdaExpr>(vinit->IgnoreImplicit());
+          }
+      if (calledLambda && calledLambda->capture_size() && lambdaVD) {
+        VarDecl* dgVD = nullptr;
+        std::string dgName = "_d_" + lambdaVD->getNameAsString();
```
vgvassilev, line 1213: Problem: the derivative lambda is found by scanning m_Derivative's declarations for a variable named `_d_` plus the lambda's name. A user variable that already carries that name, or two lambdas with the same name in sibling scopes, picks the wrong one. The reverse path had this lookup before; this copies it into forward mode, and getEnclosingCaptureClone adds a second by-name fallback.

How to address it: record the derivative VarDecl when it is built, in m_DeclReplacements or a map keyed by the primal VarDecl, and look it up there. Then both by-name lookups can go.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2329,6 +2376,30 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
         OverloadedDerivedFn = BuildCallExprToFunction(
             pullbackFD, pullbackCallArgs, CUDAExecConfig);
       } else if (MD && isLambdaCallOperator(MD)) {
+        if (calledLambda) {
+          for (const LambdaCapture& Capture : calledLambda->captures()) {
```
vgvassilev, line 2380: Problem: this loop and the one in BaseForwardModeVisitor::VisitCallExpr are the same loop: find the snapshot for a by-copy capture, else the enclosing clone, then its adjoint. They differ only in taking the address of the adjoint here.

How to address it: one helper that returns the (value, adjoint) pair per capture; the reverse caller wraps the adjoint in an address-of.

test/Gradient/Lambdas.C
```diff
@@ -173,6 +179,35 @@ double f6(double i) {
 // CHECK-NEXT: }
 
 
+double f7(double i, double j) {
```
vgvassilev, line 182: Problem: f7 is the only reverse-mode case whose body reads the captured value. f3 and f4 capture c but never use it, so their new `_d_c += _d_t` lines check plumbing, not a derivative, and f4's by-reference `d` is never read either. There is no by-reference capture read in the body, no capture created inside a loop, and neither of the two shapes named in the review body.

How to address it: add one case each: a by-reference capture the body reads, with a CHECK on `*_d_d0 += ...`; a lambda created inside a loop; and the mutable and written-by-reference shapes as expected-error tests if they are diagnosed. f3 and f4 can keep their numbers and lose the plumbing CHECK lines.

vgvassilev (commented):
Problem: the branch is 168 commits behind master and conflicts in VisitorBase.h, ReverseModeVisitor.cpp and both test files. Since it was written, master gained VisitorBase::LambdaBuilder (https://github.com/vgvassilev/clad/pull/1973), which ReverseModeVisitor::buildDerivedLambda and buildClonedLambda now use. The 110-line hand-rolled lambda introduction in buildPushforwardLambda and the DeclRefExpr rewrite loop added to the reverse builder are duplicates of that scaffolding.

How to address it: rebase first, then build the pushforward lambda with LambdaBuilder and drop the rewrite loop. The inline notes below assume that rebase; the code they point at will move but the points stay.

Problem: a captureless derivative lambda fed value-and-adjoint pairs cannot represent closure state that changes between calls, and nothing here says no. Two shapes go wrong silently. A mutable lambda called twice: the pullback receives the creation snapshot _t0 on every call while the primal clone has already doubled its own copy, so the second call differentiates at the wrong point. A by-reference capture written inside the body: the pullback gets a value copy, and TBR treats the base of a const call operator as unmodified (TBRAnalyzer.cpp:467), so the enclosing sweep neither sees nor restores the write. Neither shape is in the tests; https://github.com/vgvassilev/clad/pull/2064 carries both as mutable_capture and mutating_reference_capture.

How to address it: decide the contract. Either diagnose both shapes, a mutable call operator and an assignment to a by-reference capture inside the body, so the user gets an error instead of a wrong number, or extend the design to cover them. Landing with the diagnostics is fine; landing silent is not. Say which in the description, and how this PR relates to https://github.com/vgvassilev/clad/pull/2064, which targets the same issue.

I did not build or run anything. CI was green on the July 11 head, before the baseline check existed.

Fermats-Last-Theorem:
@vgvassilev Rebased and refactored with LambdaBuilder. I have added diagnostics for mutable and by-ref-write lambdas. Their support can be handled in subsequent PRs.

include/clad/Differentiator/VisitorBase.h
```diff
@@ -266,6 +268,12 @@ namespace clad {
     /// See the example inside ForwardModeVisitor::VisitDeclStmt.
     std::unordered_map<const clang::VarDecl*, clang::VarDecl*>
         m_DeclReplacements;
+    /// Maps a (lambda, by value captured variable) to its creation snapshot
+    llvm::DenseMap<std::pair<const clang::LambdaExpr*, const clang::VarDecl*>,
+                   clang::VarDecl*>
+        m_LambdaCaptureSnapshots;
+    /// Maps a primal lambda variable to its generated derivative closure.
+    llvm::DenseMap<const clang::VarDecl*, clang::VarDecl*> m_LambdaDerivatives;
```
vgvassilev, line 276: Problem: this map holds what two existing maps already hold. At the end of DifferentiateVarDecl the code registers `m_Variables.emplace(VDClone, adjointInfoFrom(valueDx))`, and `valueDx` comes from the `BuildDeclRef(VDDerived)` built further down, which runs after the lambda branch has reassigned `VDDerived` to the derivative closure, so that entry already points at it. The line below registers `m_DeclReplacements[VD] = VDClone`. So every visitor now carries a third map for a lookup the other two already answer.

How to address it: delete it and resolve the closure at the call site as `m_Variables` keyed by `m_DeclReplacements[VD]`, which is what #2064 does for the same lookup. That also lets the call site use `buildAdjoint` instead of building a DeclRefExpr by hand, so it picks up the clone-on-read accessor rather than sharing a raw node.

include/clad/Differentiator/VisitorBase.h
```diff
@@ -266,6 +268,12 @@ namespace clad {
     /// See the example inside ForwardModeVisitor::VisitDeclStmt.
     std::unordered_map<const clang::VarDecl*, clang::VarDecl*>
         m_DeclReplacements;
+    /// Maps a (lambda, by value captured variable) to its creation snapshot
+    llvm::DenseMap<std::pair<const clang::LambdaExpr*, const clang::VarDecl*>,
+                   clang::VarDecl*>
+        m_LambdaCaptureSnapshots;
```
vgvassilev, line 274: Problem: this map is needed, since the same variable captured by two lambdas needs two snapshots, but the key does not need to be a pair. Every capture has its own FieldDecl in the closure type, reachable through `getCaptureFields`, and that one pointer already identifies both the lambda and the variable.

How to address it: key on the capture's FieldDecl and the pair goes away. Worth noting either way that both of these sit in VisitorBase, so every visitor carries them while only two subclasses ever use them. That is a placement question rather than a defect, and they are here because the two shared helpers are.

lib/Differentiator/VisitorBase.cpp
```diff
@@ -1738,4 +1747,83 @@ namespace clad {
...
+    auto it = m_DeclReplacements.find(capVD);
+    if (it != m_DeclReplacements.end() && it->second)
+      return it->second;
+    // The capture's clone is not always in m_DeclReplacements, so fall back to
+    // resolving it by name.
+    LookupResult R(m_Sema, capVD->getDeclName(), capVD->getLocation(),
+                   Sema::LookupOrdinaryName);
+    if (m_Sema.LookupName(R, getCurrentScope()) && R.isSingleResult())
+      if (auto* found = dyn_cast<VarDecl>(R.getFoundDecl()))
+        return found;
```
vgvassilev, line 1761: Problem: this resolves a capture by name when the map misses, and a name lookup finds the wrong variable when the capture is shadowed by a local of the same name in the derivative. The comment beside `m_DeclReplacements[VD] = VDClone` in DifferentiateVarDecl says the map is recorded unconditionally so that it "makes the map authoritative and lets the fallback retire", so this puts back the kind of fallback that line exists to remove.

How to address it: find which capture is missing from the map rather than resolving it by name. If the miss is real and cannot be fixed here, an assert with a note saying which case reaches it is better than a silent wrong binding.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4034,6 +4073,13 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
   }
 
   StmtDiff ReverseModeVisitor::VisitDeclStmt(const DeclStmt* DS) {
+    // Refuse an unsupported lambda before building any derivative for it.
+    for (const Decl* D : DS->decls())
+      if (const auto* VD = dyn_cast<VarDecl>(D))
+        if (const auto* LE = dyn_cast_or_null<LambdaExpr>(VD->getInit()))
```
vgvassilev, line 4079: Problem: this test is stricter than the forward one, so the two modes refuse different sets of lambdas. Forward mode unwraps ExprWithCleanups before looking for the LambdaExpr; this reads the initializer directly. A lambda that captures an object with a non-trivial destructor is wrapped, which I confirmed on an AST dump, so reverse mode never runs the check on it. `isLambdaDS` above uses the same bare test, so such a lambda was not handled specially in reverse before this change either, and the effect here is a diagnostic that does not fire rather than a new wrong answer.

How to address it: unwrap the initializer the way the forward site does. Since three places now peel the same wrappers, this one, the forward site, and the call-site lookup that finds `calledLambda`, a small `getLambdaInit(const VarDecl*)` used by all three is better than a third copy.

vgvassilev (commented):
The four points from the last round are addressed: the branch is rebased to one commit and builds its closures with LambdaBuilder, the contract is stated and enforced by diagnoseUnsupportedLambda with a test for both refused shapes, the two capture loops are one helper, and f8 and f9 now read a by-copy and a by-reference capture in the body with checks on the adjoint. What is left is smaller, and most of it is about the state the change carries. The inline notes cover that.

Problem: no test builds a closure inside a loop. The forward test runs a loop inside a lambda body, which is a different shape. A closure created per iteration is where the capture snapshot meets the per-iteration adjoint reset in DifferentiateVarDecl, and nothing exercises that today.

How to address it: one reverse-mode function that declares the lambda inside a for loop and calls it there, with a check on the generated body and a number. If it does not work yet, the diagnostics added here are the place to say so.

I did not build or run anything. Every check passes except codecov/patch, which is a coverage threshold rather than a test, and the baseline check confirms the new tests fail on master.

Issue #367, opened 2022-02-01
Clad generates functions that cannot be compiled afterwards


parth-07:
Some of the reverse mode derived functions generated by Clad cannot be compiled afterwards because they contain a variable declaration after the goto labels (used for return statements), and thus give the following error: 

```
error: cannot jump from this goto statement to its label
note: jump bypasses variable initialization
```

One example of this issue from the `test/Gradient/Gradients.C`: 
```cpp
double f_decls3(double x, double y) {
  double a = 3 * x;
  double c = 333 * y;
  if (x > 1)
    return 2 * a;  // We add a goto statement in the forward pass 
  else if (x < -1)
    return -2 * a;  // We add a variable declaration for the return value followed by a goto statement. Both in the forward pass.
  double b = a * a;  // We add a variable declaration in the forward pass
  return b;  // We add a variable declaration for the return value followed by a  goto statement. Both in the forward pass.
}
```

One way to resolve this issue is by declaring all the variables currently being declared after a `goto` statement, at the beginning of the function. We already use a similar idea for the derived variables

utkarshraj779:
@parth-07 the thing i can do in this function is to add the variable whose name is double b;
after that the thing we can impliment is
 double a = 3 * x;
  double c = 333 * y;
  double b=0;
  if (x > 1)
    goto jump1;
  else if (x < -1)
    goto jump2;
  else
  goto jump3;
  
  jump1:
    b= 2*a;
  jump2:
    b= -2*a;
  jump3:
    b=a*a;
  return b;
i think this can work

ro4i7:
Hello @parth-07 @vgvassilev, may this can solve this issue, Please give me feedback on this solution. 

To resolve this issue, we need to declare all the variables currently being declared after a goto statement at the beginning of the function. This will ensure that the variable initialization happens before the goto statement is executed. Here's how we can modify the function `f_decls3` to fix this issue:
```double f_decls3(double x, double y) {
  double a, b, c;
  a = 3 * x;
  c = 333 * y;
  if (x > 1)
    goto label1;
  else if (x < -1) {
    double temp = -2 * a;
    goto label2;
  }
  b = a * a;
label1:
  return 2 * a;
label2:
  return temp;
}
```
[...]

PR #1861, opened 2026-06-30
Replace VisitReturnStmt goto/label with named-lambda call

vgvassilev:
The previous reverse-mode encoding emitted a label inside the master reverse and a goto into it from each early-return point. This violates [stmt.dcl]/3 (https://eel.is/c++draft/stmt.dcl#3) whenever a local with a non-trivial initializer or destructor sits between the goto source and its target — the reason DifferentiateWithClad and the forward-mode counterpart have to bypass Sema::ActOnFinishFunctionBody on the derived function (https://github.com/vgvassilev/clad/issues/367).

Replace the encoding. At each early return, emit a NullStmt marker and add the adjoint-seed Reverse to the current reverse block (no label wrapping). At function-body finalization, when m_EarlyReturnMarkers is non-empty: materialise the master reverse as a [&]-capture lambda bound to an `auto`-typed VarDecl; walk the forward block to patch each marker with `{ _rev(); return; }`; emit the saved tail-return seed and `_rev();` at the natural-return position. The auto deduction makes the generated source render as `auto _rev0 = [&] {...};` rather than the unspellable closure type.

Captures: the body's DeclRefExprs were built outside the lambda scope during the original Visit, so Sema's [&] capture-default scan does not detect them. Collect every outer VarDecl referenced in the body, pass them as explicit captures (Intro.Captures), and inside the lambda scope rebuild each captured DeclRefExpr via Sema::BuildDeclRefExpr so the
RefersToEnclosingVariableOrCapture bit is set on the new DRE — codegen relies on that bit to translate references into closure FieldDecl accesses per [expr.prim.lambda.capture]/12.

m_EarlyReturnMarkers and m_NaturalReturnSeed are reset at the start of each DifferentiateWithClad call to avoid leaking state across multiple functions handled by the same visitor instance.

Status: verified correct under an asan-instrumented clad build (the f3 early-return case in test/Gradient/Loops.C produces gradient {6.00}). A release-build crash in EmitDeclRefLValue does not reproduce under+asan, indicating a layout-sensitive bug elsewhere rather than an issue with the AD logic; tracked as a follow-up.

Scaffolding for https://github.com/vgvassilev/clad/issues/367 — Sema::ActOnFinishFunctionBody restoration lands in a follow-up.

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -73,6 +73,17 @@ namespace clad {
     /// the reverse mode we also accumulate Stmts for the reverse pass which
     /// will be executed on return.
     std::vector<Stmts> m_Reverse;
+    /// Markers emitted at early-return sites in the forward block. Each is
+    /// patched with `{ _rev(); return; }` at function-body finalization,
+    /// where `_rev` is a [&] lambda wrapping the master reverse sweep.
+    /// Replaces the goto/label encoding (vgvassilev/clad#367) so the body
+    /// stays well-formed under [stmt.dcl]/2 and Sema::ActOnFinishFunctionBody.
+    llvm::SmallPtrSet<clang::Stmt*, 4> m_EarlyReturnMarkers;
+    /// The tail-return's adjoint-seed Reverse stmt, captured when the function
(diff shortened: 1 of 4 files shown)


guitargeek:
Landed as 7514327a

PR #1864, opened 2026-07-01
Fix nullptr dereference introduced by #1819

SahilPatidar:
In #1819, we prevented Clad’s consumer from replacing (stealing) the existing consumer when running under Cling’s incremental compilation. In that configuration, `m_Multiplexer` was left as nullptr.

Normally this is not an issue because Cling’s incremental compilation path does not invoke HandleTranslationUnit(). However, other code path also creates other CompilerInstances using the same invocation options. For example:
```cpp
GenerateModuleFromModuleMapAction Action;
Instance.ExecuteAction(Action);
```
This path eventually calls HandleTranslationUnit(), where Clad (and a few other places) dereferences `m_Multiplexer` without checking whether it is nullptr, resulting in a segmentation fault.

This PR fixes the issue by adding check around `m_Multiplexer` usage.

### DIFF
--- a/tools/ClangPlugin.cpp
+++ b/tools/ClangPlugin.cpp
@@ -182,14 +182,16 @@ void InitTimers();
       auto& MultiplexC = cast<MultiplexConsumer>(m_CI.getASTConsumer());
       auto& RobbedCs = ACCESS(MultiplexC, Consumers);
       assert(RobbedCs.back().get() == this && "Clad is not the last consumer");
+
+      std::vector<std::unique_ptr<ASTConsumer>> StolenConsumers;
       const auto& Macros = m_CI.getPreprocessorOpts().Macros;
       const bool IsCling = llvm::any_of(
           Macros, [](const auto& Macro) { return Macro.first == "__CLING__"; });
       if (IsCling && m_CI.getPreprocessor().isIncrementalProcessingEnabled()) {
         std::swap(RobbedCs.front(), RobbedCs.back());
+        m_Multiplexer.reset(new MultiplexConsumer(std::move(StolenConsumers)));
         return;
       }
-      std::vector<std::unique_ptr<ASTConsumer>> StolenConsumers;
 
       // The range-based for loop in MultiplexConsumer::Initialize has
       // dispatched this call. Generally, it is unsafe to delete elements while
--- a/tools/ClangPlugin.cpp
+++ b/tools/ClangPlugin.cpp
@@ -189,7 +189,7 @@ void InitTimers();
           Macros, [](const auto& Macro) { return Macro.first == "__CLING__"; });
       if (IsCling && m_CI.getPreprocessor().isIncrementalProcessingEnabled()) {
         std::swap(RobbedCs.front(), RobbedCs.back());
-        m_Multiplexer.reset(new MultiplexConsumer(std::move(StolenConsumers)));
+        m_Multiplexer = std::make_unique<MultiplexConsumer>(std::move(StolenConsumers));
         return;
       }
 
--- a/tools/ClangPlugin.cpp
+++ b/tools/ClangPlugin.cpp
@@ -183,7 +183,6 @@ void InitTimers();
       auto& RobbedCs = ACCESS(MultiplexC, Consumers);
       assert(RobbedCs.back().get() == this && "Clad is not the last consumer");
 
-      std::vector<std::unique_ptr<ASTConsumer>> StolenConsumers;
       const auto& Macros = m_CI.getPreprocessorOpts().Macros;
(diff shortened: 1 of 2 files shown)


SahilPatidar:
Should we add a check around this as well?
```cpp
    bool shouldSkipFunctionBody(clang::Decl* D) override {
      return m_Multiplexer->shouldSkipFunctionBody(D);
    }
```

vgvassilev:
> Should we add a check around this as well?
> 
> ```c++
>     bool shouldSkipFunctionBody(clang::Decl* D) override {
>       return m_Multiplexer->shouldSkipFunctionBody(D);
>     }
> ```

Not sure. Do we have a way to test this?

SahilPatidar:
No, only on the Cling side, since `m_Multiplexer` will be nullptr when Cling is enabled.

vgvassilev:
> No, only on the Cling side, since `m_Multiplexer` will be nullptr when Cling is enabled.

Ok, maybe we should -- why we did not run into these problems before?

SahilPatidar:
Previously, the Clad consumer was inside `DeclCollector`, and `Initialize()` was never called on it. Because of that, no consumer replacement happened. The Clad consumer stayed inside `DeclCollector`, where `m_Multiplexer == nullptr`.

The issue comes up when another part of the code reuses the same InvocationOptions with Cling enabled. In that case, the consumer is created through Clang’s include-directive code path, which creates a new Clad consumer using the same InvocationOptions but new CI instance. Previously, there was no check on the Clad side to prevent replacing the existing consumer. This was okay because Cling deliberately never called Initialize() in its own code path, and it didn’t need to handle Clad plugin cases outside code path. The plugin consumer setup was entirely handled by Cling: it retrieved the plugin consumer, inserted it before the CodeGenerator, and then passed it to DeclCollector.

With the new implementation, we rely on Clang’s code path to set up the plugin consumer. This means Cling no longer controls how the consumer is initialized. Since this path always calls Initialize(), I added the check to preserve the previous behavior. Now, if InvocationOptions has Cling enabled, Initialize() does not replace the existing consumer which means m_Multiplexer will remain nullptr with any code path with Cling enabled.

We could also use the original Clad logic, but the problem is that the AST transformation happens after DeclCollector is called. If Clad takes ownership of the consumer, the Clad consumer runs first, and the AST is transformed afterward. Since those transformations can modify the AST, we would have to call `HandleTranslationUnit()` again wherever the consumer is used. Otherwise, the new changes would not be visible to the other consumers at the right time. For example, template instantiations trigger getASTConsumer().Handle*() callbacks later, which was one of the issues I ran into. I tried this approach with Cling, but it caused test failures. I added extra HandleTranslationUnit() calls in a few places, but some tests were still failing.

With the current approach, the AST is still transformed after the Clad consumer runs. The only difference is that we don’t replace the existing consumers, and we insert the Clad consumer at the front of the consumer list of CI. So far, I haven’t seen any Clad-related test failures with this approach. However, this approach is only safe as long as the AST transformations never cause any issue later.

If that happens, I think we should handle the consumer order more carefully. Ideally, it would be:

* DeclCollector
* Clad consumer
* CodeGenerator

That would probably require keeping the CodeGenerator outside DeclCollector. Even with that approach, we would still need to prevent the consumer-stealing logic.

That’s why I wanted to getting feedback on the cling PR before merging the Clad-side changes. Since the original design placed Clad inside DeclCollector, I thought it would be good to get input from others to make sure moving Clad outside DeclCollector and running it before DeclCollector won’t cause problems later because of the AST transformations.

vgvassilev:
> That’s why I wanted to getting feedback on the cling PR before merging the Clad-side changes.

I thought that Shahzad said it was okay, right?

SahilPatidar:
I’m not saying the logic we came up with may be wrong. It is actually working fine in the CI build. I was just thinking about the AST transformations and the order in which CladConsumer is called.

vgvassilev:
Ok, so we land this as is?

SahilPatidar:
I wanted to know if we retested on CMSSW after I updated the PR. If yes, then we can land this, since this logic is only for Cling.

Also, what about `shouldSkipFunctionBody`? I haven’t added any check for it. If any code path calls it (mostly from clang side), we’ll have to add a check there as well.

vgvassilev:
> I wanted to know if we retested on CMSSW after I updated the PR. If yes, then we can land this, since this logic is only for Cling.
> 
> Also, what about `shouldSkipFunctionBody`? I haven’t added any check for it. If any code path calls it (mostly from clang side), we’ll have to add a check there as well.

Let’s add a protection, merge this and ask Shahzad to retrigger under the other pr.

vgvassilev (approved):
Lgtm!

PR #1867, opened 2026-07-02
[cmake] Guard `-Wno-macro-redefined` behind `NOT MSVC`

devajithvs:
When building clad as an external project, MSVC does not recognise -W flags and errors out with:

```
  cl : command line error D8021: invalid numeric argument '/Wno-macro-redefined'
```


devajithvs:
Maybe we also want to suppress this warning on windows with: https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4005?view=msvc-170

vgvassilev (approved):
LGTM!

PR #1868, opened 2026-07-03
[UA]Replace m_LoopMem back-edge test with a per-block worklist fixpoint.

leetcodez:
This PR fixes a bug in `UsefulAnalyzer` where functions containing multiple loops hang compilation indefinitely under `-enable-ua`. 
    
`m_LoopMem` was previously declared as a single state variable that was never cleared. 
This converts `m_LoopMem` from a single `VarsData` instance into a per-block `std::vector<VarsData>` state. This allows
each loop's back-edge test to independently converge on its own local fixpoint without leaking state globally across the entire function.

**TESTS:**
- Added `test/Analyses/UsefulNonTermination.cpp`: Verifies that compilation now terminates correctly and declares the correctness of the generated gradient using `CHECK-EXEC`,

### DIFF
--- a/lib/Differentiator/UsefulAnalyzer.cpp
+++ b/lib/Differentiator/UsefulAnalyzer.cpp
@@ -7,6 +7,7 @@ namespace clad {
 void UsefulAnalyzer::Analyze(const FunctionDecl* FD) {
   // Build the CFG (control-flow graph) of FD.
   m_BlockData.resize(m_AnalysisDC->getCFG()->size());
+  m_LoopMem.resize(m_AnalysisDC->getCFG()->size());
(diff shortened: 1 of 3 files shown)


vgvassilev (approved):
Lgtm!

vgvassilev:
The tag should be upfront and [UA]

leetcodez:
> The tag should be upfront and [UA]

Done.

vgvassilev:
The pr description and the commit log need to be the same.

leetcodez:
> The pr description and the commit log need to be the same.

Got it, will make sure things are consistent moving forward.

vgvassilev (approved):
LGTM!

PR #1869, opened 2026-07-04
[test] Move stl-cmath's test_func helpers into TestUtils. NFC

leetcodez:
- Moved 1D helpers from stl-cmath.cpp to shared TestUtils.h to kill duplication.
- Added multi-var central-FD helpers like fd_gradient and dot for independent checking.
- Added DotProductIdentity.cpp harness to test forward vs reverse vs FD agreementon multi-variable targets under both default and -enable-ua paths. 
- This establishesa baseline gradient testing harness to catch regressions in upcoming activity-analysis PRs.

### DIFF
--- a/test/Analyses/DotProductIdentity.cpp
+++ b/test/Analyses/DotProductIdentity.cpp
@@ -1,0 +1,54 @@
+// RUN: %cladclang %s -I%S/../../include -oDotProductIdentity.out
+// RUN: ./DotProductIdentity.out | %filecheck_exec %s
+// RUN: %cladclang -Xclang -plugin-arg-clad -Xclang -enable-ua %s \
+// RUN:     -I%S/../../include -oDotProductIdentity_ua.out
+// RUN: ./DotProductIdentity_ua.out | %filecheck_exec %s
+// CHECK-EXEC-NOT: FAIL
+
+#include "clad/Differentiator/Differentiator.h"
+#include "../TestUtils.h"
+#include <cmath>
+#include <cstdio>
+
+double h3(double a, double b, double c) {
+    return a * b + std::sin(c) * a;
+}
+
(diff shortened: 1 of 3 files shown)


vgvassilev:
What is the reason for this change?

leetcodez:
> What is the reason for this change?

The goal was to add a test harness that catches silently wrong gradients by cross-checking reverse mode against forward mode (they're computed independently).
Currently clad's gradient tests only check handwritten expected values on a few cases, this adds an automatic check that generalizes to any function.

vgvassilev:
https://github.com/vgvassilev/clad/blob/master/test/Features/stl-cmath.cpp has similar infrastructure. This seems to be duplicating parts of it.

vgvassilev:
I don’t see how the newly added test is improving coverage. These test utils should be used in the context of actual changes.

leetcodez:
@vgvassilev 
Sorry for the delay. Both of your points are addressed and I've restructured the PR.
On the duplication: get_tolerance, get_h, numerical_derivative and test_func are now defined once in test/TestUtils.h, the local copies in stl-cmath.cpp are deleted, and its macros call test_utils::test_func wich is the 53 removed lines.

Now for the coverage ,So I've dropped DotProductIdentity.cpp and the finite-difference helpers from this PR. They'll show up with the first analysis change that actually needs an independent oracle -- seeding useful analysis from all requested outputs.

TestUtils.h now includes Differentiator.h, because test_func names clad::CladFunction and clad::gradient in its signature and the header should stay self-contained.
Do you want this PR on its own, or mixed into the analysis PR that first uses the helpers. I'm happy to close this and carry the move there instead.

vgvassilev (approved):
LGTM!

PR #1870, opened 2026-07-04
Fix reverse division tape use in loops

### LINKED ISSUE #1865: Reverse mode: wrong gradient for an inline `/` divisor accumulated in a loop (`push` emitted instead of `pop`)
# Reverse mode: wrong gradient for an inline `/` divisor accumulated in a loop (`push` emitted instead of `pop`)

## Summary

In reverse mode, `clad::gradient` produces a **silently wrong** derivative (no error, no
warning) when the active input appears in the **divisor of a `/`** written as an **inline
expression**, and that division is **accumulated inside a `for` loop of ≥ 2 iterations**:

```cpp
double f(double *p) {
    double a = p[0], s = 0.0;
    for (int i = 0; i < 2; ++i) { double x = i + 1.0; s += 1.0 / (x + a * a); }
    return s;
}
```

Binding the divisor to a named local variable (`double den = x + a*a; s += 1.0/den;`) makes
the gradient correct. Naming the *quotient* instead (`double r = 1.0/(x + a*a); s += r;`) does
**not** help.

The generated reverse code re-emits `clad::push(_t1, <divisor expr>)` in the adjoint sweep where it should emit `clad::pop(_t1)` — so the forward-taped divisor is never consumed, and the divisor is recomputed from the un-restored live variables (their final-iteration values). The adjoint is therefore stale and roughly independent of the iteration count.

## Environment

- clad **v2.4** (commit `201abc0c88387c9d5ed67571e0385c763b1237b2`), bundled with ROOT 6.41.01
- Driven via ROOT/cppyy (Cling), `-std=c++20`, Linux x86-64

## Reproducer (ROOT macro)

Self-contained ROOT macro — `root -l -b -q clad_reciprocal_bug.C` (also attached
in this repo). The `f_bad`/`f_good` pair is the minimal trigger; the rest is a
characterization battery. Each case prints the reverse-mode gradient next to a
central finite difference at `a = 1.3`.

```cpp
#include "Math/CladDerivator.h"
#include <cstdio>
#include <cmath>

// ---- minimal trigger: 2-iteration loop, inline divisor ----
double f_bad(double *p) {                 // WRONG gradient
    double a = p[0], s = 0.0;
    for (int i = 0; i < 2; ++i) { double x = i + 1.0; s += 1.0 / (x + a * a); }
    return s;
}
double f_good(double *p) {                // CORRECT: divisor bound to a local
    double a = p[0], s = 0.0;
    for (int i = 0; i < 2; ++i) { double x = i + 1.0; double den = x + a * a; s += 1.0 / den; }
    return s;
}

// ---- iteration count: correct at N=1, wrong for N>=2 ----
double L1(double *p){ double a=p[0],s=0.0; for(int i=0;i<1;++i){double x=i+1.0; s+=1.0/(x+a*a);} return s; }
double L2(double *p){ double a=p[0],s=0.0; for(int i=0;i<2;++i){double x=i+1.0; s+=1.0/(x+a*a);} return s; }
double L3(double *p){ double a=p[0],s=0.0; for(int i=0;i<3;++i){double x=i+1.0; s+=1.0/(x+a*a);} return s; }
double L4(double *p){ double a=p[0],s=0.0; for(int i=0;i<4;++i){double x=i+1.0; s+=1.0/(x+a*a);} return s; }

// ---- what does / doesn't trigger it (all 4-iteration loops) ----
double t_nameden(double *p){ double a=p[0],s=0.0; for(int i=0;i<4;++i){double x=i+1.0; double d=x+a*a; s+=1.0/d;}      return s; } // OK
double t_namequo(double *p){ double a=p[0],s=0.0; for(int i=0;i<4;++i){double x=i+1.0; double r=1.0/(x+a*a); s+=r;}    return s; } // WRONG
double t_pow    (double *p){ double a=p[0],s=0.0; for(int i=0;i<4;++i){double x=i+1.0; double d=x+a*a; s+=std::pow(d,-1.0);} return s; } // OK
double t_exp    (double *p){ double a=p[0],s=0.0; for(int i=0;i<4;++i){double x=i+1.0; s+=std::exp(-a*a*x);}           return s; } // OK
double t_numer  (double *p){ double a=p[0],s=0.0; for(int i=0;i<4;++i){double x=i+1.0; s+=(a*x)/x;}                   return s; } // OK

#define CHECK(F) do {                                                          \
    auto g = clad::gradient(F);                                                \
    double p[1] = {1.3}, d[1] = {0.0}; g.execute(p, d);                        \
    double h = 1e-6, pp[1] = {p[0]+h}, pm[1] = {p[0]-h};                       \
    double fd = (F(pp) - F(pm)) / (2*h);                                       \
    printf("%-9s %-6s clad=%+.6f  fd=%+.6f\n", #F,                             \
        (std::fabs(d[0]-fd) < 1e-4*std::fmax(1.0,std::fabs(fd))) ? "OK":"WRONG",\
        d[0], fd);                                                             \
} while (0)

void clad_reciprocal_bug() {
    CHECK(f_bad);   CHECK(f_good);
    CHECK(L1); CHECK(L2); CHECK(L3); CHECK(L4);
    CHECK(t_nameden); CHECK(t_namequo); CHECK(t_pow); CHECK(t_exp); CHECK(t_numer);
}
```

Output (validated on ROOT 6.41.01 / clad v2.4):

```
f_bad     WRONG  clad=-0.381901  fd=-0.550260
f_good    OK     clad=-0.550260  fd=-0.550260
L1        OK     clad=-0.359310  fd=-0.359310   # 1 iteration: correct
L2        WRONG  clad=-0.381901  fd=-0.550260   # >=2 iterations: wrong
L3        WRONG  clad=-0.354608  fd=-0.668463
L4        WRONG  clad=-0.321225  fd=-0.748769
t_nameden OK     clad=-0.748769  fd=-0.748769   # named divisor -> fix
t_namequo WRONG  clad=-0.321225  fd=-0.748769   # naming the quotient does NOT fix
t_pow     OK     clad=-0.748769  fd=-0.748769   # pow(den,-1) with named den
t_exp     OK     clad=-0.717856  fd=-0.717856   # exp(), no division
t_numer   OK     clad=+4.000000  fd=+4.000000   # active var only in numerator
```

Analytic check for `f_bad` at `a = 1.3`:
`f = 1/(1+a²) + 1/(2+a²)`, `df/da = -2a/(1+a²)² - 2a/(2+a²)² = -0.5502600`.

Note the wrong value is nearly flat in `N` (−0.38, −0.35, −0.32) while the truth scales
(−0.55, −0.67, −0.75): the reverse sweep is not accumulating the divisor's adjoint per
iteration.

Elvand-Lie:
## Summary
- Materialize the reverse-sweep divisor once before reusing it in reverse-mode division adjoints.
- Avoid reusing the forward `clad::push(...)` expression while generating reverse code.
- Add regression coverage for inline divisor loops and named quotient cases.

Fixes: #1865

## Tests
- `python3 /usr/lib/llvm-18/build/utils/lit/lit.py -q --param clad_site_config=/home/elvand/build-clad-issue1865/test/lit.site.cfg /mnt/c/CLAD/clad/test/Regressions/issue-1865.cpp`
- `python3 /usr/lib/llvm-18/build/utils/lit/lit.py -sv --show-unsupported --param clad_site_config=/home/elvand/build-clad-issue1865/test/lit.site.cfg /mnt/c/CLAD/clad/test/Regressions` (16 passed, 1 unsupported CUDA runtime test: `issue-1624.cu`)

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2690,9 +2690,12 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       auto RDelayed = DelayedGlobalStoreAndRef(R, /*prefix=*/"_t",
                                                /*forceStore=*/true);
       StmtDiff& RResult = RDelayed.Result;
+      Expr* RRev = RResult.getRevSweepAsExpr();
+      if (dfdx())
+        RRev = StoreAndRef(RRev, direction::reverse);
       Expr* dl = nullptr;
       if (dfdx())
-        dl = BuildOp(BO_Div, dfdx(), RResult.getExpr());
+        dl = BuildOp(BO_Div, dfdx(), Clone(RRev));
       Ldiff = Visit(L, dl);
       StmtDiff LStored = Ldiff;
       // Catch the pop statement and emit it after
@@ -2715,8 +2718,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
         // produced instead of 1 / (R * R).
         Expr* dr = nullptr;
         if (dfdx()) {
-          Expr* RxR = BuildParens(
-              BuildOp(BO_Mul, RResult.getExpr(), RResult.getExpr()));
+          Expr* RxR = BuildParens(BuildOp(BO_Mul, RRev, Clone(RRev)));
           dr = BuildOp(BO_Mul, dfdx(),
                        BuildOp(UO_Minus,
                                BuildParens(BuildOp(
@@ -2882,7 +2884,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
                           direction::reverse);
         if (isInsideLoop)
           addToCurrentBlock(LCloned, direction::forward);
-        Expr* RxR = BuildParens(BuildOp(BO_Mul, RStored, RStored));
+        Expr* RxR = BuildParens(BuildOp(BO_Mul, RStored, Clone(RStored)));
         Expr* dr = BuildOp(BO_Mul, oldValue,
                            BuildOp(UO_Minus, BuildOp(BO_Div, LCloned, RxR)));
         dr = StoreAndRef(dr, direction::reverse);
--- a/test/Regressions/issue-1865.cpp
+++ b/test/Regressions/issue-1865.cpp
@@ -1,0 +1,55 @@
+// RUN: %cladclang -std=c++17 -I%S/../../include %s -o %t 2>&1 | %filecheck %s
+// RUN: %t | %filecheck_exec %s
+// XFAIL: valgrind
+
+#include "clad/Differentiator/Differentiator.h"
+#include <cstdio>
+
+double inline_divisor_2(double* p) {
+  double a = p[0], s = 0.0;
+  for (int i = 0; i < 2; ++i) {
+    double x = i + 1.0;
+    s += 1.0 / (x + a * a);
+  }
+  return s;
+}
+
+// CHECK: void inline_divisor_2_grad(double *p, double *_d_p) {
+// CHECK: double _r{{[0-9]+}} = clad::pop(_t{{[0-9]+}});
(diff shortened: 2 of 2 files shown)


guitargeek (approved):
Thank you very much! This indeed fixes the problem, reproduced with the ROOT macro in #1865.

FYI, @vgvassilev

Issue #1873, opened 2026-07-07
[crash] -enable-va: null Expr* deref in ReverseModeVisitor::VisitDeclStmt on alias-carried write.


leetcodez:
### Description
Under `-enable-va`, differentiating a function that writes through a pointer alias (`*p = x * x;`) crashes the compiler with a null `Expr*` dereference in `ReverseModeVisitor::VisitDeclStmt`. The UA-only path (`-enable-ua`) on the same shape compiles cleanly.
   
### Minimal reproducer
```cpp
#include "clad/Differentiator/Differentiator.h"
double f(double x) {
  double y = 0;
  double* p = &y;
  *p = x * x;          
  return y;
}
int main() {
  auto grad = clad::gradient(f, "x");
}
```

Command: clang++ -fplugin=/path/to/lib/clad.so -Xclang -plugin-arg-clad -Xclang -enable-va -I include crash.cpp -c
###
Backtrace
```
#0 0x00007c7b5f5a63ff llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/lib/x86_64-linux-gnu/libLLVM-18.so.18.1+0xda63ff)
#1 0x00007c7b5f5a4539 llvm::sys::RunSignalHandlers() (/lib/x86_64-linux-gnu/libLLVM-18.so.18.1+0xda4539)
#2 0x00007c7b5f4f0267 (/lib/x86_64-linux-gnu/libLLVM-18.so.18.1+0xcf0267)
#3 0x00007c7b5e045330 (/lib/x86_64-linux-gnu/libc.so.6+0x45330)
#4 0x00007c7b66e95473 clang::Expr::IgnoreCasts() (/usr/lib/llvm-18/bin/../lib/libclang-cpp.so.18.1+0xe95473)
#5 0x00007c7b5b57668a clad::ReverseModeVisitor::VisitDeclStmt(clang::DeclStmt const*) /home/utsav/projects/clad/lib/Differentiator/ReverseModeVisitor.cpp:3330:43
#6 0x00007c7b5b559ed1 clang::StmtVisitorBase<llvm::make_const_ptr, clad::ReverseModeVisitor, clad::StmtDiff>::Visit(clang::Stmt const*) /usr/lib/llvm-18/include/clang/AST/StmtNodes.inc:1474:1
[...]

PR #1874, opened 2026-07-07
[benchmark] Add activity-analysis headroom cases measuring master

leetcodez:
[benchmark] Add activity-analysis headroom cases measuring master

Add six microbenchmarks measuring allocation headroom for activity-analysis changes: a positive control, interprocedural passive work, a dead struct field, a dead array chain, branch-guarded work, and useful-but-not-varied values. Each case provides a baseline for a different kind of reduction.

Keep the kernel definitions in `benchmark/ActivityWasteKernels.h` so the benchmark and its lit correctness test exercise the same functions. The test checks analytic gradients at -O0 and -O3 -DNDEBUG, including both branch paths, zero derivative components, repeated calls with reset state, and the scratch-array upper bound. Numerical checks remain active when NDEBUG is defined.

Reset the external sink before each affected benchmark invocation and observe it after the call. Allocate input and derivative buffers outside the counted region. Snapshot allocation totals before publishing counters, so allocations made by counter-map insertion do not contaminate AllocBytes.

Measured tape-block counts on the local 64-bit LLVM/Clang 18.1.8 build:

| Case              | Range             | Tape blocks | Reduction target          |
|-------------------|-------------------|-------------|----------------------------|
| positive control  | n 64 -> 4096      | 0 -> 4      | allocation control         |
| interproc passive | work 100 -> 5000  | 50 -> 250   | interprocedural summaries |
| dead struct field | n 64 -> 4096      | 1 -> 12     | field-sensitive activity   |
| dead array chain  | n 1000 -> 10000   | 1 -> 10     | array-aware pruning        |
| branch-guarded    | n 64 -> 4096      | 0 -> 4      | usefulness analysis        |
| useful-not-varied | reps 100 -> 10000 | 3 -> 30     | variedness analysis        |

Reproduce with:
`./benchmark/ActivityWaste --benchmark_counters_tabular=true`

For this build the measured blocks are 8224 bytes. Block counts change in steps, not continuously with the range parameter: the interprocedural case uses 50 blocks for work 100, 125 and 625, then 150 at 3125. The n=64 positive-control and branch-guarded cases allocate no tape blocks, so that size is not an allocation-counter control. 
The harness follows `benchmark/MemoryComplexity.cpp`, with floating-point per-iteration division and snapshot-before-publication accounting. That existing benchmark is unchanged; sharing its instrumentation is outside this PR.

Related to issue #1873


leetcodez:
Hi @vgvassilev,

Bringing this back up for review after a thorough cleanup and rebase on current master:
1.. Benchmark fixes:
   - Corrected branch_guarded so the recurrence is bounded rather than diverging to inf.
   - Made dead_array_chain's scratch buffer function-local, ensuring the tape records genuinely dead work rather than an actively requested input.
2. Exact metrics: Re-measured the baseline table using exact tape-block counts (each block is 8224 bytes). This turns the table into step functions across range sizes rather than float averages.

**These benchmarks establish the fixed baseline against which the upcoming activity-analysis reduction PRs can measure their individual deltas.

vgvassilev:
Thank you for doing so! Are these benchmarks the minimal subset that is needed?

leetcodez:
Yes, for the activity-analysis changes I’m planning, this is the minimal set I had in mind:
one case each for calls, struct fields, arrays, branch-guarded work, and useful/not-varied values. The positive control is there to check that the counters still pick up the needed tape.

vgvassilev:
Ok, please address clang-tidy and make the comments tighter -- I feel they are redundantly terse. Also we need to make sure that the code is still valid not only faster -- we will need to have some levels of assertions but that should happen as a unit or lit test.

leetcodez:
@vgvassilev 
Updated with shorter comments and the tidy fixes. I also added a lit test that shares the benchmark kernels and checks the gradients against expected values, including both branch paths. It passes at -O0 and -O3 with NDEBUG, so the checks stay active in optimized builds. The allocation counts are unchanged.  Local clang-tidy 18 is clean.

leetcodez:
Clang-tidy has been addressed.

leetcodez:
Hi @vgvassilev please take a look at this PR when you get the time to do so.

PR #1876, opened 2026-07-07
Fix StmtClone and loop restore-tracker guard for SmallPT AD.

a-elrawy:
Deep-clone CXX construct/temporary expression arguments in StmtClone. Disable restore-tracker reverse_forw inside loops when no tracker is set. Add StmtCloneConstructRepro and ReverseModeLoopPullbackRepro lit tests.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -2016,6 +2016,10 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
             (utils::GetValueType(lastParamType) == trackerType);
       }
     }
+    if (isInsideLoop && usingRestoreTracker && !m_RestoreTracker) {
+      calleeFnForwPassFD = nullptr;
+      usingRestoreTracker = false;
+    }
 
     // FIXME: consider moving non-diff analysis to DiffPlanner.
     bool nonDiff = clad::utils::hasNonDifferentiableAttribute(CE);
--- a/lib/Differentiator/StmtClone.cpp
+++ b/lib/Differentiator/StmtClone.cpp
@@ -174,14 +174,22 @@ DEFINE_CREATE_EXPR(CXXConstCastExpr,
                     Clone(Node->getSubExpr()), Node->getTypeInfoAsWritten(),
                     Node->getOperatorLoc(), Node->getRParenLoc(),
                     Node->getAngleBrackets()))
-DEFINE_CREATE_EXPR(
-    CXXConstructExpr,
-    (Ctx, CloneType(Node->getType()), Node->getLocation(),
-     Node->getConstructor(), Node->isElidable(),
-     clad_compat::makeArrayRef(Node->getArgs(), Node->getNumArgs()),
-     Node->hadMultipleCandidates(), Node->isListInitialization(),
-     Node->isStdInitListInitialization(), Node->requiresZeroInitialization(),
-     Node->getConstructionKind(), Node->getParenOrBraceRange()))
+Stmt* StmtClone::VisitCXXConstructExpr(CXXConstructExpr* Node) {
+  llvm::SmallVector<Expr*, 4> clonedArgs;
+  clonedArgs.reserve(Node->getNumArgs());
+  for (unsigned i = 0, e = Node->getNumArgs(); i < e; ++i)
+    clonedArgs.push_back(Clone(Node->getArg(i)));
+
+  CXXConstructExpr* result = CXXConstructExpr::Create(
+      Ctx, CloneType(Node->getType()), Node->getLocation(),
+      Node->getConstructor(), Node->isElidable(),
+      clad_compat::makeArrayRef(clonedArgs.data(), clonedArgs.size()),
+      Node->hadMultipleCandidates(), Node->isListInitialization(),
+      Node->isStdInitListInitialization(), Node->requiresZeroInitialization(),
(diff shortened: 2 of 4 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2016,6 +2016,10 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
             (utils::GetValueType(lastParamType) == trackerType);
       }
     }
+    if (isInsideLoop && usingRestoreTracker && !m_RestoreTracker) {
```
vgvassilev, line 2019: This guard is far wider than the case that motivated it, and what it changes is the answer, not the crash. The condition holds for any call inside any loop whose reverse_forw takes a clad::restore_tracker, whenever we are not already building a reverse_forw -- which is most of them. Clearing calleeFnForwPassFD makes the big `if` below false, so control reaches the "Recreate the original call expression" path at the end of the function and clad emits the plain primal call. The tracker is then never filled, whatever the callee overwrote is never restored, and the pullback replays from post-call values. That is a silently wrong gradient, with no diagnostic, which is worse than the crash it was meant to avoid.

On current master it also removes the mechanism 5bea5283 added for exactly this situation. test/Gradient/RecordRestore.C calls scale(s) inside a for loop and checks for `scale_reverse_forw(s, _d_s, clad::back(_tracker0))`; with this guard in place that call is not emitted and those CHECK lines stop matching.

lib/Differentiator/StmtClone.cpp
```diff
@@ -174,14 +174,22 @@ DEFINE_CREATE_EXPR(CXXConstCastExpr,
...
                     Node->getAngleBrackets()))
-DEFINE_CREATE_EXPR(
-    CXXConstructExpr,
-    (Ctx, CloneType(Node->getType()), Node->getLocation(),
-     Node->getConstructor(), Node->isElidable(),
-     clad_compat::makeArrayRef(Node->getArgs(), Node->getNumArgs()),
-     Node->hadMultipleCandidates(), Node->isListInitialization(),
-     Node->isStdInitListInitialization(), Node->requiresZeroInitialization(),
-     Node->getConstructionKind(), Node->getParenOrBraceRange()))
+Stmt* StmtClone::VisitCXXConstructExpr(CXXConstructExpr* Node) {
```
vgvassilev, line 177: Master already has this function, loop and all, down to a comment saying the arguments must not be shared with the source; it arrived in b75bba2c. The same is true of VisitCXXTemporaryObjectExpr below. Both hunks should disappear in the rebase.

test/Gradient/ReverseModeLoopPullbackRepro.C
```diff
@@ -0,0 +1,104 @@
+// RUN: %cladclang %s -I%S/../../include -oReverseModeLoopPullbackRepro.out 2>&1
```
vgvassilev, line 1: This test cannot fail for the reason the change exists. main computes gcx, never looks at it, and returns 0, and neither RUN line pipes anything into %filecheck. So it catches a crash and nothing else -- revert the guard and, unless clad crashes, the test still passes. The guard's entire effect is on the value of the gradient, so give the test a number to check: compare grad.execute against a central difference of loop_vec_radiance, or FileCheck the generated code for the call you expect to see.

Separately, `XFAIL: valgrind` never fires. lit matches XFAIL against config.available_features only, and clad's test/lit.cfg defines no valgrind feature -- under valgrind it appends -vg to the target triple instead, which XFAIL no longer looks at. Either add the feature in lit.cfg or drop the line; as written it reads like protection that is not there.

vgvassilev (commented):
Both halves of this have landed on master already, by different commits, so nothing here applies as written. The argument cloning in StmtClone is b75bba2c, "Do not reuse AST nodes in generated derivatives." -- it rewrote VisitCXXConstructExpr and VisitCXXTemporaryObjectExpr the same way and for the same reason. The loop case for restore_tracker is 5bea5283, "Snapshot call state per loop iteration and survive the pullback replay." -- it keeps the reverse_forw and gives every iteration its own tracker through a clad::tape, rather than dropping the reverse_forw. The branch is 226 commits behind master and GitHub reports it as conflicting; the 30 green checks are from 7 July, against a base that predates both commits. Please rebase and see what survives. My reading is that only the two tests do, and they need work before they are worth keeping -- see the comment on the loop one.

On how this sits next to the other StmtClone work in flight, since a maintainer will want to know that before merging any of it: this does not collide with #2094, and it does not help #2088 either. #2094 adds four cases for statement kinds StmtClone has none for -- BinaryConditionalOperator, AttributedStmt, CXXParenListInitExpr and CXXInheritedCtorInitExpr -- placed after VisitSubstNonTypeTemplateParmExpr. This pull request rewrites two cases that already existed, in a different part of the file. The two sets of kinds are disjoint and the two regions do not overlap, so the order they land in does not matter. What #2088 is actually about is StmtClone::VisitStmt, which asserts in a debug build and returns null under NDEBUG for any kind nobody wrote a case for, while callers use the result without checking. Nothing here touches that path, so it neither improves nor entrenches it.

I read the code rather than building it, so I have not reproduced the crash the loop guard was written for, and I cannot say whether it is still there now that 5bea5283 has landed. If it is, the reproducer is worth keeping -- but as a test against the current tracker code, not as a test for this guard.

vgvassilev:
This seems superseded by b75bba2c -- apologies for the slow response. Feel free to reopen if I am missing something.

PR #1877, opened 2026-07-08
Add SmallPTDiff demo

a-elrawy:
Added a SmallPTDiff demo with full path-tracing radiance through clad, plus a test validating AD vs FD (camera, patch, light, mirror). Original SmallPT.cpp is unchanged.

### DIFF
--- a/demos/ComputerGraphics/smallpt/SmallPTDiffCommon.h
+++ b/demos/ComputerGraphics/smallpt/SmallPTDiffCommon.h
@@ -1,0 +1,189 @@
+#ifndef SMALLPT_DIFF_COMMON_H
+#define SMALLPT_DIFF_COMMON_H
+
+#include "clad/Differentiator/Differentiator.h"
+
+#include <cmath>
+#include <cstdlib>
+
+#define non_differentiable __attribute__((annotate("non_differentiable")))
+
+struct Vec {
+  double x, y, z;
+  Vec(double x_ = 0, double y_ = 0, double z_ = 0) : x(x_), y(y_), z(z_) {}
+  Vec operator+(const Vec& b) const { return Vec(x + b.x, y + b.y, z + b.z); }
+  Vec operator-(const Vec& b) const { return Vec(x - b.x, y - b.y, z - b.z); }
+  Vec operator*(double b) const { return Vec(x * b, y * b, z * b); }
+  double operator*(const Vec& b) const { return x * b.x + y * b.y + z * b.z; }
+  Vec operator%(const Vec& b) const {
+    return Vec(y * b.z - z * b.y, z * b.x - x * b.z, x * b.y - y * b.x);
+  }
+  Vec mult(const Vec& b) const { return Vec(x * b.x, y * b.y, z * b.z); }
+  Vec norm() const;
+};
+
+namespace clad {
+namespace custom_derivatives {
+namespace class_functions {
+inline void operator_plus_pullback(const Vec* a, const Vec& b, Vec d_out,
+                                   Vec* d_a, Vec* d_b) {
+  d_a->x += d_out.x;
+  d_a->y += d_out.y;
+  d_a->z += d_out.z;
+  d_b->x += d_out.x;
+  d_b->y += d_out.y;
+  d_b->z += d_out.z;
+}
+inline void operator_minus_pullback(const Vec* a, const Vec& b, Vec d_out,
+                                    Vec* d_a, Vec* d_b) {
+  d_a->x += d_out.x;
+  d_a->y += d_out.y;
+  d_a->z += d_out.z;
+  d_b->x -= d_out.x;
+  d_b->y -= d_out.y;
+  d_b->z -= d_out.z;
+}
+inline void operator_star_pullback(const Vec* a, double b, Vec d_out, Vec* d_a,
+                                   double* d_b) {
+  d_a->x += d_out.x * b;
+  d_a->y += d_out.y * b;
+  d_a->z += d_out.z * b;
+  *d_b += a->x * d_out.x + a->y * d_out.y + a->z * d_out.z;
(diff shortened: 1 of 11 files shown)


demos/ComputerGraphics/smallpt/SmallPTDiffCommon.h
```diff
@@ -0,0 +1,193 @@
...
+  d_a->x += d_out * b.x;
+  d_a->y += d_out * b.y;
+  d_a->z += d_out * b.z;
+  d_b->x += d_out * a->x;
+  d_b->y += d_out * a->y;
+  d_b->z += d_out * a->z;
+}
+inline void operator_percent_pullback(const Vec* a, const Vec& b, Vec d_out,
+                                      Vec* d_a, Vec* d_b) {
+  d_a->x += d_out.y * b.z - d_out.z * b.y;
```
vgvassilev, line 65: The x component of this adjoint has the wrong sign. For c = a % b, a.x appears in c.y with coefficient -b.z and in c.z with coefficient +b.y, so the accumulation is d_a.x += d_out.z*b.y - d_out.y*b.z. The two lines under it are right, which is what makes this look like a typo rather than a convention: d_a.y and d_a.z are b x d_out, while d_a.x as written is d_out x b. The three d_b lines are right too.

It is on a live path, not a dead one. diff_radiance computes `Vec v = w % u;` with both operands depending on the parameter being differentiated, and v goes into the diffuse bounce direction. If the AD/FD check passes as it stands, then either that path is not reached at the sample counts used or the 5% tolerance is covering it. Either way a demo is the code people copy into their own project.

```suggestion
  d_a->x += d_out.z * b.y - d_out.y * b.z;
```

demos/ComputerGraphics/smallpt/SmallPTDiffValidate.h
```diff
@@ -0,0 +1,232 @@
...
+  grad.execute(cam_x, cam_y, cam_z, &g_x, &g_y, &g_z);
+
+  bool pass = smallpt_essentially_equal(
+      g_x, smallpt_central_fd3(diff_objective_demo_aa, cam_x, cam_y, cam_z, 0));
+  pass &= smallpt_essentially_equal(
+      g_y, smallpt_central_fd3(diff_objective_demo_aa, cam_x, cam_y, cam_z, 1));
+  pass &= smallpt_essentially_equal(
+      g_z, smallpt_central_fd3(diff_objective_demo_aa, cam_x, cam_y, cam_z, 2));
+
+  printf("SMALLPT_GRAD_FD_PASS=%d\n", pass ? 1 : 0);
```
vgvassilev, line 59: When this fails on a bot, all anyone gets is SMALLPT_GRAD_FD_PASS=0. Print the two numbers as well. This is clad's gradient against a central difference of a path tracer at 5% relative tolerance, so a real regression and a finite difference whose two evaluations landed on opposite sides of a silhouette edge or a Russian-roulette outcome produce the same single digit -- and the second is not hypothetical with h=1e-4 applied to cam_z=295.6 or to mirror_x. Printing g and the finite difference on failure is the difference between a bot failure someone can read and one someone has to reproduce locally. The same applies to the other four checks in this header.

test/Gradient/SmallPTDiff.C
```diff
@@ -0,0 +1,29 @@
+// RUN: %cladclang %s -I%S/../../include -I%S/../../demos/ComputerGraphics/smallpt -oSmallPTDiff.out
+// RUN: ./SmallPTDiff.out | %filecheck_exec %s
+// XFAIL: valgrind
```
vgvassilev, line 3: This line does nothing. lit matches XFAIL against config.available_features only, and clad's test/lit.cfg defines no valgrind feature -- when running under valgrind it appends -vg to the target triple instead, which XFAIL no longer consults. So the test does run under valgrind today, and it is a path tracer rendered once for the gradient and twice more for each finite difference. CI does have a valgrind job (build-valgrind, on push, already budgeted at around two hours).

Worth deciding what you actually want before respelling it. XFAIL means lit expects a failure and reports an unexpected pass as an error, which is probably not it; UNSUPPORTED is the directive for "do not run here". Either way the feature has to exist in test/lit.cfg first.

vgvassilev (commented):
Reading this as a draft, the first thing that would need to be true is that it is one change. Right now it is three: the two library fixes from #1876, that pull request's two tests, and the demo. Both library fixes have since landed on master by other routes -- the StmtClone argument cloning as b75bba2c, and the loop restore_tracker as 5bea5283, which keeps the reverse_forw and tapes a fresh tracker per iteration instead of dropping the reverse_forw the way the hunk here does. So the first step is a rebase onto master. If the demo still works without the two library hunks, drop them and this becomes a demo pull request that can be judged as one. If it does not work, that is the more interesting finding and deserves its own pull request with a small reproducer, rather than being carried inside a demo.

On whether the demo will rot: it will not, and that is the right call. demos/ is not built by CMake at all -- test/Misc/RunDemos.C is the only thing that compiles any of them, so adding the RUN line there is what makes the demo part of check-clad. demos/ErrorEstimation is in the same position. The thing that actually needs CI confidence is not the demo but test/Gradient/SmallPTDiff.C, which runs a path tracer and compares clad's gradient to a central difference on every configuration clad tests. This branch has had no CI runs at all, so nobody has seen that pass anywhere. I would want it green across the matrix, and under the valgrind job in particular, before this leaves draft. Comments below on two things that would make that test easier to trust.

One thing that is not this pull request's problem but is worth knowing: clad's docs have no page listing the demos (#2112), so a fourteenth demo will be exactly as hard to find as the other thirteen. Given how much is here, that seems a shame.

I read the code rather than building or running it, so I have not seen the AD and FD numbers agree. Everything below about the validation comes from reading it.

PR #1880, opened 2026-07-09
Fix unused _d_i adjoints in hessian with -enable-va

### LINKED ISSUE #1665: Don't generate adjoints of counters in `weightedSum`
When differentiating ``weightedSum`` for hessians, we generate adjoints of `i`, and they are diagnosed as unused even by Clang. This issue comes down to enabling activity analysis for hessians.

MdSaifAliMolla:
shouldHaveAdjointForw now checks VariedDecls when EnableVariedAnalysis is true, and VariedDecls seeding covers forward, hessian, and hessian_diagonal modes. HandleNestedDiffRequest runs VariedAnalyzer for reverse-mode nested requests instead of unconditionally zeroing analysis flags.

Fixes #1665

### DIFF
--- a/include/clad/Differentiator/DerivativeBuilder.h
+++ b/include/clad/Differentiator/DerivativeBuilder.h
@@ -113,6 +113,7 @@ struct DerivativeAndOverload {
     clang::ASTContext& m_Context;
     DerivedFnCollector& m_DFC;
     clad::DynamicGraph<DiffRequest>& m_DiffRequestGraph;
+    OwnedAnalysisContexts& m_AllAnalysisDC;
     std::unique_ptr<utils::StmtClone> m_NodeCloner;
     clang::NamespaceDecl* m_BuiltinDerivativesNSD;
     clang::NamespaceDecl* m_NumericalDiffNSD;
@@ -178,7 +179,8 @@ struct DerivativeAndOverload {
   public:
     DerivativeBuilder(clang::Sema& S, plugin::CladPlugin& P,
                       DerivedFnCollector& DFC,
-                      clad::DynamicGraph<DiffRequest>& DRG);
+                      clad::DynamicGraph<DiffRequest>& DRG,
+                      OwnedAnalysisContexts& ADC);
     ~DerivativeBuilder();
     /// Fuction to set the error diagnostic printing value for numerical
     /// differentiation.
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -26,6 +26,7 @@
 #include "clang/AST/RecursiveASTVisitor.h"
 #include "clang/AST/Type.h"
 #include "clang/Analysis/AnalysisDeclContext.h"
+#include "clang/Analysis/CFG.h"
 #include "clang/Basic/DiagnosticSema.h"
 #include "clang/Basic/IdentifierTable.h"
 #include "clang/Basic/LLVM.h" // isa, dyn_cast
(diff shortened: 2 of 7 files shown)


include/clad/Differentiator/DerivativeBuilder.h
```diff
@@ -119,6 +121,10 @@ struct DerivativeAndOverload {
     /// A flag to keep track of whether error diagnostics are requested by user
     /// for numerical differentiation.
     bool m_PrintNumericalDiffErrorDiag = false;
+    /// Stores AnalysisDeclContexts built for nested requests that don't go
+    /// through DiffPlanner's traversal.
+    std::vector<std::unique_ptr<clang::AnalysisDeclContext>>
```
vgvassilev: Which requests do not go via the DiffPlanner? Ideally these should be zero.

MdSaifAliMolla: the nested reverse-mode request created inline by the Hessian forward→reverse pipeline bypasses DifffPlanner's normal traversal.

vgvassilev: I'd prefer fixing that issue, rather than working it around.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -690,10 +691,13 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
...
-    if (!EnableUsefulAnalysis)
-      return true;
-    auto found = m_UsefulRunInfo.UsefulDecls.find(VD);
-    return found != m_UsefulRunInfo.UsefulDecls.end();
+    if (EnableUsefulAnalysis) {
+      auto found = m_UsefulRunInfo.UsefulDecls.find(VD);
+      return found != m_UsefulRunInfo.UsefulDecls.end();
+    }
+    if (EnableVariedAnalysis)
+      return getVariedDecls().find(VD) != getVariedDecls().end();
```
vgvassilev: On master this set means something else in forward mode. It is seeded only when the function has a pushforward candidate (needsCallActivity), and the comment in DiffRequest::isVaried says those markings must feed only shouldHavePushforward. So under -enable-va a call-free function gets an empty set and loses every adjoint, and one with calls keeps extra ones. After the rebase, the forward and hessian seeding has to go through seedVariedDirection, and the pre-pass markings must not reach this.

vgvassilev, line 699: Problem: on master this set means something different in forward mode. It is seeded only when the function has a pushforward candidate (needsCallActivity), and the comment in DiffRequest::isVaried says those markings must feed only shouldHavePushforward. Read here, a call-free function under -enable-va gets an empty set and loses every adjoint, and a function with calls keeps adjoints the pre-pass over-marked.

How to address it: after the rebase, seed forward and hessian requests through seedVariedDirection, and keep the pre-pass markings out of the set this reads, either by seeding into a separate set or by only consulting it when the analysis was run for the differentiation direction. A forward-mode test with no calls under -enable-va, with a CHECK-EXEC on the number, would catch the empty-set case.

vgvassilev (commented):
Problem: the branch conflicts with master, and master has since taken a different shape in exactly this area. The planner now owns seeding and running the varied analysis (seedVariedDirection, runVariedAnalysis, and analyzeDirection for a hessian's per-row forward request). HandleNestedDiffRequest already calls m_Scheduler.Plan, which lands in DiffCollector::PlanNestedRequest; that function builds the analysis context for TBR and carries a FIXME saying the varied analysis is still forced off for nested requests. So PrepareRequestAnalysis duplicates code that now exists, and the DerivativeBuilder change works around a hook the planner already provides.

How to address it: rebase, delete PrepareRequestAnalysis and the DerivativeBuilder change, and instead make PlanNestedRequest seed and run the varied analysis when the request has EnableVariedAnalysis set, using seedVariedDirection and runVariedAnalysis. In the hessian visitor, pass EnableVariedAnalysis to the reverse request the same way EnableTBRAnalysis is passed today. That answers the July thread, which asked for the fix to live in the planner rather than around it.

I did not build or run anything. The one red CI row is a self-hosted CUDA run from July.

PR #1881, opened 2026-07-09
Group the node-reuse diagnostics so they can be silenced.

vgvassilev:
The AST-integrity checks added with the no-shared-node work emit a warning a user cannot turn off, and in a release build their findSharedNode/findPrimalSharedNode walks run on every generated derivative even when nobody wants the diagnostic -- pure overhead for a check that only ever fires on a clad bug.

Route both diagnostics into the "clad-plugin-node-reuse" group (a subgroup of clad's "clad-plugin" group) through the getCustomDiagID group overload, so -Wno-clad-plugin-node-reuse -- or the wider -Wno-clad-plugin / -Wno-plugin -- silences them. A release build then consults isIgnored and skips the walks entirely when the group is off, so the checks cost nothing once silenced; a debug build always runs them so the asserts still catch a regression.

The group overload is a recent clang addition, so a compile-time trait detects it and older clang keeps the ungrouped two-argument form.

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -51,8 +51,44 @@
 #include <cstddef>
 #include <memory>
 #include <string>
+#include <type_traits>
+#include <utility>
 
 using namespace clang;
+
+namespace {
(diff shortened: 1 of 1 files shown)


vgvassilev:
Land when/if https://github.com/llvm/llvm-project/pull/208538 lands.

PR #1882, opened 2026-07-11
Adds missing reverse mode pullback for the vector 

Vedant2005goyal:
This PR adds the missing reverse-mode pullback for basic std::vector initialization.
**Example:**
```cpp
std::vector<int> vec(size);
std::vector<int> vec1(size,allocated);
```


vgvassilev (approved):
LGTM!

PR #1885, opened 2026-07-11
Fix min/max pushforwards in Hessian generation

### LINKED ISSUE #1872: `clad::hessian` fails when `std::max` or similar functions are used
This is a blocker for the RooFit usecase of `clad::hessian`.

Reproducer as a ROOT macro:
```c++
// clad_hessian_std_max.C

#include <iostream>

double my_func(double x)
{
   return std::max(0.5, x * x);
}

#pragma clad ON
void requests()
{
   // Both of these work fine:
   //   clad::differentiate(my_func, "x");
   //   clad::gradient(my_func, "x");
   // This one fails to compile:
   clad::hessian(my_func, "x");
}
#pragma clad OFF

void clad_hessian_std_max()
{
   double h = 0;
   my_func_hessian(1.5, &h);
   // Expected output: 2 (for x*x branch). Instead: compilation of the
   // clad-generated code fails with the errors quoted above.
   std::cout << h << std::endl;
}
```

Gives you an error like this:
```txt
Processing clad_hessian_std_max.C...
In file included from input_line_8:1:
root/clad_hessian_std_max.C:36:8: error: call to implicitly-deleted default constructor of 'ValueAndPushforward<const double &, const double &>'
double my_func(double x)
       ^
root/root_build/etc//cling/plugins/include/clad/Differentiator/BuiltinDerivatives.h:21:5: note: default constructor of 'ValueAndPushforward<const double &, const double &>' is implicitly deleted because field 'value' of reference type 'const double &' would not be initialized
  T value;
    ^
<<< cling interactive line includer >>>:1:1: error: cannot assign to non-static data member 'pushforward' with const-qualified type 'const double &'

^
root/root_build/etc//cling/plugins/include/clad/Differentiator/BuiltinDerivatives.h:22:5: note: non-static data member 'pushforward' declared const here
  U pushforward;
  ~~^~~~~~~~~~~
```

Elvand-Lie:
## Summary

- return `std::min` and `std::max` pushforward results by value
- preserve exact custom-derivative matches first, then retry only eligible `ValueAndPushforward<const T&, const T&>` returns with value types
- cover two-argument and explicit-comparator min/max Hessians

Fixes: #1872

## Testing

- `Regressions/issue-1872.cpp`
- `FirstDerivative/BuiltinDerivatives.C`
- `ForwardMode/STLCustomDerivatives.C` confirms reference-returning derivatives retain exact matching
- full regression suite: 18 passed, 1 CUDA unsupported
- changed-line `clang-format --dry-run --Werror`
- `git diff --check`

## Performance

Exact-HEAD versus patched, 11 alternating Release compile runs after warm-up:

- `STLCustomDerivatives.C`: 1343.6 ms -> 1390.1 ms (+3.46%, overlapping ranges)
- `BuiltinDerivatives.C`: 1307.6 ms -> 1271.4 ms (-2.77%)

At `-O3`, all four Hessian paths inline without `ValueAndPushforward` call or object overhead.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -1038,6 +1038,39 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
             utils::MatchOverloadType(S, dTy, Found, FailedCandidates))
       return overload;
 
+    // Retry ValueAndPushforward<const T&, const T&> as
+    // ValueAndPushforward<T, T> after the exact match fails. This keeps
+    // reference-returning custom derivatives preferred while allowing
+    // assignable min/max results in higher-order differentiation.
+    do {
+      const auto* FPT = dTy->getAs<FunctionProtoType>();
+      if (!FPT)
+        break;
+      auto* CTSD = dyn_cast_or_null<ClassTemplateSpecializationDecl>(
+          FPT->getReturnType()->getAsCXXRecordDecl());
+      TemplateDecl* VPDecl =
+          utils::LookupTemplateDeclInCladNamespace(S, "ValueAndPushforward");
+      if (!CTSD || CTSD->getSpecializedTemplate() != VPDecl)
+        break;
+      const auto& tArgs = CTSD->getTemplateArgs();
+      QualType valueTy = tArgs[0].getAsType();
+      QualType pushforwardTy = tArgs[1].getAsType();
+      if (!valueTy->isLValueReferenceType() ||
+          !valueTy.getNonReferenceType().isConstQualified() ||
+          !C.hasSameType(valueTy, pushforwardTy))
+        break;
+      QualType relaxedTy = valueTy.getNonReferenceType().getUnqualifiedType();
+      QualType relaxedRetTy =
+          utils::InstantiateTemplate(S, VPDecl, {relaxedTy, relaxedTy});
+      QualType relaxedDTy = C.getFunctionType(
+          relaxedRetTy, FPT->getParamTypes(), FPT->getExtProtoInfo());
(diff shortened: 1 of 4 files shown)


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1038,6 +1038,39 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
...
+      QualType relaxedRetTy =
+          utils::InstantiateTemplate(S, VPDecl, {relaxedTy, relaxedTy});
+      QualType relaxedDTy = C.getFunctionType(
+          relaxedRetTy, FPT->getParamTypes(), FPT->getExtProtoInfo());
+      TemplateSpecCandidateSet RelaxedCandidates(R.CallContext->getBeginLoc(),
+                                                 /*ForTakingAddress=*/false);
+      if (Expr* overload =
+              utils::MatchOverloadType(S, relaxedDTy, Found, RelaxedCandidates))
+        return overload;
+    } while (false);
```
vgvassilev: This seems to be just a single iteration why the do-while-stmt?

Elvand-Lie: I think the do-while(false) wasn't intended to loop I intentionally used it as a breakable scope. 

Because we have to fall through to the diagnostic code at the bottom on failure, using break allowed me to keep the early exit checks perfectly flat without deep nesting. 

But after I noticed the code styling guide from Github actions I've now simplified the type-checking logic down to just three conditions, I've replaced it with a standard nested if in the latest push.

vgvassilev (approved):
Lgtm!

PR #1891, opened 2026-07-14
Add AD support for std::thread.

a-elrawy:
Adds forward- and reverse-mode AD for `std::thread` (construct, join, joinable; detach is rejected). Forward mode launches the worker’s pushforward so side effects through `std::ref` (and similar) update derivatives; reverse mode runs the worker pullback after join, in the constructor’s reverse sweep. Covers free functions and named functors, including zero-argument functors; lambdas and unresolved callables are diagnosed.

### DIFF
--- a/include/clad/Differentiator/STLBuiltins.h
+++ b/include/clad/Differentiator/STLBuiltins.h
@@ -817,9 +817,47 @@ template <typename T>
 clad::ValueAndAdjoint<T&, T&> conversion_operator_reverse_forw(
     clad::Tag<T&>, const ::std::reference_wrapper<T>* x,
     const ::std::reference_wrapper<T>* dx) elidable_reverse_forw;
+
+template <class T>
+inline clad::ValueAndPushforward<const T&, const T&>
+get_pushforward(const ::std::reference_wrapper<const T>* w,
+                const ::std::reference_wrapper<const T>* d_w) noexcept {
+  return {w->get(), d_w->get()};
+}
+
+template <class T>
+inline clad::ValueAndPushforward<T&, T&>
+get_pushforward(const ::std::reference_wrapper<T>* w,
+                const ::std::reference_wrapper<T>* d_w) noexcept {
+  return {w->get(), d_w->get()};
+}
+
+template <class T>
+inline clad::ValueAndPushforward<T&, T&>
+operator_T_amp_pushforward(const ::std::reference_wrapper<T>* w,
+                           const ::std::reference_wrapper<T>* d_w) noexcept {
+  return {w->get(), d_w->get()};
+}
+
 } // namespace class_functions
 
 namespace std {
+
+// std::ref and std::cref
+
+template <class T>
+inline clad::ValueAndPushforward<::std::reference_wrapper<T>,
+                                 ::std::reference_wrapper<T>>
+ref_pushforward(T& t, T& d_t) noexcept {
+  return {::std::ref(t), ::std::ref(d_t)};
+}
+
+template <class T>
+inline clad::ValueAndPushforward<::std::reference_wrapper<const T>,
+                                 ::std::reference_wrapper<const T>>
+cref_pushforward(const T& t, const T& d_t) noexcept {
+  return {::std::cref(t), ::std::cref(d_t)};
+}
 
 // tie and maketuple forward mode
 
--- a/include/clad/Differentiator/STLBuiltins.h
+++ b/include/clad/Differentiator/STLBuiltins.h
@@ -5,12 +5,15 @@
 #include <clad/Differentiator/Array.h>
 #include <clad/Differentiator/BuiltinDerivatives.h>
 #include <clad/Differentiator/FunctionTraits.h>
+#include <cstddef>
 #include <functional>
 #include <initializer_list>
 #include <iterator>
 #include <memory>
+#include <thread>
 #include <tuple>
 #include <type_traits>
+#include <utility>
 #include <vector>
 
 #define elidable_reverse_forw __attribute__((annotate("elidable_reverse_forw")))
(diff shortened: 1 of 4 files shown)


lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -2195,6 +2195,32 @@ BaseForwardModeVisitor::VisitCXXConstructExpr(const CXXConstructExpr* CE) {
     derivedArgs.push_back(argDiff.getExpr_dx());
   }
 
+  // std::thread(f, args...): pass f's pushforward in the d_f slot.
+  if (!clonedArgs.empty() && !derivedArgs.empty()) {
+    QualType recTy = CE->getType().getNonReferenceType().getCanonicalType();
+    if (const auto* RD = recTy->getAsCXXRecordDecl()) {
+      if (RD->getName() == "thread" && RD->isInStdNamespace() &&
```
vgvassilev: Why do we have to special case here?

a-elrawy: Constructing `std::thread` doesn’t call the callable, so `VisitCallExpr ` never nest-diffs it. The callable’s tangent is just `{}`/`nullptr`, not `F_pushforward`, which the custom derivative needs to run AD through the worker. We can generalize this later if you prefer.

lib/Differentiator/CladUtils.cpp
```diff
@@ -882,6 +883,12 @@ namespace clad {
     bool isCopyable(const clang::CXXRecordDecl* RD) {
       if (RD->defaultedCopyConstructorIsDeleted())
         return false;
+      // Also treat inaccessible copy ctors as non-copyable (e.g. libc++
+      // thread).
+      for (const clang::CXXConstructorDecl* Ctor : RD->ctors())
+        if (Ctor->isCopyConstructor() &&
+            (Ctor->isDeleted() || Ctor->getAccess() != clang::AS_public))
```
vgvassilev: **This changes a shared function, and nothing in this PR tests the change.**

`isCopyable` now answers "no" for any class whose copy constructor is
private or protected. That is a reasonable answer. The problem is who else
asks: six places in `ReverseModeVisitor` call `isCopyable`, and they now
take a different path for every class with a private copy constructor.
This PR only adds forward-mode tests, so nothing here checks reverse mode.

Please move this to its own commit with a reverse-mode test.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -2186,15 +2187,78 @@ StmtDiff BaseForwardModeVisitor::VisitBreakStmt(const BreakStmt* stmt) {
...
+namespace {
+const FunctionDecl* findCallOperator(Sema& SemaRef, CXXRecordDecl* RD) {
+  if (!RD)
+    return nullptr;
+  DeclarationName DN =
+      SemaRef.getASTContext().DeclarationNames.getCXXOperatorName(OO_Call);
+  LookupResult R(SemaRef, DN, noLoc, Sema::LookupMemberName);
+  R.suppressDiagnostics();
+  SemaRef.LookupQualifiedName(R, RD);
+  if (!R.isSingleResult())
```
vgvassilev, line 2376: **clad stays silent when it cannot find the thread's function.**

`resolveThreadCallable` returns nothing for a generic lambda, an
overloaded `operator()`, a `std::function`, or a `std::bind` result. When
that happens the code just skips ahead. No derivative is generated and
nothing is reported. The user gets a wrong number back and no reason to
suspect it.

A wrong answer is worse than an error here. Report it with
`DiagnosticsEngine::Report`, pointing at the constructor and naming the
callable that could not be resolved.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -2186,15 +2187,78 @@ StmtDiff BaseForwardModeVisitor::VisitBreakStmt(const BreakStmt* stmt) {
...
+        pushforwardFnRequest.Function = callableFD;
+        pushforwardFnRequest.Mode = GetPushForwardMode();
+        pushforwardFnRequest.BaseFunctionName =
+            utils::ComputeEffectiveFnName(callableFD);
+        pushforwardFnRequest.VerboseDiags = false;
+        if (const auto* MD = dyn_cast<CXXMethodDecl>(callableFD))
+          pushforwardFnRequest.Functor = MD->getParent();
+        FunctionDecl* pushforwardFD =
+            m_Builder.HandleNestedDiffRequest(pushforwardFnRequest);
+        if (pushforwardFD && !isa<CXXMethodDecl>(callableFD))
```
vgvassilev, line 2438: **A derivative is generated and then thrown away.**

`HandleNestedDiffRequest` runs for every callable it resolves, but the
result is only used when the callable is not a class method. For a functor
it is dropped. I think that is on purpose: the functor's derivative is
reached later through `operator_call_pushforward`, so the call is only
here to make clad emit that method.

If so, please say that in a comment. As written it looks like a bug, and
whoever reads it next will either "fix" it or delete it.

vgvassilev (commented):
This adds forward-mode differentiation for `std::thread`: it finds the
function a thread was started with, generates that function's derivative,
and gives the thread a value/derivative pair. The approach looks right.
I would split it up before landing it.

**Scope: this is three changes in one pull request.**
The title says `std::thread`, but about half the diff is
`std::reference_wrapper` -- the `get_pushforward` and
`operator_T_amp_pushforward` overloads in `STLBuiltins.h`, and the
`f_ref`, `f_cref`, `f_implicit`, `f_twice_ref` and `mul_refs` tests. The
`isCopyable` edit is a third. Each is useful and each can be reviewed on
its own. Three commits would be easier to review and easier to bisect.

Three more findings are on the lines they concern.

What I could not check: I read the change rather than building it, so I do
not know whether the generated code compiles. The test checks 20 runtime
results but only 5 lines of generated code, so a future change to how the
thread pair is built would only be noticed if the numbers moved. I also
could not tell whether the `constructor_pushforward` overload taking
`F&& f, ..., F&& d_f` is ever used: `F` is deduced from both arguments, so
they must have the same type, and I found no test that reaches it.

---
These comments come from `ai-vgvassilev`: written by an AI reviewer and posted after @vgvassilev read them.

include/clad/Differentiator/STLBuiltins.h
```diff
@@ -870,10 +873,227 @@ template <typename T>
...
+template <class A0>
+inline clad::ValueAndPushforward<::std::thread, ::std::thread>
+constructor_pushforward(clad::Tag<::std::thread>, void (* /*f*/)(A0), A0 a0,
+                        void (*pf)(A0, A0), A0 d_a0) {
+  return thread_detail::build<
+      clad::ValueAndPushforward<::std::thread, ::std::thread>>(pf, a0, d_a0);
+}
+
+template <class F, class A0,
+          ::std::enable_if_t<::std::is_class<::std::decay_t<F>>::value, int> = 0>
```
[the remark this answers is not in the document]
a-elrawy: Kept is_class::value for C++14; is_class_v fails the -std=c++14 lit runs.

include/clad/Differentiator/STLBuiltins.h
```diff
@@ -870,10 +873,227 @@ template <typename T>
...
+inline clad::ValueAndPushforward<::std::thread, ::std::thread>
+constructor_pushforward(clad::Tag<::std::thread>, F&& f, A0&& a0, F&& /*d_f*/,
+                        A0&& /*d_a0*/) {
+  return thread_detail::build<
+      clad::ValueAndPushforward<::std::thread, ::std::thread>>(
+      ::std::forward<F>(f), ::std::forward<A0>(a0));
+}
+
+template <class F, class A0,
+          ::std::enable_if_t<!::std::is_class<::std::decay_t<F>>::value, int> = 0>
```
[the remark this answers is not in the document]
a-elrawy: Kept is_class::value for C++14; is_class_v fails the -std=c++14 lit runs.

include/clad/Differentiator/STLBuiltins.h
```diff
@@ -870,10 +873,227 @@ template <typename T>
...
+  return thread_detail::build<
+      clad::ValueAndPushforward<::std::thread, ::std::thread>>(
+      [f = ::std::forward<F>(f), d_f = ::std::forward<F>(d_f),
+       a0 = ::std::forward<A0>(a0), d_a0 = ::std::forward<A0>(d_a0)]() mutable {
+        f.operator_call_pushforward(a0, &d_f, d_a0);
+      });
+}
+
+template <class F, class A0,
+          ::std::enable_if_t<!::std::is_class<::std::decay_t<F>>::value, int> = 0>
```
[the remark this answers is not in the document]
a-elrawy: Kept is_class::value for C++14; is_class_v fails the -std=c++14 lit runs.

vgvassilev:
We still need the reverse of this.

a-elrawy:
> We still need the reverse of this.

I've started working on it. Should I make a new PR for the reverse mode or extend the current one?

vgvassilev:
> > We still need the reverse of this.
> 
> I've started working on it. Should I make a new PR for the reverse mode or extend the current one?

I’d prefer to be a single PR.

include/clad/Differentiator/STLBuiltins.h
```diff
@@ -870,10 +873,233 @@ template <typename T>
...
+inline clad::ValueAndPushforward<::std::thread, ::std::thread>
+constructor_pushforward(clad::Tag<::std::thread>, void (* /*f*/)(),
+                        void (*pf)()) {
+  return thread_detail::build<
+      clad::ValueAndPushforward<::std::thread, ::std::thread>>(pf);
+}
+
+template <class F, class DF>
+inline clad::ValueAndPushforward<::std::thread, ::std::thread>
+constructor_pushforward(clad::Tag<::std::thread>, F&& f, DF&& /*d_f*/) {
```
vgvassilev: A callable with no arguments runs the primal only. This overload accepts any class type, ignores d_f, and starts the thread on f. So a functor whose zero-argument operator() mutates a value it holds by reference differentiates to 1 instead of 2, with no diagnostic. A lambda takes the same path, and nothing enforces the 'not yet supported' in the description. The one-argument overload below is right because it calls operator_call_pushforward; add the same for zero arguments, or diagnose it in VisitCXXConstructExpr the way an unresolved callable is.

vgvassilev, line 919: Problem: a callable with no arguments runs the primal only. This overload accepts any class type, ignores d_f, and starts the thread on f. A functor whose zero-argument operator() mutates a value it holds by reference therefore differentiates to 1 where the answer is 2, and nothing is reported. A lambda takes the same path; the description says lambdas are not supported, but nothing rejects them.

How to address it: mirror the one-argument overload below, which is right because it calls operator_call_pushforward. Add a zero-argument overload constrained on is_class that runs `f.operator_call_pushforward(&d_f)` in the worker. If lambdas cannot be handled that way yet, reject them in VisitCXXConstructExpr with the same diagnostic used for an unresolved callable, and add a test for the functor case.

vgvassilev (commented):
Problem: the branch no longer applies. Its middle commit, the isCopyable change with PrivateCopy.C, is already on master as https://github.com/vgvassilev/clad/commit/1a25a38634977b76158d3c04623ce5eecef68ab3 and was adjusted by https://github.com/vgvassilev/clad/commit/628517d0464b38a948653604b8af6a3930e01b9d, so the two collide. Separately, every CI row is red for a reason unrelated to this change: the lit config at the time used execute_external=True, which newer lit rejects; master has fixed that since. CI's own baseline check does show all three new tests failing on master, so they observe the change. The reverse-mode half asked for in the thread, in this same PR, is still missing.

How to address it: rebase onto master and drop the isCopyable commit, keeping the reference_wrapper and std::thread commits. That alone should turn CI green. Then add the reverse-mode support in this PR as agreed.

I did not build or run anything.

include/clad/Differentiator/STLBuiltins.h
```diff
@@ -876,21 +879,526 @@ void lock_pullback(const ::std::weak_ptr<T>* p, ::std::shared_ptr<T> dthis,
...
+  return thread_detail::build<
+      clad::ValueAndPushforward<::std::thread, ::std::thread>>(
+      ::std::forward<F>(f));
+}
+
+template <class A0>
+inline clad::ValueAndPushforward<::std::thread, ::std::thread>
+constructor_pushforward(clad::Tag<::std::thread>, void (* /*f*/)(A0), A0 a0,
```
vgvassilev: Forward mode cannot construct a thread whose worker takes an argument by reference. `A0` is deduced twice here, once from the callable's parameter list and once from the value passed, and the only idiom that reaches a `T&` parameter is `std::thread(w, std::ref(x))` with `void w(double&)`. Then the two deductions give `double&` and `std::reference_wrapper<double>` and this overload drops out. The `F&& f, ..., F&& d_f` overloads cannot take over, because for a free function `d_f` is the generated pushforward and its type differs from the callable's, so `F` conflicts too. Nothing matches, `BuildCustomDerivativeConstructorPFCall` returns null, and forward mode falls back to building `std::thread{w_pushforward, _d_rx}` from the derived arguments, which fails inside `<thread>` with an error pointing at code the user never wrote.

Every worker in `test/ForwardMode/Concurrency.C` takes `std::reference_wrapper<double>` by value, which is why this does not show. `test/Gradient/ConcurrencyThread.C` uses `double&` workers throughout, so the forward-mode version of those same functions would not build. By-value and pointer parameters are fine; it is `T&` that has no path.

Giving the derivative parameters their own template parameters -- `class DF`, `class DA0`, `class DA1` -- instead of reusing `F` and `A0` fixes both halves at once, and would also let the roughly twenty overloads here collapse into far fewer. A forward-mode test with a `double&` worker would hold it.

lib/Differentiator/CladUtils.cpp
```diff
@@ -1015,6 +1016,53 @@ namespace clad {
...
+          SemaRef.getASTContext().DeclarationNames.getCXXOperatorName(OO_Call);
+      LookupResult R(SemaRef, DN, noLoc, Sema::LookupMemberName);
+      R.suppressDiagnostics();
+      SemaRef.LookupQualifiedName(R, RD);
+      if (!R.isSingleResult())
+        return nullptr;
+      return dyn_cast<CXXMethodDecl>(R.getFoundDecl());
+    }
+
+    const FunctionDecl* resolveThreadCallable(Sema& SemaRef, const Expr* E) {
```
vgvassilev, line 1051: `std::function` resolves here and should not. The comment at the call site says `std::function` and `std::bind` cannot be resolved, but `findCallOperator` looks up `operator()` and both libstdc++ and libc++ declare exactly one on `std::function`, so `isSingleResult()` is true and the type-erased `operator()` comes back. clad then nest-diffs a body whose entire content is a call through a stored function pointer, and the user gets a derivative of zero with no diagnostic. `std::bind` really is rejected, because `_Bind` has several `operator()` overloads, so the two cases behave differently despite the comment treating them as one. This is the case the 2026-08-25 review asked to be reported rather than skipped, so it is worth either rejecting `std::function` by name or requiring the resolved `operator()` to have a body clad can see.

The last fallback also fires on `std::thread`'s own move constructor. `std::thread t2(std::move(t1))` arrives here with a `std::thread` argument, finds no `operator()`, and the caller reports "failed to resolve callable of type 'std::thread'". Skipping the resolve when the first argument is already a `std::thread` would avoid rejecting a legal program with a misleading message.

lib/Differentiator/CladUtils.cpp
```diff
@@ -1222,6 +1270,9 @@ namespace clad {
         return !T->getPointeeType().isConstQualified();
       if (const auto* RT = T->getAs<RecordType>()) {
         const RecordDecl* RD = RT->getDecl();
+        // Including reference_wrapper<const T>, whose field is a const T*.
+        if (RD->isInStdNamespace() && RD->getName() == "reference_wrapper")
```
vgvassilev: `isMemoryType` is documented as "true if T allows to edit any memory", and `std::reference_wrapper<const T>` does not -- the comment added here says so and returns true anyway. The predicate is shared. `TBRAnalyzer` uses it to decide whether a callee can write through a parameter, and `hasMemoryTypeParams` uses it from three places in `DiffPlanner` and one in `ReverseModeVisitor`. None of those were updated, and they now take a different path for every function with a `reference_wrapper` parameter.

I could not decide whether that is the point of the change or a side effect. The TBR one looks deliberate and may well be needed -- without it TBR would not know that `read_ref` can write through its argument, and `f_write_through` would be at risk. If so, please say that here, because as written this reads as a local fix for `needsReverseForw`, which this pull request adds for exactly this kind of rule and where the `reference_wrapper` case could otherwise live. Either way the const case wants its own sentence: right now the comment explains why the answer should be no and the code answers yes.

lib/Differentiator/CladUtils.cpp
```diff
@@ -1571,17 +1649,19 @@ namespace clad {
           FnTypes.push_back(utils::GetParameterDerivativeType(S, mode, PVDTy));
       }
 
+      // Type tag before the object prepended below; a constructor has none.
+      if (mode == DiffMode::reverse_mode_forward_pass &&
+          (isa<CXXConversionDecl>(FD) || isa<CXXConstructorDecl>(FD))) {
```
vgvassilev, line 1664: This changes the signature every custom `conversion_operator_reverse_forw` has to have. It was `(clad::Tag<R>, const C* obj, const C* d_obj)` and is now `(const C* obj, clad::Tag<R>, const C* d_obj)`. In-tree only `std::reference_wrapper` has one and it is updated in the same commit, so nothing here breaks. Out of tree the failure is a "no matching function" at a call site in generated code, with nothing to suggest the order moved -- and `KokkosBuiltins.h`, `CladtorchBuiltins.h` and `ThrustBuiltins.h` are the obvious places for one to appear.

If the reorder is needed so the new `reference_wrapper<const T>` overload can be declared, I could not see why the old order could not carry it, since the tag deduces `T` from `const T&` either way. If it is a deliberate tidy-up, it is a convention change and belongs in its own NFC commit with a line in the message, so that someone hitting the error later can find it.

vgvassilev (commented):
I checked this against issue #2114, where clad's OpenMP reverse sweep races because clad parallelises the sweep itself and two threads then write the same `_d_x[i]`. This change does not have that shape, and I think the design is right. Reverse mode runs the worker's primal on a real thread in the forward pass, joins it there through `join_reverse_forw`, and then runs the worker pullback inline on the main thread at the constructor's point in the reverse sweep. Nothing in the reverse sweep runs concurrently, so no two threads ever write the same adjoint. Forward mode does run derivative code on the worker thread, but a tangent is only reachable where its variable is, so a primal that is free of races gives a tangent that is free of races. That is worth a sentence in the commit message, because a reader will assume the opposite and go looking for the race.

Tests that observe the change: `test/Gradient/ConcurrencyThread.C` is the real one. `f_thread_ptr` and `f_thread_ref` expect 6.0000 and `f_thread_functor_ref` expects 2.0000; drop the worker-pullback insertion and all three come back 0 or 1. In forward mode `test/ForwardMode/Concurrency.C` has `f_thread_scale` at 2.0000 and `f_thread_functor_0` at 2.0000, which are the two numbers that moved when the pushforward was corrected. `test/Gradient/Concurrency.C` contains no thread at all -- it observes the two `reference_wrapper` commits, not the thread work. The two `ThreadCallableError.C` files observe the new diagnostics. All five use real substitutions (`%cladclang`, `%filecheck`, `%filecheck_exec`), so they do run.

One point from the 2026-08-25 review is still open: the `constructor_pushforward` overloads shaped `F&& f, ..., F&& d_f` deduce `F` from both arguments and cannot be selected when the two differ. The first comment below says what that costs rather than re-arguing it.

What I could not check: I read the change rather than building it, so I do not know whether the generated code compiles in the cases the tests do not cover. I also could not work out what makes the `isCopyable` half of the new `canTape` guard in `DifferentiateCallArg` necessary. The `isFunctionType` half clearly is -- `f_thread_noop` passes a function lvalue through `std::thread`'s `F&&`, which collapses to a reference to function type and would otherwise be taped -- but I found no test that reaches the non-copyable branch, and skipping the store/restore there is silent rather than diagnosed.

PR #1893, opened 2026-07-15
Enable restore_tracker usage in device kernels

Vedant2005goyal:
This PR enables `clad::restore_tracker` to work correctly inside GPU kernels. The previous implementation relied on STL containers such as `std::vector` and `std::map`, which are not supported in device code, causing compilation failures or runtime crashes when restore_tracker was used within GPU kernels.

### DIFF
--- a/include/clad/Differentiator/RestoreTracker.h
+++ b/include/clad/Differentiator/RestoreTracker.h
@@ -7,6 +7,12 @@
 #include <map>
 #include <utility>
 #include <vector>
+#ifndef Max_Records
+#define Max_Records 64
+#endif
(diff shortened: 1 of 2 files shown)


vgvassilev:
Can we explain in the commit message the idea of that change. Also maybe update the pr description.

Vedant2005goyal:
> Can we explain in the commit message the idea of that change. Also maybe update the pr description.

Done

vgvassilev:
Let’s add a test that did not work with cuda in the past and now passes.

vgvassilev (approved):
Lgtm!

PR #1911, opened 2026-07-20
Use the variable-template type traits in zero_impl. NFC

vgvassilev:
clang-tidy's modernize-type-traits flags the ::value spelling of std::is_trivially_destructible now that clad requires C++17; use the _v form.

The c-array NOLINT named only the cppcoreguidelines alias, so the modernize spelling of the same check still fired on the memcpy staging buffer. Name both.


vgvassilev:
This is wrong -- we can't just use c++17 and our bots just told us so.

PR #1917, opened 2026-07-21
Register a cloned lambda body's nested declarations before remapping.

### LINKED ISSUE #1916: clad::differentiate crashes when a lambda contains a loop
Reproducer:
``` cpp
double f(double x) {
    auto g = [](double t) {
        double s = 0;
        for (int k = 0; k < 3; ++k) s += t;
        return s;
    };
    return g(x);
}
```
The assert added in #1912 catches it,
```Assertion `!Report.UseBeforeDecl && "clad generated a use of a variable before its declaration"' failed.
clad referenced 'k' before its declaration while differentiating 'f_darg0'```

Cloning the lambda body gives the for initialization a fresh k but leaves the condition and increment on the original, and `buildClonedLambda` only remaps top level decls, so the counter is missed.

### PR DESCRIPTION
**Pull request by Fermats-Last-Theorem**
Cloning a lambda body gives the for initialization a fresh k but leaves the condition and increment on the original, because `buildClonedLambda` only remaps top level declarations, and only after the `ReferencesUpdater` has run.

This PR registers the (original, clone) pairs at any depth, before the remap.

Fixes #1916

### DIFF
--- a/test/ForwardMode/Lambdas.C
+++ b/test/ForwardMode/Lambdas.C
@@ -42,6 +42,16 @@ double fn4(double x) {
     return _f(x);
 }
 
+double fn5(double x) {
+    auto _f = [](double _x) {
+        double s = 0;
+        for (int k = 0; k < 3; ++k)
+            s += _x * _x;
+        return s;
+    };
+    return _f(x);
+}
+
 int main() {
     auto fn0_dx = clad::differentiate(fn0, 0);
     printf("Result is = %.2f\n", fn0_dx.execute(7)); // CHECK-EXEC: Result is = 14.00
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/ForwardMode/Lambdas.C [hunk 1]
```diff
@@ -62,4 +72,8 @@ int main() {
     auto fn4_dx = clad::differentiate(fn4, 0);
     printf("Result is = %.2f\n", fn4_dx.execute(7)); // CHECK-EXEC: Result is = 28.00
     printf("Result is = %.2f\n", fn4_dx.execute(-1)); // CHECK-EXEC: Result is = -4.00
+
+    auto fn5_dx = clad::differentiate(fn5, 0);
+    printf("Result is = %.2f\n", fn5_dx.execute(7)); // CHECK-EXEC: Result is = 42.00
+    printf("Result is = %.2f\n", fn5_dx.execute(-1)); // CHECK-EXEC: Result is = -6.00
```
vgvassilev, line 78: Did you verify these results with finite differences?

Fermats-Last-Theorem: Yes, central differences match to `~1e-8`

vgvassilev (approved):
Lgtm!

PR #1926, opened 2026-07-26
Add missing clone  of CXXBindTemporaryExpr in StmtClone

fogsong233:
### Add `CXXBindTemporaryExpr` support in `StmtClone`.

 Clang inserts this implicit AST node when a temporary object has a non-trivial destructor. Since `StmtClone` did not handle it, cloning nested member calls on such temporaries fell through to `VisitStmt` and could eventually crash while updating references.

  ### Possible follow-up

  StmtClone currently implements cloning by manually handling individual AST node kinds. This makes it possible to miss implicit nodes such as CXXBindTemporaryExpr.

  A longer-term solution could be to refactor cloning and declaration-reference replacement around Clang's TreeTransform. It already recursively transforms and rebuilds AST nodes through Sema, including the implicit temporary and cleanup machinery, and could address this class of missing-node failures more systematically.

  The main obstacle is that TreeTransform.h is currently a private, non-installed Clang header under clang/lib/Sema. It also depends on other private headers such as CoroutineStmtBuilder.h and TypeLocBuilder.h. Using it in Clad's standalone build would therefore require vendoring version-matched private headers, requiring a matching Clang source tree, or first making this infrastructure public in Clang.

### DIFF
--- a/include/clad/Differentiator/StmtClone.h
+++ b/include/clad/Differentiator/StmtClone.h
@@ -120,6 +120,7 @@ namespace utils {
     DECLARE_CLONE_FN(CXXThrowExpr)
     DECLARE_CLONE_FN(CXXConstructExpr)
     DECLARE_CLONE_FN(CXXTemporaryObjectExpr)
+    DECLARE_CLONE_FN(CXXBindTemporaryExpr)
     DECLARE_CLONE_FN(MaterializeTemporaryExpr)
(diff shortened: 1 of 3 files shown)


vgvassilev (approved):
Lgtm!

vgvassilev (approved):
LGTM!

PR #1929, opened 2026-07-28
Add support for SourceLocExpr

Vedant2005goyal:
This PR extends Clad's AST visitor to support SourceLocExpr, enabling differentiation of code containing assert statements and other source-location builtins for both forward and reverse mode.


vgvassilev:
We need a test.

Vedant2005goyal:
> We need a test.

Done!

vgvassilev (approved):
LGTM!

PR #1932, opened 2026-07-29
Zero-initialize non-independent parameter adjoints and introduce `clad::zero_like`

fogsong233:
While investigating the following case:

```cpp
double replace_and_square(std::vector<double> values, double x) {
  values[0] = x;
  return values[0] * values[0];
}

auto gradient = clad::gradient(replace_and_square, "x");
```

I found that even though `values` is not selected as an independent parameter, Clad still needs an internal `_d_values` to propagate the derivative through the assignment.

For `values = {10.0}` and `x = 3.0`, the expected derivative is `6.0`. However, `_d_values` was initialized by copying the primal value and was never cleared, so its initial value leaked into the reverse pass and produced `16.0`.

This means that every internal adjoint created by Clad, including adjoints for non-independent parameters, must represent a valid zero tangent before the reverse pass starts.

While fixing this, I found that the current API does not have enough expressive power to construct such an adjoint correctly for every type.

For a record parameter of type `T`, the existing logic is roughly:

```cpp
// T has direct pointer or reference fields:
T _d_param = T{};

// Otherwise, if T is copyable:
T _d_param(param);
clad::zero_init(_d_param);
```

Neither option is generally correct.

For example, libc++ implements `std::vector` using direct pointer fields. The field-based check therefore selects:

```cpp
std::vector<double> _d_values{};
```

but this loses the vector’s runtime structure: `_d_values` is empty while `values` may contain several elements.

At the same time, simply ignoring pointer fields and always copying is also unsafe. A user-defined type may contain an internal pointer referring to its own storage, and its copy constructor may leave the copied pointer referring to the primal object.

In other words, field layout alone cannot tell us whether copying a type creates a valid, independently owned adjoint.

`clad::zero_init` cannot fully solve this either. Its current purpose is to reset an already existing adjoint while preserving its structure. It is also used for arrays, loop resets, and user-provided specializations. Changing it into an adjoint-construction operation would change its existing semantics and break backward compatibility.

This PR therefore introduces a separate customization point:

```cpp
auto adjoint = clad::zero_like(primal);
```

The intended contract of `zero_like` is:

- create a new adjoint with independent storage;
- preserve runtime structure relevant to differentiation, such as size, shape, allocator, device, and internal invariants;
- initialize all differentiable values to zero;
- never modify or alias mutable storage from the primal.

Types can provide their own overload:

```cpp
namespace clad {

MyType zero_like(const MyType& value) {
  // Construct an independently owned zero adjoint.
}

} // namespace clad
```

For `std::vector`, the implementation copies the vector to obtain independent storage and preserve its size, then applies `zero_init` to its elements:

```cpp
template <class T, class Allocator>
std::vector<T, Allocator>
zero_like(const std::vector<T, Allocator>& value) {
  std::vector<T, Allocator> result(value);
  zero_init(result);
  return result;
}
```

A Tensor backend can similarly implement this by delegating to an operation such as `at::zeros_like`.

When Clad creates an internal adjoint from an existing primal value, it now tries a viable `clad::zero_like` overload first. If no overload is available, it preserves the previous copy/value-initialization logic as a backward-compatible fallback.

This also removes the need for a `std::vector`-specific exception in `ReverseModeVisitor`: knowledge about how to construct a vector adjoint now lives in `STLBuiltins.h`, where it belongs.

Finally, this API provides an important building block for improving `clad::gradient`. Currently, users must allocate zero adjoints themselves and pass `_d_xxx` pointers to `execute`. With `zero_like`, Clad can eventually construct those terminal adjoints from the primal arguments and return them directly:

```cpp
auto gradients = gradient.execute(args...);
```

This PR does not change that interface yet, but it provides the missing construction operation needed to implement it correctly for vectors, tensors, and user-defined structured types.

### DIFF
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -210,8 +210,10 @@ CUDA_HOST_DEVICE auto back(TapeType& of) -> decltype(of.back()) {
   return of.back();
 }
 
-  /// The purpose of this function is to initialize adjoints
-  /// (or all of its iteratable elements) with 0.
+  /// Reset values in an already-constructed adjoint (or its iterable elements)
+  /// to zero in place. This is the primitive used by the default zero_like
+  /// implementation. Provide an overload when the default recursive or
+  /// byte-wise zeroing would not preserve a type's required structure.
   namespace zero_init_detail {
   template <class T> struct iterator_traits : std::iterator_traits<T> {};
   template <> struct iterator_traits<void*> {};
@@ -281,13 +283,15 @@ CUDA_HOST_DEVICE auto back(TapeType& of) -> decltype(of.back()) {
 
   template <class T> CUDA_HOST_DEVICE void zero_init(T& t) { zero_impl(t); }
 
-  /// Create a zero adjoint with the same relevant structure as \p value.
+  /// Construct a zero adjoint with the same relevant structure as \p value.
   ///
-  /// Record and container types should provide an overload when constructing
-  /// their zero adjoint requires preserving runtime structure such as shape,
-  /// size, allocator, or device information. The reverse-mode visitor prefers
-  /// this customization point when it must create an internal adjoint from an
-  /// existing primal value.
+  /// Conceptually, zero_like(value) returns a value structurally like value
+  /// with zero_init applied. This is the adjoint-construction extension point;
+  /// provide an overload when copy-then-zero is incorrect, for example for
+  /// owning or reference-counted storage, device memory, or types whose shape,
+  /// allocator, or other runtime metadata needs special handling. The
+  /// reverse-mode visitor prefers this customization point when constructing
+  /// an internal adjoint from an existing primal value.
   // Keep this header compatible with C++14; some Clad tests and clients include
   // it in that language mode.
   // NOLINTBEGIN(modernize-type-traits)
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -31,6 +31,9 @@
 #endif
 #include <type_traits>
 #include <utility>
+// Keep std::valarray's non-member begin/end overloads visible when is_range is
+// defined below.
+#include <valarray>
 #ifndef __CUDACC__
 #include <mutex>
 #endif
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -31,6 +31,9 @@
 #endif
 #include <type_traits>
 #include <utility>
+// Keep std::valarray's non-member begin/end overloads visible when is_range is
+// defined below.
+#include <valarray> // NOLINT(misc-include-cleaner)
 #ifndef __CUDACC__
 #include <mutex>
 #endif
@@ -207,25 +210,27 @@ CUDA_HOST_DEVICE auto back(TapeType& of) -> decltype(of.back()) {
   return of.back();
 }
 
-  /// The purpose of this function is to initialize adjoints
-  /// (or all of its iteratable elements) with 0.
-  namespace zero_init_detail {
-  template <class T> struct iterator_traits : std::iterator_traits<T> {};
-  template <> struct iterator_traits<void*> {};
-  template <> struct iterator_traits<const void*> {};
+/// Reset values in an already-constructed adjoint (or its iterable elements)
+/// to zero in place. This is the primitive used by the default zero_like
+/// implementation. Provide an overload when the default recursive or
+/// byte-wise zeroing would not preserve a type's required structure.
+namespace zero_init_detail {
+template <class T> struct iterator_traits : std::iterator_traits<T> {};
+template <> struct iterator_traits<void*> {};
+template <> struct iterator_traits<const void*> {};
 
-  template <class T, class It>
-  std::integral_constant<
-      bool, !std::is_same<typename std::remove_cv<T>::type,
-                          typename iterator_traits<It>::value_type>::value>
-  is_range_check(It first, It last);
+template <class T, class It>
+std::integral_constant<
+    bool, !std::is_same<typename std::remove_cv<T>::type,
+                        typename iterator_traits<It>::value_type>::value>
+is_range_check(It first, It last);
 
-  template <class T>
-  decltype(is_range_check<T>(std::begin(std::declval<const T&>()),
-                             std::end(std::declval<const T&>())))
-  is_range(int);
-  template <class T> std::false_type is_range(...);
-  } // namespace zero_init_detail
+template <class T>
+decltype(is_range_check<T>(std::begin(std::declval<const T&>()),
+                           std::end(std::declval<const T&>())))
+is_range(int);
+template <class T> std::false_type is_range(...);
+} // namespace zero_init_detail
 
   template <class T>
   struct is_range : decltype(zero_init_detail::is_range<T>(0)) {};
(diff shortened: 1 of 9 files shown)


lib/Differentiator/VisitorBase.cpp
```diff
@@ -1243,6 +1245,54 @@ namespace clad {
...
+        OverloadCandidateSet CandidateSet(noLoc,
+                                          OverloadCandidateSet::CSK_Normal);
+        Scope* S = m_Sema.getScopeForContext(m_Sema.CurContext);
+        auto* ULE = cast<UnresolvedLookupExpr>(Find.Expression);
+        m_Sema.buildOverloadedCallSet(S, UnresolvedLookup, ULE, ARargs, noLoc,
+                                      &CandidateSet, &Result);
+        OverloadCandidateSet::iterator Best = nullptr;
+        OverloadingResult OverloadResult =
+            CandidateSet.BestViableFunction(m_Sema, noLoc, Best);
+        if (OverloadResult != 0U)
```
vgvassilev:
```suggestion
        if (OverloadResult != OR_Success)
```

lib/Differentiator/VisitorBase.cpp
```diff
@@ -1243,6 +1245,54 @@ namespace clad {
...
+          return nullptr;
+      }
```
vgvassilev: Can we avoid the hand-roled overload resolution here? Also worth a one-line assertion that, with suppressDiagnostics(), the failure path emits no stray notes the whole point is a silent optional fallback. If clad already has an overload-probe helper for custom derivatives, reusing it avoids a second copy of this logic.

include/clad/Differentiator/STLBuiltins.h
```diff
@@ -11,12 +11,15 @@
...
 #include <tuple>
 #include <type_traits>
+#include <valarray>
 #include <vector>
 
 #define elidable_reverse_forw __attribute__((annotate("elidable_reverse_forw")))
 
 namespace clad {
 
+template <class T> struct is_range;
```
vgvassilev: Do we really need this forward declaration -- seems fragile.

include/clad/Differentiator/Differentiator.h
```diff
@@ -278,6 +278,34 @@ CUDA_HOST_DEVICE auto back(TapeType& of) -> decltype(of.back()) {
...
+                                      int> = 0>
+  CUDA_HOST_DEVICE T zero_like(const T&) {
+    return T();
+  }
+
+  /// Default zero_like implementation for copyable ranges.
+  template <class T, std::enable_if_t<is_range<T>::value &&
+                                          std::is_copy_constructible<T>::value,
+                                      int> = 0>
+  T zero_like(const T& value) {
```
vgvassilev, line 357: This copies the whole primal only to overwrite it with zeros — for an N-element container that's N element copy-ctors + two data passes. For sized, size-constructible ranges you can build T(value.size()) and skip the copy while keeping zero_init for correctness:

```patch
From 4949edc8882e8864a0466a53390a65674907a580 Mon Sep 17 00:00:00 2001
From: Vassil Vassilev <v.g.vassilev@gmail.com>
Date: Wed, 29 Jul 2026 16:51:17 +0000
Subject: [PATCH] [Differentiator] Build sized zero_like adjoints without a
 primal copy.

The default zero_like for a range copies the primal and then zeroes the
copy, so building an adjoint for an N-element container touches the data
twice and runs N element copy-constructors whose results are immediately
overwritten.

For a range that exposes size() and a size-constructor -- vector, deque,
list, valarray -- construct N value-initialized elements directly and
zero_init them; this drops the primal copy while zero_init keeps element
types that do not value-initialize to zero correct. Ranges without a
size-constructor (forward_list, std::array, the associative containers)
keep the copy-then-zero fallback.

Follows up on the zero_like customization point from #1932.
---
 include/clad/Differentiator/Differentiator.h | 42 +++++++++++--
 test/Misc/ZeroLikeSized.C                    | 63 ++++++++++++++++++++
 2 files changed, 101 insertions(+), 4 deletions(-)
 create mode 100644 test/Misc/ZeroLikeSized.C

diff --git a/include/clad/Differentiator/Differentiator.h b/include/clad/Differentiator/Differentiator.h
index 80f74b37..78c1584f 100644
--- a/include/clad/Differentiator/Differentiator.h
+++ b/include/clad/Differentiator/Differentiator.h
@@ -295,10 +295,44 @@ CUDA_HOST_DEVICE auto back(TapeType& of) -> decltype(of.back()) {
     return T();
   }
 
-  /// Default zero_like implementation for copyable ranges.
-  template <class T, std::enable_if_t<is_range<T>::value &&
-                                          std::is_copy_constructible<T>::value,
-                                      int> = 0>
+  // Detect a range that has size() and a T(size_type) constructor, so a zero
+  // adjoint can be built from just the size instead of copying the primal.
+  // C++14-safe overload dispatch, mirroring is_range above.
+  namespace zero_like_detail {
+  template <class T>
+  auto has_sized_ctor(int)
+      -> decltype(static_cast<void>(std::declval<const T&>().size()),
+                  static_cast<void>(T(std::declval<typename T::size_type>())),
+                  std::true_type{});
+  template <class T> std::false_type has_sized_ctor(...);
+  } // namespace zero_like_detail
+
+  template <class T>
+  struct has_sized_ctor : decltype(zero_like_detail::has_sized_ctor<T>(0)) {};
+
+  /// Default zero_like for a sized, size-constructible range (vector, deque,
+  /// list, valarray, ...): construct `size` value-initialized elements -- no
+  /// copy of the primal's values -- then zero_init for correctness on element
+  /// types that do not value-initialize to zero.
+  template <class T,
+            std::enable_if_t<is_range<T>::value &&
+                                 std::is_copy_constructible<T>::value &&
+                                 has_sized_ctor<T>::value,
+                             int> = 0>
+  T zero_like(const T& value) {
+    T result(value.size());
+    zero_init(result);
+    return result;
+  }
+
+  /// Fallback zero_like for copyable ranges without a size-constructor
+  /// (forward_list, std::array, associative containers, ...): copy the
+  /// structure, then zero it.
+  template <class T,
+            std::enable_if_t<is_range<T>::value &&
+                                 std::is_copy_constructible<T>::value &&
+                                 !has_sized_ctor<T>::value,
+                             int> = 0>
   T zero_like(const T& value) {
     T result(value);
     zero_init(result);
diff --git a/test/Misc/ZeroLikeSized.C b/test/Misc/ZeroLikeSized.C
new file mode 100644
index 00000000..30f706bc
--- /dev/null
+++ b/test/Misc/ZeroLikeSized.C
@@ -0,0 +1,63 @@
+// RUN: %cladclang %s -I%S/../../include -o %t
+// Lock in the C++14 compatibility of the zero_like SFINAE.
+// RUN: %cladclang -std=c++14 %s -I%S/../../include -o %t14
+// RUN: %t14 | %filecheck_exec %s
+
+// A sized, size-constructible range (std::vector) builds its zero adjoint from
+// the size alone -- no copy of the primal's values -- while ranges without a
+// size-constructor (std::forward_list, std::array) fall back to copy-then-zero.
+// All must still yield an all-zero, same-shape result.
+
+#include "clad/Differentiator/STLBuiltins.h"
+
+#include <array>
+#include <cstdio>
+#include <forward_list>
+#include <vector>
+
+// Counts copies and does NOT value-initialize to zero: proves the sized path
+// avoids the primal copy yet still zeroes via zero_init.
+struct Counted {
+  double v = 7;
+  static long copies;
+  Counted() = default;
+  Counted(const Counted& o) : v(o.v) { ++copies; }
+  Counted& operator=(const Counted&) = default;
+};
+long Counted::copies = 0;
+
+namespace clad {
+void zero_init(Counted& c) { c.v = 0; }
+} // namespace clad
+
+template <class R> bool all_zero(const R& r) {
+  for (const auto& e : r)
+    if (e.v != 0)
+      return false;
+  return true;
+}
+
+int main() {
+  std::vector<Counted> vec(1000);
+  for (auto& c : vec)
+    c.v = 3.14;
+  Counted::copies = 0;
+  auto dvec = clad::zero_like(vec); // sized path: constructs, never copies
+  std::printf("%d %ld\n", (int)(all_zero(dvec) && dvec.size() == vec.size()),
+              Counted::copies);
+  // CHECK-EXEC: 1 0
+
+  std::forward_list<Counted> fl(3); // no size() -> copy fallback
+  for (auto& c : fl)
+    c.v = 2.0;
+  std::printf("%d\n", (int)all_zero(clad::zero_like(fl)));
+  // CHECK-EXEC: 1
+
+  std::array<Counted, 4> arr{}; // no size-ctor -> copy fallback
+  for (auto& c : arr)
+    c.v = 5.0;
+  std::printf("%d\n", (int)all_zero(clad::zero_like(arr)));
+  // CHECK-EXEC: 1
+}
-- 
2.39.5 (Apple Git-154)
```

include/clad/Differentiator/Differentiator.h
```diff
@@ -278,6 +278,34 @@ CUDA_HOST_DEVICE auto back(TapeType& of) -> decltype(of.back()) {
...
+  /// size, allocator, or device information. The reverse-mode visitor prefers
+  /// this customization point when it must create an internal adjoint from an
+  /// existing primal value.
+  // Keep this header compatible with C++14; some Clad tests and clients include
+  // it in that language mode.
+  // NOLINTBEGIN(modernize-type-traits)
+  template <class T, std::enable_if_t<std::is_arithmetic<T>::value ||
+                                          std::is_enum<T>::value,
+                                      int> = 0>
+  CUDA_HOST_DEVICE T zero_like(const T&) {
```
vgvassilev, line 301: Two customization points now coexist (zero_init in-place vs zero_like return-new) and the boundary is undocumented. Add to the doc-comment + docs/: "zero_like(x) returns a value structurally like x with zero_init applied; override it when copy-then-zero is wrong — owning storage, refcounts, device memory." Consider documenting zero_like as the adjoint-construction extension point and zero_init as the primitive it builds on.

We can add the missing section on zero_init, too.

test/Gradient/STLNonIndependentAdjoint.C
```diff
@@ -0,0 +1,54 @@
+// RUN: %cladclang %s -I%S/../../include -o %t 2>&1 | %filecheck %s
```
vgvassilev, line 1: zero_like's SFINAE is deliberately C++14-safe (the NOLINT(modernize-type-traits)), but these run one mode. Add -std=c++14 and -std=c++20 RUN lines so a future _v-trait/if constexpr creep can't silently regress C++14 clients.

Same holds for the other test.

vgvassilev (commented):
This actually looks very good!

include/clad/Differentiator/Differentiator.h
fogsong233: I am concerned that this optimization is not generally correct. For example, for an `std::vector<std::vector<double>>` with an `m x n` (or ragged) shape, `T(value.size())` preserves only the outer size. It constructs `m` empty inner vectors, and `zero_init` only zeros existing elements, so it cannot recover the inner sizes.

This particular case could be addressed by recursively applying the same construction to each element. In the longer term, however, that would still not make size-based construction safe as a generic rule. A size is not a complete structural descriptor: stateful allocators or memory resources, comparators, hash functions, key-equality objects, and other runtime metadata may be lost.

For example, although Clad does not appear to support differentiating `std::unordered_map` today, it is a concrete false positive for `has_sized_ctor`. It has an `unordered_map(size_type bucket_count)` constructor, so `T(value.size())` constructs an empty container whose bucket count merely happens to equal the primal size. It drops all keys and values and resets the hash, equality, and allocator state.

Therefore, I think recursively reconstructing each element from the corresponding primal value can serve as the default `zero_like` strategy for sized ranges. This would preserve nested sizes without copying primal values. Types whose relevant structure cannot be fully reconstructed this way, such as those carrying stateful allocators or other runtime metadata, should provide type-specific `zero_like` overloads or explicitly opt into a more appropriate construction policy.

vgvassilev: That’s a fair point. However, I am really worried about the performance hit we are going to take with this feature. What are our alternative optimization options?

vgvassilev:
Can we make progress on that and add the missing tests to fix coverage?

vgvassilev:
Can you squash into a single commit with an appropriate commit message?

vgvassilev (approved):
Lgtm!

PR #1938, opened 2026-08-02
Fix uninitialized record assignment in reverse mode

### LINKED ISSUE #1283: Gradient: Fix missing struct initialization to structs declared inside a loop but not originally initialized
Example:
```
#include <stdio.h>


typedef struct {
    double val;
} Struct;

void fn1(double a) {
    for (int i=0;i<1;i++){
        Struct s;
    }
}


int main(){
    auto grad = clad::gradient(fn1);
    return 0;
}
```

Derivative:
```
void fn1_grad(double a, double *_d_a) {
    int _d_i = 0;
    int i = 0;
    clad::tape<Struct> _t1 = {};
    Struct _d_s = {0.};
    Struct s = {0.};
    unsigned long _t0 = 0UL;
    for (i = 0; ; i++) {
        {
            if (!(i < 1))
                break;
        }
        _t0++;
        clad::push(_t1, std::move(s)) , s = ; // error
    }
    for (;; _t0--) {
        {
            if (!_t0)
                break;
        }
        {
            _d_s = {0.};
            s = clad::pop(_t1);
        }
    }
}
```

After fix:
```
void fn1_grad(double a, double *_d_a) {
    int _d_i = 0;
    int i = 0;
    clad::tape<Struct> _t1 = {};
    Struct _d_s = {0.};
    Struct s = {0.};
    unsigned long _t0 = 0UL;
    for (i = 0; ; i++) {
        {
            if (!(i < 1))
                break;
        }
        _t0++;
        clad::push(_t1, std::move(s)) , s = {0.}; // fix
    }
    for (;; _t0--) {
        {
            if (!_t0)
                break;
        }
        {
            _d_s = {0.};
            s = clad::pop(_t1);
        }
    }
}
````

### PR DESCRIPTION
**Pull request by Elvand-Lie**
## Summary

- replace declaration-only empty constructor expressions with Clad’s existing zero initializer before building promoted assignments
- preserve explicit list initialization and constructor calls with arguments
- add regression coverage for record declarations inside loops

## Testing

- verified the issue #1283 reproducer fails on `origin/master` and passes with this change
- manually A/B tested implicit aggregates, default member initializers, explicitly defaulted records, explicit list initialization, and user-defined constructors
- ran the full lit suite: 174 passed, 1 expected platform failure, 25 unsupported, and 2 failures independently reproduced on `origin/master`
- ran `DifferentiatorHTests` and `MiscTests`: 5/5 passed
- ran `git-clang-format-18`: no changes
- ran `clang-tidy-18` on the changed production-code range: no diagnostics
- ran `git diff --check`

Fixes: #1283

### DIFF
--- a/test/Regressions/issue-1283.cpp
+++ b/test/Regressions/issue-1283.cpp
@@ -1,0 +1,41 @@
+// RUN: %cladclang %s -I%S/../../include -o %t 2>&1 | %filecheck %s
+// RUN: %t | %filecheck_exec %s
+
+#include "clad/Differentiator/Differentiator.h"
+
+extern "C" int printf(const char*, ...);
+
+struct Struct {
+  double val;
+};
+
+double fn(double a) {
+  double result = 0;
+  for (int i = 0; i < 3; ++i) {
+    Struct s;
+    s.val = a * (i + 1);
+    result += s.val;
+  }
+  return result;
+}
+
+int main() {
+  auto grad = clad::gradient(fn);
+  double d_a = 0;
+  grad.execute(3, &d_a);
+  printf("%.2f\n", d_a); // CHECK-EXEC: 6.00
+}
+
+// CHECK: void fn_grad(double a, double *_d_a) {
+// CHECK-NEXT:     int _d_i = 0;
+// CHECK-NEXT:     int i = 0;
+// CHECK-NEXT:     Struct _d_s = {0.};
+// CHECK-NEXT:     Struct s = {0.};
+// CHECK-NEXT:     double _d_result = 0.;
+// CHECK-NEXT:     double result = 0;
+// CHECK-NEXT:     unsigned {{int|long|long long}} _t0 = 0;
+// CHECK-NEXT:     for (i = 0; i < 3; ++i) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/Regressions/issue-1283.cpp [hunk 1]
```diff
@@ -0,0 +1,21 @@
...
+  for (int i = 0; i < 1; ++i) {
+    Struct s;
+  }
+}
+
+int main() {
+  auto grad = clad::gradient(fn);
+}
+
+// CHECK-LABEL: void fn_grad
```
vgvassilev: Why not using CHECK: and CHECK-NEXT? We also need to check the execution results.

Elvand-Lie: Updated in a9fe9a55. The regression now uses `CHECK`/`CHECK-NEXT` for the generated loop, builds and runs the test executable, and checks the derivative result (`6.00`) with `CHECK-EXEC`. I also verified that the revised lit test passes with this patch and fails against current `origin/master` at the original `s = ;` line.

vgvassilev (approved):
LGTM!

PR #1939, opened 2026-08-04
Clone CXXStdInitializerListExpr nodes

### LINKED ISSUE #1931: Regression: crash when reverse-differentiating a nested function that has a `#pragma clad checkpoint loop`
Clad master crashes when reverse-differentiating a function that has a `#pragma clad checkpoint loop` for-loop calling a templated helper with local array arguments (e.g. `multiVarGaussian(2, t1, t2, xlArr)`).

This regression makes Clad `master` currently unusable for RooFit.

Reproducer as a standalone ROOT macro:
```C++
#include <TInterpreter.h>
void min2()
{
   const char *helpers = R"CODE(
#include <cmath>
namespace repro {
template <typename XArray, typename MuArray, typename CovArray>
double multiVarGaussian(int n, XArray x, MuArray mu, CovArray covI)
{
   double result = 0.0;
   for (int i = 0; i < n; ++i)
      for (int j = 0; j < n; ++j)
         result += (x[i] - mu[i]) * covI[i * n + j] * (x[j] - mu[j]);
   return std::exp(-0.5 * result);
}
}
)CODE";
   gInterpreter->Declare(helpers);

   const char *funcCode = R"CODE(
double roo_codegen_1(double* params, double const* obs, double const* xlArr) {
  double t2[]{params[0],params[1]};
  double sum = 0.0;
  #pragma clad checkpoint loop
  for(int loopIdx1 = 0; loopIdx1 < obs[1]; loopIdx1++) {
    double t1[]{obs[static_cast<int>(obs[0]) + loopIdx1], obs[static_cast<int>(obs[12]) + loopIdx1]};
    const double t3 = repro::multiVarGaussian(2, t1, t2, xlArr + 0);
    sum += t3;
  }
  return sum;
}
)CODE";
   if(!gInterpreter->Declare(funcCode)){ printf("declare failed\n"); return; }
   printf("function declared OK\n");
   gInterpreter->Declare("#include <Math/CladDerivator.h>\n");
   const char *request = "#pragma clad ON\nvoid req(){ clad::gradient(roo_codegen_1, \"params\"); }\n#pragma clad OFF";
   bool ok = gInterpreter->Declare(request);
   printf("SURVIVED ok=%d\n",(int)ok);
}
```

Fermats-Last-Theorem:
Brace initialized local arrays produce a CXXStdInitializerListExpr node that StmtClone had no visitor for and caused a crash. This PR adds the clone handler.

Fixes #1931

### DIFF
--- a/include/clad/Differentiator/StmtClone.h
+++ b/include/clad/Differentiator/StmtClone.h
@@ -122,6 +122,7 @@ namespace utils {
     DECLARE_CLONE_FN(CXXConstructExpr)
     DECLARE_CLONE_FN(CXXTemporaryObjectExpr)
     DECLARE_CLONE_FN(CXXBindTemporaryExpr)
+    DECLARE_CLONE_FN(CXXStdInitializerListExpr)
     DECLARE_CLONE_FN(MaterializeTemporaryExpr)
     DECLARE_CLONE_FN(PseudoObjectExpr)
     DECLARE_CLONE_FN(OpaqueValueExpr)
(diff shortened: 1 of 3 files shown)


vgvassilev (approved):
Lgtm!

Issue #1940, opened 2026-08-05
Regression: error with referenced 'fact' before its declaration


guitargeek:
A new warning popped up in ROOT unit tests when using Clad `master` with the ROOT unit tests, leading also to wrong results in some cases.

Reproducer as a ROOT macro:

```c++
/// "clad referenced 'y' before its declaration ...; this is a clad bug"
///
/// Because of the early `return`, clad wraps the reverse sweep in a lambda
/// `_rev0` so it can be called from the early-return path.  The lambda is
/// emitted *above* the declarations of the primals it reads:
///
///     auto _rev0 = [&] { y = _t0; ... _d_i += y * _r_d0; ... };
///     { _cond0 = k == 0; if (_cond0) { _rev0(); return; } }
///     int n = k;            // <-- n, y, i, _t0 only declared here
///     double y = n + x;
///     double i = n;
///     double _t0 = y;
///     y *= i;
///
/// The gradient then comes out as 0 instead of 3, so this warning is not
/// benign.  Dropping the early return (see `good` below) makes clad hoist the
/// declarations above `_rev0` and both the warning and the wrong value go away.
///
/// Found via RooFit::Detail::MathFuncs::binomial() (ROOT's inlined
/// TMath::Binomial), where the warning names 'fact' and appears while
/// generating `binomial_pullback`.
///
/// Run with:  root -b -q -l cladIssueRefBeforeDecl.C
/// clad master @ 80f6c90f (v2.4-60-g80f6c90f)

#include <Math/CladDerivator.h>
#include <cstdio>

double bad(double x, int k)
{
   if (k == 0)
      return 1;
   int n = k;
   double y = n + x;
   double i = n;
   y *= i;
   return y;
}

double good(double x, int k)
{
   int n = k;
   double y = n + x;
   double i = n;
   y *= i;
   return y;
}

void cladIssueRefBeforeDecl()
{
   double d = 0.;

   // y = (n + x) * n with n = k = 3, so dy/dx = 3
   clad::gradient(bad, "x").execute(2., 3, &d);
   std::printf("with early return   :  %g   (expected 3)\n", d);

   d = 0.;
   clad::gradient(good, "x").execute(2., 3, &d);
   std::printf("without early return:  %g   (expected 3)\n", d);

   clad::gradient(bad, "x").dump();
   clad::gradient(good, "x").dump();
}
```
The output is:
```txt
Processing cladIssueRefBeforeDecl.C...
[...]

PR #1928, opened 2026-07-27
Fix dangling tracker references in early-return recursive functions

Vedant2005goyal:
This PR fixes a segmentation fault encountered while differentiating recursive functions with early returns (e.g., backtracking type functions). When differentiating such functions Clad wraps the reverse sweep in a lambda ```(auto _rev0 = [&])```. Previously, ```clad::restore_tracker``` declarations and ValueAndAdjoint temporaries generated for operator[] calls were emitted inside nested forward-sweep blocks and it leads to null pointer dereferencing, resulting in a segmentation fault.

PR #1943, opened 2026-08-06
Zero-initialize hoisted variables in GlobalStoreAndRef and StoreAndRef

Vedant2005goyal:
When early returns are present, variables assigned inside nested scopes  are hoisted to the top level so they can be safely captured by the lambda. Previously, these variables were left uninitialized, which could lead to capturing garbage memory if an early return triggered before their assignment ultimately leading to crash
.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -4231,7 +4231,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       // this global there (a branch _cond flag) asks for zero-init so the
       // unreached case reads false -- the same zero-state clad already relies
       // on for adjoints and loop counters.
-      if (zeroInit)
+      if (zeroInit || m_DiffReq.hasEarlyReturns())
         SetDeclInit(VD, getZeroInit(Type));
       addToBlock(decl, m_Globals);
       // Use a fresh ref for the store-back LHS; Ref is returned to the caller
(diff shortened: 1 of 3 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4288,6 +4288,8 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       bool isFnScope = getCurrentScope()->isFunctionScope() ||
                        m_DiffReq.Mode == DiffMode::reverse_mode_forward_pass;
       if (isFnScope) {
+        if (m_DiffReq.hasEarlyReturns())
+          SetDeclInit(VD, getZeroInit(Type));
```
vgvassilev: Can we cover this with a test?

Vedant2005goyal: Wouldn’t it be better to add a small test function to the existing test in PR #1928 instead of creating a separate test for this?

vgvassilev: Yes, I agree.

vgvassilev:
Can you check if this fixes https://github.com/vgvassilev/clad/issues/1940?

Vedant2005goyal:
> Can you check if this fixes #1940?

No the issue still persists

lib/Differentiator/ReverseModeVisitor.cpp
Vedant2005goyal: Great! I’ll add the function after that PR is merged.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4288,9 +4288,13 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
```
vgvassilev: This seems still uncovered. Can we do that as part of this PR?

Vedant2005goyal: Ok i have added the patch to increase coverage

vgvassilev (approved):
LGTM!

vgvassilev:
Can you rebase this PR?

vgvassilev:
Can you rebase again?

PR #1944, opened 2026-08-06
Discard delayed calls after a repl error or undo on clang < 22

aaronj0:
in incremental mode a parse error or %undo discards the declarations of the current ptu, but clad's consumer has already queued delayed calls over them; replaying the queue to the multiplexer hands codegen a broken ast and crashes when run with CppInterOp + clang-repl (https://github.com/compiler-research/CppInterOp/pull/830). fast-forward delayed calls index past the queued calls when diagnostics report an error so they are never sent.

clang >= 22 is not affected (this was fixed by llvm/llvm-project#182044), so the workaround is gated on CLANG_VERSION_MAJOR < 22. Tested across LLVM 20, 21 and 22, all CppInterOp unittests pass.

### DIFF
--- a/test/Misc/Repl.cpp
+++ b/test/Misc/Repl.cpp
@@ -1,6 +1,8 @@
 // REQUIRES: clang-repl
-// UNSUPPORTED: clang-13, clang-14, clang-15, clang-17
-// RUN: cat %s | %clang-repl -Xcc -fplugin=%cladlib -Xcc -I%S/../../include | FileCheck %s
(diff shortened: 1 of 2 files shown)


aaronj0:
@vgvassilev this PR is ready for review.

vgvassilev:
> @vgvassilev this PR is ready for review.

Out cuda builders are offline. I need to enable them first.

aaronj0:
> Out cuda builders are offline. I need to enable them first.

Thanks, no rush

vgvassilev (approved):
LGTM!

PR #1945, opened 2026-08-07
Find LLD before Clang to support emscripten-built LLVM

aaronj0:
Emscripten builds of the interpreter link `lldWasm/lldCommon`, so the generated ClangTargets.cmake references LLD targets and `find_package(Clang)` fails unless the LLD package is found first. Find LLD before Clang, honoring LLD_DIR when set and hinting off LLVM_DIR otherwise. The lookup is QUIET, so native builds without LLD are unaffected. This allows cross-compiling clad against an emscripten-built LLVM/Clang for wasm.


vgvassilev (approved):
LGTM!

PR #1949, opened 2026-08-09
Migrate from shell.nix to up-to-date nix flake

guitargeek:
Nix flakes are the modern way to declare development environment, and this commit migrates the former shell.nix (which was very out-of-date) to a flake.

Now, if you use nix with flakes, you'll be able to instantiate the development shell environment with `nix develop`.

In the environment, you can build clad and run its test suite:

```txt
nix develop
cmake -S . -B build $CMAKE_FLAGS
cmake --build build -j$(nproc)
cmake --build build --target check-clad
```

Provided environment variables:
  - `CMAKE_FLAGS` : flags for configuring with CMake, mostly LLVM plumbing.

The environment is fully reproducible across systems, as the inputs are pinned in the `flake.lock` file. That makes is guaranteed to work. Note that developers can also override the pinned nixpkgs commits in the lock file if they would like the inputs to follow their system nixpkgs for example.


vgvassilev (approved):
Lgtm!

PR #1950, opened 2026-08-10
Give each delayed-store placeholder a distinct value

### LINKED ISSUE #1941: Regression: wrong gradient for `a * c[idx] * b` when `idx` is a compound expression
This is a regression in Clad `master` that breaks the RooFit unit tests.

Reproducer as a standalone Clad macro:
```C++
/// Wrong gradient for `a * c[idx] * b` when `idx` is a compound expression.
///
/// The adjoint of the left factor is emitted with the *middle* factor
/// substituted for it:
///
///     _d_x += 1 * c[i * 2] * c[i * 2];      // clad, wrong
///     _d_x += (x + 1.) * c[i * 2] * 1;
///
/// With a plain subscript (`c[i]`) or a literal one (`c[2]`) the very same
/// expression is differentiated correctly, so the trigger is the compound
/// index expression.
///
/// Found via RooFit::Detail::MathFuncs::multiVarGaussian(), which sums
/// `(x[i] - mu[i]) * covI[i * n + j] * (x[j] - mu[j])`.
///
/// Run with:  root -b -q -l cladIssueWrongProductPullback.C
/// clad master @ 80f6c90f (v2.4-60-g80f6c90f)

#include <cstdio>

double bad(double x, double const *c, int i)
{
   return (x + 1.) * c[i * 2] * (x + 1.);
}

double good(double x, double const *c, int i)
{
   return (x + 1.) * c[i] * (x + 1.);
}

void cladIssueWrongProductPullback()
{
   double c[4] = {0., 0., 3., 0.};
   double d = 0.;

   // d/dx of (x+1)^2 * c[2] at x = 5  ==  2 * 6 * 3  ==  36
   clad::gradient(bad, "x").execute(5., c, 1, &d);
   std::printf("c[i * 2]:  %g   (expected 36)\n", d);

   d = 0.;
   clad::gradient(good, "x").execute(5., c, 2, &d);
   std::printf("c[i]    :  %g   (expected 36)\n", d);

   clad::gradient(bad, "x").dump();
   clad::gradient(good, "x").dump();
}
```

The output is:
```txt
Processing cladIssueWrongProductPullback.C...
c[i * 2]:  27   (expected 36)
c[i]    :  36   (expected 36)
```

guitargeek:
DelayedGlobalStoreAndRef breaks the circular dependency in `x * y` -- each operand's adjoint needs the other's value -- by returning a placeholder literal that DelayedStoreResult::Finalize later splices out.

b75bba2c ("Do not reuse AST nodes in generated derivatives.") made de-sharing callsites CloneNode the placeholder, so PlaceholderReplacer started matching it by value rather than by pointer. Every placeholder was synthesized with the same value, ~0U, so value-equal placeholders became indistinguishable. In a nested product the outer placeholder is still live when the inner one is finalized, and the inner Finalize then spliced its operand into the outer placeholder's occurrences too:

```c++
double f(double x, const double* c, int i) {
  return (x + 1.) * c[i * 2] * (x + 1.);
}

*_d_x += 1 * c[i * 2] * c[i * 2];   // wrong, left operand clobbered
```

Hand each placeholder a serial number so no two are ever value-equal. The base is kept below 2^24 so that base + serial is exactly representable in `float` as well -- values near 2^32 round consecutive placeholders back together, which left the `float` case broken.

A plain `c[i]` subscript is not UsefulToStore, so it never takes the placeholder path; that is why only compound indices were affected. Found via RooFit's multiVarGaussian(), which sums
`(x[i] - mu[i]) * covI[i * n + j] * (x[j] - mu[j])`.

Fixes #1941

**It was confirmed that with this commit, Clad `master` works again for the RooFit usecase, modulo the problem reported in #1940.**

### DIFF
--- a/include/clad/Differentiator/ReverseModeVisitor.h
+++ b/include/clad/Differentiator/ReverseModeVisitor.h
@@ -120,6 +120,7 @@ namespace clad {
 
     unsigned outputArrayCursor = 0;
     unsigned numParams = 0;
+    unsigned m_PlaceholderCount = 0;
     llvm::SmallVector<clang::Expr*, 1> m_Pullback;
     const char* funcPostfix() const {
       if (m_DiffReq.Mode == DiffMode::jacobian)
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -4430,11 +4430,15 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       // VISITED forward value at every occurrence -- so a stored condition
       // `_cond0` (not the raw operand) is used, preserving correctness. The
       // placeholder may be cloned by de-sharing callsites; PlaceholderReplacer
-      // matches it structurally (by value) so clones are replaced too.
-      // Its value is arbitrary but distinctive to aid
-      // debugging if one ever leaks.
+      // matches it structurally (by value) so clones are replaced too. Its
+      // value is arbitrary but distinctive to aid debugging if one ever leaks,
+      // but it must not be shared among placeholder: nested multiplications
+      // keep the outer placeholder live while the inner one is finalized. The
+      // base stays below 2^24 so that base + serial is exact in `float` too.
+      constexpr unsigned placeholderBase = 0xBAD000;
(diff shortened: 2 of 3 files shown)


vgvassilev (approved):
LGTM!

PR #1951, opened 2026-08-10
Fix use-before-declaration in early-return reverse-mode lambda

### LINKED ISSUE #1940: Regression: error with referenced 'fact' before its declaration

guitargeek:
When a differentiated function has early returns, the master reverse sweep is wrapped in a shared [&] lambda invoked from every return path. LambdaCaptures::orderCaptureDecls hoists any local the lambda reads to before that lambda, but it only handled captured decls whose initializer was already computable at the hoist point ("self-contained"); anything else (e.g. `double y = n + x;`, where `n` isn't itself captured) was left after the lambda while still being referenced inside it.

That is more than a diagnostic: Clang's CodeGen, given a lambda body referencing a local before any DeclStmt introduces it, silently emitted the variable as a disconnected `internal global` instead of a real by-reference capture, decoupling the lambda's view of the variable from its actual computed value and corrupting the gradient.

Regression introduced by https://github.com/vgvassilev/clad/commit/7514327a0c44ee037297e7596297b361f964dae1 ("Encode reverse-mode early returns with a named lambda"), which added orderCaptureDecls with this gap.

Fix: a captured decl that can't move whole is now split into a zero-initialized declaration hoisted before the lambda (giving it a real, in-scope binding) and a plain assignment left at the original program point (preserving control flow and side-effect order for the early return). A decl whose only "availability" comes from such a placeholder is not marked available for downstream decls, so a value snapshot depending on it also goes through the split path instead of capturing the placeholder.

Hoisting is now decided per decl rather than per DeclStmt, so a mixed statement like `double a = n + x, c = 3.;` no longer has to stay behind wholesale on account of its uncaptured neighbour: `a` is split out and `c` is regrouped into a DeclStmt at the point its order requires. An array decl is the one thing that cannot be split -- there is no whole-object assignment to split into, and an element-wise copy would need ReverseModeVisitor::BuildArrayAssignment, out of this shared VisitorBase utility's reach -- so it is kept in place, and only when it would otherwise be split: a captured array that can move whole (an adjoint's `double _d_arr[2] = {0}`) still does.

Fixes https://github.com/vgvassilev/clad/issues/1940

### DIFF
--- a/include/clad/Differentiator/VisitorBase.h
+++ b/include/clad/Differentiator/VisitorBase.h
@@ -290,10 +290,14 @@ namespace clad {
       void collect(llvm::ArrayRef<clang::Stmt*> Body);
       /// A `[&]` capture binds a variable at the lambda's definition point, so
       /// every captured decl must precede the lambda. \p Prefix and \p Suffix
-      /// are the forward block split at the lambda's insertion point; move each
-      /// captured DeclStmt from Suffix to the end of Prefix when its
-      /// initializer references only names already live there -- the function's
-      /// parameters, \p AlreadyLive (decls emitted earlier), and Prefix.
+      /// are the forward block split at the lambda's insertion point. Each
+      /// captured decl in \p Suffix moves to the end of \p Prefix: whole, if
+      /// its initializer references only names already live there (the
+      /// function's parameters, \p AlreadyLive, and \p Prefix); otherwise
+      /// split into a zero-initialized declaration (moved) plus an assignment
+      /// left at the original spot, so the real value is still computed where
+      /// the original control flow put it. Array decls, which have no
+      /// whole-object assignment to split into, stay in \p Suffix.
       void orderCaptureDecls(llvm::SmallVectorImpl<clang::Stmt*>& Prefix,
                              llvm::SmallVectorImpl<clang::Stmt*>& Suffix,
                              llvm::ArrayRef<clang::Stmt*> AlreadyLive);
--- a/lib/Differentiator/VisitorBase.cpp
+++ b/lib/Differentiator/VisitorBase.cpp
@@ -26,6 +26,7 @@
 #include "clang/AST/RecursiveASTVisitor.h"
 #include "clang/AST/Stmt.h"
 #include "clang/AST/TemplateBase.h"
+#include "clang/AST/Type.h"
 #include "clang/Basic/OperatorKinds.h"
 #include "clang/Basic/SourceManager.h"
 #include "clang/Basic/Specifiers.h"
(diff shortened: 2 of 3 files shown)


vgvassilev (approved):
LGTM! Thank you, @guitargeek!

PR #1952, opened 2026-08-10
Allow safe qualification conversions in custom derivative lookup

### LINKED ISSUE #1947: A const-qualified parameter makes a custom derivative silently unusable
Custom-derivative lookup matches candidates against the expected signature by **exact type identity**, not call compatibility. A pullback that declares a read-only parameter as `const double*` where the primal has `double*` — a qualification conversion any real call would accept ([conv.qual]), and the spelling clang-tidy's `readability-non-const-parameter` actively recommends — is rejected:

- for **free functions** with a hard error ("user-defined derivative ... was  provided but not used"), refusing perfectly reasonable code;
- for **methods and overloaded operators** *silently*: the diagnostic is  suppressed (`enableDiagnostics` in `getOverloadExpr`, DiffPlanner.cpp), clad  falls back to differentiating the primal body, and the gradient quietly changes value. `-Wall` shows nothing.

The silent case is the dangerous one: adding a `const` to a pullback parameter changes numerical results with no diagnostic. The loud case is a usability trap: clang-tidy pushes every clad user toward the rejected spelling (it did exactly this on the pullback_state PR, in `benchmark/PullbackState.cpp`).

## Reproducer 1 — free function, loud rejection

```cpp
// const-pullback-rejected.C
#include <cstdio>

#ifdef CONST_P
#define P_QUAL const
#else
#define P_QUAL
#endif

void vop(double* p) {} // primal: no-op, so df/dx is 1 without the pullback

// The pullback only reads *p; const is the natural (and clang-tidy
// readability-non-const-parameter recommended) spelling.
void vop_pullback(P_QUAL double* p, double* d_p) { *d_p += 5.; }
} // namespace clad::custom_derivatives

double f(double x) {
  double v = x;
  vop(&v);
  return v;
}

int main() {
  auto grad = clad::gradient(f, "x");
  double dx = 0;
  grad.execute(3., &dx);
  std::printf("df/dx = %.2f  (6 = custom pullback used, 1 = it was ignored)\n",
              dx);
}
```

```
$ clang++ -std=c++17 -fplugin=clad.so const-pullback-rejected.C -I<clad>/include -o r.out
$ ./r.out
df/dx = 6.00  (6 = custom pullback used, 1 = it was ignored)

$ clang++ -std=c++17 -DCONST_P -fplugin=clad.so const-pullback-rejected.C -I<clad>/include -o r.out
const-pullback-rejected.C:26:3: error: user-defined derivative for 'vop' was provided but not used; expected signature 'void (double *, double *)' does not match
   26 |   vop(&v);
      |   ^
const-pullback-rejected.C:21:6: note: candidate 'vop_pullback' has type mismatch at 1st parameter (expected 'double *' but has 'const double *')
```

## Reproducer 2 — method, silent wrong gradient

```cpp
// const-pullback-silent.C
#include <cstdio>

#ifdef CONST_P
#define P_QUAL const
#else
#define P_QUAL
#endif

struct Box {
  void stash(double* p) {} // primal: no-op
};

namespace clad::custom_derivatives::class_functions {
void stash_pullback(Box* obj, P_QUAL double* p, Box* d_obj, double* d_p) {
  *d_p += 5.;
}
} // namespace clad::custom_derivatives::class_functions

double f(double x) {
  Box b;
  double v = x;
  b.stash(&v);
  return v;
}

int main() {
  auto grad = clad::gradient(f, "x");
  double dx = 0;
  grad.execute(3., &dx);
  std::printf("df/dx = %.2f  (6 = custom pullback used, 1 = it was ignored)\n",
              dx);
}
```

```
$ clang++ -std=c++17 -Wall -fplugin=clad.so const-pullback-silent.C -I<clad>/include -o s.out
$ ./s.out
df/dx = 6.00  (6 = custom pullback used, 1 = it was ignored)

$ clang++ -std=c++17 -Wall -DCONST_P -fplugin=clad.so const-pullback-silent.C -I<clad>/include -o s.out
$ ./s.out                       # compiles with no warning of any kind
df/dx = 1.00  (6 = custom pullback used, 1 = it was ignored)
```

Verified on clang/LLVM 21.1.8 and 22.1.7 (behavior identical) at clad master with the pullback_state commit.

## Where the behavior comes from

- `utils::MatchOverloadType` (lib/Differentiator/CladUtils.cpp): non-template candidates are compared with `ASTContext::hasSameFunctionTypeIgnoringExceptionSpec(FD->getType(), FnTy)` — strict type identity. Template candidates go through
  `Sema::DeduceTemplateArguments(FTD, ..., FnTy, ...)`, i.e. deduction against a *function type* ([temp.deduct.funcaddr]-style), which is likewise exact. The model is "taking the address of a function of this type", not "calling it with these arguments".
- `getOverloadExpr` (lib/Differentiator/DiffPlanner.cpp): when no candidate matches, the "provided but not used" error is emitted only when
  `enableDiagnostics` holds — which excludes `CXXMethodDecl`s, overloaded operators, `BaseFunctionName == "forward"`, and error-estimation requests (deliberately, because STL/Kokkos custom-derivative names collide). For all of those, a mismatched candidate silently degrades to differentiating the primal body.

## Proposed direction

Full call-viability matching is too loose — parameter identity is load-bearing where aliasing/out-params are involved (a `reverse_forw` taking `T` by value where the primal takes `T&` must stay rejected). The safe middle ground:

1. Accept a non-template candidate whose parameters differ from the expected signature only by a qualification conversion in the safe direction ([conv.qual]) — i.e. the candidate parameter is *more* const-qualified at each pointer/reference level. Checkable per parameter with `Sema::IsQualificationConversion(expected, candidate)`. Return type and arity stay exact.
2. Templates: either extend later via call-mode deduction ([temp.deduct.call]), or document that templated custom derivatives must spell parameter types exactly.
3. The existing mismatch note already pinpoints the parameter; attach a `FixItHint` removing the offending qualifier so the loud case is self-healing.
4. Independently: when `enableDiagnostics` is off and candidates for the *exact base name* exist but none matched, consider at least a warning — silent fallback from a one-qualifier mismatch is very hard to debug (this is the same failure shape as the templated-thrust `sort_by_key` regression: candidates exist, none match, gradients silently change).

## Workaround

Mirror the primal's parameter types exactly and suppress the tidy check:

```cpp
// NOLINTNEXTLINE(readability-non-const-parameter): clad matches a custom
// derivative against the primal's exact parameter types; const here breaks it.
void op_pullback(double* p, double* d_p) { *d_p += 2.0 * *p; }
```

Tests to ship with a fix: positive (const-qualified pullback accepted and used, for a free function and a method) and negative (arity/return/aliasing mismatches still rejected with the existing notes).

Elvand-Lie:
## Problem

Custom derivative lookup required exact function type identity. A pullback that safely adds `const` to a pointer parameter was therefore unusable: the free-function case emitted a hard error, while the method case silently ignored the custom pullback and produced `1.00` instead of `6.00`.

## Fix

For non-template candidates, accept a parameter difference only when Clang recognizes the expected-to-candidate change as a qualification conversion. Reconstructing the candidate type with the expected parameter list keeps the return type, arity, variadic state, calling convention, and other function properties exact. Template deduction and reference binding remain unchanged.

This intentionally follows Clang qualification-conversion semantics rather than being const-only: safe nested-pointer qualification and adding `volatile` are accepted, while unsafe `double**` to `const double**`, qualifier removal, and pointer-to-value changes remain rejected. Reference parameters stay exact because reference binding uses separate compatibility rules and is outside this issue-specific change.

Clang still performs normal overload resolution. If equally ranked qualified candidates are ambiguous, Clad now preserves the ambiguity diagnostic and returns cleanly instead of assuming overload resolution produced a `CallExpr` and asserting.

The regression checks generated calls and executes free-function, method, safe nested-pointer, volatile, exact-overload-preference, and pushforward cases. Diagnostics pin the unsafe and ambiguous boundaries.

## Testing

- untouched-upstream RED: const pointer, safe nested-pointer, volatile, and pushforward candidates are rejected; patched runtime output is `{6.00, 6.00}` and `{3.00, 4.00, 5.00, 6.00}`
- unsafe deep-pointer qualification is rejected on both untouched upstream and the patch
- plain Clang confirms `double**` to `const double**` is invalid and `double**` to `const double* const*` is valid
- const and volatile candidates with no exact overload now emit `call to ambiguous_op_pullback is ambiguous` without a Clad assertion; untouched upstream rejects both as mismatched
- focused lit coverage: 10/10 passed (issue regression, custom derivative diagnostics, function calls, member functions, and error estimation)
- unit binaries: 5/5 tests passed
- `clang-format-18 --dry-run --Werror` and `git diff --check` passed
- `clang-tidy-18` with changed-line filters: no diagnostics in either changed production file
- full `check-clad`: 178 passed, 1 expected failure, 24 unsupported, and 2 failures (`issue-1815`, `issue-1855`) that reproduce with the saved untouched-upstream plugin at `b59b4439`; both assert in unrelated loop-checkpoint code at `ClangPlugin.cpp:312`

Fixes: #1947

### DIFF
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -1031,6 +1031,37 @@ namespace clad {
       return cast<TemplateDecl>(TapeR.getFoundDecl());
     }
 
+    static bool MatchFunctionTypeWithQualifiedParams(Sema& S,
+                                                     QualType ExpectedFnTy,
+                                                     QualType CandidateFnTy) {
+      ASTContext& C = S.getASTContext();
+      if (C.hasSameFunctionTypeIgnoringExceptionSpec(ExpectedFnTy,
+                                                     CandidateFnTy))
+        return true;
+
+      const auto* Expected = ExpectedFnTy->getAs<FunctionProtoType>();
+      const auto* Candidate = CandidateFnTy->getAs<FunctionProtoType>();
+      if (!Expected || !Candidate ||
+          Expected->getNumParams() != Candidate->getNumParams())
+        return false;
+
+      for (unsigned I = 0; I < Expected->getNumParams(); ++I) {
+        QualType From = Expected->getParamType(I);
+        QualType To = Candidate->getParamType(I);
+        bool ObjCLifetimeConversion = false;
+        if (!C.hasSameType(From, To) &&
+            !S.IsQualificationConversion(From, To, /*CStyle=*/false,
+                                         ObjCLifetimeConversion))
+          return false;
+      }
+
+      QualType CandidateWithExpectedParams = C.getFunctionType(
+          Candidate->getReturnType(), Expected->getParamTypes(),
+          Candidate->getExtProtoInfo());
+      return C.hasSameFunctionTypeIgnoringExceptionSpec(
+          ExpectedFnTy, CandidateWithExpectedParams);
+    }
+
     Expr* MatchOverloadType(Sema& S, QualType FnTy, LookupResult& Overloads,
                             TemplateSpecCandidateSet& FailedCandidates) {
       CXXScopeSpec SS;
@@ -1072,7 +1103,7 @@ namespace clad {
         if (!FD)
           continue;
         // Overload is just a FunctionDecl, check if the signature matches.
-        if (C.hasSameFunctionTypeIgnoringExceptionSpec(FD->getType(), FnTy))
+        if (MatchFunctionTypeWithQualifiedParams(S, FnTy, FD->getType()))
           return S.BuildDeclarationNameExpr(SS, Overloads, /*ADL=*/false).get();
       }
       return nullptr;
--- a/test/Diagnostics/CustomDerivativeDiagnostics.C
+++ b/test/Diagnostics/CustomDerivativeDiagnostics.C
@@ -1,5 +1,4 @@
 // RUN: %cladclang %s -I%S/../../include -fsyntax-only -Xclang -verify 2>&1 | %filecheck %s
-
 #include "clad/Differentiator/Differentiator.h"
 
 namespace helpers {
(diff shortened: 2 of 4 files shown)


lib/Differentiator/CladUtils.cpp
```diff
@@ -1072,7 +1099,8 @@ namespace clad {
         if (!FD)
           continue;
         // Overload is just a FunctionDecl, check if the signature matches.
-        if (C.hasSameFunctionTypeIgnoringExceptionSpec(FD->getType(), FnTy))
+        if (C.hasSameFunctionTypeIgnoringExceptionSpec(FD->getType(), FnTy) ||
+            MatchFunctionTypeWithQualifiedParams(S, FnTy, FD->getType()))
```
vgvassilev: Can we sink `C.hasSameFunctionTypeIgnoringExceptionSpec` into `MatchFunctionTypeWithQualifiedParams` and have a single call to it with an early return?

test/Diagnostics/CustomDerivativeDiagnostics.C
```diff
@@ -1,4 +1,5 @@
 // RUN: %cladclang %s -I%S/../../include -fsyntax-only -Xclang -verify 2>&1 | %filecheck %s
+// expected-error@* {{call to 'ambiguous_op_pullback' is ambiguous}}
```
vgvassilev: Why is this not at the relevant line where it is produced?

vgvassilev (approved):
Lgtm!

PR #1884, opened 2026-07-11
Fix null tangents for const pointer subscripts

Elvand-Lie:
## Summary

- guard nullable const-pointer tangent subscripts with `_d_ptr != nullptr`
- parenthesize the conditional so it composes correctly with surrounding arithmetic
- add a nested differentiation regression for inactive `const T*` tangents

Fixes: #1871

## Testing

- `Regressions/issue-1871.cpp`
- full regression suite: 18 passed, 1 CUDA unsupported
- verified neighboring `Regressions/issue-1804.cpp` higher-order generation
- changed-line `clang-format --dry-run --Werror`
- `git diff --check`

## Performance

At `-O3`, the guard is eliminated in the reproducer's outer derivative because the null tangent is known at the call site. The previous code instead invoked undefined behavior and optimized the derivative to an empty function.

PR #1953, opened 2026-08-11
Read a null tangent as an identically zero derivative

### LINKED ISSUE #1804: Regression since Clad 2.2: `clad::hessian` fails for nested functions with `double const *` paramters

### LINKED ISSUE #1871: `clad::differentiate` segfaults on nested function that forwards constant array parameter
Since we're now trying to use Clad-generated Hessians in RooFit, forward mode bugs like these become now apparent.

FYI, @vgvassilev.

```c++

#include <iostream>

double inner_func(double *params, double const *obs)
{
   return params[0] * params[0] * obs[0];
}

double outer_func(double *params, double const *obs)
{
   // The crash only happens if the differentiated function forwards the
   // non-differentiated pointer parameter to a callee. Inlining the body of
   // inner_func here works fine:
   // return params[0] * params[0] * obs[0];
   return inner_func(params, obs);
}

#pragma clad ON
void requests()
{
   clad::differentiate(outer_func, "params[0]");
}
#pragma clad OFF

void clad_pushforward_null_tangent()
{
   double params[]{0.5};
   double obs[]{3.0};
   // Expected output: 3 (= 2 * params[0] * obs[0]). Instead: segfault.
   std::cout << outer_func_darg0_0(params, obs) << std::endl;
}
```
This segfaults:
```txt
Processing clad_pushforward_null_tangent.C...

===========================================================
There was a crash.
This is the entire stack trace of all threads:
===========================================================
#0  0x00007ffff729a03e in __internal_syscall_cancel () from /nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib/libc.so.6
#1  0x00007ffff729a084 in __syscall_cancel () from /nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib/libc.so.6
#2  0x00007ffff730dd4f in wait4 () from /nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib/libc.so.6
#3  0x00007ffff72586cd in do_system () from /nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib/libc.so.6
#4  0x00007ffff7d2dc62 in TUnixSystem::StackTrace (this=0x5555555a7370) at /home/rembserj/code/root/root_src/core/unix/src/TUnixSystem.cxx:2424
#5  0x00007ffff7d31ae3 in TUnixSystem::DispatchSignals (this=0x5555555a7370, sig=kSigSegmentationViolation) at /home/rembserj/code/root/root_src/core/unix/src/TUnixSystem.cxx:3651
#6  <signal handler called>
#7  0x00007ffff71ad078 in ?? ()
#8  0x00007ffffffe7200 in ?? ()
#9  0x00007ffff71ad0c6 in ?? ()
#10 0x00007ffffffe7230 in ?? ()
#11 0x00007ffff71ad191 in ?? ()
#12 0x00007ffffffe7230 in ?? ()
#13 0x3fe0000000000000 in ?? ()
#14 0x4008000000000000 in ?? ()
#15 0x00007ffff71ab268 in ?? ()
#16 0x00007ffffffe72a0 in ?? ()
#17 0x00007fffebf304da in cling::IncrementalExecutor::executeWrapper (this=<optimized out>, function=..., returnValue=0x7ffffffe7ae0) at /home/rembserj/code/root/root_src/interpreter/cling/lib/Interpreter/IncrementalExecutor.cpp:217
===========================================================


The lines below might hint at the cause of the crash. If you see question
marks as part of the stack trace, try to recompile with debugging information
enabled and export CLING_DEBUG=1 environment variable before running.
You may get help by asking at the ROOT forum https://root.cern/forum
preferably using the command (.forum bug) in the ROOT prompt.
Only if you are really convinced it is a bug in ROOT then please submit a
report at https://root.cern/bugs or (preferably) using the command (.gh bug) in
the ROOT prompt. Please post the ENTIRE stack trace
from above as an attachment in addition to anything else
that might help us fixing this issue.
===========================================================
#7  0x00007ffff71ad078 in ?? ()
#8  0x00007ffffffe7200 in ?? ()
#9  0x00007ffff71ad0c6 in ?? ()
#10 0x00007ffffffe7230 in ?? ()
#11 0x00007ffff71ad191 in ?? ()
#12 0x00007ffffffe7230 in ?? ()
#13 0x3fe0000000000000 in ?? ()
#14 0x4008000000000000 in ?? ()
#15 0x00007ffff71ab268 in ?? ()
#16 0x00007ffffffe72a0 in ?? ()
#17 0x00007fffebf304da in cling::IncrementalExecutor::executeWrapper (this=<optimized out>, function=..., returnValue=0x7ffffffe7ae0) at /home/rembserj/code/root/root_src/interpreter/cling/lib/Interpreter/IncrementalExecutor.cpp:217
===========================================================
```

guitargeek:
Forward mode has to supply a tangent for every argument of a call it differentiates. For a const-qualified pointer parameter it cannot synthesize one, because the size is unknown, so it passes a null tangent instead:

  inner_func_pushforward(params, obs, (double[3]){1., 0., 0.}, nullptr);

The pushforward then read `_d_obs[0]` unconditionally and dereferenced the null pointer. #1805 fixed the type mismatch that used to mask this, leaving the null dereference behind; at -O0 it segfaults, and with optimizations the undefined behaviour lets the derivative bodies be deleted outright, which is why hessians silently came out all zeros.

Guard such reads so they produce the zero the null tangent stands for. The guard is limited to derivative parameters standing for const- qualified pointer parameters, which are exactly the ones clad nulls out: a non-const pointer parameter is either the independent variable or a hard error, and so is never null and generates no guard.

The reverse sweep needs no change of its own. The pullback of `cond ? A : 0` is `if (cond) <pullback of A>`, so the accumulation into the null adjoint of a null tangent ends up guarded by the same condition, and no existing generated code changes.

The regression test only compiled its input, and its CHECK line never ran for want of a FileCheck pipe, so the crash went unnoticed. Run it and check the hessian against the analytic values.

Fixes #1804

### DIFF
--- a/include/clad/Differentiator/BaseForwardModeVisitor.h
+++ b/include/clad/Differentiator/BaseForwardModeVisitor.h
@@ -65,20 +65,14 @@ public:
 
   virtual void ExecuteInsidePushforwardFunctionBlock() {}
 
-  /// Whether \p tangent is one clad itself may pass as a null pointer: the
-  /// derivative parameter standing for a const-qualified pointer parameter.
-  /// Clad cannot synthesize a tangent buffer for such a parameter, because its
-  /// size is unknown, so a pushforward call receives a null tangent for it,
-  /// spelling "the derivative of this argument is identically zero". Every
-  /// other tangent is either a local buffer or a pointer the caller has to
-  /// provide, and is therefore never null.
-  bool mayBeNullTangent(const clang::Expr* tangent) const;
-
   /// \returns \p read, the derivative expression dereferencing \p tangent,
   /// wrapped so that it evaluates to \p zero instead of dereferencing a null
-  /// \p tangent. Returns \p read unchanged when \p tangent cannot be null.
-  clang::Expr* GuardNullTangentRead(const clang::Expr* tangent,
-                                    clang::Expr* read, clang::Expr* zero);
+  /// \p tangent. \p ptr is the primal pointer expression \p tangent is the
+  /// tangent of; the request tells whether its tangent may be null, and \p read
+  /// is returned unchanged when it cannot be.
+  clang::Expr* GuardNullTangentRead(const clang::Expr* ptr,
+                                    clang::Expr* tangent, clang::Expr* read,
+                                    clang::Expr* zero);
 
   virtual StmtDiff
   VisitArraySubscriptExpr(const clang::ArraySubscriptExpr* ASE);
--- a/include/clad/Differentiator/BaseForwardModeVisitor.h
+++ b/include/clad/Differentiator/BaseForwardModeVisitor.h
@@ -3,6 +3,7 @@
 
 #include "Compatibility.h"
 #include "VisitorBase.h"
+#include "clang/AST/Expr.h"
 #include "clang/AST/ExprCXX.h"
 #include "clang/AST/OpenMPClause.h"
 #include "clang/AST/RecursiveASTVisitor.h"
@@ -63,6 +64,21 @@ public:
   }
 
   virtual void ExecuteInsidePushforwardFunctionBlock() {}
+
+  /// Whether \p tangent is one clad itself may pass as a null pointer: the
+  /// derivative parameter standing for a const-qualified pointer parameter.
+  /// Clad cannot synthesize a tangent buffer for such a parameter, because its
+  /// size is unknown, so a pushforward call receives a null tangent for it,
+  /// spelling "the derivative of this argument is identically zero". Every
+  /// other tangent is either a local buffer or a pointer the caller has to
+  /// provide, and is therefore never null.
+  bool mayBeNullTangent(const clang::Expr* tangent) const;
+
+  /// \returns \p read, the derivative expression dereferencing \p tangent,
+  /// wrapped so that it evaluates to \p zero instead of dereferencing a null
+  /// \p tangent. Returns \p read unchanged when \p tangent cannot be null.
+  clang::Expr* GuardNullTangentRead(const clang::Expr* tangent,
+                                    clang::Expr* read, clang::Expr* zero);
 
   virtual StmtDiff
   VisitArraySubscriptExpr(const clang::ArraySubscriptExpr* ASE);
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -49,6 +49,7 @@
 #include <cstddef>
 #include <memory>
 #include <optional>
+#include <set>
 #include <string>
 #include <utility>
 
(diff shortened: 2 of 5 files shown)


include/clad/Differentiator/BaseForwardModeVisitor.h
```diff
@@ -31,6 +34,13 @@ class BaseForwardModeVisitor
                                           clang::OMPClause*>,
       public VisitorBase {
   unsigned m_IndependentVarIndex = ~0;
+  /// Pointer variables of the function being differentiated whose tangent may
+  /// be null, as found by CollectMaybeNullPointers().
+  llvm::SmallPtrSet<const clang::VarDecl*, 8> m_MaybeNullPointerVars;
```
vgvassilev: Can we avoid that state here? If we are doing some visitation probably we should do it inside the DiffPlanner and record it in the DiffRequest information probably.

guitargeek: The state is avoided now.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -419,6 +419,94 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
...
+      if (Def && Def->hasBody()) {
+        std::set<const VarDecl*>& MaybeNull = m_NullTangentInfo.MaybeNullPtrs;
+        // Clad has no tangent buffer to pass for a const-qualified pointer
+        // parameter; the pointers of the body derived from one inherit that.
+        for (const ParmVarDecl* PVD : Def->parameters()) {
+          QualType T = PVD->getType();
+          if (T->isPointerType() && T->getPointeeType().isConstQualified())
+            MaybeNull.insert(PVD);
+        }
+        MaybeNullPointerPropagator(MaybeNull).runToFixpoint(Def->getBody());
```
vgvassilev, line 529: It thought we could use the visitation pass that already exists in the DiffPlanner/DiffCollector. We don't need to revisit the binary operators again unless somehow we require some other state created by it..

vgvassilev (approved):
LGTM. Consider merging #1884 before.

PR #1956, opened 2026-08-11
Support references and const locals across an early-return lambda

### LINKED ISSUE #1954: Reverse mode emits `double &_d_x = {};` for a reference variable when the function has an early return
## Summary

`clad::gradient()` fails to compile a function that (a) has an early `return` and (b) declares a reference to a local that is then used in a further statement. The generated code binds a non-const lvalue reference to an empty init list:

```
error: non-const lvalue reference to type 'double' cannot bind to an initializer list temporary
warning: clad internally reused a 'DeclRefExpr' AST node while differentiating 'f_grad';
         this is a clad bug -- please report it at https://github.com/vgvassilev/clad
```

This bites in practice through `clad::hessian()`, because clad's own forward-mode output contains exactly such a reference variable
(`double &_t1 = _t0.value;`, bound to the `.value` of a `ValueAndPushforward`). So every `clad::hessian()` of a function that calls a math builtin after an early return fails, even though `clad::gradient()` of the same function is fine.

We hit this in ROOT: `clad::hessian()` of a likelihood using `ROOT::Math::lognormal_pdf` cannot be generated.

## Minimal reproducer (reverse mode only, no Hessian)

```cpp
double f(double x) {
   if (x <= 0)
      return 0.0;      // remove this and it compiles
   double y = x * x;
   double &t1 = y;     // reference to a local
   double t2 = t1 * t1;// the reference is used in a further statement
   return t2;
}

clad::gradient(f);      // error
```

## Reproducer via the Hessian (the way we ran into it)

```cpp
double pdf(double x, double m) {
   if (x <= 0)
      return 0.0;
   return std::sqrt(x) * m * m;
}

clad::gradient(pdf);    // fine
clad::hessian(pdf);     // error, in 'pdf_darg0_grad'
```

`clad::differentiate(pdf, 0).dump()` shows where the reference comes from:

```cpp
double pdf_darg0(double x, double m) {
    double _d_x = 1;
    double _d_m = 0;
    if (x <= 0)
        return 0.;
    ValueAndPushforward<double, double> _t0 =
        clad::custom_derivatives::std::sqrt_pushforward(x, _d_x);
    double &_t1 = _t0.value;      // <-- reference variable
    double _t2 = _t1 * m;         // <-- used in a further statement
    return (_t0.pushforward * m + _t1 * _d_m) * m + _t2 * _d_m;
}
```

Reverse-differentiating this function is what fails.

## Original ROOT-level symptom

```cpp
double func(double *params) {
   return ROOT::Math::lognormal_pdf(params[0], params[1], params[2]);
}
clad::hessian(func, "params[0:2]");
```

```
PdfFuncMathCore.h:476:17: warning: clad internally reused a 'BinaryOperator' AST node
      while differentiating 'lognormal_pdf_pushforward'
PdfFuncMathCore.h:476:17: error: non-const lvalue reference to type 'double' cannot bind
      to an initializer list temporary
warning: clad internally reused a 'MemberExpr' AST node while differentiating
      'lognormal_pdf_pushforward_pullback'
```

`ROOT::Math::lognormal_pdf` has the same shape: an early `return 0.0` guard, then `std::log`/`std::abs`/`std::sqrt`/`std::exp` inside a nested binary expression.

## What is and is not needed to trigger it

Each of these on its own is fine; the failure needs all three together.

| Ingredient | Needed? |
|---|---|
| early `return` in the function | yes — replacing it with `if (x <= 0) m = 1.0;` compiles |
| a reference bound to a **local** | yes — `double &t1 = x;` (a parameter) compiles |
| the reference used in a **further statement** | yes — using it directly in the `return` compiles |
| reverse mode | yes — forward mode alone is fine |
| the Hessian / a math builtin | no — only how the reference appears in practice |

guitargeek:
A function with early returns keeps its master reverse sweep in a shared [&] lambda, so every local the sweep reads must be declared before it. LambdaCaptures::orderCaptureDecls hoists such a local by splitting it: a placeholder declaration before the lambda plus an assignment left at the original program point. Two kinds of declaration do not survive that split as spelled, and clad emitted code that does not compile:

```c++
  double f(double x) {
     if (x <= 0) return 0.0;
     double y = x * x;
     double &t1 = y;      // -> double &t1 = {};
     double t2 = t1 * t1;
     return t2;
  }
```

A reference cannot be declared unbound and re-seated later; the same body with `const double t1 = y + 1;` cannot take the assignment either.

Fix: DifferentiateVarDecl clones a function-scope reference of an early-return function as a pointer -- the shape a reference promoted to the function global scope already gets -- and the split drops the const of the clone it hoists. A reference bound to a temporary keeps its reference type: there is no lvalue for the pointer to point at, and clad supports neither encoding of one.

That alone left the gradient crashing on the early path. The lambda runs the whole reverse sweep on every return path, including the part undoing statements the taken path never reached, so it reads and writes through pointers the forward sweep never assigned. Those accesses carry zero adjoints and cannot change the gradient, but a null placeholder makes them a null dereference. That is reachable on master already, through a reference promoted from a nested scope -- Gradients.C's f_ref_in_rhs carried it latent, executed on one path only. A moved pointer declaration now holds a fresh zero dummy of the pointee type (BuildDereferenceablePlaceholder), one dummy per declaration: sharing would let a dead statement's write reach a later dead read, and from there a real adjoint.

Fixes #1954

### DIFF
--- a/lib/Differentiator/VisitorBase.cpp
+++ b/lib/Differentiator/VisitorBase.cpp
@@ -21,6 +21,7 @@
 #include "clang/AST/Decl.h"
 #include "clang/AST/DeclCXX.h"
 #include "clang/AST/Expr.h"
+#include "clang/AST/ExprCXX.h"
 #include "clang/AST/NestedNameSpecifier.h"
 #include "clang/AST/OperationKinds.h"
 #include "clang/AST/RecursiveASTVisitor.h"
@@ -454,6 +455,21 @@ namespace clad {
         bool Hoistable = VD && contains(VD);
         if (Hoistable && selfContained(VD)) {
           flushKept();
+          // Moved wholesale, the decl may still carry no real initial value:
+          // default init of a scalar or of a trivially-default-constructible
+          // record leaves indeterminate bits. In place that was fine -- the
+          // original program wrote them before any read -- but hoisted above
+          // the lambda they become readable by the reverse sweep on a path
+          // that returns before those writes. The adjoints there are zero, so
+          // only definedness is at stake; a zero initializer settles it.
+          const Expr* Init = VD->getInit();
+          bool Indeterminate = !Init;
+          if (const auto* CE = dyn_cast_or_null<CXXConstructExpr>(Init))
+            Indeterminate = CE->getConstructor()->isDefaultConstructor() &&
+                            CE->getConstructor()->isTrivial();
+          if (Indeterminate)
+            if (Expr* Zero = m_V.getZeroInit(VD->getType()))
+              m_V.SetDeclInit(VD, Zero);
           Prefix.push_back(m_V.BuildDeclStmt(VD));
           note(Prefix.back());
           continue;
@@ -470,8 +486,27 @@ namespace clad {
         // selfContained() is true for an initializer-less decl, so reaching
         // here means there is an initializer to move into the assignment.
         Expr* Init = VD->getInit();
+        // A const clone cannot take that assignment. Placeholder plus one
+        // assignment still leave the clone a single value for the rest of its
+        // lifetime, so dropping the qualifier costs nothing. (A decl promoted
+        // to the function global scope is split the same way and loses its
+        // const up front, in ReverseModeVisitor::DifferentiateVarDecl.)
+        QualType VDTy = VD->getType();
+        if (VDTy.isLocalConstQualified()) {
+          VDTy.removeLocalConst();
+          VD->setType(VDTy);
+          VD->setTypeSourceInfo(
+              m_V.m_Context.getTrivialTypeSourceInfo(VDTy, noLoc));
+        }
         Expr* DeclRef = m_V.BuildDeclRef(VD);
-        m_V.SetDeclInit(VD, m_V.getZeroInit(VD->getType()));
+        // The lambda reads the placeholder on a path that returns before the
+        // assignment below, so a pointer needs a dereferenceable one; its
+        // dummy declaration goes into Prefix ahead of VD's.
+        Expr* Placeholder =
+            m_V.BuildDereferenceablePlaceholder(VD->getType(), Prefix);
+        if (!Placeholder)
+          Placeholder = m_V.getZeroInit(VD->getType());
+        m_V.SetDeclInit(VD, Placeholder);
         flushKept();
         Prefix.push_back(m_V.BuildDeclStmt(VD));
         // Not noted: the hoisted decl only holds a placeholder until the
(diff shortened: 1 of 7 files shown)


lib/Differentiator/VisitorBase.cpp
```diff
@@ -712,6 +731,20 @@ namespace clad {
...
 
+  Expr* VisitorBase::BuildDereferenceablePlaceholder(
+      QualType PtrTy, llvm::SmallVectorImpl<Stmt*>& Block) {
+    const auto* PT = PtrTy->getAs<PointerType>();
+    if (!PT || !m_DiffReq.hasEarlyReturns())
+      return nullptr;
+    // Only for a pointee clad can bring into existence by itself.
+    QualType Pointee = PT->getPointeeType().getUnqualifiedType();
+    if (!Pointee->isScalarType())
+      return nullptr;
```
vgvassilev, line 764: This seems to be dead code but maybe we can trigger with a test.

vgvassilev (approved):
Lgtm!

vgvassilev:
Isn’t the last push contradicting: we enable const T& but change the pullbacks to take T?

guitargeek:
> Isn’t the last push contradicting: we enable const T& but change the pullbacks to take T?

Here is the LLMs take on it:

The last push was a rebase onto master, which had just picked up #1885, which seems to be in tension at first, but that PR and this are on opposite sides of the lookup.

`getOverloadExpr` asks `utils::GetDerivativeType` what a pushforward for `std::max` should look like, and for pushforwards that is `ValueAndPushforward<oRetTy, PushFwdTy>` with `oRetTy` taken verbatim from the original declaration. `std::max` returns `const T&`, so the expected type is `ValueAndPushforward<const T&, const T&>(...)`, and `MatchOverloadType` deduces against exactly that.

That `const T&` in the *result* is what breaks #1872: at second order clad differentiates the generated first-derivative code, which default-constructs and assigns the `ValueAndPushforward` temporary -- impossible with two reference members ("implicitly-deleted default constructor", "cannot assign to non-static data member `pushforward`"). Hence the by-value return.

On its own that change would make the builtin stop matching the expected signature and get silently dropped. The `DiffPlanner` hunk closes that gap: after the exact deduction fails, retry with the *expectation* relaxed to `ValueAndPushforward<T, T>`. So we aren't enabling `const T&` derivatives -- we still accept the `const T&`-returning *original*, we just no longer require its pushforward to propagate the reference.

The retry runs after the exact match rather than relaxing upfront so the genuinely reference-returning pushforwards in `STLBuiltins.h` (const `operator[]`, `at`, `front`, `back`) keep matching as they do today. They have the same second-order limitation, but that's a separate change.

The in-code comment is what invites the misreading, so I've amended the commit in this PR to to reword it.

vgvassilev:
> > Isn’t the last push contradicting: we enable const T& but change the pullbacks to take T?
> 
> Here is the LLMs take on it:
> 
> The last push was a rebase onto master, which had just picked up #1885, which seems to be in tension at first, but that PR and this are on opposite sides of the lookup.
> 
> `getOverloadExpr` asks `utils::GetDerivativeType` what a pushforward for `std::max` should look like, and for pushforwards that is `ValueAndPushforward<oRetTy, PushFwdTy>` with `oRetTy` taken verbatim from the original declaration. `std::max` returns `const T&`, so the expected type is `ValueAndPushforward<const T&, const T&>(...)`, and `MatchOverloadType` deduces against exactly that.
> 
> That `const T&` in the _result_ is what breaks #1872: at second order clad differentiates the generated first-derivative code, which default-constructs and assigns the `ValueAndPushforward` temporary -- impossible with two reference members ("implicitly-deleted default constructor", "cannot assign to non-static data member `pushforward`"). Hence the by-value return.
> 
> On its own that change would make the builtin stop matching the expected signature and get silently dropped. The `DiffPlanner` hunk closes that gap: after the exact deduction fails, retry with the _expectation_ relaxed to `ValueAndPushforward<T, T>`. So we aren't enabling `const T&` derivatives -- we still accept the `const T&`-returning _original_, we just no longer require its pushforward to propagate the reference.
> 
> The retry runs after the exact match rather than relaxing upfront so the genuinely reference-returning pushforwards in `STLBuiltins.h` (const `operator[]`, `at`, `front`, `back`) keep matching as they do today. They have the same second-order limitation, but that's a separate change.
> 
> The in-code comment is what invites the misreading, so I've amended the commit in this PR to to reword it.

Ah, ok. I think I get it now.

vgvassilev:
Maybe we can capture that discussion/explanation in a tight comment…

guitargeek:
> Maybe we can capture that discussion/explanation in a tight comment…

Good idea! The explanation is now in the code.

PR #1957, opened 2026-08-11
Do not fabricate tangents for aliases of read-only data

### LINKED ISSUE #1955: `clad::hessian()` drops the const qualifier for a local `const double *` alias of a const pointer parameter
## Summary

A function that binds a local `const double *` to a const pointer parameter cannot be Hessian-differentiated. Clad builds the adjoint of that local as a non-const `double *` and initializes/assigns it from the `const double *` primal:

```
error: cannot initialize a variable of type 'double *' with an rvalue of type 'const double *'
error: assigning to 'double *' from 'const double *' discards qualifiers
```

`clad::gradient()` and `clad::differentiate()` on the same function both succeed. Only the forward-over-reverse composition fails.

There is no way to avoid the const on the parameter: making it `double *` instead is rejected outright with

```
error: dependent non-const pointer and array parameters are not supported;
       differentiate w.r.t. 'xlArr' or mark it const
```

so the only shape clad accepts for the parameter is the one that then breaks the Hessian.

## Minimal reproducer

```cpp
double f(double *params, double const *xlArr) {
   double const *t5 = xlArr;      // local alias -- the whole trigger
   return params[0] * t5[0];
}

clad::gradient(f, "params");       // ok
clad::differentiate(f, "params[0]"); // ok
clad::hessian(f, "params[0]");     // error
```

Removing the alias and writing `params[0] * xlArr[0]` compiles.

## What is and is not needed

| Ingredient | Needed? |
|---|---|
| local `const double *` alias of the parameter | yes — indexing the parameter directly compiles |
| pointer arithmetic in the alias | no — `= xlArr` is enough |
| the alias inside a loop / `#pragma clad checkpoint loop` | no — it only adds the second diagnostic |
| naming the alias rather than passing the expression to the callee | yes — `interp(xlArr + 1, params[0])` compiles, `t5 = xlArr + 1; interp(t5, ...)` does not |
| `clad::hessian` | yes — gradient and forward mode each succeed alone |

A `const double &` local alias (`double const &t5 = xlArr[1];`) fails too, with
a sibling diagnostic:
`non-const lvalue reference to type 'double' cannot bind to a temporary of type 'double'`.

## Where the alias comes from in practice

We hit this in ROOT. `RooFit`'s code generation emits, for a `PiecewiseInterpolation` (`roofit/codegen/src/CodegenImpl.cxx:212`):

```cpp
double roo_codegen_0(double* params, double const* obs, double const* xlArr) {
   ...
   #pragma clad checkpoint loop
   for (int loopIdx1 = 0; loopIdx1 < obs[1]; loopIdx1++) {
      unsigned int t4 = RooFit::Detail::MathFuncs::uniformBinNumber(...);
      double const* t5 = xlArr + 2 + 1 * t4;
      double const* t6 = xlArr + 4 + 1 * t4;
      double t7 = *(xlArr + 0 + t4);
      double total_t3 = RooFit::Detail::MathFuncs::flexibleInterp(4, t8, 1, t5, t6, 1, t7, 0);
      ...
   }
}
```

`clad::gradient` of this function works; `clad::hessian` fails on the `t5`/`t6` declarations. Reproducer for that exact shape:

```cpp
double f(double *params, double const *obs, double const *xlArr) {
   double sum = 0.0;
#pragma clad checkpoint loop
   for (int i = 0; i < 2; i++) {
      unsigned int t4 = i;
      double const *t5 = xlArr + 2 + 1 * t4;
      sum += params[0] * t5[0];
   }
   return sum;
}
clad::hessian(f, "params[0]");
```

## Note on the intermediate

`clad::differentiate(f, "params[0]").dump()` for the minimal case gives

```cpp
double f_darg0_0(double *params, const double *xlArr) {
    const double *_d_t5 = nullptr;
    const double *t5 = xlArr + 1;
    return 1. * t5[0] + params[0] * (_d_t5 ? _d_t5[0] : 0.);
}
```

so the forward pass keeps both the primal and its tangent `const`-qualified. Note that hand-writing this function and calling `clad::gradient()` on it does *not* reproduce the error, so the problem is not simply reverse mode over this source — it is specific to the Hessian pipeline's internal representation.

guitargeek:
A local `const double*` or `const double&` bound to a const pointer parameter has no tangent, because the parameter it aliases cannot have one. Forward mode fabricated one regardless: a null pointer literal typed `const double*`, or a reference bound to the literal 0.

The null tangent was then dereferenced in the generated derivative, and both shapes broke `clad::hessian()`, whose reverse pass builds adjoints without the const and so tried to initialize a `double*` from a `const double*`, or bind a `double&` to a temporary.

Leave such declarations without a tangent instead, matching what reverse mode already does, so uses of them derive to a plain zero. Reverse mode gets the same lvalue check for reference adjoints, which fixes `const double& r = 5.;` breaking `clad::gradient()` on its own.

Fixes #1955

### DIFF
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -1609,17 +1609,28 @@ BaseForwardModeVisitor::DifferentiateVarDecl(const VarDecl* VD,
                                   initDiff.getExpr(), VD->isDirectInit());
   // FIXME: Create unique identifier for derivative.
   Expr* initDx = initDiff.getExpr_dx();
-  if (VD->getType()->isPointerType() && !initDx) {
+  QualType VDType = VD->getType();
+  // A reference tangent needs an lvalue to bind to; the literal 0 that e.g.
+  // `const double& r = 5.;` derives to is not one.
+  if (VDType->isLValueReferenceType() && initDx && !initDx->isLValue())
+    initDx = nullptr;
+  // References and pointers-to-const only alias storage they do not own, so
+  // without an initializer tangent there is nothing to alias. Fabricating one
+  // yields a null tangent that later gets dereferenced, or a reference bound
+  // to a temporary; leave them without a tangent and let uses derive to zero,
+  // as reverse mode does in ReverseModeVisitor::DifferentiateVarDecl.
+  bool isAlias =
+      VDType->isLValueReferenceType() ||
+      (VDType->isPointerType() && VDType->getPointeeType().isConstQualified());
+  if (VDType->isPointerType() && !initDx && !isAlias) {
     // initialize with nullptr.
-    // NOLINTBEGIN(cppcoreguidelines-owned-memory)
-    initDx =
-        new (m_Context) CXXNullPtrLiteralExpr(VD->getType(), VD->getBeginLoc());
-    // NOLINTEND(cppcoreguidelines-owned-memory)
+    // NOLINTNEXTLINE(cppcoreguidelines-owning-memory)
+    initDx = new (m_Context) CXXNullPtrLiteralExpr(VDType, VD->getBeginLoc());
   }
   VarDecl* VDDerived = nullptr;
-  if (m_DiffReq.shouldHaveAdjointForw(VD))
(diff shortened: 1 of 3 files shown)


vgvassilev (approved):
LGTM!

PR #1959, opened 2026-08-11
Zero-initialize loop counters in functions with early returns

### LINKED ISSUE #1958: `clad::hessian`: reverse sweep of a loop runs on an early-return path that never entered it, popping an empty tape
## Summary

When a function returns early on a path that never reaches a later loop, `clad::hessian` still runs that loop's reverse sweep. The forward sweep never pushed anything and never initialised the loop counter, so the reverse sweep iterates a garbage number of times and pops empty `clad::tape`s.

`clad::gradient` on the same function is fine — only the forward-over-reverse composition (`*_pushforward_pullback`) is affected.

The bad read only segfaults once the code is optimised. With the cling JIT at `.O 0` the uninitialised counter happens to be zero, the reverse loop is skipped and the Hessian comes out numerically correct; at `.O 1` and above it crashes.

## Reproducer

```cpp
// cladHessianEarlyReturnLoop.C
#include <iostream>

// `n` is only known at run time as far as the differentiated code is concerned.
inline double poly(double *coefs, int n)
{
   int degree = n - 1;

   if (degree == 2)                                 // taken
      return coefs[2] + coefs[1] + coefs[0];

   double t = 1.0;                                  // never reached for n == 3
   double result = coefs[0];
   for (int i = 1; i < degree; i++) {
      result = result + t * coefs[i];
      t *= coefs[0];
   }
   return result;
}

double f(double *params)
{
   return poly(params, 3);
}

#pragma clad ON
void cladHessianEarlyReturnLoop()
{
   auto h = clad::hessian(f, "params[0:2]");
   std::cout << "hessian requested" << std::endl;

   double params[3] = {0.3, 0.7, 0.2};
   double matrix[9] = {};
   h.execute(params, matrix);
   std::cout << "hessian evaluated" << std::endl;
}
#pragma clad OFF
```

```
root.exe -l -b -q cladHessianEarlyReturnLoop.C            # segfault
root.exe -l -b -q -e ".O 0" cladHessianEarlyReturnLoop.C  # works
```

We could only test this through cling, because in our build clad is statically linked into `libCling` and no standalone plugin is available. We would expect the same source to reproduce when compiled ahead of time at `-O1`, but we have not verified that.

## Backtrace

Obtained with `CLING_DEBUG=1 root.exe -l -b -q -e ".O 1" cladHessianEarlyReturnLoop.C` (debug info *and* optimisation — `CLING_DEBUG=1` alone forces `-O0` and hides the crash):

```
#7  clad::pop<double, 64ul, 1024ul, false, false> (to=...)
        at clad/Differentiator/Differentiator.h:97
#8  poly_pushforward_pullback(...)::{lambda()#1}::operator()() const
        at cladHessianEarlyReturnLoop.C:25
#9  poly_pushforward_pullback (coefs=<optimized out>, n=3, ...)
        at cladHessianEarlyReturnLoop.C:25
#10 f_darg0_0_grad_0_0 (...)
#11 f_hessian_0 (...)
```

## Generated code

From `-Xclang -plugin-arg-clad -Xclang -fdump-derived-fn`. The declaration of the loop counter, at the top of the pullback:

```cpp
inline void poly_pushforward_pullback(double *coefs, int n, /* ... */) {
    bool _cond0 = false;
    int _d_d_i = 0;
    int _d_i = 0;
    unsigned long _t0;                 // <-- never initialised here
    int _d_i0 = 0;
    int i = 0;
    clad::tape<double> _t1 = {};
    clad::tape<double> _t2 = {};
    clad::tape<double> _t3 = {};
    clad::tape<double> _t4 = {};
    /* ... */
    auto _rev0 = [&] {
        {
            for (; _t0; _t0--) {       // <-- iterates a garbage number of times
                i--;
                /* ... clad::pop(_t1) ... clad::pop(_t4) ... */
```

and the end of the same function:

```cpp
    {
        _cond0 = degree == 2;
        if (_cond0) {
            _rev0();                   // <-- reverse sweep of a loop that never ran
            return;
        }
    }
    {
        _d_i = 0;
        _t0 = 0;                       // <-- the only place _t0 is initialised
        for (i = 1; i < degree; i++) {
            _t0++;
            clad::push(_t1, _d_result);
            /* ... */
        }
    }
    /* ... */
    _rev0();
}
```

`_rev0` holds the reverse sweep of the loop *and* of the statements before it, so it legitimately has to run on the early-return path — but on that path `_t0` is uninitialised and all four tapes are empty.

## What changes the outcome

| Variant | Result |
|---|---|
| `clad::hessian`, JIT at `.O 1` / `.O 2` / `.O 3` | segfault in `clad::pop` |
| `clad::hessian`, JIT at `.O 0` | works, values match a numerical Hessian |
| `clad::gradient` on the same function | works |
| loop moved into the `else` of a single-return `if` | works |

The last row is what makes the diagnosis concrete: the problem is not the loop or the tape as such, it is the loop's reverse sweep running on a path where the forward sweep never entered the loop.

## Suggested fix

Emitting `unsigned long _t0 = 0;` instead of `unsigned long _t0;` would make the reverse loop a no-op on the early-return path, which is the correct behaviour here. Splitting the loop's reverse sweep out of `_rev0` so it is not called at all on that path would be the more thorough fix.

## Where we hit it

Two functions in ROOT's RooFit, both shaped the same way — special-cased early returns first, a general loop afterwards:

- `RooFit::Detail::MathFuncs::bernstein`, special cases for degree 0/1/2. A `RooBernstein` with three coefficients has degree 2, so the loop never runs and every Hessian of a likelihood containing one crashes.
- `RooFit::Detail::MathFuncs::binomial`, which returns early for `n < k` and for `k == 0 || n == k`.

guitargeek:
A loop counter is created through GlobalStoreAndRef, which on the non-function-scope path hoists the VarDecl into m_Globals without an initializer and leaves the value in a separate `_t0 = 0;` assignment in the forward sweep, right before the loop.

When the function has early returns, the reverse sweep of every statement -- the loop's included -- lives in the shared [&] lambda invoked from each return path. An exit taken before the loop is reached never runs that forward-sweep reset, yet still calls the lambda, so `for (; _t0; _t0--)` reads an uninitialized counter, iterates a garbage number of times and pops tapes the forward sweep never pushed to.

The bad read only misbehaves once optimized; at -O0 the counter happens to be zero, the reverse loop is skipped and the result is correct.

GlobalStoreAndRef already takes a zeroInit flag for exactly this, added for branch _cond flags; the counter simply was not passing it. Pass it when the function has early returns, so the counter reads zero on a path that never reached the loop and its reverse sweep becomes a no-op -- the same zero state clad already relies on for the adjoints and tapes that reverse sweep would otherwise consume.

Reported against clad::hessian, where the forward-over-reverse composition puts the loop in a *_pushforward_pullback that takes this hoisted-decl path; a plain gradient of the same source is unaffected because there the counter's decl keeps its initializer.

Fixes #1958

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -4500,9 +4500,13 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
   ReverseModeVisitor::LoopCounter::LoopCounter(ReverseModeVisitor& RMV)
       : m_RMV(RMV) {
     ASTContext& C = m_RMV.m_Context;
-    m_Ref = m_RMV.GlobalStoreAndRef(m_RMV.getZeroInit(C.IntTy),
-                                    clad_compat::getSizeType(C), "_t",
-                                    /*force=*/true);
+    // The counter's reset lives in the forward sweep, which an early return
+    // taken before the loop never reaches -- while the master reverse sweep
+    // still runs. Zero-init makes the reverse loop a no-op there.
+    m_Ref = m_RMV.GlobalStoreAndRef(
+        m_RMV.getZeroInit(C.IntTy), clad_compat::getSizeType(C), "_t",
+        /*force=*/true,
+        /*zeroInit=*/m_RMV.m_DiffReq.hasEarlyReturns());
   }
 
   StmtDiff ReverseModeVisitor::VisitWhileStmt(const WhileStmt* WS) {
--- a/test/Hessian/EarlyReturns.C
+++ b/test/Hessian/EarlyReturns.C
@@ -1,0 +1,61 @@
+// RUN: %cladclang %s -I%S/../../include -oEarlyReturns.out 2>&1 | %filecheck %s
+// RUN: ./EarlyReturns.out | %filecheck_exec %s
+// RUN: %cladclang -Xclang -plugin-arg-clad -Xclang -disable-tbr %s -I%S/../../include -oEarlyReturns.out
+// RUN: ./EarlyReturns.out | %filecheck_exec %s
+
+#include "clad/Differentiator/Differentiator.h"
+
+#include "../TestUtils.h"
+
+double mat3[9];
+clad::array_ref<double> mat3_ref(mat3, 9);
+double mat4[16];
+clad::array_ref<double> mat4_ref(mat4, 16);
+
+// For degree == 2 the early return is taken and the loop is never entered,
+// so the forward sweep neither pushes onto the loop's tapes nor resets its
+// counter. The master reverse sweep still runs on that path, so the counter
+// has to read zero there -- otherwise the reverse loop iterates a garbage
+// number of times and pops empty tapes (vgvassilev/clad#1985).
+inline double poly(double* coefs, int n) {
+  int degree = n - 1;
+
+  if (degree == 2)
+    return coefs[2] * coefs[1] + coefs[1] * coefs[0];
+
+  double t = 1.;
+  double result = coefs[0];
+  for (int i = 1; i < degree; i++) {
+    result = result + t * coefs[i];
+    t *= coefs[0];
+  }
+  return result;
+}
+
+// Takes the early return: c2 * c1 + c1 * c0.
+double earlyBeforeLoop(double* c) { return poly(c, 3); }
+
+// Falls through to the loop: c0 + c1 + c0 * c2.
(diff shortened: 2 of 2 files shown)


vgvassilev:
To reproduce locally in the failure you can use these: https://github.com/compiler-research/ci-workflows/blob/main/docs/developer-guide.md

vgvassilev (approved):
Lgtm!

PR #1961, opened 2026-08-11
Keep member calls with active args differentiable on a non-tangent base

### LINKED ISSUE #1960: Forward mode drops the derivative of a member call whose base object has no tangent
## Summary

When forward mode differentiates a member or operator call, it now bails out with a zero derivative as soon as the *base object* has no tangent — regardless of whether the call *arguments* are active:

```cpp
struct Functor { double operator()(double x) const { return x * x; } };
Functor gFunc;                                // read-only, no tangent
double helper(double x) { return gFunc(x); }  // ... but x is active

clad::differentiate(helper, "x").execute(3.); // gives 0, expected 6
```

`gFunc` really has no tangent, but `gFunc(x)` still depends on `x`, so the call does contribute to the directional derivative. Clad even generates the correct `operator_call_pushforward` — it just never calls it.

This has three consequences, in increasing order of visibility:

1. `clad::differentiate` silently returns a wrong (zero) derivative, and so does `clad::gradient` whenever it routes a scalar callee through its pushforward. `clad::hessian` silently produces wrong rows.
2. If the call sits one level below the function being differentiated, `clad::hessian` generates *ill-formed* code: the pullback of that pushforward is called with the `_d_this` argument missing.
3. If the base is reached through a `reinterpret_cast` — the shape a JIT-compiled wrapper produces when it refers to an object living in the host process — clad either emits "cannot take the address of an rvalue" or segfaults outright.

Symptom 2 is what makes ROOT's `testRooFuncWrapper` fail when RooFit asks for a Hessian of a likelihood that contains a `RooFunctorBinding`.

## Versions

* clad `master` at `743f5abe`, and the same behaviour on top of it at `2c6ed32d`.
* Introduced by c18a748a *"Yield a zero derivative when a forward-mode call base has no tangent"* (2026-07-16). **clad v2.4 is not affected.**
* clang/LLVM 22.1, run through ROOT's cling; nothing in the reproducers is ROOT-specific, so each file works both as `clang++ -std=c++17 -fplugin=libclad.so file.cpp` and as `root -l -b -q file.cpp`.

## Reproducer 1 — silently wrong derivatives

```cpp
#include <cstdio>

struct Functor {
   double operator()(double x) const { return x * x; }
};

// gFunc is only read, never differentiated with respect to, so it has no tangent.
Functor gFunc;

double helper(double x) { return gFunc(x); }
double func(double *p) { return gFunc(p[0]) * p[1]; }

int main()
{
   printf("d/dx helper(3)     = %g   (expected 6)\n", clad::differentiate(helper, "x").execute(3.));

   double p[2] = {3., 5.};
   printf("d/dp0 func(3, 5)   = %g   (expected 30)\n", clad::differentiate(func, "p[0]").execute(p));

   double g[2] = {0., 0.};
   clad::gradient(func, "p").execute(p, g);
   printf("grad func(3, 5)    = %g %g   (expected 30 9)\n", g[0], g[1]);

   double h[4] = {0., 0., 0., 0.};
   clad::hessian(func, "p[0:1]").execute(p, h);
   printf("hessian func(3, 5) = %g %g %g %g   (expected 10 6 6 0)\n", h[0], h[1], h[2], h[3]);
}
```

Output:

```
d/dx helper(3)     = 0   (expected 6)
d/dp0 func(3, 5)   = 0   (expected 30)
grad func(3, 5)    = 30 9   (expected 30 9)
hessian func(3, 5) = 0 0 6 0   (expected 10 6 6 0)
```

Reverse mode gets it right — it fabricates a zero adjoint `Functor _d_gFunc = {};` for the read-only global and calls `operator_call_pullback` normally. Only the forward pass is wrong, which is why the Hessian loses exactly the rows that come from a pushforward.

Providing a custom derivative does not help: with

```cpp
namespace clad::custom_derivatives::class_functions {
clad::ValueAndPushforward<double, double>
operator_call_pushforward(const Functor *f, double x, const Functor *, double d_x)
{
   return {(*f)(x), 2 * x * d_x};
}
}
```

in scope, `clad::differentiate(helper, "x")` still returns 0 — the bail-out happens before any custom pushforward is looked up.

## Reproducer 2 — `clad::hessian` generates ill-formed code

Same code, but with the member call one level below the differentiated function:

```cpp

struct Functor {
   double operator()(double x) const { return x * x; }
};

Functor gFunc;

double helper(double x) { return gFunc(x); }
double func(double *p) { return helper(p[0]) * p[1]; }

int main()
{
   auto h = clad::hessian(func, "p[0:1]");
}
```

```
error: too few arguments to function call, expected 4, have 3
double helper(double x) { return gFunc(x); }
       ^                         ~~~~~
note: 'operator_call_pullback' declared here
```

`-fdump-derived-fn` shows both halves of the problem — the unused pushforward and the zero tangent:

```cpp
clad::ValueAndPushforward<double, double>
operator_call_pushforward(double x, const Functor *_d_this, double _d_x) const {
    return {x * x, _d_x * x + x * _d_x};   // correct, but never called
}
clad::ValueAndPushforward<double, double> helper_pushforward(double x, double _d_x) {
    return {gFunc(x), 0.};                 // tangent hardcoded to 0
}
```

The reverse pass over `helper_pushforward` then has to pull back the `gFunc(x)` call that was left in the primal. `operator_call_pullback` is declared as

```cpp
void operator_call_pullback(double x, double _d_y, Functor *_d_this, double *_d_x) const;
```

but the emitted call passes only three arguments — the adjoint of the base is dropped, because at that point there is none.

## Reproducer 3 — the shape that JIT-generated code produces

RooFit's code generation refers to a host-process object by casting its address, and this reproduces ROOT's `testRooFuncWrapper` failure verbatim:

```cpp

struct Functor {
   double operator()(double x) const { return x * x; }
};

double helper(double x) { return reinterpret_cast<Functor const *>(0x1234)->operator()(x); }
double func(double *p) { return helper(p[0]) * p[1]; }

int main()
{
   auto h = clad::hessian(func, "p[0:1]");
}
```

```
error: cannot take the address of an rvalue of type 'int'
error: too few arguments to function call, expected 8, have 7
note: 'operator_call_pushforward_pullback' declared here
```

## Reproducer 4 — segfault when the cast operand is an object address

```cpp

struct Functor {
   double operator()(double x) const { return x * x; }
};

Functor gFunc;

double helper(double x) { return reinterpret_cast<Functor const *>(&gFunc)->operator()(x); }

int main()
{
   auto d = clad::differentiate(helper, "x");
}
```

```
error: cannot take the address of an rvalue of type 'void'
```

```
#7  clang::Sema::BuildCXXNamedCast(...)
#8  clad::BaseForwardModeVisitor::VisitCXXNamedCastExpr (BaseForwardModeVisitor.cpp:1864)
#11 clad::BaseForwardModeVisitor::VisitCallExpr (BaseForwardModeVisitor.cpp:1090)
```

`&gFunc` has no tangent, so `VisitUnaryOperator`'s `UO_AddrOf` branch builds
`&<void rvalue>`, which fails and yields a null `Expr*`; `VisitCXXNamedCastExpr` then
feeds that null into `Sema::BuildCXXNamedCast`:

```cpp
  Expr* castExprDiff =
      m_Sema
          .BuildCXXNamedCast(Loc, CastKind, TSI, subExprDiff.getExpr_dx(),
                             Brackets, Range)
          .get();
```

(This last one may deserve its own issue — happy to split it out.)

## Where it comes from

`lib/Differentiator/BaseForwardModeVisitor.cpp`, `VisitCallExpr` (master, around
line 1080):

```cpp
      baseDiff = Visit(baseOriginalE);
      Expr* baseDerivative = baseDiff.getExpr_dx();
      // A base that does not depend on the differentiation variable has no
      // tangent, hence no pushforward to call, and contributes nothing to the
      // directional derivative.
      if (!baseDerivative || baseDerivative->getType()->isVoidType())
        return StmtDiff(Clone(CE),
                        getZeroInit(CE->getType().getNonReferenceType()));
```

The premise in the comment only holds when the call's *arguments* are non-active as well, which is the case in the regression test added with that commit (`test/ForwardMode/NonActiveCallBase.cpp` calls `g(i)` with an `int` index). As soon as an argument is active, the entire call — arguments included — is silently zeroed.

## Suggested direction

* Keep the bail-out only when no call argument has a tangent either. Otherwise continue as before and pass a zero/absent tangent for the base, mirroring what reverse mode already does when it synthesizes `Functor _d_gFunc = {};` for a read-only global.
* Whatever placeholder the base tangent gets, the pushforward and pullback *calls* must keep the `_d_this` slot filled so the emitted call matches the declaration — this is what makes `clad::hessian` fail to compile in reproducer 2 and 3.
* Guard the null `Expr*` in `VisitCXXNamedCastExpr` so a base without a tangent cannot crash the plugin.

guitargeek:
Since c18a748a, forward mode yields a zero derivative as soon as the base object of a member/operator call has no tangent, before the call arguments are even visited. That is only correct when no argument is active either: gFunc(x) with a read-only global gFunc still depends on x, so the derivative was silently zeroed, hessians lost the rows routed through a pushforward, and no custom pushforward was ever looked up (issue #1960).

Defer the bail-out until the arguments have been visited. If none of them has a tangent, keep returning zero. Otherwise call the pushforward with a zero-initialized placeholder filling the _d_this slot, mirroring the zero adjoint reverse mode synthesizes for a read-only base.

The reverse pass over such a pushforward hit the second half of the bug: when the base has no adjoint -- as in hessians, where global-adjoint discovery does not run over the clad-generated pushforward -- the adjoint-of-this argument was dropped from the pullback call, emitting ill-formed code ("too few arguments to function call") and skewing the CallArgDx indexing used to classify differentiable params. Synthesize the same kind of zero placeholder there, and treat a non-addressable zero literal (the int 0 a cast pointer base differentiates to) as an absent adjoint instead of taking its address.

Also guard the forward-mode UO_AddrOf and CXXNamedCastExpr visitors against subexpressions without a tangent: they used to build &<void rvalue> and feed a null Expr* into Sema::BuildCXXNamedCast, which emitted "cannot take the address of an rvalue" and could segfault on
reinterpret_cast<T*>(&global)->call(x).

Fixes #1960.

### DIFF
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -1077,15 +1077,44 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
         baseOriginalE = OCE->getArg(0);
       baseDiff = Visit(baseOriginalE);
       Expr* baseDerivative = baseDiff.getExpr_dx();
-      // A base that does not depend on the differentiation variable has no
-      // tangent, hence no pushforward to call, and contributes nothing to the
-      // directional derivative.
-      if (!baseDerivative || baseDerivative->getType()->isVoidType())
-        return StmtDiff(Clone(CE),
-                        getZeroInit(CE->getType().getNonReferenceType()));
-      if (!baseDerivative->getType()->isPointerType())
-        baseDerivative = BuildOp(UnaryOperatorKind::UO_AddrOf, baseDerivative);
-      diffArgs.push_back(baseDerivative);
+      if (!baseDerivative || baseDerivative->getType()->isVoidType()) {
+        // The base does not depend on the differentiation variable and has no
+        // tangent. The call may still contribute to the directional derivative
+        // through its arguments: consult the activity information of the
+        // DiffRequest to find out. If no argument is varied either, the call
+        // contributes nothing.
+        bool anyVariedArg = false;
+        for (unsigned i = isa<CXXOperatorCallExpr>(CE) ? 1 : 0,
+                      e = CE->getNumArgs();
+             i < e; ++i) {
+          const Expr* arg = CE->getArg(i);
+          if (utils::IsDifferentiableType(arg->getType()) &&
+              m_DiffReq.isVaried(arg)) {
+            anyVariedArg = true;
+            break;
+          }
+        }
+        if (!anyVariedArg)
+          return StmtDiff(Clone(CE),
+                          getZeroInit(CE->getType().getNonReferenceType()));
+        // The arguments drive the derivative through the pushforward, whose
+        // `_d_this` slot must stay filled: pass a zero-initialized
+        // placeholder, mirroring the zero adjoint reverse mode synthesizes
+        // for such a base.
+        QualType baseTy = baseOriginalE->getType();
+        if (baseTy->isPointerType())
+          baseTy = baseTy->getPointeeType();
+        baseTy = baseTy.getNonReferenceType().getUnqualifiedType();
+        VarDecl* dBaseVD = BuildVarDecl(baseTy, "_d_base", getZeroInit(baseTy));
+        addToCurrentBlock(BuildDeclStmt(dBaseVD));
+        diffArgs.push_back(
+            BuildOp(UnaryOperatorKind::UO_AddrOf, BuildDeclRef(dBaseVD)));
+      } else {
+        if (!baseDerivative->getType()->isPointerType())
+          baseDerivative =
+              BuildOp(UnaryOperatorKind::UO_AddrOf, baseDerivative);
+        diffArgs.push_back(baseDerivative);
+      }
     }
   }
 
@@ -1390,8 +1419,13 @@ StmtDiff BaseForwardModeVisitor::VisitUnaryOperator(const UnaryOperator* UnOp) {
         op, ConstantFolder::synthesizeLiteral(literalTy, m_Context, /*val=*/0));
   } else if (opKind == UnaryOperatorKind::UO_AddrOf) {
     Expr* derivedOp = diff.getExpr_dx();
-    if (derivedOp)
-      derivedOp = BuildOp(opKind, diff.getExpr_dx());
+    // A subexpression without a tangent of its own (e.g. an object that does
+    // not depend on the differentiation variable) yields a synthesized zero
+    // init, which is an rvalue and has no address.
+    if (derivedOp && !derivedOp->getType()->isVoidType())
+      derivedOp = BuildOp(opKind, derivedOp);
+    else
+      derivedOp = nullptr;
     return StmtDiff(op, derivedOp);
   } else if (opKind == UnaryOperatorKind::UO_LNot) {
     Expr* zero = getZeroInit(UnOp->getType());
@@ -1844,11 +1878,14 @@ BaseForwardModeVisitor::VisitCXXNamedCastExpr(const CXXNamedCastExpr* NCE) {
           .BuildCXXNamedCast(Loc, CastKind, TSI, subExprDiff.getExpr(),
                              Brackets, Range)
           .get();
-  Expr* castExprDiff =
-      m_Sema
-          .BuildCXXNamedCast(Loc, CastKind, TSI, subExprDiff.getExpr_dx(),
-                             Brackets, Range)
-          .get();
+  // A subexpression without a tangent (e.g. the address of an object that
+  // does not depend on the differentiation variable) has nothing to cast.
+  Expr* subExprDx = subExprDiff.getExpr_dx();
+  Expr* castExprDiff = nullptr;
+  if (subExprDx && !subExprDx->getType()->isVoidType())
+    castExprDiff =
+        m_Sema.BuildCXXNamedCast(Loc, CastKind, TSI, subExprDx, Brackets, Range)
+            .get();
   return StmtDiff(castExpr, castExprDiff);
 }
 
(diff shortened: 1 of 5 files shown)


vgvassilev:
Can we add a tight test hitting the uncovered branches?

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1243,6 +1254,27 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
     }
   }
 
+  if (!missingBaseTangentTy.isNull()) {
+    // The base has no tangent. If no call argument has one either, the call
```
vgvassilev: Shouldn’t that analysis go in the DiffRequest information? That seems the right place to me.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1063,6 +1063,9 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
   // Represents `StmtDiff` result of the 'base' object if are differentiating
   // a direct or indirect (operator overload) call to member function.
   StmtDiff baseDiff;
+  // Type of the base object when it has no tangent of its own, so that a
+  // zero-initialized placeholder can be synthesized for the `_d_this` slot.
+  QualType missingBaseTangentTy;
```
vgvassilev: Hi @guitargeek, this was the state I hoped to move to the diff planning part if possible.

guitargeek: Done in the latest push!

vgvassilev (approved):
Lgtm!

PR #1962, opened 2026-08-13
Hoist reverse_forw scaffolding to function scope

guitargeek:
Three fixes for the same bug shape: a variable that a `reverse_forw` call declares in the current forward block, while the matching statement in the reverse sweep lives in a different block. When the call sits inside a branch or a loop, the reverse-sweep reference names a variable that is out of scope there, which trips the use-before-declaration integrity check (assert in debug, warning in release) and produces uncompilable derivatives.

The three declarations, one commit each:

- **`GlobalStoreAndRef` in `reverse_mode_forward_pass`** assumed the current block is the function body and merged declaration with initializer. It isn't: a short-circuit condition builds a nested `if (L) _cond = R` scaffold, and the condition rebuilt from the stored flag is placed in the enclosing block. Take the shortcut only when the forward block stack is actually at the function body. Found via the `reverse_forw` of SOFIE's `Gemm_Call` pushforward.
- **The restore tracker** (`_tracker0`) is now declared at function scope, with a `clear()` at the former declaration point to preserve the per-visit semantics. Found via the pullback of SOFIE's `doInfer` helper.
- **The `ValueAndAdjoint` result store** is hoisted the same way, as an unset declaration plus an assignment at the call point. Results holding references cannot be declared unset, so for those we keep the struct store block-local and hoist pointers to the referents instead — member access through a reference field yields the referent, which outlives the block, and both sweeps go through the pointers.

New coverage in `test/Gradient/ReverseForw.C` for each case; the existing expectations there shift where the hoists changed the printed output.

Two limitations are left as FIXMEs in the code: a single hoisted tracker still restores only the last iteration's state for a reversed loop (unchanged from before this PR; full generality needs per-iteration taping), and a `reverse_forw` result mixing reference and non-reference members — only producible by a custom `reverse_forw` — keeps the block-local form, since a pointer into that struct would dangle.

### DIFF
--- a/include/clad/Differentiator/RestoreTracker.h
+++ b/include/clad/Differentiator/RestoreTracker.h
@@ -1,6 +1,10 @@
 #ifndef CLAD_DIFFERENTIATOR_RESTORETRACKER_H
 #define CLAD_DIFFERENTIATOR_RESTORETRACKER_H
 
+// CUDA_HOST_DEVICE is only used under __CUDACC__, so include-cleaner cannot
+// see the use in a non-CUDA parse.
+#include "clad/Differentiator/CladConfig.h" // IWYU pragma: keep
+
 #include <cassert>
 #include <cstdint>
 #include <cstring>
@@ -38,9 +42,9 @@ class restore_tracker {
   size_t m_cnt = 0, m_off = 0;
 
 public:
-  __host__ __device__ restore_tracker() = default;
+  CUDA_HOST_DEVICE restore_tracker() = default;
 
-  template <typename T> __host__ __device__ void store(const T& val) {
+  template <typename T> CUDA_HOST_DEVICE void store(const T& val) {
     for (size_t i = 0; i < m_cnt; ++i)
       if (m_meta[i].addr == (char*)&val)
         return;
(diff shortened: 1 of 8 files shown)


vgvassilev:
For the sake of history, maybe we should split that into multiple atomic PRs?

guitargeek:
I'll do that. I just wanted to fast-forward to a state that works for RooFit and in the Clad CI, and then I'll split off PRs from there

vgvassilev:
> I'll do that. I just wanted to fast-forward to a state that works for RooFit and in the Clad CI, and then I'll split off PRs from there

Makes sense!

include/clad/Differentiator/RestoreTracker.h
```diff
@@ -61,6 +61,10 @@ class restore_tracker {
       std::memcpy(m_meta[i].addr, m_buf + m_meta[i].off, m_meta[i].size);
     m_cnt = m_off = 0;
   }
+
+  // Drop all records without writing anything back. Emitted where a
+  // block-local tracker declaration used to (re-)initialize the tracker.
+  __host__ __device__ void clear() { m_cnt = m_off = 0; }
```
vgvassilev: We had a macro for these things -- I can't remember why we are not using it here. Maybe it was an oversight.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2800,25 +2807,73 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+        // block than the forward one, so a block-local store would leave the
+        // reverse-sweep reference out of scope; a function with early
+        // returns wraps the reverse sweep in lambdas, which must not capture
+        // block-locals either. Hoist the store to function scope in both
+        // cases. A result whose members are references cannot be hoisted
+        // whole -- a reference member cannot be declared unset and assigned
+        // afterwards -- so store it block-locally and hoist pointers to the
+        // referents, which outlive the block; both sweeps then go through
+        // those. (With early returns GlobalStoreAndRef tapes ValueAndAdjoint
+        // stores instead, which also handles reference members.)
```
vgvassilev: Can we use simpler english (shorter sentences) to improve this comment and possibly tighten it?

vgvassilev (approved):
LGTM!

PR #1963, opened 2026-08-14
Add support for WebAssembly build of clad

aaronj0:
Emscripten builds are static-only (emscripten LLVM has LLVM_ENABLE_PLUGINS=OFF), so there is no clad
MODULE to drive the lit suite or the plugin-compiled unittests and the test tree was skipped entirely.
The Basic unittests exercise only the headers, so they now cross-compile and run under node; Misc and
Kokkos stay native-only.

Adds a ubuntu-24.04 job that builds clad against a prebuilt emscripten LLVM 22 (the llvm-wasm artifact
CppInterOp's emscripten CI uses, emsdk 4.0.9) with LLVM_ENABLE_WERROR=On and runs the Basic unittests
under node. Also verified against an emscripten-built LLVM 20.1.8 locally; the native suites are
unaffected (LLVM 21).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -607,3 +607,40 @@ jobs:
         name: observes-change-summary
         path: observes-change-summary.json
         if-no-files-found: ignore
+
+  build-wasm:
+    timeout-minutes: 30
+    name: ubu24-emscr-runtime22
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v7
+
+    - name: Setup emsdk
+      uses: emscripten-core/setup-emsdk@v16
+      with:
+        # The emsdk pin must match the llvm-wasm cell.
+        version: '4.0.9'
+
+    - name: Setup LLVM/Clang/LLD built for wasm
+      uses: compiler-research/ci-workflows/actions/setup-recipe@main
+      with:
+        recipe: llvm-wasm
+        version: '22'
+        os: ubuntu-24.04
+        arch: x86_64
+
+    - name: Build Clad for wasm
+      run: |
+        LLVM_BUILD_DIR="$(pwd)/install/build"
+        mkdir obj && cd obj
+        emcmake cmake -DCMAKE_BUILD_TYPE=Release \
+          -DLLVM_DIR=$LLVM_BUILD_DIR/lib/cmake/llvm \
+          -DClang_DIR=$LLVM_BUILD_DIR/lib/cmake/clang \
+          -DLLD_DIR=$LLVM_BUILD_DIR/lib/cmake/lld \
+          -DCLAD_BUILD_STATIC_ONLY=ON \
+          -DLLVM_ENABLE_WERROR=On \
+          $GITHUB_WORKSPACE
+        emmake make -j$(nproc)
+
+    - name: Run the unittests under node
+      run: emmake make -j$(nproc) -C obj check-clad
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -309,6 +309,12 @@ add_definitions( -D_GNU_SOURCE
   -DCLAD_INSTDIR_INCL="${CLAD_BINARY_DIR}/include" )
 
 option(CLAD_BUILD_STATIC_ONLY "Does not build shared libraries. Useful when we have LLVM_ENABLE_PLUGINS=OFF (eg. Win or CYGWIN)" OFF)
+
+if (EMSCRIPTEN)
+  # For the test executables and googletest only; the LLVM-style archive
+  # targets append -fno-exceptions, which emcc >= 4 rejects alongside wasm-EH.
+  set(CLAD_EXTRA_WASM_FLAGS "-fwasm-exceptions" CACHE STRING "Extra flags for wasm")
+endif()
 if (NOT CLAD_BUILD_STATIC_ONLY AND NOT LLVM_ENABLE_PLUGINS)
   message(FATAL_ERROR "LLVM_ENABLE_PLUGINS is set to OFF. Please build clad with -DCLAD_BUILD_STATIC_ONLY=ON.")
 endif()
@@ -378,6 +384,8 @@ if (NOT CLAD_BUILD_STATIC_ONLY)
     set(CMAKE_CXX_FLAGS ${stored_cxx_flags})
     set(CMAKE_C_FLAGS ${stored_c_flags})
   endif()
+elseif (EMSCRIPTEN AND NOT CLAD_DISABLE_TESTS)
+  add_subdirectory(unittests)
 endif()
 
 # Workaround for MSVS10 to avoid the Dialog Hell
--- a/cmake/modules/AddClad.cmake
+++ b/cmake/modules/AddClad.cmake
@@ -26,36 +26,40 @@ function(ENABLE_CLAD_FOR_TARGET executable)
     message(FATAL_ERROR "'${executable}' is not a valid target.")
   endif()
 
-  # Add the clad plugin
-  target_compile_options(${executable} PUBLIC -fplugin=$<TARGET_FILE:clad>)
+  if (TARGET clad)
+    # Add the clad plugin
+    target_compile_options(${executable} PUBLIC -fplugin=$<TARGET_FILE:clad>)
 
-  # Debugging. Emitting the derivatives' source code.
-  #target_compile_options(${executable} PUBLIC "SHELL:-Xclang -plugin-arg-clad"
-  #  "SHELL: -Xclang -fdump-derived-fn")
+    # Debugging. Emitting the derivatives' source code.
+    #target_compile_options(${executable} PUBLIC "SHELL:-Xclang -plugin-arg-clad"
+    #  "SHELL: -Xclang -fdump-derived-fn")
 
-  # Debugging. Emit llvm IR.
-  #target_compile_options(${executable} PUBLIC -S -emit-llvm)
+    # Debugging. Emit llvm IR.
+    #target_compile_options(${executable} PUBLIC -S -emit-llvm)
 
-  # Debugging. Optimization misses.
-  #target_compile_options(${executable} PUBLIC "SHELL:-Xclang -Rpass-missed=.*inline.*")
+    # Debugging. Optimization misses.
+    #target_compile_options(${executable} PUBLIC "SHELL:-Xclang -Rpass-missed=.*inline.*")
+
+    add_dependencies(${executable} clad)
+
+    # If clad.so changes we don't need to relink but to rebuild the source files.
+    # $<TARGET_FILE:clad> does not work for OBJECT_DEPENDS.
+    set (CLAD_SO_PATH "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/clad${CMAKE_SHARED_LIBRARY_SUFFIX}")
+    set_source_files_properties(${source_files} PROPERTY OBJECT_DEPENDS ${CLAD_SO_PATH})
+  endif()
 
   # Clad requires us to link against these libraries.
-  target_link_libraries(${executable} PUBLIC stdc++ pthread m)
+  if (NOT EMSCRIPTEN)
+    target_link_libraries(${executable} PUBLIC stdc++ pthread m)
+  endif()
 
   target_include_directories(${executable} PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
   set_property(TARGET ${executable} PROPERTY RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
-
-  add_dependencies(${executable} clad)
 
   if (NOT CLAD_BUILT_STANDALONE)
     # We are probably building clad with clang. Make sure we've built clang.
     add_dependencies(${executable} clang)
   endif()
-
-  # If clad.so changes we don't need to relink but to rebuild the source files.
-  # $<TARGET_FILE:clad> does not work for OBJECT_DEPENDS.
-  set (CLAD_SO_PATH "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/clad${CMAKE_SHARED_LIBRARY_SUFFIX}")
-  set_source_files_properties(${source_files} PROPERTY OBJECT_DEPENDS ${CLAD_SO_PATH})
 
   # Add dependencies to executable
   if(ARG_DEPENDS)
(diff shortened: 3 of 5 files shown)


aaronj0:
Results of the WASM run can be seen here: https://github.com/vgvassilev/clad/actions/runs/31791210431/job/94738307180?pr=1963

.github/workflows/ci.yml
```diff
@@ -549,3 +549,53 @@ jobs:
...
+  # unittests under node. CLAD_BUILD_STATIC_ONLY is mandatory here: the
+  # emscripten LLVM has LLVM_ENABLE_PLUGINS=OFF, so there is no clad MODULE
+  # to load into a host compiler and the lit suite cannot run; the Basic
+  # unittests are the executable coverage (see unittests/CMakeLists.txt).
+  build-wasm:
+    timeout-minutes: 120
+    name: ubu24-emscr-runtime22
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v4
```
mcbarton:
```suggestion
    - uses: actions/checkout@v7
```
Just a suggestion, ai generated cis tend to use outdated actions, so something to watch out for in the future. It can slo be good for stability to pin it to an individual commit (this is llvms policy) and in this case it would be @3d3c42e5aac5ba805825da76410c181273ba90b1

.github/workflows/ci.yml
```diff
@@ -549,3 +549,53 @@ jobs:
...
+
+    - name: Setup emsdk
+      run: |
+        # Matches the emsdk the llvm-wasm cell was built with
+        # (ci-workflows recipes/llvm-wasm/build.py).
+        git clone --depth=1 https://github.com/emscripten-core/emsdk.git
+        ./emsdk/emsdk install 4.0.9
+        ./emsdk/emsdk activate 4.0.9
```
mcbarton: Might be worth changing to Emscriptens new `setup-emsdk` Github action. That way you can drop all the  `source ./emsdk/emsdk_env.sh` lines

.github/workflows/ci.yml
```diff
@@ -549,3 +549,53 @@ jobs:
...
+          -DCLAD_BUILD_STATIC_ONLY=ON \
+          -DLLVM_ENABLE_WERROR=On \
+          $GITHUB_WORKSPACE
+        emmake make -j$(nproc)
+        emmake make -j$(nproc) CladUnitTests
+
+    - name: Run the unittests under node
+      run: |
+        source ./emsdk/emsdk_env.sh
+        $EMSDK_NODE obj/unittests/Basic/DifferentiatorHTests.js
```
mcbarton: Just curious as to whether this is currently only expected to run one test? Asking this since that is what the ci displays (this may be what you are expecting)
```
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from DifferentiatorH
[ RUN      ] DifferentiatorH.GetLength
[       OK ] DifferentiatorH.GetLength (0 ms)
[----------] 1 test from DifferentiatorH (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (1 ms total)
[  PASSED  ] 1 test.
```

mcbarton: Using something similar to CppInterOp you can make it so the check-clad target automatically used emsdks node (see https://github.com/compiler-research/CppInterOp/blob/8d624c621a4b95e36ff73ac708c85a768287478f/unittests/CMakeLists.txt#L80)

.github/workflows/ci.yml
```diff
@@ -549,3 +549,53 @@ jobs:
...
+          -DLLD_DIR=$LLVM_BUILD_DIR/lib/cmake/lld \
+          -DCLAD_BUILD_STATIC_ONLY=ON \
+          -DLLVM_ENABLE_WERROR=On \
+          $GITHUB_WORKSPACE
+        emmake make -j$(nproc)
+        emmake make -j$(nproc) CladUnitTests
+
+    - name: Run the unittests under node
+      run: |
+        source ./emsdk/emsdk_env.sh
```
mcbarton: Without `set -e` the ci doesn't pick up failures in tests (CppInterOp for example has it here https://github.com/compiler-research/CppInterOp/blob/8d624c621a4b95e36ff73ac708c85a768287478f/.github/actions/Build_and_Test_CppInterOp/action.yml#L486). You may need this in combination with the `EXIT_RUNTIME` flag https://emscripten.org/docs/tools_reference/settings_reference.html#exit-runtime . CppInterOps has this flag set to 1 implicitly due to the `--emrun` flag which helps it with its browser tests.

AI has a habit of suggesting `set -euo pipefail` but this is overkill.

.github/workflows/ci.yml
```diff
@@ -549,3 +549,53 @@ jobs:
...
+        mkdir obj && cd obj
+        emcmake cmake -DCMAKE_BUILD_TYPE=Release \
+          -DLLVM_DIR=$LLVM_BUILD_DIR/lib/cmake/llvm \
+          -DClang_DIR=$LLVM_BUILD_DIR/lib/cmake/clang \
+          -DLLD_DIR=$LLVM_BUILD_DIR/lib/cmake/lld \
+          -DCLAD_BUILD_STATIC_ONLY=ON \
+          -DLLVM_ENABLE_WERROR=On \
+          $GITHUB_WORKSPACE
+        emmake make -j$(nproc)
+        emmake make -j$(nproc) CladUnitTests
```
mcbarton: Does `CladUnitTests` not imply all the targets needed to run the tests? If yes then there is no need for these 2 `emmake make` commands.

.github/workflows/ci.yml
```diff
@@ -549,3 +549,53 @@ jobs:
...
 
     steps: *build_steps
+
+  # Cross-compile clad for wasm32-unknown-emscripten and run the header
+  # unittests under node. CLAD_BUILD_STATIC_ONLY is mandatory here: the
+  # emscripten LLVM has LLVM_ENABLE_PLUGINS=OFF, so there is no clad MODULE
+  # to load into a host compiler and the lit suite cannot run; the Basic
+  # unittests are the executable coverage (see unittests/CMakeLists.txt).
+  build-wasm:
+    timeout-minutes: 120
```
mcbarton: CppInterOp set this value for the tests through the timeout_value for the tests I believe https://github.com/compiler-research/CppInterOp/blob/8d624c621a4b95e36ff73ac708c85a768287478f/unittests/CMakeLists.txt#L91 . This means it works outside the ci.

mcbarton (commented):
Just some comments from a glancing look at this PR.

.github/workflows/ci.yml
aaronj0: Nice, Thanks!

.github/workflows/ci.yml
aaronj0: Yes, I used this as a sanity test. I'll build the rest as a follow-up

.github/workflows/ci.yml
aaronj0: Looks like the workflow run steps default to `bash -e {0}`, and the default link flags (on my system thats emsdk 4.0.9), node propagates main's return value. So a failing gtest run should fail the job. `EXIT_RUNTIME` only gates atexit/destructor teardown on exit. -Based on my understanding, `--emrun` implies it for the browser flow, which this job doesn't use.

.github/workflows/ci.yml
aaronj0: CladUnitTests links only gtest, so on its own it does no't build the clad archives (you can test this). The first make is what verifies the static libraries cross-compile, which is what a wasm consumer of clad takes

.github/workflows/ci.yml
aaronj0: Done

.github/workflows/ci.yml
aaronj0: Done

.github/workflows/ci.yml
mcbarton: `--emrun` is used in CppInterOp so that the terminal can pick up the standard output and error when running the tests in a browser. It also implies `EXIT_RUNTIME=1`. This setting also happens to help running the test in node (the html run with emrun is just a basic html which loads the test js). node may propagate the exit value, but based on my experience without the `set -e` the ci won't pick up a non zero return value (or at the very least crash because of the non zero value). Best way to see if the ci will pick up a failure is to test it. I suggest you put in a commit where GetLength should fail, and see if the Emscripten ci picks it up.

vgvassilev:
Why we are not use the ci-workflows infrastructure?

.github/workflows/ci.yml
```diff
@@ -549,3 +549,40 @@ jobs:
...
+        # The emsdk pin must match the llvm-wasm cell.
+        version: '4.0.9'
+
+    - name: Setup LLVM/Clang/LLD built for wasm
+      uses: compiler-research/ci-workflows/actions/setup-recipe@main
+      with:
+        recipe: llvm-wasm
+        version: '22'
+        os: ubuntu-24.04
+        arch: x86_64
```
aaronj0, line 630: @vgvassilev I am using the ci-workflows build. See here.

vgvassilev: Ah, ok. @mcbarton, is that PR good to go?

.github/workflows/ci.yml
aaronj0: In the case you mention of CppInterOp we use --emrun's browser flow which isn't used here, so that doesn't apply. Here we use ctest, so this is taken care of regardless: the make command fails as with standard ctest and the step (run steps also default to `bash -e {0}`). Node propagates the test's exit status with the default link flags.

I just pushed a temp commit proving this: https://github.com/vgvassilev/clad/actions/runs/31875243503/job/94989882585?pr=1963), so nothing actually depends on `set -e`
                                                                                            
https://github.com/compiler-research/CppInterOp/pull/830 will actually leverage this and then run the added "CladTest" as a wasm module in the browser, where we actually gain from testing the browser flow, since clad runs inside interpreter's side module there. In that case clad is statically linked into the libclangCppInterOp side module and API calls go through the JIT, so the browser-specific machinery (side module loading, dlopen, --emrun) is exercised. The point of this added support is to prove that clad _cross-compiles_ for WASM and its header-only unittests pass there, and node covers that.

.github/workflows/ci.yml
mcbarton: @vgvassilev Once @aaronj0 has reverted his commit he used to demonstrate to me that the ci can catch node failures we are good to go.

There is the possibility this built library isn't fully compatible with CppInterOp due to the lack of the WASM_BIGINT and an equivalent of the CPPINTEROP_EXTRA_WASM_FLAGS flags (used for pickng the exception model at the moment) here https://github.com/compiler-research/CppInterOp/blob/d00d52364e13533fa0c8ecddec0813e1e22f36bf/lib/CppInterOp/CMakeLists.txt#L245 but that can be an iteration on this PR.

When it comes to have a proper test suite (since this PR currently contains just a smoke test) I suggest the flags here https://github.com/compiler-research/CppInterOp/blob/d00d52364e13533fa0c8ecddec0813e1e22f36bf/unittests/CppInterOp/CMakeLists.txt#L30 are used for the test executable for both  consistency's sake with CppInterOp, xeus-cpp and llvm, and the fact some tests will fail just because of Emscriptens defaults for some of these values

CMakeLists.txt
```diff
@@ -309,6 +309,12 @@ add_definitions( -D_GNU_SOURCE
   -DCLAD_INSTDIR_INCL="${CLAD_BINARY_DIR}/include" )
 
 option(CLAD_BUILD_STATIC_ONLY "Does not build shared libraries. Useful when we have LLVM_ENABLE_PLUGINS=OFF (eg. Win or CYGWIN)" OFF)
+
+if (EMSCRIPTEN)
+  # For the test executables and googletest only; the LLVM-style archive
+  # targets append -fno-exceptions, which emcc >= 4 rejects alongside wasm-EH.
+  set(CLAD_EXTRA_WASM_FLAGS "" CACHE STRING "Extra flags for wasm")
```
mcbarton: `CLAD_EXTRA_WASM_FLAGS` should be `-fwasm-exceptions` to match Cppinterop https://github.com/compiler-research/CppInterOp/blob/75d876881ac92b3d077b60d86cfd2907c5ad2b34/CMakeLists.txt#L21 and what the llvm build should be (see build instructions here https://github.com/compiler-research/CppInterOp/blob/75d876881ac92b3d077b60d86cfd2907c5ad2b34/Emscripten-build-instructions.md?plain=1#L105)

.github/workflows/ci.yml
aaronj0: The CppInterOp WASM module consumes clad's static archives (whole-archive), as mentioned in the PR description, there is no actual shared object or module for clad, so I don't understand the library compatibility concern, since both flags do not actually apply at that level.

`WASM_BIGINT` is a link option, `.a` files carry no link options. CppInterOp applies it when it links `libclangCppInterOp.so`.

The exception model doesn't take effect either: clad is built as an llvm library (as a clang plugin), and llvm appends `-fno-exceptions` after flags are passed in (`LLVM_ENABLE_EH` is off by default and in the llvm-wasm build). no EH objects link into either exception model which is set by the consuming module's link step (CppInterOp already does for `libclangCppInterOp.so`)

In general, we cannot apply the same flags across every project's WASM builds, since they are different kinds of libraries.

CMakeLists.txt
aaronj0: Done, the test executables default to `-fwasm-exceptions` now which will be useful in the future if we expect tests that require exception handling

mcbarton (commented):
@vgvassilev This PR looks ok to me. I suggest you get a second opinion off @anutosh491 though with regards to the flags for the static library.

vgvassilev:
Ok. Let’s wait to see what Anutosh has to say. We can always rely on a post-review feedback if he doesn’t have bandwidth.

anutosh491:
Hey guys, 

I do have a perfectly fine build for clad on emscripten-forge. I remember failing to get it to work with xeus-cpp-lite but for sure the build was fine.

https://github.com/emscripten-forge/recipes/tree/main/recipes/recipes_emscripten/clad

Also I built it as a shared lib as we see that managing side modules have been easier with xeus-cpp being the main module.

I can quickly try to get the above build working with xeus-cpp-lite if that helps. I think it should be close.

vgvassilev:
Does that change the outcome of this PR? Eg. should this PR wait until something else happens?

mcbarton:
Worked out today that my `WASM_BIGINT` comment is a mute point. Although `-sWASM_BIGINT=1` didn't used to be the default for Emscripten, it is now (no idea if it was changed between 3.1.73 and 4.0.9). An issue was also raised in the last 24 hours about `WASM_BIGINT=0` being removed as an option (see https://github.com/emscripten-core/emscripten/issues/27565#issue-5187023583). I think it is ok for this PR to progress as is. Any issue with xeus-cpp using Clad, we'll cross that bridge when we come to it. I don't think its a problem which effects this PR. After this PR is merged in, the next stage would be to prove that the library in this PR can be used with CppInterOp. 

The issue with `LLVM_ENABLE_PLUGINS=OFF` pointed out in the PR description is because until recently Emscripten didn't officially support shared libraries (CppInterOp gets around this issue by using a hack suggested by one of Emscripten maintainers). Emscripten managed to support them officially as of this PR https://github.com/emscripten-core/emscripten/pull/25930, so once Clad, llvm, CppInterOp, xeus-cpp, etc migrate to a newer emsdk, Clad shared library should be able to be built, and further functionality added.

vgvassilev:
@aaronj0, can you rebase this PR?

aaronj0:
> @aaronj0, can you rebase this PR?

Done

PR #1964, opened 2026-08-15
 Adds differenitation support for lgamma functions

Vedant2005goyal:
Adds builtin pushforwards for lgamma,lgammaf and lgammal.

### DIFF
--- a/include/clad/Differentiator/BuiltinDerivatives.h
+++ b/include/clad/Differentiator/BuiltinDerivatives.h
@@ -1254,6 +1254,25 @@ CUDA_HOST_DEVICE void beta_pullback(T x, T y, U d_z, T* d_x, T* d_y) {
   if (d_y)
     *d_y += b * (clad_digamma(y) - psi_xy) * d_z;
 }
+template <typename T, typename dT>
+CUDA_HOST_DEVICE ValueAndPushforward<T, dT>
+lgamma_pushforward(T x, dT d_x) noexcept {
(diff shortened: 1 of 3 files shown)


vgvassilev:
Can we fix the missing coverage?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2549,10 +2554,17 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
         revForwAdjointArgs.push_back(CloneNode(argDiff.getRevSweepAsExpr()));
 
       CallArgDx.push_back(argDiff.getExpr_dx());
-      if (m_DiffReq.shouldBeRecorded(arg))
+      bool requiresTape = m_DiffReq.shouldBeRecorded(arg);
+      if (m_DiffReq.Function &&
+          m_DiffReq.Function->getNameAsString() == "objective") {
+        if (FD->getNameAsString() == "subtract" ||
+            FD->getNameAsString() == "Qtimesx") {
+          requiresTape = true;
```
vgvassilev: I believe we should not be hardcoding this.

Vedant2005goyal: Yes you are right. But making this global would require structural changes in TBRAnalyzer. Could we address that as part of a future PR?”

vgvassilev (approved):
LGTM!

PR #1965, opened 2026-08-16
Index the delayed-calls planning loop to survive re-entrant appends

guitargeek:
The TU-end planning loop iterated std::deque m_DelayedCalls with a range-for while its body can grow the container: Plan() can trigger an implicit instantiation (clad::Tag, while parsing the differentiate-call arguments) whose HandleTagDeclDefinition notification lands in AppendDelayed mid-loop. deque::push_back keeps element references valid but invalidates iterators, so whenever the push made the deque grow its block map the loop's cached iterators dangled and the next element read faulted. Whether that reallocation happens depends on the total number of delayed calls, which is why the crash only reproduced on the ubuntu-22.04 llvm-12/13 CI rows (deterministic there, at HandleTranslationUnit+137 on the movups copying the DelayedCallInfo) and nowhere else. Iterate by index instead; groups appended mid-loop are then planned as well, which is a no-op for decls outside the clad-activation interval.

### DIFF
--- a/tools/ClangPlugin.cpp
+++ b/tools/ClangPlugin.cpp
@@ -638,9 +638,14 @@ void InitTimers();
       // In case of diagnostics, don't bother, just let the compiler finish.
       if (!m_CI.getDiagnostics().hasErrorOccurred()) {
         // Traverse all collected DeclGroupRef only once to create the static
-        // graph.
-        TimedAnalysisRegion R("Rest of TU");
-        for (auto DCI : m_DelayedCalls)
+        // graph. Planning can trigger implicit instantiations (e.g. clad::Tag
+        // when parsing the differentiate-call arguments) whose consumer
+        // notifications append to m_DelayedCalls mid-loop; deque::push_back
+        // keeps element references valid but invalidates iterators, so index
+        // instead of iterating (the appended groups are then planned too).
+        // NOLINTNEXTLINE(modernize-loop-convert)
+        for (size_t i = 0; i < m_DelayedCalls.size(); ++i) {
+          const DelayedCallInfo DCI = m_DelayedCalls[i];
           for (Decl* D : DCI.m_DGR) {
             if (const auto* FD = dyn_cast<FunctionDecl>(D))
               if (FD->isConstexpr())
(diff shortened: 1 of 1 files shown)


tools/ClangPlugin.cpp
```diff
@@ -638,16 +638,22 @@ void InitTimers();
...
-        TimedAnalysisRegion R("Rest of TU");
-        for (auto DCI : m_DelayedCalls)
+        // graph. Planning can trigger implicit instantiations (e.g. clad::Tag
+        // when parsing the differentiate-call arguments) whose consumer
+        // notifications append to m_DelayedCalls mid-loop; deque::push_back
+        // keeps element references valid but invalidates iterators, so index
+        // instead of iterating (the appended groups are then planned too).
+        // NOLINTNEXTLINE(modernize-loop-convert)
+        for (size_t i = 0; i < m_DelayedCalls.size(); ++i) {
+          const DelayedCallInfo DCI = m_DelayedCalls[i];
```
vgvassilev, line 648: @voyager-jhk, I believe clang-tidy needs tuning here. In that particular case, we signal that the loop will change from the inside (https://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop). That means if we evaluate the (condition check) we very likely expect the data structure to change within the loop and thus `modernize-loop-convert` yields a false positive. Can you confirm and maybe submit a bug to clang-tidy upstream?

vgvassilev (approved):
LGTM!

tools/ClangPlugin.cpp
voyager-jhk: Sounds good. I'll look into it.

vgvassilev: Thank you, if you can file a bug upstream would be great!

voyager-jhk: @vgvassilev turns out this is a known clang-tidy limitation. The check only looks at the loop body's own text, so it can't see that the container gets mutated a few calls deep inside a callback.
Already tracked upstream: https://github.com/llvm/llvm-project/issues/154764 (no fix in sight).

vgvassilev: Thanks for investing!

PR #1966, opened 2026-08-16
Defer decl groups delivered while a planning traversal is in flight

guitargeek:
DiffCollector::Walk asserts that no traversal is already running, but it can be re-entered through the AST reader: a name lookup issued mid-traversal (LookupCustomDerivativeDecl, the analyses) makes the reader deserialize pending module decls and pass the interesting ones straight to the consumers, which lands in
CladPlugin::HandleTopLevelDecl -> Plan -> Walk while the outer walk -- and possibly its top-most request -- is still on the stack:

  DiffCollector::VisitCallExpr
    LookupCustomDerivativeDecl
      Sema::LookupQualifiedName
        ASTReader::FindExternalVisibleDeclsByName
          ASTReader::PassInterestingDeclsToConsumer
            CladPlugin::HandleTopLevelDecl
              DiffScheduler::Plan
                DiffCollector::Walk   // re-entered

Seen with ROOT's runtime C++ modules, where the second RooFit codegen fit in a process hits a lookup that flushes pending module decls. Traversing the group immediately would interleave collector state, so park it and drain it when the outermost walk finishes; deserialized decls lie outside the clad-activation interval, so the deferred traversal is a no-op for them beyond bookkeeping. The plugin's immediate constexpr processing and its no-multiplexer finalization are guarded the same way -- re-entrant ProcessDiffRequest would clobber the graph's current processing node.

### DIFF
--- a/include/clad/Differentiator/DiffPlanner.h
+++ b/include/clad/Differentiator/DiffPlanner.h
@@ -405,14 +405,30 @@ struct RequestOptions {
 
     bool m_IsTraversingTopLevelDecl = true;
 
+    /// True while Walk is traversing a DeclGroupRef. A traversal can trigger
+    /// name lookups that make the ASTReader deserialize pending module decls
+    /// and hand them to the consumers, re-entering Walk; such groups are
+    /// parked in m_DeferredDGRs and traversed once the active walk finishes.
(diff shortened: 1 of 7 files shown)


vgvassilev (approved):
Amazing! Thank you!

PR #1967, opened 2026-08-16
Clone the callee and arguments into the numerical-diff fallback call

guitargeek:
When forward mode cannot derive a call (e.g. one whose arguments are all literals, for which the planner schedules no pushforward), it rebuilds the primal call and asks GetSingleArgCentralDiffCall for a forward_central_difference call to use as the derivative. The callee it passes is taken straight out of the rebuilt primal call, and the primal argument expressions are appended as the trailing call arguments, so the generated pushforward held the same callee ImplicitCastExpr and the same argument subtree under two parents:

  double _t6 = std::sqrt(2 * M_PI);   // primal
  ... numerical_diff::forward_central_difference(
          std::sqrt /*shared*/, 2 * M_PI /*clone*/, 0, 0,
          2 * M_PI /*shared*/) * 0 ...

Seen with clad::hessian of RooFit's lognormal likelihoods; the shared nodes trip the derivative integrity check (and any later in-place edit of one user would corrupt the other). Clone the callee and every forwarded argument; the middle targetArg was already cloned.

### DIFF
--- a/lib/Differentiator/VisitorBase.cpp
+++ b/lib/Differentiator/VisitorBase.cpp
@@ -1018,9 +1018,12 @@ namespace clad {
     bool isSupported = argType->isArithmeticType();
     if (!isSupported)
       return nullptr;
-    // Build function args.
+    // Build function args. The callee and argument expressions the caller
+    // hands in may still be parented by the rebuilt primal call (the forward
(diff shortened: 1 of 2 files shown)


vgvassilev (approved):
LGTM!

PR #1969, opened 2026-08-16
Parenthesize compound reverse-sweep exprs in the printed pullbacks

guitargeek:
A compound assignment whose right-hand side StoreAndRef passes through un-stored (`l *= a + b` with an unchanging RHS) embeds that expression directly in the `_r_d * R` pullback product. The built AST is correct -- the product holds the sum as one operand -- but nothing inserts a ParenExpr, and Clang's printer only prints parentheses that exist as nodes, so the dumped derivative mis-associates:

  _d_C[0] += _r_d0 * alpha + beta;   // AST: _r_d0 * (alpha + beta)

Anything that recompiles printed derivatives (Derivatives.cpp generation, CladFunction::dump()-based workflows) gets wrong code. Wrap the reverse-sweep operands of the mul- and div-assign products and rev-pass values in BuildParens, which only parenthesizes exprs that print in binary/ternary form.

BuildParens itself treated any two-argument overloaded operator call as binary; subscript and call operators are postfix and never need parentheses, so stop wrapping those. This also drops the noise parens vector forward mode used to print around `_d_vector_arr[i]`.


vgvassilev (approved):
Lgtm!

PR #1972, opened 2026-08-16
Fix reverse-mode comma expression result handling

Elvand-Lie:
## Problem

Reverse-mode differentiation of the built-in comma operator `(L, R)` correctly propagated the incoming adjoint to `R`, but returned the left operand as the expression's reverse-sweep value and derivative result reference.

This could make enclosing expressions use the discarded left operand. For example, differentiating `(a, b) * c` produced a contribution based on `a` instead of `b`.

It also affected comma expressions used as lvalues, such as `(a, b) = c`, where the assignment must operate on `b` and its adjoint.

## Fix

Use the differentiated right operand for both:

- the primal value needed by the reverse sweep; and
- the derivative/adjoint result reference returned to enclosing expressions.

The left operand is still visited with a zero adjoint, preserving its evaluation and side effects.

## Testing

Added `test/Gradient/CommaOperator.C` covering:

- comma expressions used as values;
- side effects in the left operand;
- comma expressions used as lvalue assignment targets;
- nested and chained comma expressions.

Validation:

- `llvm-lit -sv test/Gradient/CommaOperator.C`
- `ninja check-clad-gradient`
- `ninja check-clad-firstderivative`
- `ninja check-clad-arrays`
- `ninja check-clad-hessian`
- `ninja check-clad-jacobian`
- `ninja check-clad-nestedcalls`
- `git-clang-format --binary clang-format-18 --diff origin/master`
- `git diff --check origin/master`

All supported tests passed; the Gradient suite has one unsupported OpenMP test in the local environment due to missing OpenMP headers.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -3514,8 +3514,8 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
           ConstantFolder::synthesizeLiteral(m_Context.IntTy, m_Context, 0);
       Rdiff = Visit(R, dfdx());
       Ldiff = Visit(L, zero);
-      valueForRevPass = Ldiff.getRevSweepAsExpr();
-      ResultRef = Ldiff.getExpr();
+      valueForRevPass = Rdiff.getRevSweepAsExpr();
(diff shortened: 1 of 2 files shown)


vgvassilev (approved):
Lgtm!

PR #1973, opened 2026-08-17
Share the lambda-introduction scaffolding between both builders

guitargeek:
buildDerivedLambda and buildClonedLambda each carried their own copy of the sequence Sema performs while parsing a lambda: introduce the closure, push its call operator as the current declaration context, describe the parameter list through a Declarator, then open the body scope. The two copies were identical for ~50 lines, most of it the 25-argument DeclaratorChunk::getFunction call, and they only diverge in the closure's captures, the type of its call operator, how the parameters are produced, and how the body is filled.

Move the shared part into a VisitorBase::LambdaScaffold helper that also owns the parser-level state Sema keeps referring to until ActOnLambdaExpr. Each builder is now its introducer setup, a parameter-building callback, and its own body construction. No functional change.

**This is a code consolidation commit, no functional changes**

### DIFF
--- a/include/clad/Differentiator/VisitorBase.h
+++ b/include/clad/Differentiator/VisitorBase.h
@@ -13,6 +13,7 @@
 
 #include "clang/AST/Decl.h"
 #include "clang/AST/Expr.h"
+#include "clang/AST/ExprCXX.h"
 #include "clang/AST/RecursiveASTVisitor.h"
 #include "clang/AST/Stmt.h"
 #include "clang/AST/StmtVisitor.h"
@@ -21,6 +22,7 @@
 #include "clang/Basic/Lambda.h"
 #include "clang/Basic/OperatorKinds.h"
 #include "clang/Basic/Specifiers.h"
+#include "clang/Basic/Version.h"
 #include "clang/Sema/DeclSpec.h"
 #include "clang/Sema/Ownership.h"
 #include "clang/Sema/ParsedAttr.h"
@@ -30,6 +32,12 @@
 #include "llvm/ADT/SmallPtrSet.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/PrettyStackTrace.h"
+
+// For the LambdaBuilder
+#if CLANG_VERSION_MAJOR > 16
+#include "llvm/ADT/STLFunctionalExtras.h"
+#include "llvm/Support/SaveAndRestore.h"
+#endif // CLANG_VERSION_MAJOR > 16
 
 #include <array>
 #include <cassert>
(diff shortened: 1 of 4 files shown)


include/clad/Differentiator/VisitorBase.h
```diff
@@ -305,6 +307,41 @@ namespace clad {
       bool contains(clang::VarDecl* VD) const { return m_Captures.count(VD); }
     };
 
+#if CLANG_VERSION_MAJOR > 16
+    /// Builds a new lambda closure with an explicit parameter list, following
+    /// the order Sema relies on while parsing one -- deviating from it leaves
+    /// Sema in a state that only crashes much later. Used by
+    /// `buildClonedLambda` and `ReverseModeVisitor::buildDerivedLambda`.
+    class LambdaScaffold {
```
vgvassilev:
```suggestion
    class LambdaBuilder {
```

vgvassilev (commented):
Thanks, this is really a good improvement.

include/clad/Differentiator/VisitorBase.h
guitargeek: I did the rename to `LambdaBuilder` for the class name, and `LBuilder` for the instances. Is that okay, or what are the usual conventions there?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2192,19 +2108,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
-                DerivedBody /*,*/
-                    CLAD_COMPAT_CLANG17_ActOnLambdaExpr_getCurrentScope_ExtraParam(
-                        *this))
-            .get();
-
-    endScope();
-    endScope();
-    endScope();
-    return lambda;
+    return LBuiler.finish(DerivedBody);
```
vgvassilev: I am wondering if we don't need some RAII logic in case this gets some early return somewhere in future..

vgvassilev: That is, maybe we should assert in the destructor if `finish` was not called.

include/clad/Differentiator/VisitorBase.h
vgvassilev: I think that's fine -- we follow more-or-less the clang coding conventions.

vgvassilev (approved):
LGTM % my assert in the dtor comment.

guitargeek:
The coverage hole was preexisting. It just got unearthed by the refactor. I'll add test coverage anyway

PR #1974, opened 2026-08-17
Declare clad's derivatives inline

guitargeek:
Hessian mode derives the function once in forward mode per direction and reverse-differentiates each of those, cloning the first-order body into the second. Nothing ever calls `f_darg0_<i>` afterwards, yet the backend got one dead function per direction, each roughly the size of the primal -- on a 41-parameter RooFit likelihood, 17% of all emitted derivative AST and 1.19x on the time to produce the hessian.

CodeGen defers a function until a DeclRefExpr asks for it, so why were these exempt? clad cloned the primal's inline-ness, and a plain `double f(...)` yields a strong external definition, which ASTContext::DeclMustBeEmitted requires this TU to emit.

But a derivative is not the primal. It is a small function clad regenerates in every translation unit that asks for it, and the handwritten pushforwards and pullbacks in BuiltinDerivatives.h are already inline for that reason. So clone every derivative inline at the FunctionDecl::Create and CXXMethodDecl::Create calls instead of carrying the primal's specifier over, and no mode needs a case of its own. Being discardable, CodeGen drops whatever nothing references, and a later clad::differentiate(f, "x") reusing the cached declaration brings its definition back through the DeclRefExpr in its own call.

The user-facing f_darg0, f_grad, f_hessian and f_jacobian are inline as well. The CladFunction that names one sits in the same translation unit, so its DeclRefExpr keeps the definition alive.

The exception is a derivative whose name a user already defined by hand: [dcl.inline]p6 makes an inline declaration that follows a definition ill-formed, and Hessian/BuiltinDerivatives.C does exactly that with clad::custom_derivatives::f7_darg1_grad. registerDerivative already looks the name up for Sema's redeclaration merge, so drop the specifier there when that lookup finds a non-inline definition.

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -80,6 +80,16 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
     if (R.Function && !R.Function->getPrimaryTemplate())
       S.LookupQualifiedName(Previous, dFD->getParent());
 
+    // Derivatives are declared inline, but a hand-written custom derivative may
+    // already define this name, and [dcl.inline]p6 forbids an inline
+    // declaration that follows a definition. Drop the specifier in that case.
+    auto definedNotInline = [](const NamedDecl* ND) {
+      const auto* PrevFD = dyn_cast<FunctionDecl>(ND);
+      return PrevFD && PrevFD->isDefined() && !PrevFD->isInlined();
+    };
+    if (std::any_of(Previous.begin(), Previous.end(), definedNotInline))
+      dFD->setInlineSpecified(false);
+
     // Check if we created a top-level decl with the same name for another
     // class.
     // FIXME: This case should be addressed by providing proper names and
@@ -134,7 +144,7 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
       returnedFD = CXXMethodDecl::Create(
           m_Context, CXXRD, noLoc, name, functionType, TSI,
           SC CLAD_COMPAT_FunctionDecl_UsesFPIntrin_Param(FD),
-          FD->isInlineSpecified(), FD->getConstexprKind(), noLoc);
+          /*isInlineSpecified=*/true, FD->getConstexprKind(), noLoc);
       // Generated member function should be called outside of class definitions
       // even if their original function had different access specifier.
       returnedFD->setAccess(AS_public);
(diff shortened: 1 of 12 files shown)


include/clad/Differentiator/DiffPlanner.h
```diff
@@ -305,6 +305,12 @@ struct DiffRequest {
   /// This will be particularly useful for pushforward and pullback functions.
   bool DeclarationOnly = false;
 
+  /// A boolean marking a derivative that only exists to be differentiated
+  /// again: nothing ever calls it, so it is kept in the AST but not sent to
+  /// CodeGen. Emitting it would hand the backend a dead function roughly the
+  /// size of the primal.
+  bool NotCodeGenerated = false;
```
vgvassilev: That goes against the way clang and C++ work. Clang's CodeGen does not eagerly generate code for every function -- it starts decides what code to generate from a given `DeclRefExpr` in the function body. I believe something more subtle happens here, that is, somehow codegen decides that this function cannot be deferred. Look at `ASTContext::`[DeclMustBeEmitted](https://clang.llvm.org/doxygen/classclang_1_1ASTContext.html#a3b9cd9f4bc5349ff4afd1f98c6ceca28)

guitargeek: You're right! There was indeed a more idiomatic solution using `FD->setImplicitlyInline(true);`, that I have used now in the most recent force-push.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -736,6 +736,15 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
...
         registerDerivative(OFD, m_Sema, request);
+      // An intermediate derivative inherits the primal's linkage, and a plain
+      // `double f(...)` gives a strong external definition, which C++ requires
+      // to be emitted (ASTContext::DeclMustBeEmitted). Marking it inline makes
+      // it discardable instead, so CodeGen keeps it deferred until a
+      // DeclRefExpr shows up. A custom derivative is user-written, so its
+      // linkage is not ours to change.
+      if (request.IsIntermediate)
+        if (auto* FD = dyn_cast_or_null<FunctionDecl>(result.derivative))
+          FD->setImplicitlyInline(true);
```
vgvassilev: Actually it'd make sense to mark them explicitly as inlined functions. They are small helpers so it'd be probably better the builder to do that properly -- this is the same anyway for the non-templated pullbacks and `reverse_forw`s in the `.h` files.

guitargeek: Done!

vgvassilev (approved):
Lgtm!

guitargeek:
Thanks for the review! I guess I need to update a few more references though

PR #1976, opened 2026-08-17
Generate the clad::array element-wise assignments from macros

guitargeek:
clad::array and clad::array_ref each carry a family of compound assignment operators per operand kind: a scalar, a raw pointer, an array_ref, an array of a different element type, and an unevaluated array_expression. Within a family the four operators differ in exactly one token -- the loop body, the assert, the enable_if and the return are copied verbatim -- so the two headers spent ~340 lines saying the same thing 38 times. The copies had already started to drift: array_ref's four array_ref overloads all assert "for carrying out addition assignment", including the three that are not addition.

Replace each family with a generator macro taking the operator token, so a family reads as one body plus a list of the tokens it is instantiated for. The macros are #undef'd at the end of the header that defines them. The resulting overload set is unchanged; only the shared assert message is reworded to "compound assignment".

Three overloads stay written out because they do not fit their family: array's operator=(T*), which treats a null pointer as all-zero; array's operator= from an array_ref of a different element type, the one assignment that crosses element types; and array_ref's operator=(const array_ref<T>&), which has to stay non-templated to catch self-assignment.

Verified by AST-printing both headers before and after and diffing the declared operator signatures (identical up to one parameter name), and by running test/Misc/CladArray.C, test/Misc/CladMatrix.C and a TU exercising every overload under clang and gcc, C++17 and C++20, with asserts on -- byte-identical output in every case.

**This is a code consolidation commit, no functional changes**

### DIFF
--- a/include/clad/Differentiator/Array.h
+++ b/include/clad/Differentiator/Array.h
@@ -18,6 +18,70 @@ template <typename T> class array_ref;
 #define PUREFUNC __attribute__((pure))
 #endif
 
+// The element-wise assignments of clad::array come in families that differ
+// only in the operator token, passed to these generators as `op`: one of `+=`,
+// `-=`, `*=`, `/=`, or `=` where plain assignment has the same shape.
+//
+// Three checks fire on the generators below and cannot be followed here:
+// cppcoreguidelines-macro-usage, because `op` is an operator token and each
+// expansion declares a different operator overload; bugprone-macro-parentheses,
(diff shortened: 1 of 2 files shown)


vgvassilev (approved):
That does have some readability implications but I do not think  it should be a huge challenge to read through. What it improves is consistency in iterations, etc. LGTM!

guitargeek:
Yes, that's always a tradeoff. But here, I think the win in consistency outweighs readability concerns from the extra indirection

PR #1977, opened 2026-08-17
Use CLAD_SOURCE_DIR and CLAD_BINARY_DIR in lib/Differentiator

guitargeek:
VERSION, the version control file and Version.inc were located relative to CMAKE_SOURCE_DIR and CMAKE_BINARY_DIR, which are clad's own directories only when clad happens to be the top level project. CLAD_SOURCE_DIR and CLAD_BINARY_DIR are ours in every configuration, so use those and drop the LLVM_EXTERNAL_CLAD_SOURCE_DIR special case they needed.

This changes where two generated files end up, on purpose:

Version.inc now goes to CLAD_BINARY_DIR/include instead of CMAKE_BINARY_DIR/include. Those are the same directory in a standalone build, but for a clad built as an external project of LLVM the old path was inside the LLVM binary directory. That is neither the directory clad puts on its include path nor the one it installs from, so clad was picking the header up only because LLVM happens to add its own binary include directory globally, and clad's install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/include/) never shipped it. Both now line up.

VCSVersion.inc now describes clad. find_first_existing_vc_file and the CLAD_SOURCE_DIR passed to GenerateVersionFromVCS.cmake were given CMAKE_SOURCE_DIR, so an external-project build searched the host checkout and stamped the host's git revision and remote into CLAD_REVISION and CLAD_REPOSITORY. clad::getCladRevision() and clad::getCladRepositoryPath() there reported the LLVM checkout instead of ours.

While the generated headers are being sorted out, give cladDifferentiator the directory VCSVersion.inc is generated into. Version.cpp includes it by its bare name and so far relied on CMAKE_INCLUDE_CURRENT_DIR, which clad only sets on the standalone path, so the include did not resolve once a host project added clad with add_subdirectory().

### DIFF
--- a/lib/Differentiator/CMakeLists.txt
+++ b/lib/Differentiator/CMakeLists.txt
@@ -1,18 +1,16 @@
-# The VC revision include that we want to generate.
-if (LLVM_EXTERNAL_CLAD_SOURCE_DIR)
-  set(version_file ${LLVM_EXTERNAL_CLAD_SOURCE_DIR}/VERSION)
-else()
-  set(version_file ${CMAKE_SOURCE_DIR}/VERSION)
-endif(LLVM_EXTERNAL_CLAD_SOURCE_DIR)
+# The VC revision include that we want to generate. Note that CLAD_SOURCE_DIR
+# and CLAD_BINARY_DIR are ours also when clad is only a part of a bigger build,
+# whereas CMAKE_SOURCE_DIR and CMAKE_BINARY_DIR belong to whoever is on top.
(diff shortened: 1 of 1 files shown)


vgvassilev (approved):
Lgtm if clad’s getVersion-related api return the same after this patch.

guitargeek:
The API returns the same in the standalone and in-tree builds, bit-for-bit. In the llvm-in-tree build the version string is the same and the revision/repository fields now report clad instead of LLVM, which is the point of the patch.

PR #1979, opened 2026-08-17
Null check a tangent pointer at the root of its pointer arithmetic

guitargeek:
Clad cannot size a tangent buffer for a `const double*` parameter, so a pushforward is handed a null tangent for it. Null spells "the derivative of this argument is identically zero", and forward mode null checks the reads through it.

Pointer arithmetic destroys that encoding, because `nullptr + 1` is 0x8 rather than null. The guard was still emitted -- the planner does mark the variable as possibly having a null tangent -- it just tested a pointer the arithmetic had already made non-null, so the read went ahead and dereferenced a small integer. That is what made the failure quiet rather than loud. These are the shapes RooFit's code generation emits for a multi-channel likelihood, so every such hessian crashed at run time.

Add getPointerArithmeticRoot, which walks an expression back to the declaration its pointer value originates in, and KeepTangentNullness, which rebuilds a derived tangent to evaluate to null whenever the tangent it came from is null. Guard the five places a tangent outlives the expression that built it: variable initializers, call arguments, assignments, compound assignments and increments.

Null-ness originates at the root and pointer arithmetic preserves it, so the root is what the check has to test. Testing it there also keeps the condition free of the arithmetic, which matters because the condition is cloned. `p += n` and `p++` update the pointer they are rooted in and are walked through; plain `p = q` is rooted in `q` instead, so the assignment case guards the derived right-hand side while the in-place cases guard the whole update.

Widening the walk also makes the pre-existing read guard fire on `*(p++)`, where the tangent already carries a step. Such a tangent is both tested and read through, so it is stored first and the guard tests the stored value -- one step, one test.

### DIFF
--- a/include/clad/Differentiator/BaseForwardModeVisitor.h
+++ b/include/clad/Differentiator/BaseForwardModeVisitor.h
@@ -74,6 +74,27 @@ public:
                                     clang::Expr* tangent, clang::Expr* read,
                                     clang::Expr* zero);
 
+  /// \returns the declaration \p E is derived from by pointer arithmetic, or
+  /// null when \p E is not rooted in one. `xlArr + 1`, `xlArr += 1` and
+  /// `xlArr++` all yield `xlArr`; plain `q = xlArr + 1` does not, being rooted
+  /// in its right-hand side. A tangent pointer is null at its root, so the
+  /// root is what a null check has to test.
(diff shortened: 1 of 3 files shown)


vgvassilev (approved):
Lgtm!

guitargeek:
Read the docs seems stuck. Everything else passed.

PR #1980, opened 2026-08-17
Track the independent-variable offset in one place

guitargeek:
Vector forward mode and Jacobian mode each computed the offset of a parameter into the flat vector of independent variables by hand, three times over: in VectorForwardModeVisitor::Derive, in its VisitReturnStmt, and in JacobianModeVisitor::Derive. Every copy accumulated the running sum of array sizes, folded the scalar count into a literal, combined the two, and cloned expressions on both sides before splicing them into the derivative.

The clones are what made the duplication expensive. The running sum keeps growing after an offset is taken from it, and each offset is spliced into a different part of the derivative, so handing out the accumulator itself would give an AST node two parents. The same applies to the size expression: it is already used by the identity matrix or array slice built for that parameter, so the accumulator has to take a copy. Each of the three sites had to know all of this independently.

Replace them with an IndVarOffsetTracker that owns the accumulation and clones on the way in and on the way out, so the one-parent invariant is structural instead of restated at each site.

The tracker stores a VectorForwardModeVisitor* rather than a VisitorBase*, because m_Context and BuildOp are protected in VisitorBase and cannot be reached through a base pointer. Its methods are defined in the .cpp so ConstantFolder.h stays out of the public header.

The two sites that compute the *total* independent-variable count (VectorForwardModeVisitor.cpp and JacobianModeVisitor.cpp) use the same combine formula but hand the accumulated expression off instead of cloning it, so they are left alone; routing them through the tracker would orphan a node per array parameter.

No functional change: the Jacobian, VectorMode and ArrayInputsVectorForwardMode tests check the generated derivative source line by line with FileCheck and pass unmodified, so the emitted AST is identical. Verified against a debug build, where the AST-integrity checks assert rather than warn.

**This is a refactor to consolidate the code, no functional changes**

### DIFF
--- a/include/clad/Differentiator/VectorForwardModeVisitor.h
+++ b/include/clad/Differentiator/VectorForwardModeVisitor.h
@@ -4,6 +4,9 @@
 #include "BaseForwardModeVisitor.h"
 #include "DerivativeBuilder.h"
 
+#include "clang/AST/Expr.h"
+
+#include <cstddef>
 #include <unordered_map>
 
 namespace clad {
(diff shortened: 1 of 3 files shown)


vgvassilev:
Shouldn’t that be refactored in the DiffPlanner and the DiffRequest?

guitargeek:
> Shouldn’t that be refactored in the DiffPlanner and the DiffRequest?

The offset isn't a property of the primal, so there's nothing here for the planner to compute. What the tracker accumulates are expressions in the *derivative's* AST — `_d_vector_p.rows()` in Jacobian mode, `_d_x.size()` in vector forward mode — member calls on the derivative function's own parameters. Those parameters are built by the visitor (`BuildVectorModeParams`, and the parameter loop in `JacobianModeVisitor::Derive`), so they don't exist while `DiffPlanner` runs; `DiffRequest` only ever holds nodes from the primal and the call site. The sizes are runtime values too, so the offset isn't a compile-time number at all.

The static half — is this independent variable an array or a scalar, and in what order — is already on the request: `DVI` gives the order, array-ness is `utils::isArrayOrPointerType(param->getType())`. Recording a per-`DVI` offset on the request would store that one bool, and each visitor would still have to build the same sum of `.size()` calls, with the clone bookkeeping this PR consolidates unchanged.

If the concern is where the helper *lives* rather than when it runs, I'm happy to lift `IndVarOffsetTracker` out of `VectorForwardModeVisitor` into its own header so Jacobian mode doesn't reach into a nested class of the vector-forward visitor. And if you did mean the planner, could you say which fact you'd want recorded on the request? I may be reading the suggestion too narrowly.

vgvassilev (approved):
Ok, makes sense. Lgtm!

PR #1983, opened 2026-08-18
Let TBR run on dynamically scheduled reverse-mode requests

guitargeek:
HandleNestedDiffRequest force-disabled TBR, varied and useful analysis for every lazily-scheduled request, because such requests never pass through the static TU walk that builds an AnalysisDeclContext for them.

TBR does not need the planner to run it: DiffRequest::shouldBeRecorded() starts the analysis on demand and only needs m_AnalysisDC to be set. Build one in PlanNestedRequest -- which already exists to give these requests the planning information the TU walk never produced -- and keep TBR enabled whenever we got a context. The varied and useful analyses do have to be run eagerly by the planner, so they stay off and the FIXME shrinks to cover just them.

The context is built only for the reverse-mode family, the modes that consult shouldBeRecorded(), and only for a Function the analyzer will actually accept: TBRAnalyzer seeds its state from request.Function's parameters and matches them against the CFG, so the request's Function has to be the definition the CFG comes from -- a redeclaration's ParmVarDecls would silently match nothing -- and a lambda call operator, which shouldBeRecorded() refuses to analyze for the same kind of reason, is left alone. The context is cached per function, so planning the same nested request again from another call site reuses one CFG.

What this buys is dropping the stores the reverse sweep never reads back, so the size of the win depends on the function. Two existing tests show it: flexibleInterp_pullback loses the tape for its loop-carried 'total' entirely -- the accumulation is linear, so the old value is never needed -- and pow_pushforward_pullback loses three temporaries and two restores. The new hessian test tapes four values instead of eight. Hessians benefit most often, because their inner reverse pass runs once per requested direction over a function the size of the first-order derivative; HessianModeVisitor asks for TBR there by inheriting the flag from the hessian request.

Running TBR on lazily-scheduled requests also means running it on clad-generated functions, whose synthesized statements carry no source location. The debug-only location dump at the end of TBRAnalyzer::Analyze called PresumedLoc::getLine() unconditionally, which asserts on an invalid location and aborts the compiler in any assertions-enabled build; it took out Regressions/issue-1955.cpp. It now skips what it cannot print.

Verified with a full lit run (221 tests), and by checking that the new test fails again once HandleNestedDiffRequest force-disables TBR.

### DIFF
--- a/include/clad/Differentiator/DiffPlanner.h
+++ b/include/clad/Differentiator/DiffPlanner.h
@@ -17,6 +17,7 @@
 #include "clang/Analysis/AnalysisDeclContext.h"
 #include "clang/Basic/SourceLocation.h"
 
+#include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/DenseSet.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringSwitch.h"
@@ -424,6 +425,11 @@ struct RequestOptions {
     /// Essentially needed for prolonging the lifetime of
     /// unique_ptr<clang::AnalysisDeclContext>.
     OwnedAnalysisContexts& m_AllAnalysisDC;
+    /// The contexts PlanNestedRequest built, keyed by the function they
+    /// describe, so that repeated planning of the same lazily-scheduled
+    /// request reuses one context (and one CFG) instead of building another.
+    llvm::DenseMap<const clang::FunctionDecl*, clang::AnalysisDeclContext*>
+        m_NestedAnalysisDC;
     /// If set it means that we need to find the called functions and
     /// add them for implicit diff.
     ///
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -413,10 +413,13 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
     FunctionDecl* derivative = this->FindDerivedFunction(request);
     if (!derivative) {
       alreadyDerived = false;
-      // FIXME: Our analyses are closely tied to the DiffPlanner. Dynamic
-      // derivatives don't have m_AnalysisDC. We should either disable
-      // dynamic scheduling or build m_AnalysisDC here.
-      request.EnableTBRAnalysis = false;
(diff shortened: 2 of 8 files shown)


vgvassilev:
Does that relate in any shape of form to #1880?

guitargeek:
> Does that relate in any shape of form to #1880?

Yes! Same lines, disjoint halves.

#1880 and this PR both rewrite the `FIXME` block in `HandleNestedDiffRequest` that force-disables the analyses for lazily scheduled requests, but they turn on different ones. #1880 turns on the varied analysis so hessians stop emitting adjoints for loop counters (#1665). This PR turns on TBR only, and deliberately leaves `EnableVariedAnalysis`/`EnableUsefulAnalysis` off: those two have to be run eagerly by the planner, while TBR starts itself from `shouldBeRecorded()` and only needs `m_AnalysisDC` to be set.

The other difference is where the `AnalysisDeclContext` comes from. #1880 factors the construction into `PrepareRequestAnalysis` and threads `OwnedAnalysisContexts` down into `DerivativeBuilder`. Here it is built in `PlanNestedRequest`, the planner hook that already exists to give nested requests the planning information the TU walk never produced, so `DerivativeBuilder` keeps knowing nothing about analysis contexts. The context is cached per `FunctionDecl`, so planning the same nested request from a second call site reuses one CFG.

So the two compose rather than compete. If this one lands first, #1880 gets smaller: the context is already on the request, and its diff shrinks to running `VariedAnalyzer` on it, plus the `shouldHaveAdjointForw` and varied-seeding changes in `DiffPlanner.cpp` which are independent of this PR. It needs a rebase either way.

One thing more to note: #1880's new branch does not set `request.EnableTBRAnalysis = false`, so with `-enable-va` it already enables TBR on lazily scheduled requests as a side effect. It would then hit the crash this PR fixes. The debug-only location dump at the end of `TBRAnalyzer::Analyze` calls `PresumedLoc::getLine()` on the invalid locations of clad-generated statements, which asserts and takes out `Regressions/issue-1955.cpp` in any assertions-enabled build. That fix is needed by either PR.

vgvassilev:
Ok, makes sense -- we can use that as a baseline and rebase the other following the same design...

vgvassilev (approved):
LGTM!

PR #1984, opened 2026-08-18
Recognize a constant-zero tangent that has been bound to a variable

guitargeek:
VisitCallExpr already knows that a call whose argument tangents are all zero contributes nothing, and emits the plain call with a zero derivative instead of asking for a pushforward. The check went through IsZeroOrNullValue, which pattern-matches literals -- but a tangent has almost always been bound to a declaration by the time it reaches a call, so what the check actually sees is a reference to a `const double _d_t6 = 0;`.

This matters most when a request seeds one direction at a time, as a hessian does: there every tangent except the seeded one is exactly such a zero-initialized constant, and each direction was paying for the full chain of pushforwards anyway. The parameters have to be const-qualified for the fold to fire. A tangent that can still be assigned to is never folded, which is what stops a tangent that is reassigned later -- in a loop, say -- from being wrongly read as zero.

Add an IsZeroOrNullValue overload that constant-folds the expression and use it here. EvaluateAs* refuses side effects by default, so an expression that could observe or change anything on the way is not folded and the answer stays conservative.

Folding also widens an existing behavior that is worth stating plainly. The zero-derivative path keeps the primal call in the StmtDiff, but a _darg function returns only the derivative half, so the call is dropped along with it. That was already true of a literal-zero tangent; it now fires for every const-qualified parameter, so adding const to a signature can stop a side-effecting callee from running:

```c++
    int calls = 0;
    double g(double a) { ++calls; return a * a; }
    double f(const double x, const double y) { return g(y) + x * x; }
```

f_darg0 no longer calls g, and calls stays 0 where it used to reach 1. The derivative is unaffected -- an elided call contributes nothing to it -- and this is the intended trade for not differentiating through a term that cannot contribute.

Two tests change their expected output accordingly. CallArguments.C loses one pushforward in f_const_args_func_7_darg0; its non-const twin func_8 keeps both, which is exactly the distinction the fold draws. TFormula.C loses the Exp pushforward in the directions that do not seed p[0], where the tangent is `-0.` -- a UnaryOperator the literal matcher never caught. Hessian/ZeroTangentCalls.C is new and pins the hessian case.

### DIFF
--- a/include/clad/Differentiator/CladUtils.h
+++ b/include/clad/Differentiator/CladUtils.h
@@ -446,6 +446,12 @@ namespace clad {
 
     bool IsZeroOrNullValue(const clang::Expr* E);
 
+    /// Like IsZeroOrNullValue(E), but also folds \p E as a constant
+    /// expression, so that a reference to a `const` variable initialized to
+    /// zero -- the shape a tangent takes once it has been bound to a
+    /// declaration -- is recognized as zero too.
+    bool IsZeroOrNullValue(const clang::Expr* E, const clang::ASTContext& C);
+
     bool IsMemoryDeallocationFunction(const clang::FunctionDecl* FD);
 
     /// Returns true if QT is a non-const reference type.
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -1357,8 +1357,12 @@ StmtDiff BaseForwardModeVisitor::VisitCallExpr(const CallExpr* CE) {
       for (unsigned i = 0, e = diffArgs.size(); i < e; ++i) {
         Expr* dArg = diffArgs[i];
         // If argDiff.expr_dx is nullptr or is a constant 0, then the derivative
-        // of the function call is 0.
-        if (!clad::utils::IsZeroOrNullValue(dArg)) {
+        // of the function call is 0. Constant-fold rather than pattern-match a
+        // literal: by the time a tangent reaches a call it has usually been
+        // bound to a `const` variable, and seeding one direction of a
+        // multi-directional request leaves every other tangent exactly that --
+        // a zero-initialized constant.
+        if (!clad::utils::IsZeroOrNullValue(dArg, m_Context)) {
           allArgsHaveZeroDerivatives = false;
           break;
         }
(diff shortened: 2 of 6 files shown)


vgvassilev:
Isn’t that an analysis at its core that affects the overall diff plan? The decision whether or not to differentiate g should be in the DiffPlanner right?

guitargeek:
> Isn’t that an analysis at its core that affects the overall diff plan? The decision whether or not to differentiate g should be in the DiffPlanner right?

Agreed that this belongs in the planner -- the two FIXMEs at that site say the same thing, and this PR doesn't move the decision, it only fixes how zero is recognized at the site that already makes it.

I tried routing it through `DiffRequest::isVaried()` (which `VisitCallExpr` already consults for the member-call base a bit further up) and it doesn't hold together yet. Varied analysis is off by default, so the elision would never fire; and with `-enable-va` the varied set isn't populated for forward-mode requests, so every argument reads as inactive, every call gets elided, and the hessian comes back as 0,0,0,0 instead of 12,0,0,18. Extending the seeding at `DiffPlanner.cpp:1433` to forward mode isn't enough either.

So making activity carry this needs forward-mode requests seeded from `DVI` and `VariedAnalyzer` run per-direction for the spawned `darg` requests first. Happy to do that as a follow-up -- would you rather this one wait for it, or land as the narrow recognition fix it is?

vgvassilev:
> > Isn’t that an analysis at its core that affects the overall diff plan? The decision whether or not to differentiate g should be in the DiffPlanner right?
> 
> Agreed that this belongs in the planner -- the two FIXMEs at that site say the same thing, and this PR doesn't move the decision, it only fixes how zero is recognized at the site that already makes it.
> 
> I tried routing it through `DiffRequest::isVaried()` (which `VisitCallExpr` already consults for the member-call base a bit further up) and it doesn't hold together yet. Varied analysis is off by default, so the elision would never fire; and with `-enable-va` the varied set isn't populated for forward-mode requests, so every argument reads as inactive, every call gets elided, and the hessian comes back as 0,0,0,0 instead of 12,0,0,18. Extending the seeding at `DiffPlanner.cpp:1433` to forward mode isn't enough either.
> 
> So making activity carry this needs forward-mode requests seeded from `DVI` and `VariedAnalyzer` run per-direction for the spawned `darg` requests first. Happy to do that as a follow-up -- would you rather this one wait for it, or land as the narrow recognition fix it is?

For me the important part is whoever follows through to make sure we don't forget to move it. Let's land this one and see..

vgvassilev (approved):
LGTM!

PR #1985, opened 2026-08-18
Drop two tests fully subsumed by their constexpr twins

guitargeek:
test/Hessian/testhessUtility.C and test/Jacobian/testUtility.C are line-for-line copies of test/Hessian/constexprTest.C and test/Jacobian/constexprTest.C with the constexpr keyword removed: same functions, same CHECK blocks, same expected values. The constexpr variants exercise a strict superset (they additionally assert that constexpr is carried over to the generated hessian/jacobian), and the plain non-constexpr hessian and jacobian paths are covered by Hessian/Hessians.C and Jacobian/Jacobian.C.

Verified with the full lit suite on a Debug build: 215 tests, 0 failures.


vgvassilev (approved):
I love PRs with negative code addition ratio: LGTM!

PR #1988, opened 2026-08-18
Enable the constant folder in forward mode

guitargeek:
ConstantFolder has been dead code: its only caller constructed it with m_Enabled hardcoded to false. Forward mode carries one seeded direction, so most tangents reaching an operator are a constant zero, and the product/quotient rules expand into `0 * b + a * 0` chains that keep the whole primal subtree alive -- in a hessian, once per direction. Folding collapses them to a zero that cascades through the rest of the expression. This is 1.84x on hessian generation for a 41-parameter RooFit likelihood, and the generated code also evaluates faster.

Enabling the folder safely needs a set of accompanying fixes:

- Build only the product-rule terms whose tangent is not a constant zero, and store only the operands the surviving terms read; storing an operand only a dropped term reads would keep the primal alive as a dead temporary. When both tangents are zero, the whole rule -- for division including the denominator -- is a plain zero and nothing is stored. Term selection also keeps every fold() input safe to re-evaluate, which the bail-out below relies on.

- Refuse a fold whose result is an lvalue when the input was not. A tangent like `_d_x + 0` folding to the bare `_d_x` next to a prvalue primal argument changes which parameters the callee's pullback is requested for in reverse-over-forward; RooFit's poissonIntegral fails to build without this.

- Never treat an expression with side effects as a constant: a zero tangent with a side effect must not be dropped, and a constant-valued subexpression with a side effect must not be replaced by its value.

- Synthesize a bool literal when a folded constant is boolean. An IntegerLiteral of type bool is malformed and crashes clang's StmtPrinter (it finds no integer suffix for the type).

- Keep ParenExpr around any loosely binding subexpression the fold hands back (conditionals, overloaded operators), not just binary operators, so dumped source reparses to the AST it came from.

Test expectations are regenerated from the emitted derivatives, and ForwardMode/ConstantFolding.C pins each of the new branches. Verified with the full lit suite against both the release and the debug build (live AST-integrity asserts).

### DIFF
--- a/lib/Differentiator/BaseForwardModeVisitor.cpp
+++ b/lib/Differentiator/BaseForwardModeVisitor.cpp
@@ -1528,50 +1528,88 @@ BaseForwardModeVisitor::VisitBinaryOperator(const BinaryOperator* BinOp) {
   StmtDiff Ldiff = Visit(BinOp->getLHS());
   StmtDiff Rdiff = Visit(BinOp->getRHS());
 
+  // Fold the derivative expression algebraically. Forward mode carries one
+  // seeded direction, so most tangents reaching an operator are a constant
+  // zero and the product/quotient rules expand into `0 * b + a * 0` chains
+  // that keep the whole primal subtree alive. Folding them away is what stops
+  // an n-direction request (a hessian) from generating n copies of the parts
+  // of the function that the direction does not reach.
   ConstantFolder folder(m_Context);
   auto opCode = BinOp->getOpcode();
   Expr* opDiff = nullptr;
 
   // The operand primals are also parented by the forward value; clone each so
-  // the derivative expression owns a distinct subtree.
-  auto deriveMul = [this](StmtDiff& Ldiff, StmtDiff& Rdiff) {
-    Expr* LHS = BuildOp(BO_Mul, BuildParens(Ldiff.getExpr_dx()),
-                        BuildParens(CloneNode(Rdiff.getExpr())));
+  // the derivative expression owns a distinct subtree. A term whose tangent
+  // factor is a constant zero is not built at all rather than left for the
+  // folder to drop: the caller then skips storing the primal that term reads,
+  // and fold() may hand back its whole input unfolded, so the expression given
+  // to it must never contain a primal that is unsafe to re-evaluate.
+  auto mulTerm = [this](Expr* LHS, Expr* RHS) {
+    return BuildOp(BO_Mul, BuildParens(LHS), BuildParens(RHS));
+  };
 
-    Expr* RHS = BuildOp(BO_Mul, BuildParens(CloneNode(Ldiff.getExpr())),
-                        BuildParens(Rdiff.getExpr_dx()));
-
+  auto deriveMul = [this, &mulTerm](StmtDiff& Ldiff, StmtDiff& Rdiff,
+                                    bool dropsLTerm = false,
+                                    bool dropsRTerm = false) {
+    Expr* LHS = dropsLTerm
+                    ? nullptr
+                    : mulTerm(Ldiff.getExpr_dx(), CloneNode(Rdiff.getExpr()));
+    Expr* RHS = dropsRTerm
+                    ? nullptr
+                    : mulTerm(CloneNode(Ldiff.getExpr()), Rdiff.getExpr_dx());
+    if (!LHS)
+      return RHS;
+    if (!RHS)
+      return LHS;
     return BuildOp(BO_Add, LHS, RHS);
   };
 
-  auto deriveDiv = [this](StmtDiff& Ldiff, StmtDiff& Rdiff) {
-    Expr* LHS = BuildOp(BO_Mul, BuildParens(Ldiff.getExpr_dx()),
-                        BuildParens(CloneNode(Rdiff.getExpr())));
+  auto deriveDiv = [this, &mulTerm](StmtDiff& Ldiff, StmtDiff& Rdiff,
+                                    bool dropsRTerm = false) {
+    // The `dL * R` term is always built: when both tangents are zero the
+    // caller emits a plain zero instead of calling here, so the denominator
+    // survives and R is stored -- a zero dL term is then safe to keep for the
+    // folder.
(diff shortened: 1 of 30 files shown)


lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -1528,30 +1528,55 @@ BaseForwardModeVisitor::VisitBinaryOperator(const BinaryOperator* BinOp) {
...
-  // the derivative expression owns a distinct subtree.
-  auto deriveMul = [this](StmtDiff& Ldiff, StmtDiff& Rdiff) {
-    Expr* LHS = BuildOp(BO_Mul, BuildParens(Ldiff.getExpr_dx()),
-                        BuildParens(CloneNode(Rdiff.getExpr())));
-
-    Expr* RHS = BuildOp(BO_Mul, BuildParens(CloneNode(Ldiff.getExpr())),
-                        BuildParens(Rdiff.getExpr_dx()));
-
+  // the derivative expression owns a distinct subtree. A term whose tangent
+  // factor is a constant zero is not built at all rather than left for the
```
vgvassilev, line 1543: I wonder if the clad guardrails forced that decision by complaining we leave unused ast nodes?

test/CUDA/ForwardMode.cu
```diff
@@ -89,21 +89,21 @@ double fn1(double i, double j) {
 // CHECK-NEXT:     double *_d_device_c = nullptr;
 // CHECK-NEXT:     double *device_c = nullptr;
 // CHECK-NEXT:     unsigned long _t0 = sizeof(double);
-// CHECK-NEXT:     ValueAndPushforward<cudaError_t, cudaError_t> _t1 = clad::custom_derivatives::cudaMalloc_pushforward(&device_a, n * _t0, &_d_device_a, _d_n * _t0 + n * sizeof(double));
+// CHECK-NEXT:     ValueAndPushforward<cudaError_t, cudaError_t> _t1 = clad::custom_derivatives::cudaMalloc_pushforward(&device_a, n * _t0, &_d_device_a, _d_n * _t0 + n * 8UL);
```
vgvassilev: That is potentially problematic because makes the code not portable across platforms where sizeof(X) varies.

guitargeek: God point, fixed in the next push!

lib/Differentiator/BaseForwardModeVisitor.cpp
guitargeek: No, the integrity checks only flag sharing (a node with two parents, or one spliced from the still-live primal AST); an orphaned `0 * b` subtree left behind for the folder to delete would be invisible to them. What forces the up-front decision is the stores: to build that term at all we'd first  `StoreAndRef` the primal operand it reads, and the folder can only delete expressions, not the already-emitted `_t0 = <primal>;` statement. So the primal would survive as a dead temporary, which is exactly what the fold is meant to prevent in the hessian case. A second constraint: `fold()` can hand back its whole input unfolded (the value-category bail-out), so everything passed to it must be safe to re-evaluate in full; deciding terms up front keeps unsafe primals out of that expression.

vgvassilev: I thought I have also implemented a check for the dead nodes but now that I think about it there was some level of back and forth maybe nothing landed... For cling/clang-repl that's something we need to find out because I assume the ConstantFolder will contribute quite a bit to the overall session memory...

guitargeek: Indeed, the check for dead nodes never landed.

I also asked Claude to do some benchmarking to answer the session memory question, report below.

So it looks like this PR actually *helps* to reduce session memory.

---------------------

I looked into the session-memory question with numbers. Two observations first:

1. Clang AST nodes are bump-allocated in the `ASTContext` arena and never freed
   individually, so a dead-node check couldn't reclaim anything — the only
   lever is how many nodes get allocated in the first place.
2. This PR mostly *avoids allocating* dead nodes rather than creating them:
   the term selection in `VisitBinaryOperator` never builds a product-rule
   term whose tangent is a constant zero (and skips its `StoreAndRef`), so the
   dropped subtrees — including the cloned primal operands — are never
   allocated. What the identity folds do orphan is small glue (a
   `BinaryOperator`, a zero literal, an implicit cast) that would have been
   allocated *and kept reachable from the emitted derivative* without the
   folder.

To quantify it, I compiled a product/quotient-heavy `clad::hessian` over a
12-element array parameter with the same clad build and folding toggled by a
temporary env var, and compared `-Xclang -print-stats` output against a
baseline TU with no differentiation request. Clad-attributable allocations:

|                 | folder on | folder off |
|-----------------|-----------|------------|
| Stmt/Expr nodes | 39,455    | 72,515     |
| Stmt/Expr bytes | 1.12 MB   | 2.06 MB    |
| Decl bytes      | 105 KB    | 142 KB     |

The arena totals count every allocation, orphaned or not, so the ~46%
reduction is net of any dead nodes the fold leaves behind. The emitted
derivative — the part that stays live for the whole cling session — shrinks by
about the same factor, and the effect is per forward direction, so it grows
with the parameter count. So for clang-repl/cling the folder should be a clear
memory win; the residual orphaned glue could only be avoided by mutating nodes
in place, which the folder already does where the single-parent invariant
allows.

vgvassilev: Thanks for doing this analysis.

vgvassilev (approved):
Lgtm!

guitargeek:
CI is saying:
```txt
   No platform saw these fail with the change reverted, so they cannot be
  its regression test:
  
    test/CUDA/ForwardMode.cu
  
  Either the shape they use does not reach the changed code, or they pin
  behavior this change does not affect -- in which case they belong in
  their own commit.
```
but this is a false positive here, because the change needed to get its reference results updated. Maybe it's because the runner of the "Tests observe the change" workflow doesn't run the tests in `test/CUDA`?

guitargeek:
One more force-push to address the coverage report

PR #1994, opened 2026-08-19
Ci observes change cuda

vgvassilev:
test/CUDA files can never be shown to fail without the change that updates them: the only rows able to run them are excluded from the observes-change check, and on every other row lit skips them and the checker reads that skip as a pass. #1988 was told its updated CUDA reference results were not evidence about the change they came with.

The first commit stops a row that never ran a test from voting on it, so an unmet REQUIRES: reports "unmeasured" instead of "cannot fail" — this affects all 26 REQUIRES:-gated tests (16 cuda-runtime, 6 Enzyme, 2 OpenMP, 1 clang-repl, 1 asserts). The second lets the CUDA rows run the check and makes the verdict wait for the self-hosted matrix.

### DIFF
--- a/.github/actions/observes-change/observes-change-verdict.py
+++ b/.github/actions/observes-change/observes-change-verdict.py
@@ -3,7 +3,9 @@
 
 A test counts as observing the change if it failed at baseline on any row,
 since a defect can be invisible everywhere but under valgrind. The check
-fails only for a touched test that failed nowhere.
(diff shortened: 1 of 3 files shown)


guitargeek (commented):
Thank you!

vgvassilev:
Moving forward -- github seems to have instabilities and the cancelled jobs were stuck. This PR does not really require full runs.

PR #1996, opened 2026-08-20
Resolve each hessian direction's custom forward derivative independently

guitargeek:
Hessian mode derives the function once in forward mode per requested direction, and the planner builds those forward requests by mutating one DiffRequest in a loop. LookupCustomDerivativeDecl writes CustomDerivative only when it finds one, so the field survives into the next iteration: a direction whose custom derivative exists leaks it into every direction after it.

A hessian of f(x, y) with a user-registered f_darg0 but no f_darg1 then takes both of its rows from f_darg0 -- the second row is the gradient of df/dx where it should be the gradient of df/dy -- and clad emits no diagnostic. Reset the field before each lookup so every direction resolves its own.

Existing coverage misses this because test/Hessian/BuiltinDerivatives.C registers custom derivatives for both directions of f7, where a leak is indistinguishable from a correct lookup. The new test registers one for the first direction only and scales it by 1000, so the leak shows in both the generated f_hessian body and the hessian values: without the fix the second row comes out {2000, 6000} instead of {6, 0}.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -1764,6 +1764,10 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
               forwRequest.UpdateDiffParamsInfo(m_Sema);
               if (!forwRequest.DVI.empty())
                 forwRequest.DVI.back().TotalCapacity = indexInterval.Finish;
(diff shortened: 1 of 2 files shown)


vgvassilev (approved):
LGTM!

PR #1997, opened 2026-08-20
Cover the dependent non-const pointer diagnostic in hessian mode

guitargeek:
A hessian w.r.t. a subset of the parameters leaves the parameters outside that subset without a tangent. Forward mode reads a missing tangent as an identically zero derivative only through a const pointee, so it rejects a dependent non-const pointer parameter instead: "dependent non-const pointer and array parameters are not supported".

Hessian mode reaches that diagnostic through the per-direction forward requests it schedules, but no test pinned it there -- test/Hessian/ ArrayErrors.C covers only the missing-index-interval error, and its pointer parameter is const. Add the case, so that the behavior is guarded before anything changes how hessian mode derives.

### DIFF
--- a/test/Hessian/ArrayErrors.C
+++ b/test/Hessian/ArrayErrors.C
@@ -6,9 +6,16 @@ double f(double a, const double *b) {
   return b[0] * a + b[1] * a + b[2] * a;
 }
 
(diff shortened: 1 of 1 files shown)


vgvassilev (approved):
LGTM!

PR #1998, opened 2026-08-20
Compute hessian sizes and diagnostics before deriving

guitargeek:
HessianModeVisitor::Derive walked the parameters once and did two unrelated things per parameter: it accumulated the row length of the hessian matrix and diagnosed an array parameter that came without an index interval, and it derived that parameter's second-order function. The two are interleaved, so the shape of the result is not known until the last derivative has been built, and a request that is going to be rejected still pays for the derivatives of every parameter ahead of the bad one.

Split the walk in two: one loop computes IndependentArgsSize and TotalIndependentArgsSize and emits the missing-index-interval diagnostic, and a second loop derives. Nothing else changes -- the derivation loop visits the same parameters in the same order and Merge receives the same three values.

The one observable difference is on the error path: clad::hessian(f, "a, b") with an array b no longer emits the second-order function for a before rejecting the call. The diagnostic and the exit code are unchanged, which test/Hessian/ArrayErrors.C pins.

The NOLINTNEXTLINE on the do-while is not new debt: the loop is moved verbatim, and moving it puts the line in the diff where clang-tidy's cppcoreguidelines-avoid-do-while would report it.

### DIFF
--- a/lib/Differentiator/HessianModeVisitor.cpp
+++ b/lib/Differentiator/HessianModeVisitor.cpp
@@ -12,9 +12,11 @@
 #include "clad/Differentiator/ErrorEstimator.h"
 #include "clad/Differentiator/StmtClone.h"
 
+#include "clang/AST/Decl.h"
 #include "clang/AST/Expr.h"
 #include "clang/AST/TemplateBase.h"
 #include "clang/Basic/Diagnostic.h"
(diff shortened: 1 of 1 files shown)


vgvassilev (approved):
Lgtm!

PR #2000, opened 2026-08-20
Support discrete compound assignments in reverse mode

Elvand-Lie:
### Description

ReverseModeVisitor entered its assignment path for all Clang assignment operators but explicitly handled only `=`, `+=`, `-=`, `*=`, and `/=`. Remainder, bitwise, and shift compound assignments (`%=`, `&=`, `|=`, `^=`, `<<=`, `>>=`) fell through to an `llvm_unreachable`.

This PR adds discrete compound assignment support in reverse mode:

1. **Primal Result Materialization & Lvalue Reference Preservation**: `(L op= R)` is evaluated exactly once in the forward pass inside an lvalue-reference declaration `T& _ref = (L op= R);`. A primal result override returns this reference directly as the forward expression result, preserving C++ lvalue semantics for nested operations.
2. **Reverse Value Snapshot**: Snapshot the converted post-assignment value `T _t = _ref;` (or `clad::tape<T>` inside loops) using `GlobalStoreAndRef` for use in the reverse sweep.
3. **No Operator Reconstruction**: Eliminates binary operator reconstruction during the reverse sweep, preventing double execution and avoiding Clad shared AST node integrity violations.
4. **Clean Adjoint Storage**: Restructures `oldValue` adjoint storage to avoid emitting unused `_r_d` declarations for discrete operations.
5. **Coverage**: Adds tests in `test/Gradient/CompoundBitwiseOps.C` covering all six operators, nested assignments under parent `*=`, narrowing conversions, loops, and side-effectful RHS expressions.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1034,9 +1034,23 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
             direction::reverse);
       // Materialize rev before closing the arm reverse block (Pop placement)
       // only when the conditional reverse value is consumed.
-      Expr* Rev =
-          requestArmRev ? ExprDiff.getRevSweepAsExpr() : nullptr;
+      Expr* Rev = requestArmRev ? ExprDiff.getRevSweepAsExpr() : nullptr;
       Expr* Fwd = ExprDiff.getExpr();
+      // Compositionally stabilize any side-effectful forward value under its
+      // branch only when the branch's reverse value is consumed (requestArmRev)
+      // so the side effects execute exactly once under this branch's condition,
+      // and the expression used downstream (in condExpr and any reverse
+      // calls/pullbacks) is pure. Discarded ternaries do not force tape/store.
+      if (requestArmRev && Fwd && Fwd->HasSideEffects(m_Context)) {
+        if (Fwd->isLValue() && Fwd->getObjectKind() != OK_BitField) {
+          Expr* addr = BuildOp(UO_AddrOf, Fwd);
+          Expr* storeAddr =
+              GlobalStoreAndRef(addr, /*prefix=*/"_t", /*force=*/true);
+          Fwd = BuildOp(UO_Deref, storeAddr);
+        } else {
+          Fwd = GlobalStoreAndRef(Fwd, /*prefix=*/"_t", /*force=*/true);
+        }
+      }
       Expr* Dx = ExprDiff.getExpr_dx();
       CompoundStmt* RCS = endBlock(direction::reverse);
       Stmt* ForwardResult = endBlock(direction::forward);
@@ -1074,7 +1088,8 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       addToCurrentBlock(Reverse, direction::reverse);
 
     // Arms already materialized above when requestArmRev; reuse those exprs
-    // (do not re-request rev after BuildIf -- that would mis-place discrete Pop).
+    // (do not re-request rev after BuildIf -- that would mis-place discrete
+    // Pop).
     Expr* condExpr =
         m_Sema
             .ActOnConditionalOp(noLoc, noLoc, CloneNode(condStored),
@@ -1967,26 +1982,15 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       result.updateRevSweep(moveCall);
     }
 
-    // Save cloned arg in a "global" variable, so that it is accessible from
-    // the reverse pass.
-    // For example:
-    // ```
-    // // forward pass
-    // _t0 = a;
-    // modify(a); // a is modified so we store it
-    //
-    // // reverse pass
-    // a = _t0;
-    // modify_pullback(a, ...); // the pullback should always keep `a` intact
-    // ```
-    // FIXME: Handle storing data passed through pointers and structures.
-    // FIXME: Improve TBR to handle these stores.
+    bool isDiscrete = argDiff.hasRevSweepBuild();
     bool passByRef = paramTy->isLValueReferenceType() &&
                      !paramTy.getNonReferenceType().isConstQualified();
-    if (passByRef && m_DiffReq.shouldBeRecorded(arg)) {
-      // Finalize discrete FwdWrapper before CloneNode(getExpr()). Without
-      // this, in-place upgrade never runs (argDiff dies) or runs after the
-      // clone and StoreAndRestore keeps a stale Paren(Assign).
+    bool shouldRecordRef = passByRef && m_DiffReq.shouldBeRecorded(arg);
+
+    if (isDiscrete || shouldRecordRef) {
+      // Finalize discrete FwdWrapper before CloneNode(getExpr()) or call
+      // assembly. Without this, in-place upgrade never runs (argDiff dies) or
+      // runs after the clone and CallArgs keeps a stale Paren(Assign).
       argDiff.prepareForFwdClone();
       Expr* E = argDiff.getExpr();
       // Side-effectful args (e.g. discrete assign) must run once before
@@ -2017,6 +2021,23 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
           argDiff.updateStmt(CloneNode(returnExprs[0]));
         }
       }
+    }
+
+    // Save cloned arg in a "global" variable, so that it is accessible from
+    // the reverse pass.
+    // For example:
+    // ```
+    // // forward pass
+    // _t0 = a;
+    // modify(a); // a is modified so we store it
+    //
+    // // reverse pass
+    // a = _t0;
+    // modify_pullback(a, ...); // the pullback should always keep `a` intact
+    // ```
+    // FIXME: Handle storing data passed through pointers and structures.
+    // FIXME: Improve TBR to handle these stores.
+    if (shouldRecordRef) {
       // argDiff.getExpr() is also returned below as the call argument; clone
       // it for the store/restore so the node is not shared with the call.
       StmtDiff pushPop = StoreAndRestore(CloneNode(argDiff.getExpr()));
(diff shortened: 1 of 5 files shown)


test/Gradient/CompoundBitwiseOps.C
```diff
@@ -0,0 +1,302 @@
...
+
+// 2. &= operator
+double f_and(double x, int y) {
+  int n = y;
+  n &= 7;
+  return x * n;
+}
+
+// CHECK-LABEL: f_and_grad
+// CHECK: int &_ref0 = n &= 7;
```
vgvassilev, line 27: Can we make sure we check all the bodies with CHECK-NEXT? Also the tests should be tight (almost surgical).

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3384,6 +3392,26 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
         std::tie(Ldiff, Rdiff) = std::make_pair(LCloned, RResult);
+      } else if (opCode == BO_RemAssign || opCode == BO_AndAssign ||
+                 opCode == BO_OrAssign || opCode == BO_XorAssign ||
+                 opCode == BO_ShlAssign || opCode == BO_ShrAssign) {
+        if (ResultRef && !isPointerOp) {
+          Expr* zero = getZeroInit(ResultRef->getType());
+          Expr* assign_zero = BuildOp(BO_Assign, CloneNode(ResultRef), zero);
+          addToCurrentBlock(assign_zero, direction::reverse);
+        }
+        Rdiff = Visit(R, static_cast<Expr*>(nullptr));
```
vgvassilev, line 3403:
```suggestion
        Rdiff = Visit(R);
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3384,6 +3392,26 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
             BuildOp(BO_Div, BuildParens(Rdiff.getRevSweepAsExpr()),
                     BuildParens(Ldiff.getRevSweepAsExpr()));
         std::tie(Ldiff, Rdiff) = std::make_pair(LCloned, RResult);
+      } else if (opCode == BO_RemAssign || opCode == BO_AndAssign ||
+                 opCode == BO_OrAssign || opCode == BO_XorAssign ||
+                 opCode == BO_ShlAssign || opCode == BO_ShrAssign) {
```
vgvassilev, line 3397:
```suggestion
      } else if (isDiscreteAssign) {
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3268,19 +3270,25 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+      // For pointer types or discrete ops, no need to store old derivatives.
       // ResultRef is returned as the adjoint and, for nested compound
       // assignments, reused as the next level's ResultRef; clone it for the
       // old-value store so the stores do not share the node.
-      if (indepSides)
-        oldValue = CloneNode(ResultRef);
-      else if (!isPointerOp)
-        oldValue = StoreAndRef(CloneNode(ResultRef), direction::reverse, "_r_d",
-                               /*forceDeclCreation=*/true);
+      if (!isDiscreteAssign) {
```
vgvassilev, line 3284: If `!isDiscreteAssign` do we really need to pass through the other code?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3384,6 +3392,26 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+          Expr* assign_zero = BuildOp(BO_Assign, CloneNode(ResultRef), zero);
+          addToCurrentBlock(assign_zero, direction::reverse);
+        }
+        Rdiff = Visit(R, static_cast<Expr*>(nullptr));
+        Expr* assignExpr = BuildOp(opCode, Ldiff.getExpr(), Rdiff.getExpr());
+        QualType valueType =
+            utils::getNonConstType(Ldiff.getExpr()->getType(), m_Sema);
+        QualType referenceType = m_Context.getLValueReferenceType(valueType);
+        Expr* forwardResult =
+            StoreAndRef(assignExpr, referenceType, direction::forward, "_ref",
```
vgvassilev, line 3409: Binding a reference to the result of the assignment does not work when the left side is a bit-field, and the user sees a compiler error in code they did not write.

`int &_ref0 = n %= y;` is fine for an ordinary variable. For `s.flags &= mask;` where `flags` is a bit-field, it is ill-formed: a non-const lvalue reference cannot bind to a bit-field ([class.bit]/5). The generated code will not compile, and the error will point into clad's output rather than at the line the user wrote.

Bit-fields are not common in code people differentiate, so this may be worth no more than a guard that falls back to the old path when `ResultRef` refers to one, which `FieldDecl::isBitField()` answers.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3384,6 +3392,26 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
             BuildOp(BO_Div, BuildParens(Rdiff.getRevSweepAsExpr()),
                     BuildParens(Ldiff.getRevSweepAsExpr()));
         std::tie(Ldiff, Rdiff) = std::make_pair(LCloned, RResult);
+      } else if (opCode == BO_RemAssign || opCode == BO_AndAssign ||
+                 opCode == BO_OrAssign || opCode == BO_XorAssign ||
+                 opCode == BO_ShlAssign || opCode == BO_ShrAssign) {
+        if (ResultRef && !isPointerOp) {
```
vgvassilev, line 3398: This decides whether the adjoint gets zeroed, and only one side of it is ever taken by a test.

Every function in the new test assigns to a plain local `int`: `n %= y`, `n &= 7`, `mask %= 4`. None uses an array element, a pointer dereference, or a struct member, so `isPointerOp` is false every time and the zeroing always happens. The path where it is skipped is not exercised at all.

That matters because skipping the zero is what would leave a stale adjoint behind, and a stale adjoint gives a wrong gradient rather than a crash. A function doing `arr[i] %= y` and then reading `arr[i]`, with a `CHECK-EXEC` on the result, would say whether the skip is right.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3384,6 +3392,26 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+        Rdiff = Visit(R, static_cast<Expr*>(nullptr));
+        Expr* assignExpr = BuildOp(opCode, Ldiff.getExpr(), Rdiff.getExpr());
+        QualType valueType =
+            utils::getNonConstType(Ldiff.getExpr()->getType(), m_Sema);
+        QualType referenceType = m_Context.getLValueReferenceType(valueType);
+        Expr* forwardResult =
+            StoreAndRef(assignExpr, referenceType, direction::forward, "_ref",
+                        /*forceDeclCreation=*/true);
+        valueForRevPass =
+            GlobalStoreAndRef(CloneNode(forwardResult), valueType, "_t",
```
vgvassilev, line 3412: Is this stored value ever read for a discrete operation?

The adjoint of the left side is set to zero rather than propagated, and the right side is visited with a null adjoint, so as far as I can tell nothing in the reverse sweep needs the post-assignment value. If that is right, `force=true` inside a loop pushes one value per iteration onto the tape and never pops it, which costs memory proportional to the trip count for nothing.

I could not confirm this either way from the diff, since the test's `CHECK` lines do not cover the generated loop body. If the value is needed, a sentence saying what reads it would help; if it is not, dropping the store would make the branch smaller.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3230,18 +3347,92 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+              utils::GetInnermostReturnExpr(Edx, dxExprs);
+              if (dxExprs.size() == 1)
+                Ldiff.updateStmtDx(CloneNode(dxExprs[0]));
+            }
+          } else if (E->getObjectKind() != OK_BitField) {
+            auto* storeE = GlobalStoreAndRef(BuildOp(UO_AddrOf, E));
+            Ldiff.updateStmt(BuildOp(UO_Deref, storeE));
+            if (Expr* Edx = Ldiff.getExpr_dx()) {
+              if (Edx->getObjectKind() != OK_BitField) {
+                auto* storeDx = GlobalStoreAndRef(BuildOp(UO_AddrOf, Edx));
```
vgvassilev: This adds a pointer store for every assignment whose left side has side effects, which includes every std::vector element write, and a tape in loops (see the STLCustomDerivatives.C changes). No test's number changes because of it. If a case needs it, an adjoint whose designator mutates, like `*_d_p++`, put that case in a test and store only then. The member-of-subscript branch above already gates on the designator's side effects.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3230,18 +3347,92 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+                      BuildArraySubscript(CloneNode(ASEdx->getBase()), idxsDx);
+                  Expr* newEdx = utils::BuildMemberExpr(
+                      m_Sema, getCurrentScope(), newBaseDx, fieldName);
+                  // Visit(L) may already have emitted an AddAssign whose LHS is
+                  // the old adjoint; rewrite those reverse stmts onto the
+                  // stabilized adjoint so reverse does not re-run i++ / p++.
+                  Expr* oldEdx = Edx;
+                  for (Stmt*& S : getCurrentBlock(direction::reverse)) {
+                    if (auto* BO = dyn_cast<BinaryOperator>(S)) {
+                      if (BO->getLHS() == oldEdx)
```
vgvassilev: Patching reverse statements that were already emitted, by pointer identity, is fragile. It catches only a top-level BinaryOperator whose left side is exactly this node. A parenthesized or nested one is missed, and the reverse sweep re-runs `i++`. Stabilizing the designator before Visit(L) emits anything removes the patch.

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev, line 3428: Problem: this branch now stores the adjoint's address for every assignment whose left side has side effects. A std::vector element write counts, because operator[] is a call, so `a[1] = x * x` gains a pointer store per assignment and a tape inside loops; the STLCustomDerivatives.C diff shows both. Re-reading `_d_a[1]` in the reverse sweep was already correct, and no test's numeric result changes because of this store.

How to address it: store the adjoint's address only when re-evaluating it would repeat a side effect, that is when the designator itself mutates something, like `_d_p++` or `_d_b[i++]`. The member-of-subscript branch above already checks the base and the index separately for side effects. Apply that check here instead of HasSideEffects on the whole expression, and add one test where the designator mutates, so the store has a test that observes it.

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev, line 3396: Problem: this loop rewrites reverse statements that Visit(L) already emitted, by comparing each statement's left side with the old adjoint pointer. That matches only a top-level BinaryOperator whose left operand is exactly that node. If the adjoint is wrapped in parentheses or a cast, or the statement sits inside a compound statement, nothing matches, no error is raised, and the reverse sweep keeps the old adjoint, which re-runs `i++`.

How to address it: stabilize the designator before visiting the left side. Compute stabBase and stabIdx first, build the stabilized primal and adjoint expressions from them, and hand those to the assignment path, so Visit(L) emits its statements against the stabilized adjoint from the start. Then there is nothing to patch afterwards and this loop can go.

vgvassilev (commented):
Problem: this PR carries four separate changes, and two of them change the code clad generates for programs that never use a discrete operator. The updated expectations in Assignments.C and STLCustomDerivatives.C come from those two. First, every assignment whose left side has side effects now also stores the address of its adjoint. Second, a by-reference call argument with side effects is now emitted once before the call. The third reverse-sweep slot in StmtDiff, and the lazy paren, cast, unary and comma visitors, are infrastructure: they change no output on their own. Reviewed as one change this is hard to follow, and a later regression bisects to a 2600-line commit.

How to address it: split into a series. (1) The StmtDiff slot and the lazy visitors as an NFC commit, with existing tests unchanged. (2) The adjoint-address store, with a test whose gradient is wrong without it. (3) The call-argument change, with the same kind of test. (4) The six operators with their tests: one function per operator plus the loop and nesting cases. The other forty functions in CompoundBitwiseOps.C exercise the machinery from (1) to (3) and can move into those commits. The two inline notes below are about (2) and will move with it.

I did not build or run anything. CI is green on all 44 checks, and the baseline check confirms the new test fails on master.

PR #2001, opened 2026-08-20
Diagnose non-void functions in Jacobian mode

### LINKED ISSUE #1306: Jacobian Generates `return` in `void` Function
# Jacobian Generates `return` in `void` Function

`clad::jacobian(...)` applied to a scalar-returning function causes invalid C++ to be generated.

## Problem

Clad assumes the function returns `void`, but it actually returns a scalar (`double`),
so the generated `*_jac` function has a `void` return type **but contains a `return` statement**, causing a compiler error.
Users get a cryptic C++ compiler error later, rather than a clear Clad diagnostic
```error: void function 'model_and_loss_jac' should not return a value [-Wreturn-type]```

##  Minimal Reproducer

```cpp

double model_and_loss(double x0, double x1, double w0, double w1, double b, double t, double* out) {
    double y = w0 * x0 + w1 * x1 + b;
    double diff = y - t;
    return 0.5 * diff * diff; 
}

int main() {
    clad::matrix<double> J(1, 3);
    double loss_output[1];
    clad::jacobian(model_and_loss).execute(1.0, 2.0, 0.5, -0.3, 0.1, 1.5, loss_output, &J);
    return 0;
}

### PR DESCRIPTION
**Pull request by Elvand-Lie**
### Description

Clad's Jacobian API generates a void derivative function and exposes outputs through pointer, reference, or array parameters.

Requesting `clad::jacobian` for a non-void primal function previously proceeded into derivative generation and emitted invalid C++, resulting in a cryptic downstream compiler error.

This PR adds a direct diagnostic in `DerivativeBuilder::Derive` that rejects non-void primal functions before invalid C++ is generated, explaining that Jacobian mode currently requires a void-returning primal function and that outputs must be provided through pointer, reference, or array parameters.

- **Current Contract**: Strict void-returning primal requirement.
- **Coverage**: Adds tests in `test/Jacobian/NonVoidReturnDiagnostic.C` covering scalar returns, output pointers with scalar returns (reproducing #1306), and status-returning functions to document current behavior.
- **Follow-up Work**: Selective status-return support, if desired by maintainers, involves separate return-control-flow and output-classification semantics and can be evaluated as a follow-up feature.

### Related Issues
Fixes #1306

### CODE REVIEW

**Comment by Elvand-Lie** [comment]
Before I finalize the #1306 diagnostic, I would like to confirm the intended Jacobian return-value contract.

The current patch rejects every non-void primal. That handles differentiable scalar returns, for which the generated Jacobian ABI has no return-derivative output slot, but it also rejects status-returning C-style functions such as:

    int status_output(double x, double* out) {
      *out = x * x;
      return 0;
    }

I propose the following narrower rule:

- keep the generated Jacobian function void-returning;
- reject differentiable non-void returns with a clear diagnostic;
- allow bool, integral, or enum status returns when Jacobian mode already recognizes a differentiable output parameter;
- evaluate the status-return expression for side effects and lower it to a bare return, preserving early-return control flow.

This avoids inventing an ABI slot for scalar-return derivatives while preserving common status-code APIs.

Would you prefer this selective status-return policy, or should Jacobian primal functions be strictly void-returning?

**Comment by Elvand-Lie** [comment]
After reviewing the merged vector-forward Jacobian design and its void-returning derivative ABI, I am keeping this PR scoped to the current void-returning primal contract. Selective status-return support would require separate return-control-flow and output-classification semantics, so I will leave that as follow-up feature work unless maintainers prefer otherwise.

File: lib/Differentiator/DerivativeBuilder.cpp [hunk 1]
```diff
@@ -694,6 +694,23 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
...
+      // incompatible return.
+      if (request.Mode == DiffMode::jacobian &&
+          !FD->getReturnType()->isVoidType()) {
+        SourceLocation L = request.CallContext->getBeginLoc();
+        diag(DiagnosticsEngine::Error, L,
+             "jacobian mode currently requires function %0 to return void; "
+             "provide differentiable outputs through pointer, reference, or "
+             "array "
+             "parameters")
+            << FD;
```
vgvassilev: Optional: a note at the return type would take the user to the line they need to change. The error names the function but points at the clad::jacobian call. After the error, something like `diag(DiagnosticsEngine::Note, FD->getReturnTypeSourceRange().getBegin(), "%0 declared here with return type %1") << FD << FD->getReturnType();`, plus a matching expected-note line in the test.

vgvassilev:
Can we squash the changes into a single commit with a proper commit log.

lib/Differentiator/DerivativeBuilder.cpp
```diff
@@ -591,6 +591,29 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
...
+        SourceLocation NoteLoc = FD->getReturnTypeSourceRange().getBegin();
+        if (NoteLoc.isInvalid())
+          NoteLoc = FD->getLocation();
+        diag(DiagnosticsEngine::Note, NoteLoc,
+             "%0 declared here with return type %1")
+            << FD << FD->getReturnType();
+        return {};
+      }
```
vgvassilev, line 615: We should probably move this check earlier in the diffplanner -- likely around ProcessInvocationArgs.

vgvassilev: ping.

vgvassilev (approved):
LGTM!

Elvand-Lie:
Hi @vgvassilev, this is updated against the latest master and CI is green. The previous approval predates the latest head, so if possible could this PR have one final look/approval? Thanks.

PR #2004, opened 2026-08-21
Decide in the planner which calls a forward direction differentiates

guitargeek:
**Follows up on https://github.com/vgvassilev/clad/pull/1984#issuecomment-5330149662**

Forward mode worked out for itself whether a call could contribute to the derivative: VisitCallExpr folded the tangents it had just built and, when they were all zero, emitted the plain call with a zero derivative. Two FIXMEs at that site asked for the decision to move to the planner.

The planner now runs varied analysis for forward-mode requests, seeded from the direction the request differentiates along, and VariedAnalyzer records for each call whether any argument is varied. DiffRequest::shouldHavePushforward answers the question the visitor used to answer itself; a call the analysis never saw answers "differentiate it". Only forward-mode requests record and consume this: the requests -enable-va analyzes are unseeded or seeded for a caller, and their verdicts would zero out direction-agnostic pushforwards. A body without a plain call skips the analysis, and a hessian analyzes each parameter once, not once per array index. The record is monotone within a run, so later passes over a loop body can only add varied calls, never lose them.

Activity does not care whether a tangent is a const-qualified zero, so the elision now fires for plain parameters too; four tests lose the pushforward of the direction they do not seed. The tangent fold stays as an index-level refinement: a direction seeding p[1] leaves the whole of p varied, and only the tangents know f(p[0]) is inactive.

A "not varied" verdict now deletes derivatives by default, so the analyzer is made conservative where it does not track state: a DeclRefExpr naming a function (an assert, reachable on master with -enable-va), a call through a function pointer (a crash on a null getDirectCallee()), and member state reached through `this`, which is not tracked per field and can carry the derivative -- this also covers functors differentiated with respect to a field, whose FieldDecl the seed set cannot hold. ActivityForward.cpp pins these with and without -enable-va, which used to run the forward analysis unseeded; ZeroTangentCalls.C pins the once-per-parameter array hessian.

### DIFF
--- a/include/clad/Differentiator/DiffPlanner.h
+++ b/include/clad/Differentiator/DiffPlanner.h
@@ -147,6 +147,9 @@ private:
   mutable struct ActivityRunInfo {
     std::set<const clang::VarDecl*> VariedDecls;
     std::set<const clang::Stmt*> VariedS;
+    /// Whether a call of the primal has a varied argument. Only filled for
+    /// requests varied analysis has run on; see shouldHavePushforward().
+    llvm::DenseMap<const clang::CallExpr*, bool> VariedCalls;
     bool HasAnalysisRun = false;
   } m_ActivityRunInfo;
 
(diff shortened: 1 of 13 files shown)


vgvassilev (approved):
LGTM!

PR #2024, opened 2026-08-24
Let lit tell compiling CUDA apart from running it, and compile it in CI.

vgvassilev:
cuda-runtime looks for libcudart.so, and every CUDA test asks for it, so a machine with the CUDA headers but no runtime library and no device is treated as having no CUDA at all and skips the lot -- including the compile half, which such a machine runs perfectly well. That is every hosted runner, so nothing outside the two self-hosted rows has ever compiled a .cu, and the step that was supposed to give a hosted row CUDA could not run at all: its condition read `!matrix.os == 'self-hosted'`, which `!` binding tighter than `==` makes always false.

Split the question. cuda-compile is available wherever the headers are and carries what a compile needs; cuda-runtime keeps libcudart.so and what linking needs. A test asks for cuda-compile, %cudaldflags degrades to -fsyntax-only when there is nothing to link against, and only the line that runs the built binary is spelled out as conditional -- lit wraps each RUN in a brace group, so a substitution cannot swallow the trailing FileCheck.

Provision the headers on three hosted rows, at 12.6, 12.8 and 13.0, since version skew is where CUDA breaks and headers cost 19 MB. Compiling against headers alone means the release is one CI picked, so a compiler that does not recognise it says nothing about the code and the warning is silenced -- only there; where a real toolkit is installed the mismatch is worth hearing. ubu22-clang13-runtime13-cuda loses its flag rather than gaining a version, clang 13 predating all three, and the self-hosted rows keep theirs, still linking and running for real.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,9 +41,9 @@ jobs:
           - { name: ubu24-clang16-runtime17, os: ubuntu-24.04, compiler: clang-16, clang-runtime: '17' }
           - { name: ubu24-gcc14-runtime18, os: ubuntu-24.04, compiler: gcc-14, clang-runtime: '18' }
           - { name: ubu24-clang17-runtime19, os: ubuntu-24.04, compiler: clang-17, clang-runtime: '19' }
-          - { name: ubu24-clang19-runtime20, os: ubuntu-24.04, compiler: clang-19, clang-runtime: '20' }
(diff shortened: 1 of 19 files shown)


vgvassilev:
[Tests observe the change (pull_request)](https://github.com/vgvassilev/clad/actions/runs/32811017489/job/97698100393?pr=2024) is a false positive -- this is mostly refactoring of the ci rather than anything else.

PR #2025, opened 2026-08-24
Report what an analysis left in the derivative, where it left it.

vgvassilev:
Clad's analyses decide what a derivative has to keep, and a value one of them cannot prove dead is paid for on every call. Nothing said which values those were. The cost appears in generated code, and generated code had no position a diagnostic could name: every node reaches for the same placeholder, the start of the main file, so a caret about it lands on the user's first line.

Give the code somewhere to be pointed at. A derivative is rendered into a buffer registered with the SourceManager, and each of its statements can then be named by the line and column its text occupies there -- the same mechanism that backs interactive input in cling and clang-repl, and macro expansions in Swift. The rendering is built on first use and cached per function, so a compilation that reports nothing never pays for one, and it is entered from the differentiation that asked for the derivative, so the place clang prints above a diagnostic is the line a reader would edit.

On top of that, -Rclad-analysis=<name>. It is the -Rpass-missed analogue for an AST-level plugin, which clang's own remark machinery cannot serve because that runs in the backend over LLVM IR and never sees one. Each remark points at the statement still in the derivative and carries a note back to the expression in the user's code whose value is being kept -- the half they can act on. Why the value survived is a separate sentence for an analysis that ran and could not prove it dead and for one that was switched off; saying the former when the latter is true would be false.

-fdump-generated-source prints the rendering and the position of every statement in it, which is how the mapping is checked without a diagnostic to hang it on.

### DIFF
--- a/docs/userDocs/source/user/DevelopersDocumentation.rst
+++ b/docs/userDocs/source/user/DevelopersDocumentation.rst
@@ -158,6 +158,71 @@ temporarily the differences in the produced outputs with:
    cmake --build . --target check-clad-execonly
 
 
+Inspecting the generated code
+=============================
+
+Clad builds the derivative as an AST, so its statements belong to no file and
+have nothing a diagnostic can point at. Clad renders a derivative into a
+buffer it registers with the ``SourceManager``, which gives every generated
+statement a line and a column, and two switches report through it. Both are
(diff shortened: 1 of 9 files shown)


vgvassilev:
The code coverage misses are pre-existing and if we keep adding as part of this pr it will change the nature of the pr itself.

PR #2029, opened 2026-08-25
Treat inaccessible copy constructors as non-copyable

a-elrawy:
Treat private/protected copy constructors as non-copyable in `utils::isCopyable`. Add a reverse-mode test so `ReverseModeVisitor` moves out of `constructor_reverse_forw` instead of copying.


vgvassilev (approved):
Lgtm!

PR #2030, opened 2026-08-26
Make CUDA builtin variables passive

Vedant2005goyal:
CUDA built-in variables such as threadIdx, blockIdx, blockDim, and gridDim are hardware execution context variables and should not be treated as active. This PR marks them as passive in Activity and TBR analysis, avoiding unnecessary adjoint allocations and tape entries.

### DIFF
--- a/include/clad/Differentiator/CladUtils.h
+++ b/include/clad/Differentiator/CladUtils.h
@@ -540,6 +540,10 @@ namespace clad {
     /// includes.
     clang::Expr* BuildEnzymeActivityMarkerRef(clang::Sema& semaRef,
                                               llvm::StringRef name);
+    /// Returns true if the expression represents a CUDA built-in variable
+    /// like threadIdx, blockIdx, blockDim, or gridDim.
+    bool isCUDABuiltinVariable(const clang::Expr* E,
+                               const clang::ASTContext& Context);
     } // namespace utils
     } // namespace clad
 
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -1930,6 +1930,23 @@ namespace clad {
       } while (E->IgnoreImplicit() != E);
       return false;
     }
+    bool isCUDABuiltinVariable(const clang::Expr* E,
+                               const clang::ASTContext& Context) {
+      if (const auto* DRE = clang::dyn_cast<clang::DeclRefExpr>(E)) {
+        const clang::ValueDecl* VD = DRE->getDecl();
+        const clang::IdentifierInfo* II = VD->getIdentifier();
+        if (!II)
+          return false;
+        llvm::StringRef Name = II->getName();
+
+        if (Name == "threadIdx" || Name == "blockIdx" || Name == "blockDim" ||
+            Name == "gridDim") {
+          const clang::SourceManager& SM = Context.getSourceManager();
+          return SM.isInSystemHeader(VD->getLocation());
+        }
+      }
+      return false;
+    }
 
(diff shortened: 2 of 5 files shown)


include/clad/Differentiator/CladUtils.h
```diff
@@ -538,6 +538,10 @@ namespace clad {
     /// matches by name to decide which arguments are differentiated.
     /// They are declared in EnzymeBuiltins.h, which Differentiator.h
     /// includes.
+    bool isCUDABuiltinVariable(const clang::Expr* E,
+                               const clang::ASTContext& Context);
+    /// Returns true if the expression represents a CUDA built-in variable
+    /// like threadIdx, blockIdx, blockDim, or gridDim.
     clang::Expr* BuildEnzymeActivityMarkerRef(clang::Sema& semaRef,
                                               llvm::StringRef name);
```
vgvassilev: The doc comments are crossed. The comment above this line documents BuildEnzymeActivityMarkerRef, and the comment for isCUDABuiltinVariable sits below it, where Doxygen attaches it to the Enzyme helper. Keep each comment directly above its own declaration:
```suggestion
    clang::Expr* BuildEnzymeActivityMarkerRef(clang::Sema& semaRef,
                                              llvm::StringRef name);
    /// Returns true if the expression represents a CUDA built-in variable
    /// like threadIdx, blockIdx, blockDim, or gridDim.
    bool isCUDABuiltinVariable(const clang::Expr* E,
                               const clang::ASTContext& Context);
```

lib/Differentiator/CladUtils.cpp
```diff
@@ -1917,6 +1917,25 @@ namespace clad {
...
       return false;
     }
+    bool isCUDABuiltinVariable(const clang::Expr* E,
+                               const clang::ASTContext& Context) {
+      if (const auto* ME = clang::dyn_cast<clang::MemberExpr>(E))
+        E = ME->getBase()->IgnoreParenImpCasts();
+
+      if (const auto* DRE = clang::dyn_cast<clang::DeclRefExpr>(E)) {
+        const clang::ValueDecl* VD = DRE->getDecl();
+        llvm::StringRef Name = VD->getName();
```
vgvassilev: NamedDecl::getName() asserts when the name is not a plain identifier, and both callers can hand this a DeclRefExpr whose name is not one. Neither analyzer overrides TraverseUserDefinedLiteral, so the default traversal reaches the DeclRefExpr naming a literal's operator"" (a `1.0i` from <complex>, or a chrono literal), and this call is now the first thing that runs on it. VariedAnalyzer used to be protected by its dyn_cast<VarDecl> check, which the new early return sits in front of; TBRAnalyzer went through getIDSequence, which bails on a non-variable. The assert is inline in Decl.h, so any Debug build of clad trips it. getIdentifier() returns null for such a name instead:
```suggestion
        const clang::ValueDecl* VD = DRE->getDecl();
        // getName() asserts unless the name is a plain identifier, and the
        // callee of a UserDefinedLiteral is a DeclRefExpr naming operator"".
        const clang::IdentifierInfo* II = VD->getIdentifier();
        if (!II)
          return false;
        llvm::StringRef Name = II->getName();
```

vgvassilev:
@Vedant2005goyal, the failure means that the test did not fail without your new code.

Vedant2005goyal:
> @Vedant2005goyal, the failure means that the test did not fail without your new code.

You are right actually earlier test dont contain the -enable-va so the varied analysis part was not working but now it should pass it

vgvassilev (approved):
Lgtm!

PR #2034, opened 2026-08-27
[docs] update supported llvm versions and fix spelling errors

hardikxk:
Updated the llvm version for clad as per the CMakeLists.txt. There are also some spelling errors like duplicated words or incomplete ones which are fixed in this patch.

NFC change only modifying the docs.


vgvassilev (approved):
LGTM! Thank you, @hardikxk!

PR #2040, opened 2026-08-30
Declare only clad-internal derivatives inline

guitargeek:
863e29dd ("Declare clad's derivatives inline") declared every derivative inline so that CodeGen drops the ones nothing calls, like a hessian's intermediate pushforward. But an inline definition must also be emitted by every translation unit that references it. In an interpreter the unit that requests a derivative and a later unit naming it are separate, so each such unit runs CodeGen over the whole pullback tree again.

Only the root of the request graph is exposed to this, being the one derivative whose name clad hands back to the caller. Give it its primal's inline-ness again -- the property that decides whether a second unit naming it is a redefinition in the first place -- and leave every derivative clad calls but never names discardable, as 863e29dd made them. The intermediate pushforward is still dropped, so hessian generation keeps what that commit bought it.

The CHECK lines for root derivatives lose the inline specifier again; Regressions/hessian-intermediate-codegen.cpp still checks that the intermediate pushforward carries it and that nothing referencing it fails to link.

### DIFF
--- a/lib/Differentiator/DerivativeBuilder.cpp
+++ b/lib/Differentiator/DerivativeBuilder.cpp
@@ -97,6 +97,17 @@ static void registerDerivative(Decl* D, Sema& S, const DiffRequest& R) {
     if (std::any_of(Previous.begin(), Previous.end(), definedNotInline))
       dFD->setInlineSpecified(false);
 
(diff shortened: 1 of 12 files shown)


vgvassilev (approved):
LGTM!

PR #2047, opened 2026-09-04
Fix operand cloning in pointer arithmetic 

Vedant2005goyal:
Fix operand cloning in pointer arithmetic to clone only the primal operand that is actually shared. Previously only the right operand was cloned unconditionally which could miss shared operands on the left and incorrectly clone newly created derivative nodes leading to invalid AST clones and missing adjoint assignments for pointer allocations.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -3442,9 +3442,10 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
         Expr* derivedL = nullptr;
         Expr* derivedR = nullptr;
         ComputeEffectiveDOperands(Ldiff, Rdiff, derivedL, derivedR);
-        // derivedR is the scalar offset, already used by the forward op above;
-        // clone so the derivative op does not share it.
-        derivedR = CloneNode(derivedR);
+        if (derivedL == Ldiff.getExpr())
+          derivedL = CloneNode(derivedL);
+        if (derivedR == Rdiff.getExpr())
+          derivedR = CloneNode(derivedR);
         if (opCode == BO_Sub)
           derivedR = BuildParens(derivedR);
         return StmtDiff(op, BuildOp(opCode, derivedL, derivedR),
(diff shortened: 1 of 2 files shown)


vgvassilev:
We need a human in the loop as the owner of the code. Can you remove Claude as a co-author. It is a tool not a human.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3455,8 +3463,18 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+        // As above, clone only the primal operand shared with the forward
+        // op. Here that can only ever be the right one: the substitution
+        // happens on the side that is not a pointer, and an assignment whose
+        // left side is not a pointer cannot have a pointer on its right.
+        // `p = new T[n]` reaches here with derivedR holding the adjoint
+        // allocation this expression just built; StmtClone has no case for a
+        // CXXNewExpr, so cloning it asserted in an assertions build and,
+        // without them, produced a type-less node that BuildOp rejected --
+        // dropping the whole adjoint assignment and leaving the adjoint
+        // pointer null for the pullback to write through.
```
vgvassilev: We also do not want lengthy pointless comments. Please keep them tight and only at the places that are really needed.

test/Gradient/PointerAllocation.C
```diff
@@ -0,0 +1,44 @@
...
+
+// Assigning a `new` expression to a pointer must allocate the adjoint buffer
+// alongside it. The adjoint allocation used to be cloned before it was
+// emitted, and StmtClone has no case for a CXXNewExpr, so the whole adjoint
+// assignment was dropped and the adjoint pointer stayed null.
+
+#include "clad/Differentiator/Differentiator.h"
+#include <cstdio>
+
+double allocInBody(double x) {
```
vgvassilev: Is there a need for a new test? Isn’t it better to find an similar test to update?

vgvassilev (commented):
We should simplify the commit log so that it’s shorter and understandable by an engineer. It needs to explain why and not what. The diff makes the “what” obvious.

test/Gradient/PointerAllocation.C
Vedant2005goyal: Yes we do need a seperate test for it, I have tried to incorporate it into other tests but it didnt workout

vgvassilev:
The commit message seems still off -- ask claude to look how the other commit messages are organized

Vedant2005goyal:
> The commit message seems still off -- ask claude to look how the other commit messages are organized

Done

vgvassilev (approved):
LGTM!

PR #2050, opened 2026-09-05
Vary what a pointer writes through, not only the pointer.

### LINKED ISSUE #1873: [crash] -enable-va: null Expr* deref in ReverseModeVisitor::VisitDeclStmt on alias-carried write.

ShresthSamyak:
Fixes: #1873

## The crash

Under `-enable-va`, differentiating a function that writes to a variable through a pointer alias crashes clang with a null `Expr*` dereference in `ReverseModeVisitor::VisitDeclStmt`:

```cpp
double f(double x) {
  double y = 0;
  double* p = &y;
  *p = x * x;
  return y;
}
```

## Root cause

The alias information the analysis needed was already being collected, but never reached the decision that consumes it.

`VariedAnalyzer::TraverseDeclStmt` gives `double* p = &y` a `REF_TYPE` `VarData` whose dependency set is `{y}`, and `AnalysisBase::setIsRequired` walks those edges to set the pointee's varied bit. But whether a declaration is given an adjoint is decided by `DiffRequest::shouldHaveAdjoint`, which reads the request's varied-decl set — and only `addVariedDecl` fills that. `TraverseDeclRefExpr` called it on the syntactic declaration, `p`, and stopped.

So `p` came out varied and `y` did not. `y` was denied an adjoint, differentiating `&y` produced nothing for `_d_p` to be initialised with, and `VisitDeclStmt` dereferenced that null initialiser.

## Why the null guard alone is not the fix

Guarding the dereference stops the crash and produces a **silently wrong gradient**: with `y` still passive, the dependency `return y` carries is gone along with its adjoint. Measured on the reproducer above at `x = 3` (expected `6.00`):

| Build | Result |
|---|---|
| master `e871c9f2` | crash — `Expr::IgnoreCasts()` in `VisitDeclStmt` |
| null guard only | compiles, returns `0.00` |
| this PR | `6.00` |

## The change

- **`ActivityAnalyzer`** — `addVariedDeclWithAliases` walks the same `REF_TYPE` edges into the varied-decl set, transitively and cycle-safe, so everything a varied pointer may alias is given an adjoint.
- **`ReverseModeVisitor`** — guard the adjoint's initialiser before dereferencing it. It holds whatever differentiating the primal's initialiser produced, and nothing promises that is non-null. (`AllocCallInfo::recognize` was already null-safe, so the guard belongs at the caller.)

## Tests

Three cases in `test/Analyses/ActivityReverse.cpp`: a write through a pointer, through an array element reached by two levels of indirection, and through a reference binding. They assert the gradient value rather than only that the compiler survives, which is what distinguishes the crash from the silently-wrong-answer failure.

That file compiles the same source with and without `-enable-va` and exec-checks both, and `TEST1` uses `clad::gradient<clad::opts::enable_va>`, so the new cases exercise the analysis in both RUN lines. Verified load-bearing: with the fix reverted the test fails with three crash signatures.

## Scope

This is deliberately the conservative direction. `getDependencySet` is not a points-to analysis, so handing an adjoint to everything a varied pointer *may* alias gives up some pruning. That seemed the right trade against a wrong derivative, but I am happy to narrow it if you would rather the analysis bail only where it cannot resolve an alias.

The same gap should in principle affect a field reached through a pointer and a pointer passed to a callee; those are not addressed here.

## Testing done

`ninja check-clad` on LLVM/Clang 18.1.3, Ubuntu 24.04: **217 passed, 0 failed**, 30 unsupported (CUDA, Enzyme, OpenMP, 32-bit, clang-repl — toolchains absent locally). `git clang-format-18` reports no changes.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -3981,9 +3981,17 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
           else {
             VarDecl* VDDerived = VDDiff.getDecl_dx();
             declsDiff.push_back(VDDerived);
-            if (Stmt* memsetCall = CheckAndBuildCallToMemset(
-                    BuildDeclRef(VDDerived),
-                    VDDerived->getInit()->IgnoreCasts()))
+            // Without an initializer the adjoint is left for the reverse
+            // sweep to read uninitialized, which is a wrong gradient rather
+            // than a missing one. Refuse instead of answering.
+            if (!VDDerived->getInit())
+              diag(DiagnosticsEngine::Error, VD->getLocation(),
+                   "derivative of the initializer of '%0' is not available; "
+                   "the computed gradient would be incorrect")
+                  << VD->getName();
+            else if (Stmt* memsetCall = CheckAndBuildCallToMemset(
+                         BuildDeclRef(VDDerived),
+                         VDDerived->getInit()->IgnoreCasts()))
               memsetCalls.push_back(memsetCall);
             // Track this pointer's allocation size in bytes so an in-place
             // realloc of it can be undone in the reverse sweep. The shadow is
--- a/lib/Differentiator/ActivityAnalyzer.cpp
+++ b/lib/Differentiator/ActivityAnalyzer.cpp
@@ -10,6 +10,7 @@
 #include "clang/AST/Type.h"
 #include "clang/Basic/LLVM.h"
 
+#include "llvm/ADT/SmallPtrSet.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/Support/Casting.h"
 
@@ -109,6 +110,27 @@ void VariedAnalyzer::AnalyzeCFGBlock(const CFGBlock& block) {
       if (merge(succData.get(), m_BlockData[block.getBlockID()].get()))
         m_CFGQueue.insert(succ->getBlockID());
     }
+  }
+}
+
+void VariedAnalyzer::addVariedDeclWithAliases(const VarDecl* VD) {
+  llvm::SmallVector<const VarDecl*, 4> worklist{VD};
+  // What has been queued already. Two pointers initialised from the same
+  // variable reconverge on it, so a declaration can be reached by more than
+  // one path; admitting each one once is what bounds the walk.
+  llvm::SmallPtrSet<const VarDecl*, 4> seen{VD};
+  while (!worklist.empty()) {
+    const VarDecl* cur = worklist.pop_back_val();
+    m_DiffReq.addVariedDecl(cur);
+    // setIsRequired already walks a REF_TYPE's targets to set their varied
+    // bit; this walks the same edges to reach the request's varied-decl set,
+    // which is what decides whether a declaration is given an adjoint.
+    const VarData* data = getVarDataFromDecl(cur);
+    if (data && data->m_Type == VarData::REF_TYPE)
+      // NOLINTNEXTLINE(cppcoreguidelines-pro-type-union-access)
+      for (const VarDecl* target : *data->m_Val.m_RefData)
+        if (target && seen.insert(target).second)
+          worklist.push_back(target);
   }
 }
 
@@ -427,7 +449,7 @@ bool VariedAnalyzer::TraverseDeclRefExpr(DeclRefExpr* DRE) {
 
   if (m_Varied && m_Marking) {
     setVaried(DRE);
-    m_DiffReq.addVariedDecl(VD);
+    addVariedDeclWithAliases(VD);
     markExpr(DRE);
   } else if (m_Marking)
     setVaried(DRE, false);
--- a/lib/Differentiator/ActivityAnalyzer.h
+++ b/lib/Differentiator/ActivityAnalyzer.h
@@ -37,6 +37,12 @@ class VariedAnalyzer : public clang::RecursiveASTVisitor<VariedAnalyzer>,
   std::set<const clang::Stmt*>& m_ResSet;
   void markExpr(const clang::Stmt* S) { m_ResSet.insert(S); }
   void setVaried(const clang::Expr* E, bool isVaried = true);
+  /// Records \p VD as varied, and with it every declaration \p VD may alias.
+  /// A write reaching a pointer or reference varies what it refers to, not
+  /// only the pointer itself, so the pointee needs an adjoint just as much.
+  /// The alias targets are the ones TraverseDeclStmt recorded in the REF_TYPE
+  /// dependency set, followed transitively for a pointer bound to a pointer.
+  void addVariedDeclWithAliases(const clang::VarDecl* VD);
   void AnalyzeCFGBlock(const clang::CFGBlock& block);
   void TraverseAllStmtInsideBlock(const clang::CFGBlock& block);
 
--- a/test/Analyses/ActivityReverse.cpp
+++ b/test/Analyses/ActivityReverse.cpp
@@ -608,6 +608,35 @@ double f15(double x) { return f15_1(x, pick); }
 // CHECK-NEXT:     }
 // CHECK-NEXT: }
 
+// A write reaching a variable through a pointer varies that variable, not
+// only the pointer holding its address. Leaving the pointee passive used to
+// deny it an adjoint, which both crashed the reverse mode on the adjoint's
+// missing initializer and dropped the dependency `return y` carries.
+double f16(double x) {
+  double y = 0;
+  double* p = &y;
+  *p = x * x;
+  return y;
+}
+
+// The same through an array element and one more level of indirection, so a
+// pointee reached transitively is not left behind either.
+double f17(double x) {
+  double y[2] = {0, 0};
+  double* p = &y[1];
+  double** q = &p;
+  **q = x * x * x;
+  return y[1];
+}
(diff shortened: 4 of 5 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3873,10 +3873,15 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
-            if (Stmt* memsetCall = CheckAndBuildCallToMemset(
-                    BuildDeclRef(VDDerived),
-                    VDDerived->getInit()->IgnoreCasts()))
-              memsetCalls.push_back(memsetCall);
+            // An adjoint need not have an initializer: DifferentiateVarDecl
+            // stores whatever differentiating the primal's initializer
+            // produced, and that is null when the initializer has no adjoint
+            // to speak of. There is no allocation to mirror in that case, so
+            // there is nothing for memset to zero.
+            if (Expr* derivedInit = VDDerived->getInit())
```
vgvassilev: This trades a compiler crash for a silently wrong derivative. With the analysis change none of the three new tests reaches this branch, since each adjoint pointer now has an initializer to mirror, so the branch is untested. When it does fire, the adjoint is emitted as `double *_d_p;` with no initializer and the reverse sweep reads and writes through it. The description's own measurement shows what that looks like: 0.00 instead of 6.00, with nothing reported. Any alias the analysis still misses, or any other initializer whose derivative comes back null, would now take this path instead of crashing. Either drop the guard, so a null adjoint initializer keeps failing loudly until the analysis covers the case, or make the miss visible: an assert naming the contract, or a diagnostic through DiagnosticsEngine at the declaration so the user learns the gradient cannot be trusted. Skipping the memset quietly is the one option that hides it.

vgvassilev (commented):
Thank you for your first contribution!

lib/Differentiator/ReverseModeVisitor.cpp
ShresthSamyak: You are right, and the measurement in my own description was the argument against the guard. Dropped it; the dereference is back as it was, with an assert stating the contract above it:

```cpp
// A declaration given an adjoint must have given it an initializer:
// DifferentiateVarDecl stores whatever differentiating the primal's
// initializer produced, and a null there means the analysis called the
// declaration varied while denying the initializer's operands an adjoint.
// Emitting the adjoint uninitialized and letting the reverse sweep read
// through it would answer with a wrong derivative and no diagnostic, so fail
// here instead, where the contract is stated.
assert(VDDerived->getInit() &&
       "adjoint of a varied declaration has no initializer");
```

So an alias the analysis still misses names the broken invariant in an asserts build, and keeps failing where it already did without them. Nothing is skipped quietly.

Checked both directions rather than assuming, on a build with `NDEBUG` removed:

- with the analysis change reverted, the reproducer stops at
  `Assertion 'VDDerived->getInit() && "adjoint of a varied declaration has no initializer"' failed`
  rather than at the segfault
- with it in place, `check-clad` is 225 passed, 0 failed, and the assert does not fire anywhere in the suite

On the branch being untested: that is right and it stays right, since the assert is a contract rather than a path the tests are meant to reach. If you would rather have a diagnostic through `DiagnosticsEngine` so a release build says something too, I am happy to write that instead — an assert seemed the smaller claim while the analysis is the thing being fixed.

Rebased onto master.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3981,6 +3981,16 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
             declsDiff.push_back(VDDerived);
+            // A declaration given an adjoint must have given it an
+            // initializer: DifferentiateVarDecl stores whatever
+            // differentiating the primal's initializer produced, and a null
+            // there means the analysis called the declaration varied while
+            // denying the initializer's operands an adjoint. Emitting the
+            // adjoint uninitialized and letting the reverse sweep read through
+            // it would answer with a wrong derivative and no diagnostic, so
+            // fail here instead, where the contract is stated.
+            assert(VDDerived->getInit() &&
```
vgvassilev: We should reduce the comment to a more meaningful subset and also consider turning this assert into a diagnostics message (eg. error: or warning:) or a optimization remark hint.

ShresthSamyak: Both done. The assert is now an error through `DiagnosticsEngine`, and the comment is three lines:

```cpp
// Without an initializer the adjoint is left for the reverse sweep to read
// uninitialized, which is a wrong gradient rather than a missing one.
// Refuse instead of answering.
if (!VDDerived->getInit())
  diag(DiagnosticsEngine::Error, VD->getLocation(),
       "derivative of the initializer of '%0' is not available; "
       "the computed gradient would be incorrect")
      << VD->getName();
else if (Stmt* memsetCall = ...)
```

Error rather than warning or a remark, because the state it reports is not a pruning decision the user might want to hear about and carry on past -- the derivative that would come out is wrong. Nothing that worked before is refused by it either: every input that reaches this line crashed until now.

It also turns out to be testable, which the assert was not. A conditional yielding a pointer to a pointer gets no derivative for its initializer, so it lands here:

```cpp
double** r = c ? &p : &q;
```

That happens with the activity analysis off as well, so it is a limitation of its own rather than anything this PR introduces -- on master it is the same crash as #1873, from a different direction. `test/Diagnostics/VariedAdjointInitializer.C` pins the message with `-verify`, so the branch you flagged as unreachable by the other three tests now has one of its own. Happy to drop that test if you would rather not fix the shape of the limitation in place; it does document a real gap.

`check-clad` is 226 passed, 0 failed.

vgvassilev (approved):
LGTM!

PR #2057, opened 2026-09-07
[NFC][UA] Rebase UsefulAnalyzer onto the AnalysisBase lattice

leetcodez:
[NFC][UA] Rebase UsefulAnalyzer onto the AnalysisBase lattice

Replace UsefulAnalyzer's duplicated state storage and traversal machinery with the copy-on-write VarsData lattice from AnalysisBase.

Link blocks through first-discovery m_Prev chains. Direct continuations are reprocessed when their parent is visited; other incoming edges are re-enqueued only when merging introduces a previously unseen variable.

Represent whole-variable usefulness through scalar presence markers, independent of field, pointee, or declaration types. UsefulAnalyzer queries key presence rather than payload values, and its scalar payload bits remain false.

This is intended to be a no-functional-change port. Existing lit coverage in UsefulForward.cpp and UsefulNonTermination.cpp exercises generated-code and termination behavior. The merged #2060 UsefulLattice.cpp test additionally pins the observable default-versus-enable-ua behavior.

The port depends on #2058 reporting insertion of a previously unseen variable as stable merge growth. It does not depend on #2058's array-default behavior or true/false payload-shadowing decisions.

Testing:

• ActivityLoopCarry.cpp: passed (exit code 0)
• UsefulForward.cpp: passed (exit code 0)
• UsefulNonTermination.cpp: passed (exit code 0)
• UsefulLattice.cpp: passed (exit code 0)
• check-clad: 282 passed, 34 unsupported, 0 failures (316 total)

Depends on: #2058

### DIFF
--- a/unittests/Analyses/AnalysisBaseTest.cpp
+++ b/unittests/Analyses/AnalysisBaseTest.cpp
@@ -1,0 +1,195 @@
+#include "AnalysisBase.h"
+
+#include "gtest/gtest.h"
+
+namespace {
+class MergeAccess : public clad::AnalysisBase {
+public:
+  MergeAccess() : AnalysisBase(nullptr) {}
+  using AnalysisBase::merge;
+};
+
+class AnalysisBaseMergeTest : public testing::Test {
+protected:
+  MergeAccess Analysis;
+  // merge treats declaration keys as opaque; nullptr is a supported key.
+  static constexpr const clang::VarDecl* Key = nullptr;
+
+  clad::VarData bit(bool Value) {
+    clad::VarData Data;
+    Data.m_Type = clad::VarData::FUND_TYPE;
+    Data.m_Val.m_FundData = Value;
+    return Data;
+  }
+
+  void put(clad::VarsData& Data, bool Value) { Data[Key] = bit(Value); }
+  bool value(clad::VarsData& Data) { return Data[Key].m_Val.m_FundData; }
+
+  // 0 = absent, 1 = present/false, 2 = present/true.
+  int visible(clad::VarsData* Data) {
+    for (; Data; Data = Data->m_Prev) {
+      auto Found = Data->find(Key);
+      if (Found != Data->end())
+        return Found->second.m_Val.m_FundData ? 2 : 1;
+    }
+    return 0;
+  }
+};
+
+TEST_F(AnalysisBaseMergeTest, ReportsMissingArrayElement) {
+  clad::VarData Target, Source;
+  Target.m_Type = Source.m_Type = clad::VarData::ARR_TYPE;
+  Target.m_Val.m_ArrData = std::make_unique<clad::ArrMap>();
+  Source.m_Val.m_ArrData = std::make_unique<clad::ArrMap>();
+  clad::ProfileID Default, Index;
+  Index.AddInteger(1);
+  (*Target.m_Val.m_ArrData)[Default] = bit(false);
+  (*Source.m_Val.m_ArrData)[Default] = bit(false);
+  (*Source.m_Val.m_ArrData)[Index] = bit(true);
+
+  EXPECT_TRUE(Analysis.merge(Target, Source));
+  EXPECT_TRUE(Target[Index]->m_Val.m_FundData);
+  EXPECT_FALSE(Analysis.merge(Target, Source));
+}
+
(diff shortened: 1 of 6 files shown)


vgvassilev (commented):
The port looks correct. A variable can only be added to the lattice, never removed. The discovery links form a tree. A merge re-queues a block only when it brings a new variable. So the loop ends, and it ends at the same result the old code reached. The tests that would catch a wrong port are the ones with CHECK lines under -enable-ua: UsefulForward.cpp and UsefulNonTermination.cpp. Two remarks below have no line to sit on.

If this change keeps behaviour, put NFC in the title. The description already says the tests do not tell the old analyzer from the new one, which is the same claim. I tried to find a case where the two differ and could not. The old copy-back (mergeData := targetData in mergeVarsData) could give a block the marks of a sibling block it cannot reach. But in the structured CFGs I traced, those marks arrive after the block has already been traversed, so nothing reads them. If you know a case where the results differ, that case needs a test more than the two incomplete-type tests do. Those two pass on master as well, because the old analyzer never looked at a type either.

This port needs only one of the four AnalysisBase hunks it inherits from #2058: the one that reports a newly copied variable as a change (the else branch in the first loop of merge over VarsData). The other three hunks are being discussed there: the array element copy, local versus inherited state, and the ancestor merge report. None of them can affect this analyzer, because every bit in its lattice is false. So #2058 can change in any of those places without breaking this.

I did not build or run anything. CI is green on all 41 configurations for this head.

unittests/Analyses/AnalysisBaseTest.cpp
```diff
@@ -0,0 +1,195 @@
...
+#include "gtest/gtest.h"
+
+namespace {
+class MergeAccess : public clad::AnalysisBase {
+public:
+  MergeAccess() : AnalysisBase(nullptr) {}
+  using AnalysisBase::merge;
+};
+
+class AnalysisBaseMergeTest : public testing::Test {
```
vgvassilev, line 12: Do we need a unittest infrastructure for this analyses? I suspect most of that is reachable via simple lit-level tests, right?

leetcodez: Thanks for the review. I removed the UsefulAnalyzer-specific unit tests from this PR. Both incomplete-type cases also pass with the old analyzer and hence do not distinguish this NFC port.

unittests/Analyses/UsefulAnalyzerTest.cpp has been deleted, and unittests/Analyses/CMakeLists.txt now matches its [#2058](https://github.com/vgvassilev/clad/pull/2058) prerequisite state.

Retained AnalysisBaseTest.cpp in prerequisite [#2058](https://github.com/vgvassilev/clad/pull/2058) because it directly verifies internal merge contract -- including copy-down behavior, shadowing, array defaults and change reporting that are not cleanly isolated through derivative-code lit tests.

UsefulAnalyzer's generated-code behavior remains covered by UsefulForward.cpp and UsefulNonTermination.cpp. Prerequisite [#2060](https://github.com/vgvassilev/clad/pull/2060) additionally checks the default-versus--enable-ua code-generation difference in UsefulLattice.cpp.

leetcodez:
Thanks for the review. I have updated this change to an explicit [NFC] port.

- The two UsefulAnalyzer-specific unit tests were removed because they also pass with the old analyzer and do not distinguish the port.
- The delta against #2058 is now limited to UsefulAnalyzer.cpp and UsefulAnalyzer.h.
- UsefulAnalyzer depends only on #2058 reporting a newly copied variable as a change. Its lattice query key presence and all scalar payload bits remain false, so the other payload-related merge decisions in #2058 do not affect its traversal.
- A 24-case LegacyUsefulAnalyzer-versus-UsefulAnalyzer comparison produced identical usefulness sets. UsefulForward.cpp, UsefulNonTermination.cpp, and the #2060 UsefulLattice.cpp checks all pass.
- The 11 AnalysisBase tests pass, and check-clad passes with 230 tests, 30 unsupported and no failures.

leetcodez:
Update: #2058 has since replaced the AnalysisBase GTest infrastructure with a focused lit regression. This PR is now restacked onto the latest #2058, and its delta is limited to UsefulAnalyzer.cpp and UsefulAnalyzer.h.

PR #2058, opened 2026-09-07
[VA] Preserve accumulated state and report lattice merge changes

leetcodez:
[VA] Preserve accumulated state and report lattice merge changes

Make AnalysisBase merges report newly introduced stable state while preserving locally accumulated values during ancestor reconciliation.

VariedAnalyzer uses the returned change flag to drive re-analysis. TBRAnalyzer ignores that flag but consumes the merged state, so preserving local state is relevant to both consumers.

Lattice and Merge Semantics:

- Missing array elements are materialized from the target default element through VarData::operator[] and recursively merged, preserving the conservative default safety floor.
- Local VarsData entries are preserved while inherited entries are reconciled.
- Stable lattice growth is reported to VariedAnalyzer so affected CFG blocks are re-analysed.
- The final ancestor refresh is deliberately not reported as worklist growth because block traversal may immediately clear that transient state.

Testing and Validation:

- Replaced the implementation-level AnalysisBase GTests with a focused lit regression that observes loop-carried activity through the generated derivative.
- ActivityLoopCarry.cpp produces df/dx = 1 with the patch.

### DIFF
--- a/lib/Differentiator/AnalysisBase.cpp
+++ b/lib/Differentiator/AnalysisBase.cpp
@@ -287,7 +287,7 @@ bool AnalysisBase::merge(VarData& targetData, VarData& mergeData) {
     for (auto& pair : *mergeData.m_Val.m_ArrData) {
       auto it = targetData.m_Val.m_ArrData->find(pair.first);
       if (it == targetData.m_Val.m_ArrData->end())
-        (*targetData.m_Val.m_ArrData)[pair.first] = pair.second.copy();
+        isMod = merge(*targetData[pair.first], pair.second) || isMod;
     }
     return isMod;
   }
@@ -455,8 +455,10 @@ bool AnalysisBase::merge(VarsData* targetData, VarsData* mergeData) {
     if (found) {
       if (merge(*found, *pair.second))
         isModified = true;
-    } else
+    } else {
       (*targetData)[pair.first] = pair.second->copy();
+      isModified = true;
+    }
   }
 
   // For every variable in collected targetData predecessors, search it inside
@@ -474,8 +476,11 @@ bool AnalysisBase::merge(VarsData* targetData, VarsData* mergeData) {
         while (branch) {
           auto it = branch->find(pair.first);
           if (it != branch->end()) {
-            (*targetData)[pair.first] = pair.second->copy();
-            merge((*targetData)[pair.first], it->second);
+            // Preserve local state; copying inherited state is not a change.
+            if (targetData->find(pair.first) == targetData->end())
+              (*targetData)[pair.first] = pair.second->copy();
+            isModified =
+                merge((*targetData)[pair.first], it->second) || isModified;
             break;
           }
           branch = branch->m_Prev;
@@ -494,6 +499,8 @@ bool AnalysisBase::merge(VarsData* targetData, VarsData* mergeData) {
       while (branch) {
         auto it = branch->find(pair.first);
         if (it != branch->end()) {
+          // Refresh the local state, but do not report this as worklist
+          // growth: block traversal may immediately clear it again.
           merge(pair.second, it->second);
           break;
         }
--- a/unittests/Analyses/AnalysisBaseTest.cpp
+++ b/unittests/Analyses/AnalysisBaseTest.cpp
@@ -1,0 +1,195 @@
+#include "AnalysisBase.h"
+
+#include "gtest/gtest.h"
+
+namespace {
+class MergeAccess : public clad::AnalysisBase {
+public:
+  MergeAccess() : AnalysisBase(nullptr) {}
+  using AnalysisBase::merge;
+};
+
+class AnalysisBaseMergeTest : public testing::Test {
+protected:
+  MergeAccess Analysis;
+  // merge treats declaration keys as opaque; nullptr is a supported key.
+  static constexpr const clang::VarDecl* Key = nullptr;
+
+  clad::VarData bit(bool Value) {
+    clad::VarData Data;
+    Data.m_Type = clad::VarData::FUND_TYPE;
+    Data.m_Val.m_FundData = Value;
+    return Data;
+  }
+
+  void put(clad::VarsData& Data, bool Value) { Data[Key] = bit(Value); }
+  bool value(clad::VarsData& Data) { return Data[Key].m_Val.m_FundData; }
+
+  // 0 = absent, 1 = present/false, 2 = present/true.
+  int visible(clad::VarsData* Data) {
+    for (; Data; Data = Data->m_Prev) {
+      auto Found = Data->find(Key);
+      if (Found != Data->end())
+        return Found->second.m_Val.m_FundData ? 2 : 1;
+    }
+    return 0;
+  }
+};
+
+TEST_F(AnalysisBaseMergeTest, ReportsMissingArrayElement) {
+  clad::VarData Target, Source;
+  Target.m_Type = Source.m_Type = clad::VarData::ARR_TYPE;
+  Target.m_Val.m_ArrData = std::make_unique<clad::ArrMap>();
+  Source.m_Val.m_ArrData = std::make_unique<clad::ArrMap>();
+  clad::ProfileID Default, Index;
+  Index.AddInteger(1);
+  (*Target.m_Val.m_ArrData)[Default] = bit(false);
+  (*Source.m_Val.m_ArrData)[Default] = bit(false);
+  (*Source.m_Val.m_ArrData)[Index] = bit(true);
+
+  EXPECT_TRUE(Analysis.merge(Target, Source));
+  EXPECT_TRUE(Target[Index]->m_Val.m_FundData);
+  EXPECT_FALSE(Analysis.merge(Target, Source));
+}
+
+TEST_F(AnalysisBaseMergeTest, PreservesArrayDefaultForMissingElement) {
+  clad::VarData Target, Source;
+  Target.m_Type = Source.m_Type = clad::VarData::ARR_TYPE;
+  Target.m_Val.m_ArrData = std::make_unique<clad::ArrMap>();
+  Source.m_Val.m_ArrData = std::make_unique<clad::ArrMap>();
+  clad::ProfileID Default, Index;
+  Index.AddInteger(1);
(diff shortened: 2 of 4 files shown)


lib/Differentiator/AnalysisBase.cpp
```diff
@@ -474,8 +478,11 @@ bool AnalysisBase::merge(VarsData* targetData, VarsData* mergeData) {
         while (branch) {
           auto it = branch->find(pair.first);
           if (it != branch->end()) {
-            (*targetData)[pair.first] = pair.second->copy();
-            merge((*targetData)[pair.first], it->second);
+            // Preserve local state; copying inherited state is not a change.
+            if (targetData->find(pair.first) == targetData->end())
+              (*targetData)[pair.first] = pair.second->copy();
+            isModified =
+                merge((*targetData)[pair.first], it->second) || isModified;
```
vgvassilev, line 483: This makes the merged result less conservative in one shape, and that deserves to be a decision rather than a side effect.

I modelled merge over all 243 ways five blocks can hold one variable (absent, false, true) and diffed old against new. Exactly two configurations change. Root=false, Left=false, Target=true is the bug you are fixing: the old code overwrote the local true, the new code keeps it. Root=false, Left=true, Target=false goes the other way: the old code copied Left's true down and produced true, the new code keeps the local false and produces false.

In TBR, true means the value has to be stored so the reverse sweep can read it back. So the second shape is a store the analysis used to ask for and now does not. Concretely: a variable is marked required in a block between the join and the common ancestor, the join block cleared it on an earlier pass, and the ancestor has it clear. clad now skips the store, the reverse sweep reads whatever overwrote the value, and the gradient is wrong with nothing crashing.

The new answer may well be the correct one, since a local entry shadows its ancestors everywhere else in this file and getVarDataFromDecl returns the first hit walking up. But the commit message argues only the recovery direction, not this one, and the lit test above is what would settle it.

lib/Differentiator/AnalysisBase.cpp
```diff
@@ -286,8 +286,10 @@ bool AnalysisBase::merge(VarData& targetData, VarData& mergeData) {
     }
     for (auto& pair : *mergeData.m_Val.m_ArrData) {
       auto it = targetData.m_Val.m_ArrData->find(pair.first);
-      if (it == targetData.m_Val.m_ArrData->end())
+      if (it == targetData.m_Val.m_ArrData->end()) {
         (*targetData.m_Val.m_ArrData)[pair.first] = pair.second.copy();
+        isMod = true;
```
vgvassilev: Copying the element in verbatim can drop the safety floor arrays carry.

An array VarData keeps one element under the default ProfileID standing for indices not known at compile time, and VarData::operator[] materialises a missing index as a copy of that element so an unknown index is never read back as not-required. This loop bypasses that. Merge a Target holding Default true into a Source holding Default false and Index false, and Target ends up with Index false written down, where reading Index before the merge would have given true. The comment above setIsRequired states the rule this breaks: if we are unsure whether an index is used, we assume it is.

Inserting through VarData::operator[] and then merging fixes it, and gives an accurate flag at the same time: isMod = true as written fires even when the copied element leaves every observable value alone.

lib/Differentiator/AnalysisBase.cpp
```diff
@@ -494,7 +501,7 @@ bool AnalysisBase::merge(VarsData* targetData, VarsData* mergeData) {
       while (branch) {
         auto it = branch->find(pair.first);
         if (it != branch->end()) {
-          merge(pair.second, it->second);
+          isModified = merge(pair.second, it->second) || isModified;
```
vgvassilev: Feeding this result into isModified removes the property the worklist relies on to terminate.

VariedAnalyzer has no iteration cap. ActivityAnalyzer.cpp:109 re-queues a block whenever merge returns true, and that flag is the only thing that ends the loop. TBRAnalyzer has m_BlockPassCounter holding it to three passes; VariedAnalyzer has no equivalent.

What this line now reports is the ancestor's bit being merged into an entry that the block's own traversal will clear again, because VariedAnalyzer::TraverseDeclRefExpr calls setVaried(DRE, false) when a variable is assigned something non-varied. A cycle can then settle into merge sets the bit and says changed, the block is re-analysed and clears it, merge sets it again. Before this patch the phase-3 result was discarded, so the flag only fired for changes a re-traversal could not undo.

Worth either convincing yourself a bound exists and writing it down, or adding one.

vgvassilev (commented):
This is a ai-based review I mostly agree with.

unittests/Analyses/AnalysisBaseTest.cpp
```diff
@@ -0,0 +1,154 @@
...
+    for (auto* Data : {&Root, &Left, &Right, &Target, &Source}) {
+      int State = States % 3;
+      States /= 3;
+      if (State)
+        put(*Data, State == 2);
+    }
+    int Before = visible(&Target);
+    int Incoming = visible(&Source);
+    int Expected =
+        Before == 2 || Incoming == 2 ? 2 : (Before || Incoming ? 1 : 0);
```
vgvassilev, line 179: This is where the decision asked for in the thread on AnalysisBase.cpp line 485 is already taken. visible() returns the first hit walking up, so Expected says a local false beats an ancestor's true, and the sweep then asserts merge agrees in all 243 configurations. If the answer goes the other way, this is the line that changes. Either way, the two shapes named in that thread (Root=false, Left=true, Target=false and its mirror) are worth pulling out of the loop into named tests, so the choice is visible in a test name rather than inside a formula.

lib/Differentiator/AnalysisBase.cpp
leetcodez: I have validated the local-shadow contract: local entries in `VarsData` shadow ancestor entries along the `m_Prev` chain, directly mirroring the lookup semantics of `getVarDataFromDecl()`.

This behavior is tested by two dedicated named unit tests in `unittests/Analyses/AnalysisBaseTest.cpp`:
- `LocalTrueShadowsInheritedFalse`: confirms that a local `true` is preserved when merging across ancestor branches that inherit `false`.
- `LocalFalseShadowsInheritedTrue`: confirms that a local `false` correctly shadows an inherited ancestor `true` without being overwritten.

To evaluate the impact on TBR store/restore decisions, I constructed an end-to-end differential corpus covering branch kills, nonlinear joins, loop-carried overwrites, and nested conditional kills across positive, negative, and zero floating-point inputs. The tested parent and candidate generated byte-identical TBR code, so this corpus is preservation evidence rather than a distinguishing witness. No required tape stores were omitted in these tested cases, and all tested analytical gradients remained correct.

lib/Differentiator/AnalysisBase.cpp
leetcodez: On the array default safety floor. I replaced the direct copy of missing elements with:
```cpp
if (it == targetData.m_Val.m_ArrData->end())
  isMod = merge(*targetData[pair.first], pair.second) || isMod;
```
By accessing `targetData[pair.first]` via `VarData::operator[]`, the missing element is materialized from the target's default `ProfileID` element before merging. The recursive merge then ensures that a conservative `true` default in the target cannot be overwritten by an explicit `false` from the source branch, preserving the safety floor. Furthermore, `isMod` reflects true lattice growth rather than mere map key insertion.

I added a dedicated unit test `PreservesArrayDefaultForMissingElement` in `unittests/Analyses/AnalysisBaseTest.cpp`. In a negative control, the direct-copy implementation failed:
```
Value of: Analysis.merge(Target, Source) -> Actual: true, Expected: false
Value of: Target[Index]->m_Val.m_FundData -> Actual: false, Expected: true
```
With the candidate's `operator[]` + recursive merge, `PreservesArrayDefaultForMissingElement` passes cleanly, and existing test `ReportsMissingArrayElement` continues to pass.

lib/Differentiator/AnalysisBase.cpp
leetcodez: I agree that reporting Phase 3 refreshes as worklist growth can violate the termination property. Phase 3 still merges the ancestor state into the local entry, but its result is no longer added to isModified. AncestorRefreshDoesNotSignalWorklistGrowth verifies that the state changes while merge() returns false.

A bounded exploratory CFG corpus terminated without oscillation. The goto/label cases emitted existing unsupported-statement warnings, so I do not claim a universal CFG proof or rely on those cases as semantic coverage.

unittests/Analyses/AnalysisBaseTest.cpp
leetcodez: Done. I have made the semantic policy explicit:
1. Renamed `PreservesLocalBitAcrossAncestorMerge` to `LocalTrueShadowsInheritedFalse`.
2. Added the explicit mirror test `LocalFalseShadowsInheritedTrue`:
```cpp
TEST_F(AnalysisBaseMergeTest, LocalFalseShadowsInheritedTrue) {
  clad::VarsData Root, Parent, Target, Source;
  Parent.m_Prev = Source.m_Prev = &Root;
  Target.m_Prev = &Parent;
  put(Root, false);
  put(Parent, true);
  put(Target, false);

  EXPECT_FALSE(Analysis.merge(&Target, &Source));
  EXPECT_FALSE(value(Target));
  EXPECT_FALSE(Analysis.merge(&Target, &Source));
  EXPECT_FALSE(value(Target));
}
```
This explicitly documents the contract that local entries in `VarsData` shadow ancestors in `m_Prev` (aligning with `getVarDataFromDecl()`).

Both named tests pass on the candidate diff and fail against the legacy overwrite behavior. The exhaustive 243-configuration sweep in `ScalarForksJoinAndThenStabilize` is retained alongside them.

unittests/Analyses/AnalysisBaseTest.cpp
```diff
@@ -0,0 +1,195 @@
+#include "AnalysisBase.h"
```
vgvassilev: We do not need tests for the particular implementation details. Unless I am missing something we should be good enough with a tight set of lit-based tests.

leetcodez: Done. I removed the AnalysisBase GTests and replaced them with a focused lit regression under test/Analyses/.

vgvassilev (approved):
LGTM!

PR #2059, opened 2026-09-08
Fix null body in generated IfStmt

Vedant2005goyal:
A braceless branch whose body carries no derivative can result in a null branch and crash clang:

```cpp
if (x > 0)
  delete[] d;
```
VisitCXXDeleteExpr saves the delete for the end of the derivative and returns no statement, so the branch becomes an empty block. unwrapIfSingleStmt then turns the empty block into null, which becomes the then-branch of the IfStmt clad builds, ultimately leading to a crash.


vgvassilev (approved):
LGTM!

PR #2060, opened 2026-09-08
[Tests] Make UsefulAnalyzer effect observable in UsefulLattice

leetcodez:
[Tests] Make UsefulAnalyzer effect observable in UsefulLattice

Pipe captured codegen dumps in UsefulLattice.cpp through FileCheck to
verify carry_darg0 retains the unused loop counter adjoint (_d_i) by default
and drops it under -enable-ua for both -O0 and -O3.

### DIFF
--- a/test/Analyses/UsefulLattice.cpp
+++ b/test/Analyses/UsefulLattice.cpp
@@ -1,0 +1,139 @@
+// RUN: %cladclang -O0 %s -I%S/../../include -o %t.base > %t.base.codegen
+// RUN: %filecheck --check-prefix=CHECK-BASE %s < %t.base.codegen
+// RUN: %t.base | %filecheck_exec %s
+// RUN: %cladclang -O0 -Xclang -plugin-arg-clad -Xclang -enable-ua %s -I%S/../../include -o %t.ua > %t.ua.codegen
+// RUN: %filecheck --check-prefix=CHECK-UA %s < %t.ua.codegen
+// RUN: %t.ua | %filecheck_exec %s
+// RUN: %cladclang -O3 -DNDEBUG %s -I%S/../../include -o %t.opt > %t.opt.codegen
+// RUN: %filecheck --check-prefix=CHECK-BASE %s < %t.opt.codegen
+// RUN: %t.opt | %filecheck_exec %s
+// RUN: %cladclang -O3 -DNDEBUG -Xclang -plugin-arg-clad -Xclang -enable-ua %s -I%S/../../include -o %t.ua-opt > %t.ua-opt.codegen
+// RUN: %filecheck --check-prefix=CHECK-UA %s < %t.ua-opt.codegen
+// RUN: %t.ua-opt | %filecheck_exec %s
+
+#include "clad/Differentiator/Differentiator.h"
+
+#include <cmath>
+#include <cstdio>
+#include <cstdlib>
+#include <initializer_list>
+
+double carry(double x, int n) {
+  double a = x, b = 0, c = 0;
+  for (int i = 0; i < n; ++i) {
+    c = b;
+    b = a;
+    a = x;
+  }
+  return c;
+}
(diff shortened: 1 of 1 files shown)


test/Analyses/UsefulLattice.cpp
```diff
@@ -0,0 +1,128 @@
+// RUN: %cladclang -O0 %s -I%S/../../include -o %t.base > %t.base.codegen
+// RUN: %t.base | %filecheck_exec %s
+// RUN: %cladclang -O0 -Xclang -plugin-arg-clad -Xclang -enable-ua %s -I%S/../../include -o %t.ua > %t.ua.codegen
```
vgvassilev, line 4: The two -enable-ua lines can catch an adjoint dropped wrongly but not one kept wrongly. The only visible difference the analysis makes in this file is in the dumped derivative code, and every RUN line redirects that to a %t.*.codegen file nothing reads. The loop counters are the concrete case: by default forward mode emits `int _d_i = 0;` (test/FirstDerivative/Loops.C:24) and under -enable-ua it does not (UsefulForward.cpp:116), so carry, nested, exits and do_loop all come out different in the two modes, and this test passes either way. Since the dump is already captured, running it through FileCheck with a prefix per mode for one of those functions, an absent `_d_i` for carry under -enable-ua and a present one by default, would make the analysis observable here rather than only its worst failure.

leetcodez: Updated `test/Analyses/UsefulLattice.cpp` to pipe the captured `%t.*.codegen` files through `FileCheck`:

```cpp
// RUN: %cladclang -O0 %s -I%S/../../include -o %t.base > %t.base.codegen
// RUN: %filecheck --check-prefix=CHECK-BASE %s < %t.base.codegen
// RUN: %t.base | %filecheck_exec %s
// RUN: %cladclang -O0 -Xclang -plugin-arg-clad -Xclang -enable-ua %s -I%S/../../include -o %t.ua > %t.ua.codegen
// RUN: %filecheck --check-prefix=CHECK-UA %s < %t.ua.codegen
// RUN: %t.ua | %filecheck_exec %s
// RUN: %cladclang -O3 -DNDEBUG %s -I%S/../../include -o %t.opt > %t.opt.codegen
// RUN: %filecheck --check-prefix=CHECK-BASE %s < %t.opt.codegen
// RUN: %t.opt | %filecheck_exec %s
// RUN: %cladclang -O3 -DNDEBUG -Xclang -plugin-arg-clad -Xclang -enable-ua %s -I%S/../../include -o %t.ua-opt > %t.ua-opt.codegen
// RUN: %filecheck --check-prefix=CHECK-UA %s < %t.ua-opt.codegen
// RUN: %t.ua-opt | %filecheck_exec %s
```

Directly following `carry()`, I added:
```cpp
// CHECK-BASE-LABEL: double carry_darg0(double x, int n) {
// CHECK-BASE: int _d_i = 0;
// CHECK-BASE-LABEL: double arms_darg0

// CHECK-UA-LABEL: double carry_darg0(double x, int n) {
// CHECK-UA-NOT: _d_i
// CHECK-UA-LABEL: double arms_darg0
```

I validated all four compilation and execution modes (`base_O0`, `ua_O0`, `base_O3`, `ua_O3`), and all FileCheck and numerical assertions pass. I also verified the negative control: running `CHECK-UA` against `%t.base.codegen` fails as expected with `CHECK-UA-NOT: excluded string found in input: _d_i`.

vgvassilev:
In addition, please make sure that the comments are tight and only at the non-obvious places. We should also apply the same to the commit message.

leetcodez:
Done. I removed the redundant comment in `UsefulLattice.cpp` and tightened the commit message and PR description to focus strictly on the FileCheck behavior.

vgvassilev (approved):
LGTM!

PR #2064, opened 2026-09-11
fix lambda captures

fogsong233:


lib/Differentiator/ReverseModeForwPassVisitor.cpp
```diff
@@ -26,6 +26,32 @@ ReverseModeForwPassVisitor::ReverseModeForwPassVisitor(
     DerivativeBuilder& builder, const DiffRequest& request)
     : ReverseModeVisitor(builder, request) {}
 
+Expr* ReverseModeForwPassVisitor::buildDerivedLambda() {
```
vgvassilev, line 33: Problem: this is a second copy of ReverseModeVisitor::buildDerivedLambda. Both build a LambdaIntroducer, call LambdaBuilder::start with GetDerivativeType and BuildParams, open a block, differentiate the body and finish. They differ in three details: this one sets no Intro.Range, passes ExplicitResult=true, and forces LCD_ByRef. The CI crash on LLVM 21 and older sits in LambdaBuilder::start called from here.

How to address it: one builder, in VisitorBase or ReverseModeVisitor, taking the introducer and a body callback; the two callers pass only what differs (the capture default and the body). Then the range is set in one place and the crash has one place to be fixed.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1951,54 +1957,113 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+    LambdaReq.CallContext = m_DiffReq.CallContext;
+    LambdaReq.VerboseDiags = m_DiffReq.VerboseDiags;
+    for (const auto* Param : LambdaReq.Function->parameters())
+      LambdaReq.DVI.push_back(Param);
 
     ReverseModeVisitor NestedVisitor(m_Builder, LambdaReq);
-    Expr* lambdaE = NestedVisitor.buildDerivedLambda(LE);
+    LambdaCaptures Captures(*this);
+    Captures.collect({LE->getBody()});
+    ExprMutationAnalyzer Mutations(*LE->getBody(), m_Context);
```
vgvassilev, line 1990: Problem: ExprMutationAnalyzer is used to compute one boolean, MutatesCaptures. It costs an include from clang/Analysis and, in tools/CMakeLists.txt, linking the clangAnalysis and clangASTMatchers archives into every static host. clad already tracks which parameters a function modifies for TBR (getModifiedParams), and the forward-pass lambda is emitted anyway whenever the call operator returns a memory type.

How to address it: answer the question with what is already here, a small visitor over assignments and non-const calls on the captured variables, or the TBR modified set, and drop the two archives and the include.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1951,54 +1957,113 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+      m_DiffReq.getVariedDecls().insert(Snapshot);
+      m_DiffReq.getToBeRecorded().insert(SnapshotDecl);
+      StmtDiff SnapshotDiff = VisitDeclStmt(SnapshotDecl);
+      addToCurrentBlock(SnapshotDiff.getStmt());
+      auto* SnapshotClone = m_DeclReplacements[Snapshot];
+      Snapshots.emplace_back(Variable, SnapshotClone);
+      if (Variable->isInitCapture())
+        m_DeclReplacements[Variable] = SnapshotClone;
+    }
+    NestedVisitor.m_DeclReplacements = m_DeclReplacements;
```
vgvassilev, line 2037: Problem: the parent's state is copied into NestedVisitor here and again into ForwVisitor ten lines down, four maps each time, and m_Variables is copied wholesale, so every adjoint of the enclosing function is duplicated into each nested visitor.

How to address it: one helper that seeds a nested visitor from its parent (declaration map, lambda forward passes, tracker owner, and a view of the adjoints rather than a copy), used for both.

test/Gradient/LambdaCaptures.C
```diff
@@ -0,0 +1,344 @@
+// RUN: %cladclang %s -I%S/../../include -o %t
+// RUN: %t | %filecheck_exec %s
+// RUN: %cladclang %s -I%S/../../include -Xclang -plugin-arg-clad -Xclang -disable-tbr -Xclang -plugin-arg-clad -Xclang -disable-va -o %t.no-analysis
+// RUN: %t.no-analysis | %filecheck_exec %s
+// RUN: %cladclang %s -I%S/../../include -Xclang -plugin-arg-clad -Xclang -enable-va -o %t.va
```
vgvassilev, line 5: Problem: the third configuration runs every file a third time with varied analysis on, and nothing in this change is specific to that analysis; the one line touching it inserts a snapshot into the varied set, which the default run exercises too. Across five files that is fifteen compile-and-run passes.

How to address it: keep the two configurations Lambdas.C uses, default and no-analysis.

test/Gradient/LambdaInitialization.C
```diff
@@ -0,0 +1,80 @@
...
+}
+
+double braced_reference(double x) {
+  auto inner{([&value = x] { return value * value; })};
+  double first = inner();
+  x *= 3;
+  return first + inner();
+}
+
+double nested_mutating_reference(double x) {
```
vgvassilev, line 39: Problem: this file re-runs cases from LambdaCaptureState.C with parentheses or braces around the lambda. This function is nested_mutation, nested_mutable_capture is nested_snapshot, parenthesized_reference and braced_reference are read_before_mutation and read_reference_initializer, and the numbers match one for one.

How to address it: the only new path here is the transparent brace and the parenthesized initializer, and parenthesized_capture and braced_capture cover both. Keep those two in Lambdas.C and drop the other four.

vgvassilev (commented):
Problem: this is at least six changes in one PR, with no description, no issue number, and three commits whose titles do not say what they do. (1) StmtClone fixes: case and default labels registered with the cloned switch, the if-init and the switch condition variable cloned. (2) A second cloning mechanism: rebuild hooks in StmtClone and a rewritten buildClonedLambda. (3) Three new diagnostics for closures clad cannot handle. (4) TBR: closure reference fields recorded as dependencies, lambda call operators treated as instance calls. (5) Restore-tracker ownership across frames (the Owner argument and m_RestoreTrackerOwner). (6) The fix itself: capture snapshots and a forward-pass lambda, plus a new dependency on ExprMutationAnalyzer. And a planner change that drops a request after a diagnostic. The same code also duplicates itself: two copies of buildDerivedLambda, the same visitor state copied twice, and the "braces around a lambda are transparent" peel written once in VisitInitListExpr and again in TBRAnalyzer. https://github.com/vgvassilev/clad/pull/1859 has targeted the same issue (https://github.com/vgvassilev/clad/issues/1769) since June and nothing here says how the two relate.

How to address it: split into a series and land the independent parts first, each with its own small test: (1), (4), (5) and the planner guard are self-contained fixes; (3) is a diagnostics commit with LambdaCopies.C; (2) and (6) are the fix. Say in the description which issue this fixes and whether it supersedes https://github.com/vgvassilev/clad/pull/1859. The inline notes point at the duplicated code; each one shrinks the fix.

Problem: CI is red on every finished row on LLVM 21 and older. LambdaCaptureState.C crashes clang on its first case, read_before_mutation; the trace ends in LambdaBuilder::start called from ReverseModeForwPassVisitor::buildDerivedLambda. The baseline check itself passes, so the tests do observe the change.

How to address it: see the inline note on the second buildDerivedLambda. It is the one copy that builds the introducer without a range and with an explicit result type, so whichever of those the older Sema rejects, one shared builder removes the difference.

Problem: the tests are 53 functions in 5 new files, each compiled and run three times, and only numbers are checked. master's test/Gradient/Lambdas.C already has 7 cases with 124 CHECK-NEXT lines and two configurations. Inside the new files the same scenario recurs: nested_mutating_reference is nested_mutation with parentheses (both 1360), nested_mutable_capture is nested_snapshot (400), parenthesized_reference and braced_reference are read_before_mutation and read_reference_initializer (40), loop_reference_initializer, repeated_loop_calls and reference_return_loop all compute 336, and mutable_capture, mutable_alias, mutable_initialized_capture and default_and_initialized_capture share one body (80). CopyValue is defined twice. About a dozen cases are run a second time with only the sign flipped where no branch depends on the sign. Nothing checks the generated code, which is what this change is about.

How to address it: fold the cases into test/Gradient/Lambdas.C. Keep one function per mechanism: by-copy snapshot, by-reference capture, init-capture, mutable, nested closure, closure in a loop, switch in a closure, reference return, closure alias. Give the snapshot and forward-pass cases CHECK-NEXT on the generated body, check numbers for the rest, and keep a sign flip only where a branch depends on it. Keep LambdaCopies.C as the diagnostics test. Two configurations, as Lambdas.C has, are enough.

I did not build or run anything.

fogsong233:
I ran into incorrect gradients for lambdas with captures while working on my Torch integration task, so I opened this PR. I missed the existing work in #1859—sorry for the overlap.
Since this is outside the main scope of my task, I’ll close this PR for now and focus on the Torch integration.

PR #2065, opened 2026-09-11
Carry a type alias into the derivative instead of refusing it.

vgvassilev:
A `using` or `typedef` in the body of a differentiated function was reported as an unsupported declaration kind and dropped, while the statements around it went on naming the type by it. That rules out the way performance-portable
code is written: the alias is where the precision, the width or the index type is chosen once, and every declaration below it is spelled in those terms.

Both modes now re-declare the alias in the derivative with the type it was written with, rather than the type it resolves to, so a derivative of a template still names the type that template chose. Reverse mode puts it at the top of the derivative, because the declarations that use it are promoted to function scope too and an alias reads no variable that would keep it from moving there.

A local class stays unsupported: it has members to derive, which an alias does not.


vgvassilev:
The failures seem unrelated.

PR #2066, opened 2026-09-11
Preserve reverse-forward results used by return statements

fogsong233:
`hasUnusedReturnValue` treats a non-expression parent as evidence that a call’s result is unused. However, `ReturnStmt` consumes that result. This misclassification can cause reverse-forward generation to omit the value/adjoint pair needed to build the enclosing function’s return statement.
Treat `ReturnStmt` as a use of the call result, allowing the existing reverse-forward handling to preserve both components.
Add a standalone regression test that checks the generated return statement and verifies the gradient through a record-returning helper.

### DIFF
--- a/lib/Differentiator/CladUtils.cpp
+++ b/lib/Differentiator/CladUtils.cpp
@@ -1925,6 +1925,10 @@ namespace clad {
         // If the parent is a Stmt but not an Expr, then the result is not used.
         if (!S)
           return false;
+        // A call returned directly is still consumed by the enclosing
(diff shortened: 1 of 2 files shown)


vgvassilev (approved):
Lgtm!

fogsong233:
@vgvassilev It seems that the docs/readthdocs.org:clad has some problem? I seem the fail is due to network.

vgvassilev:
> @vgvassilev It seems that the docs/readthdocs.org:clad has some problem? I seem the fail is due to network.

Yes. We can ignore it.

PR #2067, opened 2026-09-11
Preserve differentiation through default arguments and member initializers

fogsong233:
Default arguments and default member initializers can hide global dependencies from activity analysis, causing reverse mode to miss gradient contributions. Traverse their underlying expressions, including semantic initializer lists, so these dependencies are retained.
Also reset constructor tracking while differentiating call arguments, preventing argument constructors from inheriting the enclosing variable’s initialization context. This applies consistently to omitted and explicit arguments.

### DIFF
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -1898,6 +1898,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       bool isCUDAKernel) {
     StmtDiff result;
     StmtDiff argDiff{};
+    const auto* defaultArg = dyn_cast<CXXDefaultArgExpr>(arg);
     // FIXME: We handle parameters with default values by setting them
     // explicitly. However, some of them have private types and cannot be set.
     // For this reason, we ignore std::__nat. We need to come up with a
@@ -1920,13 +1921,14 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       // is done to reduce cloning complexity and only clone once. The type is
       // same as the call expression as it is the type used to declare the
       // _gradX array
-      QualType dArgTy =
-          utils::getNonConstType(CloneType(arg->getType()), m_Sema);
-      Expr* init = getStdInitListSizeExpr(arg);
+      QualType argType =
+          defaultArg ? utils::GetValueType(param->getType()) : arg->getType();
+      QualType dArgTy = utils::getNonConstType(CloneType(argType), m_Sema);
+      Expr* init = defaultArg ? nullptr : getStdInitListSizeExpr(arg);
       bool shouldCopyInitialize = false;
       if (!init) {
         if (const CXXRecordDecl* CRD = dArgTy->getAsCXXRecordDecl())
-          shouldCopyInitialize = utils::isCopyable(CRD);
+          shouldCopyInitialize = !defaultArg && utils::isCopyable(CRD);
         // Temporarily initialize the object with `*nullptr` to avoid
         // a potential error because of non-existing default constructor.
         if (shouldCopyInitialize) {
@@ -2012,8 +2014,11 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
         dArgRef = BuildDeclRef(dArgDeclCUDA);
       }
       result.updateStmtDx(dArgRef);
-      // Visit using uninitialized reference.
-      argDiff = Visit(arg, BuildDeclRef(dArgDecl));
+      if (defaultArg)
+        argDiff = {Clone(defaultArg->getExpr()), getZeroInit(dArgTy)};
+      else
+        // Visit using uninitialized reference.
+        argDiff = Visit(arg, BuildDeclRef(dArgDecl));
       if (shouldCopyInitialize) {
         if (Expr* dInit = argDiff.getExpr_dx())
           SetDeclInit(dArgDecl, dInit);
--- a/test/Gradient/DefaultArgumentAdjoint.C
+++ b/test/Gradient/DefaultArgumentAdjoint.C
@@ -1,0 +1,42 @@
+// RUN: %cladclang %s -I%S/../../include -o %t 2>&1 | %filecheck %s
+// RUN: %t | %filecheck_exec %s
+
+#include "clad/Differentiator/Differentiator.h"
+
+#include <cstdio>
+#include <optional>
+
+double configured_scale(double value,
+                        ::std::optional<int> factor = ::std::nullopt) {
+  return value * factor.value_or(2);
+}
+
+namespace clad::custom_derivatives {
+
+void configured_scale_pullback(
+    double /*value*/, ::std::optional<int> factor, double d_output,
+    double* d_value, ::std::optional<int>* /*d_factor*/) {
+  *d_value += factor.value_or(2) * d_output;
+}
+
+} // namespace clad::custom_derivatives
+
+double use_default_factor(double value) { return configured_scale(value); }
+
+// CHECK: void use_default_factor_grad(double value, double *_d_value) {
+// CHECK-NEXT:     {
+// CHECK-NEXT:         double _r0 = 0.;
+// CHECK-NEXT:         std::optional<int> _r1 = {};
+// CHECK-NEXT:         clad::custom_derivatives::configured_scale_pullback(value, ::std::nullopt, 1, &_r0, &_r1);
+// CHECK-NEXT:         *_d_value += _r0;
+// CHECK-NEXT:     }
+// CHECK-NEXT: }
+
+int main() {
+  auto gradient = clad::gradient(use_default_factor);
+  double derivative = 0.0;
+  gradient.execute(3.0, &derivative);
+  std::printf("%.1f\n", derivative);
+}
+
+// CHECK-EXEC: 2.0
--- a/test/Gradient/DefaultArgumentAdjoint.C
+++ b/test/Gradient/DefaultArgumentAdjoint.C
@@ -1,21 +1,32 @@
 // RUN: %cladclang %s -I%S/../../include -o %t 2>&1 | %filecheck %s
-// RUN: %t | %filecheck_exec %s
+// RUN: %t
 
 #include "clad/Differentiator/Differentiator.h"
 
-#include <cstdio>
-#include <optional>
+struct DefaultFactor {
+  struct Construct {};
 
-double configured_scale(double value,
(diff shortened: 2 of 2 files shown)


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1920,13 +1921,14 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
-      Expr* init = getStdInitListSizeExpr(arg);
+      QualType argType =
+          defaultArg ? utils::GetValueType(param->getType()) : arg->getType();
+      QualType dArgTy = utils::getNonConstType(CloneType(argType), m_Sema);
+      Expr* init = defaultArg ? nullptr : getStdInitListSizeExpr(arg);
       bool shouldCopyInitialize = false;
       if (!init) {
         if (const CXXRecordDecl* CRD = dArgTy->getAsCXXRecordDecl())
-          shouldCopyInitialize = utils::isCopyable(CRD);
+          shouldCopyInitialize = !defaultArg && utils::isCopyable(CRD);
```
vgvassilev, line 1931: Shouldn’t that sink in isCopyable?

fogsong233: Do you mean moving the !defaultArg check into the surrounding control flow, so that isCopyable is only called for explicitly provided arguments? 
Since whether an argument is a default argument is call-site information, while isCopyable only receives a CXXRecordDecl, I think utils::isCopyable should remain a general type-level utility.

test/Gradient/DefaultArgumentAdjoint.C
```diff
@@ -0,0 +1,42 @@
+// RUN: %cladclang %s -I%S/../../include -o %t 2>&1 | %filecheck %s
+// RUN: %t | %filecheck_exec %s
+
+#include "clad/Differentiator/Differentiator.h"
+
+#include <cstdio>
```
vgvassilev: Can we make this test standalone without requiring header inclusion?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1920,13 +1921,14 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
       // is done to reduce cloning complexity and only clone once. The type is
       // same as the call expression as it is the type used to declare the
       // _gradX array
-      QualType dArgTy =
-          utils::getNonConstType(CloneType(arg->getType()), m_Sema);
-      Expr* init = getStdInitListSizeExpr(arg);
+      QualType argType =
+          defaultArg ? utils::GetValueType(param->getType()) : arg->getType();
+      QualType dArgTy = utils::getNonConstType(CloneType(argType), m_Sema);
```
vgvassilev: Problem: this is the identity. A CXXDefaultArgExpr already carries the parameter's type, so `arg->getType()` and `param->getType()` are the same here, and `GetValueType` only strips pointers, references and arrays, none of which reach this branch. It is also the one place the change could go wrong: for a pointer parameter `GetValueType` would give `T` where the old code gave `T*`, so if that path ever became reachable the adjoint temporary would have the wrong type.

How to address it: restore the original two lines.
```suggestion
      QualType dArgTy =
          utils::getNonConstType(CloneType(arg->getType()), m_Sema);
```

lib/Differentiator/ReverseModeVisitor.cpp
vgvassilev: Problem: this guard has no effect, which answers the question on this line. With it removed and a copyable record type, `shouldCopyInitialize` is true, the declaration starts as `T _r1 = *(T*)nullptr;`, and then the block below calls `SetDeclInit` with `argDiff.getExpr_dx()`, which for a default argument is `getZeroInit(dArgTy)`. The declaration ends up as `T _r1 = {};`, exactly what the guarded path produces, and exactly what this PR's own CHECK line expects.

How to address it: delete it. Then the question of whether it belongs in `isCopyable` does not arise.
```suggestion
          shouldCopyInitialize = utils::isCopyable(CRD);
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -2012,8 +2014,11 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
         dArgRef = BuildDeclRef(dArgDeclCUDA);
       }
       result.updateStmtDx(dArgRef);
-      // Visit using uninitialized reference.
-      argDiff = Visit(arg, BuildDeclRef(dArgDecl));
+      if (defaultArg)
+        argDiff = {Clone(defaultArg->getExpr()), getZeroInit(dArgTy)};
```
vgvassilev: Problem: this is the line that fixes the bug, and it also throws away any derivative the default expression carries. A default argument cannot name the caller's locals, but it can name a global, and clad differentiates globals. For `double g = 2; double f(double x, double c = g) { return x * c; }`, master seeds the visit with the adjoint temporary, so `_d_g` accumulates; with this change the adjoint is a fresh zero and `_d_g` stays at zero, with nothing reported. It also means a CXXDefaultArgExpr is now handled two ways: `VisitCXXDefaultArgExpr` differentiates through it, this branch does not.

How to address it: if treating a default argument as a constant is the intent, say so in a comment here and add the global case as a test with the zero it now produces, so the choice is recorded rather than inferred. If it is not the intent, the narrower fix is to keep visiting and make `VisitCXXDefaultArgExpr` handle whatever the conversion produces, which also keeps the two paths in agreement.

vgvassilev (commented):
Problem: the type change this PR is named after does not do anything. Sema has already converted a default argument to the parameter type by the time clad sees it, so arg->getType() on a CXXDefaultArgExpr is the parameter type. I dumped the AST of this PR's own example and of the case in the description: the node's type is OptionalFactor, not DefaultFactor, and for std::optional<int> o = std::nullopt it is std::optional<int>, not std::nullopt_t. The branch being edited is also reached only for by-value parameters, because utils::IsReferenceOrPointerArg sends pointers, arrays and references down the other path, so GetValueType is the identity on everything that gets here. That leaves one line that changes behaviour: the argDiff assignment, which stops differentiating the default expression and gives it a zero adjoint.

How to address it: drop the type swap and the init and isCopyable edits, keep the argDiff branch, and retitle. Then say in the description what actually goes wrong, which is what visiting the default expression produces rather than what type it has. Two of the three inline notes are that reduction; the third is about the line that does the work.

I did not build or run anything, so I could not reproduce the failure that prompted this. CI is green on all 44 checks, and the "Tests observe the change" check passes, so the new test does fail on master. What it observes is the argDiff line, not the type.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -1516,6 +1516,9 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
 
   StmtDiff
   ReverseModeVisitor::VisitCXXDefaultArgExpr(const CXXDefaultArgExpr* DE) {
+    // This initializes a call argument, even when the call initializes a local.
```
vgvassilev: A default argument is one of several ways to reach a constructor that is not the VarDecl's initialiser, and this covers only that one. Nothing resets the flag for an explicitly written argument: in `double y = f(A(p, q));` it is still set when `VisitCXXConstructExpr` reaches `BuildConstructorCall`, which then takes the `ActOnParenListExpr` branch and builds a parenthesised list where a call argument belongs. I could not find a test for that shape, so I do not know whether it is already broken. If the reset went where call arguments are visited it would cover both cases, and this override would not have to know about constructors at all.

vgvassilev (commented):
`CXXDefaultInitExpr` has the same blind spot and this change leaves it open. RecursiveASTVisitor skips it for the same reason it skips `CXXDefaultArgExpr`: both have an empty `children()`, and both `DEF_TRAVERSE_STMT` bodies are guarded by `shouldVisitImplicitCode()`, which neither `VariedAnalyzer` nor `VariedChecker` turns on. clad does differentiate through it -- `ReverseModeVisitor::VisitCXXDefaultInitExpr` is `Visit(DIE->getExpr(), dfdx())`, the same shape as the default-argument one. So a default member initialiser that reads a global loses its dependency exactly the way a default argument did. In `double g; struct S { double a = g; }; double f(double x) { g = x; S s{}; return s.a; }` the analyser's `TraverseInitListExpr` hands RAV the `CXXDefaultInitExpr` and stops there, and `isVaried` answers no. That is two more overrides in the two files you already touched and one more case in the test. Better done here than filed as a second bug.

Forward mode reads both predicates you changed, and the test does not exercise it. `DiffRequest::isVaried` is what `BaseForwardModeVisitor.cpp:1153` asks before concluding a call with no base tangent contributes nothing, and `recordCallActivity`, which the varied analyser now fills differently, is what `shouldHavePushforward` answers at `BaseForwardModeVisitor.cpp:1355`. Every case in `DefaultArgumentAdjoint.C` goes through `clad::gradient`. One `clad::differentiate` over the `scalar::f1` shape would cover the forward side; as it stands, a wrong answer there is silent.

I read the code and did not build it. Each of the three hunks does have a case that fails without it, which I traced by hand: `scalar::f2` on the default RUN line for the `DiffPlanner` hunk -- with every argument defaulted, `isVaried` answering false made the whole call non-differentiable and left `_d_b` at zero; `scalar::f2` under `-enable-va` for the analyser hunk; and `effects::f2` for the `SaveAndRestore`, the only case in the file where the call initialises a local and the default argument's constructor takes more than one argument. This change is independent of #2068: the two branches diverge from a common base, neither contains the other, the hunks do not touch, and this one is green on all 43 checks including "Tests observe the change", which #2068 fails. It can land first and on its own.

vgvassilev:
One more thing, now that the earlier rounds have been addressed: the title no longer describes the change.

"Fix differentiation of default arguments with implicit conversions" was written for the type change in the first commit — `GetValueType(arg->getType())` in place of the parameter type. That change is gone, and rightly so: Sema has already converted a default argument to the parameter type before clad sees it, so it was the identity.

What the pull request does now is two different things. The varied analyser was losing a dependency through `CXXDefaultArgExpr`, because RecursiveASTVisitor skips it unless `shouldVisitImplicitCode()` is on, so a default argument reading a global answered "not varied". And `m_TrackVarDeclConstructor` stayed set for a constructor reached through a default argument rather than as a variable's initialiser.

Could you retitle it to say that? Something like "Keep the varied analysis from losing default-argument dependencies" would let the commit log point at what landed. Worth squashing the six commits while you are there — three of them are `Update lib/Differentiator/ReverseModeVisitor.cpp` and `fix test among clang multiple version`, which will not mean much to whoever bisects here later.

Asked by ai-vgvassilev.

PR #2068, opened 2026-09-12
Use zero_like to build default reverse-forward adjoints

fogsong233:
This complements #1932 by generating a default reverse-forward result for memory types returned by value when no suitable reverse_forw helper is available. For a class or container type with a valid zero_like customization, users can define that type-level customization once instead of writing separate reverse_forw helpers for each ordinary value-returning function. Explicit reverse_forw helpers remain necessary when runtime aliasing or additional per-call state must be preserved.

This is a part of my GSoC task, which greatly helps the customized functions related to torch::Tensor.

### DIFF
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -1798,6 +1798,10 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
       for (const auto& decl : m_ParentReq->getVariedDecls())
         request.addVariedDecl(decl);
 
+    // Keep the request whose body contains E. m_ParentReq is replaced with
+    // the callee request below while its body is traversed, but the planning
+    // fact belongs to the containing body.
+    DiffRequest* bodyRequest = m_ParentReq;
     llvm::SaveAndRestore<DiffRequest*> Saved(m_ParentReq, &request);
     if (request.Function->getDefinition())
       request.Function = request.Function->getDefinition();
@@ -1877,6 +1881,11 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
 
       //  Recurse into call graph.
       TraverseFunctionDeclOnce(request.Function);
+
+      // The reverse-forward request for this function is built from the same
+      // primal body as its pullback. Carry the call-site planning facts over
+      // so both visitors consume the same decision.
+      forwPassRequest.inheritDefaultReverseForwCallsFrom(request);
 
       if (requestTBR) {
         TimedAnalysisRegion R("TBR " + request.BaseFunctionName);
@@ -1992,6 +2001,16 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
       forwPassRequest.UseRestoreTracker = shouldUseRestoreTracker;
       QualType returnType = request->getReturnType();
       bool hasCustomReverseForw = LookupCustomDerivativeDecl(forwPassRequest);
+
+      // Whether the default reverse-forward adjoint is applicable is a
+      // call-site planning decision. Record it on the request that owns this
+      // body; ReverseModeVisitor only consumes the result while generating
+      // the call. A custom reverse_forw remains authoritative.
+      if (bodyRequest && !hasCustomReverseForw &&
+          (bodyRequest->Mode == DiffMode::reverse_mode_forward_pass ||
+           !nonDiff) && utils::isMemoryType(returnType) &&
+          !returnType->isReferenceType() && !returnType->isPointerType())
+        bodyRequest->recordDefaultReverseForwCall(E);
 
       if (hasCustomReverseForw ||
           (!hasCustomPullback &&
--- a/lib/Differentiator/DiffPlanner.cpp
+++ b/lib/Differentiator/DiffPlanner.cpp
@@ -2008,8 +2008,9 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
       // the call. A custom reverse_forw remains authoritative.
       if (bodyRequest && !hasCustomReverseForw &&
           (bodyRequest->Mode == DiffMode::reverse_mode_forward_pass ||
-           !nonDiff) && utils::isMemoryType(returnType) &&
-          !returnType->isReferenceType() && !returnType->isPointerType())
+           !nonDiff) &&
+          utils::isMemoryType(returnType) && !returnType->isReferenceType() &&
+          !returnType->isPointerType())
         bodyRequest->recordDefaultReverseForwCall(E);
 
       if (hasCustomReverseForw ||
--- a/lib/Differentiator/ReverseModeVisitor.cpp
+++ b/lib/Differentiator/ReverseModeVisitor.cpp
@@ -3046,9 +3046,7 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
     // independent, structurally correct adjoint.
     // An explicit reverse_forw above still takes precedence when a call needs
     // to save additional state or preserve reference/aliasing semantics.
-    bool needsDefaultAdjoint =
-        !nonDiff || m_DiffReq.Mode == DiffMode::reverse_mode_forward_pass;
-    if (needsDefaultAdjoint && needsForwPass &&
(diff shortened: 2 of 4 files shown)


vgvassilev:
Can we try to hit the test coverage with a minimal change to the existing tests or a small new one?

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3028,17 +3028,45 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+    // A call returning a memory type by value needs no handwritten
+    // reverse-forward helper when a zero_like customization exists: the
+    // original call supplies the primal and clad::zero_like supplies an
+    // independent, structurally correct adjoint.
+    // An explicit reverse_forw above still takes precedence when a call needs
+    // to save additional state or preserve reference/aliasing semantics.
+    bool needsDefaultAdjoint =
+        !nonDiff || m_DiffReq.Mode == DiffMode::reverse_mode_forward_pass;
+    if (needsDefaultAdjoint && needsForwPass &&
+        !returnType->isReferenceType() && !returnType->isPointerType()) {
```
vgvassilev: Isn't that a DiffPlanner level decision? Can it move there?

fogsong233: This block does not decide whether an adjoint is needed. That decision has already been made by the surrounding reverse-mode context and request planning.

This block only constructs an adjoint when no explicit reverse_forw is available. It first checks whether zero_like is viable and uses it to construct the adjoint; if not, it preserves the existing getZeroInit fallback.

Previously, Clad synthesized a type-based zero value, such as 0, 0.0f, or {0, 0} for an aggregate. The new behavior considers zero_like first, which allows types such as Torch containers to provide their own structurally correct adjoint initialization.

I initially make this due to Torch::Tensor, but I agree that this flag is too restrictive and a little weried. While after thinking,  i think zero_like is the general customization point for constructing adjoints, I will revise the code to try it for all eligible adjoint values. An explicit reverse_forw still takes precedence, and getZeroInit remains the fallback when no viable zero_like overload exists.

fogsong233: I reconsidered this approach. Using `zero_like` universally introduces runtime overhead, especially for scalar types, while its main benefit is for types whose adjoints depend on runtime shape or other structural metadata. We could add a special-case branch to use `getZeroInit` for scalars before trying `zero_like`, but that would make the logic more ad hoc.
I would therefore keep `zero_like` narrowly scoped. In other words, we should only use it where it is explicitly needed for structurally correct adjoint construction. Types that do not need `zero_like` should continue using the existing `getZeroInit` path, while types that require shape or metadata preservation can opt into `zero_like` explicitly.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3028,17 +3028,43 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+    // to save additional state or preserve reference/aliasing semantics.
+    if (m_DiffReq.shouldGenerateDefaultReverseForw(CE) && needsForwPass &&
+        !returnType->isReferenceType() && !returnType->isPointerType()) {
+      auto storeForReverse = [this](Expr* value, llvm::StringRef prefix) {
+        if (isInsideLoop)
+          return GlobalStoreAndRef(value, prefix, /*force=*/true);
+        return StoreAndRef(value, direction::forward, prefix,
+                           /*forceDeclCreation=*/true);
+      };
+      Expr* callRef = storeForReverse(call, /*prefix=*/"_t");
```
vgvassilev, line 3053: Problem: the store is emitted before anything has checked that `zero_like` is viable, so when it is not the stored value is pure cost. Inside a loop `storeForReverse` calls `GlobalStoreAndRef` with force, which builds a tape and pushes the returned object on every iteration, and the adjoint handed back on that path is a fresh `getZeroInit`, so nothing in the reverse sweep ever reads what was pushed. For the types this PR is aimed at, a container or a tensor, that keeps every intermediate result alive for the length of the loop.

This PR reaches that path itself. `clad::zero_like` has overloads only for arithmetic and enum types and for ranges, `FallbackMemoryValue` is neither, and `make_fallback_memory_value` is called inside a `for` loop, so the tape is built there. Nothing notices because the test has no CHECK on that function's generated code.

How to address it: decide viability before storing, ideally in the planner as the review body suggests, so this block runs only when the adjoint will actually be built. Whichever way it lands, add CHECK lines for `fallback_memory_loss` so the generated code of the fallback is pinned and a future change cannot quietly put the tape back.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -3028,17 +3028,43 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+
+      if (Expr* zeroLike = GetCladZeroLike(CloneNode(callRef))) {
+        Expr* adjointRef = storeForReverse(zeroLike, /*prefix=*/"_r");
+        return StmtDiff(callRef, adjointRef);
+      }
+      call = callRef;
+    }
+
+    if (OCE)
```
vgvassilev, line 3061: Problem: an operator call can now come back with an adjoint where it never had one. Before this change a `CXXOperatorCallExpr` returned `StmtDiff(call)` with a null derivative and left the function immediately; now the new block sits in front of that early return, so a marked operator call returns `StmtDiff(callRef, adjointRef)` instead. Nothing in the planner's condition excludes operators, and an `operator+` returning a container by value is exactly the shape that qualifies. Callers that branch on whether the derivative is null take a different path than before, and no test covers it.

How to address it: if giving operator calls a default adjoint is intended, and for a container-returning `operator+` it probably is, add one such case to the test so the new behaviour is recorded. If it is not intended, move the `if (OCE) return StmtDiff(call);` above the new block so operators keep their old path.

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1993,6 +2002,17 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
       QualType returnType = request->getReturnType();
       bool hasCustomReverseForw = LookupCustomDerivativeDecl(forwPassRequest);
 
+      // Whether the default reverse-forward adjoint is applicable is a
+      // call-site planning decision. Record it on the request that owns this
+      // body; ReverseModeVisitor only consumes the result while generating
+      // the call. A custom reverse_forw remains authoritative.
+      if (bodyRequest && !hasCustomReverseForw &&
+          (bodyRequest->Mode == DiffMode::reverse_mode_forward_pass ||
```
vgvassilev: Problem: this disjunct makes the same call behave differently depending on which body it sits in, and nothing says why. Everywhere else a non-differentiable call is skipped, which is what keeps `make_memory_value(1.0)` in `nonactive_memory_loss` on the old path and gives that test its CHECK. Inside a reverse-forward body the same call would instead be stored and given a `zero_like` adjoint. If that is load-bearing, a reader cannot tell what breaks without it; if it was added to make one case pass, it is the kind of condition that outlives its reason.

How to address it: say in the comment which case needs it, or drop it and see whether anything fails. A non-differentiable call inside a reverse-forward body, with a CHECK on the generated code, would settle it either way.

vgvassilev (commented):
Problem: the decision is now split across the planner and the visitor, and the two halves do not agree on when it applies. The planner marks a call knowing the return type is a memory type and that there is no custom reverse_forw, but not whether clad::zero_like has a viable overload for that type. The visitor learns that only after it has already emitted the store, which is the inline note below. The visitor then re-checks needsForwPass && !isReferenceType && !isPointerType, which is the same condition the planner already applied, since needsForwPass is utils::isMemoryType(returnType) two hundred lines above, under a comment that reads "FIXME: Decide this in the diff planner".

How to address it: put the overload probe in the planner too. It already calls LookupCustomDerivativeDecl there, and GetCladZeroLike decides viability with m_Builder.noOverloadExists, so the same probe can run at planning time against the return type. Then shouldGenerateDefaultReverseForw means what its name says, the visitor block loses its guard and its early-out, and there is one copy of the condition rather than two that can drift.

I did not build or run anything. CI is green on all 44 checks and the "Tests observe the change" check passes, so the new test does fail on master.

PR #2070, opened 2026-09-14
Differentiate vector::resize in reverse mode.

Vedant2005goyal:
The current implementation stores the old size in a global vector, which can cause multiple definitions and is not thread-safe. By the time the pullback runs, the vector has been restored to its size before the resize, so the old size can be read directly from it. Value-initializing new adjoints instead of filling them with 0 allows the element type to be a class type.
For example:-
```cpp
struct Foo {
  Foo() = default;
};

std::vector<Foo> v;
v.resize(5);  // Works with value-initialization.
```


vgvassilev (approved):
Lgtm!

PR #2071, opened 2026-09-14
Give pointers returned by const methods their adjoints

Vedant2005goyal:
Use returnsAdjoint when determining whether a returned value needs an adjoint instead of relying on isMemoryType, which incorrectly excludes const pointers. Keep the adjoint type non-const so the generated pullback can write to it.

```cpp
struct st {
    int value;
    const int* func() const {
        return &value;
    }
};
```
Currently Clad did not generate an adjoint for st::func() because its return type is const int*, causing the generated pullback to fail.


lib/Differentiator/CladUtils.cpp
```diff
@@ -1214,6 +1214,10 @@ namespace clad {
       }
     }
 
+    bool returnsAdjoint(QualType T) {
+      return isMemoryType(T) || T->isPointerType();
```
vgvassilev, line 1218: Problem: the two halves overlap, so the expression is harder to read than the rule it encodes. `isMemoryType` answers a pointer with `!getPointeeType().isConstQualified()`, which `T->isPointerType()` already subsumes, so the pointer branch of `isMemoryType` is dead here. What is left after the overlap is: any pointer, a reference whose referent is writable, or a record holding memory. That is the actual rule, and a reader has to derive it.

How to address it: say the rule in the doc comment on the declaration rather than leaving it to be inferred, and consider spelling the predicate out directly so the reader sees the three cases. Worth naming why a pointer now qualifies regardless of constness too: the pullback writes through the adjoint, not through the returned pointer, so the primal's constness does not decide it.

test/Gradient/MemberFunctions.C
```diff
@@ -792,6 +792,50 @@ float fn12(const B b, const float* in) {
 // CHECK-NEXT:      b.scale_pullback(in, &res, _d_b, &_d_res);
 // CHECK-NEXT:  }
 
+struct C {
+  double d[2];
+  const double* data() const { return d; }
```
vgvassilev, line 797: Problem: the test uses `const double*` while the report this fixes is about `const int*`. I ran your own example on master and it fails with a different message than the double case, `cannot initialize a parameter of type 'int *' with an rvalue of type 'const int *'`, because the adjoint of an int member is an int rather than a double. The mechanism you changed is type-agnostic so it very likely covers both, but the type that was reported is the one not being tested.

How to address it: add the reported shape as a second case, a struct with an `int` member and a `const int* func() const`, so the fix is pinned against the case that prompted it.

vgvassilev (commented):
Problem: a const method returning a const reference fails the same way, and this change does not cover it. I built the analogue of your example on master, const double& first() const { return d[0]; } called through double use(const double&), and it fails with cannot take the address of an rvalue of type 'const double'. The cause is the same one you found: isMemoryType answers a reference by asking whether the referent is writable, so a const reference returns false and no reverse-forward is generated, exactly as it did for a const pointer. returnsAdjoint restores the pointer half of that and leaves the reference half as it was, so the next report of this reopens the issue one line away from the code this PR touches.

How to address it: make the predicate symmetric, isMemoryType(T) || T->isPointerType() || T->isReferenceType(), with a test for the const-reference return beside the const-pointer one. If references turn out to need more than that, say so in the commit message and leave them out deliberately, so the asymmetry is a decision rather than an oversight.

I did not build or run the change itself. CI is green on all 44 checks and the baseline check passes, so the new test does fail on master. Everything I checked about master's behaviour, including the two failures quoted here, comes from running it against a local build of master, not of this branch.

vgvassilev (approved):
LGTM!

Issue #2075, opened 2026-09-15
Forward mode discards the tangent of an output parameter


vgvassilev:
`clad::differentiate` cannot return the derivative of a result written through a reference parameter. It computes it and throws it away.

```cpp
void fn(double i, double j, double k, double& res) { res = i * j * k; }

  auto d = clad::differentiate(fn, "i");
  double res = 0;
  d.execute(2, 3, 4, res);   // res == 24; dres/di is nowhere
```

  generates

```cpp
void fn_darg0(double i, double j, double k, double &res) {
  double _d_i = 1;
  double _d_j = 0;
  double _d_k = 0;
  double _d_res = 0;
  double _t0 = i * j;
  _d_res = (_d_i * j + i * _d_j) * k + _t0 * _d_k;   // correct
  res = _t0 * k;
}                                                      // and dropped
```

`_d_res` is right. It is a dead local, and the function returns void, so the caller cannot read it.

The derivative itself is not missing. The pushforward already models a reference output, taking `double& _d_res` and assigning through it, so the value is reachable by wrapping the call:

```cpp
double outer(double i) { double r = 0; fn(i, 3, 4, r); return r; }
clad::differentiate(outer, "i").execute(2);   // 12.00
```

Reverse mode needs no workaround: name the output in the argument list and seed its adjoint. That path is covered by #2073 and is what closes #396.

## Why the direct form does not work

Two deliberate decisions meet here.

`BaseForwardModeVisitor::SetupDerivativeParameters` returns at line 282 when the mode is `DiffMode::forward`, before any derivative parameter is added, so `f_darg0` receives the primal's parameters and nothing else. `GenerateSeeds` then materialises every tangent as a local.

`ExtractDerivedFnTraitsForwMode` in FunctionTraits.h:758 pins that from the other side: the derived type is the primal's own signature with references and pointers stripped.

## Path forward

Two options, both interface changes rather than fixes.

  1. Give the forward derivative tangent parameters for non-const reference and pointer parameters, matching what the pushforward already does. The most faithful option and the most invasive: it changes the shape of every forward derivative, so the traits, `CladFunction::execute`, the overload builder and the 35 existing `_darg` CHECK lines that print a reference parameter all move with it.

  2. Narrower: when the primal returns void, let the forward derivative return the tangent of its single differentiable output parameter. Only
[...]

PR #2080, opened 2026-09-15
Return the tangent of an output parameter in forward mode.

ShresthSamyak:
Implements option 2 from #2075: a void-returning primal's forward derivative returns the tangent of the parameter it writes through, instead of computing it into a local the caller cannot read.

```cpp
void fn(double i, double j, double k, double& res) { res = i * j * k; }

auto d = clad::differentiate(fn, "i");
double res = 0;
double dres = d.execute(2, 3, 4, res);   // res == 24, dres == 12
```

which is what the wrapper in the issue already produced, without the wrapper.

## What moves

The signature is decided in two places and they have to agree:

- `utils::GetDerivativeType` returned `FD->getType()` for forward mode. It now builds a type returning the tangent of the sole output parameter when the primal returns void.
- `ExtractDerivedFnTraitsForwMode` has to say the same, or `CladFunction::execute` calls through a type whose return it has wrong. Written recursively rather than as a fold expression: `test/` compiles these headers at `-std=c++14` in three places.

`BaseForwardModeVisitor` then returns the tangent `GenerateSeeds` already created.

## Scope, and why

**An output parameter is a non-const lvalue reference to a floating point type.** An `int&` a function counts through is not one, nor is a reference to an enum, which has no tangent for the derivative to find. Pointers are left out -- how many elements one addresses is not something the signature says, so there is no tangent of known shape to return.

**Member functions are left out.** `ExtractDerivedFnTraitsForwMode` describes a member's derivative through the primal's own member pointer type, which has no room for a changed return type. I checked what happens without the exclusion: the generated code returns `double` while the traits still say `void`, so a caller that ignores the result calls through a wrongly-typed pointer. A tangent reachable only through a wrapper is better than that, so methods, lambdas and functors keep today's behaviour.

**More than one output parameter is an error, not a guess.** Forward mode hands back one tangent; any choice among several outputs is arbitrary, and the wrapper still says which one is wanted. A function with no output parameter keeps its void return and is untouched -- which is why `fn_with_no_return` and `c_fwd` needed no changes, contrary to what the issue expected.

## Still open

I asked two questions on the issue that this PR answers one way; both are cheap to change if you prefer the other:

- the tangent type is plain `double` for `double&`, rather than the `ValueAndPushforward` pair the pushforward produces
- pointers are out of scope, as above

If option 1 lands later, the void-returning case becomes a special case of tangent parameters and the multi-output diagnostic goes away.

## Testing

`ninja check-clad` on LLVM/Clang 18.1.3, Ubuntu 24.04, asserts enabled: **287 passed, 0 failed**. The reproducer returns 12.00 at `-std=c++14`, `c++17` and `c++20`. `git clang-format-18` against the merge base reports no changes.

- `test/ForwardMode/ReferenceArguments.C` -- the shape above, a static member function, a const reference that is not an output, a void function with no output parameter keeping its void return, and an early `return;` against the tail return. It executes and checks both the returned tangent and the output parameter.
- `test/ForwardMode/OutputParamErrors.C` -- the multi-output diagnostic, an `int&` counter and an enum reference that are not outputs, and a `void` custom `_darg0` that no longer matches, with `-verify`.
- `test/Documentation/CustomDerivatives/OutputParameter.cpp` -- a tangent-returning custom `_darg0` being used rather than clad deriving the primal. It is the source `docs/userDocs/source/user/CustomDerivatives.rst` quotes, so the documented signature is compiled and run.


ShresthSamyak:
All three of the Copilot findings were real. I reproduced each before changing anything, and each was a wrong answer rather than a style point, so thank you to whoever wired that up. Pushed fixes for all three, plus tests that fail without them.

**Static member functions.** The one that mattered most. `&A::sf` has a plain function pointer type, so the traits took the free-function path and declared a tangent return, while the AST side skipped every `CXXMethodDecl` and generated void. `execute` then read a register nothing had written:

```
before:  ret=9.00 out=9.00      (9.00 is `out` arriving by luck, not the derivative)
after:   ret=6.00 out=9.00
```

The rule now lives in one place, `utils::CanReturnOutputTangent`, which both sides call: void return, and not an *instance* method. A static method is addressed as a function pointer and so is in scope; an instance method is described through a member pointer type that cannot carry a changed return type and so is not.

**Early return.** Worse than losing the tangent -- it lost the path. `VisitReturnStmt` drops a valueless return, so the derivative ran past it:

```cpp
void early(double x, double& out) { if (x < 0) return; out = x * x; }
```

```
before:  if (x < 0) { }          // and then wrote out = x*x regardless
         d(-2, out) -> ret=-4.00, out=4.00
after:   if (x < 0) return _d_out;
         d(-2, out) -> ret=0.00,  out=-1.00   (primal leaves out alone)
```

Worth noting this half is a pre-existing bug this feature exposed rather than caused: the unconditional write was there before, with only the tangent missing to make it visible.

**C-variadic functions.** Same mismatch as the static methods, via `void(Args..., ...)` not matching the `void(Args...)` specialization. Added the specialization as suggested, so those now get the tangent too rather than disagreeing about it.

On the clang-tidy suggestions: I took the four `_t` ones, which are C++14. I have left the two `_v` ones (`is_arithmetic_v`, `is_same_v`) -- they are C++17, and this header is compiled as C++14 by part of the suite, which is the reason the trait is recursive in the first place. Checked rather than assumed:

```
$ clang++ -std=c++14 -fsyntax-only -  <<< 'static_assert(std::is_arithmetic_v<double>, "");'
error: no template named 'is_arithmetic_v' in namespace 'std'
```

There is now a comment above the trait saying so, so the next run does not re-litigate it.

`check-clad` is 235 passed, 0 failed, and the reproducer still returns 12.00 at `-std=c++14`, `c++17` and `c++20`. Rebased onto master.

ShresthSamyak:
The one red row, `ubu24-gcc13-runtime20-benchmarks`, is not this change. The build died fetching an external project:

```
fatal: unable to access 'https://gitlab.inria.fr/tapenade/tapenade.git/':
       Failed to connect to gitlab.inria.fr port 443 after 133423 ms
Failed to clone repository: 'https://gitlab.inria.fr/tapenade/tapenade.git'
Build step for tapenade_kit failed: 2
```

There is no compiler diagnostic anywhere in that log, and the same row is green on master either side of this. Rebased onto `afa45bf6`, which gives it another go.

Everything else is green now, including the two that were not: `codecov/patch` is 100% of the diff, and clang-tidy says LGTM -- the remaining `modernize-type-traits` suggestions are the two `_v` spellings, which are C++17 and break the C++14 rows, so they are silenced with a NOLINT block naming that reason rather than left to be re-posted on every push.

`check-clad` locally: 281 passed, 0 failed.

lib/Differentiator/CladUtils.cpp
```diff
@@ -1461,8 +1480,27 @@ namespace clad {
...
       ASTContext& C = S.getASTContext();
-      if (mode == DiffMode::forward)
+      if (mode == DiffMode::forward) {
+        // A function returning void leaves its result in a parameter, and the
+        // tangent of that parameter is the only thing its derivative has to
+        // give back. Returning it is what keeps the derivative reachable: the
+        // signature is otherwise the primal's, so the tangent would be a local
+        // the caller cannot read. More than one such parameter is left to the
+        // caller of this function to diagnose -- there is no way to choose.
+        if (utils::CanReturnOutputTangent(FD)) {
```
vgvassilev, line 1503: A third place decides this signature, and it is not in step with the other two. `getOverloadExpr` in DiffPlanner.cpp calls `GetDerivativeType` with `forCustomDerv=true`, and this branch ignores that flag, so the type it hands back is also the one custom derivatives are matched against. The match is `MatchFunctionTypeWithQualifiedParams`, which compares whole function types, return type included -- `DiagnoseSignatureMismatch` even has a "has different return type" note for it. So a user who today writes `void f_darg0(double, double&)` next to a void `f` stops being found after this change, and clad quietly differentiates `f`'s definition instead of using the derivative they wrote. That is the kind of thing a custom derivative usually exists to prevent. Either honour `forCustomDerv` here and keep the old type for lookup, or accept the new signature for custom derivatives too and say so in the documentation.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -143,6 +143,26 @@ DerivativeAndOverload BaseForwardModeVisitor::Derive() {
...
+  // parameter it writes through. With several such parameters there is no one
+  // tangent to return, and picking one would be a guess, so say so rather than
+  // hand back a derivative whose result cannot be read. Differentiating a
+  // wrapper that returns the wanted output remains the way to ask for it.
+  if (m_DiffReq.Mode == DiffMode::forward &&
+      utils::CanReturnOutputTangent(FD)) {
+    llvm::SmallVector<const ParmVarDecl*, 2> outputParams;
+    utils::CollectOutputParams(FD, outputParams);
+    if (outputParams.size() > 1) {
+      SourceLocation L = m_DiffReq->getLocation();
```
vgvassilev: The error points at the function rather than at the `clad::differentiate` call that asked for it. `m_DiffReq->getLocation()` is the primal's declaration. Nothing is wrong with the function -- it is a perfectly good void function -- and if it came from a header the user does not own, the caret lands in someone else's file with no indication of which line of theirs caused it. The three diagnostics a few lines above in this same function already use `m_DiffReq.Args->getBeginLoc()`, which is the argument list of the `clad::differentiate` call. A note at the function, pointing at the parameters it writes through, would finish it off. The `-verify` test encodes the current location, so it moves with this.

```suggestion
      SourceLocation L = m_DiffReq.Args ? m_DiffReq.Args->getBeginLoc()
                                        : FD->getLocation();
```

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -201,6 +221,14 @@ DerivativeAndOverload BaseForwardModeVisitor::Derive() {
...
     else
       addToCurrentBlock(BodyDiff);
+    // The primal wrote its result through a parameter and returned nothing, so
+    // GetDerivativeType gave the derivative that parameter's tangent as its
+    // return type. Hand the tangent back; without this it stays a local the
+    // caller has no way to read.
+    if (Expr* outputTangent = GetOutputParamTangent())
+      addToCurrentBlock(
+          m_Sema.ActOnReturnStmt(noLoc, outputTangent, getCurrentScope())
+              .get());
```
vgvassilev: Two things here. `noLoc` leaves the generated `return` with no source location, and `VisitorBase::BuildReturnStmt` exists for exactly this -- its comment says "the `return` keyword is not one the user wrote, so the location comes from here rather than from the caller". Master has spent several commits moving generated nodes onto real locations so a debugger can step through clad's output, and this adds two that have none; the other is the same call in `VisitReturnStmt`.

Second, this runs unconditionally after the body, so a primal that already ends in `return;` gets a second, unreachable `return` -- `VisitReturnStmt` emitted one for it a moment earlier. Checking whether the block already ends in a `ReturnStmt` would keep dead code out of the user's translation unit.

```suggestion
    if (Expr* outputTangent = GetOutputParamTangent())
      addToCurrentBlock(BuildReturnStmt(outputTangent));
```

vgvassilev (commented):
The rule for what counts as an output parameter is wider than the design it serves, and it is wide in a way that reaches an assert. `CollectOutputParams` accepts any non-const lvalue reference whose pointee is `isRealType()`, and Clang's `isRealType()` is true for a complete unscoped enum. `IsOutputParam` on the header side asks `std::is_arithmetic`, which is false for an enum. So `void f(double i, Status& s)` has one output parameter on the AST side and none in the traits -- the same disagreement the static-method fix closed, reopened through a different door. The crash comes before the wrong signature does: `IsDifferentiableType` refuses enums, so `GenerateSeeds` never builds a tangent for that parameter, and `GetOutputParamTangent` then trips its own `assert(found != m_Variables.end())`. An assertion failure on a valid program, in any asserts build.

`int&` and `bool&` are the same predicate seen from the other side. Both are real and both are arithmetic, so the two halves agree, but neither is a tangent anyone asked for. The visible cost is the new error: `void f(double x, double& out, int& iterations)` has two output parameters by this rule, so code that compiles today stops compiling, and the user is told to write a wrapper for a function that has exactly one real output. Narrowing both halves to floating point -- `isRealFloatingType()` on the AST side, `std::is_floating_point` in the traits -- removes the enum divergence, the assert and the spurious error together. I have not put that as a suggestion inline, because changing one half on its own makes the disagreement worse rather than better.

I read this rather than building it, so the assert above is traced through `CollectOutputParams` to `GetDerivativeType` to `GenerateSeeds` to `GetOutputParamTangent` in the source, not observed running. Nothing in `test/` differentiates a void primal through a reference parameter that is not `double&`, so neither the enum case nor the two-output refusal has a test that would notice either way. CI is green on all 43 rows.

This review was produced by ai-vgvassilev.

ShresthSamyak:
All four taken. I built the two you traced rather than ran, and both reproduce exactly as you described.

**The predicate was too wide.** Confirmed before changing anything:

```
void f(double i, Status& s)           -> Assertion `found != m_Variables.end() && found->second.Decl
void f(double x, double& out, int& n) -> error: ... writes through 2 parameters ...
```

The first is the assert you traced, on a valid program, in any asserts build. The second refuses a function with exactly one real output. Both halves are now floating point -- `isRealFloatingType()` on the AST side, `std::is_floating_point` in the traits -- and both cases compile. The comment on each half now says the other exists and has to agree with it, since this is the second time the two drifted apart.

**The diagnostic location.** Taken as suggested, with the note:

```
d_two.cpp:3:48: error: attempted to differentiate 'two', which returns void and writes through 2 parameters; ...
    3 | int main() { auto d = clad::differentiate(two, "i"); }
      |                                                ^~~
d_two.cpp:2:28: note: 'a' is written through here
d_two.cpp:2:39: note: 'b' is written through here
```

**`BuildReturnStmt` and the unreachable return.** Both taken. The generated returns carry a location now, and a primal already ending in `return;` gets one return rather than two -- the block is checked for a trailing `ReturnStmt` before the tail one is added. `test/ForwardMode/ReferenceArguments.C` pins that shape.

**Custom derivatives.** You are right that `getOverloadExpr` passes `forCustomDerv=true` and this branch ignores it, and right that the matched type changes. One correction to the consequence, though: clad does not quietly differentiate the definition. It refuses, and says what it wanted:

```
error: user-defined derivative for 'g' was provided but not used;
       expected signature 'double (double, double &)' does not match
note: candidate 'g_darg0' has different return type ('double' expected but has 'void')
```

So it is a breaking change for anyone with such a custom derivative, but a loud and actionable one rather than a silent substitution.

I considered honouring `forCustomDerv` and keeping the old type for lookup, and did not, because it reopens the hole the static-method fix closed: the traits would still describe the derivative as returning a tangent, so a `void` custom derivative found by that lookup would be called through a wrongly typed pointer. The two have to agree, and the lookup type is one of them. So the new signature applies to custom derivatives too, and `docs/userDocs/source/user/CustomDerivatives.rst` now says so with an example.

Happy to reverse that if you would rather the old signature keep matching -- it would want the traits taught about it as well, which is a bigger change than this PR.

New coverage for the narrowing in `test/ForwardMode/OutputParamErrors.C`: an `int&` counter and an enum reference, each alongside one real output, both differentiating rather than refusing.

`check-clad` is 285 passed, 0 failed; the reproducer returns 12.00 at `-std=c++14`, `c++17` and `c++20`; format clean against the merge base. Rebased onto master.

docs/userDocs/source/user/CustomDerivatives.rst
```diff
@@ -44,6 +44,29 @@ a pointer type. This case will be explained in more detail later.
...
+   .. code-block:: cpp
+
+      void f(double x, double& out);
+
+      namespace clad {
+      namespace custom_derivatives {
+        // `double`, not `void`: the tangent of `out` is what the caller reads.
+        double f_darg0(double x, double& out);
+      }
+      }
```
vgvassilev: That needs to align with the new docs infrastructure which now puts these into a runnable state in test/Documentation.

vgvassilev (commented):
All four are answered, and two of them better than I asked. Both halves of the predicate moved to floating point together, which closes the enum divergence, the assert behind it and the spurious two-output refusal in one move, and `OutputParamErrors.C` now pins all three cases including the `int&` counter and the `Status&`. The error points at the `clad::differentiate` call with a note on each parameter it writes through. The generated return goes through `BuildReturnStmt` and is skipped when the block already ends in one.

The tests are stronger than I expected: `ReferenceArguments.C` executes and checks both the returned tangent and the output parameter, so the claim that the caller can now read the tangent is observed rather than asserted, and `early_darg0` and `tail_darg0` cover the return-guard from both sides.

One thing is still open, and it is the finding I raised about `forCustomDerv`. You resolved it by deciding the custom-derivative signature changes too and documenting that, which is a defensible answer and better than a special case -- and I confirmed the documented claim is true: a mismatching `_darg` is not silently dropped, `DiffPlanner.cpp:1496` reports it was provided but not used and `DiagnoseSignatureMismatch` names the signature expected. But nothing tests it. The change alters what signature a user's custom `_darg` must have for a void primal, and the only thing standing between a user and a quietly unused derivative is a diagnostic no test in this pull request exercises. Two short cases would pin it: a `void f(double, double&)` with a user-written `void f_darg0(double, double&)`, expecting the provided-but-not-used error, and one with `double f_darg0(double, double&)`, checking it is picked up instead of clad deriving the primal. I looked for an existing test that would catch a regression here and there is none -- no void-returning custom `_darg` exists anywhere in `test/`, which is also why CI is green.

I read this rather than building it, so the diagnostic path above is traced through the sources on master rather than observed.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -143,6 +143,35 @@ DerivativeAndOverload BaseForwardModeVisitor::Derive() {
...
+           "attempted to differentiate '%0', which returns void and writes "
+           "through %1 parameters; forward mode returns a single tangent, so "
+           "differentiate a wrapper returning the output you need")
+          << FD->getNameAsString() << (unsigned)outputParams.size() << L;
+      for (const ParmVarDecl* PVD : outputParams)
+        diag(DiagnosticsEngine::Note, PVD->getLocation(),
+             "'%0' is written through here")
+            << PVD->getNameAsString() << PVD->getLocation();
+      return {};
+    }
```
vgvassilev: This will need to move in the DiffPlanner.

lib/Differentiator/BaseForwardModeVisitor.cpp
```diff
@@ -316,6 +357,30 @@ void BaseForwardModeVisitor::SetupDerivativeParameters(
   }
 }
 
+Expr* BaseForwardModeVisitor::GetOutputParamTangent() {
```
vgvassilev, line 331: That seems to be a good candidate for the DiffRequest.

ShresthSamyak:
All four taken, and the custom-derivative behaviour is now measured rather than documented.

**The two cases you asked for.** Both directions, against a build of master and a build of this branch:

```
double g_darg0(double x, double& out)     before: error: ... expected signature
                                                  'void (double, double &)' does not match
                                          after:  used; ret=300.00 out=9.00

void   g_darg0(double x, double& out)     before: compiles, runs, out=42.00
                                          after:  error: user-defined derivative for 'g' was
                                                  provided but not used; expected signature
                                                  'double (double, double &)' does not match
                                          note:   candidate 'g_darg0' has different return type
```

The `300.00` is deliberate: the custom derivative returns `100 * x`, which clad's own derivative never would, so the value is what shows whose body ran. That case is now `test/Documentation/CustomDerivatives/OutputParameter.cpp`, which executes and also checks the dumped derivative is the user's. The `void` case is two lines in `test/ForwardMode/OutputParamErrors.C`, expecting the error and the note.

So the answer to "nothing tests it" is that both halves fail on master now: the first does not compile, the second compiles and silently uses the wrong derivative.

**Docs infrastructure.** The `code-block` is gone; the note is a `literalinclude` of the marked region of that same test, the way `Pushforward.cpp` and the rest are quoted. The example is therefore compiled and run by lit rather than being prose that happens to look like C++. Built the docs to check the include resolves rather than assuming: `sphinx-build -W` on `docs/userDocs/source` is clean and the rendered page carries the snippet.

**The diagnostic moved to the planner.** It is in `DiffCollector::VisitCallExpr` now, next to the other checks on a request that has just been built, and the request is dropped there rather than in the visitor. The location is unchanged, so `OutputParamErrors.C` needed no edit -- which is also what says the move did not change what the user sees.

One consequence worth naming: the planner sees only the differentiation the user wrote, where the visitor saw every forward request. A multi-output void function reached through a call is a pushforward, which has its own place for both tangents and was never affected; I checked that such a call and a `clad::gradient` of the same function still compile and run.

**`GetOutputParamTangent` moved to `DiffRequest`.** `getOutputTangentParam()` answers which parameter the derivative returns the tangent of, and `getOutputParams()` the whole list for the diagnostic; `utils::GetDerivativeType` decides the type by the same rule. The visitor now reads the answer instead of recomputing it, and its `assert(outputParams.size() == 1)` is gone with it: the request returns null when there is not exactly one, so a request that reaches the visitor without the planner's check keeps the primal's void return rather than asserting.

`check-clad` is 287 passed, 0 failed; format clean against the merge base.

PR #2093, opened 2026-09-16
Fix octahedron distance function in hyperbolic_func

Kuba70:
### Summary
Fixes the signed distance function (SDF) for the octahedron geometry in `hyperbolic_func` within `SmallPT.cpp`.

### Details
- Replaced the astroid/superellipsoid power formula (`pow(..., 2./3.)`) with the exact octahedron SDF formula utilizing `std::abs()` (`|x| + |y| + |z| - r`).
- Removed outdated `//TODO` comments regarding distance function validation.
- Added code comments explaining the geometry transformation and distance calculation.

### Testing
- Verified via direct point evaluation on rotated boundary vertices. The updated function returns `0` (within double precision limits `~1e-16`), whereas the old implementation resulted in an error margin of `~0.225`.


demos/ComputerGraphics/smallpt/SmallPT.cpp
```diff
@@ -217,12 +217,13 @@ double h_func_dz(const Vec& p, const Vec& p0, double r) {
...
 double hyperbolic_func(const Vec& p, const Vec& p0, double r) {
-  return pow((p.x - p0.x) * cos_a + (p.z - p0.z) * sin_a, 2. / 3.) +
-         pow(p.y - p0.y, 2. / 3.) +
-         pow((p.x - p0.x) * -sin_a + (p.z - p0.z) * cos_a, 2. / 3.) -
-         pow(r, 2. / 3.);
+  double x_rot = (p.x - p0.x) * cos_a + (p.z - p0.z) * sin_a;
+  double y_rot = p.y - p0.y;
+  double z_rot = -(p.x - p0.x) * sin_a + (p.z - p0.z) * cos_a;
+
+  return std::abs(x_rot) + std::abs(y_rot) + std::abs(z_rot) - r;
```
vgvassilev, line 226: This is not a distance, and two things here read it as one. `|x| + |y| + |z| - r` is the distance to the nearest face plane multiplied by sqrt(3). `ImplicitSolid::intersect` sphere-traces by stepping `t += fabs(distance_func(pt))`, which needs a value that never overshoots the surface. And `HyperbolicSolid::normal` returns the clad gradient without normalising it, on the comment that a signed distance function has a unit gradient; the gradient of this one is a rotated (1, 1, 1) with signs, whose length is exactly sqrt(3), so every shading dot product comes out 1.73 times too large. Scaling by 1/sqrt(3) fixes both. This only applies if the octahedron is the shape you settle on.

```suggestion
  // 1/sqrt(3): |x|+|y|+|z|-r is the distance to the nearest face plane times
  // sqrt(3). Scaling it down never overshoots the surface, which the sphere
  // trace in ImplicitSolid::intersect needs, and makes the gradient a unit
  // vector, which HyperbolicSolid::normal assumes.
  return (std::abs(x_rot) + std::abs(y_rot) + std::abs(z_rot) - r) *
         0.5773502691896258;
```

vgvassilev (commented):
This changes one of four places that spell out the same solid, and leaves the other three saying the old thing. `HyperbolicSolid::distance_func`, further down this file, still holds the `pow(..., 2./3.)` formula, and that is the one the renderer uses: `ImplicitSolid::intersect` ray-marches by stepping `fabs(distance_func(pt))`, while `hyperbolic_func` is only ever called to get the normal. The demo would now trace one surface and shade it with the normals of a different one. The three `h_func_d*` functions under `TEST_TYPE_BY_HAND` are derivatives of the 2/3-power formula too, so that path splits the same way.

Before changing those, it is worth deciding which solid was meant, because I do not think it was an octahedron. The surface `x^(2/3) + y^(2/3) + z^(2/3) = r^(2/3)` is the astroidal ellipsoid, whose other name is the hyperbolic octahedron. That is where `hyperbolic_func` and `HyperbolicSolid` get their names, and it explains the TODO you removed: it does not look like an octahedron because it is not one, it has concave faces. The bug the TODO was chasing is somewhere else. `pow` with a negative base and a non-integer exponent is a domain error and returns NaN, so every point outside the all-positive corner of the rotated frame evaluated to NaN and there was nothing to see. Wrapping each base, `pow(std::abs(x_rot), 2./3.)` and so on, fixes that, keeps the solid the file is named after, and leaves the hand-written derivatives correct. Putting a real octahedron here instead is a reasonable thing to want, but then say so in the commit message and change the other three places with it. Worth noting the check in the description does not separate the two: the six axis vertices lie on both surfaces, so the old formula returns 0 there as well.

On tests, `test/Misc/RunDemos.C` compiles this file, so a build break here would be caught. Nothing runs it, and `scene[]` never constructs a `HyperbolicSolid`, so no test in the tree observes the maths in either direction. That is fine for a demo fix, but it does mean review is the only check this gets. The head commit also has no check runs at all, so nothing has compiled it yet.

I read the code rather than building or rendering it.

This review was produced by ai-vgvassilev.

vgvassilev:
Closing this in favour of #2142, and the reason is about the object rather than about your patch.

`HyperbolicSolid` has never been rendered. `scene[]` is nine spheres, and `git log -S"new HyperbolicSolid"` across the whole history of the file returns nothing, so the class has never been constructed at any point. `test/Misc/RunDemos.C` compiles this file and nothing runs it, so no test observes the mathematics either way. Fixing the geometry of a solid nobody has ever seen is not worth the four places it would have to be changed in.

For the record of why the octahedron is not the answer, in case it comes up again. The surface `x^(2/3) + y^(2/3) + z^(2/3) = r^(2/3)` is the astroidal ellipsoid, also called the hyperbolic octahedron, which is where `hyperbolic_func` and `HyperbolicSolid` get their names. The TODO you removed was chasing a real defect, but not that one: `pow` with a negative base and a non-integer exponent is a domain error returning NaN, so every point outside one octant of the rotated frame evaluated to NaN and there was nothing to see. Wrapping each base, `pow(std::abs(x_rot), 2./3.)`, fixes the rendering, keeps the solid the file is named after, and leaves the three hand-written `h_func_d*` derivatives correct. The octahedron formula leaves them differentiating a function that no longer exists, and leaves `HyperbolicSolid::distance_func` -- the one `ImplicitSolid::intersect` actually calls -- on the old formula, so the demo would trace one surface and shade it with the normals of another.

If you would like to carry on here, #2142 is the deletion, and the more interesting work is in `SmallPTDiff` (#1877). Thank you for the check in the description; it is a good habit, and the reason it did not separate the two candidates is that the six axis vertices lie on both surfaces, so the old formula returns zero there as well.

Written by ai-vgvassilev.

PR #2094, opened 2026-09-16
Clone the statements clad says it cannot differentiate.

### LINKED ISSUE #2088: Unsupported statements crash instead of being diagnosed
Clad warns that a statement kind is unsupported and then segfaults while cloning that same statement.

```cpp
double f(double x) { return x ?: 1.0; }
int main() { clad::differentiate(f, "x"); }
```

```
warning: statement kind 'BinaryConditionalOperator' is not supported
Stack dump:
2. Building code for '<double f(double x)>[name=f, order=1, mode=forward, args='"x"']'
 #6 clang::RecursiveASTVisitor<clad::utils::ReferencesUpdater>::TraverseStmt(clang::Stmt*, ...)
 #7 clad::VisitorBase::Clone(clang::Stmt const*)
 #8 clang::StmtVisitorBase<..., clad::BaseForwardModeVisitor, clad::StmtDiff>::Visit(...)
clang++: error: clang frontend command failed with exit code 139
```

The cause is the fallback that reports the diagnostic, `BaseForwardModeVisitor.cpp:463` and `ReverseModeVisitor.cpp:850`:

```cpp
StmtDiff VisitStmt(const Stmt* S) {
  diagUnsupported(S);
  // Unknown stmt, just clone it.
  return StmtDiff(Clone(S));
}
```

Having said it cannot handle the node, it clones it anyway, and `ReferencesUpdater` crashes walking the clone. Four constructs reach this, all with the same stack:

| Construct | Reproducer | Modes |
|---|---|---|
| `?:` | `return x ?: 1.0;` | forward and reverse |
| `[[likely]]` on a statement | `if (r > 0) [[likely]] { r *= 2; }` | forward and reverse |
| parenthesised aggregate init | `struct P { double a, b; }; P p(x, 2.0);` | forward and reverse |
| inherited constructor | `struct D : B { using B::B; };` | reverse only |

Not every unsupported statement crashes -- `throw`, labels and `goto` are diagnosed and the compile finishes cleanly, which is the intended behaviour. Whatever makes those safe is what these four need.

Environment: clad 96d3cb2b, clang 22.1.8, macOS x86_64.

ShresthSamyak:
Fixes: #2088

`StmtClone::VisitStmt` was `assert(0); return 0;`, so a statement kind with no case there cloned to null and `ReferencesUpdater` walked it. This closes that two ways: the fallback stops returning null, and the four kinds the issue names get a clone each.

## The fallback

A kind with no case has no clone to give, and handing back null gives the caller something it can neither use nor detect. It now says so, as an error, and stands a placeholder in the tree so it stays well formed until the compilation stops:

```
warning: statement kind 'GCCAsmStmt' is not supported
error: clad cannot differentiate this function: no clone exists for statement kind 'GCCAsmStmt'
```

That is the difference between fixing four instances and fixing the bug: `CXXFoldExpr`, `GenericSelectionExpr`, `AtomicExpr`, `CXXNoexceptExpr` and `GCCAsmStmt` still have no case, and none of them crashes any more. `test/Regressions/issue-2088-unclonable.cpp` pins it on `GCCAsmStmt`; measured both ways, the same input asserts in a debug build and segfaults in a release one without the change.

## The four clones

| construct | node |
| --- | --- |
| `x ?: 1.0` | `BinaryConditionalOperator` |
| `if (c) [[likely]] { }` | `AttributedStmt` |
| `P p(x, 2.0);` | `CXXParenListInitExpr` |
| `struct D : B { using B::B; };` | `CXXInheritedCtorInitExpr` |

Three are transcriptions of the constructor. `a ?: b` is not: it evaluates `a` once and names it with an `OpaqueValueExpr` that both the condition and the true branch point at, so cloning those on their own leaves them pointing into the original. `VisitPseudoObjectExpr` already solves this with `m_OVESubst`, so this reuses it. The common sub-expression and the opaque value's source are one node and the constructor asserts they are, so the clone is made once and shared -- an asserts build is what caught that.

`CXXParenListInitExpr` arrived in clang 16 (P0960R3), so its clone and its test are behind `CLANG_VERSION_MAJOR >= 16`. Its union form carries the initialised member in the slot the array filler occupies, so the clone carries whichever of the two is set.

With a clone in place each of the four is diagnosed and the compile finishes, and the derivative keeps the statement:

```cpp
double f_darg0(double x) {
    double _d_x = 1;
    double _d_y;
    double y = x ?: 1.;
    return _d_y * 2;
}
```

An unsupported expression sitting directly in a `return` still fails to compile, saying so rather than crashing: the fallback produces no tangent to return, which is what being unsupported means. Returning a zero tangent instead would answer with a derivative that is quietly wrong, so I have left it.

## Testing

`ninja check-clad` on LLVM/Clang 18.1.3, Ubuntu 24.04, asserts enabled: **287 passed, 0 failed**. `git clang-format-18` against the merge base reports no changes.

- `test/Regressions/issue-2088.cpp` -- the three version-independent kinds, `?:` in both forward and reverse mode so the two clones a reverse derivation makes exercise the opaque-value sharing.
- `test/Regressions/issue-2088-paren-init.cpp` -- the clang 16 kind, aggregate and union.
- `test/Regressions/issue-2088-unclonable.cpp` -- the fallback.

Each fails without the change; I reverted and re-ran rather than assuming.


ShresthSamyak:
Taken, with one honest caveat about how much it mattered.

`clad_compat::ExprSetDeps(clonedOVE, OVE)` is now there. `VisitOpaqueValueExpr` does it for the opaque values it clones itself, so the hand-built one here was simply inconsistent with it, and the fix is a line.

I did try to make the omission bite before adding it, and could not:

```cpp
template <typename T> T tf(T x) { T y = x ?: T(1); return y * 2; }
clad::differentiate(tf<double>, "x");
```

clad differentiates the instantiation, where nothing is dependent any more, so every bit `ExprSetDeps` copies is already false and copying them changes nothing observable. So this is a consistency fix rather than a demonstrated bug -- worth making, but I would rather say that than imply I found a failing case.

Worth noting the same omission is in `VisitPseudoObjectExpr`, which builds its opaque values inline the same way and does not set their dependence either. I have left that alone as out of scope here; happy to include it if you would rather it went in one go.

Also took the `misc-include-cleaner` suggestions -- `clang/AST/Expr.h` and `clang/AST/Stmt.h` in the header for the two node names, `llvm/ADT/DenseMap.h` in the implementation. I have left the `cppcoreguidelines-owning-memory` ones: `new (Ctx)` is placement new into the `ASTContext` allocator and is how every clone in this file is built, so silencing it here and nowhere else would be the odd thing.

Rechecked after the change: the four reproducers behave as before, `check-clad` is 282 passed, 0 failed, and format is clean against the merge base.

include/clad/Differentiator/StmtClone.h
```diff
@@ -132,6 +134,12 @@ namespace utils {
     DECLARE_CLONE_FN(ConstantExpr)
     DECLARE_CLONE_FN(ValueStmt)
     DECLARE_CLONE_FN(LambdaExpr)
+    DECLARE_CLONE_FN(BinaryConditionalOperator)
+    DECLARE_CLONE_FN(AttributedStmt)
+    DECLARE_CLONE_FN(CXXInheritedCtorInitExpr)
+#if CLANG_VERSION_MAJOR > 16
+    DECLARE_CLONE_FN(CXXParenListInitExpr)
+#endif // CLANG_VERSION_MAJOR > 16
```
vgvassilev: `CXXParenListInitExpr` is a clang 16 node, not a clang 17 one, so this guard leaves the crash in place on a version clad supports and tests. It landed in llvm/llvm-project 40c52159d3ee ("[clang][C++20] P0960R3 and P1975R0: Allow initializing aggregates from a parenthesized list of values"), whose first release tag is llvmorg-16.0.0, and the class is in `clang/include/clang/AST/ExprCXX.h` at llvmorg-16.0.0 with byte-for-byte the same `Create` signature it has at 17. `.github/workflows/ci.yml` builds a `clang-runtime: '16'` row (`ubu24-gcc14-runtime16`). The same `#if` in lib/Differentiator/StmtClone.cpp needs the same change.

```suggestion
#if CLANG_VERSION_MAJOR >= 16
    DECLARE_CLONE_FN(CXXParenListInitExpr)
#endif // CLANG_VERSION_MAJOR >= 16
```

lib/Differentiator/StmtClone.cpp
```diff
@@ -310,6 +311,78 @@ DEFINE_CLONE_EXPR(SubstNonTypeTemplateParmExpr,
...
+  inits.reserve(Node->getInitExprs().size());
+  for (Expr* init : Node->getInitExprs())
+    inits.push_back(init ? cast<Expr>(Clone(init)) : nullptr);
+
+  auto* result = CXXParenListInitExpr::Create(
+      Ctx, inits, CloneType(Node->getType()),
+      Node->getUserSpecifiedInitExprs().size(), Node->getInitLoc(),
+      Node->getBeginLoc(), Node->getEndLoc());
+  if (Expr* filler = Node->getArrayFiller())
+    result->setArrayFiller(cast<Expr>(Clone(filler)));
```
vgvassilev, line 387: A `CXXParenListInitExpr` that initialises a union carries the chosen member in the same slot the array filler lives in, and the clone drops it. `union U { double a; int b; }; U u(x);` in C++20 builds a node whose `getInitializedFieldInUnion()` names the field; the clone's is null, and CodeGen reads it -- `CGExprAgg.cpp` treats a null field as an empty union, which in an asserts build fires "Only unnamed bitfields or ananymous class allowed" and in a release build emits no initialisation at all. That is the crash this pull request is closing, reached through a different construct. `VisitInitListExpr` further down this file already carries both halves and explains in a comment why aggregate emission needs them. The two setters share one storage slot, so it has to be an else.

```suggestion
  if (Expr* filler = Node->getArrayFiller())
    result->setArrayFiller(cast<Expr>(Clone(filler)));
  else if (FieldDecl* fieldInUnion = Node->getInitializedFieldInUnion())
    result->setInitializedFieldInUnion(fieldInUnion);
```

test/Regressions/issue-2088-paren-init.cpp
```diff
@@ -0,0 +1,17 @@
+// RUN: %cladclang -std=c++20 %s -I%S/../../include -fsyntax-only -Xclang -verify
+// UNSUPPORTED: clang-11, clang-12, clang-13, clang-14, clang-15, clang-16
```
vgvassilev: clang 16 is the first version that has the node, and clad's CI builds a clang 16 row, so 16 is the version this test most needs to run on -- as written it is skipped exactly where the guard above leaves the crash unfixed. The comment a few lines down calling it "a clang 17 node" wants the same correction.

```suggestion
// UNSUPPORTED: clang-11, clang-12, clang-13, clang-14, clang-15
```

vgvassilev (commented):
This and #2119 fix the same issue two different ways, and this is the better of the two places to fix it. The null comes out of `StmtClone::VisitStmt`, so a case there covers the kind wherever it appears, including nested inside a construct clad does support; a deny list in the visitor fallback only sees the statement at the top. Whoever lands one of these should close the other rather than take both.

The fallback itself is untouched, though, and it is still `assert(0); return 0;`. `StmtClone` has a case for 75 statement kinds and clang has a couple of hundred, so the next kind a user writes crashes in exactly the same way -- `CXXFoldExpr`, `GenericSelectionExpr`, `AtomicExpr`, `CXXNoexceptExpr` and `GCCAsmStmt` have no case today either. Four new cases fix four instances of the bug rather than the bug. Closing it means `VisitStmt` stops handing a null back to its caller: either it returns something that can sit in a tree, or the caller learns the clone failed and declines to build the derivative at all.

On whether the tests observe the change: `test/Regressions/issue-2088.cpp` does. Revert the patch and clad hits the assert in a debug build and walks a null pointer in a release one, so the test fails either way. `test/Regressions/issue-2088-paren-init.cpp` does not observe it on the one version where it matters; see the comment on its `UNSUPPORTED` line.

Two things I did not check by running them. I read the change rather than building it, so the LLVM 16 point below rests on the clang 16 headers and on `.github/workflows/ci.yml`, not on a build. And `x ?: 1.0` is only differentiated in forward mode in the test, so the opaque-value sharing the new code exists for -- the two clones a reverse-mode derivation makes, forward sweep and pullback -- is not reached by anything here. A `clad::gradient(elvis, "x")` alongside the existing call would reach it, and the shared-node check in `DerivativeBuilder.cpp` runs even on a derivation that emitted a diagnostic, so it would catch a slip.

ShresthSamyak:
All four taken. Checked the two you flagged as read-not-run, and both hold.

**The fallback.** You are right that four cases fix four instances. `StmtClone::VisitStmt` no longer returns null: it reports that no clone exists for the kind, as an error, and leaves a placeholder so the tree stays well formed until the compilation stops. An error rather than a warning because clad cannot build this derivative at all, and rather than a node that quietly stands in for the statement, because that is a wrong answer dressed as an answer.

`GCCAsmStmt`, from your list, measured both ways on the same input:

```
before (debug)    Assertion `0 && "clone not fully implemented"' failed
before (release)  segfault in ReferencesUpdater
after             error: clad cannot differentiate this function: no clone
                         exists for statement kind 'GCCAsmStmt'
```

`test/Regressions/issue-2088-unclonable.cpp` pins that, and says in a comment that it wants a different kind rather than deleting if `GCCAsmStmt` ever gains a clone. `check-clad` is unchanged at 0 failures, so nothing was relying on the silent null.

**clang 16, not 17.** Correct, and my "clang 17 node" claim in the description was wrong; fixed there too. Checked rather than taking it on trust:

```
$ gh api repos/llvm/llvm-project/contents/clang/include/clang/AST/ExprCXX.h?ref=llvmorg-16.0.0 | grep -c 'class CXXParenListInitExpr'
1
```

with the same `Create` signature as at 17, and `ci.yml:40` is the `ubu24-gcc14-runtime16` row. Both `#if`s are now `>= 16` and the test's `UNSUPPORTED` line drops `clang-16`, so it runs on the version that most needs it.

**The union field.** Also correct. `setArrayFiller` and `setInitializedFieldInUnion` write the same slot, so it is an `else if`, and `issue-2088-paren-init.cpp` now covers `union U { double a; int b; }; U u(x);` alongside the aggregate.

**Reverse mode on `?:`.** Added `clad::gradient(elvis, "x")` next to the existing `differentiate`. It does reach the second clone -- the diagnostic count on that line went from one to two, which is what told me the sweep really is a separate derivation -- and the shared-node check stays quiet, so the opaque-value sharing is doing its job rather than being untested scaffolding.

On #2119: agreed that whichever lands, the other should close. I have no stake in which.

Rebased onto master. `check-clad` is 287 passed, 0 failed, format clean against the merge base.

ShresthSamyak:
Chasing the codecov gap found a bug in my own patch, so this push is more than a coverage fix.

The uncovered lines were the array half of the `CXXParenListInitExpr` clone. Trying to cover them showed the clone is wrong for arrays:

```cpp
double f(double x) { double arr[3](x, 2.0); return arr[0] + arr[2]; }
```

is valid C++20 that plain clang accepts, and with the clone in place it became

```
error: cannot initialize an array element of type 'double' with an rvalue of type 'double[3]'
```

The same function written `double arr[3]{x, 2.0}` compiles, which is what points at the cause: clad's array handling recognises `InitListExpr` and not this node, so even a faithful clone arrives at Sema as an element initialiser carrying the whole array's type. That is a gap in the array path rather than in the clone, and bigger than this pull request.

So the clone now refuses arrays and says so, and aggregates and unions keep working:

```
warning: statement kind 'CXXParenListInitExpr' is not supported
error: clad cannot differentiate this function: no clone exists for statement kind 'CXXParenListInitExpr'
error: cannot initialize an array element of type 'double' with an rvalue of type 'double[3]'
```

The second error is the cascade from carrying on after the first, and I have left it rather than hide it -- `issue-2088-paren-init.cpp` expects both, so a later change to either has to say so. Refusing is not a fix for arrays; it is the difference between a message naming clad and a message that reads as though the user's own line were wrong.

Two corrections to my previous comment, both mine:

- I ran the earlier expression probes against the wrong branch and drew a conclusion from a build without this change in it. On the right branch, `noexcept`, `_Generic`, `typeid`, `__builtin_bit_cast` and a default member initialiser all abort on an unrelated `Assertion 'isNull()... Cannot retrieve a NULL type pointer'` before `StmtClone` is ever reached. That one is not this patch's -- it reproduces on master unchanged -- but it does mean the `Expr` arm of the new fallback has no test: every expression kind I could find that lacks a clone dies earlier, for another reason. The arm has to exist, since `VisitorBase::Clone` casts its result to `Expr*`, and I would rather say it is untested than imply otherwise.
- That earlier assert is worth its own issue if you agree it is real; happy to open one with the reproducers.

`check-clad` is 287 passed, 0 failed; format clean against the merge base.

test/Regressions/issue-2088-paren-init.cpp
```diff
@@ -0,0 +1,44 @@
+// RUN: %cladclang -std=c++20 %s -I%S/../../include -fsyntax-only -Xclang -verify
+// UNSUPPORTED: clang-11, clang-12, clang-13, clang-14, clang-15
```
vgvassilev:
```suggestion
// UNSUPPORTED: clang-12, clang-13, clang-14, clang-15
```
clang-11 is not supported anyway in clad's master.

ShresthSamyak: Taken -- `CMakeLists.txt:42` is `set(CLANG_MIN_SUPPORTED 12.0)` and the check at line 155 refuses anything below it, so a clang-11 row cannot exist to skip.

Ten other tests still carry the same stale `clang-11`, from `test/Regressions/issue-1736.cpp` to `test/Gradient/Lambdas.C` (one of them also lists `clang-10`). Left alone here; happy to sweep them in a separate NFC commit if you want it done.

Also amended the commit message, which still called `CXXParenListInitExpr` a clang 17 node and did not mention the fallback or the array refusal, and added the `Fixes:` line.

`check-clad` reads 288 passed, 0 failed. I quoted 287 last time; I re-ran the unpatched commit alongside this one and both give 288 with an identical unsupported set, so the difference is in how I measured earlier, not in anything this push changed. Format clean against the merge base.

Issue #888, opened 2024-05-08
Incorrect gradient computation for non-zeroed derivatives


vaithak:
Function 1:
```cpp
double f1(double x) {
  x = -x;
  return x;
}
```

Function 2:
```cpp
double f2(double x) {
  double y = -x;
  return y;
}
```

Both the functions are essentially doing `f(x) = -x`, so the `df/dx = -1`. 
If we don't zero out the derivative parameter when passing to the `.execute` method, the expectation is that it will accumulate the result, i.e. `_d_param += dfdx`. Doing so for the above functions results in different outputs.
Code for reproducing the issue:
```cpp
double f2(double x) {
  x = -x;
  return x;
}

int main() {
  double x = 1.0;
  double dx = 0;

  // test with f1
  auto f1_dx = clad::gradient(f1);
  f1_dx.execute(x, &dx); // test with zeroed dx
  show(dx);

  f1_dx.execute(x, &dx); // test with non-zeroed dx
  show(dx);

  // test with f2
  dx = 0;
  auto f2_dx = clad::gradient(f2);
  f2_dx.execute(x, &dx); // test with zeroed dx
  show(dx);

  f2_dx.execute(x, &dx); // test with non-zeroed dx
  show(dx);
  return 0;
}
```

Outputs:
```
dx = -1
dx = -2
dx = -1
dx = 0
```

vgvassilev:
That is a feature, not a bug because often (in case of ROOT) we were told that is is with accumulating especially for vector values functions…

I agree it looks like a bug but I am not sure how we can improve the situation. Maybe clarify in the documentation?

vaithak:
> That is a feature, not a bug because often (in case of ROOT) we were told that is is with accumulating especially for vector values functions…

I don't mind the accumulation, but the accumulation seems to be happening incorrectly in the second case.

vaithak:
Python frameworks like Pytorch accumulate too, until you call `.zero_grad` method (or something similar), but the accumulation result seems to be incorrect here.

vgvassilev:
Ok, sorry I missed that part!

PR #2095, opened 2026-09-16
Fix #958: Prevent user variables starting with '_' from being renamed

### LINKED ISSUE #958: Unnecessary identifier in VarDecl that starts with "_"

ravindra-RKB:
## Description
This pull request resolves #958.

Previously, any user-defined variable starting with an underscore (e.g., `__b`) was incorrectly treated as an internal intermediate variable (like `_t0`) because of the broad `countedName` check in `VisitorBase::CreateUniqueIdentifier`. This resulted in Clad automatically renaming valid user variables to `__b0`, which caused confusion and cluttered the generated code.

This patch updates the `countedName` heuristic in `lib/Differentiator/VisitorBase.cpp` to specifically match known internal prefixes used by Clad (such as `_t`, `_r`, `_acc`, `_dummy`, `_temp`, `_chk`, `_cond`), instead of blindly targeting any name starting with `_`. 

## Related Issue
Fixes #958 

## Proposed Changes
- Narrowed the `countedName` condition in `VisitorBase::CreateUniqueIdentifier` to explicitly check for internal Clad prefixes.
- Added a regression test `test/Regressions/issue-958.cpp` using `FileCheck` to ensure user variables starting with `_` retain their exact original names during differentiation.

## Verification
- [x] Added `lit` and `FileCheck` tests to verify the generated AST code.
- [x] Verified that this change aligns with the expected behavior described in the original issue.


lib/Differentiator/VisitorBase.cpp
```diff
@@ -607,8 +607,11 @@ namespace clad {
                        !nameBase.starts_with("__end") &&
                        !nameBase.starts_with("__begin");
     bool countedName =
-        nameBase.starts_with("_") && !nameBase.starts_with("_d_") &&
-        !nameBase.starts_with("_delta_") && isRangedVar && nameBase != "_this";
+        (nameBase.starts_with("_t") || nameBase.starts_with("_r") ||
+         nameBase.starts_with("_acc") || nameBase.starts_with("_chk") ||
+         nameBase.starts_with("_dummy") || nameBase.starts_with("_temp") ||
+         nameBase.starts_with("_cond")) &&
+        !nameBase.starts_with("_this") && isRangedVar;
```
vgvassilev: The list leaves out names clad does generate, and they stop being numbered as a result. `_grad` (ReverseModeVisitor.cpp:3132), `_state` (:2343), `_numRevIterations` (:5347) and `_<user variable>_size` (:4205) all match none of these seven prefixes, so each comes out bare instead of `_grad0`, `_state0` and so on. Five tests check for the numbered form and would fail: test/Gradient/NumDiff.C on `_grad0`, test/Gradient/PullbackState.C, test/Gradient/PullbackStateByRef.C and test/CUDA/ThrustSortByKey.cu on `_state0`, and test/Gradient/Loops.C on `_numRevIterations0` at three places. Adding four more entries would turn those green, but it is the same trade the previous code made in the other direction -- the list has to be complete and nothing makes it stay complete.

test/Regressions/issue-958.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+// CHECK: double fn_darg0(double x) {
+// CHECK-NEXT:    double _d_x = 1;
+// CHECK-NEXT:    double _d___b = _d_x;
+// CHECK-NEXT:    double __b = x;
+// CHECK-NEXT:    return _d___b;
+// CHECK-NEXT: }
+
+int main() {
+    auto df = clad::differentiate(fn, "x");
```
vgvassilev, line 27: `CreateUniqueIdentifier` is shared by every differentiation mode, and only forward mode is exercised here. Reverse mode renames the primal through `BuildGlobalVarDecl` at ReverseModeVisitor.cpp:3982, so `clad::gradient(fn, "x")` on this same function is a second case the fix has to carry, and nothing checks it. One more call and one more CHECK block covers it.

vgvassilev (commented):
This is not a second attempt at the #958 fix in #2096, it is the same one. f74fe7d4 here and b418c9a5 there are byte-identical patches, and so are 34ae9336 and df86acb2; #2096 is this pull request plus one further commit for #888. Only one of the two should carry #958, and keeping it here and rebasing #2096 down to its #888 commit is the cleaner split, because the two issues are unrelated and a bisect wants them apart. Either way the defect below is in both.

The larger question is whether a list of prefixes is the right shape for this fix. A name string cannot tell clad's own temporary from a variable the user wrote, but the call site always knows which it is holding. `BuildVarDecl(Type, "_t", ...)` passes a literal clad chose; `BaseForwardModeVisitor::DifferentiateVarDecl` (BaseForwardModeVisitor.cpp:1762) and `ReverseModeVisitor` (ReverseModeVisitor.cpp:3982, through `BuildGlobalVarDecl`) pass `VD->getNameAsString()`, which is whatever the user typed, and ReverseModeVisitorOpenMP.cpp:212 and JacobianModeVisitor.cpp:94 do the same with a loop variable and a parameter. Move the distinction to the caller -- a parameter on `CreateUniqueIdentifier`, or a second entry point for user-written names -- and there is no list to write, nothing to keep in sync as clad grows new internal names, and the cases this patch still gets wrong go away with it: a user who writes `double _temp = x;` or `_cond`, `_acc`, `_chk`, `_dummy`, or anything at all starting `_t` or `_r`, still has the variable renamed.

The new test does observe the change, which is worth saying since it often is not true. Reverted, `CreateUniqueIdentifier("__b")` numbers the user's own variable and the derivative reads `double __b0 = x;`, so the `CHECK-NEXT` on line 14 fails.

I read the change rather than building it, and CI has not reported on this head yet, so the broken tests below come from reading the CHECK lines, not from a run.

PR #2096, opened 2026-09-16
Fix #888: Avoid overwriting user adjoints for pass-by-value parameters in reverse mode

### LINKED ISSUE #888: Incorrect gradient computation for non-zeroed derivatives

ravindra-RKB:
Fixes #888

### Problem
In reverse mode, when a primal function reassigns a pass-by-value parameter (e.g., `x = -x`), the generated gradient function was incorrectly zeroing out the caller's accumulated adjoint pointer. 

Because `x` was mapped directly to `*_d_x`, the assignment visitor would emit `*_d_x = 0;` to reset the adjoint after moving the derivative. This destroyed the initial adjoint state provided by the user in `&dx`.

### Solution
For pass-by-value parameters, we should not directly alias the caller's adjoint variable during the internal reverse sweep. 
=>The generated gradient now declares a local variable to safely accumulate the internal adjoint (e.g., `double _local_d_x = 0;`).
=>All internal assignments and uses of the parameter modify this local adjoint.
=> At the very end of the reverse sweep, we accumulate this local result back into the caller's pointer: `if (_d_x) *_d_x += _local_d_x;`.

This correctly preserves the chain-rule state (e.g. allowing `f1_dx.execute(x, &dx)` to accumulate properly over multiple calls or with non-zero initial values) without affecting pass-by-reference parameters which rightfully alias the caller's memory.

A lit test (`test/Regressions/issue-888.cpp`) has been added to verify the generated gradient code structure.


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -506,6 +506,47 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
     }
 
+    llvm::SmallVector<Stmt*, 4> paramAccumulations;
+    if (m_DiffReq.Mode == DiffMode::reverse && !m_ExternalSource) {
+      for (const DiffInputVarInfo& VarInfo : m_DiffReq.DVI) {
+        const ParmVarDecl* oPVD = VarInfo.param;
+        if (!utils::isArrayOrPointerType(oPVD->getType()) &&
+            !oPVD->getType()->isReferenceType()) {
+          ParmVarDecl* PVD = nullptr;
+          for (auto* p : params) {
```
vgvassilev: `params` is a local of `ReverseModeVisitor::Derive`, which closes about ninety lines above this one, and there is no member of that name. It is not in scope here, so this file does not compile as written. The loop a few dozen lines up does the same name lookup over `m_Derivative->parameters()`.

```suggestion
          for (auto* p : m_Derivative->parameters()) {
```

lib/Differentiator/VisitorBase.cpp
```diff
@@ -607,8 +607,11 @@ namespace clad {
                        !nameBase.starts_with("__end") &&
                        !nameBase.starts_with("__begin");
     bool countedName =
-        nameBase.starts_with("_") && !nameBase.starts_with("_d_") &&
-        !nameBase.starts_with("_delta_") && isRangedVar && nameBase != "_this";
+        (nameBase.starts_with("_t") || nameBase.starts_with("_r") ||
+         nameBase.starts_with("_acc") || nameBase.starts_with("_chk") ||
+         nameBase.starts_with("_dummy") || nameBase.starts_with("_temp") ||
+         nameBase.starts_with("_cond")) &&
+        !nameBase.starts_with("_this") && isRangedVar;
```
vgvassilev: This turns a deny list into an allow list, and every generated name outside those seven prefixes loses the counter that kept it unique. Two that clad produces today are `_grad`, the array for a numerical-diff gradient, and `_state`, the pullback state object. Neither starts with `_t`, `_r`, `_acc`, `_chk`, `_dummy`, `_temp` or `_cond`. `test/Gradient/NumDiff.C` expects `double _grad0[2]`, and `test/Gradient/PullbackState.C` and `test/Gradient/PullbackStateByRef.C` expect `_state0`, so those three should fail on this commit.

The loop underneath does not cover for it. It only disambiguates against names `LookupName` can see in the current scope, and clad later hoists declarations to function scope, so two `_grad` born in sibling scopes can still land as two declarations of one name in the same generated function. That is a redefinition error in code the user cannot see.

For #958 the property you want is that the name came from the user's source, not that it starts with one of a list of prefixes. Deciding from the decl being named rather than from its spelling also means nobody has to remember to extend the list the next time someone adds a prefix.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -506,6 +506,47 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+
+          m_Variables[PVD] = {VDDerived, AdjointInfo::Plain};
+
+          Expr* dPVD_ref = BuildDeclRef(dPVD);
+          Expr* local_ref = BuildDeclRef(VDDerived);
+          Expr* deref_dPVD = BuildOp(UO_Deref, dPVD_ref);
+          Expr* add_assign = BuildOp(BO_AddAssign, deref_dPVD, local_ref);
+          Stmt* if_stmt = clad_compat::IfStmt_Create(
+              m_Context, noLoc, false, nullptr, nullptr, dPVD_ref, noLoc, noLoc,
+              add_assign, noLoc, nullptr);
```
vgvassilev: `dPVD_ref` ends up in two places in the same tree: as the condition of the `if`, and under the `*` in `*_d_x += _local_d_x`. One `Expr` with two parents is what `findSharedNode` in `lib/Differentiator/ASTIntegrity.h` is there to catch, and clang's traversals assume a node has a single parent. Build the second reference separately.

```suggestion
          Expr* local_ref = BuildDeclRef(VDDerived);
          Expr* deref_dPVD = BuildOp(UO_Deref, BuildDeclRef(dPVD));
          Expr* add_assign = BuildOp(BO_AddAssign, deref_dPVD, local_ref);
          Stmt* if_stmt = clad_compat::IfStmt_Create(
              m_Context, noLoc, false, nullptr, nullptr, BuildDeclRef(dPVD),
              noLoc, noLoc, add_assign, noLoc, nullptr);
```

test/Regressions/issue-888.cpp
```diff
@@ -0,0 +1,23 @@
+
+// RUN: %clang++ -fplugin=%clad_plugin_path -fplugin-arg-clad-print-derivatives %s -c -o /dev/null 2>&1 | FileCheck %s
```
vgvassilev, line 2: `%clang++` and `%clad_plugin_path` are not substitutions in `test/lit.cfg`; it defines `%cladclang`, `%cladnumdiffclang` and `%cladclang_cuda`. lit leaves an unrecognised `%name` in the command line, so the shell looks for a program called `%clang++` and the test fails before FileCheck sees anything. As it stands, nothing in this branch observes the #888 fix.

A check on the generated text is also the weaker test here. #888 is that a second `execute` into a non-zero `dx` gave 0 instead of -2, so add a second RUN that runs the binary and asserts those two numbers with `%filecheck_exec`. That fails if the fix is reverted, where a check on the generated spelling only fails if the spelling changes.

```suggestion
// RUN: %cladclang %s -I%S/../../include -oissue-888.out 2>&1 | %filecheck %s
```

vgvassilev (commented):
There are two independent fixes in this branch. b418c9a5 rewrites `CreateUniqueIdentifier` for #958, which is about user variables that start with an underscore being renamed. ec85866e changes how a by-value parameter's adjoint is built, for #888. Neither needs the other, they fail in different ways, and a gradient that goes wrong six months from now will bisect to whichever of the two happens to be the commit it lands on. Please send them as two pull requests, each with its own test. Everything below is about one commit or the other rather than about the pair.

Two things I could not check because I read the change rather than building it. The new loop in `DifferentiateWithClad` excludes arrays, pointers and references, so a by-value class parameter falls through into it, and `BuildParams` gives those `AdjointInfo::ParenDeref`, so clad does handle them. For one of those the epilogue would be `*_d_v += _local_d_v`, which needs `operator+=` on the user's type, and `std::vector<double>` has none. That is an error inside generated code, which is the worst kind to hand a user. Restricting the loop to scalar types, or a test with a by-value struct parameter, would settle it. I also could not run the suite, so where I say below that a test fails, that is from reading its expected output.

PR #2097, opened 2026-09-16
ci :- add testing mode to compile produced source code (#714)

### LINKED ISSUE #714: [ci] Compile the produced code as part of the test suite
Currently clad tests are organized in two ways. Clad can be attached to clang and it can produce an executable file which we run and/or it can produce source code which we verify with FileCheck.

The problem with verification statement by statement is that if we make a core change the changes in the output checks are painful.

To mitigate this problem we can implement a testing mode which pipes the produced source code to the compiler and if it compiles we can assume the test output is reasonable. We will still have a number of tests that make sure the code is still readable for humans but that should be more focused and won't need that much manual interventions.

Aaravanand00:
Fixes #714

### Summary
Implements a new testing mode in lit (`%cladcompile` and `%cladpipecompile`) that extracts and compiles the produced derived C++ source code using `clang++ -fsyntax-only`.

This fulfills the remaining testing mode requirement of #714 (following PR #875) by providing an automated way to verify that Clad's generated derivative code compiles cleanly without requiring line-by-line `FileCheck` assertions.

### Key Changes
1. **`test/lit.cfg`**: Added lit substitutions:
   - `%cladcompile`: Runs Clad in syntax-only compilation verification mode (`-fsyntax-only`).
   - `%cladpipecompile`: Runs `compile_produced_code.py` to extract dumped derived source code and pipe it directly to `clang++ -std=c++17 -fsyntax-only -xc++ -`.
2. **`test/compile_produced_code.py`**: Created a test pipeline helper script that extracts dumped derived C++ source code and verifies clean compilation.
3. **`test/ValidCodeGen/ValidCodeGen.C`**: Added `%cladcompile` test run line.

### Verification
- [x] Verified Python syntax for `test/compile_produced_code.py`.
- [x] Added `%cladcompile` test run mode in `test/ValidCodeGen/ValidCodeGen.C`.


Aaravanand00:
@vgvassilev ptal...

vgvassilev:
That is not the design we are looking for. Maybe pick a simpler issue to feed to ai.

Aaravanand00:
Hi @vgvassilev

Apologies for jumping straight to a PR without discussing the design first. I searched for an open discussion channel to check your preferred approach beforehand but couldn't find one so I put together this implementation as a starting proposal.

Just to clarify, I take code quality seriously I'm currently an LFX Mentee at OpenSSF (working on the gittuf project) and have over 80 merged PR in gittuf and across various open source orgs (including Open Robotics), primarily focused on testing infrastructure. All my work is manually written, tested and reviewed.

My initial thought with the Python helper script was to keep the test runner changes lightweight and isolated without modifying Clad's core C++ plugin driver flags directly. However, I completely understand if you'd prefer a native C++ compiler plugin flag / Lit format design instead.

Could you share the specific design or entry point you have in mind for this? I had become more happy to refactor and implement it according to your vision for Clad.....

PR #2098, opened 2026-09-16
[Fix] Reject structs with reference data members in Reverse Mode (#2082)

### LINKED ISSUE #2082: Reference data members give a gradient twice the correct value
A struct with a reference data member gets a gradient exactly twice the correct one, with no diagnostic.

```cpp
#include <cstdio>

struct Ref {
  double& a;
  Ref(double& x) : a(x) {}
};

double f(double x) {
  Ref r(x);
  return r.a * 2.0;
}

int main() {
  auto g = clad::gradient(f);
  double dx = 0;
  g.execute(3.0, &dx);
  printf("dx=%g (expected 2)\n", dx);   // prints dx=4
}
```

With two members it doubles both:

```cpp
struct R {
  double &a, &b;
  R(double& x, double& y) : a(x), b(y) {}
};
double f(double x, double y) { R r(x, y); return r.a * r.b; }
// g.execute(3.0, 4.0, &dx, &dy)  ->  dx=8 dy=6, expected dx=4 dy=3
```

The factor of two looks like the adjoint arriving twice: once through the member access and once through the constructor pullback, since the reference member and the parameter it binds to are the same object and each path adds into it.

There is a second, separate symptom for the same feature. Without a user-defined constructor the type is an aggregate, and then clad generates code that default-constructs it:

```cpp
struct R { double &a, &b; };
double f(double x, double y) { R r{x, y}; return r.a * r.b; }
```

```
<clad generated code>:98305:1: error: call to implicitly-deleted default constructor of 'R'
```

A type with reference members has no default constructor, so the adjoint object clad creates for it is ill-formed. The error points into generated code, which gives a user nothing to act on.

Either case would be fine if clad rejected reference data members with a diagnostic naming the member. The silent factor of two is the part worth fixing first: it is a wrong answer that looks like a right one.

`docs/userDocs/source/user/UsingClad.rst` lists reference members among the unsupported class-type features, but nothing enforces that, so the documentation is the only thing standing between a user and a wrong gradient.

Environment: clad 014586bd, clang 22.1.8 (Homebrew), macOS x86_64.

Kuba70:
### Summary
This PR addresses **#2082** by implementing a guard check in `ReverseModeVisitor::VisitDeclStmt` to reject C++ structures containing reference data members (`isReferenceType()`) with a proper diagnostic error.

### Problem
Previously, Clad did not validate whether a `struct` or `class` contained reference data members before differentiating local variables in Reverse Mode. This caused two symptoms:
1. **Double-counting of gradients:** The adjoint was accumulated twice (once through member access and once through constructor pullback), leading to silent incorrect calculations.
2. **Compilation failure in generated code:** For aggregate types, Clad attempted to call an implicitly deleted default constructor for the adjoint object, causing cascade errors pointing into Clad's generated code.

### Solution
- Added a validation loop in `ReverseModeVisitor::VisitDeclStmt` right before `DifferentiateVarDecl` is invoked.
- Iterates through fields of `CXXRecordDecl` and checks if any field `isReferenceType()`.
- Emits a diagnostic error `reference data members are not supported` at the field location and returns an empty `StmtDiff()`, stopping further invalid code generation.
- Added a regression test in `test/Regressions/issue-2082.cpp`.

### Test Plan
Verified using `lit`:
```bash
make clad -j$(nproc)
lit -v test/Regressions/issue-2082.cpp
```

Closes #2082


lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4114,10 +4114,25 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
     // double _d_y = _d_x; double y = x;
+
+
     for (auto* D : DS->decls()) {
       if (auto* VD = dyn_cast<VarDecl>(D)) {
-        DeclDiff<VarDecl> VDDiff;
 
+        // Reject structures containing reference data members to prevent
+        // incorrect gradient generation / double-counting errors (Issue #2082).
+        if (const auto* RD = VD->getType()->getAsCXXRecordDecl()) {
```
vgvassilev, line 4127: This matches more than the structs a user writes. A lambda's closure type is a `CXXRecordDecl` too, and a capture by reference is stored as a reference-typed field, so `auto g = [&x](double t) { return x * t; };` inside a differentiated function becomes a hard error with this change. clad supports that today - `VisitorBase::buildClonedLambda` handles `LCK_ByRef` on purpose - and every lambda in `test/Gradient/Lambdas.C` is capture-free, so no test in the tree would catch the regression.

Skipping lambdas is the minimum. The wider point is that the predicate asks whether the type has a reference field, when the question is whether clad will build an adjoint for it that double-counts. A struct whose reference member is never touched on the differentiated path is rejected too.

```suggestion
        if (const auto* RD = VD->getType()->getAsCXXRecordDecl();
            RD && !RD->isLambda()) {
```

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4114,10 +4114,25 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
...
+          for (const clang::FieldDecl* FD : RD->fields()) {
+            if (FD->getType()->isReferenceType()) {
+
+             diag(DiagnosticsEngine::Error, FD->getLocation(), "reference data members are not supported");
```
vgvassilev: The message names nothing and the location is the field inside the struct, which can be in a header the user does not own. What they see is `reference data members are not supported` pointing at a line of somebody else's code, with no mention of the variable or the gradient that triggered it. #2082 asks for a diagnostic naming the member.

Report the error at `VD->getLocation()` and pass the variable and its type as arguments, then add a `DiagnosticsEngine::Note` at `FD->getLocation()` for the member itself. That is the shape the rest of this file uses: the `dependent non-const pointer and array parameters` error earlier in the file passes its subject with `<< param`.

test/Regressions/issue-2082.cpp
```diff
@@ -0,0 +1,16 @@
+// RUN: %cladclang -std=c++17 -fsyntax-only -I%S/../../include -Xclang -verify %s
+
+#include "clad/Differentiator/Differentiator.h"
+
+struct Ref {
+  double& a; // expected-error {{reference data members are not supported}}
+};
+
+double f(double x) {
+  Ref r{x};
```
vgvassilev, line 10: This is the aggregate case, which already fails today with `call to implicitly-deleted default constructor of 'Ref'`. The case #2082 calls the one worth fixing first is the other one: give `Ref` a constructor (`Ref(double& x) : a(x) {}`) and build it with `Ref r(x)`. That compiles and silently returns 4 where 2 is right.

The check keys on the type rather than on how the variable is initialised, so it should cover that too, but nothing here shows it does, and that is the variant whose old behaviour was a wrong answer rather than an error. Worth adding as a second struct in this file.

The test as written does observe the change: revert the guard and `-verify` fails, because the expected error is gone and clang's deleted-constructor errors are unexpected.

vgvassilev (commented):
The guard only runs on local variable declarations in reverse mode. A parameter of the same struct type, a member of it, and the same code under `clad::differentiate` all take other paths and keep returning the doubled gradient with no diagnostic. The pull request says it closes #2082, and #2082 is about a wrong number rather than about one spelling of the declaration, so either the check needs to cover those too or the pull request should say plainly that only the local-declaration case is handled.

Rejecting is a user-visible capability regression, and the commit message should say why rejecting is preferred to differentiating these types correctly. The issue does ask for exactly this, so the direction is fine, but a user who hits the new error will go looking for that reasoning and the commit is where they will look. `docs/userDocs/source/user/UsingClad.rst:546` already says a class cannot have pointer or reference data members, so this enforces one half of a rule that was already written down; a line in the docs pointing at the new error would finish the job.

I read the change and did not build it, so I could not confirm that the new error fires exactly once per differentiated function. That matters here because the new test uses `-Xclang -verify`, which fails if the same error is reported twice.

One mechanical thing that will turn CI red: the `diag` call is indented one space short of its block and runs well past 80 columns, and the hunk adds two blank lines and a `// ----` banner. `git-clang-format HEAD~` fixes the formatting.

Kuba70:
Hi @vgvassilev  I made a few changes based on your feedback:

I added a FindReferenceDataMember() helper that recursively checks a CXXRecordDecl for reference data members, including nested records (e.g. A -> B -> Ref). It skips lambda closure types with RD->isLambda().
I changed BuildParams() from void to bool. I didn't find another suitable path for validating function parameters before they are processed, so I use the return value to safely propagate the failure back to Derive(). If an unsupported reference data member is found, BuildParams() returns false and Derive() returns an empty DerivativeAndOverload instead of continuing with partially constructed parameters.
I improved the diagnostics so the error points to the parameter/variable and the note points directly to the offending reference member, e.g.:
error: parameter 'r' has unsupported reference data member
note: reference member 'a' declared here
I tested the following cases:
a reference data member in a local variable,
a Ref struct passed as a parameter to clad::gradient,
multiple reference data members (double& a, b),
nested reference data members (A -> B -> Ref),
constructor initialization,
aggregate initialization,
and lambda capture by reference.

For the invalid reference-member cases, compilation now terminates safely with the diagnostic instead of continuing into invalid derivative generation / crashing.

I also investigated the lambda case. A lambda with a reference capture currently crashes Clad with the UseBeforeDecl assertion in DerivativeBuilder.cpp. I tested the exact same case both with my changes and on the original origin/master, and it crashes in both versions. So it doesn't appear to be caused by my changes, but I don't know the root cause of that issue yet.

PR #2118, opened 2026-09-18
Ignore redundant #pragma clad OFF/DEFAULT

### LINKED ISSUE #2116: A second '#pragma clad OFF' asserts in CladPragmaHandler
Two `#pragma clad OFF` in a row abort a build with assertions:

```cpp
#pragma clad ON
#pragma clad OFF
#pragma clad OFF
int main() { return 0; }
```

```
$ clang++ -fsyntax-only -Xclang -add-plugin -Xclang clad \
    -Xclang -load -Xclang <path>/clad.dylib -I<clad>/include double-off.cpp
Assertion failed: (CladEnabledRange.back().getEnd().isInvalid()),
function HandlePragma, file ClangPlugin.cpp, line 120.
```

`CladPragmaHandler::HandlePragma` (tools/ClangPlugin.cpp:108-124) treats the
two directives differently. A second `ON` is ignored on purpose while a region
is open:

```cpp
// If a second ON is seen, ignore it if the interval is open.
if (CladEnabledRange.empty() || CladEnabledRange.back().getEnd().isValid())
  CladEnabledRange.push_back(R);
```

A second `OFF` asserts instead, and without assertions it moves the end of the
already-closed region to the later `OFF`, so the recorded region silently
grows over the text between the two directives.

Turning off what is already off is not an error the user can act on: the
region ends up closed either way, which is what was asked for. Ignoring the
second `OFF`, the way a second `ON` is ignored, would do.

Where this shows up: ROOT wraps every request as
`#pragma clad ON` / `clad::gradient(f);` / `#pragma clad OFF`, and includes
`Math/CladDerivator.h` followed by another `#pragma clad OFF`, so interpreted
code that turns clad off once more of its own accord aborts the interpreter.
I hit it writing a cling test, with the LLVM that has assertions on.

Reproduced at 48be98c7 against an assertions build of LLVM 21.

Elvand-Lie:
## Summary

A second `#pragma clad OFF` (or `DEFAULT`) while clad is already disabled asserted in debug builds and silently extended the closed interval otherwise. Match the existing second-`ON` ignore behavior: only close an open interval.

Fixes: #2116

## Test plan

- [x] Extended `test/Misc/Pragma.C` with redundant OFF/DEFAULT/ON sequences
- [ ] CI `check-clad` / Misc.Pragma


test/Misc/Pragma.C
```diff
@@ -1,9 +1,19 @@
 // RUN: %cladclang -fsyntax-only -Xclang -verify %s
 
-#pragma clad DEFAULT // FIXME: Moving DEFAULT after ON/OFF makes the test crash.
+#pragma clad DEFAULT
 #pragma clad ON
 #pragma clad OFF
 
+// A second OFF/DEFAULT while clad is already off is ignored: it must neither
+// assert nor extend the already closed interval.
+#pragma clad OFF
```
vgvassilev, line 9: This shows that a second OFF no longer asserts, which is the only half of the change a test can see, and it sees it only in a build with assertions. CI sets BUILD_TYPE=Release everywhere except the debug_build rows, so on most rows this file passes with the fix reverted. Nothing to change here -- see the body for why the other half is not writable yet.

vgvassilev (approved):
One thing worth knowing while this is in your hands: the recorded range hardly reaches behaviour at all. `DiffCollector::isInInterval`, lib/Differentiator/DiffPlanner.cpp:433, is

```cpp
if (E.isValid() && clad_compat::SourceManager_isPointWithin(SM, Loc, B, E))
  return true;
else if (SM.isBeforeInTranslationUnit(B, Loc))
  return true;
```

and the second branch fires for a closed interval too, so any location after the first ON counts as enabled and a later OFF does not stop a request from being collected. Differentiator.h ends with `#pragma clad ON` at line 1117, so every unit that includes it has such a first interval. It reads as if the `else` was meant to be the open-interval case, `E.isInvalid() && ...`. Two consequences for this change. The half of it that is not the assert -- the end of a closed region no longer moving to a later OFF -- has no effect anyone can observe, so the test below can only show that nothing asserts. And the FIXME at the bottom of Pragma.C, enumerate decls and clad:: calls between on/off/default regions, cannot be discharged until that is settled. Worth its own issue rather than this one; it is also what would make ROOT's wrapping of each request in ON/OFF mean what it looks like it means.

The change itself is right. The assert cannot fire again, because its condition is now the guard on the one statement it protected, and an OFF with no open region is still ignored rather than taken for something else, which is what the issue asked for. Removing the FIXME from the first line is fair -- the sequence it warned about is in the file now.

I read the change rather than building it.

PR #2119, opened 2026-09-18
Clone unsupported stmts except StmtClone-unsafe kinds

### LINKED ISSUE #2088: Unsupported statements crash instead of being diagnosed

Elvand-Lie:
## Summary

`VisitStmt` must keep cloning unsupported-but-clonable nodes (for example `CXXTryStmt`) so existing derivatives that preserve try/catch keep working.

Issue #2088 crashes only for kinds `StmtClone` cannot implement. This PR restores `Clone(S)` as the default after `diagUnsupported`, and drops only:

- `BinaryConditionalOperator` (GNU `x ?: y`)
- `AttributedStmt` (`[[likely]]`, etc.)
- `CXXInheritedCtorInitExpr`
- `CXXParenListInitExpr` (Clang 16+, parenthesized aggregate init)

The deny list lives in one shared helper used by forward and reverse visitors.

Fixes: #2088

## Test plan

- [x] `test/Regressions/issue-2088.cpp` — `?:` forward+reverse, `[[likely]]` forward+reverse, inherited ctor reverse
- [x] `test/Regressions/issue-2088-paren-init.cpp` — paren aggregate init forward+reverse (Clang 16+)
- [ ] CI: `FirstDerivative/DiffInterface.C`, `ForwardMode/VectorModeInterface.C` stay green (try/catch still cloned)


lib/Differentiator/VisitorBase.cpp
```diff
@@ -680,6 +681,25 @@ namespace clad {
...
   }
 
+  bool VisitorBase::isKnownStmtCloneUnsupported(const clang::Stmt* S) {
+    // These are the four kinds from issue #2088 that StmtClone has no case
+    // for. Cloning one from the VisitStmt fallback asserts in
+    // StmtClone::VisitStmt or, with asserts off, yields a malformed node that
+    // ReferencesUpdater crashes on, so the callers diagnose and drop them
+    // instead. This is a fixed deny list, not a general StmtClone support
+    // query.
+    if (isa<BinaryConditionalOperator>(S) || isa<AttributedStmt>(S) ||
```
vgvassilev, line 691: `AttributedStmt` is not a construct clad cannot differentiate. `[[likely]]` and `[[unlikely]]` carry no runtime meaning, so the statement under the attribute is ordinary code the visitors already handle. Listing the kind here throws that statement away: in `f_likely` in the new test the `AttributedStmt` is the whole then-branch, so `r *= 2` is gone from the generated derivative and the function answers 1 where the derivative is 2, with only a warning to say so. Visiting the sub-statement instead would differentiate it correctly and would not need a clone at all.

test/Regressions/issue-2088.cpp
```diff
@@ -0,0 +1,68 @@
...
+  return x ?: 1.0; // expected-warning {{statement kind 'BinaryConditionalOperator' is not supported}}
+}
+
+// `[[likely]]` on a statement wraps that statement in an AttributedStmt, which
+// StmtClone has no case for. Covered in both directions below.
+// expected-warning@* 1+ {{statement kind 'AttributedStmt' is not supported}}
+double f_likely(double x) {
+  double r = x;
+  if (r > 0) [[likely]] {
+    r *= 2;
```
vgvassilev, line 38: Nothing in this file observes what the derivative now contains -- it passes as long as the compile finishes and the warning appears. With the `AttributedStmt` dropped, this statement is no longer in `f_likely`'s derivative, so the generated function answers 1 instead of 2 and the test is still happy. `%cladclang` already passes `-fdump-derived-fn`, so piping the run through `%filecheck` and adding CHECK lines, the way DiffInterface.C does for the try block, would put the new behaviour in front of the reader instead of leaving it to be found downstream.

test/Regressions/issue-2088.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+double f_inherited(double x) {
+  Derived d(x);
+  return d.b;
+}
+
+int main() {
+  clad::gradient(fwd, "x");
+  clad::differentiate(fwd, "x");
+  clad::gradient(rev, "x");
```
vgvassilev, line 64: `fwd` is asked for in both modes a few lines up, `rev` only in reverse. Was `clad::differentiate(rev, "x")` left out because it does not compile? An unsupported expression in a return position leaves forward mode with no tangent to return, and clad emits C++ that the user's compiler then has to accept, so a `double` function with no return statement is a build failure on code they cannot see. Either cover it here or say in the description that it is still open.

vgvassilev (commented):
The deny list guards only the node the fallback happens to land on, and `Clone` is a deep clone, so the same crash sits one level below it. Forward mode has no `VisitCXXTryStmt`, so a `try` block reaches this very fallback and is cloned whole -- `test/FirstDerivative/DiffInterface.C` pins that, with its `statement kind 'CXXTryStmt' is not supported` warning and the CHECK lines showing the try block reappearing in the derivative. Put one of the four kinds inside such a block and nothing has changed:

```cpp
double f(double x) {
  try { double y = x ?: 1.0; return y; } catch (...) { return 0; }
}
int main() { clad::differentiate(f, "x"); }
```

`isKnownStmtCloneUnsupported` says no for the `CXXTryStmt`, `Clone` descends into its body, and `StmtClone::VisitStmt` asserts on the `BinaryConditionalOperator` exactly as in the issue. The list also covers only the four kinds the issue happened to report. A C++20 rewritten comparison -- `a < b` built from `operator<=>` -- is a `CXXRewrittenBinaryOperator`, which has no handler in either visitor and no `StmtClone` case, so it crashes the same way and is not on the list.

On the two approaches, since #2094 fixes the same issue a different way: #2094 gives `StmtClone` a case for each of the four kinds, and that is the sounder of the two. The null comes out of `StmtClone::VisitStmt`, so a case there fixes every position the kind can appear in, the nested one above included, and it keeps the behaviour the issue calls intended -- diagnose, then carry the statement over unchanged. That is also why `throw`, `goto` and labels are safe today: `StmtClone` can reproduce them, not because anything drops them. So "drop those the way throw, goto and labels finish" in the new comments describes the opposite of what those kinds do. Dropping is a different contract, and DiffInterface.C's CHECK lines are an existing test that asks for the old one.

#2094 gets one thing wrong that this change gets right. It guards `CXXParenListInitExpr` with `CLANG_VERSION_MAJOR > 16`, but the node landed in LLVM 16 (llvm/llvm-project 40c52159d3ee, first tag llvmorg-16.0.0), and clad's CI has a `clang-runtime: '16'` row, so clang 16 would keep crashing there with its test marked UNSUPPORTED. The `>= 16` here is the right bound. Neither change fixes the general case: `StmtClone::VisitStmt` still asserts, and still returns null under NDEBUG, for every kind nobody has written a case for. Letting the caller see that a clone failed, and deciding there, would fix the class rather than four instances of it.

I read the change rather than building it, and I took the `-verify` counts in the new tests from CI being green rather than working them out myself.

Issue #731, opened 2024-01-21
Provide an alternative to clad::differentiate


vgvassilev:
This would allow us to avoid inclusion of Differentiator.h. We have several approaches: a compiler flag, a pragma directive or maybe even an APINote.

Might simplify the work required in #730

vgvassilev:
@Errant404, as we discussed during the proposal writing period, that'd be a good warmup task. Consider,

```c
#include <stdio.h>
// Can't include:
//#include <clad/Differentiator/Differentiator.h>

double add(double i){
    return i + i;
}

double fn(double i, double start){
    double k = start, sum;
    for (int a=0; a < 3; a++) {
        k = start;
        sum = add(k);
    }
    return i;
}

#pragma clad gradient fn //
void fn_grad(double i, double start, double *d_i);

int main() {
    double i = 1;
    double d_i = 0;
    fn_grad(i, 0, &d_i);
    //clad::gradient(fn, "i"); // Can't call it as that's not c

    return 0;
}
```

I think we should implement the `#pragma clad gradient` infrastructure rather easily in the pragma we will need anyway for openmp.

cc: @mvassilev

Aaravanand00:
Hi @vgvassilev,

I've submitted a PR for #731. To allow differentiation without needing `#include <clad/Differentiator/Differentiator.h>`, I added support for `#pragma clad gradient <fn>` and `#pragma clad differentiate <fn>` in `ClangPlugin.cpp`. 

It looks up the target function in the AST scope during translation unit processing and generates the derivatives. I've also added tests to verify it works as expected.

ptal and if any changes needed let me know...

PR #2121, opened 2026-09-18
[Frontend] Provide #pragma clad gradient and differentiate directives as alternative entry points (#731)

Aaravanand00:
### Summary
This PR addresses issue #731 by introducing `#pragma clad gradient <fn_name>` and `#pragma clad differentiate <fn_name>` as alternative entry points for automatic differentiation in Clad.

This allows differentiating C and C++ functions without needing to include `<clad/Differentiator/Differentiator.h>` or call template APIs like `clad::gradient(...)` / `clad::differentiate(...)`.

### Key Changes
- **Clang Frontend Plugin (`tools/ClangPlugin.cpp`)**:
  - Extended `CladPragmaHandler` to parse `#pragma clad gradient <fn>` and `#pragma clad differentiate <fn>` (supporting both bare `fn` and parenthesized `(fn)` forms).
  - Added token validation and clear diagnostics for invalid syntax or missing arguments.
  - Extended `CladPlugin::HandleTranslationUnit` to perform Sema/AST scope lookups for target functions, construct `DiffRequest` nodes, and trigger derivative generation.
- **Tests**:
  - Added lit test `test/Pragmas/PragmaGradient.cpp` verifying headerless differentiation.
  - Updated pragma error expectations in `test/Misc/Pragma.C`.

### Verification
- Tested pragma parsing and AST derivative emission.
- Verified headerless execution of generated derivative functions (`add_grad`, `add_darg0`).


test/Pragmas/PragmaGradient.cpp
```diff
@@ -0,0 +1,27 @@
+// RUN: %cladclang -c %s -o %t.o
+// RUN: %clang %t.o -o %t.exe
+// RUN: %t.exe | %FileCheck %s
```
vgvassilev: These RUN lines do not run. clad's test/lit.cfg defines neither %clang nor %FileCheck. It does rewrite the bare words `clang` and `FileCheck` wherever they appear, including inside `%clang` and `%FileCheck`, so the commands come out as `%/path/to/clang ...` and `%/path/to/FileCheck ...`, each with a stray leading `%` and neither an executable. The substitutions this project has are %cladclang for compiling, and %filecheck / %filecheck_exec for checking.

```suggestion
// RUN: %cladclang %s -o %t
// RUN: %t | %filecheck_exec %s
```

%filecheck_exec looks for CHECK-EXEC, so the `// CHECK: grad = 2.0, df = 2.0` line further down has to become `// CHECK-EXEC: grad = 2.0, df = 2.0` for this to check anything.

test/Pragmas/PragmaGradient.cpp
```diff
@@ -0,0 +1,27 @@
...
+double add(double i) {
+    return i + i;
+}
+
+#pragma clad gradient add
+#pragma clad differentiate add
+
+// Declare the signature of the generated derivatives
+extern "C" {
+    void add_grad(double i, double* _grad);
```
vgvassilev: clad generates add_grad as an ordinary C++ function in the global namespace, so its symbol is mangled. These declarations have C language linkage, so they name different symbols and the link step cannot resolve them. Dropping the extern "C" would fix the link.

The larger point is what #731 is asking for. A user who still has to hand-write the derivative's prototype has traded an include for a prototype that goes silently stale the first time the signature changes -- and getting it wrong is a link error with no explanation, which is worse than the include. The pragma earns its place when it puts the declaration in scope itself, so that `add_grad` is simply callable on the next line. That needs the declaration created when the pragma is seen rather than at the end of the translation unit, which is a different shape from what this patch does, so it is worth settling before the rest.

vgvassilev (commented):
The pragma requests are processed outside the scheduler's graph, so the feature can only work for a function that calls nothing. ProcessDiffRequest is called directly, with no getScheduler().getGraph().setCurrentProcessingNode(request) around it. When clad derives a function whose body contains a call, DerivativeBuilder::HandleNestedDiffRequest ends at AddEdgeToGraph (lib/Differentiator/DerivativeBuilder.cpp:489 and 967), which is DynamicGraph::addEdgeToCurrentNode; with no current node set it adds nothing, so the callee's pullback is never generated and the derivative body calls a function that does not exist. The loop in FinalizeTranslationUnit at tools/ClangPlugin.cpp:1010-1020 is the protocol for this. Add each pragma request to the graph as a source node and let that loop drive it, rather than calling ProcessDiffRequest yourself. The test differentiates add, which calls nothing, so it cannot see this.

The DiffRequest is missing nearly everything a clad::gradient request carries, and I do not think the generated function can have the name the test expects. request.Args is left null, so UpdateDiffParamsInfo returns at its `if (!diffArgs || !FD)` guard (lib/Differentiator/DiffPlanner.cpp:658) and DVI stays empty. The "differentiate with respect to every parameter" default is Case 3 of that same function and is reached only when Args is a CXXDefaultArgExpr, which a hand-built request does not have. ComputeDerivativeName then hits `if (DVI.empty()) return "<no independent variable specified>";` (DiffPlanner.cpp:1039), so the derivative is not named add_grad at all. BaseFunctionName, CallContext, and the TBR, varied, useful and loop analysis switches that ProcessInvocationArgs copies out of RequestOptions (DiffPlanner.cpp:1113-1146) are all unset as well, so -fno-tbr and the other analysis flags would not reach a pragma request and it would produce a different derivative from clad::gradient on the same function. Going through the same request-building code the clad:: calls use is what keeps the two from drifting.

The new `if (PragmaTok.isNot(tok::identifier))` guard is a real bug fix on its own -- a bare `#pragma clad` with nothing after it dereferences a null IdentifierInfo today -- but it is not part of this feature. It wants its own commit ahead of this one, with its own line in test/Misc/Pragma.C. Note also that #2118 is already changing this same handler and this same test file for #2116, so the two will conflict; landing the small fix separately keeps that easy to sort out.

One thing I could not check: CI has not run on 2178f4e -- the head commit has no check runs at all -- so none of this has been through the test suite.

Aaravanand00:
@vgvassilev changes was done ptall....

PR #2122, opened 2026-09-19
Fix #2116: Ignore duplicate `#pragma clad OFF` in CladPragmaHandler

### LINKED ISSUE #2116: A second '#pragma clad OFF' asserts in CladPragmaHandler

ravindra-RKB:
Fixes #2116

### Problem
When `#pragma clad OFF` is used twice in a row, the `CladPragmaHandler` would abort the build with an assertion failure (`assert(CladEnabledRange.back().getEnd().isInvalid())`). It attempted to close an already-closed clad region instead of gracefully ignoring the redundant directive, which is problematic since generated or wrapped code (like in ROOT) can legitimately emit an extra `OFF` pragma.

### Solution
Modified `CladPragmaHandler` to ensure that it only sets the end location if the current region's end is actually invalid (i.e. currently open). If it is already closed, the duplicate `#pragma clad OFF` is simply ignored (just like a duplicate `ON` is).

Added a regression test to verify that multiple OFF pragmas no longer crash the build.


Elvand-Lie:
Hi @ravindra-RKB I think I have an open work on this. Possible to not make it duplicate?

ravindra-RKB:
Hi @Elvand-Lie, sorry about that! I didn't realize there was already an open PR for this issue when I was working on it. I'll close this one to avoid any duplication. Great minds think alike! 😄

Issue #2099, opened 2026-09-17
Reverse mode drops the cast when building the adjoint of a pointer cast


vgvassilev:
Reverse mode builds the adjoint of a pointer cast as a `<recovery-expr>`, and the generated code does not compile. There is no diagnostic naming the construct.

```cpp
#include "clad/Differentiator/Differentiator.h"
struct B { double v; };
struct D : B {};
double f(double x) {
  D d; d.v = x * x;
  B* b = &d;
  D* p = static_cast<D*>(b);
  return p->v;
}
int main() { clad::gradient(f); }
```

```
<clad generated code>:98344:1: error: cannot initialize a variable of type 'D *' with an lvalue of type 'B *'
```

`-fdump-derived-fn` shows why. The primal cast is cloned correctly and the adjoint beside it is not:

```cpp
B *_d_b = &_d_d;
B *b = &d;
D *_d_p = <recovery-expr>(_d_b);      // adjoint: the cast was dropped
D *p = static_cast<D *>(b);           // primal: fine
```

`dynamic_cast` behaves the same way. Forward mode compiles and runs this example, so it is reverse mode only.

### Why there is no warning, which may be the more important half

`CXXStaticCastExpr` derives from `CXXNamedCastExpr`, which a visitor handles, so it never reaches `VisitStmt` and never reaches `diagUnsupported`. Clad's contract elsewhere is that it names a construct it cannot handle; that contract only covers nodes that fall all the way through to the catch-all.

Any node deriving from a handled base is absorbed by that base's handler with no signal at all, whether or not anybody considered it. `UserDefinedLiteral` is another: it derives from `CallExpr`, `VisitCallExpr` absorbs it, and differentiating `x * 2.0_k` happens to give the right answer -- but that is a coincidence nobody verified, not a decision.

Handling a family at its base is deliberate and correct where it was intended: `VisitCallExpr` `dyn_cast`s to `CXXMemberCallExpr` and `CXXOperatorCallExpr` in seven places, so it is written for them. The problem is that intentional base-handling and accidental absorption are indistinguishable, and the accidental kind is silent. A node kind a future Clang release adds under an existing base inherits that base's treatment on upgrade, with no diagnostic and no test failure.

[...]

PR #2123, opened 2026-09-19
Fix #2099: Cast the adjoint expression in ReverseModeVisitor

ravindra-RKB:
Resolves the code generation part of #2099.

### Problem
Reverse mode builds the adjoint of a pointer cast as a `<recovery-expr>` causing a compilation error.
While `VisitCStyleCastExpr` properly rebuilt the cast for the adjoint expression (`getExpr_dx()`), `VisitCXXNamedCastExpr` and `VisitCXXFunctionalCastExpr` updated the primal statement but entirely missed applying the cast to their adjoint expressions. This led to invalid types when dealing with adjoints of pointer casts.

### Solution
Modified `VisitCXXNamedCastExpr` and `VisitCXXFunctionalCastExpr` in `ReverseModeVisitor.cpp` to correctly rebuild and apply the cast to the adjoint expression (via `updateStmtDx`) using the identical `TypeSourceInfo` and `CastKind` from the primal cast.

Added a regression test to verify that `static_cast` works correctly in reverse mode without triggering a compilation error.


test/Regressions/issue-2099.cpp
```diff
@@ -0,0 +1,21 @@
+
+// RUN: %clang++ -fsyntax-only -Xclang -add-plugin -Xclang clad -Xclang -load -Xclang %clad_plugin_path %s 2>&1 | FileCheck %s
```
vgvassilev: This test never runs, so nothing here observes the fix. clad's test/lit.cfg defines neither %clang++ nor %clad_plugin_path. lit rewrites the bare word `clang` wherever it appears, including inside `%clang++`, so the command becomes `%/path/to/clang++ ...` with a stray leading `%` and is not an executable; `%clad_plugin_path` is passed through verbatim. There is also no -I for clad's include directory, so the #include would not resolve even if the command ran. Use the substitutions the other tests in this directory use: %cladclang already adds the plugin, -std=c++17 and -fdump-derived-fn, and %filecheck already implies --implicit-check-not "error:".

```suggestion
// RUN: %cladclang -I%S/../../include %s -o %t 2>&1 | %filecheck %s
```

With that line, the CHECK-NOT below is redundant, and FileCheck needs at least one directive to run. Replace it with a positive check on the generated adjoint -- something of the shape `// CHECK: D *_d_p = static_cast<D *>(_d_b);` -- so the test also fails when the adjoint compiles but is wrong. As written it could only ever catch a hard error.

lib/Differentiator/ReverseModeVisitor.cpp
```diff
@@ -4425,6 +4437,15 @@ Expr* ReverseModeVisitor::getStdInitListSizeExpr(const Expr* E) {
             .get();
     subExprDiff.updateStmt(castExpr);
 
+    Expr* castExprDx = subExprDiff.getExpr_dx();
```
vgvassilev: Casting the adjoint with the primal's written type is wrong whenever the two types differ, and clad makes them differ for const pointers. DifferentiateVarDecl builds the adjoint's type with utils::getNonConstType (line 3823), which strips the pointee const, so for

```cpp
double v = x * x;
double* p = &v;
const double* q = static_cast<const double*>(p);
```

the adjoint declaration is `double *_d_q` and this code hands it `static_cast<const double *>(_d_p)`, of type `const double *`. That example compiles today and would stop compiling. Building the cast against utils::getNonConstType of the written type, rather than reusing TSI, keeps the target in step with the adjoint's declared type.

Two smaller things fall out of the same line. When Sema refuses the cast, the error lands on the user's own static_cast with nothing saying clad generated the expression. And BuildCXXNamedCast returning an invalid result makes .get() null, so updateStmtDx silently clears the adjoint. The functional-cast hunk above has the same shape.

vgvassilev (commented):
The direction is right. VisitCStyleCastExpr already casts the adjoint, and these two visitors should match it.

There is a test in the tree that asserts this bug, and this change breaks it. test/Misc/Cling.cpp declares the same B, D and f on lines 12-15 -- it is the reproducer from #2099 -- and the CHECK block under f_req asserts `error: cannot initialize a variable of type 'D *' with an lvalue of type 'B *'`, which is exactly the error you are removing. The build-cling job in .github/workflows/ci.yml runs `lit ... obj/test/Misc/Cling.cpp`, so that row will fail. That test exists to show a diagnostic pointing into code clad generated, so it needs a construct clad still cannot handle rather than a deleted CHECK block. Please update it here, in the same commit.

Two things I could not check. I did not build, so the const case in the comment below is read off getNonConstType (lib/Differentiator/CladUtils.cpp:613) and the adjoint type at lib/Differentiator/ReverseModeVisitor.cpp:3823 rather than observed. And no CI has run on 27bae01 -- the head commit has no check runs at all -- so nothing has yet seen the two added lines that end in trailing whitespace, which the clang-format job rejects.

PR #2128, opened 2026-09-19
Use std::numeric_limits for NaN in builtin derivatives

Vedant2005goyal:
clad_digamma and clad_trigamma return NaN for non-positive integer arguments using the NAN macro from math.h. The availability of this macro is implementation-dependent.

Use std::numeric_limits<T>::quiet_NaN() instead to avoid relying on the NAN macro and ensure BuiltinDerivatives.h remains compatible with environments where NAN is not defined.

P.S. This also allows the test suite to compile and run on newer macOS versions whose SDKs do not provide NAN.


vgvassilev (approved):
LGTM!

PR #2129, opened 2026-09-20
Run useful analysis only on defined functions

Vedant2005goyal:
Currently, useful analysis runs on every request, including declared-only functions. This causes Clad to request a CFG for a function without a body, leading to a crash.

Only run the analysis when the function has a body else it is skipped.


vgvassilev (approved):
Lgtm!

lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1901,7 +1901,7 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
         runVariedAnalysis(request, AnalysisDC.get());
       }
 
-      if (m_TopMostReq->EnableUsefulAnalysis) {
+      if (m_TopMostReq->EnableUsefulAnalysis && request->isDefined()) {
```
vgvassilev: On a second thought: shouldn’t that check be part of the analysis itself?

vgvassilev (approved):
LGTM!

PR #2132, opened 2026-09-20
[docs] Repair and extend the internal doxygen output.

vgvassilev:
The internal docs site had no front page. `doxygen.intro` carries the `@mainpage` and is listed in INPUT, but `.intro` is not an extension doxygen recognises and EXTENSION_MAPPING was empty, so doxygen indexed the file and extracted nothing: the published page at /internalDocs/html/index.html has an empty `<div class="contents">`, and the developer guide links contributors straight at it. Mapping the extension makes doxygen read the block it was always meant to read.

The site also offered nothing a tool could read, and navigation through a menu bar and the browser back button. It now emits a tag file, so another doxygen project can cross-link into clad, and XML, which is what anything indexing an API reads; both land beside the HTML in the directory conf.py publishes. XML_PROGRAMLISTING stays off: it embeds a second copy of every source file and doubles the XML, while the per-entity records it does not touch already carry the line numbers of each body.

The rest is deletion. HTML_STYLESHEET and HTML_HEADER are empty, so `doxygen.css` and `doxygen.header` have never reached the site; `manpage.css` is for man output nothing generates and imports an `../llvm.css` that does not exist here; `docs/internalDocs/sphinx/` is a second Sphinx project no CMakeLists refers to. EXCLUDE named `lib/UserInterface/textinput`, a cling path clad does not have, and MATHJAX_RELPATH pointed at a host that stopped serving years ago, for formulas the tree does not contain.


vgvassilev:
clang-format can be ignored for this one.

PR #2138, opened 2026-09-20
Handle kernel launches during varied analysis

Vedant2005goyal:
Kernel launches were not being analyzed like regular function calls, so variation passed through pointer arguments could be missed. This caused values modified by a kernel launch to remain marked as non-varied. As a result, their adjoints and seeds could be dropped, producing an incorrect zero gradient.

This PR makes kernel launches go through the same call handling, so variations introduced through their arguments are propagated correctly.


lib/Differentiator/DiffPlanner.cpp
```diff
@@ -1599,6 +1599,14 @@ static QualType GetDerivedFunctionType(const CallExpr* CE) {
     return false;
   }
 
+  bool DiffCollector::TraverseCUDAKernelCallExpr(CUDAKernelCallExpr* KCE) {
+    WalkUpFromCUDAKernelCallExpr(KCE);
+    TraverseStmt(KCE->getCallee());
+    for (Expr* arg : KCE->arguments())
+      TraverseStmt(arg);
+    return true;
+  }
```
vgvassilev: This override has exactly one effect that the default does not, and nothing in the patch says what it is. RecursiveASTVisitor already defines TraverseCUDAKernelCallExpr with an empty body, so the default walks every child and still reaches DiffCollector::VisitCallExpr through WalkUpFromCUDAKernelCallExpr and WalkUpFromCallExpr. The one difference is that CallExpr::children() includes the pre-argument holding the <<<...>>> configuration and CallExpr::arguments() does not, so this version keeps the collector out of the configuration call and out of the dim3 temporaries it builds. That looks like the intent, and it agrees with the rest of clad, which clones the configuration rather than differentiating it (ReverseModeVisitor.cpp:2227). Please write that reason down, and say which request the old path was creating that this one stops.

Two smaller things in the same body. The results of WalkUpFromCUDAKernelCallExpr and of each TraverseStmt are dropped, so a nested traversal that asks to stop cannot; every other Traverse in this file returns what it got. And the declaration added in DiffPlanner.h has no comment, where the members either side of it do.

test/CUDA/ActivityAnalysis.cu
```diff
@@ -33,6 +33,24 @@ __global__ void func(double* out, double x) {
 // CHECK: val = val * 2.;
 // CHECK: }
 
+__global__ void square(double* out, const double* in) {
+    out[0] = in[0] * in[0];
+}
+
+double func1(double* in) {
+  double tmp[1] = {0};
+  square<<<1, 1>>>(tmp, in);
```
vgvassilev, line 42: Which half of the change does this test observe? The whole file runs with -enable-va, so the new CHECK lines exercise the ActivityAnalyzer override and the DiffPlanner one together. Please check whether the test still passes with only the DiffPlanner.cpp hunk reverted. If it does, that hunk has no regression test and wants one.

Separately, tmp is host memory being handed to a kernel, so this launch would fault if it ever ran. Nothing runs it today, since the test gradients func1 and never calls the result, but pinning a code path on a program that could not execute is fragile. Allocating tmp with cudaMalloc costs one line and makes it a program someone can also run.

vgvassilev (commented):
The two halves of this are not the same kind of change, and only one of them explains itself. The ActivityAnalyzer half is plainly right: VariedAnalyzer keeps its call handling in TraverseCallExpr, clang's default traversal of a kernel call never reaches that override, so until now a launch was analysed as if it had no arguments. The DiffPlanner half needs a sentence saying what it is for; see the comment on it.

One thing I could not check: every test under test/CUDA is UNSUPPORTED on this machine, so nothing here comes from a run. The claims below come from reading clang's RecursiveASTVisitor.h and Expr.h. For what it is worth, the new CHECK lines should run on a compile-only CUDA row, since the derivative is dumped by -fdump-derived-fn; only the CHECK-EXEC lines need a runtime.

PR #2153, opened 2026-09-23
Keep restore_tracker records in a single buffer

Vedant2005goyal:
Avoid the overhead of storing each scalar in a map node with a separate vector allocation. Store one record per address with all bytes in a single buffer, matching the existing CUDA implementation.


vgvassilev:
Can you elaborate about the motivation of this patch?

PR #2154, opened 2026-09-23
Do not tape in a forward pass with no reverse sweep

Vedant2005goyal:
A reverse_forw function has no reverse sweep, so pushing primals onto the tape is unnecessary. Instead, store the value in a local variable.


vgvassilev:
I am surprised that this does not need test updates.

